PR #250, opened 2014-12-30
contrib/sigtest.sh - testing signature

smoe:
The autoupdating does not work for me - no exact idea why this may be so. I have hence modified sign.sh to confirm that I was using the right secret key for the public key referenced in the site.conf.
<b>Update</b>: The autoupdate works now - very happy!

### DIFF
--- a/contrib/sigtest.sh
+++ b/contrib/sigtest.sh
@@ -1,0 +1,40 @@
+#!/bin/sh
+
+if [ $# -eq 0 -o "-h" = "$1" -o "-help" = "$1" -o "--help" = "$1" ]; then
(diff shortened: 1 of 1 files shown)


tcatm:
I'd suggest specifying the public key right on the commandline.

When testing the script I couldn't get it to work without moving `rm -f $upper $lower` to saner locations (right before the script exits).

NeoRaider:
Another thing (if it is still planned to get this merged): `contrib/sign.sh` has been fixed to work with paths containing spaces a while ago, this script should be fixed in the same way.

tcatm:
Indirectly merged by #500

PR #251, opened 2015-01-01
Explaining how to access running nodes

smoe:


tcatm:
Thanks! Could you 

- move it to a separate section (i.e. not in authorized-keys.rst, maybe user/accessing-nodes.rst),
- rewrite the examples from the perspective of a mobile device connected to the client WLAN (I think that's the more common case) and
- try to reduce references to batman (that's, at least in theory, exchangeable)?

rubo77:
finding out the ipv6 can be done by this also: http://wiki.freifunk.in-kiel.de/wiki/IPv6_broadcast

tcatm:
@smoe and I decided to move this to the freifunk wiki instead.

PR #255, opened 2015-01-03
Create inspect-network-traffic.rst

smoe:
This document provides general description on how to approach the network from gateways/routers/clients. This provides some material for the still ongoing discussion about how much how difficult is to gather personal information in the Freifunk and the Internet in general - shall transport some understanding on why end-to-end cryptography is so important. General network security may improve, too, with individuals learning to see/understand pathologic states of the network earlier.

### DIFF
--- a/docs/features/inspect-network-traffic.rst
+++ b/docs/features/inspect-network-traffic.rst
@@ -1,0 +1,164 @@
+Network Traffic
+===============
+
+In general, Freifunk does not filter network traffic. Users are not monitored.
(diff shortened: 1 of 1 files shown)


NeoRaider:
This information is not specific to Gluon (but rather to batman-adv based mesh networks), and Gluon is not only used for Freifunk anymore, but also scientific networks, so the Freifunk wiki is a much more appropriate place (but you can maybe add a link at http://wiki.freifunk.net/Freifunk_Firmware_Gluon , so Gluon users will find it)

PR #265, opened 2015-01-08
Add support for Netgear WNDR3700/3800

flokli:
basic wifi meshing with neighbouring nodes
tested, with TL-WDR4300 (both bands working)

client WLAN (if applicable)
works

client lan on LAN ports (if applicable)
works

mesh-vpn on WAN port
works

mesh-on-wan
Not tested

re-entering configmode by pressing the button for >3 seconds
works

entering failsafe mode
works

whether the MAC address seen by Gluon is the one printed on the device
yes

lua -e "print(require('platform_info').get_image_name())"
netgear-wndr3700v2

### DIFF
--- a/targets/ar71xx-generic/profiles.mk
+++ b/targets/ar71xx-generic/profiles.mk
@@ -134,3 +134,11 @@ $(eval $(call GluonModel,WZRHPG450H,wzr-hp-g450h-squashfs,buffalo-wzr-hp-g450h))
 # WZR-HP-AG300H/WZR-600DHP
 $(eval $(call GluonProfile,WZRHPAG300H))
 $(eval $(call GluonModel,WZRHPAG300H,wzr-hp-ag300h-squashfs,buffalo-wzr-hp-ag300h-wzr-600dhp))
+
+## Netgear
+
+# WNDR3700v2/3800
+$(eval $(call GluonProfile,WNDR3700))
+$(eval $(call GluonProfileFactorySuffix,WNDR3700,.img))
+$(eval $(call GluonModel,WNDR3700,wndr3700v2-squashfs,netgear-wndr3700v2))
+$(eval $(call GluonModel,WNDR3700,wndr3800-squashfs,netgear-wndr3800))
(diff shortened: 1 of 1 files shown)


CodeFetch:
There are several versions of the WNDR3700. Some of them have Wifi-chips that can't work with Gluon at all. Following versions could work:
WNDR3700 v1, v2, v4 and maybe (I don't know the hardware) v5. V4 can't work as it has a Broadcom Wifi chip that cannot create virtual access points.

Seems that you actually have tested a WNDR3800? There is only one version of this device.

flokli:
I tested with netgear-wndr3700v2.

Am 9. Januar 2015 17:22:07 MEZ, schrieb CodeFetch <notifications@github.com>:

flokli:
we have the netgear-wndr3700v2 running for one week so far, without any problems.

targets/ar71xx-generic/profiles.mk
```diff
@@ -134,3 +134,12 @@ $(eval $(call GluonModel,WZRHPG450H,wzr-hp-g450h-squashfs,buffalo-wzr-hp-g450h))
...
 $(eval $(call GluonProfile,WZRHPAG300H))
 $(eval $(call GluonModel,WZRHPAG300H,wzr-hp-ag300h-squashfs,buffalo-wzr-hp-ag300h-wzr-600dhp))
+
+## Netgear
+
+# WNDR3700v2/3800
+$(eval $(call GluonProfile,WNDR3700))
+$(eval $(call GluonProfileFactorySuffix,WNDR3700,.img))
+$(eval $(call GluonModel,WNDR3700,wndr3700-squashfs,netgear-wndr3700))
+$(eval $(call GluonModel,WNDR3700,wndr3700v2-squashfs,netgear-wndr3700v2))
```
NeoRaider: Please add a comment that the 3700v2 will advertise itself as a 3800, and that the auto updater will thus use the 3800 image (and that this is okay as the 3700v2 and 3800 sysupgrade images are identical)

NeoRaider: You can ignore this comment, I seems I was wrong about this.

PR #274, opened 2015-01-16
docs/site-example: cleanup

flokli:
- generalize, to make it more clear that the site config
   needs to be adapted to the local community.
 - remove some host names and public keys used in Lübeck
 - suggest private space mac adresses
 - change fastd method to salsa2012+umac, which is the recommended default
 - add config for a fastd peer reachable by both ipv4 and ipv6
 - order GLUON_SITE_PACKAGES alphabetically

### DIFF
--- a/docs/site-example/site.conf
+++ b/docs/site-example/site.conf
@@ -10,14 +10,14 @@
   hostname_prefix = 'freifunk',
 
   -- Name of the community.
-  site_name = 'Freifunk Lübeck',
+  site_name = 'Freifunk Entenhausen',
 
   -- Shorthand of the community.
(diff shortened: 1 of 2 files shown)


kokel:
`gluon-mesh-batman-adv-14` should be the first package in site.mk

DerDerwish:
Why is it everywhere "Entenhausen" and in in line 60 "musterhausen"?

flokli:
@kokel, @DerDerwish: fixed. Also fixed some typos in msg_pubkey.

Not quite sure if we should keep "the mail us your pubkey to keys@..." text or use something more general.

PR #286, opened 2015-01-25
add some missing commata

kokel:


NeoRaider:
In my opinion, these commas aren't missing, as all example blocks are only fragments of the site configuration. Each block for itself is a valid Lua expression, adding commas at the end would change that...

@tcatm, you opinion?

kokel:
I thought that these are all site.conf snippets and if someone will copy these snippets inside his own site.conf it will fail...

tcatm:
I don't think they were meant for copypasting. The example site.conf is included in the documentation now so those snippets aren't needed anymore.

Issue #226, opened 2014-11-05
Build fails on targets with no factory image


tcatm:
In case there is only a sysupgrade image the build will fail. This happens on pure OpenWrt targets (like VoCore) where only a sysupgrade image is needed.

NeoRaider:
Fixed by e93173c45cf4a48b4692ab6a23390f97f7b5832b.

PR #296, opened 2015-02-17
Add support for ALLNET ALL0315N

FRNK-BK:
Built, flashed and tested. SUCCESS.
This patch takes advantage of patch https://github.com/freifunk-gluon/gluon/commit/e93173c45cf4a48b4692ab6a23390f97f7b5832b which solved issue https://github.com/freifunk-gluon/gluon/issues/226 by introducing "GluonProfileFactorySuffix".


targets/ar71xx-generic/profiles.mk
```diff
@@ -169,3 +169,10 @@ $(eval $(call GluonProfileFactorySuffix,WNDR3700,.img))
 $(eval $(call GluonModel,WNDR3700,wndr3700-squashfs,netgear-wndr3700))
 $(eval $(call GluonModel,WNDR3700,wndr3700v2-squashfs,netgear-wndr3700v2))
 $(eval $(call GluonModel,WNDR3700,wndr3800-squashfs,netgear-wndr3800))
++
++## Allnet
++
++# ALL0315N
++$(eval $(call GluonProfile,ALL0315N,uboot-envtools,rssileds))
```
NeoRaider: The two packages need to be separated by a space, not a comma, otherwise rssileds will be ignored.

targets/ar71xx-generic/profiles.mk
```diff
@@ -169,3 +169,10 @@ $(eval $(call GluonProfileFactorySuffix,WNDR3700,.img))
...
++$(eval $(call GluonProfileFactorySuffix,ALL0315N,))
```
NeoRaider: Is it correct that there are no factory images for this device? Does it use the same images for factory and sysupgrade, or how does the initial installation work?

FRNK-BK: Initially the device comes with a pure OpenWRT based firmware. Sysupgrade can be flashed directly.

targets/ar71xx-generic/profiles.mk
FRNK-BK: So it will have been ignored in my build. I'll make a new one with space separation to check if it still works and what the leds are doing...

NeoRaider: Okay. By the way, there's no need to open a new pull request for fixups, you can just force-push over the branch the pull request was created from to update the request.

FRNK-BK: I made a new build with space as seperator and it (still) worked.

FRNK-BK:
This pull refers to original commit and to the seperator fixup.

NeoRaider:
Please squash the commits in to a single one.

FRNK-BK:
How do I do this via github website?

FRNK-BK:
It's getting worse. I'll start a new approach...

NeoRaider:
You should use the `git rebase -i` command (or just `git reset` and a new commit) ... the github website isn't really usable for non-trivial operations.

FRNK-BK:
Yes, I switched to command line and did "git rebase -i"... but I mixed up references, did some merges... and f**ked it up... So, here's the new pull request: https://github.com/freifunk-gluon/gluon/pull/297

NeoRaider:
As I wrote before, it is never necessary to open a new pull request. You can just completely replace a request by force-pushing over the old one...

PR #297, opened 2015-02-18
Add support for ALLNET ALL0315N

FRNK-BK:
Built, flashed and successully tested. This patch takes advantage of patch freifunk-gluon@e93173c which solved issue freifunk-gluon#226 by introducing "GluonProfileFactorySuffix". Packages are separated by a space, not a comma. The ALL0315N initially has a OpenWRT based firmware and there is no need for a factory image. The sysupgrade can be flashed instantly.


NeoRaider:
Thanks, pulled.

PR #307, opened 2015-03-07
Add support for GL-iNet devicAdd support for GL-iNet devicee

MaStr:
Adds support for GL-iNet devices.

http://wiki.openwrt.org/toh/gl-inet/gl-inet

With BB only 8MB firmware files, but they can be flashed on the newer 16MB devices without problems.
Can be bought here: https://www.gl-inet.nl/ and on amazon

### DIFF
--- a/targets/ar71xx-generic/profiles.mk
+++ b/targets/ar71xx-generic/profiles.mk
@@ -4,7 +4,7 @@
 
 # GL-iNet 1.0 ; BB is the 8MB version, works on 16M too
 $(eval $(call GluonProfile,GLINET))
-$(eval $(call GluonModel,GLINET,gl-inet-v1-squashfs,gl-inet-v1.0))
+$(eval $(call GluonModel,GLINET,gl-inet-v1-squashfs,gl-inet-v1))
 
 
 ## TP-Link
--- a/targets/ar71xx-generic/profiles.mk
+++ b/targets/ar71xx-generic/profiles.mk
@@ -1,5 +1,12 @@
 # List of hardware profiles
 
+## GL-iNet
+
+# GL-iNet 1.0 ; BB is the 8MB version, works on 16M too
+$(eval $(call GluonProfile,GLINET))
+$(eval $(call GluonModel,GLINET,gl-inet-v1-squashfs,gl-inet-v1.0))
+
+
 ## TP-Link
 
 # CPE210/220/510/520


targets/ar71xx-generic/profiles.mk
```diff
@@ -1,5 +1,12 @@
 # List of hardware profiles
 
+## GL-iNet
+
+# GL-iNet 1.0 ; BB is the 8MB version, works on 16M too
+$(eval $(call GluonProfile,GLINET))
+$(eval $(call GluonModel,GLINET,gl-inet-v1-squashfs,gl-inet-v1.0))
```
NeoRaider: The image filename `gl-inet-v1.0` looks wrong. As far as I can tell the correct name is `gl-connect-inet-v1` (the name is essential for the autoupdater).

targets/ar71xx-generic/profiles.mk
```diff
@@ -1,5 +1,12 @@
 # List of hardware profiles
 
+## GL-iNet
+
+# GL-iNet 1.0 ; BB is the 8MB version, works on 16M too
```
NeoRaider: I'd like to backport the GL-iNET changes from CC (as it should be done anyways when we upgrade to CC) instead of using the 8M image for both versions.

Currently OpenWrt is missing code to disinguish the versions for the autoupdater, I'll whip up a patch to achieve that.

MaStr:
I can change the name to gl-inet-v1 , if you want to. I tried to stick close this one

```` $(eval $(call GluonModel,CPE510,cpe210-220-510-520-squashfs,tp-link-cpe220-v1.0)) ````
This is no issue.

On PirateBox we are using a backport from CC with the 16MB, too (but we are using AA source base). 

I'm unsure I can backport it, because I don't have much experience with that.

NeoRaider:
No problem, I'll do the backport.

Please refer to http://gluon.readthedocs.org/en/latest/dev/hardware.html#adding-profiles for the image naming.

MaStr:
yes, I read that part, but not briefly enough :( 
Better with that commit?

NeoRaider:
@MaStr, can you check what the following command outputs on the 16MB version?

    source /lib/functions/system.sh; mtd_get_part_size firmware

MaStr:
````
root@OpenWrt:/# source /lib/functions/system.sh; mtd_get_part_size firmware
16580608
````

best regards Matthias

NeoRaider:
I've pushed support for the GL-inet, together with a patch to detect the correct model which isn't upstream yet. It has to be built with `BROKEN=1` for now. Does

now print the correct string `gl-inet-6416a-v1`? If it does, I'll remove the BROKEN check and send the patch upstream.

MaStr:
I'm not quite sure, what you want me to do.
Should I build the patched version and run the command on the device, or should I use the trunk and run the command from above?

regards Matthias

NeoRaider:
Build the current Gluon master with `BROKEN=1` and run the command with the generated Gluon image.

MaStr:
Built "ab3322079c26576ca25a80d90ff9d8fb9e92365d" with BROKEN=1 and site configuration of FreiFunk Wiesbaden. 

````
root@New-Horizons:~# lua -e 'print(require("platform_info").get_image_name())'
gl-inet-6416a-v1
````

Best regards Matthias

NeoRaider:
Okay, I've removed the BROKEN flag and sent the model detection patch upstream.

MaStr:
Thanks for adding the device.

Issue #294, opened 2015-02-13
Wrong model name on WNDR3700v2


aeickho:
On NETGEAR WNDR3700v2

The model name is  **NETGEAR NDR3700v2** and not **NETGEAR WNDR3700v2**
(cat /tmp/sysinfo/model)

this has an affect for autoupdate

tmp/sysinfo/model on (v1) is ok NETGEAR WNDR3700

348a819711fe6594bce5bc3703900871f4e2b6d3

NeoRaider:
What is the output of `cat /proc/cpuinfo` on this device?

aeickho:
root@alex_test-10:~# cat /proc/cpuinfo
system type             : Atheros AR7161 rev 2
machine                 : NETGEAR WNDR3700/WNDR3800/WNDRMAC
processor               : 0
cpu model               : MIPS 24Kc V7.4
BogoMIPS                : 452.19
wait instruction        : yes
microsecond timers      : yes
tlb_entries             : 16
extra interrupt vector  : yes
hardware watchpoint     : yes, count: 4, address/irw mask: [0x0000, 0x0ff8, 0x0ff8, 0x0ff8]
isa                     : mips1 mips2 mips32r1 mips32r2
ASEs implemented        : mips16
shadow register sets    : 1
kscratch registers      : 0
core                    : 0
VCED exceptions         : not available
VCEI exceptions         : not available

aeickho:
root@alex_test-10:~# cat /tmp/sysinfo/model 
NETGEAR NDR3700v2

NeoRaider:
Okay, could you provide a dump of your ART partition? You can find out which partition the ART is in `/proc/mtd` and then just dump it via e.g. `cat /dev/mtd3 > /tmp/art.dump`.

aeickho:
root@alex_test-10:~# cat /proc/mtd 
dev:    size   erasesize  name
mtd0: 00050000 00010000 "u-boot"
mtd1: 00020000 00010000 "u-boot-env"
mtd2: 00f80000 00010000 "firmware"
mtd3: 00112440 00010000 "kernel"
mtd4: 00e6dbc0 00010000 "rootfs"
mtd5: 00c00000 00010000 "rootfs_data"
mtd6: 00010000 00010000 "art"
root@alex_test-10:~# cat /dev/mtd6 > /tmp/art.dump

    https://www2.artbanause.de/art.dump

NeoRaider:
[...]

PR #309, opened 2015-03-10
Art for WNDR3700v2 type 2, patch

aeickho:
issues #294 temporary patch to build for WNDR3700v2 type 2 :) 
i'm not sure, if patch format is ok


NeoRaider:
I'd prefer to fix this properly. Unfortunately, I haven't been able to aquire an ART dump of an old NETGEAR device which doesn't contain the full model name in the ART, but I'll write a fix in the next few days, based on the information I have.

PR #316, opened 2015-03-24
add subject and body in default-mailto link

rubo77:
Add a default subject and body that is transferred to your mail client as default text, so the user has to type less when creating the mail to send the fastd-key to the community mailinglist.   
I am not sure if the `<%=key?>` part works, maybe this has to be adapted, so the key is in the link

### DIFF
--- a/docs/site-example/i18n/de.po
+++ b/docs/site-example/i18n/de.po
@@ -23,7 +23,7 @@ msgstr ""
 "kann sich dein Knoten mit dem Entenhausener Mesh-VPN zu verbinden. Bitte "
 "schicke dazu diesen Schlüssel und den Namen deines Knotens "
 "(<em><%=hostname%></em>) an "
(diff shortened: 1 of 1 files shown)


mweinelt:
I have built the same thing and tested it successfully.
```
<a href=\"mailto:keys@darmstadt.freifunk.net?subject=Anmeldung:%20<%=hostname%>&amp;body=%23%20<%=hostname%>%0A%23%20<%=sysconfig.primary_mac%>%0Akey%20%22<"%=pubkey%>%22%3B\">keys@darmstadt.freifunk.net</a>
```
(https://github.com/freifunk-darmstadt/site-ffda/blob/master/i18n/de.po#L24)

which results in the quite neutral
```
# nodename
# node primary mac
key "node-public-key";
```

in the mail body, which can be directly copied to a fastd hosts file.

tcatm:
We'd like to keep the example config short and readable. User's are free to change this in their communities as needed.

PR #317, opened 2015-03-24
docs: site-example: change default batman-adv version to 15

deavmi:
...refore we should change the example.

neoraider suggested that I change my own config to batman-adv-15, so therefore we should change the example.

he said that batman-adv-15 is stable enough now, so why not go with something more stable and newer. :D


NeoRaider:
I'd recommend changing the commit site to something like

    docs: site-example: change default batman-adv version to 15

deavmi:
@NeoRaider Okay, thanks :smile:

NeoRaider:
You changed the pull request title, not the commit title...

PR #319, opened 2015-03-26
fix: Fix handling of filenames with spaces

BenBE:
When filenames contain spaces this might cause problems to create a proper signature.


NeoRaider:
As discussed, I've applied the patch with fixed commit message.

PR #323, opened 2015-04-02
Create 0003-add-config-parsing-for-no_rebroadcast-option.patch

RubenKelevra:


NeoRaider:
I've added the fix myself a bit differently, as I didn't want to add a new patch file. (Also, you broke the patch file with your second commit as you didn't adjust the patch header...)

PR #324, opened 2015-04-02
Create 0005-reset-hop_penalty-to-previous-rating.patch

RubenKelevra:
Batman-adv's default hop_penalty was changed from 15 to 30, which makes some large meshes with many hops unusable, because the TQ for the gateway drops below zero.

This patch restores the old behaviour which is well tested.


NeoRaider:
The hop penalty can be nicely configured in `/etc/config/batman-adv`, no need to patch the code for that.

NeoRaider:
I think adding an option for the hop penalty to `site.conf` would be approprate.

PR #330, opened 2015-04-04
ar71xx: Meraki MR12/MR62 support

tcatm:


patches/openwrt/0032-ar71xx-support-for-Meraki-MR12-and-MR62.patch
```diff
@@ -0,0 +1,306 @@
...
+@@ -158,6 +158,14 @@ hornet-ub)
+ 	ucidef_set_led_usbdev "usb" "USB" "alfa:blue:usb" "1-1"
+ 	;;
+ 
++mr12)
++	ucidef_set_led_netdev "wan" "WAN" "mr12:green:wan" "eth0"
++	ucidef_set_led_wlan "wlan1" "WLAN1" "mr12:green:wifi1" "phy0assoc"
++	ucidef_set_led_wlan "wlan2" "WLAN2" "mr12:green:wifi2" "phy0assoc"
++	ucidef_set_led_wlan "wlan3" "WLAN3" "mr12:green:wifi3" "phy0assoc"
++	ucidef_set_led_wlan "wlan4" "WLAN4" "mr12:green:wifi4" "phy0tpt"
```
NeoRaider: I think this should be changed to use rssileds like some other models with multiple WLAN LEDs do.

targets/ar71xx-generic/profiles.mk
```diff
@@ -199,3 +199,10 @@ $(eval $(call GluonModel,ALL0315N,all0315n,allnet-all0315n))
 $(eval $(call GluonProfile,GLINET))
 $(eval $(call GluonModel,GLINET,gl-inet-6408A-v1,gl-inet-6408a-v1))
 $(eval $(call GluonModel,GLINET,gl-inet-6416A-v1,gl-inet-6416a-v1))
+
+## Meraki
+
+# Meraki MR12/MR62
+$(eval $(call GluonProfile,MR12))
+$(eval $(call GluonProfileFactorySuffix,MR12))
```
NeoRaider: Hmm, we should really try to provide an image which can be flashed over TFTP if possible. As the sysupgrade image is just flashed linearly by OpenWrt, have you tried just flashing it in one part through the bootloader?

tcatm:
I'm going to rebase this on top of the patches that have been merged into OpenWrt (acdf1a2592a14e11aab99996133c836dbef1f92a).

PR #337, opened 2015-04-14
fix syntax of setup_mode

RubenKelevra:


NeoRaider:
Thanks.

Please always include a prefix like "docs: " in your commit messages, so you don't need to look at the changed files to understand that it only affects the documentation. I've fixed up the message for you this time.

RubenKelevra:
thanks for your hint. :)

PR #345, opened 2015-04-25
make wireless-disable state survive updates and allow site option

do9xe:
This Option allows the "disabled"-option in "/etc/config/wireless" to survive updates or specify it in the site.conf.

The new code looks, weather the network is disabled, if not it looks for the site-option. In case none of those options is set it enables the network.

The site-options have to look like this, both options can be true or false.
```
client_disabled = true,
mesh_disabled = false,
```

### DIFF
--- a/package/gluon-mesh-batman-adv-core/files/lib/gluon/upgrade/320-gluon-mesh-batman-adv-core-wireless
+++ b/package/gluon-mesh-batman-adv-core/files/lib/gluon/upgrade/320-gluon-mesh-batman-adv-core-wireless
@@ -16,6 +16,21 @@ local function configure_radio(radio, index, config)
   local client = 'client_' .. radio
   local mesh = 'mesh_' .. radio
 
+  local disable_state_client = false
+  local disable_state_mesh = false
+
+  if uci:get('wireless', client) then
+    disable_state_client = uci:get_bool('wireless', client, "disabled")
+  elseif config.client_disabled then
+    disable_state_client = true
+  end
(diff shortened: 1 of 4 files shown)


package/gluon-mesh-batman-adv-core/files/lib/gluon/upgrade/320-gluon-mesh-batman-adv-core-wireless
```diff
@@ -15,6 +15,21 @@ local function configure_radio(radio, index, config)
 
   local client = 'client_' .. radio
   local mesh = 'mesh_' .. radio
+  
+  local disable_state_client = false
+  local disable_state_mesh = false
+
+  if (uci:get_bool('wireless', client, "disabled") then
```
NeoRaider: There is an opening ( too much.

package/gluon-mesh-batman-adv-core/files/lib/gluon/upgrade/320-gluon-mesh-batman-adv-core-wireless
```diff
@@ -15,6 +15,21 @@ local function configure_radio(radio, index, config)
...
+  	disable_state_client = true
+  else if config.client_disabled
+  	disable_state_client = config.client_disabled
+  end
+
+	if (uci:get_bool('wireless', mesh, "disabled") then
```
NeoRaider: Broken indentation

PR #352, opened 2015-04-26
nodeinfo: mesh interface classes (wireless, tunnel, other)

tcatm:
This will introduce a new nodeinfo object, network.mesh.bat0.interfaces,
containing any of the the following subordinated objects:

- wireless
- tunnel
- other

Each of these objects contains a (possibly empty) list of MAC addresses
(lowercase, colon-notation) corresponding to a interface of the given
class. Combined with a batman graph it is thus possible to mark
sub-graphs as "wireless" or "vpn".

The previously used object mesh_intefaces is superseded by this new
object structure and mesh_interfaces will be removed in a future Gluon
release.

### DIFF
--- a/package/gluon-mesh-batman-adv-core/files/lib/gluon/announce/nodeinfo.d/network/mesh/bat0/interfaces
+++ b/package/gluon-mesh-batman-adv-core/files/lib/gluon/announce/nodeinfo.d/network/mesh/bat0/interfaces
@@ -1,0 +1,52 @@
+local list = util.exec('batctl if')
+
+local wireless = {}
+local tunnel = {}
+local other = {}
+
+local function get_address(t, ifname)
+  pcall(
+    function()
+      table.insert(t, util.trim(fs.readfile('/sys/class/net/' .. ifname .. '/address')))
+    end
+  )
+end
+
+local function is_wireless(ifname)
+  local type = fs.stat('/sys/class/net/' .. ifname .. '/wireless', 'type')
+
+  return type == 'directory'
+end
+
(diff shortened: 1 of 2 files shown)


package/gluon-mesh-batman-adv-core/files/lib/gluon/announce/nodeinfo.d/network/mesh/bat0/interfaces
```diff
@@ -0,0 +1,52 @@
...
+  return type == 'regular'
+end
+
+function nil_table(t)
+  if next(t) ~= nil then
+    return t
+  else
+    return nil
+  end
+end
```
NeoRaider: All these functions should be local

package/gluon-status-page/files/lib/gluon/status-page/www/cgi-bin/status
```diff
@@ -22,9 +22,30 @@ function neighbours(ifname)
...
+          if interfaces["wireless"] then
+            for _, mac in ipairs(interfaces["wireless"]) do
+              macs[mac] = data
+            end
+          end
+          if interfaces["tunnel"] then
+            for _, mac in ipairs(interfaces["tunnel"]) do
+              macs[mac] = data
+            end
+          end
```
NeoRaider: This could be shortened a bit by adding a helper where `data` is in scope, as there are four very similar pieces of code:

    local function add_macs(list)
      if list then
        for _, mac in ipairs(list) do
          macs[mac] = data
        end
      end
    end

PR #353, opened 2015-04-26
gluon-luci-node-role: add translations

kokel:
Internationalization for gluon package ``gluon-luci-node-role``.


package/gluon-luci-node-role/files/usr/lib/lua/luci/controller/admin/noderole.lua
```diff
@@ -1,5 +1,5 @@
 module("luci.controller.admin.noderole", package.seeall)
 
 function index()
-	entry({"admin", "noderole"}, cbi("admin/noderole"), "Verwendungszweck", 20)
+	entry({"admin", "noderole"}, cbi("admin/noderole"), "Purpose", 20)
```
NeoRaider: Hmm, I don't really like the term "purpose" here. Can we just keep calling this "Node role"?

package/gluon-luci-node-role/files/usr/lib/lua/luci/model/cbi/admin/noderole.lua
```diff
@@ -1,23 +1,23 @@
...
 local site = require 'gluon.site_config'
+local i18n = require "luci.i18n"
 local uci = luci.model.uci.cursor()
 local config = 'gluon-node-info'
 
 -- where to read the configuration from
 local role = uci:get(config, uci:get_first(config, "system"), "role")
 
-f = SimpleForm("role", "Verwendungszweck")
+f = SimpleForm("role", i18n.translate("Node Role"))
```
NeoRaider: The R should be an r in "Node Role".

package/gluon-luci-node-role/i18n/de.po
```diff
@@ -0,0 +1,30 @@
...
+
+msgid "Role"
+msgstr "Rolle"
+
+msgid ""
+"If this node has a special role within the freifunk network you can specify "
+"this role here. Please find out about the available roles and their impact "
+"first. Only change the role if you know what you are doing."
+msgstr ""
+"Wenn dein Freifunk-Router eine besondere Rolle im Freifunk Netz einnimmt, "
```
NeoRaider: War zwar vorher schon so, aber trotzdem auch hier eine Anmerkung: Wir sollten immer von Freifunk-Knoten, nicht Freifunk-Routern sprechen, damit es einheitlich ist.

package/gluon-luci-node-role/i18n/de.po
```diff
@@ -0,0 +1,30 @@
...
+"kannst du diese hier angeben. Bringe bitte zuvor in Erfahrung welche "
```
NeoRaider: Erfahrung, welche

package/gluon-luci-node-role/i18n/de.po
```diff
@@ -0,0 +1,30 @@
...
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
+
+msgid "Purpose"
+msgstr "Verwendungszweck"
+
+msgid "Node Role"
+msgstr "Knoten Rolle"
```
NeoRaider: As I suggested to change both "Purpose" and "Node Role" to "Node role", these two would become one. I think we can go with "Verwendungszweck". If we want to keep the translation closer to the English version, this should be "Knotenrolle" or "Knoten-Rolle", not "Knoten Rolle".

package/gluon-luci-node-role/files/usr/lib/lua/luci/model/cbi/admin/noderole.lua
```diff
@@ -1,23 +1,23 @@
...
 f.reset = false
 f.template = "admin/expertmode"
-f.submit = "Fertig"
+f.submit = i18n.translate("Finish")
```
NeoRaider: You can just leave `f.submit` unset, then it will default to "Submit" and "Absenden". I think it makes most sense to go with LuCI's defaults here.

You should also remove the line `f.reset = false` to match the other expert mode modules, which all have the reset button.

kokel:
Thanks for the audit. I have implemented all your suggestions.

package/gluon-luci-node-role/files/usr/lib/lua/luci/model/cbi/admin/noderole.lua
```diff
@@ -1,23 +1,22 @@
...
 
 -- where to read the configuration from
 local role = uci:get(config, uci:get_first(config, "system"), "role")
 
-f = SimpleForm("role", "Verwendungszweck")
-f.reset = false
+f = SimpleForm("role", i18n.translate("Node role"))
 f.template = "admin/expertmode"
-f.submit = "Fertig"
+f.submit
```
NeoRaider: The f.submit line should be removed completely. I don't even think this is correct Lua...

kokel:
Now the ``f.submit`` line is removed ...

NeoRaider:
Another thing I noticed is that this doesn't allow translating the role names. So maybe the configuration could be changed a bit to add support for that:

    roles = {
        default = 'foo',
        list = {
            'foo',
            'bar',
        }
    }

The translated strings could then be looked up from `site/i18n/*.po` as `gluon-node-role:role:foo` and `gluon-node-role-role:bar`.

kokel:
Yeah, would be nice, actually. But I would need some time to implement this because i18n and lua aren't my friends, yet. Or do you have a hint for me which could quicken my work?

NeoRaider:
If you want, we can merge the changes like this, and I'll make the role names translatable.

kokel:
I agree, so feel free to merge ...

NeoRaider:
Okay, everything merged, role display strings made translatable, documentation updated.

PR #356, opened 2015-04-27
gluon-mesh-vpn-fastd: extend site configuration to allow multiple and nested peer groups

### LINKED ISSUE #354: fastd-mesh-vpn: nested peer groups
The peer specification format in the `site.conf` should be extended to allow nested peer groups with different peer limits.

NeoRaider:
This extends the `fastd_mesh_vpn` block in the `site.conf` to allow configuration of multiple and nested peer groups. This requires moving the `backbone` table into a new table `groups` in existing site configurations. The name `backbone` isn't a keyword anymore, any other group name might be used as well.

Example:

    fastd_mesh_vpn = {
      methods = {'salsa2012+umac'},
      mtu = 1426,
      groups = {
        backbone = {
          limit = 2,
          peers = { ... },
          groups = { ... subgroups ... },
        }
      }
    }

Fixes #354

### DIFF
--- a/docs/site-example/site.conf
+++ b/docs/site-example/site.conf
@@ -90,24 +90,38 @@
     -- List of crypto-methods to use.
     methods = {'salsa2012+umac'},
     mtu = 1426,
-    backbone = {
-      -- Limit number of connected peers to reduce bandwidth.
-      limit = 2,
+    groups = {
+      backbone = {
+        -- Limit number of connected peers to reduce bandwidth.
+        limit = 2,
 
-      -- List of peers.
-      peers = {
-        peer1 = {
-          key = 'xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx',
+        -- List of peers.
+        peers = {
+          peer1 = {
+            key = 'xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx',
 
-          -- This is a list, so you might add multiple entries.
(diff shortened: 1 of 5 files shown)


viisauksena:
what exactly does it mean when i have something like
group { first { somthg - limit=1 peers=somethg} second {somethg - limit=1 peers=somethg}
a Node in Freifunk will then take 1 peer out of every group? and if it looses conn. to one of them it try another one of these group?
just that i get i right .. thx

NeoRaider:
@viisauksena, yes, that's correct. In addition, nested groups are supported, which can be useful to apply more complex peer limits.

viisauksena:
i have an serious issue here .. when i do the reordering (you can see it here) 
i got real trouble while the old fastd config seems to be merged within a normal sysupgrade (and thats i guess the same with autoupdate (keeping config)) 
but that leads in 2 times the same fastd key in one config - which make fastd disable both - making the gateway turn off 8from the perspective of the node)
any idea?

mmalte:
There is a workaround for this nasty behaviour:

https://github.com/freifunk-gluon/gluon/issues/557

You have to keep the old peer until this is fixed:
https://github.com/ffac/site/commit/246cff3134a52e67e85ce5f4126112cff16f875d

A-Kasper:
I guess you've created a new site.conf and changed naming or hierachie part
in fastd.
1. you should just connect to one peer at a time. this masslivy reduces load
2. you should add the old names of the peers in the site.conf and clear
them out. then you shouldn't have any more duplicates

2016-01-16 12:49 GMT+01:00 viisauksena <notifications@github.com>:

NeoRaider:
More to the point: Gluon won't touch "unknown" groups on updates, but all known groups are replaced completely. As the old group was called "backbone", your new config must contain a group called "backbone" as well to replace old peers; this group "backbone" may be empty to just remove all peers of this group.

PR #361, opened 2015-05-01
gluon-mesh-batman-adv-core: fix the data types in the check_site.lua

do9xe:
As we changed the ```check_site.lua``` I forgot to change the data types of the ```mesh_disabled``` and the ```client_disabled``` lines. This mistake stopped gluon from compiling, when using this options. This PR fixes the data types to boolean.


NeoRaider:
Another thing I overlooked before merging: You *really* need to learn to write correct commit messages. This one doesn't start with the `gluon-mesh-batman-adv-core:` prefix (the pull request title does, but the commit message is much more important than the pull request title). Someone looking through the commit log should understand what each commit does by looking at the messages only.

PR #368, opened 2015-05-12
batman-adv / alfred: update version to 2015.0

RubenKelevra:


NeoRaider:
As we're close to the release of Gluon 2015.1, we don't accept any non-{bugfix,documentation} commits anymore, so this will have to wait till after the release.

RubenKelevra:
Well, sure, you can merge it after this release. Were running a bit out of sync with your releases, so we gonna try batman-adv 2015.0 in our networks and provide a feedback.

2014.4 seem to have only one bug at the moment, were arp w/ and w/o the "dat" seem to have problems on our gateways, when they run for a longer time (over 4 weeks).

So we try out 2015.0 on our gateways atm, and having a this update for the nodes in queue.

ohrensessel:
how do you update nodes from batman 14 to 15?
do you have a specialized autoupdater?

NeoRaider:
We've updated to the current openwrt-routing master a while ago, which includes batman-adv 2015.0.

Issue #322, opened 2015-03-31
Weird bridge behaviour


NeoRaider:
We've had multiple reports of weird bridge behaviour with Gluon v2014.4, for example:
* `br-wan` stops responding to neighbour solicitations occasionally
* `br-wan` doesn't forward DHCP correctly when the LAN ports are moved from `br-lan` to `br-wan` occasionally

The neighbour discovery issue might be caused by multicast issues, but that doesn't really explain the DHCP problem. Another possible cause might be be netfilter/conntrack bugs.

T-X:
I can confirm this issue and it's quite easy to reproduce:

Take a recent stable gluon version. Look up the ipv6 link-local address on br-wan (ip -6 a s dev br-wan). Try to ping6 that address from a host connected to the WAN port - it will fail.

On a 4.0.0-rc3 kernel I couldn't reproduce that issue. The following things are workarounds on gluon after which IPv6 seems to work again (any one works):

* echo 0 > /sys/class/net/br-wan/bridge/multicast_snooping
* ifconfig br-wan promisc
* /etc/init.d/firewall stop and wait a few seconds (probably up to 125 seconds until the next MLD report)

Especially as the last one is a valid workaround I suspect your last assumption might be the right one. So that netfilter is preventing multicast reports to reach the bridge kernel module.

NeoRaider:
Okay, the first one seems like the most workable solution.

@T-X, did you also have a look at the DHCP issue, which shouldn't involve multicast at all, but only broadcast?

T-X:
Just noticed, that there's never been a gluon release with this commit yet:

"kernel: bridge: multicast: backport a few more fixes for 3.10"
(http://git.openwrt.org/?p=14.07/openwrt.git;a=commit;h=60abf1305e4c8cc8c5673dfd98b214e9ceab8ca5)

I'm going to ask THiSCO whether the issue still exists for him with a recent gluon-master build.

thisco-de:
@T-X I first noticed the problem with a Gluon build from master built some time around the beginning or mid of March and then reproduced it with 2014.4. As the patch you mention dates back to January, it should thus be included in the build (at least the patch files exist in my build-dir).

[...]

Issue #278, opened 2015-01-19
Enable bridge multicast snooping


T-X:
This would for one thing spare us a bunch of multicast packets with no listener in the mesh in general. Secondly, it'll spare us some more multicast packets to be broadcasted on the slow, 1MBit/s wifi AP interfaces.

~~To do:~~
- ~~Update to recent barrier breaker (see & close: #191)~~
- ~~Enable bridge multicast snooping again (while leaving the broken multicast-to-unicast feature disabled)~~
- ~~Increase "/sys/class/net/$br0/bridge/hash_max" (and maybe hash_elasticity, too? maybe hash_max = 2048 and hash_elasticity = 8?)~~

UPDATE: _Strategy change:_ Instead of increasing hash_max, let's segment the IGMP/MLD broadcast domains with ip(6)tables: Filter IGMP/MLD packets from and to bat0 and set /sys/class/net/bat0/brport/multicast_router to 1.

The advantage of this is that we'd spare IGMP/MLD overhead over the VPN. And every node could/would run a querier on br-client, making multicast snooping work in a decentral way without relying on the gateways (where we'd have otherwise had to place the querier).

New ToDos:
- Fix multicast-to-unicast configuration in OpenWRT (e.g. enable ap-isolation + bridge-hairpinning). See: https://lists.openwrt.org/pipermail/openwrt-devel/2015-June/033398.html
- Implement filter rules (also set /proc/sys/net/bridge/bridge-nf-call-ip{6,}tables)
- Set multicast_router to 1 for bat0
- Reenable multicast-snooping, including querier

(Optional/later: Implement Multicast-Router-Discovery in the bridge and send MRD-RAs on AP interface and LAN-Ports to 100% support scenarios with multicast-snooping switches / multicast routers behind the LAN-Ports, too)

NeoRaider:
I've submitted this patch for netifd: https://lists.openwrt.org/pipermail/openwrt-devel/2015-January/030864.html

T-X:
Hm, is the intention to have the querier enabled on all nodes? That won't work nicely, otherwise there'd be one poor node chosen as the selected querier which would get taken down by multicast traffic. In my opinion the querier should be enabled on gateways only.

Instead it would be nice to have hash_max configurable and setting it to 8192. Each hash entry should take less than 100KB, so it'd take about 800KB max. I thought about hash_elasticity again and I think we can leave that one unchanged.

[...]

PR #369, opened 2015-05-12
batman-adv: enable multicast optimisations per default

RubenKelevra:


T-X:
Hi @RubenKelevra,

On the one hand, technically, this shouldn't make any difference yet. On the other hand, it might be a good idea to have the batman-adv default as the default in Gluon, too. So that we could get there step by step and wouldn't need to hunt for a bug after enabling all that at once.

Also see #322 and #278 to see what else needs to be done to make this work (next to actually implement batman-adv multicast optimizations for bridged scenarios, too).

NeoRaider:
Closing, see #278.

RubenKelevra:
Thanks for pointing this out, we going to try it anyway on 2014.4.x-Branch

PR #370, opened 2015-05-12
batman-adv: reset hop_penalty to pre "batman-adv 2014.4"-rating

RubenKelevra:
The higher ratings make large meshes unreliable and might confuse map-applications which calculate with the TQ to gateway.

### DIFF
--- a/patches/packages/routing/0004-reset-hop_penalty-to-previous-rating.patch
+++ b/patches/packages/routing/0004-reset-hop_penalty-to-previous-rating.patch
@@ -1,0 +1,43 @@
+From 431594c451d853353d035df47a36657a7d707ef7 Mon Sep 17 00:00:00 2001
+From: RubenKelevra <ruben@freifunk-nrw.de>
+Date: Fri, 3 Apr 2015 13:17:27 +0200
+Subject: [PATCH] add patch which resets hop_penalty to previous rating
+
+---
+ ...enalty-to-previous-rating-which-is-well-t.patch | 25 ++++++++++++++++++++++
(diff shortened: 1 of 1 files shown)


NeoRaider:
This setting can be made in `/etc/config/batman-adv`, it doesn't make any sense to patch it in the module code...

I agree though that we should use the same hop penalty for batman-adv compat 14 and 15. @T-X, do you have an opinion about the value to use? I guess going back to the value 15 would be the safest.

T-X:
@NeoRaider: I'm also for Ruben's suggestion to set it to 15 (= 30 effectively for single radio devices) for now. Haven't done any thorough measurements yet, but so far it seemed to me that TQ values are getting too low / out of horizon with batman-adv 2013.4, so not quite sure about the new value yet, too.

RubenKelevra:
Well, the only reason why it was changed in upstream was, that commercial networks with dualband-wifis on a large setup have to long routes calculated. So they change it to cut to long routes in this particular setup.

I think patching the definition in the source code is the safest way to be backward-compatible. Beyond that, sure, it's nice to make this configurable in /etc/config as well.

NeoRaider:
Ensuring that the value is always set in /etc/config is enough, we don't patch things when we can avoid it.

So, either we could set it to 15 unconditionally, or make it configurable in `site.conf`, with a default of 15 if unset.

NeoRaider:
I've set the value to 15 now, making this configurable could be a task for a future Gluon version (as we're very close to the Gluon 2015.1 release)

PR #371, opened 2015-05-12
gluon-status-page: API

tcatm:
This adds the API used for the new status page. It doesn't yet add the frontend but it will allow the frontend running on other nodes to access data from this node. The total number of concurrent connections is limited to 12, limiting the number of users to about 2 or 3 (~4 connections per user).

Later versions of the should probably convert scripts to pure C to further reduce load.

### DIFF
--- a/package/gluon-status-page/files/lib/gluon/status-page/www/cgi-bin/dyn/neighbours-nodeinfo
+++ b/package/gluon-status-page/files/lib/gluon/status-page/www/cgi-bin/dyn/neighbours-nodeinfo
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+echo 'Access-Control-Allow-Origin: *'
+
+batctl if | cut -d: -f1 | grep -qxF "$QUERY_STRING" || exit 1
+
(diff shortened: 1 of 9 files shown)


package/gluon-status-page/files/lib/gluon/status-page/www/cgi-bin/dyn/neighbours-nodeinfo
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+echo 'Access-Control-Allow-Origin: *'
+
+batctl if | grep -q "^$QUERY_STRING:"  || exit 1
```
NeoRaider: Using the query string as a grep argument without `-F` seems like very bad idea.

tcatm: I'm going to replace that line with

    batctl if | cut -d: -f1 | grep -qxF "$QUERY_STRING" || exit 1

with a force push.

PR #378, opened 2015-05-17
[RFC] add initial Peer-to-peer VPN support

ohrensessel:
This pull request is not yet ready for merging, but somehow a request for comments.

What I want to achieve here: Allow node owners to connect mesh clouds, which do not have a possibility to mesh because of distance, via a Peer-to-peer VPN directly, without taking the way through a gateway. By that we would introduce a new possibility for decentralization.

Why manual configuration: Any form of automation would introduce a (possibly centralized) list of hostnames/IPs of DSL connections of the node owners participating in P2P VPN somewhere, which we do not want.

Open problems:
- How are P2P VPN connections displayed on the map?
- At the moment Mesh VPN has to be enabled to use P2P VPN.
- ...

Please comment and give recommendations.


kokel:
@ohrensessel Quite cool idea!
I would suggest to call this feature not P2P because it can confuse a lot of people thinking this belongs to torrent stuff e.g.
While we are everytime talking about nodes: N2N - node-to-node VPN.

ohrensessel:
@kokel mh yes maybe you are right regarding the naming. thinking about it :)

NeoRaider:
Nice work. I had such a fastd connection a few years ago, when the Freifunk Lübeck network was still small and Gluon didn't exist yet...

Note that currently the upgrade script won't touch any peer groups that aren't defined in the `site.conf` (nor toplevel peers without a enclosing group), so it is already possible for users to add peers in addition to the predefined ones. It would probably make sense to integrate this into the `gluon-mesh-vpn-fastd` package and just add a way to disable the preconfigured peers/peer groups.

The LuCI part could then be added to the package `gluon-luci-mesh-vpn-fastd`.

ohrensessel:
thanks @NeoRaider for your feedback.

can you elaborate your idea a little further? the LuCi part I wrote should be turned in a generall fastd config page in the expert mode, where node owners have the possibility to add P2P peers (in a special group as I did it) and where they can disable existing peer groups, set a fixed port, maybe later also the MTU?

NeoRaider:
Well, there is already a package `gluon-luci-mesh-vpn-fastd` in Gluon 2015.1 which allows to configure the `null` method. I think it would make sense to add this to the same package; while modularity is good, splitting the expert mode in too many small packages is just confusing.

ohrensessel:
ok I will have a look and refactor it

NeoRaider:
Thinking about this again, I'm not completely sure if we want the same fastd instance for both centralized and decentralized peerings. While it would save a few 100KB of RAM on the node, it would require to disable `no_rebroadcast` for the fastd instance (in most cases), also impacting the centralized peerings.

ohrensessel:
The reason using the same fastd instance was the argument, that we have
only single core routers...
using a second fastd instance is configuration-wise not that much of a
problem I suppose.
As the P2P VPN is something for "experts" only, one could argue that giving
the node owner a hint what kind of problems might arise (even in the case
two instances are used) is sufficient (more "grundrauschen" on the dsl line
because more tunnels, ...)

2015-05-18 21:01 GMT+02:00 Matthias Schiffer <notifications@github.com>:

NeoRaider:
I think the additional ~500KB RAM would be the only measuable overhead of the second fastd instance. The CPU load would be about the same as fastd is context-switching all the time anyways... So in my opinion the advantages of a second fastd instance outweigh the disadvantages.

christf:
@ohrensessel are you still working on this? Getting this merged would be a great step forward.

rubo77:
Is this already available as a community package somewhere?

CodeFetch:
@christf I'd love to see your package being merged. Is there anything other than the status page that needs to be updated? Can you please open up a new pull-request with your changes as @ohrensessel does not respond so we can close this referencing this pull-request as superseded for archival purposes?

ohrensessel:
for me, this was more a proof of concept. Do not expect anything from me. therefore, closing is ok for me.

rubo77:
I think this brilliant Idea should not be closed. Not until there is a follow-up PR (of someone who has the expertise to maintain this too)

CodeFetch:
@rubo77 I'm going to make a new pull-request today.

PR #380, opened 2015-05-17
fixed typos

kjokjo:


NeoRaider:
I don't really think it makes sense to fix up old release notes, as we always link them as http://gluon.readthedocs.org/en/v2015.1/releases/v2015.1.html etc. As the Git tag is included in the links, changing the RST source won't change the linked release notes. Thus, I guess we'll have to live with the issues in the release notes, they have been released and can't be changed.

PR #381, opened 2015-05-18
Altitude in meters above streetlevel

rubo77:
It is not easy to determine the exact altitude above NN. In openstreetmap the streetlevel already has the altitude information, so the altitude a user can determine by himself would be the altitude above streetlevel. For scripts that would plot these coordinates the altitudes can simply be added to the base level.

### DIFF
--- a/package/gluon-config-mode-geo-location/i18n/de.po
+++ b/package/gluon-config-mode-geo-location/i18n/de.po
@@ -27,7 +27,7 @@ msgid "Longitude"
 msgstr "Längengrad"
 
 msgid "Altitude"
-msgstr "Höhenmeter über Normalnull"
+msgstr "Höhenmeter über dem Boden"
 
 msgid "Show node on the map"
 msgstr "Knoten auf der Karte anzeigen"


NeoRaider:
I don't think getting accurate values about ground is much easier than getting the altitude about NN when you are a few levels above ground floor. And adding an inaccurate estimate of the altitude about ground with a possible inaccurate ground altitude won't give you any more accurate height values.

In addition, "ground level" or "street level" isn't well-defined at all when there are bridges and similar structures.

rubo77:
What do you want to achieve with this altitude vlaue? How should a user determin the altitude by himself?

I thought about the information for others to place their node in the optimal position to reach other nodes, when they know the altitude. 

Maybe we should change the whole field then to "storey" or "structural level"?

NeoRaider:
@rubo77, see this ridiculously dramatic video: https://www.youtube.com/watch?v=uc2V9zKxEUQ

tcatm:
For the altitude value to have any meaning we need to define a reference. We've decided to go with [WGS84](http://en.wikipedia.org/wiki/World_Geodetic_System) as we were using it for latitude and longitude already.

Determining the altitude in WGS84 is trivial for anyone equipped with a GPS-capable smartphone. Just hold the phone at the position of the node and copypaste the values. For indoor nodes some manual interpolation might be necessary but that's certainly doable for most people. You may also consult a (printed) map which includes height information and add the building's height to it.

In cases where it is impossible to determine the altitude the field may be left empty.

rubo77:
So we should have a button: "show altitude of streetlevel" in the mesh viewer or a link to a map, that has this easily accessible

PR #387, opened 2015-05-30
Add primary mac address and fastd config syntax to shown entry

Kunsi:
Added those values to make it easier to directly add config files to git, so it's no longer needed to manually adjust the pasted lines.


NeoRaider:
I don't really like the format, the exact config file syntax is a technical detail, not something a user should see. (And I think solving this by using scripts to maintain the key list would be a nice solution if you want to avoid manually touching the entries)

I can see though that some communities might want to adjust the key display format, some might even not want to display it at all (if they use fastd in a accept-all configuration). So I guess dropping the "the-key" div altogether and moving it into the `gluon-config-mode:pubkey` translation string might be the best solution to make this as flexible as possible.

Kunsi:
Should be fixed with #457

PR #391, opened 2015-06-08
gluon-core: add script for node-info at login

do9xe:
# general idea
In the IRC-Channel we came up with an idea to let the node throw out some useful information on login via command-line to make it easier to debug a node. In the following picture you can see a proof-of-concept version of the idea. It contains the core-information and the fastd-information.

node-info

# how it works
With this pull request come three important files an directories: ```/etc/profiles```, ```/lib/gluon/node-info``` and ```/lib/gluon/info.d```.
```/etc/profiles``` is called at login and is modified to call ```/lib/gluon/node-info```, which again calls every script in the directory ```/lib/gluon/info.d```. This allows other packages (like fastd) to provide a script into the directory, which is executed at login to display the current status of the service.
Because the info-scripts (except ```core```) are provided by the packages there is no need to write code and check weather a package is installed or not.

# ToDo-list
- [x] rework the code for ```core``` to make the uptime look prettier
- [ ] provide scripts for all packages, where info is useful
- [ ] define a maximum width for a line

This pull request can be merged, but some commits for more scripts will follow anyway.

### DIFF
--- a/package/gluon-core/files/lib/gluon/node-info
+++ b/package/gluon-core/files/lib/gluon/node-info
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+for script in /lib/gluon/info.d/*; do
+	"$script"
+done
--- a/package/gluon-core/files/etc/profile
+++ b/package/gluon-core/files/etc/profile
@@ -1,0 +1,16 @@
+#!/bin/sh
+[ -f /etc/banner ] && cat /etc/banner && /lib/gluon/node-info
+
+export PATH=/usr/bin:/usr/sbin:/bin:/sbin
+export HOME=$(grep -e "^${USER:-root}:" /etc/passwd | cut -d ":" -f 6)
+export HOME=${HOME:-/root}
+export PS1='\u@\h:\w\$ '
+
+[ -x /bin/more ] || alias more=less
+[ -x /usr/bin/vim ] && alias vi=vim || alias vim=vi
+
+[ -z "$KSH_VERSION" -o \! -s /etc/mkshrc ] || . /etc/mkshrc
+
+[ -x /usr/bin/arp ] || arp() { cat /proc/net/arp; }
+[ -x /usr/bin/ldd ] || ldd() { LD_TRACE_LOADED_OBJECTS=1 $*; }
+
--- a/package/gluon-core/files/lib/gluon/info.d/core
+++ b/package/gluon-core/files/lib/gluon/info.d/core
@@ -1,0 +1,9 @@
+#!/usr/bin/lua
+local util = require("luci.util")
+local sys = require("luci.sys")
+
+local hours = math.floor(sys.uptime() / 3600)
+local minutes = math.floor(sys.uptime() / 60) - (hours * 60)
+
+io.write(" your nodes uptime: " .. hours.. " hours, ".. minutes  .. " minutes\n")
+io.write(" " .. string.rep("-",53) .."\n")
--- a/package/gluon-mesh-vpn-fastd/files/lib/gluon/info.d/fastd
+++ b/package/gluon-mesh-vpn-fastd/files/lib/gluon/info.d/fastd
@@ -1,0 +1,64 @@
+#!/usr/bin/lua
+
+local util = require("luci.util")
+local fs = require("nixio.fs")
+local ltn12 = require 'luci.ltn12'
+local sys = require("luci.sys")
+local json = require("luci.json")
+local nixio = require 'nixio'
+local platform_info = require("platform_info")
+
+
+local stat, fastd_status = pcall(
+  function()
+    local fastd_sock = nixio.socket('unix', 'stream')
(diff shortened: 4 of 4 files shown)

### CODE REVIEW

File: package/gluon-core/files/lib/gluon/node-info [hunk 1]
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+for script in /lib/gluon/info.d/*; do
```
mweinelt: may change this to `banner.d`

do9xe:
Now, after i reworked the fastd script it looks like this:
http://pastebin.com/BFbvDTmk

do9xe:
This is my new suggestion for ```core```. I talked to @tcatm and we agreed, that we will only execute ```core``` at login and add a command to display the whole bunch of information so you don't get spamed at login. 
This will be, what ```core``` will look like:
```\n
root@Club-Mate:~# cat /etc/banner && /lib/gluon/info.d/core
  _______                     ________        __
 |       |.-----.-----.-----.|  |  |  |.----.|  |_
 |   -   ||  _  |  -__|     ||  |  |  ||   _||   _|
 |_______||   __|_____|__|__||________||__|  |____|
          |__| W I R E L E S S   F R E E D O M
 -----------------------------------------------------
 BARRIER BREAKER (Barrier Breaker, r45620)
 -----------------------------------------------------
  * 1/2 oz Galliano         Pour all ingredients into
  * 4 oz cold Coffee        an irish coffee mug filled
  * 1 1/2 oz Dark Rum       with crushed ice. Stir.
  * 2 tsp. Creme de Cacao
 -----------------------------------------------------
 Community: Freifunk Kiel
 Contact: DO9XE, FFKI-NOC

 primary MAC: e8:94:f6:51:cf:40
 Hardware: TP-Link TL-WR841N/ND v9
 Gluon-Version: v2015.1
 Release: 0.7
 Autoupdater: activ on branch stable
 your nodes uptime: 38 hours, 52 minutes
 -----------------------------------------------------
```

NeoRaider:
There are multiple issues with this package:

* There is no reason to make the call to `/lib/gluon/node-info` depend on the existence of `/etc/banner`
* `/etc/profile` won't be updated on sysupgrades
* `/etc/profile` is already part of the package base-files, you can't include the same file in two packages

I'd recommend editing `/etc/profile` in an upgrade script. In addition, I'd like to see this as a generic patch to source all scripts in `/etc/profile.d` like most desktop distros do. This should be realized as an OpenWrt patch (and also submitted upstream).

NeoRaider:
I'd like to see this implemented as an independent package, not in gluon-core.

Also, the `/lib/gluon` namespace is more or less reserved for packages with matching names (`/lib/gluon/FOO` should be created by the package `gluon-FOO`). It is okay for packages to drop files into the directory of another package, but there should always be one responsible package with matching name.

viisauksena:
i build something similar based on pure adding of /etc/profile
gluon version and fastd status would be nice .. at first we found important to give the node ipv4 capabilities if they dont have already (like mostly meshing nodes purely rely on v6 - which is bad for wget)

# get ipv4 if not already there, checked simply by wget call  
wget -O /dev/null --quiet http://nirgendwo.de;
if [ $? != 0 ]; then
                udhcpc -i br-client -q -S -n;
                fi
                echo " "
                echo "##########################################"
                echo -n "### IP :  "
                ifconfig |head|tail -n1
                echo    "### from ssh : "$SSH_CONNECTION
                echo -n "### uptime :  "
                uptime
                echo -n "### pubkey :  "
                /etc/init.d/fastd show_key mesh_vpn
                echo -n "### firmware :  "
                cat /lib/gluon/release
                echo -n "### hardware :  "
                cat /tmp/sysinfo/board_name
                echo -n "### connected to this node  :    "
                batctl tl |grep W |wc -l
                echo -n "### all Wifi Users in fffr  :    "
                batctl tg |grep W |wc -l
                echo    "########################################"
                echo -n "/sys/class/net/bat0/mesh/hop_penalty :    "
                echo $(cat /sys/class/net/bat0/mesh/hop_penalty)
                echo "batctl gw :     " $(batctl gw)
                echo    "########################################"
                batctl gwl
                echo    "########################################"


# just some handy aliases                                                                                 
alias cd..="cd .."
alias myip="echo -n 'ifconfig' && ifconfig |head|tail -n1 && echo 'from ssh : ' $SSH_CONNECTION"
alias status="source /etc/profile"
alias ll="ls -la"
PS1='$? ${debian_chroot:+($debian_chroot)}\[\033[01;33m\]\u@\h\[\033[00m\]:\[\033[01;34m\]\w\[\033[00m\]\$ '
~

viisauksena:
the fastd limits could be optained like this (it depends on how much grous their are)
**uci show fastd |grep limit**

NeoRaider:
This PR seems to be dead, I'm closing it. If someone wants to take over and work on the mentioned issues, please open a new one.

PR #393, opened 2015-06-10
Add /etc/profile.d support

do9xe:
OpenWrt should support an optinal /etc/profile.d directory like most other Linux distributions. This allows packages to install their own scripts into /etc/profile.d/.

### DIFF
--- a/patches/openwrt/0025-base-files-add-etc-profile.d-support.patch
+++ b/patches/openwrt/0025-base-files-add-etc-profile.d-support.patch
@@ -1,0 +1,29 @@
+From: Hendrik Lüth <hendrik@linux-nerds.de>
+Date: Wed, 10 Jun 2015 07:47:58 +0200
+Subject: base-files: add /etc/profile.d support
+
+OpenWrt should support an optinal /etc/profile.d directory like most other Linux
+distributions. This allows packages to install their own scripts into
+/etc/profile.d/.
+
+Signed-off-by: Hendrik Lüth <hendrik@linux-nerds.de>
+
+diff --git a/package/base-files/files/etc/profile b/package/base-files/files/etc/profile
+index 3dd58e1..c9e805f 100644
+--- a/package/base-files/files/etc/profile
++++ b/package/base-files/files/etc/profile
+@@ -14,3 +14,12 @@ export PS1='\u@\h:\w\$ '
+ 
+ [ -x /usr/bin/arp ] || arp() { cat /proc/net/arp; }
+ [ -x /usr/bin/ldd ] || ldd() { LD_TRACE_LOADED_OBJECTS=1 $*; }
++
++[ -z "$FAILSAFE" ] && {
++  if [ -d /etc/profile.d ]; then
++    for i in /etc/profile.d/*.sh ; do
++      if [ -r $i ]; then
++        . $i
++      fi
++    done
++  fi
++}
+


FreifunkUFO:
in openwrt its normally done inside startscripts in /etc/init.d

root@ci-serverraum:~# cat /etc/init.d/olsrd
#!/bin/sh /etc/rc.common
# Copyright (C) 2008-2013 OpenWrt.org

. $IPKG_INSTROOT/lib/functions/olsrd.sh

START=65

SERVICE_DAEMONIZE=1
SERVICE_WRITE_PID=1

CONF=/var/etc/olsrd.conf
PID=/var/run/olsrd.pid

UCI_CONF_NAME=olsrd

patches/openwrt/0009-base-files-add-etc-profile.d-support.patch
```diff
@@ -0,0 +1,27 @@
...
+index 3dd58e1..c9e805f 100644
+--- a/package/base-files/files/etc/profile
++++ b/package/base-files/files/etc/profile
+@@ -14,3 +14,12 @@ export PS1='\u@\h:\w\$ '
+ 
+ [ -x /usr/bin/arp ] || arp() { cat /proc/net/arp; }
+ [ -x /usr/bin/ldd ] || ldd() { LD_TRACE_LOADED_OBJECTS=1 $*; }
++
++if [ -d /etc/profile.d ]; then
++  for i in /etc/profile.d/* ; do
```
jplitza: At least Ubuntu and Debian use `/etc/profile.d/*.sh` here, which seems reasonable to me. Any reason to differ from that?

do9xe:
No, arch did it the same way. I didn't follow up on this on the openwrt devel list, because i had the feeling, that there was not enough interest.

jplitza:
@FreifunkUFO I have no idea what you are talking about. `/etc/init.d/*` is executed on system startup non-interactively, `/etc/profile` (and with this patch also `/etc/profile.d/*`) is executed on shell login.

As an update, the patch was [posted to the openwrt-devel ML](https://lists.openwrt.org/pipermail/openwrt-devel/2015-June/033552.html) (as demanded in #391), discussed, [modified to not start in failsafe mode](https://lists.openwrt.org/pipermail/openwrt-devel/2015-June/033594.html), criticized for no reason, [posted unmodified a third time](https://lists.openwrt.org/pipermail/openwrt-devel/2015-June/033760.html) and then ignored upstream. I think this justifies merging it locally in Gluon. @do9xe Could you update it to the latest version (with the non-failsafe clause) and rebase it to master (in the sense that you re-execute `make update-patches` so the patch number is correct)? If @tcatm and @NeoRaider (who argued in favor of the patch on openwrt-devel) don't think differently, I'd merge it then so #391 can continue.

do9xe:
I'll do this tomorrow, if this is fast enough, because we have our community meetup today :) If not, write me in IRC and I'll do it right away :)

NeoRaider:
@do9xe, please also change the glob to `/etc/profile.d/*.sh`.

bittorf:
@do9xe I was one if the people who gave feedback to you. i like it. i think it was not accepted yet, because you did not follow the OpenWrt style. I send a v4 for you:

https://lists.openwrt.org/pipermail/openwrt-devel/2015-September/035593.html

NeoRaider:
The patch has been accepted upstream, I've picked it up for Gluon.

Issue #284, opened 2015-01-24
Concept for port reconfiguration


neocturne:
We currently have some limited features to reconfigure the ports used by Gluon: "mesh-on-wan" is supported for some time, and there are conflicting pull requests for "mesh-on-lan" (freifunk-gluon/packages#63) and "lan+wan bridged" (freifunk-gluon/packages#71).

I propose a more generic approach:
- There are two ports: WAN and LAN. This might be extended to individual LAN ports in the future.
- There are three functions: uplink, mesh, client

It should be possible to assign any combination of functions to any port (with the limitation that the "client" function can't be enabled together with other functions). Enabling the uplink function on two ports will connect these ports with a bridge.

This could also be combined with the WLAN configuration as the three functions mesh, client and uplink directly correspond to the mesh WLAN, client WLAN and private WLAN.

flokli:
I would go even further:

**There are four networks:**

- uplink: The existing users' network through which the node should try to connect via fastd
- mesh: Network used for meshing. For each assignment, multiple interfaces would be needed
- client: The freifunk client network
- private-net: a network on which the node runs a dns and dhcp server and masquerades packages to the uplink network. Probably it would also make sense to do some routing into the freifunk network, this needs to be further discussed


**Available ports:**
- wan: the wan port
- lan: the lan port(s)
- wifi-mesh: the wifi interface used for meshing
- wifi-client: the wifi interface used for client accessing the freifunk network
- wifi-private: the wifi interface used for a private network

It would probably make sense to create these as lists, as on some routers, you could have multiple physical wifi interfaces or are able to address each lan port on its own.


When "connecting" ports and networks together in a "mapping", you can also select a vlan which should be used on the specific port.
On the initial boot, the node would create a "default mapping", which would be something like this:

- uplink -> wan, untagged
- mesh -> wifi-mesh, untagged
- client -> lan, untagged
- client -> wifi-client, untagged


Different functions would now look like this:

*Mesh on wan:*
remove uplink -> wan, untagged
add mesh -> wan, untagged

*Mesh on lan:*
remove client -> lan, untagged
add mesh -> lan, untagged

*Mesh on wan (different vlan):*
remove uplink -> wan, untagged
add mesh -> wan, vlan x

*lan+wan bridged:*
remove client -> lan, untagged
add uplink -> LAN, untagged

*private wifi, bridged with wan:*
add uplink -> wifi-client, untagged

*private network on lan and wifi, natting through uplink:*
[...]

Issue #163, opened 2014-08-13
Private Wifi with NAT & DHCP


mweinelt:
Is it possible to provide NAT Gateway functionality with DHCP instead of WAN Bridging for the private Wifi? Users in our university housing locations only get one IP and one lan port, so the current functionality does not work for them.

maurerle:
There currently is https://github.com/ffac/gluon-packages/tree/main/ffac-web-private-wan-dhcp which provides this functionality, but needs further improvements before going to the community-packages.

This is especially useful for cellular network routers or DSL with FB7530.
I don't know if we should keep track of this here?

PR #398, opened 2015-06-16
Add NAT features

mar-v-in:
Related to #163 and #284.
Unfinished (bridge on lan missing, translations missing, only tested on TL-841N). Primarily asking for a first feedback here :)

### DIFF
--- a/package/gluon-core/files/lib/gluon/upgrade/110-network
+++ b/package/gluon-core/files/lib/gluon/upgrade/110-network
@@ -10,7 +10,7 @@ uci:section('network', 'interface', 'wan',
 	      ifname = sysconfig.wan_ifname,
 	      type = 'bridge',
 	      igmp_snooping = 0,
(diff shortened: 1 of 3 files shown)


NeoRaider:
Some obvious issues:
* `peerdns` for the WAN interface must *never* be set to 1, or the DNS over the mesh will be completely broken. There are two dnsmasq instances on Gluon nodes, one the the mesh on port 53 and one for WAN on port 54 (that one is used by fastd only at the moment).
* Unconditionally creating an interface with address 192.168.0.1 will break WAN when the upstream network uses 192.168.0.0/24 as well

I don't think it makes sense to write any code related to network configuration until #284 is implemented, as I plan to throw out all current network setup code in Gluon for that...

Another thing: Don't base your development on the v2015.1.x branch. The stable branches are for bugfixes and well-tested feature backports only, all new features must be implemented in the master first.

PR #402, opened 2015-06-21
Bridge Multicast Snooping

T-X:
Here are a few patches which should enable bridge multicast snooping again, including the bridge "multicast-to-unicast" feature to forward multicast packets to wifi STAs via unicast. All-in-all this should reduce the overhead on the wifi quite a bit (not for the VPN yet though, that's what the upcoming batman-adv multicast optimizations for bridges should deal with soon). And a working multicast-to-unicast would lay the foundation for any multicast streaming application in the future.

This patchset basically segments the IGMP/MLD domain, keeping IGMP/MLD messages node-local and allowing to have an IGMP/MLD querier on every node instead of needing to place it in a central position (like on a gateway). I'd be curious what others think about this approach.

### DIFF
--- a/package/gluon-core/files/lib/gluon/upgrade/110-network
+++ b/package/gluon-core/files/lib/gluon/upgrade/110-network
@@ -9,11 +9,12 @@ uci:section('network', 'interface', 'wan',
 	    {
 	      ifname = sysconfig.wan_ifname,
 	      type = 'bridge',
-	      igmp_snooping = 0,
 	      peerdns = 0,
 	      auto = 1,
 	    }
 )
+uci:delete('network', 'wan', 'igmp_snooping')
+uci:delete('network', 'wan', 'multicast_querier')
 
 if not uci:get('network', 'wan', 'proto') then
   uci:set('network', 'wan', 'proto', 'dhcp')
--- a/package/gluon-core/files/lib/gluon/upgrade/110-network
+++ b/package/gluon-core/files/lib/gluon/upgrade/110-network
@@ -56,5 +56,31 @@ uci:save('network')
 uci:commit('network')
 
 
+uci:section('firewall', 'rule', 'wan_igmp',
+	    {
+	       name = 'Allow-IGMP',
+	       src = 'wan',
+	       proto = 'igmp',
+	       family = 'ipv4',
+	       target = 'ACCEPT',
+	    }
+)
+
+uci:section('firewall', 'rule', 'wan_mld',
+	    {
+	       name = 'Allow-MLD',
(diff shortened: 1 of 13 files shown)


NeoRaider:
While I haven't looked through the code yet: The `patches` subdirectory of the Gluon tree is not supposed to be touched manually, but only by the `make update-patches` command. Just add the commits to the `openwrt` git repository via `git am` or `git cherry-pick` and run update-patches.

T-X:
bridge multicast-to-unicast is enabled by default in OpenWRT and some people already gave feedback to the current state in trunk. No more issues I'm aware of and any fix or feature is a backport from OpenWRT trunk. Therefore removing the "RFC" in the subject line - feel free to merge :).

NeoRaider:
Should be rebased onto #482 after it is merged. Needs a rebase anyways, the patch numbers for the OpenWrt repo don't match anymore.

By the way, how does this handle roaming clients? Are MLD/IGMP queries sent often enough to catch such clients quickly enough?

T-X:
Roaming is a good point. I'm trying to gather some more information on that.

https://www.mail-archive.com/bridge@lists.linux-foundation.org/msg04193.html

NeoRaider:
Many of the bridge patches have been backported upstream, so this will need another rebase...

T-X:
Next to the bridge mailing list I linked above I also asked for feedback on the new IETF multicast-wifi mailing list (https://mailarchive.ietf.org/arch/msg/mcast-wifi/Ghn2cGy1oN2ZwG1qaQO9SE13g6g). Unfortunately so far, the responses from these two lists were very sparse.

The cleanest solution would be to update all mobile devices to reissue reports upon roaming (Linux currently does not...). Which is unfeasible for the short or mid term.

I was thinking about implementing some hacks within the bridge itself (similar to what I described in the linked mails) initially. But now I'm tending to simply adjust the MLD querier intervals of the bridge:

Query interval: 125s -> 20s
Query response interval: 10s -> 5s

After splitting the IGMP/MLD domain, lower intervals shouldn't cause too much burden. 20+5s for the worst case (12.5s average) is well below any TCP timeout and below the "impatient user" interval of 30s we learned about in "Softwaretechnik".

For the few people who might want to experiment with things like VoIP roaming, needing < 1s delay, I'm going to propose a patch to the Linux kernel afterwards.

I'll also split this pull request into two ones: First the IGMP/MLD domain split which should help with cases like #553 already. And will put the IGMP/MLD snooping (ref: #278) part in a second one.

NeoRaider:
Superseded by #688.

PR #408, opened 2015-07-04
Add support for TP-Link TL-WA830RE v2

oakey-dev:


oakey-dev:
https://freifunk-ingolstadt.de/map/#!n:a0f3c19f1c51

NeoRaider:
I think the commit message is a bit misleading, I'd prefer it to say something like "TP-Link TL-WA830RE v2: remove BROKEN"

oakey-dev:
done

PR #409, opened 2015-07-05
Do not use error() in gluon-neighbour-info

hwhw:
error() is a glibc specific function and should not be used in
code meant to be portable.

Fixes build on musl-libc.


tcatm:
As per https://github.com/freifunk-gluon/gluon/wiki/Developer-guidelines commit messages should be prefixed by the packagename affected.

hwhw:
Ah, thanks, I blame yesterday's heat :-) fixed.

PR #422, opened 2015-07-25
mesh-batadv-core: introduce 11s mesh, refactor wireless config

tcatm:
This is a site.conf-breaking change in regard to the wireless config.
Make sure to read http://gluon.readthedocs.org/en/latest/user/site.html
and update your site.conf accordingly!

Support for 802.11s mesh interfaces has been added. Gluon now supports
three interface types: ap, ibss and mesh. All of them are now optional
and may be configured independently in site.conf.

A sample site.conf may look like this:

    wifi24 = {
            channel = 1,
            htmode = 'HT40+',
            ap = {
                    ssid = 'luebeck.freifunk.net',
            },
            ibss = {
                    ssid = '02:d1:11:37:fc:38',
                    bssid = '02:d1:11:37:fc:38',
                    mcast_rate = 12000,
            },
            mesh = {
                    id = 'ffhl-mesh',
                    mcast_rate = 12000,
            },
    },

### DIFF
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -62,24 +62,46 @@ regdom
       regdom = 'DE'
 
 wifi24
-    WLAN Configuration of your community in the 2.4Ghz radio. Consisting
-    of ``ssid`` of your client network, the ``channel`` your community is using,
-    ``htmode``, the adhoc ssid ``mesh_ssid`` used between devices, the adhoc
-    bssid ``mesh_bssid`` and the adhoc multicast rate ``mesh_mcast_rate``.
-    Optionally ``mesh_vlan`` can be used to setup VLAN on top of the 802.11
-    ad-hoc interface. The options ``mesh_disabled`` and ``client_disabled``
-    are optional, too. They allow to disable the SSID by default, e.g. for
-    preconfigured node. This only affects first configuraton.
-    Combined in an dictionary, e.g.:
+    WLAN configuration for 2.4 GHz devices.
+    ``channel`` must be set to a valid wireless channel for your radio.
+    ``htmode`` selects the desired htmode (e.g. HT20, HT40- or HT40+).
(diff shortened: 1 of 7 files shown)


docs/user/site.rst
```diff
@@ -62,24 +62,46 @@ regdom
...
+    ``htmode`` selects the desired htmode (e.g. HT20, HT40- or HT40+).
+
+    There are currently three interface types available. You many choose to
+    configure any subset of them:
+
+    - ``ap`` creates a master interface where clients may connect
+    - ``mesh`` creates an 802.11s mesh interface with forwarding disabled
+    - ``ibss`` creates an ad-hoc interface
+
+    All interfaces may be disabled by setting ``disabled`` to ``false``.
```
leeps: This sounds wrong, you probably mean

    All interfaces may be disabled by setting ``disabled`` to ``true``.

chris007de:
@tcatm When two nodes have both ibss and mesh activated and see each other, what is the intended behavior? Do ad-hoc and 802.11s work in parallel or is 802.11s preferred automatically?

NeoRaider:
@chris007de, batman-adv is very stupid in regard to choosing the "best" link, it only uses the packet loss as metric. When both links have the same packet loss, one is chosen at random. (Just like it's almost impossible to force batman-adv to use a wired mesh connection when the wireless link is perfect as well...)

Having IBSS and 11s enabled should be avoided anyways: all OGMs and broadcast traffic (i.e. the whole background traffic) will be transmitted twice, eating up your airtime and thus bandwidth. In Lübeck, we'll probably have a migration phase of about one week in which we enable both IBSS and 11s, after that week, we'll roll out another firmware update removing the IBSS config.

rubo77:
Did you start the migration phase by now?

PR #441, opened 2015-08-17
br-client: convert option ifname to list ifname

tcatm:
Convert option ifname in br-client to use a list instead. This
simplifies adding and remove interfaces:

uci:add_to_set("network", "client", "ifname", "eth0")
uci:remove_from_set("network", "client", "ifname", "eth0")

An option ifname will be automatically converted to a list when
performing an upgrade.

Packages affected: gluon-mesh-batman-adv-core, gluon-luci-portconfig

---

I have tested it as follows:

- upgrade from master
- check portconfig to work (both enabling and disabling mesh-on-lan)
- check initial setup (i.e. sysupgrade -n) with mesh_on_lan = false
- check initial setup with mesh_on_lan = true

### DIFF
--- a/package/gluon-mesh-batman-adv-core/files/lib/gluon/upgrade/310-gluon-mesh-batman-adv-core-mesh
+++ b/package/gluon-mesh-batman-adv-core/files/lib/gluon/upgrade/310-gluon-mesh-batman-adv-core-mesh
@@ -20,22 +20,28 @@ uci:commit('batman-adv')
 
 
 if not uci:get('network', 'client') then
-  local ifname
+  uci:section('network', 'interface', 'client',
+    {
+      type = 'bridge',
+      proto = 'dhcpv6',
+      reqprefix = 'no',
+    }
+  )
+
+  uci:add_to_set('network', 'client', 'ifname', 'bat0')
 
   if sysconfig.lan_ifname and not site.mesh_on_lan then
-    ifname = sysconfig.lan_ifname .. ' bat0'
-  else
-    ifname = 'bat0'
+    uci:add_to_set('network', 'client', 'ifname', sysconfig.lan_ifname)
   end
+end
 
-  uci:section('network', 'interface', 'client',
-	      {
-		ifname = ifname,
-		type = 'bridge',
-		proto = 'dhcpv6',
-		reqprefix = 'no',
-	      }
-  )
+local ifname = uci:get('network', 'client', 'ifname')
+
+if type(ifname) == 'string' then
+  uci:delete('network', 'client', 'ifname')
+  for x in ifname:gmatch("[^%s]+") do
+    uci:add_to_set('network', 'client', 'ifname', x)
+  end
 end
 
 uci:set('network', 'client', 'igmp_snooping', 0)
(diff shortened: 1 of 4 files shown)


NeoRaider:
I think the uci list handling functions should use the same names they use in the uci cli: `add_list` and `del_list`, and also behave the same.

patches/packages/luci/0005-model.uci-add-add_to_set-remove_from_set.patch
```diff
@@ -0,0 +1,90 @@
...
++	else
++		set[value] = true
++	end
++
++	list = {}
++	for k, _ in pairs(set) do
++		table.insert(list, k)
++	end
++
++	return self:set_list(config, section, option, list)
```
NeoRaider: I've noticed that this breaks when you try to remove the last element of a set (as the `set` and `set_list` function of UCI don't allow setting an empty table). I'm not sure if this should be fixed here or changed in uci itself. Also, `set_list` is a weird wrapper around `set`, just using `set` is probably preferable.

package/gluon-mesh-batman-adv-core/files/lib/gluon/upgrade/310-gluon-mesh-batman-adv-core-mesh
```diff
@@ -20,22 +21,25 @@ uci:commit('batman-adv')
...
-  local ifname
+  uci:section('network', 'interface', 'client',
+    {
+      type = 'bridge',
+      proto = 'dhcpv6',
+      reqprefix = 'no',
+    }
+  )
+
+  uci:add_to_set('network', 'client', 'ifname', 'bat0')
```
NeoRaider: The `bat0` entry can be merged into the `section` call.

package/gluon-mesh-batman-adv-core/files/lib/gluon/upgrade/310-gluon-mesh-batman-adv-core-mesh
```diff
@@ -3,6 +3,7 @@
 local sysconfig = require 'gluon.sysconfig'
 local sysctl = require 'gluon.sysctl'
 local site = require 'gluon.site_config'
+local util = require 'luci.util'
```
NeoRaider: I guess we shouldn't add new LuCI dependencies when we're elsewhere trying to get rid of it...

package/gluon-mesh-batman-adv-core/files/lib/gluon/upgrade/310-gluon-mesh-batman-adv-core-mesh
NeoRaider: I wonder if we should copy some simple UCI helpers to gluon.util to avoid depending on LuCI?

tcatm: I have prepared a package containing all useful helpers named simple-uci (https://github.com/freifunk-gluon/packages/commit/9e81355843f7d0f7b59ff2f1fdb8980ad6bc31c2). It's a drop-in for luci.model.cbi using the same API (excluding any luci specific calls).

tcatm:
I have removed the dependency on luci.util.

PR #447, opened 2015-08-29
Announce client count separated by frequency

jplitza:
It might be nice to distinguish how many clients are connected to a node via 2.4 GHz and 5 GHz, so we could gather information about the usage of the 5 GHz band if available.

While the actual implementation is rather easy (we already used [a hackish version](https://github.com/FreifunkBremen/ffhb-packages/blob/breminale/ffhb/ffhb-breminale/files/lib/gluon/announce/statistics.d/clients) that simply parses the output of `iw dev ... station dump`), we should probably first fix the output format in the JSON. Currently, we have:

    "clients": {
      "wifi": 3,
      "total": 3
    },

I would propose the following format, which obviously isn't backwards compatible:

    "clients": {
      "wifi24": 2,
      "wifi50": 1,
      "other": 0
    },

Having a "total" field doesn't make much sense when having more than one other field, I think, because you can get the total by something like `.clients | add` (in jq), whereas it would be more cumbersome and less flexible to get the count of "other" given "wifi24", "wifi50" and "total".

(I was by the way told that @NeoRaider already approved of the general idea at CCCamp, and that he would like different types of clients to have different color tints on the map)

### DIFF
--- a/package/gluon-mesh-batman-adv-core/files/lib/gluon/announce/statistics.d/clients
+++ b/package/gluon-mesh-batman-adv-core/files/lib/gluon/announce/statistics.d/clients
@@ -1,20 +1,55 @@
-local list = io.lines("/sys/kernel/debug/batman_adv/bat0/transtable_local")
+local iwinfo = require 'iwinfo'
 
-local count = 0
-local wifi = 0
+local counts = { total = 0
+               , wifi = 0
+               , wifi24 = 0
+               , wifi5 = 0
+               }
+
+local list = io.lines("/sys/kernel/debug/batman_adv/bat0/transtable_local")
+local clients = {}
 for line in list do
   local mac, _, flags, lastseen = line:match("^ %* ([0-9a-f:]+) *(.- )%[(.-)%] +(%d+%.%d+)")
   if mac then
     if not flags:match('P') then
-      count = count + 1
+      counts.total = counts.total + 1
+      clients[mac:lower()] = true
 
       if flags:match('W') then
-        wifi = wifi +1
+        counts.wifi = counts.wifi +1
       end
     end
   end
 end
 
-return { total = count
-       , wifi  = wifi
-       }
+function count_iface_stations(iface)
+  local wifitype = iwinfo.type(iface)
+  if wifitype == nil then
+    return
+  end
+
+  local freq = iwinfo[wifitype].frequency(iface)
+  local key
+  if freq >= 2400 and freq < 2500 then
+    key = "wifi24"
+  elseif freq >= 5000 and freq < 6000 then
+    key = "wifi5"
+  else
+    return
+  end
+
+  for k, v in pairs(iwinfo[wifitype].assoclist(iface)) do
+    if clients[k:lower()] then
+      counts[key] = counts[key] + 1
+    end
+  end
+end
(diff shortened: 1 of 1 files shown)


jplitza:
Okay, this patch produces the following format:
```
"clients": {
    "wifi": 2,
    "total": 3,
    "wifi24": 2,
    "wifi5": 0
}
```
As easy as it can get.
The wifi number is still read from the W flags in batman-adv, so if this gets merged we can monitor if this is always equal to wifi24 and wifi5. total also still comes from batman-adv obviously.

package/gluon-mesh-batman-adv-core/files/lib/gluon/announce/statistics.d/clients
```diff
@@ -1,20 +1,51 @@
...
+    key = "wifi24"
+  elseif freq >= 5000 and freq < 6000 then
+    key = "wifi5"
+  else
+    return
+  end
+
+  for k, v in pairs(iwinfo[wifitype].assoclist(iface)) do
+    counts[key] = counts[key] + 1
+  end
```
NeoRaider: This can be done much simpler: `counts[key] = counts[key] + #iwinfo[wifitype].assoclist(iface)` (`# ` is the length operator in Lua)

jplitza: No, it can't. The length operator only really works on numerically indexed tables without gaps. From [the Lua manual](http://www.lua.org/manual/5.1/manual.html#2.5.5):
> The length of a table t is defined to be any integer index n such that t[n] is not nil and t[n+1] is nil; moreover, if t[1] is nil, n can be zero. For a regular array, with non-nil values from 1 to a given n, its length is exactly that n, the index of its last value. If the array has "holes" (that is, nil values between other non-nil values), then #t can be any of the indices that directly precedes a nil value (that is, it may consider any such nil value as the end of the array). 

The `assoclist` however is indexed by MAC addresses, thus `#` doesn't work.

NeoRaider:
Hmm, I get the following output on one of my test nodes:

    "clients": {
      "wifi": 1,
      "wifi24": 2,
      "total": 2,
      "wifi5": 0
    }

The reason is that there's an inactive associated client that isn't seen by batman-adv (the second client in "total" is wired). I think the best solution would be to count stations only if the MAC address is known to batman-adv as well.

jplitza:
Should be fixed now: While counting batman clients, a table with the mac addresses is created, which is later checked for when counting the wifi clients.

My output didn't change greatly, but I didn't have those strange inactive associated clients either. 😛

    "clients": {
      "wifi": 1,
      "total": 2,
      "wifi24": 1,
      "wifi5": 0
    }

bittorf:
like proposed for netjson:
https://lists.funkfeuer.at//pipermail/interop-dev/2015-September/000564.html

"properties": {
  "clients": {
    "wifi_2ghz": 8,
    "wifi_5ghz": 12,
    "wifi_total": 20,
    "ethernet": 5,
    "bluetooth": 1,
    "total": 26
  }
}


24 is the same mistake like 50. also: 5ghz goes from 4,9 to 6,1 ghz 8-)

PR #461, opened 2015-08-31
Remove dependencies on luci.util (and partly nixio.fs) from announce scripts

jplitza:


NeoRaider:
It would be nice to close all files after use and not rely on the garbage collector for that, as the Lua state will live for a long time in some contexts (e.g. respondd). It might make sense to add a utility function to gluon.util that reads a single line.

jplitza:
Okay, implemented `gluon_util.readline()` which takes a file object and returns the first line (after closing the file). It takes a file object and not a path because this way, it can also be used for the output of commands executed via `io.popen()`.

Also, `announce.lua` now necessarily contains
```
util = require 'gluon.util'
```

package/gluon-core/files/usr/lib/lua/gluon/util.lua
```diff
@@ -52,6 +52,12 @@ function replace_prefix(file, prefix, add)
 	os.rename(tmp, file)
 end
 
+function readline(fd)
+	local line = f:read('*l')
+	f:close()
+	return line
+end
```
NeoRaider: You called the argument fd, not f ...

tcatm:
`nproc` returns 0 on single CPU devices.

jplitza:
Sorry, somehow didn't see the `else` branch there. Fixed.

tcatm:
Works now. Thanks!

PR #462, opened 2015-08-31
RFC: Add contribution guidelines

jplitza:
I think Gluon could need some general contribution guidelines. As I wrote in the draft, "[t]o ease the work for the maintainers and to reduce the frustration of contributors".

I wrote down what came to my mind, but of course it is essential that the maintainers decide on something. This PR is just in case they are too lazy to write it down and I captured everything necessary (probably not the case).


tcatm:
ACK

PR #464, opened 2015-08-31
gluon-announce: build a tree of functions first

tcatm:
collect_dir() will now pre-load all files and return a function that will
collect all information.


NeoRaider:
Is the resource usage of the the C loader significantly lower? The code is much more complex, thus I'd like to avoid replacing the Lua code if possible...

tcatm:
I think it was several 10kb but I don't have the exact numbers anymore.

PR #465, opened 2015-08-31
Use respondd from gluon-alfred and gluon-status-page

jplitza:


package/gluon-alfred/files/lib/gluon/cron/alfred
```diff
@@ -1 +1 @@
-* * * * * /lib/gluon/announce/collect.lua nodeinfo | gzip | alfred -s 158; /lib/gluon/announce/collect.lua statistics | gzip | alfred -s 159; /lib/gluon/announce/collect.lua neighbours | gzip | alfred -s 160
+* * * * * gluon-neighbour-info -d ::1 -p 1001 -t 0 -r nodeinfo | gzip | alfred -s 158; gluon-neighbour-info -d ::1 -p 1001 -t 0 -r statistics | gzip | alfred -s 159; gluon-neighbour-info -d ::1 -p 1001 -t 0 -r neighbours | gzip | alfred -s 160
```
NeoRaider: What will gluon-neighbour-info do with `-t 0` when the request packet gets lost? While that shouldn't happen normally, won't that make the command block indefinitely when
* respondd is not running or
* the request or answer is dropped because of full queues?

tcatm: That's a good point. We should probably add a single-response feature:

- if a response has been received exit with 0 immediately
- if it runs into the timeout without having received a response, exit with 1

This could be generalized a little as to wait for n responses instead of one. The optional parameter could be called `-c` taking a number as its argument.

jplitza:
I implemented the proposed switch `-c <n>` with the following return code semantics:
* If `n > 0`, exit successfully iff exactly `n` packages have been received (more isn't possible anyway)
* If `n = 0` (the default), exit successfully if at least one packet hast been received.
The latter is a change from the previous behavior, but is only consistent with the proposed influence of `-c <n>` on the return code.

NeoRaider:
This pull request needs a rebase, as the status page API has been split off into a new package.

NeoRaider:
I don't like the way the old announced API without the `GET` verb is used again. The only reason we still support the old `nodeinfo` request is the old status page, which might still be in use on other nodes.

We should rather think of a proper announced API, containing:
* verbs for different request types (GET/GETZ for uncompressed/compressed data?)
* A way to partition nodes for multicast queries. Sending a request to all nodes will cause traffic peaks which could be easily avoided. My suggestion would be to have a request type including a modulus m and a remainder r, which is only answered by nodes with `nodeid mod m == r`.

jplitza:
@tcatm had the idea of making it possible to request very specific fields (like `nodeinfo.node_id` and `statistics.clients.wifi`). That would justify using only an API that makes it possible to request multiple fields at once. But as it is now, I don't see any advantage in forcing the "new" API, or discouraging the "old" API, because it is simple and works.
Anyways, I'll rebase it for now.

NeoRaider:
The old API has been deprecated for good reasons. Having requests without a verb like "GET" should be discouraged in my opinion, as it clutters the request namespace.

Another way to go forward from here might be to drop support for uncompressed replies completely (using uncompressed packets might cause further issues when the data gets too big for a single ethernet frame...). Adding decompression to our Lua deflate library would make sense anyways.

Nevertheless, we should casefully think about the API now before we finalize it. (Although there be might have another chance to fix it when we switch to NetJSON, as that will break compatiblity anyways.) Allowing to query for specific attributes sound very useful and might save lots of resouces.

Does anyone have any comments regarding my "partitioned query" proposal?

tcatm:
I believe the intention was to use "GET" as the external API while the other one is just for internal use (to avoid compressing and decompressing data on the same node).

jplitza:
#### (Un)compressed responses
I'm against dropping uncompressed responses altogether. There is no need (and in fact, no justification) for compressing the answer if it is delivered to localhost, as it is for example in #465. The lo interface typically has an MTU of 65536, which we probably won't reach too soon. And forcing compression in that case needlessly uses up CPU cycles. Using the verbs `GET` for uncompressed and `GETZ` for compressed sounds good to me.

#### Querying individual fields
Well, I definitely like the idea too. It probably involves some major overhaul of gluon-announced, but should be doable.

#### Partitioning responses
IIRC, tests by @tcatm have shown that there is no need for partitioning, as in a moderately large mesh, only very few nodes didn't respond to the first request. And AFAIR those tests didn't even include mitigations such as unsolicited neighbour advertisements to the multicast address preceding the request. So primarily, I'm against this blowup altogether because it introduces lots of complexity and problems on the client side (What if we receive answers to multiple remainders? How long do we wait between remainders? What is a good value for *m*?). Also, if it is introduced nonetheless, it should be optional so that querying direct neighbours, e.g. via WiFi, doesn't need the remainder-hassle.

Apart from the necessity, you proposal has a problem: `nodeid` isn't guaranteed to have a numeric interpretation, and if it has (MAC addresses have), nothing suggests it's evenly distributed mod some integer, does it? So it should probably be replaced by some random integer `n` that is generated on first boot or something, and to foster a uniform distribution, `m` should be fixed or from a fixed set. For example, we could pick *n∈{1,…,N!}* uniformly random, making it uniformly distributed modulo *m* for every *m∈{2,…,N}* if I'm not mistaken. The maximal realistic value of *N* probably is 12, as 13! exceeds 4 bytes. Dividing some terms from the factorial, we could also use *n∈{1,…,2⁴·3²·5·7·11·13·17·19=232792560}* to get uniform distribution modulo *m* for every *m∈{2,…,20}*. This would enable the inquirer to divide the network into up to 20 roughly evenly sized segments, at his discretion. Or (probably most reasonable) we only allow a coarser (exponentially growing) choice of *m*, choose *n∈{0,…,2³²}* and allow *m∈{2^k | 0<k≤32}*.

NeoRaider:
I have rebased and merged this pull request (patch 1/4 has become redundant after #464).

PR #470, opened 2015-09-01
Add sink support to luci-lib-jsonc and use it where needed

jplitza:


NeoRaider:
I find the way the sink is attached to the module quite awkward (and making C metatables accessible always asks for trouble).

Just having a normal function that can be used like `sink(decoder)` instead of `decoder:sink()` seems much cleaner to me.

jplitza:
It's now implemented directly in the C library as a method of the parser object. (What was the "needs work" label for?)

patches/packages/luci/0007-luci-lib-jsonc-Add-ltn12-compatible-sink-factory.patch
```diff
@@ -0,0 +1,133 @@
...
++				lua_pop(L, 2);
++				lua_pushboolean(L, true);
++				return 1;
++			}
++		}
++	}
++}
++
++static int json_parse_sink(lua_State *L)
++{
```
NeoRaider: This function is missing an argument check.

PR #472, opened 2015-09-02
odhcp6c: minor fixes

tcatm:
This helps with multiple radvds.


NeoRaider:
I usually remove the git-svn line and reset the commit author to myself when backporting patches from the OpenWrt trunk.

PR #473, opened 2015-09-02
Don't fork batctl from announce scripts

jplitza:
In particular this means to generate the list of mesh interfaces in lua (done in new module `gluon.batman_adv` by globbing `/sys/class/net/bat0/lower_*`) and to open `/sys/kernel/debug/batman_adv/bat0/gateways` instead of calling `batctl gateways`.

This PR contains the changes from #461


NeoRaider:
If we don't care that we're now iterating over all mesh interfaces and not only the ones in active state (and I don't think we care), this can be merged.

PR #478, opened 2015-09-03
Split gluon-status-page-api from gluon-status-page, rewrite parts in C

tcatm:


package/gluon-status-page-api/src/neighbours-batadv.c
```diff
@@ -0,0 +1,61 @@
...
+
+static json_object *neighbours(void) {
+  struct json_object *obj = json_object_new_object();
+
+  FILE *f;
+
+  f = fopen("/sys/kernel/debug/batman_adv/bat0/originators" , "r");
+
+  if (f == NULL) {
+    perror("Ca not open bat0/originators");
```
jplitza: "n" missing.

package/gluon-status-page-api/src/stations.c
```diff
@@ -0,0 +1,99 @@
...
+    return stations;
+
+  if (len <= 0)
+    return stations;
+
+  len /= sizeof(struct iwinfo_assoclist_entry);
+
+  struct iwinfo_assoclist_entry *entry = (struct iwinfo_assoclist_entry *)buf;
+
+  for (int i = 0; i < len; i++, entry++) {
```
NeoRaider: The variable `i` can be completely removed if the condition is changed to something like `(char*)(entry+1) <= buf+len`. The early return statement can be removed as the `for` condition will never be true if `len <= 0`.

NeoRaider:
`500-status-page` could also be renamed to `500-status-page-api`.

package/gluon-status-page-api/src/neighbours-batadv.c
```diff
@@ -0,0 +1,65 @@
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+#include <json.h>
```
NeoRaider: The recommended include for json-c is `<json-c/json.h>`, the `<json.h>` include is deprecated and will be removed from json-c in the future.

tcatm:
I've renamed the update-script.

tcatm:
#include changed to <json-c/json.h>.

package/gluon-status-page-api/src/stations.c
```diff
@@ -0,0 +1,94 @@
...
+}
+
+int main(void) {
+  char *ifname = getenv("QUERY_STRING");
+
+  if (ifname == NULL)
+    badrequest();
+
+
+  FILE *list = popen("batctl if", "r");
```
NeoRaider: The call to batctl here could be avoided, too.

NeoRaider: Especially as the device name is only validated, so it can be  looked up directly in sysfs.

tcatm:
Changed to query sysfs directly instead of using batctl if.

marvin:
Is the status-page-api going to be included in the 2015.1.2 branch or does it stay in the master branch?
We would like to use it.

tcatm:
@marvin It will not get backported to 2015.1.x but it'll be available in 2015.2.

marvin:
@tcatm thanks. when is 2015.2 going to be released?

PR #481, opened 2015-09-05
contributing: mention "rejected features" wiki page

tcatm:


jplitza:
As this seems to be more of a "list of issues that were rejected", why not make it exactly that and tag rejected issues with "rejected" and link to that filtered list?

tcatm:
We'd like to do that, too. However, the list should be a curated one like http://www.gnu.org/software/coreutils/rejected_requests.html so it's easy to spot whether a feature has already been rejected.

jplitza:
The items on that list are so short that I don't see how you can spot the feature you were about to propose any easier than on the list of only issue titles, if they are changed to something meaningful whem they are rejected.

Also, you could change this into a general "Please search for duplicates first", with the amendment "If it was rejected, it will be rejected again"

jplitza:
After discussion with @tcatm I changed the link to point to the [filtered issue list](https://github.com/freifunk-gluon/gluon/issues?q=label%3Arejected), as the issue title can be changed on rejection to state clearly which feature was rejected, and mentioned searching for duplicates in general.

rubo77:
someone has to go through all these:

https://github.com/freifunk-gluon/gluon/issues?utf8=%E2%9C%93&q=-label%3Aduplicate+-label%3Aupstream+-label%3Aquestion+-label%3Arejected+-label%3Adocumentation+-label%3A%22ACK+tcatm%22+-label%3A%22in+progress%22+is%3Aclosed+

rubo77:
someone has to go through all these:

https://github.com/freifunk-gluon/gluon/issues?utf8=%E2%9C%93&q=-merged%3A%3C2031-01-01+-label%3Aduplicate+-label%3Aupstream+-label%3Aquestion+-label%3Arejected+-label%3Adocumentation+-label%3A%22ACK+tcatm%22+-label%3A%22in+progress%22+is%3Aclosed+

PR #482, opened 2015-09-06
gluon-client-bridge: basic br-client config and wireless AP

tcatm:
This package provides br-client and sets up a wireless AP interface for
clients.

### DIFF
--- a/package/gluon-client-bridge/Makefile
+++ b/package/gluon-client-bridge/Makefile
@@ -1,17 +1,17 @@
 include $(TOPDIR)/rules.mk
 
-PKG_NAME:=gluon-mesh-batman-adv-core
+PKG_NAME:=gluon-client-bridge
 PKG_VERSION:=1
 
 PKG_BUILD_DIR := $(BUILD_DIR)/$(PKG_NAME)
 
 include $(GLUONDIR)/include/package.mk
 
-define Package/gluon-mesh-batman-adv-core
+define Package/gluon-client-bridge
   SECTION:=gluon
   CATEGORY:=Gluon
-  TITLE:=Support for batman-adv meshing (core)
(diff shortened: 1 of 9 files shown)


package/gluon-client-bridge/Makefile
```diff
@@ -0,0 +1,36 @@
...
+define Build/Compile
+endef
+
+define Package/gluon-client-bridge/install
+	$(CP) ./files/* $(1)/
+endef
+
+define Package/gluon-client-bridge/postinst
+#!/bin/sh
+$(call GluonCheckSite,check_site.lua)
```
NeoRaider: Where is this check_site.lua?

package/gluon-client-bridge/Makefile
```diff
@@ -0,0 +1,36 @@
...
+
+PKG_BUILD_DIR := $(BUILD_DIR)/$(PKG_NAME)
+
+include $(GLUONDIR)/include/package.mk
+
+define Package/gluon-client-bridge
+  SECTION:=gluon
+  CATEGORY:=Gluon
+  TITLE:=Provides a bridge and a wireless interface for clients to connect to
+  DEPENDS:=+gluon-core +gluon-radio-config
```
NeoRaider: Can we somehow loosen the dependency on gluon-radio-config? Optimally, it should be possible to create a firmware containing gluon-client-bridge and maybe even gluon-batman-adv-* without gluon-radio-config if no WLAN support is needed.

tcatm: Yes, we could do that. However, I'm not sure which package would depend on gluon-radio-config in that case. We could add it to site.mk.

NeoRaider:
Rebased and committed to master.

PR #488, opened 2015-09-10
Documentation added Python dependency

indigo423:
If you try to build with Python 3 you run in build errors. Python 2 has to be the system default.


indigo423:
Wrong branch

PR #489, opened 2015-09-10
Update build dependency documentation

indigo423:
Build requires Python 2 and libssl-dev

### DIFF
--- a/docs/user/getting_started.rst
+++ b/docs/user/getting_started.rst
@@ -27,12 +27,13 @@ freshly installed Debian Wheezy system the following packages are required:
 
 * `git` (to get Gluon and other dependencies)
 * `subversion`
+* `python2` (Python 3 doesn't work)
 * `build-essential`
 * `gawk`
 * `unzip`
 * `libncurses-dev` (actually `libncurses5-dev`)
 * `libz-dev` (actually `zlib1g-dev`)
-
+* `libssl-dev`
 
 Building the images
 -------------------


tcatm:
This should go into master. Commits to release branches get cherry-picked from there.

indigo423:
Did the first pull request against master and guys in IRC explained, it doesn't fit for master and the v2015.1.x branch is the correct branch for this pull request. What should I do?

jplitza:
Who was "guy in IRC"? Anyway, he was probably wrong regarding the branch:
The documentation goes to gluon.readthedocs.org, where you cannot select branches as documentation versions, only tags. While the issue you are raising (unmentioned build requirements) is valid for all past versions, tags cannot (and will not) be changed, so the documentation of those past versions will never contain the change. In contrast, if the change is applied to the master branch, it will soon be reflected in the "latest" version of the documentation.

(Also, I was surprised that libssl-dev is needed, but it indeed is.)

NeoRaider:
libssl-dev is a new dependency of Chaos Calmer, so it isn't needed for Gluon 2015.1.x. Not sure about python, I can't remember installing it explicitly on my build machines...

jplitza:
Yes, Python really is required by the [openwrt/include/prereq-build.mk](http://git.openwrt.org/?p=15.05/openwrt.git;a=blob;f=include/prereq-build.mk;h=211201af3da1e0aaf85a2b3701f78c3bd0d6f499;hb=HEAD#l139).

jplitza:
Actually, the sentence above the list of dependencies reads
> On a freshly installed Debian Wheezy system the following packages are required:

Now Python 2 is not only pre-installed in Debian Wheezy, it also isn't called `python2` as a package. However, I think listing
* `python` (Python 3 doesn't work!)

as a dependency is fine even though `python` is pre-installed. I commited to master an adapted version fixing that and a whitespace error that was introduced by this PR in 3f72e68. This can thus closed.

PR #491, opened 2015-09-12
Add support for x86-xen_domu target

mruettgers:


targets/x86-xen_domu/profiles.mk
```diff
@@ -0,0 +1,6 @@
+X86_XEN_MODULES := 
+
+$(eval $(call GluonProfile,GENERIC,$(X86_XEN_MODULES)))
```
NeoRaider: Please just remove the second argument to GluonProfile instead of defining an unused variable.

targets/x86-xen_domu/config
```diff
@@ -0,0 +1,2 @@
+CONFIG_TARGET_x86=y
+CONFIG_TARGET_x86_xen_domu=y
```
NeoRaider: The file `config` can be deleted, the target and subtarget options are set implicitly by Gluon for some time.

mruettgers:
The issues should be solved now.

PR #492, opened 2015-09-12
docs: Change suggested wifi24 channel to 6

rubo77:
TP-Link has a lower txpower on channel 1 and 11

### DIFF
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -88,7 +88,7 @@ wifi24
     ::
 
        wifi24 = {
-         channel = 11,
+         channel = 6,
          htmode = 'HT20',
          ap = {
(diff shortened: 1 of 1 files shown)


rubo77:
see https://forum.freifunk.net/t/tp-link-sendet-auf-kanal-1-11-schwacher/623/2

I suggest 6 as default, to stay in the context 1,6,11.

ohrensessel:
In Germany the scheme is 1,5,9,13 --> suggesting channel 5 would be better (as long as this really has an influence on the txpower)

tcatm:
As this is just an example config the specific channel used here does not matter. It may be much more useful to explain how to choose a channel (for both 2.4 and 5 ghz).

PR #493, opened 2015-09-12
docs: Suggest MTU 1280 to circumvent Kabel Deutschland problems

rubo77:


tcatm:
This should be reflacted in `site-example/site.conf`, too.

jplitza:
Why 1280 and not 1406? The latter would allow clients to have another (private) tunnel using IPv6 inside the "forced" fastd tunnel.

tcatm:
Nested tunnels can be fragmented (either by batman or in layer3 meshes using normal UDP fragmentation).

jplitza:
Putting aside my experiences from yesterday, where precisely the fragmentation apparently didn't work, it needlessly impairs performance of the inner tunnel. Also, there was some rationale to use 1426 in the original example, wasn't there? It could have been 1280 from the beginning, but it wasn't. 😉

rubo77:
I amended the change in site-example/site.conf too.

PR #494, opened 2015-09-12
docs: Suggest peer limit 1 instead of 2

rubo77:


tcatm:
This should also change site-example/site.conf accordingly.

rubo77:
I amended the same in site-example/site.conf, too.

do9xe:
Can you provide a reason, why we should change this? Have the nodes connect to two gateways works fine.

rubo77:
The reason for two uplinks was the worst case, where a gateway crashes often and the peers should have a backup link. I heard, in that case switching to the backup gateway willstill take som 5 seconds.
And if there was no 2nd uplink the node will take about 20 seconds to decide to switch to any other new uplink.

So the gain would be **just some 15 seconds** if a gateway dies (which is not a condition we desire anyway)

The drawback we have every day when all is running fine and all gateways are up and running, is: A lot of more background-Traffic.

So the default suggestion should be `limit 1` and only if you plan to build really instable gateways, that die several times a day, you should set a higher limit

tcatm:
Great! Could you change the commit message in conformance with https://github.com/freifunk-gluon/gluon/blob/master/CONTRIBUTING.md#use-descriptive-commit-messages ?

PR #508, opened 2015-10-04
gluon-announce: statistics.loadinfo has 3 averages

lemoer:
It would be nice if all three values would be announced. For monitoring purposes the 5 and 15 min values are more interesting than the 1 min.

### DIFF
--- a/package/gluon-announce/files/lib/gluon/announce/statistics.d/loadavg
+++ b/package/gluon-announce/files/lib/gluon/announce/statistics.d/loadavg
@@ -1,1 +1,10 @@
-return tonumber(util.readline(io.open('/proc/loadavg')):match('^([^ ]+) '))
+local info = fs.readfile('/proc/loadavg')
+local loadavg = {}
+local a, b, c
+a, b, c = info:match('^([^ ]+) ([^ ]+) ([^ ]+)')
+
+loadavg['1'] = tonumber(a)
+loadavg['5'] = tonumber(b)
+loadavg['15'] = tonumber(c)
+
+return loadavg


tcatm:
This breaks backwards compatibility.

For monitoring purposes you can average five resp. 15 1-min values. You should use the uptime field to check the temporal intervals between your measurements.

NeoRaider:
Well, but what if you don't want to query your nodes every minute? In large networks, less frequent queries are preferable, so having the pre-averaged values as well might be nice.

But we should probably wait until we switch to NetJSON, as that will completely break compatiblity anyways.

tcatm:
You don't have to query that often. In fact, you can average 1-min values spaced by 5 minutes each to get a good grasp about the general load (which I assume was the goal).

lemoer:
That's what I am doing right now.. Here is an example: 

70-load-1d
(nearly all routers in the net show same results)

I think this does not reveal a lot about the routers load. I startet a second test by using the 5 min's average a few minutes ago. I will share my results here, when I collected enough data.

Maybe we could use a less breaking schema. What do you think about this schema?
``` javascript
{ "statistics": {
  ...
  "loadavg": 0.45,
  "loadavg-5": 0.12,
  "loadavg-15": 0.42,
  ...
} }
```

lemoer:
Here are the results:

70-load-1d_5avg

I think the differences are definitively visible.

tobby88:
I like the solution from lemoer. The second one won't break backwards compatibility while still serving the load-averages for 5 and 15 minutes - and they can be helpful for better statistics.

tcatm:
We'd like to postpone this until after we've switched to NetJSON. Could you take a look at the NetJSON spec and check whether it is possible to express "loadavg" in multiple intervals?

PR #513, opened 2015-10-09
Add support for WHR-HP-G300N

brueggemann:
Tested wireless, meshing, config mode, reset switch, lan ports. Everything works fine.


brueggemann:
Tested autoupdater and works fine, too

PR #517, opened 2015-10-11
Add support for Ubiquity Air Gateway

fpletz:


NeoRaider:
As far as I can tell, the correct filename is `ubiquiti-airgateway`, not `ubiquiti-air-gateway`. Please always test the autoupdater (or check at least that the filename matches the one expected by the updater) before opening a pull request.

fpletz:
@NeoRaider Thanks for pointing that out. Fixed and tested the autoupdater successfully. Other stuff like client wifi, mesh wifi and mesh on wan/lan has already been tested and works.

PR #525, opened 2015-10-14
Replace gluon-simple-tc by non-Gluon-specific simple-tc

NeoRaider:


tcatm:
The fact that the limit is applied to the VPN is lost by this patch.

NeoRaider:
@tcatm, I don't understand. All limits from `/etc/config/simple-tc` are applied in the hotplug.d handler (which is now in the `simple-tc` package in the `packages` repo).

PR #542, opened 2015-11-01
Timeout 30 is to less Change to 60

Delta1977:
in some cases ( with Tunneldigger) the Bat0 interfaces need more than 30 sec to come up. 
i think 60 Second is enough.


NeoRaider:
Init scripts are not run in parallel, this is holding up the whole boot. The timeout is already much too high, it should mainly cover the time needed for duplicate address detection.

We should rather find a way to get the interface up earlier (or even better: fix alfred so it can start when the interface is not ready) than making more workarounds.

PR #544, opened 2015-11-03
Update site.rst

picardtux:
new git url


NeoRaider:
The URL should be `https://git.nordwest.freifunk.net/ffnw/siteconf/tree/master`, so the whole site config is shown and not just site.conf.

NeoRaider:
Also please make the commit message a bit more specific...

NeoRaider:
I've squashed the two commits, improved the commit messages and applied this.

PR #545, opened 2015-11-04
docs: Update v2015.2 release notes and list of supported devices

CodeFetch:
Hope this will spare someone a little time for more important things.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -81,6 +81,7 @@ ar71xx-generic
   - WZR-HP-AG300H / WZR-600DHP
   - WZR-HP-G450H
 
+
 * D-Link
 
   - DIR-825 (B1)
@@ -103,10 +104,10 @@ ar71xx-generic
 
 * TP-Link
 
-  - CPE210 (v1)
-  - CPE220 (v1)
-  - CPE510 (v1)
-  - CPE520 (v1)
+  - CPE210 (v1, v1.1)
+  - CPE220 (v1, v1.1)
+  - CPE510 (v1, v1.1)
+  - CPE520 (v1, v1.1)
   - TL-MR3020 (v1)
   - TL-MR3040 (v1, v2)
   - TL-MR3220 (v1, v2)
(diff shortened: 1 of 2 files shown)


docs/index.rst
```diff
@@ -103,10 +104,10 @@ ar71xx-generic
...
 * TP-Link
 
-  - CPE210 (v1)
-  - CPE220 (v1)
-  - CPE510 (v1)
-  - CPE520 (v1)
+  - CPE210 (v1, v1.1)
+  - CPE220 (v1, v1.1)
+  - CPE510 (v1, v1.1)
+  - CPE520 (v1, v1.1)
```
NeoRaider: These should be changed to "v1.0, v1.1"

docs/index.rst
```diff
@@ -123,17 +124,20 @@ ar71xx-generic
...
+  - TL-WR710N (v1, v2)
   - TL-WR740N (v1, v3, v4, v5)
-  - TL-WR741N/ND (v1, v2, v4, v5)
+  - TL-WR741ND (v1, v2, v4, v5)
   - TL-WR743N/ND (v1, v2)
   - TL-WR841N/ND (v3, v5, v7, v8, v9, v10)
   - TL-WR842N/ND (v1, v2)
-  - TL-WR941N/ND (v2, v3, v4, v5)
+  - TL-WR901ND (v1)
+  - TL-WR940N (v1, v2, v3)
```
NeoRaider: v1 and v2 are still marked BROKEN because they are untested, so they should not appear in the support list.

docs/releases/v2015.2.rst
```diff
@@ -1,17 +1,31 @@
...
+  - TL-WR941N (v6)
+  
+* Ubiquity
+
+  - Air Gateway
 
 New features
 ~~~~~~~~~~~~
-
+802.11s can now be used as an alternative to IBSS mode (ad-hoc). The statuspage has been reworked completely. French language support has been added. The numbers of clients connected are announced per band now (2.4/5GHz).
```
NeoRaider: I think you can leave the "New features" and "Bugfixes" sections to us. We want the explanations to be a bit more detailed.

docs/releases/v2015.2.rst
```diff
@@ -47,14 +62,12 @@ Site changes
         },
       }
 
-    The ``htmode`` option has been dropped, the channel width is now always set to 20MHz
-    (see https://github.com/freifunk-gluon/gluon/issues/487 for a discussion of this change).
-
```
NeoRaider: Please don't change that; the htmode option is gone and won't return. It is possible that we'll add support for higher channel widths again in the future, but it is not planned to make it configurable.

CodeFetch:
You just really irritated me, because I was the one who wrote the OpenWrt and Wikidevi articles stating that the hardware of the TL-WR940N v1, v2 is equal to the TL-WR941ND v4, v5.
I once compared the PCB-pictures and was pretty sure that it's the same hardware, but now I'm 100% sure as I just compared the MD5 hashes of the stock firmware images and they are equal.

TL-WR940N v1: http://www.tp-link.com/resources/software/20112159232014.zip
TL-WR941N v4: http://www.tp-link.com/resources/software/201121516435819.zip
Both have MD5 hash ceae1b710c82ce81ce6dd0acedb0a3b5

The images for WR940N v2 and WR941ND v5 are byte-equivalent, too. They even have the same filename "wr940ndv2_wr941ndv5_en_3_16_9_up_boot(150311).bin".

So you can serenely remove the broken flags for TL-WR940N v1, v2, I think.

NeoRaider:
Makes sense, I'll remove the BROKEN for these devices.

PR #547, opened 2015-11-05
ar71xx: add support for TP-Link TL-WR1043ND v3

rotanid:
add support for new revision v3 of TP-Link TL-WR1043ND
it seems to be exactly the same hardware


NeoRaider:
While adding such a simple patch to v2015.1.x would usually be no problem, all new features must be added to the master first before they are backported. In this case I doubt we'll add it to v2015.1.x anymore, as we're going to release v2015.2 soon.

I can adjust the patch for the master and submit it to the OpenWrt upstream, can you provide a log of a boot on the 1043v3?

rotanid:
ok, here's the bootlog / dmesg:
https://gist.github.com/rotanid/6e3b1271792071a5e563

PR #551, opened 2015-11-08
contrib/sign.sh: Let it fail if ecdsasign fails

corny:
Before this change sign.sh succeeded with exit status zero, even if ecdsasign was not installed.

### DIFF
--- a/contrib/sign.sh
+++ b/contrib/sign.sh
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+set -e
+
 if [ $# -eq 0 -o "-h" = "$1" -o "-help" = "$1" -o "--help" = "$1" ]; then
 	cat <<EOHELP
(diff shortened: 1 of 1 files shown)


NeoRaider:
I'd prefer these to just use `set -e` like some other scripts, so it will behave the same when the interpreter `sh` is called manually with the script files.

corny:
Ok, I updated my pull request.

NeoRaider:
Looking at the changes again, I don't think simply setting -e would make sense here.

sigtest.sh already checks the return code of ecdsaverify, and it would be nice if sign.sh would clean up (delete $lower, $upper and $manifest) even if it fails.

corny:
Ok, i removed `set -e` from `sigtest.sh`.

NeoRaider:
I've now pushed a more complete fix.

PR #552, opened 2015-11-08
gluon-mesh-batman-adv-core: keep disabled state even when the mesh interface section name is changed

### LINKED ISSUE #549: IBSS re-enabled on 2015.2 upgrade
After upgrading a node from 2015.1.x with IBSS disabled to 2015.2 (current master), the IBSS mesh mode gets activated. 

The 802.11s mesh mode is disabled in `site.conf`.

This is caused by https://github.com/freifunk-gluon/gluon/commit/2a93c580428d10724116b0d2d1238e2745715a14 which changed the interface name from `mesh` to `ibss`.

/cc @tcatm

NeoRaider:
Is makes sense to always look for both ibss_radio* and mesh_radio* sections
to determine if the meshing should be enabled when regenerating these
sections. Doing this, the disabled state will survive updates changing the
section name (either updating from pre-2015.2 while keeping IBSS, or
changing from IBSS to 11s or vice-versa).

If both ibss_radio* and mesh_radio* sections exist, the disabled state will
be kept correctly for each section, the behaviour is changed only when
creating a section that didn't exist before.

Fixes #549

### DIFF
--- a/package/gluon-mesh-batman-adv-core/files/lib/gluon/upgrade/320-gluon-mesh-batman-adv-core-wireless
+++ b/package/gluon-mesh-batman-adv-core/files/lib/gluon/upgrade/320-gluon-mesh-batman-adv-core-wireless
@@ -6,18 +6,24 @@ local util = require 'gluon.util'
 local uci = require('luci.model.uci').cursor()
 
 
-local function is_disabled(config, name)
-  local disabled = config and config.disabled
+local function is_disabled(name)
   if uci:get('wireless', name) then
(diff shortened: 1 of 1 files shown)


tcatm:
This looks okay but I have found the call to `first_non_nil()` a little confusing to read due to the nested `(config.mesh or {})`. Could you add a comment explaining the details?

NeoRaider:
I've updated the pull request with a few comments.

fpletz:
This works and fixes #549 for me. :+1: Thanks!

PR #561, opened 2015-11-16
Update getting_started.rst

Nurtic-Vibe:
OpenSSL dependency missing


tcatm:
We're pretty close to releasing 2015.2 and there won't be a 2015.1.3. Could you check whether this also applies to the current master and whether the patch is required?

Nurtic-Vibe:
This is already fixed in master branch.

PR #562, opened 2015-11-17
remove htmode from example

RubenKelevra:


NeoRaider:
Did you even read the text before the example? The example show how the config looked *before* Gluon 2015.2.

PR #571, opened 2015-11-24
Check for version of ecdsautils

alariel:
Check if the older multiple-binary verion is used, or if we have a single binary (newer version).


NeoRaider:
Okay, let's defer this until new ecdsautils actually have been released, till then, use the last release.

PR #572, opened 2015-11-24
sigtest.sh - check for version of ecdsautils

alariel:
See if we have the older multi-binary-version or if there is the newer single binary.

### DIFF
--- a/contrib/sigtest.sh
+++ b/contrib/sigtest.sh
@@ -27,9 +27,15 @@ awk "BEGIN    { sep=0 }
                 else       print > \"$lower\"}" \
     "$manifest"
 
+if [ -e '/usr/local/bin/ecdsautil' ]; then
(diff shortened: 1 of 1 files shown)


NeoRaider:
1. Hardcoding the path is no good, just use `which`
2. I think the break of compatiblity in ecdsautils is a problem. Instead of forcing users to adjust their commands, having symlinks and automatic detection of the desired command would be better (or at least scripts with the old names for compabitlity.) But that is an ecdsautils issue. @tcatm?
3. One pull request for both sign and sigtest would have been enough...

alariel:
Agreed, sorry for that - it's my first try with github. Could just get better, i think...

tcatm:
Well, there's no version of ecdsautils released using the new syntax yet. I've created https://github.com/tcatm/ecdsautils/issues/21 to track this issue. In the near future, I'd still like to deprecate the old commands, though.

PR #578, opened 2015-11-30
added how to copy images without symlinks

A-Kasper:


NeoRaider:
I don't like the wording, and I'm not sure if the Getting Started Guide is the best place for this... But I agree that having this information in our docs would be nice.

A-Kasper:
I noticed that this "how to" question recurred and just thought it would
help to add this.
I placed it in Getting Started, because the image-building process is
described here. If you build the images straight forward via the
documentation this would be the place you need this information. If you
find a better place feel free to move it.
My english is a little bit out of practice. Please don't hasitate to
rephrase and move it as you like.
It should just be a suggestion to add this information.

2015-12-08 23:14 GMT+01:00 Matthias Schiffer <notifications@github.com>:

tcatm:
For one, this should probably be formatted as a note:

```
.. note::

   Some images in the output/images ...

```

(See http://sphinx-doc.org/markup/para.html)

What's the usecase for this?

NeoRaider:
"note" is decribed as "An especially important bit of information"... and Sphinx renders it very big, so I don't think it makes sense to use it in this case.

NeoRaider:
I've added the info in 8bfb5fe1020facf13577a0ec9f80b8df4e054b5f.

PR #510, opened 2015-10-07
docs: enhanced site.conf documentation

rubo77:
Better default values and comments how to use certain config values. 
Thanks to https://github.com/ffmus/site-ffwaf/blob/master/site.conf

PR #580, opened 2015-12-04
docs: enhanced site.conf documentation and default values

rubo77:
Better default values and comments how to use certain config values.
Thanks to https://github.com/ffmus/site-ffwaf/blob/master/site.conf

spllited commits from #510


NeoRaider:
Thanks, I've taken the one commit I can agree with and commented the rest.

PR #581, opened 2015-12-04
Altitude in meters above the street

rubo77:
To achieve a homogeneous user input in all nodes add details about the reference for the altitude

### DIFF
--- a/package/gluon-config-mode-geo-location/files/lib/gluon/config-mode/wizard/0400-geo-location.lua
+++ b/package/gluon-config-mode-geo-location/files/lib/gluon/config-mode/wizard/0400-geo-location.lua
@@ -31,7 +31,7 @@ function M.section(form)
   o.datatype = "float"
   o.description = i18n.translatef("e.g. %s", "10.689901")
 
-  o = s:option(cbi.Value, "_altitude", i18n.translate("Altitude"))
+  o = s:option(cbi.Value, "_altitude", i18n.translate("Altitude above street level"))
   o.default = uci:get_first("gluon-node-info", "location", "altitude")
   o:depends("_location", "1")
   o.rmempty = true
--- a/package/gluon-config-mode-geo-location/i18n/de.po
+++ b/package/gluon-config-mode-geo-location/i18n/de.po
@@ -26,8 +26,8 @@ msgstr "Breitengrad"
 msgid "Longitude"
 msgstr "Längengrad"
 
-msgid "Altitude"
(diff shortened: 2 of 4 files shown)


NeoRaider:
I don't think we've had a consensus on this solution... and I don't like it at all.

Street level is not well-defined everywhere: What about buildings between streets with different levels? In Lübeck we have a Karstadt building (that actually had a few Freifunk nodes some years ago) that has an "upper ground floor" and a "lower ground floor". Also, small errors in the lat/lon coordinates may cause significant errors in the street level lookup if we were ever to actually do anything with the height value...

Adorfer:
i vote "+" for this PR. 
reason: streetlevel is more likely to be clear, obvious and precise than the current UELN value. 
(most people think if asked about height will think about "Normalnull" as if asked for weight ans responding in pounds.)
Off course a UELN would be gread it given. But due to the fact that it's likely that people will use GNSS-values: Those suffer by design from a DOP in z-axis about 3-10 times the lat/lon dilution. 
meaning: if coordinates are exact by 5m in lat/lon, you will have a height with precision of 40m. 
This makes it possibly totally useless in an area with regular buidings up to 30 metres.

rubo77:
We could add a passus like:

> in case there are different street levels around the house, take the average of the nearest side "

tcatm:
Maybe we could add a note below the input field:

   Altitudes are not yet well defined. Please consult with your community when filling in this information.

NeoRaider:
Hmm, something like that... the word "community" is too Freifunk-specific though.

kb-light:
Maybe one can add some configuration in the site.conf, whether this field should be displayed or not (we actually do not show this field). And a possibility to rename this field in the i18n-files, so every "community" has the ability to rename it.

rubo77:
Maybe we should completely change the **altitude** field into **floor** (Stockwerk, Stockwhich is well defined im most buildings

rubo77:
Rename Altitude to Storey #872

jplitza:
See [my comment in #872](https://github.com/freifunk-gluon/gluon/pull/872#issuecomment-243680860) ⇒ closing.

rubo77:
Ia it possible already to define the Label for altitude in the site/i18n Folder?

NeoRaider:
No, having multiple translations for the same field will yield undefined results. We'd need to remove the current strings and replace them with a placeholder.

rubo77:
https://github.com/freifunk-gluon/gluon/issues/873

PR #584, opened 2015-12-06
OpenSSL: Added source/old to PKG_SOURCE_URL

Ranlvor:
OpenSSL moves old versions of the library from http://www.openssl.org/source/
to http://www.openssl.org/source/old/1.0.2/ breaking the old links. That
behavior breaks the OpenWRT-build every time OpenSSL releases a new version.

This patch adds http://www.openssl.org/source/old/1.0.2/ to the PKG_SOURCE_URL
of OpenSSL to avoid breaking the build whenever OpenSSL releases a new
version.

### DIFF
--- a/patches/openwrt/0043-OpenSSL-Added-source-old-to-PKG_SOURCE_URL.patch
+++ b/patches/openwrt/0043-OpenSSL-Added-source-old-to-PKG_SOURCE_URL.patch
@@ -1,0 +1,24 @@
+From: Ranlvor <ranlvor@starletp9.de>
+Date: Sun, 6 Dec 2015 16:12:55 +0100
+Subject: OpenSSL: Added source/old to PKG_SOURCE_URL
+
+OpenSSL moves old versions of the library from http://www.openssl.org/source/
+to http://www.openssl.org/source/old/1.0.2/ breaking the old links. That
+behavior breaks the OpenWRT-build every time OpenSSL releases a new version.
+
+This patch adds http://www.openssl.org/source/old/1.0.2/ to the PKG_SOURCE_URL
+of OpenSSL to avoid breaking the build whenever OpenSSL releases a new
+version.
+
+diff --git a/package/libs/openssl/Makefile b/package/libs/openssl/Makefile
+index 7f0da8b..039e1ab 100644
+--- a/package/libs/openssl/Makefile
++++ b/package/libs/openssl/Makefile
+@@ -16,6 +16,7 @@ PKG_BUILD_PARALLEL:=1
+ 
+ PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+ PKG_SOURCE_URL:=http://www.openssl.org/source/ \
++	http://www.openssl.org/source/old/1.0.2/ \
+ 	ftp://ftp.funet.fi/pub/crypt/mirrors/ftp.openssl.org/source \
+ 	ftp://ftp.sunet.se/pub/security/tools/net/openssl/source/
+ PKG_MD5SUM:=38dd619b2e77cbac69b99f52a053d25a


rotanid:
good idea, i would also like to have your patch included in gluon

LeSpocky:
This is not as generic as it could be. I'm just trying with the following patch on OpenWrt master:

```
diff --git a/package/libs/openssl/Makefile b/package/libs/openssl/Makefile
index faf6816..f41f092 100644
--- a/package/libs/openssl/Makefile
+++ b/package/libs/openssl/Makefile
@@ -8,7 +8,9 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=openssl
-PKG_VERSION:=1.0.2e
+PKG_BASE:=1.0.2
+PKG_BUGFIX:=e
+PKG_VERSION:=$(PKG_BASE)$(PKG_BUGFIX)
 PKG_RELEASE:=1
 PKG_USE_MIPS16:=0
 
@@ -16,6 +18,7 @@ PKG_BUILD_PARALLEL:=0
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=http://www.openssl.org/source/ \
+   http://www.openssl.org/source/old/$(PKG_BASE)/ \
    ftp://ftp.openssl.org/source/ \
    ftp://ftp.funet.fi/pub/crypt/mirrors/ftp.openssl.org/source \
    ftp://ftp.sunet.se/pub/security/tools/net/openssl/source/
```

This is more or less how it's done in ptxdist.

LeSpocky:
Oh I see you already tried: https://lists.openwrt.org/pipermail/openwrt-devel/2015-December/037714.html ;-)

Ranlvor:
@LeSpocky Well, take a very close look at the ordering of the lines in the CC-Branch vs. the master-Branch. They are not the same. Took me quite a while to notice that while trying to figure out why I was not able to write a patch which applies cleanly to CC and master.

oni303:
Please merge this, it fixes an Error.
It is not as generic as possible but it fixes the Error and hopefully openwrt will accept the more generic patch. If they do, this patch is obsolete after an openwrt update anyway. So do not  waste time making this patch more generic.

LeSpocky:
@oni303 There's still the workaround to put the openssl tarball manually into the dl folder, so this is neither an security issue nor a major breakage, no need to fix this ASAP.

oni303:
@LeSpocky  Well integrating this workaround into  a Jenkins build server is actually painful.
Besides, having at least one branch that compiles without error should be one of the goals of  every project. Especially if this state is just two click ahead.

NeoRaider:
I like this fix, I hope it will also get merged upstream.

pinguinpfleger:
Wouldn't it be better to count up the revision number automatically instead of using outdated versions with bugs or security issues.
For example if 1.0.2d was the latest stable and meanwhile it's 1.0.2e just trying to increment the letter.
https://www.openssl.org/policies/releasestrat.html

NeoRaider:
@pinguinpfleger, no, builds are supposed to be as reproducible as possible. The same Gluon commit id should always yield the same included package versions.

jplitza:
Also, regarding the security aspect, please remember that OpenSSL isn't used for anything really critical in (stock) Gluon. It is just used to sign kernel modules and check their signatures on load. It isn't exposed via any network service or anything.

oni303:
Thanks, the build is finishing without error again.

NeoRaider:
Actually, it is not used at all. It is a build dependency of nodejs, which is used to build the status page...

pinguinpfleger:
thanks for the clarification. If it is not used at all, I'm conciliated.
(But if it is not used at all we could also use the latest stable version anyhow.)

LeSpocky:
Meanwhile integrated upstream: https://github.com/openwrt/openwrt/commit/61176282400e868e1c9174ebba460b4c6130dae4 :-)

PR #599, opened 2015-12-12
Status page load reduction

NeoRaider:
TODO:

* [x] SSE multiplexer
* [x] Stop reading batadv originators?


NeoRaider:
Depends on the sse-multiplex addition.

NeoRaider:
I've also added some more fixes for gluon-neighbour-info. The first one is actually related to sse-multiplex, the other two are general bugfixes.

tcatm:
I don't think we can apply 'gluon-mesh-batman-adv-core: remove neighbours.d/batadv' without a fallback as that is what a future ffmap-backend should use instead of alfred.

NeoRaider:
@tcatm, I guess having a script pull the originators out of the kernel to /tmp periodically (batadv-vis does this every 10s) and changing alfred, batadv-vis, gluon-status-page-api and the announced script to use this copy would allow us to keep most of our current features, including the TQ display on the status page.

NeoRaider:
Updated.

PR #603, opened 2015-12-15
Updated Patch 0043 - OpenSSL: Added source/old to PKG_SOURCE_URL

Ranlvor:
to match upstream.

The patch was accepted upstream in
Changeset 47860: https://dev.openwrt.org/changeset/47860
Git commit openwrt/openwrt@6117628 / https://github.com/openwrt/openwrt/commit/61176282400e868e1c9174ebba460b4c6130dae4


NeoRaider:
I think it would be a good idea to keep the old version of the patch. It will need to be rebased on OpenSSL updates in the Chaos Calmer branch, so we should keep the differences as simple as possible (and especially not modify the lines defining the version number).

Ranlvor:
Ok. I thought that a patch that is equal to a upstream commit makes things easier instead of harder. But I do not know your exact workflow anyway.

PR #607, opened 2015-12-30
Convert site.conf to JSON during build

NeoRaider:
No functional changes so far.


NeoRaider:
Note: as this fixes an issue with the Lua host installation, at least `make package/lua/host/clean` is necessary to make this build.

corny:
I would prefer to have a site.yml instead of a site.conf.
What is your opinion on that?

NeoRaider:
@corny, yes, sounds like a good idea, but that will have to wait until after the v2016.1 release.

corny:
Ok, great! :+1:

PR #612, opened 2016-01-03
Convert gluon-announce Lua code to C modules

NeoRaider:


NeoRaider:
I'd like to merge the c-respondd ASAP. I've verified that its output is identical to the old code, and there don't seem to be any memory leaks.

package/libgluon/src/libgluon.c
```diff
@@ -0,0 +1,111 @@
...
+                if (*in != ':')
+                        *out++ = *in;
+        } while (*in++);
+
+        return node_id;
+}
+
+char * gluon_get_interface_address(const char *ifname) {
+        const char *format = "/sys/class/net/%s/address";
+        char path[strlen(format) + strlen(ifname)];
```
jplitza: Shouldn't this be `strlen(format) + strlen(ifname) - 1`? The `%s` is replaced by `ifname` (-2) but `\0` is appended (+1).

jplitza:
Some thoughts on package naming:
* libgluon is quite general and high-profile for some helper functions. Maybe libgluonutil or libgluon-util are better?
* gluon-announce now really is a misnomer, as it no longer contains any code related to announcing anything. Instead, it simply is a basic respondd provider like every other. The only thing special about it is that it provides the cache periods (which could maybe simply be included in gluon-announced instead?)
* gluon-announced could be renamed to gluon-respondd, to emphasize that it only starts respondd.

In fact, we can now probably merge gluon-announce and gluon-announced, as using the first without respondd is impossible, as is using respondd without the latter.

package/libgluon/src/libgluon.c
NeoRaider: You're right. I pondered about this a while when writing this, as it obviously doesn't matter if we reserve 1 byte more, but I should've at least added a comment. Thinking about this again, using the exact size is probably less confusing.

NeoRaider:
I agree that gluon-announce and gluon-announced can be merged to gluon-respondd.

I'm not sure about libgluon. libgluonutil is probably a better name, I'll rename it.

jplitza:
What does commit b13e233 do in this PR? But now that it is there: Didn't you object to stopping respondd during autoupdate?

NeoRaider:
@jplitza, well, it doesn't really belong in this PR, but it is based on the other patches; I think I'll remove it again before merging and merge it individually. (And Github shows the patch order incorrectly O_o).

No, I didn't object to stopping respondd at all.

jplitza:
Sorry, my memory betrayed me. I thought you wanted to have the node behave mostly normally during download, but that was with regard to everything *except* the mapping services.

Yes, Github shows the commits in a PR ordered randomly, it's really annoying.

package/gluon-mesh-vpn-fastd/src/respondd.c
```diff
@@ -0,0 +1,301 @@
...
+#include <unistd.h>
+
+#include <sys/socket.h>
+#include <sys/un.h>
+
+
+static struct json_object * get_peer_groups(struct json_object *groups, struct json_object *peers);
+
+static struct json_object * get_fastd_version(void) {
+        FILE *f = popen("exec fastd -v", "r");
```
jplitza: Would it be too much work to do the fork ourselves and execute fastd directly, instead of spawning a shell in between?

NeoRaider: I think it isn't worth the hassle...

package/gluon-mesh-vpn-fastd/src/respondd.c
```diff
@@ -0,0 +1,301 @@
...
+
+        char *line = NULL;
+        size_t len = 0;
+
+        ssize_t r = getline(&line, &len, f);
+
+        pclose(f);
+
+        if (r >= 0) {
+                len = strlen(line);
```
jplitza: I was confused, why you don't simply use the return value of `getline()`. Maybe add a comment explaining that?

package/gluon-mesh-vpn-fastd/src/respondd.c
```diff
@@ -0,0 +1,301 @@
...
+
+                if (len && line[len-1] == '\n')
+                        line[len-1] = 0;
+        }
+        else {
+                free(line);
+                line = NULL;
+        }
+
+        struct json_object *ret = gluonutil_wrap_string(line + 6);
```
jplitza: I would like to check if `strncmp(line, "fastd ", 6) == 0` before stripping the first six bytes. If you want to do this unconditionally, at least add a comment that the first six bytes are "fastd ".

package/gluon-mesh-vpn-fastd/src/respondd.c
```diff
@@ -0,0 +1,301 @@
...
+        free(line);
+        return ret;
+}
+
+static struct json_object * get_fastd(void) {
+        bool enabled = false;
+
+        struct uci_context *ctx = uci_alloc_context();
```
jplitza: I'd propose packaging `uci_alloc_context()`, the setting of the strict flag and `uci_load()` into a libgluonutil function. It occurs four times in this PR, and especially the setting of the strict flag could easily be forgotten otherwise.

package/gluon-node-info/src/respondd.c
```diff
@@ -0,0 +1,144 @@
...
+        json_object_object_add(ret, "contact", gluonutil_wrap_string(contact));
+        return ret;
+}
+
+static struct json_object * get_system(struct uci_context *ctx, struct uci_package *p) {
+        struct json_object *ret = json_object_new_object();
+
+        const char *role = get_first_option(ctx, p, "system", "role");
+        if (role && *role)
+                json_object_object_add(ret, "role", gluonutil_wrap_string(role));
```
jplitza: Why does `get_system()` behave differently from `get_owner()` and `get_location()` in that it returns an empty JSON object instead of NULL when the corresponding setting isn't set?

package/gluon-respondd/files/etc/init.d/gluon-respondd
```diff
@@ -0,0 +1,45 @@
...
+SERVICE_WRITE_PID=1
+SERVICE_DAEMONIZE=1
+
+DEVLIST=/var/run/gluon-respondd.devs
+DAEMON=/usr/bin/respondd
+LOCK=/var/run/gluon-respondd.lock
+
+
+do_start() {
+	DEVS=$(cat $DEVLIST 2>/dev/null | while read dev iface; do echo -n " -i $dev"; done)
```
jplitza: No need to use `cat` here, simply do `while read dev iface; do ...; done < $DEVLIST`

package/gluon-respondd/files/etc/hotplug.d/iface/10-gluon-respondd
```diff
@@ -0,0 +1,34 @@
...
+ifname_to_dev () {
+	json_load "$(ubus call network.interface.$1 status)"
+	json_get_var dev device
+
+	echo "$dev"
+}
+
+case "$ACTION" in
+	ifdown)
+		sed "/ $INTERFACE$/d" $DEVLIST > $DEVLIST.new
```
jplitza: Why the temporary `.new` file instead of `sed -i`?

package/gluon-respondd/files/lib/gluon/respondd/neighbours.cache
```diff
@@ -0,0 +1 @@
+10000
```
jplitza: Didn't [you say](https://github.com/freifunk-gluon/packages/pull/125#discussion_r49883003) you want to change neighbours and statistics cache time to 900ms?

package/gluon-respondd/src/respondd.c
```diff
@@ -0,0 +1,209 @@
...
+
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+
+#include <sys/utsname.h>
+#include <sys/vfs.h>
+
+
+static struct json_object * gluonutil_version(void) {
```
jplitza: Why is this function named `gluonutil_*` when it isn't in libgluonutil?

package/gluon-core/files/lib/gluon/upgrade/040-site-code
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/lua
+
+local sysconfig = require 'gluon.sysconfig'
+local site = require 'gluon.site_config'
+
+sysconfig.site_code = site.site_code
```
jplitza: Where is this needed/where was this done before? It seems to be totally unrelated to everything else here.

package/gluon-mesh-batman-adv-core/src/respondd.c
```diff
@@ -0,0 +1,606 @@
...
+                json_object_put(val);
+}
+
+static bool interface_file_exists(const char *ifname, const char *name) {
+        const char *format = "/sys/class/net/%s/%s";
+        char path[strlen(format) + strlen(ifname) + strlen(name)];
+        snprintf(path, sizeof(path), format, ifname, name);
+
+        struct stat s;
+        return !stat(path, &s);
```
jplitza: Why not use `!access(path, F_OK)` instead of `!stat(…)`? That wouldn't require an additional temporary struct.

package/gluon-mesh-batman-adv-core/src/respondd.c
```diff
@@ -0,0 +1,606 @@
...
+
+}
+
+static struct json_object * get_mesh_subifs(const char *ifname) {
+        struct json_object *wireless = json_object_new_array();
+        struct json_object *tunnel = json_object_new_array();
+        struct json_object *other = json_object_new_array();
+
+        const char *format = "/sys/class/net/%s/lower_*";
+        char pattern[strlen(format) + strlen(ifname)];
```
jplitza: [Again](https://github.com/freifunk-gluon/gluon/pull/612#discussion_r49765063), this should be `- 1`.

package/gluon-mesh-batman-adv-core/files/lib/gluon/announce/nodeinfo.d/network/mesh_interfaces
```diff
@@ -1,13 +0,0 @@
...
-
-for ifname in batman_adv.interfaces('bat0') do
-  pcall(
-    function()
-      table.insert(interfaces, util.readline(io.open('/sys/class/net/' .. ifname .. '/address')))
-    end
-  )
-end
-
-return interfaces
```
jplitza: `nodeinfo.network.mesh_interfaces` isn't generated by the C version anymore. Obviously, the same information is contained in `nodeinfo.network.mesh.bat0.interfaces`, but in other form. Is this intended? Is it documented? Do we have something like API versions for what is returned by respondd?

package/gluon-mesh-batman-adv-core/src/respondd.c
```diff
@@ -0,0 +1,606 @@
...
+
+                json_object_object_add(obj, "gateway", json_object_new_string(addr));
+                break;
+        }
+
+        free(line);
+        fclose(f);
+}
+
+static inline bool do_ioctl(int fd, struct ifreq *ifr, void *data) {
```
jplitza: I think `do_ethtool_ioctl` would be a better name.

jplitza:
`package/gluon-mesh-batman-adv-core/src/respondd.c` has horribly mixed indentation, sometimes using spaces and sometimes tabs. All other `respondd.c` files use spaces exclusively, while respondd itself uses tabs. I could live with the discrepancy between the modules and the daemon, end even between the modules, but inside this same file, you really should settle with one of the options.

jplitza:
I have to admit, I currently don't have the leisure to dig into the documentation for iwinfo and ethtool, which means I didn't fully review the respondd.c from gluon-mesh-batman-adv-core yet. Also, I'm no good with Makefiles, so I didn't really review them either.

However, spontaneous tests with the compiled firmware looked good too. So modulo my comments, consider this ACK'ed.

package/gluon-mesh-vpn-fastd/src/respondd.c
NeoRaider: Could be done, but I think it's fine as is for now.

package/gluon-node-info/src/respondd.c
NeoRaider: This reflects the behaviour of the old Lua snippets. The `system` field should always exist, even when it is empty, while owner and location just don't exist when there's no data.

package/gluon-respondd/files/etc/init.d/gluon-respondd
NeoRaider: I think there's no way to hide the error message when $DEVLIST doesn't exist without the cat...

package/gluon-respondd/files/etc/hotplug.d/iface/10-gluon-respondd
NeoRaider: The specification of sed doesn't guarantee atomicity for `sed -i`. I think most (all?) current implementions use a temporary file for it, but making this explicit is cleaner.

package/gluon-respondd/files/lib/gluon/respondd/neighbours.cache
NeoRaider: That's the plan, but I'd like to get some more real-life testing of the code before lowering the cache times, so this will be a post-2016.1 change.

package/gluon-respondd/src/respondd.c
NeoRaider: Oops. Too much search&replace.

package/gluon-core/files/lib/gluon/upgrade/040-site-code
NeoRaider: This piece of code was added before we decided to convert site.conf to JSON (so creating a sysconfig file was the only way to get the site code from C). I forgot to remove it when rebasing...

package/gluon-mesh-batman-adv-core/files/lib/gluon/announce/nodeinfo.d/network/mesh_interfaces
NeoRaider: `nodeinfo.network.mesh_interfaces` had been deprecated when `nodeinfo.network.mesh.bat0.interfaces` was added. This was before 2015.1, and the commit message mentioned that `nodeinfo.network.mesh_interfaces` would be removed after 2015.1.

At the moment, we don't have API versions or proper documentation of the API. We plan to switch to the well-documented [NetJSON](http://netjson.org/) format some time in the future (although we'd still have Gluon-specific extensions to NetJSON which would need to be documented...)

package/gluon-mesh-batman-adv-core/src/respondd.c
NeoRaider: I'll add an `ethtool_` prefix to a few functions. I've copied the code from our lua-ethtool module we used so far, where ethtool prefixes didn't make much sense as everything was ethtool ;)

NeoRaider:
The indentation issues were a consequence of me switching from Emacs to Atom recently, which I haven't fully mastered yet... I've changed everything to tabs, as it was intended.

Thanks again for your detailed review. I think I've addressed all your comments, so I'll merge it now after a last round of testing.

PR #613, opened 2016-01-03
Display QR-Code on Reboot Page

Philhil:
When rebooting the node in config mode, there are often problems to send the pubkey via email when you have no internet connection.

Now you can display a QR-Code. The content could set in the i18n site configuration.
The QR-Code could be read and send via phone.


NeoRaider:
Nice feature. Some comments:
* It's a bit late in the release cycle, we'll merge it for 2016.1.1 or 2016.2.
* It would be nice to minimize the JS file
* Including the JS file could be made optional (by adding an additional package to site.mk). I don't think the theme package is the right place.
* The commit message of the first commit is a bit weird. The first line should explain what was done briefly, not why it was done.

NeoRaider:
I'd recommend moving the JS file into a new package, maybe called `gluon-luci-qrcode`. Instead of adding the QR code to the example files by default, you should add a manual for the package in `docs/package/gluon-luci-qrcode.rst`.

rotanid:
@Philhil are you planning on doing the requested changes so this can be merged?

mweinelt:
@Philhil Does closing represent a “no“?

Philhil:
right no i have not enough time :(

PR #620, opened 2016-01-13
omit station-dump if no IBSS-interface is present

0xDEAD:
If you manually deactivate meshing on wifi on a Freifunk-node, the status-page prints garbage because "iw dev $xxx station dump" is called with an $xxx = "" and iw returns its help page instead of the desired output.
Problem is that "for _, ifname in ipairs(interfaces) do" generates one loop-iteration, even if interfaces is empty, which an additional if can prevent


kb-light:
You mean the old status-page, right?
Which Gluon-Version did you use?
In v2015.1.x the iw help is returned by "iw dev $interface link", if type of $interface != ibss. In master this should be fixed since 753c1f9.

0xDEAD:
Urgh, damn, sorry, you are right.
I was using master-branch of FF-Munich's fork, which was not completely up to date :(

No need to merge this fix anymore!

0xDEAD:
I didn't check myself, but if you disable meshing on all interfaces (stupid but possible) this still might happen?

PR #625, opened 2016-01-17
change txpower for indoor 5GHz to max allowed value

oakey-dev:
i don't think we need this "conservative 100mW limit", but should use the allowed value of 200mW for the indoor chanels.
This should actually improve some 5GHz mesh connections.


oakey-dev:
will be first submitted to openwrt it self

PR #628, opened 2016-01-23
Add option to insert secret via keyboard input

mape2k:
Add option to insert secret via keyboard input to prevent storing private key on server. This did not break current behaviour and makes secret file optional. Also write variable "secret" in lowercase just like any other variable.

### DIFF
--- a/contrib/sign.sh
+++ b/contrib/sign.sh
@@ -2,17 +2,19 @@
 
 set -e
 
-if [ $# -ne 2 -o "-h" = "$1" -o "--help" = "$1" -o ! -r "$1" -o ! -r "$2" ]; then
+if [ $# -eq 0 -o $# -gt 2 -o "-h" = "$1" -o "--help" = "$1" -o ! -r "$1" -o \( $# -eq 2 -a ! -r "$2" \) ]; then
 	cat <<EOHELP
-Usage: $0 <secret> <manifest>
+Usage: $0 [<secret>] <manifest>
 
 sign.sh adds lines to a manifest to indicate the approval
 of the integrity of the firmware as required for automated
-updates. The first argument <secret> references a file harboring
-the private key of a public-private key pair of a developer
-that referenced by its public key in the site configuration.
-The script may be performed multiple times to the same document
-to indicate an approval by multiple developers.
+updates. The first optional argument <secret> references a
+file harboring the private key of a public-private key pair
+of a developer that referenced by its public key in the site
+configuration. If this parameter is missing, you will be
+asked to type in secret key. The script may be performed
+multiple times to the same document to indicate an approval
+by multiple developers.
 
 See also
  * edcsautils on https://github.com/tcatm/ecdsautils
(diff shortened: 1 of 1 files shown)


NeoRaider:
This doesn't make any sense. Download the manifest to your own system and sign it locally. If the secret should not be stored on the server, you should not transmit it to the server at all.

mape2k:
This may doesn't make any sense to you. Downloading manifest and sign it locally is not as simple as providing secret via ssh. The approach does not provide full security but it can be an approach instead of having key on server (which IMHO will surely be used by some guys).

rotanid:
well i somewhat agree with mape2k.
we solve it at the moment in a way that we store the key temporarily on the build server and remove it afterwards. but you could forget do delete it, so mape2ks solution would be better and more secure.
of course, there are even more secure ways to do it, but not all of them are applicable for everyone.

NeoRaider:
I'd rather say we should work on ways to allow signing on as many different systems as possible.

ecdsautils will be in the next versions of Debian and Ubuntu by default, a Fedora package seems to exist as well. If you have other systems you want to sign on, creating a signing workflow for them would be appreciated.

mape2k:
The scenario outlined by rotanid seems to be a common one. So the pull-request is a security enhancement ;)

Maybe a security note (on start) that you should use the script on a devs client (instead of on the server) could also be a good compromise to increase users consciousness of security.

jplitza:
I am still missing a full scenario where this would improve security.

As I currently see it, there are two different scenarios:
* *Automatic builds* like nightlies are performed automatically (no shit, sherlock!) on a build server. No human interference happens, and thus, this PR wouldn't improve anything.
* *Manual builds* can be done on either a local machine or on a build server. Besides the fact you have to trust the build server anyway, because otherwise it could create bogus builds with backdoors without you noticing, it isn't really much of an obstacle to download the manifest, sign it, and re-upload it, is it? It's not like you're releasing new firmwares every day.

Besides, if you really need this feature, you could probably simply pass `/dev/stdin` as secret file.

mape2k:
@jplitza: Have a look at what @rotanid says. This is a real scenario - already if it is not recommended. Downloading/Uploading manifest will sometimes to be too inconvenient. All signatories need ecdsautils on the working machine and are not all developers nor linux desktop users or familiar with building software (e.g. at least a tester has to sign the firmware). These one will copy the secret temporarily to the build server (as @rotanid confirmed). /dev/stdin as secret file is a good catch, usability maybe poor.

As this script is not an integral part of "building" firmware we now just use our script for ourselves and hope that @rotanid's scenario will not be used by too much communities.

rotanid:
there are various levels of trust in my opinion, not only "trusted" and "untrusted".
we (firmware signers in our community) trust each other, that we don't install things that might catch the key file on the few minutes it's saved on the build server - but we don't want to let there file stay there all the time, "Gelegenheit macht Diebe"
OT:
warum bitte diskutieren wir unter deutschen Muttersprachlern auf Englisch?

NeoRaider:
I don't see why we should go for a half-baked solution, instead of just making ecdsautils available for all relevant systems. I think using a precompiled ecdsautils version from a trusted source is still better than transmitting the key unnecessarily.

OT reply: Gluon is not only used by Germans. These discussions may be of interest for all Gluon users.

rotanid:
well, ok, almost convinced...

PR #630, opened 2016-01-25
Update sign.sh

Wlanfr3ak:
Change "ecdsasign" to "ecdsautil sign"

### DIFF
--- a/contrib/sign.sh
+++ b/contrib/sign.sh
@@ -35,7 +35,7 @@ awk 'BEGIN    { sep=0 }
                 else       print > "'"$lower"'"}' \
     "$manifest"
 
-ecdsasign "$upper" < "$SECRET" >> "$lower"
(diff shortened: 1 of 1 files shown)


jplitza:
1. ecdsautils still hasn't seen a release with the new program name. The upcoming packages in [Debian](https://packages.debian.org/testing/ecdsautils) and [Ubuntu](http://packages.ubuntu.com/xenial/ecdsautils) for example still feature the old syntax. And I would expect people to use either prebuilt packages or stable releases, not master branches from some Git. See also tcatm/ecdsautils#21 for a discussion about backward compatibility.
2. `contrib/sign.sh` isn't the only place where the ecdsautils are used. At least `contrib/sigtest.sh` also uses them, and if we wanted to "support" the master branch, we should consequently switch to it on the nodes as well and change the autoupdater accordingly.

PR #632, opened 2016-01-28
fix UBNT XM model detection

oakey-dev:


patches/openwrt/0048-fix-UBNT-XM-model-detection.patch
```diff
@@ -0,0 +1,73 @@
...
+ 	*"Bullet M")
+ 		name="bullet-m"
++		ubnt_xm_board_detect
+ 		;;
+ 	*"Loco M XW")
+ 		name="loco-m-xw"
+ 		;;
+ 	*"Nanostation M")
+ 		name="nanostation-m"
++		ubnt_xm_board_detect
```
NeoRaider: Please also add the new detection for the Rocket M, we'd like to switch to the proper Rocket images.

PR #634, opened 2016-01-29
Added Support for TP-Link WR843ND

Ranlvor:
Tests done to verify that this worked as intended:
 * checked plattform_info image_name
 * checked primary_mac
 * does appear in ffmap-d3
 * does mesh
 * does fastd-VPN


rotanid:
reading your Pull Request, i notice, that there are some models like your 843 and 842 and 1043 which have a "N/ND" in their description/comment, although there are only the ND ones available on TP-Links website and everywhere else.
if no one objects, i can prepare a pull request which changes this to "ND" only.

Ranlvor:
This N/ND-Stuff is weird. For example my code says

    tl-wr843nd-v1,tp-link-tl-wr843n-nd-v1

because the OpenWRT-Image is called nd-v1 while the autoupdater expects an image to be called n-nd-v1 according to the output of

To your proposed Pull Request: The software doen't realy seem to care about the N/ND-difference and so do I.

rotanid:
this is cosmetic, so i don't insist on changig that, if @NeoRaider voices reasons not to do it

PR #640, opened 2016-02-05
gluon-core: add preserve wifi channel feature

kokel:
**Old description:**
This changes introduces the new uci section 'gluon-core.wireless' with a preserve_channel option per radio:
 * preserve_channel_radio0 (boolean)
 * preserve_channel_radio1 (boolean)

By setting these options to true the wifi channel of the regarding radio will be preserved during upgrades.

**New description:**
This change introduces the new uci section 'gluon-core.wireless' with a preserve_channels option:
 * preserve_channels (boolean)

By setting this option to 1 (true) wifi channels will be preserved during upgrades.

### DIFF
--- a/docs/features/wlan-configuration.rst
+++ b/docs/features/wlan-configuration.rst
@@ -24,3 +24,11 @@ existing configuration will also be set for the new configuration.
 This allows upgrades to change from IBSS to 11s and vice-versa while retaining the
 "wireless meshing is enabled/disabled" property configured by the user regardless
 of the used mode.
+
+During upgrades the wifi channel of the 2.4GHz and 5GHz radio will be restored to the channel
+configured in the site.conf. If you need to preserve a user defined wifi channel during upgrades
+you can configure this via the uci section ``gluon-core.wireless``::
+
+  uci set gluon-core.@wireless[0].preserve_channels='1'
+
+Keep in mind that nodes running wifi interfaces on custom channels can't mesh with default nodes anymore!
--- a/package/gluon-core/files/lib/gluon/upgrade/200-wireless
+++ b/package/gluon-core/files/lib/gluon/upgrade/200-wireless
@@ -10,11 +10,21 @@ if not sysconfig.gluon_version then
   uci:delete_all('wireless', 'wifi-iface')
 end
 
+local function get_channel(radio, config)
+  if uci:get_first('gluon-core', 'wireless', 'preserve_channels') then
+    return uci:get('wireless', radio, 'channel') or config.channel
+  else
+    return config.channel
+  end
+end
+
 local function configure_radio(radio, index, config)
   if config then
+    local channel = get_channel(radio, config)
+
     uci:delete('wireless', radio, 'disabled')
 
-    uci:set('wireless', radio, 'channel', config.channel)
+    uci:set('wireless', radio, 'channel', channel)
     uci:set('wireless', radio, 'htmode', 'HT20')
     uci:set('wireless', radio, 'country', site.regdom)
   end
--- a/package/gluon-core/files/etc/config/gluon-core
+++ b/package/gluon-core/files/etc/config/gluon-core
@@ -1,0 +1,1 @@
+config wireless


Adorfer:
@kokel 
As a hotfix (to preserve the radio channel) you could do this: 

https://github.com/eulenfunk/firmware/blob/master/templates/dus/keepradiochannel.diff

tcatm:
I don't like `preserve_channel_radio0` as you have not much control of what `radio0` and `radio1` is (or whether it's `radioX` at all). Maybe a simpler approach like `preserve_channels` works better (or adding the options to the wireless config).

NeoRaider:
Well, the current code would work with any radio name, and as it is supposed to be set after flashing, the radio names would already be known. But I don't see a good usecase for setting this for each radio individually either...

package/gluon-core/files/lib/gluon/upgrade/200-wireless
```diff
@@ -10,11 +10,21 @@ if not sysconfig.gluon_version then
   uci:delete_all('wireless', 'wifi-iface')
 end
 
+local function get_channel(radio, config)
+  if uci:get_first('gluon-core', 'wireless', 'preserve_channel_' .. radio) and sysconfig.gluon_version and uci:get('wireless', radio) then
```
NeoRaider: This should not refer to sysconfig.gluon_version. Just return the current setting if there is a setting and preserve_channel is set; in any other case, return the default setting.

kokel:
Thanks for your feedback. I have merged your suggestions. For now there is one config option `gluon-core.wireless.preserve_channels` to preserve the wifi channel of all radios.

Please let me know if the alternate approach via wireless config (then this would be on a per radio basis) is cleaner or the preferred way, in your point of view. I don't care much about the appraoch, both are fine for me.

plumpudding:
Another great addition to this would be the possibility to turn this on via the site by default.

jplitza:
@plumpudding Why would that make any sense? The different channel would have to be configured via the command line or via an additional package anyway. In the same way would this setting be set.

kokel:
@plumpudding @jplitza I could imagine one possible use case for the site.conf switch ... the bofh in freifunk community x wants to reset all user defined wifi channels :stuck_out_tongue_closed_eyes: :innocent:

nazco:
Great feature!

docs/features/wlan-configuration.rst
```diff
@@ -24,3 +24,11 @@ existing configuration will also be set for the new configuration.
 This allows upgrades to change from IBSS to 11s and vice-versa while retaining the
 "wireless meshing is enabled/disabled" property configured by the user regardless
 of the used mode.
+
+During upgrades the wifi channel of the 2,4GHz and 5GHz radio will be restored to the channel
```
NeoRaider: Should be "2.4GHz", not "2,4GHz".

docs/features/wlan-configuration.rst
```diff
@@ -24,3 +24,11 @@ existing configuration will also be set for the new configuration.
...
+configured in the site.conf. If your use case is to be in need to preserve a user defined
+wifi channel during upgrades you can configure this via the uci section ``gluon-core.wireless``::
```
NeoRaider: Very weird grammar, just say: "If you need to preserve [...]"

package/gluon-core/files/lib/gluon/upgrade/200-wireless
```diff
@@ -10,11 +10,21 @@ if not sysconfig.gluon_version then
   uci:delete_all('wireless', 'wifi-iface')
 end
 
+local function get_channel(radio, config)
+  if uci:get_first('gluon-core', 'wireless', 'preserve_channels') then
+    return uci:get('wireless', radio, 'channel')
```
NeoRaider: I think the function should still fall back to config.channel if uci:get('wireless', radio, 'channel') returns nil, so previously unconfigured radios (maybe because of missing drivers... not supported at the moment, but we'd like to make the config more flexible) will get the correct channel even if preserve_channels is set. You can just add `or config.channel` to the return line above.

Adorfer:
i want the optional "site.conf": 
overwrite - take site.conf channel(s) by any means
preserve - keep existing chanels if already gluon on device (even if UCI not set)
node - do what's set in the node's UCI (and by default don't preserve)

NeoRaider:
@Adorfer

When a user changes a setting, site.conf should not override it. There are settings for which we don't know if they were changed by the user; therefore we have to add settings like "preserve_channels".

We definitely don't want to add more complex settings to site.conf. We'd like to keep everything as simple as:
* Defaults get set from site.conf
* When the user changes things, they won't be changed on updates
* When the site.conf defaults change, the settings that haven't been changed by the user should be updated where appropriate.

Following this scheme, I completely agree with @jplitza.

Adorfer:
@NeoRaider 
An Autoupdater is always a matter of trust. 
the scenario for different wifi channels is mostly "airtime optimization" in an existing rollout. 
But if you want to migrate a list of nodes to new domain, it's certainly neccesary to change radio settings, most likely including radio chanels. 
Otherwise perhaps not "override node settings" but "do not preserve if sitecode changes."

NeoRaider:
If a user sets the "preserve_channels" option, that's a willful decision to take care of the configuration themselves. You should never change any settings in spite of such a willful decision.

If some greater change requires modifying the configuration, then the users will need to fix it themselves. The same is the case if the autoupdater is disabled completely.

Adorfer:
> If some greater change requires modifying the configuration, then the users will need to fix it themselves.

Since it's our own nodes, we will then bake special firmware (as usual..)
And yes, i would like to have some kind of lightweight remote administratin interface for nodes which are my own. (and not doing this via ssh-scripts...) but that's another topic.

kokel:
@Adorfer I think your use cases are valid but can't be handled by this little feature. The migration of nodes to a complete different site configuration should be done in an extra package which implements the change of a complete different site.conf because it's not only the channel to be changed, than. This would be much more complex.

@NeoRaider I have merged your suggestions, thanks.

PR #642, opened 2016-02-06
Limit number of connected peers 1

rubo77:
It should be recommended to set the limit number of connected peers to `1` to minimize background traffic in larger networks

### DIFF
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -177,7 +177,7 @@ fastd_mesh_vpn
         groups = {
           backbone = {
             -- Limit number of connected peers from this group
-            limit = 2,
+            limit = 1,
             peers = {
               peer1 = {
                 key = 'XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX',


viisauksena:
if so - you should slightly point to the disadvantages also -  or new risks you get, see
https://forum.freifunk.net/t/fastd-tunnel-anzahl-limit-2-oder-lieber-5-traffic-impact/10280/18?u=fuzzle

Adorfer:
@viisauksena 
>"while taking extra care of having healthy fastd servers (gateway, dhcp, v6, peerlimit and Internet connectivity, or regulary drop old connections)"
this is more an issue in terms of lack of communication between batman and fastd. 
In case batman does not see a gateway over a fastd link, it is not dropped (in order to get a new one). 
So if a fastd-tunnel is stable to a non working batman gateway, this one will be never discarded. For this 2 fastd links do make sense, so batman has the choice.

NeoRaider:
We've already reduced the limit to 1 in the site-example, so doing so in site docs as well makes sense.

@rubo77, your commit message is really misleading, so I'll not merge this, but fix it myself... The first line of the commit message should be understandable without reading the rest of the message or the commit itself.

Issue #328, opened 2015-04-03
PoE passthrough via site.conf and config-mode


RubenKelevra:
Currently the command to enable PoE-Passthru is very cryptic, maybe we should add an option to enable this function via config-mode.

christian1980nrw:
bump ;-)

chris007de:
Regarding the questions whether to use site.conf or config-mode, I would strongly recommend to use the config mode, just because of convenience for the users.

NeoRaider:
We have many site.conf options which aren't supposed to ever be set in "normal" firmwares (like mesh_on_wan, authorized_keys, and a few others). It is still nice to have them to allow completely preconfiguring images for special installations. Of course, such settings should also be added to the config mode.

tcatm:
This OpenWrt commit is relevant to this issue: http://git.openwrt.org/?p=openwrt.git;a=commit;h=24d54f8b0c6b18e5ec1f8c202afdbae14d347737

To my knowledge we'd want to at least support Nanostations, CPE210 and CPE510. Are there other devices supporting PoE passthrough?

SimJoSt:
I would consider adding this option only after #245 has been resolved. At the moment everybody uses PoE-passthrough at their own risk. Making it easier to use this option would create the need for a warning message next to the option.

4ndr3:
Klasse Funktion, habe ich getestet, läuft prima! Drei bitten habe ich noch:

1. Bitte die entsprechenden uci-Befehle zum de-/aktivieren in der Doku hinzufügen: https://github.com/freifunk-gluon/gluon/wiki/Commandline-administration#poe-passthrough
2. Bitte einen warneden Text dran schreiben, dass an den LAN-Port angeschlossende Geräte 24V passives PoE abkönnen müssen oder sonst kaputt gehen
3. Das ganze steht im config mode derzeit unter der Überschrift WAN Connection. Über die Funktion müsste, um Verwirrung auszuschliessen "LAN" stehen. Schliesslich wird es das PoE durch den LAN Port durchgereicht (in der original firmware LAN1 oder secondary LAN port genannt)

Besten Dank!

PR #651, opened 2016-02-13
Backport patch for GPIO control

corny:
First prototype, still untested.
- Where is the best place to store the PoE value in the configuration?
- Is it possible to exclude this package in images for non compatible devices?

### DIFF
--- a/patches/openwrt/0039-base-files-implemented-basic-GPIO-control.patch
+++ b/patches/openwrt/0039-base-files-implemented-basic-GPIO-control.patch
@@ -1,0 +1,144 @@
+From 24d54f8b0c6b18e5ec1f8c202afdbae14d347737 Mon Sep 17 00:00:00 2001
+From: blogic <blogic@3c298f89-4303-0410-b956-a3cf2f4a3e73>
+Date: Wed, 8 Jul 2015 14:25:52 +0000
+Subject: [PATCH] base-files: implemented basic GPIO control
+
+Internal GPIO pins are used for PoE passthrough setups in multi-port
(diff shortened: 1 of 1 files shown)


NeoRaider:
* The OpenWrt trunk has support for setting GPIO pins, /etc/init.d/gpio_switch is provided by base-files. It reads its settings from /etc/config/system. We should backport the OpenWrt solution instead of building our own.
* The gpio_switch script would always exist when we backport this; but we could choose how to expose this option in the expert mode depending on the hardware. The package would always exist, but just not do anything on incompatible devices.

corny:
Thanks for your hints. I added the patch.
If I find time to create a package for the expert mode, I would prefer to create a second pull request for it.

NeoRaider:
@corny, you should also add r46922.

I also prefer to change the commit metadata when cherry-picking from the OpenWrt trunk (set myself as author using `git commit --amend --reset-author` and remove the git-svn line) to make clear that the original commiter is not responsible for our version of the patch.

NeoRaider:
Ping?

Adorfer:
having a UCI-key for "PPOE-Through" on UBNT-NSM and CPE210 would help a great deal. 
(not doing this via rc.local... this is just a hack)

NeoRaider:
I've committed a rebased version of this PR with my requested changes.

corny:
Sorry, I couldn't find time to get this done.
@Adorfer CPE210 has the same as CPE510.

NeoRaider:
CPE210 and CPE510 aren't differentiated by this code anyways, so the Gluon master now has such a setting. We'll also solve #328 soon.

PR #665, opened 2016-02-24
Status page number formatting improvements

NeoRaider:


NeoRaider:
Note: While the status page works fine in Chrome on Android, and this pull requests fixes compatiblity with Firefox for Android, the page is still broken in the default browser of my Android 5.0 phone, as it is missing both Intl and Set.

As the status page doesn't use Set a lot, I'll prepare another pull request after this one is merged (if I can the the status page to work correctly in the default browser)

PR #667, opened 2016-02-29
Status page compatiblity improvements (and some misc. improvements)

NeoRaider:
Avoid using Intl and Set.

Is now working nicely in Firefox & Chrome on desktop and Android, default WebView of Android 5.0 and Safari on an outdated MacOS X 10.9. I suspect it might also work on iPhones, but I haven't tested it.

### DIFF
--- a/package/gluon-status-page/src/js/lib/gui/statistics.js
+++ b/package/gluon-status-page/src/js/lib/gui/statistics.js
@@ -171,7 +171,7 @@ define(["lib/helper"], function (Helper) {
   }
 
   function prettyPackets(d) {
-    var v = new Intl.NumberFormat("de-DE", {maximumFractionDigits: 0}).format(d)
+    var v = Helper.formatNumberFixed(d, 0)
     return v + " Pakete/s"
   }
 
@@ -183,7 +183,7 @@ define(["lib/helper"], function (Helper) {
       prefix++
     }
 
-    d = new Intl.NumberFormat("de-DE", {maximumSignificantDigits: 3}).format(d)
+    d = Helper.formatNumber(d, 3)
     return d + " " + prefixes[prefix]
   }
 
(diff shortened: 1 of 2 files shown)


tcatm:
I don't think there's any reason not to use Set. Instead, we could add a shim to add the functionality in browsers that don't support it.

tcatm:
For the Intl API I'd prefer a patch that uses a manual function only in case the API is not available. This will make removing the fallback code much easier in the future as the API is likely to be supported better within the next years. We could introduce a common helper function to format a number, though.

package/gluon-status-page/src/js/lib/gui/statistics.js
```diff
@@ -113,10 +113,11 @@ define(["lib/helper"], function (Helper) {
         }
 
         var peers = addGroup(d, "")
-        var paths = new Set(peers.map(function (d) { return d.path } ))
+        var paths = {}
+        peers.forAach(function (d) { paths[d.path] = true })
```
tcatm: `forAach` should be `forEach`

tcatm:
Another solution might be to use babel within the workflow: https://github.com/babel/grunt-babel#usage

NeoRaider:
I don't think we should add Shims if we can easily avoid it, and we almost didn't use Set anyways. The status page should really be optimized for code size and low number of dependencies.

I'm not sure if I see the point of using the Intl API when it is available. The number formatting is encapsulated in a function anyways, so its implementation can be replaced at any time. Using two different functions depending on the availability of one just needlessly increases the code complexity.

tcatm:
I just want to be future compatible regarding the Set use. That is, I'll probably use Set sometime again when working on the status page so to avoid this issue we need to add a shim. There are no browsers not supporting Set anyway (that are still under development).

NeoRaider:
@tcatm, could we merge it like this into 2016.1.x, and look for a more complete solution for the master?

tcatm:
I'd like to leave the Set patch out for now. I can't find a recent browser that could benefit from it.

NeoRaider:
@tcatm, I think pretty much all Android devices benefit from it; most people just use the default browser. This might be fixed in Android 6 (haven't tested), but most devices are still on Android 5 or lower.

tcatm:
I just tried using babel on meshviewer. It seems very simple so we could just use it here, too. I'll have a look at it this evening.

NeoRaider:
AFAICT, Babel is only about syntax transformations, we'd still need a Shim to add support for classes like Set.

In my opinion, we should just avoid using Set, as a Shim for it would also require shimming things like iterators etc. This would significantly increase the code size, as the uglifier probably wouldn't be able to remove all of it. I can also make a few experiments tonight to see how much bigger the code would become with common shims.

NeoRaider:
Also, Babel has complex depenencies, making it unusable in the OpenWrt build process...

PR #671, opened 2016-03-02
provide contact-information for reboot-page

kb-light:
The user may enter contact-information within the config-mode. This information can for example be provided for a node-registration-link on the reboot-page.

In our community we had some confusion about asking for contact-information two times (in the config-mode and at node-registration).


tcatm:
If used in a template, will this string get URI encoded?

NeoRaider:
I think both HTML escaping and URL-encoding might be an issue for the config mode texts. We probably need to provide such functions.

tcatm:
Could we define a mechanism for submitting the complete nodeinfo object in such cases? Maybe Javascript on the reboot page could do a POST request to server (with CORS headers set correctly for this to work) and then take over the registration process.

PR #672, opened 2016-03-02
gluon-autoupdater: add function replace_patterns(url)

kb-light:
As already discussed on #gluon:
In some cases it could be helpful, to replace patterns in the autoupdater-path, like it is possible for the opkg-repos.

In my case replacement of %GS is sufficient, but there are maybe some other cases, so I took the entire list.

The function is copied from package/gluon-core/files/lib/gluon/upgrade/500-opkg.

### DIFF
--- a/package/gluon-autoupdater/files/lib/gluon/upgrade/500-autoupdater
+++ b/package/gluon-autoupdater/files/lib/gluon/upgrade/500-autoupdater
@@ -1,17 +1,38 @@
 #!/usr/bin/lua
 
+local fs = require 'nixio.fs'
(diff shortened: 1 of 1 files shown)


NeoRaider:
I don't think this is useful.

In contrast to opkg repos, which require such patterns (as they contain target-specific parts, and often need a new path for each single version, as the kernel modules may have become incompatible), the autoupdater is designed to work "forever" with the same path and one single manifest for all targets.

You mentioned the site code as your usecase, but this usecase is not very strong, as you could well generate the autoupdate URLs in the same way you generate the site code.

Issue #553, opened 2015-11-10
.deplease review timing of MLDv2 for IPv6 / RFC3810


Joachim-Wolff:
Broadcast traffic for IPv6 Multicast Listener Discovery Version 2 floods the howl mesh network every Query Interval of currently 60 seconds (icmpv6.type == 143) . Every node has to answer to the request packet (icmpv6.type == 130) during the Query Response Interval.

The recommended timing from RFC3810 is:
Query Interval, default value: 125 seconds, gluon use 60 seconds
Query Response Interval, default value: 10000 (10 seconds) seams to be below 4 seconds in gluon, based on captured traffic on a node somewhere in the mesh. Sorry, haven't found the values in the source, but also don't know exactly where I have to search for...

![x-screenshot from 2015-11-09 18-42-32](https://cloud.githubusercontent.com/assets/7549440/11062356/8eceb9ee-87ae-11e5-84d9-b52572665ded.png)

36% of the packets arrive in the 1st second, 35% in the 2nd, 23% in the 3rd and the remaining 6% within the 4rd second. Makes a peek of 350 pkt/s with ~970 nodes in the mesh, expecting each one has to be handled by the IPv6 stack... Maybe this explain the high sirq values or the load issue #446

Joachim-Wolff:
My nodes have answered all request faster than 30ms, only one exception with 65ms, so without delay...

Expecting every node do so and there is no large delay in transmitting the request on uncongested lines, there must be a fine broadcast peak. The visible time in the graph seams to be the transmitting time through the mesh, and every packet must be stored somewhere for this time, because it's broadcast and all packets are delivered successfully. Fortunately the packet size is only 158 byte :)

tcatm:
You may have a misconception about how MLDv2 works. It's up to the querier to decide on both intervals and (If I'm not mistaken here) any host on a layer2 mesh may become a querier (even a Client). Could you check which host has taken that role in your mesh and whether it is running Gluon at all?

Joachim-Wolff:
Sorry, have read only 30% of that RFC, and they always use the wording router...

But you're right, I've captured on 5 days and it always was a client (not the same, maybe smartphone) behind a Freifunk-Router that send the queries, with a Maximum Response Time of 10ms, so all works like designed... :(

So maybe change the title to "take control over MLDv2" ??
How handle other communities this issue, and are they aware of it?

T-X:
[...]

PR #674, opened 2016-03-06
Filter IGMP+MLD

T-X:
Addresses #553, prepares for #278 (#402).

### DIFF
--- a/package/gluon-ebtables-filter-multicast/files/lib/gluon/ebtables/110-mcast-allow-icmpv6
+++ b/package/gluon-ebtables-filter-multicast/files/lib/gluon/ebtables/110-mcast-allow-icmpv6
@@ -1,5 +1,3 @@
 rule 'MULTICAST_OUT -p IPv6 --ip6-protocol ipv6-icmp --ip6-icmp-type echo-request -j DROP'
 rule 'MULTICAST_OUT -p IPv6 --ip6-protocol ipv6-icmp --ip6-icmp-type 139 -j DROP'
 rule 'MULTICAST_OUT -p IPv6 --ip6-protocol ipv6-icmp -j RETURN'
-
-rule 'MULTICAST_OUT -p IPv6 --ip6-protocol 0 -j RETURN' -- hop-by-hop
--- a/package/gluon-ebtables-filter-multicast/files/lib/gluon/ebtables/110-mcast-allow-icmpv6
+++ b/package/gluon-ebtables-filter-multicast/files/lib/gluon/ebtables/110-mcast-allow-icmpv6
@@ -1,3 +1,3 @@
-rule 'MULTICAST_OUT -p IPv6 --ip6-protocol ipv6-icmp --ip6-icmp-type echo-request -j DROP'
-rule 'MULTICAST_OUT -p IPv6 --ip6-protocol ipv6-icmp --ip6-icmp-type 139 -j DROP'
-rule 'MULTICAST_OUT -p IPv6 --ip6-protocol ipv6-icmp -j RETURN'
+rule 'MULTICAST_OUT_ICMPV6 -p IPv6 --ip6-protocol ipv6-icmp --ip6-icmp-type echo-request -j RETURN'
+rule 'MULTICAST_OUT_ICMPV6 -p IPv6 --ip6-protocol ipv6-icmp --ip6-icmp-type 139 -j RETURN' -- ICMP Node Information Query
+rule 'MULTICAST_OUT_ICMPV6 -j ACCEPT'
--- a/docs/package/gluon-ebtables-segment-mld.rst
+++ b/docs/package/gluon-ebtables-segment-mld.rst
@@ -1,0 +1,16 @@
+gluon-ebtables-segment-mld
+==========================
+
+These filters drop IGMP/MLD packets before they enter the mesh and
+filter any IGMP/MLD packets coming from the mesh.
+
+IGMP/MLD have the concept of a local, elected Querier. For more
+decentralization and increased robustness, the idea of this package is
+to split the IGMP/MLD domain a querier is responsible for, allowing to
+have a querier per node. The split IGMP/MLD domain will also reduce
+overhead for this packet type, increasing scalability.
+
+Beware of the consequences of using this package though: You might need
+to explicitly, manually mark ports on snooping switches leading towards
+your mesh node as multicast router ports for now (Multicast Router
+Discovery, MRD, not implemented yet).
--- a/package/gluon-ebtables-filter-multicast/files/lib/gluon/ebtables/100-mcast-chain
+++ b/package/gluon-ebtables-filter-multicast/files/lib/gluon/ebtables/100-mcast-chain
@@ -1,1 +1,0 @@
-chain('MULTICAST_OUT', 'DROP')
--- a/package/gluon-ebtables-filter-multicast/files/lib/gluon/ebtables/300-mcast
+++ b/package/gluon-ebtables-filter-multicast/files/lib/gluon/ebtables/300-mcast
@@ -1,2 +1,0 @@
-rule 'FORWARD --logical-out br-client -o bat0 -d Multicast -j MULTICAST_OUT'
-rule 'OUTPUT --logical-out br-client -o bat0 -d Multicast -j MULTICAST_OUT'
--- a/package/gluon-ebtables-filter-multicast/files/lib/gluon/ebtables/355-mcast-drop
+++ b/package/gluon-ebtables-filter-multicast/files/lib/gluon/ebtables/355-mcast-drop
@@ -1,0 +1,1 @@
+rule ('MULTICAST_OUT -j DROP')
--- a/package/gluon-ebtables-segment-mld/Makefile
+++ b/package/gluon-ebtables-segment-mld/Makefile
@@ -1,0 +1,51 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-ebtables-segment-mld
+PKG_VERSION:=1
+PKG_RELEASE:=1
+
+PKG_BUILD_DIR := $(BUILD_DIR)/$(PKG_NAME)
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/gluon-ebtables-segment-mld
+  SECTION:=gluon
+  CATEGORY:=Gluon
+  TITLE:=Ebtables filters for IGMP/MLD packets
+  DEPENDS:=+gluon-core +gluon-ebtables
+endef
+
+define Package/gluon-ebtables-segment-mld/description
+	Gluon community wifi mesh firmware framework: Ebtables filters for
+	IGMP/MLD packets
+
+	These filters drop IGMP/MLD packets before they enter the mesh and
+	filter any IGMP/MLD packets coming from the mesh.
+
+	IGMP/MLD have the concept of a local, elected Querier. For more
+	decentralization and increased robustness, the idea of this package is
+	to split the IGMP/MLD domain a querier is responsible for, allowing to
+	have a querier per node. The split IGMP/MLD domain will also reduce
+	overhead for this packet type, increasing scalability.
+
+	Beware of the consequences of using this package though: You might need
+	to explicitly, manually mark ports on snooping switches leading towards
+	your mesh node as multicast router ports for now (Multicast Router
+	Discovery, MRD, not implemented yet).
+endef
+
+define Build/Prepare
+	mkdir -p $(PKG_BUILD_DIR)
+endef
+
+define Build/Configure
+endef
+
+define Build/Compile
+endef
+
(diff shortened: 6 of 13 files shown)


T-X:
I had tried filter first, but it isn't early enough. I want to avoid a foreign query message somewhere from the mesh to race with the bridge's querier election later, once snooping is re-enabled later. A foreign host/querier should not be able to stop a node's querier. The order of processing IGMP/MLD messages from the mesh is: PREROUTING, bridge multicast snooping and finally INPUT, IP-stack, etc.

(I know, it is a little counter intuitive that the INPUT hook isn't called before the snooping code)

package/gluon-ebtables-filter-multicast/files/lib/gluon/ebtables/110-mcast-allow-icmpv6
```diff
@@ -1,5 +1,6 @@
 rule 'MULTICAST_OUT -p IPv6 --ip6-protocol ipv6-icmp --ip6-icmp-type echo-request -j DROP'
 rule 'MULTICAST_OUT -p IPv6 --ip6-protocol ipv6-icmp --ip6-icmp-type 139 -j DROP'
+rule 'MULTICAST_OUT -p IPv6 --ip6-protocol ipv6-icmp --ip6-icmp-type 130 -j DROP' -- MLD query
+rule 'MULTICAST_OUT -p IPv6 --ip6-protocol ipv6-icmp --ip6-icmp-type 131 -j DROP' -- MLDv1 report
+rule 'MULTICAST_OUT -p IPv6 --ip6-protocol ipv6-icmp --ip6-icmp-type 143 -j DROP' -- MLDv2 report
```
NeoRaider: Could you sort the 139 between the others? And also give it a comment...

NeoRaider:
Another note for both #674 and #675: You might want to `git rebase --ignore-date` these branches to update the commit dates, they are currently all authored on September 5th, which is a bit confusing...

NeoRaider:
I'm wondering if it's okay to keep the MLD filter rules in gluon-ebtables-filter-multicast, or if we should make gluon-ebtables a dependency of gluon-mesh-batman-adv-core and put the rules there (i.e. make MLD segmentation mandatory for batman-adv in Gluon)

T-X:
What would be the advantage of making it a dependency? Maybe it is a bit of a philosophical question, too, whether Gluon is supposed to be usable for applications other than "metropolitan-size" mesh networks. For a smaller and fixed-size mesh network, for instance for a camping place, a trailer park or something like that, people might want to use Gluon without the filter package to be able to use mDNS, Samba, etc.

Also, people will probably notice pretty quickly when they might need to install the filter package once hitting the magical ~70 nodes barrier :). Right?

NeoRaider:
@T-X, that is exactly my point. The MLD segmentation should work without gluon-ebtables-filter-multicast, as not everybody might want to filter all multicast, but the MLD segmentation is generally useful.

T-X:
Ah, okay, then I misunderstood indeed :). Hm, hm... my current feeling is that just like the general multicast filtering as is is somehow specific to our Freifunk setups, so is this MLD segmentation. There is one more thing I would like to change before calling it "generic" / generally useful:

One disadvantage of the current approach is, that if you put another snooping switch, for instance your neat, expensive Cisco switch with multicast snooping and a querier on by default, and wins the querier election against the Gluon bridge of the Gluon node, hosts behind that switch might have issues to reach hosts behind the mesh. That is because hosts behind the enterprise switch will not know that it should forward any multicast to the Gluon node.

In a *wireless* community mesh network that usually is not an issue because STAs will always forward to the Gluon node and you cannot use bridges on STAs. For the LAN part in a wireless community mesh network, that is usually secondary and it is usually discouraged to place devices taking over such central, managing tasks (like DHCP for instance, too) anyway. And if they do, they at least only harm their own, local, wired LAN segment but nobody else.

I would call it "generally useful" once the following two tasks are achieved as well: Bridges autonomously sending Multicast-Router-Advertisements to ports adjacent to any bridge port manually configured as "multicast router port". And secondly, switches actually recognizing and processing MRAs (the Linux bridge, and probably many other switches too, currently does not, hasn't been implemented yet).

Does that make sense and clarify the current short-comings (which I think are negligible for wireless community mesh networks)?

NeoRaider:
Still, putting the MLD segmentation into gluon-ebtables-filter-multicast doesn't seem right to me. gluon-ebtables-filter-multicast is about blocking unwanted multicast; MLD segmentation should be possible without this, and vice-versa.

I think we should add a new package gluon-ebtables-segment-multicast or similar for this purpose.

T-X:
I hope nobody missed it, but I had updated the commits as suggested by @NeoRaider.

Did anyone (not necessarily @NeoRaider or @tcatm) have a chance to test and verify these patches yet?

package/gluon-ebtables-split-igmpmld-domain/Makefile
```diff
@@ -0,0 +1,42 @@
...
+endef
+
+define Package/gluon-ebtables-split-igmpmld/description
+	Gluon community wifi mesh firmware framework: Ebtables filters for IGMP/MLD packets
+
+	These filters drop IGMP/MLD packets before they enter the mesh and filter any IGMP/MLD packets coming from the mesh.
+
+	IGMP/MLD have the concept of a local, elected Querier. For more decentralization and increased robustness, the idea of this package is to split the IGMP/MLD domain a querier is responsible for, allowing to have a querier per node. The split IGMP/MLD domain will also reduce overhead for this packet type, increasing scalability.
+
+	Beware of the consequences of using this package though: You might need to explicitly, manually mark ports on snooping switches leading towards your mesh node as multicast router ports for now (Multicast Router Discovery, MRD, not implemented yet).
```
NeoRaider: These lines are extremely long, please wrap them to a reasonable length (such as 80 chars).

package/gluon-ebtables-split-igmpmld-domain/Makefile
```diff
@@ -0,0 +1,42 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-ebtables-split-igmpmld
```
NeoRaider: The directory is called *gluon-ebtables-split-igmpmld-domain*, but the package name is *gluon-ebtables-split-igmpmld*.

package/gluon-ebtables/files/lib/gluon/ebtables/101-dir-rules
```diff
@@ -1,2 +1,7 @@
 rule 'IN_ONLY --logical-in br-client -i ! bat0 -j DROP'
 rule 'OUT_ONLY --logical-out br-client -o ! bat0 -j DROP'
+
+rule 'OUTPUT -d Multicast --logical-out br-client -o bat0 -j MULTICAST_OUT'
+rule 'FORWARD -d Multicast --logical-out br-client -o bat0 -j MULTICAST_OUT'
+
+rule 'MULTICAST_OUT -p IPv6 --ip6-protocol ipv6-icmp -j ICMPV6_OUT'
```
NeoRaider: It is not really clear from the name that *ICMPV6_OUT* only applies to multicast packets. Maybe *MULTICAST_OUT_ICMPV6*?

package/gluon-ebtables-split-igmpmld-domain/files/lib/gluon/ebtables/105-mcast-drop-mld
```diff
@@ -0,0 +1,9 @@
+rule('ICMPV6_OUT -p IPv6 --ip6-protocol ipv6-icmp --ip6-icmp-type 130 -j DROP') -- MLD Query
+rule('ICMPV6_OUT -p IPv6 --ip6-protocol ipv6-icmp --ip6-icmp-type 131 -j DROP') -- MLDv1 Report
+rule('ICMPV6_OUT -p IPv6 --ip6-protocol ipv6-icmp --ip6-icmp-type 132 -j DROP') -- MLDv1 Done
+rule('ICMPV6_OUT -p IPv6 --ip6-protocol ipv6-icmp --ip6-icmp-type 143 -j DROP') -- MLDv2 Report
+
+rule('ICMPV6_IN -p IPv6 --ip6-protocol ipv6-icmp --ip6-icmp-type 130 -j DROP', 'nat') -- MLD Query
+rule('ICMPV6_IN -p IPv6 --ip6-protocol ipv6-icmp --ip6-icmp-type 131 -j DROP', 'nat') -- MLDv1 Report
+rule('ICMPV6_IN -p IPv6 --ip6-protocol ipv6-icmp --ip6-icmp-type 132 -j DROP', 'nat') -- MLDv1 Done
+rule('ICMPV6_IN -p IPv6 --ip6-protocol ipv6-icmp --ip6-icmp-type 143 -j DROP', 'nat') -- MLDv2 Report
```
NeoRaider: It is not clear to me why the *ICMPV6_IN* and *ICMPV6_OUT* chains are needed at all. AFAICT, all these rules could just be added to the MULTICAST_IN/OUT chains.

T-X: Of course, that could be done too. I think I had even started like that but found the "double negation" a bit confusing to read then if everything is in MULTICAST_IN/OUT directly.

Like, everything is allowed by default, then multicast is dissallowed by default if that package was added, but not everything is dissallowed, ICMPV6 is exempt. But hey, then again, there are some rules a little earlier in the same MULTICAST_IN/OUT chains, again dropping some other ICMPV6 types. And I was then already starting to loose myself with all that :).

So the idea was to instead make the rules more readable by saying right from the start that some ICMPV6 types might be allowed and people can have a closer look within the according ICMPV6 chains then, separating all non-ICMPV6 rules from any concerns with ICMPV6.

T-X:
Have written my thinking about the advantage of separating ICMPV6 rules (now MULTICAST_ICMPV6_IN/OUT) within their own chain.

NeoRaider:
Some high-level docs for the package in *docs/package* would be great.

rotanid:
@NeoRaider  isn't their already a doc file  gluon/docs/package/gluon-ebtables-segment-mld.rst in this PR?

T-X:
@rotanid: There is now, it wasn't when @NeoRaider asked for it ;). See the timeline. (sometimes github can be confusing indeed...)

rotanid:
@T-X it seems like i mixed up two dates, after further digging i found out now you added the doc only a few hours after he asked for them. but the date that could be found more easy stated something like 8th of march...

T-X:
@ohrensessel had found a bug resulting in ICMPv6 Router Advertisements being sent into the mesh again. This was because of the ACCEPT triggered in MULTICAST_OUT_ICMPV6 before the RA drop rule. Fixed this just now by moving the "OUTPUT/FORWARD -d Multicast ..." rules below the radv-rules.

NeoRaider:
I think this is fine now.

T-X:
@ohrensessel forwarded to me that Andre seemed to have had issues with this pull requests plus #688 applied on a node. In that, an ssh connecting supposedly and reproduceably broke after a short time but worked on an unpatched router.

@ohrensessel and I weren't able to reproduce the issue on our own nodes yet though. More information from Andre was requested. Maybe we should wait for @ohrensessel or Andre to either give an "Acked-by" or some more information.

rotanid:
are there any news on this @T-X @ohrensessel ? if no one can reproduce the alleged problem, i'd love to see this in 2016.2

T-X:
From IRC:

<pre>
--- Day changed Sun Jun 26 2016
08:17 < T_X> ohrensessel, andr3: sagt ihr nochmal kurz wegen #674 (filtern von IGMP+MLD) bescheid, was eure meinung dazu wäre
08:18 < T_X> also ob ihr plant, das in den nächsten tagen nochmal weiter zu testen oder ob man das eurer meinung nach mergen könnte
20:26 < ohrensessel> T_X ich meine das is ok so
20:26 < ohrensessel> kann es frühstens in 2 wochen nochmal ausgiebig testen
22:25 < T_X> ohrensessel: ok. schreibst du's noch unters ticket? (auch als gewissensprobe, am besten mit blut signieren bitte :P)
22:31 < ohrensessel> frühestens^^
22:31 < ohrensessel> ich signier da gar nix ;)
22:31 < ohrensessel> also ich nutz die patches hier bei mir schon länger
22:31 < ohrensessel> und mir is noch kein fehlverhalten aufgefallen
</pre>

So sounds like an OK'ish from @ohrensessel, maybe ;-).

rotanid:
so only an ACK from @tcatm is what's left, great :-)

NeoRaider:
@T-X, do you think this is ok to merge now?

T-X:
Yes please :-).

It'll be a good idea to first have bridge multicast snooping in Gluon for a while and not to wait too long to avoid adding it together with the batman-adv multicast optimizations for bridges (upcoming batman-adv 2016.3) at about the same time, I think.

PR #675, opened 2016-03-06
Preparations for Multicast Snooping

T-X:
This pull request adds two patches mandatory for bridge multicast snooping. For now, as multicast snooping is still disabled, they are basically a no-op.

Ref.:  #278 (#402), #674

### DIFF
--- a/package/gluon-core/files/lib/gluon/upgrade/110-network
+++ b/package/gluon-core/files/lib/gluon/upgrade/110-network
@@ -56,5 +56,31 @@ uci:save('network')
 uci:commit('network')
 
 
+uci:section('firewall', 'rule', 'wan_igmp',
+	    {
+	       name = 'Allow-IGMP',
+	       src = 'wan',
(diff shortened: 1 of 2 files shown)


package/gluon-core/files/lib/gluon/upgrade/110-network
```diff
@@ -56,5 +56,31 @@ uci:save('network')
...
+	    }
+)
+
+uci:section('firewall', 'rule', 'wan_mld',
+	    {
+	       name = 'Allow-MLD',
+	       src = 'wan',
+	       proto = 'icmp',
+	       src_ip = 'fe80::/10',
+	       icmp_type = { '130/0', '131/0', '143/0', },
```
NeoRaider: r45613 also lists `132/0` here, you don't. Intentional?

T-X:
Hm, no, wasn't intentional. Going to add it (and to #674 too).

PR #679, opened 2016-03-08
Preparations for Multicast Snooping Final

T-X:
These are supposed to be the last two patches necessary before re-activating multicast snooping.

The first one lowers the query intervals for br-client, the second one increases the robustness against potential packetloss on the wifi.

For now, as multicast snooping is still disabled, they are basically a no-op.

Ref.: #674, #675, #278 ( #402)


NeoRaider:
I think we should rather update netifd to the current git HEAD as a whole, most commits inbetween are bugfixes anyways.

NeoRaider:
I've just updated netifd to the latest version.

PR #688, opened 2016-03-15
Enable multicast snooping

T-X:
This commit enables multicast snooping on br-wan and br-client again.

Should be pulled only _after_ #674  and #679 got merged. This PR is rebased onto #679 for easier merging, the according two commits are identical to the ones in #679 .

Ref.: #278 (#402)

### DIFF
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/110-network
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/110-network
@@ -9,11 +9,12 @@ uci:section('network', 'interface', 'wan',
 	    {
 	      ifname = sysconfig.wan_ifname,
 	      type = 'bridge',
-	      igmp_snooping = 0,
+	      multicast_querier = 0,
 	      peerdns = 0,
 	      auto = 1,
 	    }
 )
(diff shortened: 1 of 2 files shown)


T-X:
Now rebased to master.

NeoRaider:
I still think the settings for the two bridges don't have a lot to do with each other, so they should be done in separate commits (then the commits can also get proper package name prefixes).

NeoRaider:
Is this supposed to happen? ```br-client: Multicast hash table maximum of 512 reached, disabling snooping: bat0``` Can we somehow disable the snooping on bat0 in the first place?

T-X:
@NeoRaider: Have you added the gluon-ebtables-segment-mld package in the site configuration of the firmware you are testing?

No, currently there is no possibility within the bridge to disable snooping on a per-port basis. There is just one per bridge multicast_snooping switch. Similarly, this slightly misleading message actually disables snooping for the whole bridge.

However, the gluon-ebtables-segment-mld package should be just as good to avoid learning about multicast listeners behind bat0 and should therefore prevent the message and action you are quoting.

PR #701, opened 2016-03-22
Do not have the ugly "-e" printed for CheckTarget

smoe:
The "echo -e" is executed in a newly spawned shell (the enclosing ()s). That shell by default is a bourne shell. And that has a built-in echo that fails to understand the -e option and just prints it. Same for dash. If you set a SHELL=/bin/bash then this would also solve the issue. I had felt that by explicitly executing the binary echo implementation this would be the easiest to grasp. Well. Make your pick.


NeoRaider:
Thanks, I've decided to copy the SHELL definition from OpenWrt's rules.mk instead, to keep the shell definition consistent.

PR #702, opened 2016-03-22
bit of eye candy in script/update.sh

smoe:
The separation by an empy line helps to dissect potential error messages.

### DIFF
--- a/scripts/update.sh
+++ b/scripts/update.sh
@@ -5,6 +5,7 @@ set -e
 . "$GLUONDIR"/scripts/modules.sh
 
 for module in $GLUON_MODULES; do
+	echo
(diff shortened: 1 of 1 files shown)


NeoRaider:
Let's rather add a heading line, that doesn't cause a weird empty line at the top and is even more useful. I'm thinking of something like:

    echo "--- Updating module '$module' ---"

A similar addition would be nice in patch.sh.

smoe:
I completely agree. I was shying off a bit.

I presume/hope you prefer to just add it yourself over me providing
something to pull from.

Cheers,

Steffen

NeoRaider:
Done.

PR #706, opened 2016-03-27
gluon-status-page: sort gateways alphabetically

### LINKED ISSUE #698: sort gateways alphabetically
The order of gateways in a community in the status page is not sorted:

It would be nice to have them sorted alphabetically ;)

corvusmo:
fixes #698

### DIFF
--- a/package/gluon-status-page/src/js/lib/gui/statistics.js
+++ b/package/gluon-status-page/src/js/lib/gui/statistics.js
@@ -22,18 +22,27 @@ define(["lib/helper"], function (Helper) {
     return el
   }
 
-  function mkRow(table, label, stream) {
-    var tr = document.createElement("tr")
+  function mkRow(table, label, stream, sorted) {
+
+    var i = -1
+
+    if (sorted) {
+      for (i = 0; i < table.rows.length; i++) {
+        if (label < table.rows[i].firstChild.textContent)
+           break
+      }
+    }
+
+    var tr = table.insertRow(i)
     var th = document.createElement("th")
     var td = streamElement("td", stream)
     th.textContent = label
     tr.appendChild(th)
     tr.appendChild(td)
-    table.appendChild(tr)
 
     tr.destroy = function () {
       td.destroy()
-      table.removeChild(tr)
+      table.tBodies[0].removeChild(tr)
     }
 
     return tr
@@ -127,7 +136,7 @@ define(["lib/helper"], function (Helper) {
                                         stream.startWith(d)
                                         .map(peer.path)
                                         .filter(function (d) { return d !== undefined })
-                                        .map(prettyPeer))
+                                        .map(prettyPeer), true)
         })
       }
     })


corvusmo:
I've updated the pr. I don't know if it is a good solution but it seems to work and as you wrote it should happen only on rare occasions.

NeoRaider:
A bit better, but I think it still doesn't catch changes that don't alter the number of entries (like a peer disappearing and a different one appering). I think it would make sense to compare the key set of *children* with the set of the *path* attributes of all peers, and if these differ in any way, delete all rows.

corvusmo:
This doesn't happen, because the disappering peer would be removed from the table right before (L118-122) the comparison, which should catch all changes when peers are added, whether others are removed simultaneously or not.

NeoRaider:
Ah, I completely overlooked that the code was already doing what I was proposing. I think this looks fine then... @tcatm?

tcatm:
I don't like the code that clears the list. It will break copy pasting from the status page. That is why all the code is really careful about ever removing elements from the DOM.

tcatm:
You could do it like this:

- refactor `mkRow` to allow usage of `insertBefore` in addition to `appendChild`
- when adding a row, iterate over all existing rows until one with a lexicographically higher ranked name in (maybe add a data attribute for quick lookup)
- if found, insert before that element, else append to the end of the table

corvusmo:
@tcatm Thank you for your suggestions and for the chance to contribute even as an unexperienced programmer.

package/gluon-status-page/src/js/lib/gui/statistics.js
```diff
@@ -22,14 +22,24 @@ define(["lib/helper"], function (Helper) {
...
+  function mkRow(table, label, stream, sorted) {
     var tr = document.createElement("tr")
     var th = document.createElement("th")
     var td = streamElement("td", stream)
     th.textContent = label
     tr.appendChild(th)
     tr.appendChild(td)
-    table.appendChild(tr)
+
+    if (sorted && table.rows.length && label < table.lastChild.firstChild.innerHTML) {
```
tcatm: Use `textContent` instead of `innerHTML` (applies to line 35 as well).

tcatm:
I just learned about `insertRow()` (https://developer.mozilla.org/en-US/docs/Web/API/HTMLTableElement/insertRow). This could simplify the code even more as both `insertBefore` and `appendChild` can be refactored to use `insertRow`.

corvusmo:
I tried to refactor the code. Unfortunately i was not able to get it completely working for the `mkRow` function. I tried two `insertRow` calls and also a single call with an "index" variable. In both cases the sorting of the gateway table was lost...

tcatm:
Could we focus on mkRow in this PR and not change surrounding code?

tcatm:
Try something like this:
```
  var i = -1
  
  if (sorted) {
  	for (i = 0; i < table.rows.length; i++) {
  		if (label < table.rows[i].firstChild.textContent)
 		   	break
 	 }
  }
  
  var tr = table.insertRow(i)
```

corvusmo:
Thank you again. Last try for today. `insertRow()` produces an additional `<tbody>` tag, so there was another adjustment necessary.

PR #707, opened 2016-03-28
Tcatm's meraki branch merged

smoe:
This patch brings the work of @tcatm with @Garunda on support for the CISCO Meraki series for Gluon in sync with OpenWRT as distributed with 2016.1.x. That latest redistributed OpenWRT release has accepted the patch that the two had previously applied for their cause and thus this merge basically removes the patch to OpenWRT while leaving the build-invocation by Gluon intact.

OpenWRT not only supports the MR12 (similar to the MR62) but also the MR16 (similar to the MR66). The firmware was successfully tested by @Garunda for the MR62. Gluon images for the MR16 were added to the patch of @tcatm. The confirmation of the functionality of the image for the MR66 is still pending.

The devices are of interest to the community as they are perceived as "enterprise class" hardware (makes a rock-solid impression, indeed) which comes with an annually renewed fairly hefty license. The Freifunk may offer an escape route for those who had signed up and want to keep their investment into the similarly expensive hardware. Used evices sell for $60 on ebay/amazon in the US. Here in the old world it is all >300 €, still.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -94,6 +94,11 @@ ar71xx-generic
   - WZR-HP-G300NH
   - WZR-HP-G450H
 
+* Cisco Meraki
+
+  - MR12 / MR62
+  - MR16 / MR66
+
 * D-Link
 
   - DIR-505 (A1)
--- a/targets/ar71xx-generic/profiles.mk
+++ b/targets/ar71xx-generic/profiles.mk
@@ -288,3 +288,16 @@ $(eval $(call GluonModel,HORNETUB,hornet-ub,alfa-hornet-ub))
 $(eval $(call GluonModelAlias,HORNETUB,alfa-hornet-ub,alfa-ap121))
 $(eval $(call GluonModelAlias,HORNETUB,alfa-hornet-ub,alfa-ap121u))
 
+## Meraki
+
+# Meraki MR12/MR62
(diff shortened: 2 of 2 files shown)


NeoRaider:
1. You mixed up @tcatm's username in the commit message... and probably confused someone, as the other username exists as well.
2. This pull request consists of two commits, where the second reverts most of the first one. Squash and rebase them into a single commit with a commit message that explains only the remaining changes.
3. I think the file names are wrong. Unless I've overlooked something, the autoupdater will look for the names *meraki-m12* and *meraki-m16*. You can add Images for the M62 and M66 using the *GluonModelAlias* macro (as it is done for a few other models in *ar71xx-generic/profiles.mk*.

smoe:
Hi Matthias,

I have force-pushed to the same branch and (think to have) done as advised.
Please kindly give it a second chance/look.

Best,
Steffen

NeoRaider:
Looks good now, can you also add the devices to the list of supported hardware in *docs/index.rst*?

smoe:
Updated, again force-pushed as single commit. The branch by Tcatm referenced is https://github.com/freifunk-gluon/gluon/tree/mr12-62 .

NeoRaider:
There's a typo in all model names in index.rst...

smoe:
Weird. Thanks for spotting this. Corrected.

Issue #648, opened 2016-02-10
Support for Mediatek/Ralink WLAN


NeoRaider:
Ralink WLAN hardware does support VIFs, but
* no VIFs with IBSS (but AP + 11s works)
* only with sequential MAC addresses (the MAC addresses of all VIFs may only differ in the last 3 bits -> 8 VIFs are supported)

The MAC address scheme deployed by Gluon doesn't not generate sequential MAC addresses (the address of a single radio are counted in byte 2, different radios in byte 3). This was done to avoid address collisions (we have to deal with devices with only a single assigned MAC address, so devices bought together often have very similar MAC addresses).

We plan to fix this by generating a new base MAC address by applying a hash function to the real primary address, and derive the VIF addresses from this base address.

**Update:**

* The effective length of such a MAC prefix is 43bits, as the last 3 bits are variable, and 2 bits are fixed (unicast / locally administered). This yields a collision probablity of 1.4e-08 for 1000 prefixes and 1.4e-06 for 10000 prefixes.
* Using a single prefix for all interfaces of a node would work, as 8 addresses are enough even for nodes with two radios.
  * 8 addresses would suffice for 4 radios for AP+11s, and 2 radios for AP+IBSS+11s (which is not recommended anyways). No device we support has more than 2 radios, and using additional radios doesn't work without manual config (so requiring manual config of the MAC addresses wouldn't be a problem)
  * Non-wireless interfaces (mesh VPN, mesh-on-WAN/LAN) need addresses as well. Conveniently, it is possible to reuse AP VIFs' addresses for these interfaces, as APs don't participate in meshing directly, so their MAC addresses aren't visible to batman-adv.
  * Example configuration for the worst case:
    * 0: client0; mesh-vpn
    * 1: mesh0
    * 2: ibss0
    * 3: client1; mesh-on-wan
    * 4: mesh1
    * 5: ibss1
    * 6: mesh-on-lan
    * 7: unused

FreifunkUFO:
your expectations are wrong: at least some mediatek/ralink hardware support ad-hoc + vAP mode.

p.s. for the hardware which doesnt support that: gluon already has queckboxes for disabling mesh or vAP on the device!?

NeoRaider:
I've updated the issue description with a more detailed plan.

[...]

PR #715, opened 2016-04-01
Fix ralink-rt305x/vocore mac schema generation.

lemoer:
The VIF mac addresses in the ralink-rt305x chip must be sequential in
the last bits, since they are using a hardware mac filter.

1. Hash the original primary mac, since the addresses of devices bought
   together could be sequential in the last byte.

2. Increment the last byte for the VIF on the vocore (```generate_mac``` routine)

Implements #648

### DIFF
--- a/package/gluon-client-bridge/files/lib/gluon/upgrade/320-gluon-client-bridge-wireless
+++ b/package/gluon-client-bridge/files/lib/gluon/upgrade/320-gluon-client-bridge-wireless
@@ -21,13 +21,9 @@ local function configure_client(config, radio, index, suffix)
 
   uci:delete('wireless', name)
 
-  if index == 0 then
-    macaddr = util.generate_mac(0)
-  elseif index == 1 then
-    macaddr = util.generate_mac(3)
-  end
+  macaddr = util.generate_mac(3*index)
 
-  if config then
+  if config and macaddr then
     uci:section('wireless', 'wifi-iface', name,
       {
         device = radio,
--- a/package/gluon-client-bridge/files/lib/gluon/upgrade/320-gluon-client-bridge-wireless
+++ b/package/gluon-client-bridge/files/lib/gluon/upgrade/320-gluon-client-bridge-wireless
@@ -21,6 +21,12 @@ local function configure_client(config, radio, index, suffix)
 
   uci:delete('wireless', name)
 
+  if index == 0 then
+    macaddr = util.generate_mac(0)
+  elseif index == 1 then
+    macaddr = util.generate_mac(3)
+  end
+
   if config then
     uci:section('wireless', 'wifi-iface', name,
       {
@@ -28,7 +34,7 @@ local function configure_client(config, radio, index, suffix)
         network = 'client',
         mode = 'ap',
         ssid = config.ssid,
-        macaddr = util.generate_mac(2, index),
+        macaddr = macaddr,
         ifname = suffix and 'client' .. suffix,
         disabled = disabled,
       }
--- a/package/gluon-core/files/usr/lib/lua/gluon/util.lua
+++ b/package/gluon-core/files/usr/lib/lua/gluon/util.lua
@@ -84,6 +84,8 @@ end
 -- 6: mesh-on-lan
 -- 7: unused
 function generate_mac(i)
+  if i > 7 or i < 0 then return nil end -- max allowed id (0b111)
+
   local hashed = string.sub(hash.md5(sysconfig.primary_mac), 0, 12)
   local m1, m2, m3, m4, m5, m6 = string.match(hashed, '(%x%x)(%x%x)(%x%x)(%x%x)(%x%x)(%x%x)')
 
@@ -96,8 +98,6 @@ function generate_mac(i)
   -- It's necessary that the first 45 bits of the mac do
   -- not vary on a single hardware interface, since some chips are using
   -- a hardware mac filter. (e.g 'ramips-rt305x')
-
-  if i > 7 then return nil end  -- max allowed id (0b111)
 
   m6 = nixio.bit.band(m6, 0xF8) -- zero the last three bits (space needed for counting)
   m6 = m6 + i                   -- add virtual interface id
--- a/package/gluon-core/files/usr/lib/lua/gluon/util.lua
+++ b/package/gluon-core/files/usr/lib/lua/gluon/util.lua
@@ -72,35 +72,36 @@ function node_id()
 end
 
 -- Generates a (hopefully) unique MAC address
--- The first parameter defines the function and the second
--- parameter an ID to add to the MAC address
--- Functions and IDs defined so far:
--- (1, 0): WAN (for mesh-on-WAN)
--- (1, 1): LAN (for mesh-on-LAN)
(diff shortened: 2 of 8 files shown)


NeoRaider:
* *primary_mac* must not be changed; only derived MAC addresses should use the hash function.
* I don't see a reason why we should make this specific for the Vocore. Just do it unconditionally for all hardware.
* The sys.exec construction isn't very nice, we should rather use something based on *popen* or look for a pure Lua solution

lemoer:
That would also be possible... 

* I can move the _hash_mac_ call  into the _generate_mac_ function and remove the vocore specific checks. Would that  be ok?
* I think a pure lua solution would be too much for this problem. I also think _popen_ is better than _sys.exec_. Are there any reasons why we should prefer a pure lua or c version?

lemoer:
It turns out that I forgot a major disadvantage of the _popen_-function. It is only able to open a subprocess either with a reading **or** a writing pipe. Are there any other possibilities in lua to open a process with _stdin_ and _stdout_ at the same time?

NeoRaider:
@lemoer, you can stitch together an extended popen out of nixio calls, similar to the code of the autoupdater: https://github.com/freifunk-gluon/packages/blob/master/admin/autoupdater/files/usr/lib/lua/autoupdater/util.lua#L25 (which doesn't use stdout, but workarounds another limitation of popen).

I'd still prefer a solution that doesn't need any additional processes; for example, we could add Lua bindings for the MD5 implementation of libubox (as we don't really need a cryptographic hash here, MD5 should be completely sufficient); implementing a simpler hash function like SipHash or Jenkins hash in Lua might be another option.

package/gluon-core/files/usr/lib/lua/gluon/util.lua
```diff
@@ -81,12 +82,25 @@ end
...
+  local m1, m2, m3, m4, m5, m6 = string.match(hashed, '(%x%x)(%x%x)(%x%x)(%x%x)(%x%x)(%x%x)')
 
-  return string.format('%02x:%02x:%02x:%s:%s:%s', m1, m2, m3, m4, m5, m6)
+  m1 = tonumber(m1, 16)
+  m3 = tonumber(m3, 16)
+  m6 = tonumber(m6, 16)
+
+  m1 = nixio.bit.bor(m1, 0x02)  -- set locally administered bit
+  m1 = nixio.bit.band(m1, 0xFE) -- unset the multicast bit
+  m3 = (m3+i) % 0x100           -- add interface identifier
```
NeoRaider: Also incrementing the third byte will yield a higher collision change than the scheme suggested in #648, as each node will use two or three instead of one 45bit prefix.

I'd like to see this function to be changed to use only a single parameter *i* which defines the index, and all uses of it to be adjusted according to #648.

lemoer:
I've tried to implement your suggestion. Comments are welcome.

package/gluon-client-bridge/files/lib/gluon/upgrade/320-gluon-client-bridge-wireless
```diff
@@ -21,14 +21,20 @@ local function configure_client(config, radio, index, suffix)
 
   uci:delete('wireless', name)
 
+  if index == 0 then
+    macaddr = util.generate_mac(0)
+  elseif index == 1 then
+    macaddr = util.generate_mac(3)
+  end
+
```
NeoRaider: Just change this to `macaddr = util.generate_mac(index * 3)`, so at least when no IBSS is used, a third WLAN adapter can be supported without problems.

package/gluon-mesh-batman-adv-core/files/lib/gluon/upgrade/320-gluon-mesh-batman-adv-core-wireless
```diff
@@ -53,14 +53,20 @@ local function configure_ibss(config, radio, index, suffix, disabled)
       )
     end
 
+    if index == 0 then
+      macaddr = util.generate_mac(2)
+    elseif index == 1 then
+      macaddr = util.generate_mac(5)
+    end
+
```
NeoRaider: Again: change to `macaddr = util.generate_mac(index * 3 + 2)`

package/gluon-mesh-batman-adv-core/files/lib/gluon/upgrade/320-gluon-mesh-batman-adv-core-wireless
```diff
@@ -85,14 +91,20 @@ local function configure_mesh(config, radio, index, suffix, disabled)
       }
     )
 
+    if index == 0 then
+      macaddr = util.generate_mac(1)
+    elseif index == 1 then
+      macaddr = util.generate_mac(4)
+    end
+
```
NeoRaider: Again: change to `macaddr = util.generate_mac(index * 3 + 1)`

package/gluon-client-bridge/files/lib/gluon/upgrade/320-gluon-client-bridge-wireless
lemoer: I also thought about this... I decided to do it this way, because it could lead to missunderstandings. Somebody could try to generate a mac for the interface _ibss4_ and the collision could be undetected for a while...

Maybe something like this?:
``` lua
if index < 3 then
  macaddr = util.generate_mac(index * 3)
end
```

I think an empty mac is more conspicuous than a colliding..

NeoRaider: Let's go with the following:

Drop the `% 8` from *generate_mac*, instead just return *nil* when *i* is too big. Change the functions which may use invalid indexes not to generate a config section at all when *generate_mac* returned *nil*.

lemoer:
I think the idea is ok. I implemented it.

package/gluon-core/files/usr/lib/lua/gluon/util.lua
```diff
@@ -71,22 +72,37 @@ function node_id()
...
+  m6 = tonumber(m6, 16)
+
+  m1 = nixio.bit.bor(m1, 0x02)  -- set locally administered bit
+  m1 = nixio.bit.band(m1, 0xFE) -- unset the multicast bit
+
+  -- It's necessary that the first 45 bits of the mac do
+  -- not vary on a single hardware interface, since some chips are using
+  -- a hardware mac filter. (e.g 'ramips-rt305x')
+
+  if i > 7 then return nil end  -- max allowed id (0b111)
```
NeoRaider: Maybe move this to the top?

NeoRaider: For the sake of completeness, checking the lower bound of 0 should also be added.

lemoer:
bump.

NeoRaider:
Looks good now. If you have tested this on a ralink-based system, we can merge this.

Please also squash this into a single commit (maybe with a less misleading commit message, as we're changing the MAC address schema for all hardware.)

NeoRaider:
I'll just do the squash myself, as Github recently gained a feature for that :)

PR #719, opened 2016-04-03
CONTRIBUTING.md: update links

kb-light:
link "packages" instead of "community repositories"
add link for rejected features to bottom link list
link "#gluon" to webirc
link the mentioned commit 2a93c58

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,8 +6,8 @@ different communities with different expectations and requirements, it is both
 essential and difficult to have contributions from the communities. While they
 are sometimes necessary to adapt Gluon to the needs of the communities, they
 also have to be adaptable enough to fit as many needs as possible. On the other
-hands, very special needs are better addressed in packages in [community
-repositories], because the Gluon maintainers would not use or test them and
+hands, very special needs are better addressed in [packages] in community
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -62,7 +61,9 @@ the pull request. This way, your change always consists of only one commit and
 can be merged in the instant everybody is content with the whole thing.
 
 
-[community repositories]: http://gluon.readthedocs.org/en/latest/user/site.html#packages
-[#gluon]: irc://irc.hackint.org/gluon
+[packages]: http://gluon.readthedocs.org/en/latest/user/site.html#packages
+[#gluon]: https://webirc.hackint.org/#gluon
```
NeoRaider: I'm not sure if we want to change this link to Webirc, as we occasionally have to set the channel to +r ...

kb-light:
I see that using irc could be a problem if #gluon is set to +r and one has no hackint account. But I can't see, how this effects webirc, because it seems to be possible to login with nickname and password. So it may be better to mention that an account is recommended.
But one can also argue that webirc is not needed, if one is using an irc-client and has an account.

I don't care, if the consent is to delete the webirc link, I will update this PR.

NeoRaider:
If the webirc supports accounts, this is fine.

PR #725, opened 2016-04-08
gluon-next-node: mark local-node ip6 as deprecated

tcatm:
Prevent the local-node ip6 from being eligible for source address
selection. This is highly undesireable in a layer 3 mesh.

(Also adds documentation for gluon-next-node)

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -53,6 +53,7 @@ Packages
    package/gluon-client-bridge
    package/gluon-ebtables-filter-multicast
    package/gluon-ebtables-filter-ra-dhcp
+   package/gluon-next-node
 
 Releases
 --------
--- a/docs/package/gluon-next-node.rst
+++ b/docs/package/gluon-next-node.rst
@@ -5,7 +5,7 @@ This package provides a virtual interface (tied to *br-client*) called *local-no
 using the same MAC, IP4 and IP6 across all nodes in a mesh. Thus, the node that
 the client is currently connected to, can always be reached under a known address.
 
-The IP6 is marked es deprecated to prevent it from being used as a source
+The IP6 is marked as deprecated to prevent it from being used as a source
 address for packages originating from a node.
 
 site.conf
--- a/docs/package/gluon-next-node.rst
+++ b/docs/package/gluon-next-node.rst
@@ -1,21 +1,22 @@
 gluon-next-node
 ===============
 
-This package provides a virtual interface (tied to *br-client*) called *local-node*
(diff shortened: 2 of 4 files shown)


rotanid:
found a typo in "marked es deprecated", should be "as"
and i would prefer "IPv4" and "IPv6" over IP4/IP6 in comments and documentation, as they are the common terms for these protocols

tcatm:
@rotanid Thanks for catching that typo!

NeoRaider:
I'm wondering if the gluon-next-node package is even still useful in a layer 3 mesh. I'd expect that br-client could just be assigned the local-node MAC and IP addresses directly, without the need for a macvlan device.

Marking the IP address as deprecated would still be necessary in this case, of course.

NeoRaider:
Also, it might make make sense to extend the builtin static protocol of netifd instead of defining our own script-based one.

tcatm:
I'd like to keep the next-node-feature in a seperate package shared by all mesh implementations. Technical, it will work when assigning the addresse directly to br-client but I then we'd need to have different client-bridge packages for batman-adv and babel based meshes.

I'd like to keep the script-based protocol for now. That's exactly how netifd is supposed to be extended.

NeoRaider:
The next-node package is very specific to layer 2 meshing (it includes lots of ebtables rules etc.). Sharing this implementation with layer 3 implementations would require us to include a lot of unneeded code... While reusing packages for everything "below" the mesh (for example the mesh VPN) should be possible, layer 2 and layer 3 meshs differ fundamentally, so reusing the client bridge and anything on top of that seems rather awkward to me.

When extending netifd, I'd rather add an attribute *deprecated* to the existing static protocol, so all other attributes can be reused. Making it a separate protocol doesn't make sense at all in my opinion.

christf:
switching the ip addresses makes this change obsolete

NeoRaider:
Not really; but this PR will need rebasing.

Having this properly supported in netifd would be preferable to the "static_deprecated" proto though.

christf:
I am sorry, I did not completely apply this patch. I made the observation that when setting the protocol as in this change to static_deprecated then the whole mesh falls apart. This is however probably due to the messing netifd handler.

NeoRaider:
I've added support for deprecated static addresses to netifd: http://patchwork.ozlabs.org/patch/704794/

As this makes pretty much all code in this PR outdated, I'll close this. When my patch (or a similar one) makes it into netifd, only a one-line change in *gluon-client-bridge* will be required.

PR #726, opened 2016-04-12
add package mesh-vpn-fastd-babel

christf:


NeoRaider:
Let's ignore fastd for now. When we have a working BABEL-based setup of everything else, we can try to make the mesh-vpn-fastd package generic enough to work with either BABEL or batman-adv. We should avoid having a separate package if we can.

christf:
On Tue, Apr 12, 2016 at 12:21:12PM -0700, Matthias Schiffer wrote:
> Let's ignore fastd for now. When we have a working BABEL-based setup
> of everything else, we can try to make the mesh-vpn-fastd package
> generic enough to work with either BABEL or batman-adv. We should
> avoid having a separate package if we can.
Apparently I misunderstood that you preferred to have a separate
package.

PR #730, opened 2016-04-21
Add support for carambola 2 board

jannic:
Device information can be found at:
http://www.8devices.com/products/carambola-2
https://wiki.openwrt.org/toh/8devices/carambola2

I only did some minimal testing of gluon on the carambola 2 development
board:

 - Config mode works
 - Connects to Wifi Mesh
 - Allows clients to connect

Notably, autoupgrade has not yet been tested.

Change to 010-primary-mac is necessary as the mac address printed
on the sticker is the one of eth0, not the wifi mac.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -78,6 +78,10 @@ Supported Devices & Architectures
 ar71xx-generic
 ^^^^^^^^^^^^^^
 
+* 8devices
+
+  - Carambola 2
+
 * ALFA Network
 
   - AP121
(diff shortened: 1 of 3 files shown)


targets/ar71xx-generic/profiles.mk
```diff
@@ -302,3 +302,8 @@ $(eval $(call GluonProfile,MR16,rssileds))
 $(eval $(call GluonProfileFactorySuffix,MR16))
 $(eval $(call GluonModel,MR16,mr16,meraki-mr16))
 $(eval $(call GluonModelAlias,MR16,meraki-mr16,meraki-mr66))
+
+# Carambola 2
```
NeoRaider: Please also add a comment with the vendor (like it's there for the other vendors)

NeoRaider:
The device should also be added to the list of supported hardware in *docs/index.rst*.

jannic:
ok.

docs/index.rst
```diff
@@ -180,6 +180,10 @@ ar71xx-generic
   - My Net N600
   - My Net N750
 
+* 8devices
+
+  - Carambola 2
```
NeoRaider: I'd rather put 8devices at the beginning of the list, numbers are usually sorted before alphabet chars.

jannic:
as you like :-)

PR #732, opened 2016-04-23
babel: populate babel-interfaces with the content of /lib/gluon/core/mesh_interfaces

christf:
this will cause babel to be run on the mesh-interfaces


christf:
[] . und - in interface-Namen durch _ ersetzen
[] alle bereits in /etc/config/babeld vorhandenen interface-sections löschen

package/gluon-mesh-babel/files/lib/gluon/upgrade/330-gluon-mesh-babel-interfaces
```diff
@@ -13,6 +14,15 @@ if site.mesh_on_wan then
   uci:add_to_set('firewall', 'mesh_babel', 'network', 'wan')
 end
 
+uci:delete_all('babeld', 'interface')
+
+pcall(function() 
+  for interface in io.lines(interfaces) do 
+    int = string.gsub(interface,'[.-]','_')
```
NeoRaider: You can change this to `interface:gsub('[.-]', '_')`. Or even better, to `interface:gsub('[^%w_]', '_')`, which will catch all invalid chars.

Also, let's give the variable a better name than *int*, something like *section*. It should also be declared *local*.

christf: done.

NeoRaider:
As I'm currently working on a proper way to handle the list of mesh interfaces independently of the mesh protocol, I've noticed to the following issue: As the list of mesh interfaces may be changed in the expert mode (mesh-on-lan/wan), either
* we need a trigger which will regenerate the UCI config from the interface list or
* we can't write this to UCI at all, but instead use our own init script to start babeld with the correct config.

I think it would be nice to be able to avoid UCI here. This would mean that we don't use */etc/config/babeld* at all.

christf:
That would mean the babel config must be generated without uci. We
already know of all the relevant interfaces, but ip prefixes and routing
table need to be identified.

The prefix can be read from /lib/gluon/site.json
As for the routing table - what should be the way here? My current babel
config (Freifunk FFM prefix) is this:
ipv6-subtrees true
Export-table 10
Import-table 255
import-table 11
Interface eth0 # (the interface name obviously is garbage, this is due to
# tests of the /lib/gluon/core/mesh_interfaces code)
Redistribute-ip fddd:5d16:b5dd:1::/64 eq 128 allow # (babel-mesh.prefix)
Redistribute-ip fddd:5d16:b5dd:2::/64 eq 128 allow # (prefix6)
Redistribute local deny 
Redistribute ip ::/0 eq 0 allow



The current config On Sun, Apr 24, 2016 at 07:47:18AM -0700, Matthias Schiffer wrote:
> As I'm currently working on a proper way to handle the list of mesh interfaces
> independently of the mesh protocol, I've noticed to the following issue: As the
> list of mesh interfaces may be changed in the expert mode (mesh-on-lan/wan),
> either

>   • we need a trigger which will regenerate the UCI config from the interface
>     list or
>   • we can't write this to UCI at all, but instead use our own init script to
>     start babeld with the correct config.

> I think it would be nice to be able to avoid UCI here. This would mean that we
> don't use /etc/config/babeld at all.

> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub*


-- 
()  ascii ribbon campaign - against html e-mail
/\  against proprietary attachments

NeoRaider:
All of this config can either be static (hardcoded) or be read from the site config.

christf:
- [ ] create custom init script fro babel

christf:
the init script of babeld is left in place because babeld comes with all interfaces disabled by default via uci.

christf:
- [ ] squash commits
- [ ] re-add init script for package gluon-mesh-babeld because it got lost in the commit-chain

PR #745, opened 2016-04-29
reduce batman dependency in status page

christf:
this requiries gluon.mesh

### DIFF
--- a/package/gluon-status-page-api/files/lib/gluon/status-page/www/cgi-bin/dyn/neighbours-nodeinfo
+++ b/package/gluon-status-page-api/files/lib/gluon/status-page/www/cgi-bin/dyn/neighbours-nodeinfo
@@ -1,13 +1,19 @@
-#!/bin/sh
+#!/usr/bin/lua 
+local mesh = require 'gluon.mesh' 
+local querystring = os.getenv("QUERY_STRING") 
 
-badrequest() {
-        echo 'Status: 400 Bad Request'
-        echo
-        exit 1
-}
+print('Access-Control-Allow-Origin: *\n') 
 
-echo 'Access-Control-Allow-Origin: *'
+for interface, _ in pairs(mesh.interfaces()) do 
+  if querystring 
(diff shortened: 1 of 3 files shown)


christf:
as tcatm requested, raising this for the master branch instead of babel.

package/gluon-status-page-api/files/lib/gluon/status-page/www/cgi-bin/dyn/neighbours-nodeinfo
```diff
@@ -8,6 +8,7 @@ badrequest() {
 
 echo 'Access-Control-Allow-Origin: *'
 
-batctl if | cut -d: -f1 | grep -qxF "$QUERY_STRING" || badrequest
+#batctl if | cut -d: -f1 | grep -qxF "$QUERY_STRING" || badrequest
```
rotanid: why leaving the old commands there? if we need them again, we could look them up in git history.
(my comment also applies to the other file changes in your commit)

christf:
right.

christf:
- [ ] re-implement shell scripts in commits in lua

tcatm:
Does this output an empty line after "Status: 400 Bad Request"?

christf:
yes it does.

NeoRaider:
The "blocked" label marks issues/PRs that depend on other issues/PRs (in this case the gluon.mesh support in #747).

rotanid:
removing the blocked label as the dependency was merged in early september.
needs a rebase now, if you still want that change, @christf

christf:
@rotanid this needs rework. gluon.mesh is not available any more instead interfaces from uci should be used having proto = gluon_mesh there is already a reference implementation in the babel init script which should be used here as well. better yet it should be put in a central place and be called from both of these scripts.

christf:
@lemoer  referencinghim so he is aware too.

christf:
this is superseeded by the work in #934

PR #747, opened 2016-05-01
[RFC] Introduce generic method for packages to provide mesh interfaces

NeoRaider:
This allows packages to configure mesh interfaces (wireless mesh over IBSS/11s, mesh VPN, mesh-on-WAN/LAN) without storing protocol-specific data in UCI. The list of interfaces is stored in JSON format (with some extra information) in */lib/gluon/core/mesh_interfaces.json* and can be queried and managed using the *gluon.mesh* Lua module.

At the moment, there is no way to configure additional mesh interfaces, so this is a RFC for now.


christf:
I like it and I feel a similar option should be created for the client-interfaces. This will lead to much more flexibility in the future for configuring client and mesh interfaces.

christf:
in the meantime I was also able to try this. it worked. :)

PR #756, opened 2016-05-08
Add D-Link DIR-505 rev. A2 as alias of D-Link DIR-505 rev. A1

copyrights:
Due to https://wiki.openwrt.org/toh/d-link/dir-505 rev. A1 and A2 are the same, but differ on outer hardware. I myself have flashed two DIR-505 rev. A2 with DIR-505 rev. A1 firmware (2016.1.4) without any problems, yet.

### DIFF
--- a/targets/ar71xx-generic/profiles.mk
+++ b/targets/ar71xx-generic/profiles.mk
@@ -193,10 +193,12 @@ endif
 
 ## D-Link
 
-# D-Link DIR-505 rev. A1
+# D-Link DIR-505 rev. A1/A2
 
 $(eval $(call GluonProfile,DIR505A1))
-$(eval $(call GluonModel,DIR505A1,dir-505-a1,d-link-dir-505-rev-a1))
+$(eval $(call GluonModel,DIR505A1,dir-505-a1-a2,d-link-dir-505-rev-a1))
+
+$(eval $(call GluonModelAlias,DIR505A1,d-link-dir-505-rev-a1,d-link-dir-505-rev-a2))
 
 # D-Link DIR-615 rev. C1
 $(eval $(call GluonProfile,DIR615C1))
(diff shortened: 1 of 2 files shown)


mweinelt:
They differ only in one minor thing afaik:
> exchangeable mains plug

copyrights:
My patch wouldn't defect hardware differences as there is no 'product-info' like on TP-Link CPE. The hardware string will still be "D-Link DIR-505 rev. A1" (if nobody change it in OpenWRT). But it works with rev. A2 and people shouldn't be (more) afraid to flash there device because it says "rev. A2".

NeoRaider:
@copyrights, properly detecting the revision for the model string is on my TODO, after that we can just call the images -a1 and -a2.

copyrights:
@NeoRaider if you need any info from a real rev. A2 hardware, let me know.

tcatm:
Wouldn't it make sense to use the same image for both and detect the model string at runtime?

NeoRaider:
@tcatm, yes, I intend to write a patch to detect the model string. When we do that, we'll have to provide images with -a1 and -a2 suffix (but the -a2 will just be an alias).

targets/ar71xx-generic/profiles.mk
```diff
@@ -193,10 +193,12 @@ endif
 
 ## D-Link
 
-# D-Link DIR-505 rev. A1
+# D-Link DIR-505 rev. A1/A2
 
 $(eval $(call GluonProfile,DIR505A1))
-$(eval $(call GluonModel,DIR505A1,dir-505-a1,d-link-dir-505-rev-a1))
+$(eval $(call GluonModel,DIR505A1,dir-505-a1-a2,d-link-dir-505-rev-a1))
```
NeoRaider: You can't just change the source filename from `dir-505-a1` to `dir-505-a1-a2`, the file won't be found like this.

I'll fix this up myself.

NeoRaider:
Fixed up and merged manually.

PR #760, opened 2016-05-10
openwrt: enabled iw connect module

oni303:
This re-enables the connect module of iw.
The connect module was disabled beside others to save some space in upstream openwrt, recently.

It is needed for the fallback [autoupdater](https://git.c3pb.de/freifunk-pb/ffho-packages/blob/master/ffho/ffho-autoupdater-wifi-fallback/) by Freifunk Hochstift or Paderborn.
The uncompressed binary size increases by 513bytes without fs-compression, a workaround for the connect module by [fuzzle](https://forum.freifunk.net/t/fork-von-ffho-autoupdater-wifi-fallback-gesucht/11423/16) is approx. 581bytes in shell script.
Both will be much smaller on the device, so to have the feature to join a wifi approx. the same space is needed. I would prefer to use the "official" rather than using a workaround.

Let me know if something in this commit needs to be changed or if there are serious doubts about it.

### DIFF
--- a/patches/openwrt/0033-reenabled-iw-connect.patch
+++ b/patches/openwrt/0033-reenabled-iw-connect.patch
@@ -1,0 +1,21 @@
+From: Georg von Zengen <vonzengen@ibr.cs.tu-bs.de>
+Date: Sun, 8 May 2016 18:55:28 +0200
+Subject: reenabled iw connect
+
+diff --git a/package/network/utils/iw/patches/304-reinsert_connect.patch b/package/network/utils/iw/patches/304-reinsert_connect.patch
+new file mode 100644
+index 0000000..4641775
+--- /dev/null
++++ b/package/network/utils/iw/patches/304-reinsert_connect.patch
+@@ -0,0 +1,11 @@
++--- a/Makefile
+++++ b/Makefile
++@@ -16,7 +16,7 @@ OBJS = iw.o genl.o event.o info.o phy.o
++ 	interface.o ibss.o station.o survey.o util.o ocb.o \
++ 	mesh.o mpath.o mpp.o scan.o reg.o version.o \
++ 	reason.o status.o link.o offch.o ps.o cqm.o \
++-	bitrate.o vendor.o
+++	bitrate.o vendor.o connect.o
++ OBJS += sections.o
++ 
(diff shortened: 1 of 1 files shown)


kb-light:
- the correct link is [ffho-autoupdater-wifi-fallback](https://git.c3pb.de/freifunk-pb/ffho-packages/tree/master/ffho/ffho-autoupdater-wifi-fallback)
- our original patch can be found [here](https://git.c3pb.de/freifunk-pb/site-ffho/blob/experimental/patches/0002-openwrt-patch-iw.patch) 
- if the workaround by fuzzle can be integrated in our script, please send patches
- (the discussion in the forum is not up to date)

oni303:
Thanks for adding the correct link and the original patch.
I think the best way to integrate fuzzles workaround is to add it into the wpa_supplicant.
But I am not sure about the advantages of a workaround, as I said before.

NeoRaider:
I'll check with the other LEDE people if we can enable the connect command again, I'd like to avoid having more non-upstreamable patches in Gluon.

kb-light:
@oni303 i think both are workarounds and i agree to @NeoRaider that adding the patch to Gluon is not the best way, even because it is just used for one (community/non Gluon) package.

The prefered solutiontion should be one of:
- adding the patch upstream
- integrate fuzzles workaround into the package.

oni303:
Ok I had a much deeper look into fallback updater yesterday. 
As far as I know by now the patch I provided here is only needed for the wpa_supplicant fake implementation but I could not find any use of the wpa_supplicant in the rest of the package.
The fallback updater itself actually uses something very similar to the workaround.
Not sure what the sens of the wpa_supplicant was in the first place.

kb-light:
@oni303 there are two reasons for the wpa_supplicant fake:
1. with the original wpa_supplicant the firmware image is to big for routers with 4MB flash
2. it is automatically executed via "wifi"

gvz:
A little work in progress update:
for now is looks like either wpa_supplicant or iw connect is necessary to be able to connect to a wifi as a client. By now I am not sure why but I will take a look at it and share my findings here.

kb-light:
I think this could be closed and we discuss it elsewhere.

PR #761, opened 2016-05-11
externalize radvd arguments for gluon-radvd

christf:


package/gluon-mesh-batman-adv-core/files/lib/gluon/radvd/arguments
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/lua
+local site = require "gluon.site_config"
+print("-i br-client -p " .. site.prefix6)
+
```
NeoRaider: There is a superfluous newline here.

NeoRaider: *package/gluon-mesh-batman-adv-core/files/lib/gluon/radvd/arguments* is missing the executable flag, this can't work.

package/gluon-radvd/files/etc/init.d/gluon-radvd
```diff
@@ -7,7 +7,7 @@ SERVICE_DAEMONIZE=1
 
 
 start() {
-	service_start /usr/sbin/uradvd -i br-client -p $(lua -e 'print(require("gluon.site_config").prefix6)')
+	service_start /usr/sbin/uradvd $(/lib/gluon/radvd/arguments)
```
NeoRaider: I think the start function should check if */lib/gluon/radvd/arguments* exists and is executable before trying to use it.

NeoRaider:
This commit could also remove the gluon-mesh-batman-adv dependency of gluon-radvd.

I'm not sure what to do with the gluon-ebtables dependency though... maybe merge the package into gluon-mesh-batman-adv-core? If we have use for ebtables without batman-adv, we could keep the package separate, but make gluon-mesh-batman-adv-core depend on gluon-ebtables, or merge gluon-ebtables into gluon-core. It must just be ensured that the ebtables are available when gluon-radvd is used in a batman-adv mesh.

christf:
is there a way of having a dependency to one of gluon-mesh-batman-adv-core or gluon-mesh-babel? because one of them must be installed for this to work.

as for the check in the init script for the existence of the argument-script:
imo this should be handled by the build system by a dependency check.

The rest of the feedback is incorporated already

NeoRaider:
I'd prefer not to have a dependency on gluon-mesh-batman-adv-* or gluon-mesh-babel, and just make gluon-radvd a no-op when neither is installed.

An alternative would be to add something like PROVIDES:=gluon-radvd-provider to gluon-mesh-batman-adv-core and gluon-mesh-babel, and make gluon-radvd depend on gluon-radvd-provider.

christf:
Right. Maybe tcatm can state of ebtables is needed für arpproxyd? 
Viele Grüße 

Christof 
-- 
Diese Nachricht wurde mobil gesendet.

Am 12. Mai 2016 02:04:17 MESZ, schrieb Matthias Schiffer <notifications@github.com>:

tcatm:
I don't see why this PR should be concerned with arpproxyd (it doesn't even exist yet!).

christf:
the start script now checks for /lib/gluon/radvd/arguments to be executable before executing it.
it fails silently if it does not exist.

NeoRaider:
Thanks, I've removed the PROVIDES dependency and applied your changes.

PR #763, opened 2016-05-12
GPIO control support: Add CPE210

corny:


NeoRaider:
cpe510 is the board name which is used for the CPE210 as well, so this patch is wrong.

PR #767, opened 2016-05-16
Add support for OpenMesh hardware available in OpenWrt CC

ecsv:
Following hardware was tested with gluon-2016.1.4

 - MR600
 - MR600v2
 - MR900
 - MR900v2
 - OM2P
 - OM2P-HS
 - OM2P-HSv2
 - OM2P-LC
 - OM2Pv2
 - OM5P
 - OM5P-AN

OpenMesh hardware is for example used at https://meshviewer.chemnitz.freifunk.net/#!v:m;n:ac867400ec02

----

The gluon-luci-admin change itself is a fix but there will be also a patch for OpenWrt/LEDE to modify the behavior of `platform_check_image_openmesh`

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -130,6 +130,16 @@ ar71xx-generic
 
   - Omega
 
+* OpenMesh
+
+ - MR600 (v1, v2)
+ - MR900 (v1, v2)
+ - OM2P (v1, v2)
+ - OM2P-HS (v1, v2)
+ - OM2P-LC
+ - OM5P
+ - OM5P-AN
+
 * TP-Link
 
   - CPE210 (v1.0, v1.1)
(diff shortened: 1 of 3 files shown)


mmalte:
I love it! I'll test my OM2P and MR200v2 in the next days!

NeoRaider:
Thanks a lot.

Regarding the GluonProfileSysupgradeSuffix settings: I'll take it like this for now, but I think it would be nice to change OpenWrt/LEDE to generate sysupgrade images identical to the factory ones, to avoid confusing users. Not having a sysupgrade image usually means that a board does not have sysupgrade support.

mmalte:
OM2P-LC and MR900v2 are working like a charm.

5GHz Radio is working simultaneous in Mesh and Client mode as expected.

PR #768, opened 2016-05-16
Gluon mesh provider

christf:
this will abstract the mesh-protocol provider from the actual protocol.


christf:
tthis should be reconsidered, once PROVIDE work reliably in packages. Until then there really is no point. Unfortunately I only learned this after raising the PR. closing for now because opwenwrt is b0rked.

PR #775, opened 2016-05-20
Add support for OpenMesh hardware available in OpenWrt trunk (2016.1.x)

ecsv:
Following hardware was added (mostly tested with gluon 2016.1.x)

* MR1750
* OM5P-AC
* OM5P-ACv2

The patches were forwarded to OpenWrt CC before they were integrated in gluon. Each patch also contains a link to the entry in patchwork.

The GluonProfileSysupgradeSuffix workaround was also replaced with proper build sysupgrade images.

Please only pull this branch when you've already integrated #774 in master. It is just to help in the backporting to v2016.1.x (if you want to do that).

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -131,12 +131,14 @@ ar71xx-generic
 
 * OpenMesh
 
+ - MR1750
  - MR600 (v1, v2)
  - MR900 (v1, v2)
  - OM2P (v1, v2)
  - OM2P-HS (v1, v2)
  - OM2P-LC
  - OM5P
+ - OM5P-AC (v1, v2)
  - OM5P-AN
 
 * TP-Link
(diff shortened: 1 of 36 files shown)


NeoRaider:
Thanks for your effort, but I think this series is a bit too massive for our stable branch. We plan to release Gluon v2016.2 till the end of June, I hope you can wait that long :)

mmalte:
These devices are built to work in Mesh environments. 

The MR900v2 I've got running is by far the fastest mesh device (not mesh vpn) I had running so far. 

If I've got you right you're still planning to include the devices from the last pull request in 2016.1.5, but  these three devices are more complex?

NeoRaider:
@mmalte, right, the other models were already supported by OpenWrt CC, which made adding support to v2016.1.x trivial. This patch series however adds more than 30 patches, so I'd rather like to avoid having it in the stable branch.

ecsv:
Yes, sounds good.

PR #777, opened 2016-05-21
respondd modules: Use proper formatting of doubles

jplitza:
Most doubles that are delivered via respondd have limited input
precision, but are converted with up to 17 digits of precision. That can
cause ugly blowups like 0.2800000000000001 in the output, which is
avoided by specifying better format strings (like "%.2f" in most cases).

### DIFF
--- a/package/gluon-mesh-batman-adv-core/src/respondd.c
+++ b/package/gluon-mesh-batman-adv-core/src/respondd.c
@@ -501,7 +501,9 @@ static struct json_object * get_batadv(void) {
 
 		struct json_object *obj = json_object_new_object();
 		json_object_object_add(obj, "tq", json_object_new_int(tq));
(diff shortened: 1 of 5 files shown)


jplitza:
By the way, alternatively, one could capture the input as double as well as as string and supply the precise string representation to [`json_object_new_double_s()`](https://github.com/json-c/json-c/blob/master/json_object.h#L646). But appeared to be more difficult to me.

NeoRaider:
Thanks :) Could you also add this to `gluon-node-info/src/respondd.c`? I think even for geocoordinates 8 decimal places are more than enough, and floating point errors should be hidden everywhere.

jplitza:
Done, including an updated filename/seqno for the patchfile. And while I was at it, also slightly simplified the code in that respondd module (in a separate commit).

PR #780, opened 2016-05-22
gluon-mesh-batman-adv-core: disable bridge port learning on bat0

T-X:
The mesh side has become fairly huge in many communities. Up to
a few thousand entries can currently be found in the forwarding
database (fdb) of a bridge for its bridge port bat0.

The bridge fdb is kind of redundant to the batman-adv global translation
table here. Therefore this patch tries to reduce memory footprint by
following an approach similar to the IGMP/MLD split patchset approach:

Make the bridge oblivious not only regarding multicast listeners towards
the mesh but with this patch unicast hosts on the mesh, too.

If the destination of an ethernet frame is known by the bridge to be a
local one, then the frame is forwarded to the according port. If it is
unknown, then the frame is forwarded to the wifi AP interface and bat0.

mac80211 and batman-adv then know whether to drop or forward a frame
further through their own book-keeping.

Note that unicast-flood is not disabled for the wifi AP bridge port, nor
is learning disabled on the wifi AP. This is mainly to keep the
configuration in UCI and according setup scripts simple ;). However, not
disalbling unicast-flood on the wifi AP interface might also give a
minor latency improvement for newly joining wifi clients.

Signed-off-by: Linus Lüssing <linus.luessing@c0d3.blue>

### DIFF
--- a/package/gluon-mesh-batman-adv-core/luasrc/lib/gluon/upgrade/310-gluon-mesh-batman-adv-core-mesh
+++ b/package/gluon-mesh-batman-adv-core/luasrc/lib/gluon/upgrade/310-gluon-mesh-batman-adv-core-mesh
@@ -51,8 +51,19 @@ uci:section('network', 'interface', 'bat0',
 		    proto = 'none',
 		    macaddr = sysconfig.primary_mac,
 		    multicast_router = 2,
+		    learning = 0,
 	    }
 )
+
+uci:delete('network', 'client_lan')
+if sysconfig.lan_ifname then
+	uci:section('network', 'interface', 'client_lan',
+		    {
+			    unicast_flood = 0,
+		    }
+	)
+	uci:set('network', 'client_lan', 'ifname', sysconfig.lan_ifname)
+end
 
 uci:save('network')
 


T-X:
PS: The included netifd patch was sent to openwrt-devel/lede-devel and is pending for review and upstream netifd inclusion.

T-X:
New version fixes a bug for the unicast-flood configuration (bug in the gluon-upgrade script part).

NeoRaider:
The netifd patch is included in Gluon now, please rebase.

package/gluon-mesh-batman-adv-core/files/lib/gluon/upgrade/310-gluon-mesh-batman-adv-core-mesh
```diff
@@ -51,9 +51,22 @@ uci:section('network', 'interface', 'bat0',
...
+uci:delete('network', 'client_lan')
+if sysconfig.lan_ifname then
+	uci:section('network', 'interface', 'client_lan',
+		    {
+			    unicast_flood = 0,
+		    }
+	)
+	for _, lanif in ipairs(lutil.split(sysconfig.lan_ifname, ' ')) do
+		uci:set('network', 'client_lan', 'ifname', lanif)
+	end
```
NeoRaider: This loop won't work if there are multiple LAN interfaces. Maybe you meant `add_to_set`?

Just adding `ifname = lutil.split(sysconfig.lan_ifname, ' ')` to the uci:section above seems simpler though, or am I overlooking something?

T-X: Hm, your two suggestions create a "list ifname" but a "list ifname eth0.1" for instance does not seem to work, unicast_flood is not set correctly upon reboot. Only an "option" works.

Is there supposed to ever be more than one lan interface?

T-X:
Rebased to master, removed netifd patch as available upstream now.

package/gluon-mesh-batman-adv-core/luasrc/lib/gluon/upgrade/310-gluon-mesh-batman-adv-core-mesh
```diff
@@ -51,9 +51,22 @@ uci:section('network', 'interface', 'bat0',
...
 
+uci:delete('network', 'client_lan')
+if sysconfig.lan_ifname then
+	uci:section('network', 'interface', 'client_lan',
+		    {
+			    unicast_flood = 0,
+		    }
+	)
+	for _, lanif in ipairs(lutil.split(sysconfig.lan_ifname, ' ')) do
+		uci:set('network', 'client_lan', 'ifname', lanif)
```
NeoRaider: 1. We should investigate why the list doesn't work here
2. In any case, this code is plain wrong. If you want to use an option, use `uci:set('network', 'client_lan', 'ifname', sysconfig.lan_ifname)`, your loop will replace the old value in each run otherwise!

PR #783, opened 2016-05-23
ar71xx-generic: Use eth0 as OpenMesh primary mac address

ecsv:
The OpenMesh devices have a sticker with the eth0 mac address on the bottom. Also all other mac addresses are calculated based on this address. Therefore, it is better to use this as primary mac address instead of the WiFi mac address.

### DIFF
--- a/package/gluon-core/files/lib/gluon/upgrade/010-primary-mac
+++ b/package/gluon-core/files/lib/gluon/upgrade/010-primary-mac
@@ -24,7 +24,15 @@ end
 
 if platform.match('ar71xx', 'generic', {'tl-wdr3600', 'tl-wdr4300'}) then
   table.insert(try_files, 1, '/sys/class/ieee80211/phy1/macaddress')
-elseif platform.match('ar71xx', 'generic', {'unifi-outdoor-plus', 'carambola2'}) then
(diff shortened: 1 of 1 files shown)


package/gluon-core/files/lib/gluon/upgrade/010-primary-mac
```diff
@@ -28,6 +28,16 @@ elseif platform.match('ar71xx', 'generic', {'unifi-outdoor-plus', 'carambola2'})
...
+elseif platform.match('ar71xx', 'generic', {'mr600', 'mr600v2',
+                                            'mr900', 'mr900v2',
+                                            'mr1750',
+                                            'om2p',
+                                            'om2pv2', 'om2p-hs', 'om2p-hsv2', 'om2p-hsv3',
+                                            'om2p-lc',
+                                            'om5p',
+                                            'om5p-an',
+                                            'om5p-ac', 'om5p-acv2'}) then
+  table.insert(try_files, 1, '/sys/class/net/eth0/address')
```
NeoRaider: There is already an elseif branch for eth0, please change that one instead of adding a new one.

ecsv:
Ok, merged those two together and replaced the commit

PR #785, opened 2016-05-24
Add autoupdater fallback Doc

PowerPan:
Update Gluon docs
Add missing parameter documentation --fallback


tcatm:
lgtm

nit: We prefer commits, that are prefixed, e.g. "docs: autoupdater's fallback option".

PowerPan:
ok i will make a new pull request with the right commit syntax style

PR #792, opened 2016-05-30
docs: how to show the list of supported targets

rubo77:
The complete list of supported targets is only shown, when you call `make` with setting an empty string for `GLUON_TARGET`


rubo77:
Maybe the empty string is only needed in case you have a line like

    GLUON_TARGET ?= ar71xx-generic

in your `site.mk`?

NeoRaider:
Correct, the new text should reflect this.

rotanid:
@rubo77 are you planning on updating this PR to reflect this?

PR #799, opened 2016-06-13
ar71xx: add GL-AR150 support

jannic:
I added support for GL-AR150 because a friend bought such a device and wanted to use it with Freifunk.

The device worked with branch v2016.1.x, and the same patch also applies to master. Unfortunately, he broke the router (via static discharge, not software-related), so I'm unable to test it with a current gluon version. Therefore, for now, I marked the device as 'BROKEN'.

### DIFF
--- a/patches/openwrt/0095-ar71xx-add-GL-AR150-support.patch
+++ b/patches/openwrt/0095-ar71xx-add-GL-AR150-support.patch
@@ -1,0 +1,326 @@
+From: Jan Niehusmann <jan@gondor.com>
+Date: Wed, 15 Jun 2016 14:57:47 +0200
+Subject: ar71xx: add GL-AR150 support
+
+patch from https://github.com/domino-team/OpenWrt-patches
+
+https://raw.githubusercontent.com/domino-team/OpenWrt-patches/master/AR150%2C%20AR300%2C%20Domino%20-%20CC1505/000-gl-ar150-cc.patch
+
+(and fixed indentation in target/linux/ar71xx/base-files/lib/ar71xx.sh)
+
+diff --git a/target/linux/ar71xx/base-files/etc/uci-defaults/01_leds b/target/linux/ar71xx/base-files/etc/uci-defaults/01_leds
+index 129f6b7..e29c1b5 100644
+--- a/target/linux/ar71xx/base-files/etc/uci-defaults/01_leds
++++ b/target/linux/ar71xx/base-files/etc/uci-defaults/01_leds
+@@ -182,6 +182,10 @@ dlan-pro-1200-ac)
+ 	ucidef_set_led_trigger_gpio "plcr" "dLAN" "devolo:error:dlan" "16" "0"
+ 	;;
+ 
++gl-ar150)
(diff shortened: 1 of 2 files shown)


patches/openwrt/0095-ar71xx-add-GL-AR150-support.patch
```diff
@@ -0,0 +1,115 @@
...
+ CONFIG_ATH79_MACH_F9K1115V2=y
++CONFIG_ATH79_MACH_GL_AR150=y
+ CONFIG_ATH79_MACH_GL_INET=y
+ CONFIG_ATH79_MACH_GS_MINIBOX_V1=y
+ CONFIG_ATH79_MACH_GS_OOLITE=y
+diff --git a/target/linux/ar71xx/generic/profiles/gl-connect.mk b/target/linux/ar71xx/generic/profiles/gl-connect.mk
+deleted file mode 100644
+index e9377db..0000000
+--- a/target/linux/ar71xx/generic/profiles/gl-connect.mk
++++ /dev/null
```
NeoRaider: Why is this file removed? We actually use the GLINET profile...

patches/openwrt/0095-ar71xx-add-GL-AR150-support.patch
```diff
@@ -0,0 +1,115 @@
...
+index 5768286..718ee0b 100755
+--- a/target/linux/ar71xx/base-files/lib/ar71xx.sh
++++ b/target/linux/ar71xx/base-files/lib/ar71xx.sh
+@@ -458,6 +458,9 @@ ar71xx_board_detect() {
+ 		name="gl-inet"
+ 		gl_inet_board_detect
+ 		;;
++	*"GL AR150")
++        name="gl-ar150"
++        ;;
```
NeoRaider: Indentation should use tabs instead of spaces, as the other entries.

jannic: Taken verbatim from upstream. I could fix that, of course.

patches/openwrt/0095-ar71xx-add-GL-AR150-support.patch
jannic: Upstream moved the definition from gl-connect.mk to gli.mk. I'm not sure why my commit is missing that file.
Probably I make some mistake when doing 'make update-patches'.
Can't reconstruct it now, as it was done several  weeks ago. But I'll have a look.
(In fact, I wonder why that didn't lead to an error when compiling gluon?)

NeoRaider: OpenWrt behaves a bit weird when profiles are missing - it will just make the non-existing profile behave like the "default" profile that contains all images. Build will take a lot longer, but it won't cause an error...

jannic: Ok that's a good explanation - and the files are probably missing because I forgot to git-add them after applying the upstream patch. (just git apply ...; git commit -a)

jannic: I regenerated the patch to contain the missing files. (See below.)

PR #803, opened 2016-06-17
ar71xx-generic: Add support for OpenMesh OM2P-HSv3/MR1750v2

ecsv:
The new OM2P-HSv3/MR1750v2 device support is only available in LEDE master. The relevant patches had to be backported to add support for them in Gluon.

All patches were also forwarded to OpenWrt for inclusion in CC and trunk.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -133,7 +133,7 @@ ar71xx-generic
 
 * OpenMesh
 
-  - MR1750
+  - MR1750 (v1, v2)
   - MR600 (v1, v2)
   - MR900 (v1, v2)
   - OM2P (v1, v2)
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -133,15 +133,15 @@ ar71xx-generic
 
 * OpenMesh
 
- - MR1750
- - MR600 (v1, v2)
- - MR900 (v1, v2)
- - OM2P (v1, v2)
- - OM2P-HS (v1, v2)
- - OM2P-LC
- - OM5P
- - OM5P-AC (v1, v2)
- - OM5P-AN
+  - MR1750
+  - MR600 (v1, v2)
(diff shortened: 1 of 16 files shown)


NeoRaider:
Two of the commits have the same commit message, even though they add support for different devices...

ecsv:
The commit message was different but the subject was the same. Should be fixed now

NeoRaider:
Another note: It seems you are not using the command `make update-patches` to generate the patch files (or as least I get different file names and slightly different commit messages in the patch files on my system). I'll fix this up when merging this time.

ecsv:
Ok, patches recreated with `make update-patches`

ecsv:
Thanks for pulling and thanks for all the hints

rotanid:
nice devices, although i couldn't find any information about the flash memory size of all of the open-mesh devices, data sheet doesn't say and there are no openwrt wiki pages :-(

mmalte:
The 802.11n devices have 64MB, the 802.11ac Devices 128MB

You find the specs here:

http://www.open-mesh.com/resources/downloads-and-resources.html#datasheets
[MR-2016-03.pdf](https://github.com/freifunk-gluon/gluon/files/321808/MR-2016-03.pdf)
[OM-2016-03.pdf](https://github.com/freifunk-gluon/gluon/files/321807/OM-2016-03.pdf)

Would be great to add this information to the openWrt Wiki

rotanid:
@mmalte , where can you find the flash memory size in these datasheets?

mmalte:
Sorry, I was looking for the RAM size.

I've got two devices, dmesg tells me:
OM2P-LC: m25p80 spi0.0: s25sl12800 (16384 Kbytes)
MR900v2: m25p80 spi0.0: mx25l12805d (16384 Kbytes)

PR #804, opened 2016-06-21
batman-adv: Fix speedy join in gateway client mode

ecsv:
Speedy join only works when the received packet is either broadcast or an 4addr unicast packet. Thus packets converted from broadcast to unicast via the gateway handling code have to be converted to 4addr packets to allow the receiving gateway server to add the sender address as temporary entry
to the translation table.

Not doing it will make the batman-adv gateway server drop the DHCP response in many situations because it doesn't yet have the TT entry for the destination of the DHCP response.


ecsv:
I would have liked to get merged through https://github.com/openwrt-routing/packages/pull/197 but it is unknown when this happens. I would still like to get it in gluon for 2016.2 to get speedy join (better) working with gluon

PR #808, opened 2016-06-29
add option to change/limit wifi rates

rotanid:
as first suggested by @RubenKelevra in #467 it might make sense to disable slow wifi rates to improve overall performance for clients.
credits also go to @Brother-Lal and @kb-light on whose work this PR is based.
it adds optional settings to the site.conf and suggests to set them to the values discussed in #467 .

build and flash tested on a WDR3500, rates are only set on 2,4GHz radio0 as it is expected.

### DIFF
--- a/docs/site-example/site.conf
+++ b/docs/site-example/site.conf
@@ -35,6 +35,12 @@
     -- Wireless channel.
     channel = 1,
 
+    -- Supported wifi rates, example removes 802.11b compatibility for better performance (optional, implies basic_rate)
+    -- supported_rates = '6000 9000 12000 18000 24000 36000 48000 54000',
+
+    -- Basic wifi rates, example removes 802.11b compatibility for better performance (optional)
+    -- basic_rate = '6000 9000 18000 36000 54000',
+
     -- ESSID used for client network.
     ap = {
       ssid = 'entenhausen.freifunk.net',
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -98,6 +98,12 @@ wifi24 \: optional
     This will only affect new installations.
     Upgrades will not changed the disabled state.
 
+    Additionally it is possible to configure the ``supported_rates`` and ``basic_rate``
+    of each radio. Both are optional, by default hostapd/driver dictate the rates.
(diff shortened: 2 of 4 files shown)


docs/user/site.rst
```diff
@@ -112,6 +118,8 @@ wifi24 \: optional
 
        wifi24 = {
          channel = 11,
+         supported_rates = '6000 9000 12000 18000 24000 36000 48000 54000',
+         basic_rate = '6000 9000 18000 36000 54000',
```
NeoRaider: I think making this a table of integers instead of a string would make more sense.

NeoRaider: (with table, I mean an array, as Lua doesn't really distinguish these)

package/gluon-core/files/lib/gluon/upgrade/200-wireless
```diff
@@ -27,6 +27,14 @@ local function configure_radio(radio, index, config)
...
     uci:set('wireless', radio, 'htmode', 'HT20')
     uci:set('wireless', radio, 'country', site.regdom)
+
+    if config.supported_rates then
+      uci:set('wireless', radio, 'supported_rates', config.supported_rates)
+    end
+
+    if config.basic_rate then
+      uci:set('wireless', radio, 'basic_rate', config.basic_rate)
+    end
```
NeoRaider: This should also remove these options when the config fields don't exist.

docs/site-example/site.conf
```diff
@@ -35,6 +35,12 @@
     -- Wireless channel.
     channel = 1,
 
+    -- Supported wifi rates, example removes 802.11b compatibility for better performance (optional, implies basic_rate)
+    -- supported_rates = '6000 9000 12000 18000 24000 36000 48000 54000',
+
+    -- Basic wifi rates, example removes 802.11b compatibility for better performance (optional)
+    -- basic_rate = '6000 9000 18000 36000 54000',
```
NeoRaider: I would not to comment these values out, so we recommend to use these values.

rotanid:
thanks for your feedback @NeoRaider , @kb-light did a new PR which includes your wishes.
closing this one.

PR #810, opened 2016-07-01
gluon-core: make wifi rates configurable

kb-light:
First commit adds a new function to check_site_lib.lua, to check whether a value is included in a given array. This is needed, to check that the list of wifi rates only contains unprohibited values.

Second commit adds the functionality to specify a list of allowed wifi rates within site.conf, which was already discussed in #467. Additionally this should fix all issues mentioned by @NeoRaider in #808.

Thanks to:
@RubenKelevra
@Brother-Lal
@rotanid

### DIFF
--- a/docs/site-example/site.conf
+++ b/docs/site-example/site.conf
@@ -35,6 +35,14 @@
     -- Wireless channel.
     channel = 1,
 
+    -- List of supported wifi rates (optional)
+    -- Example removes 802.11b compatibility for better performance
+    supported_rates = {6000, 9000, 12000, 18000, 24000, 36000, 48000, 54000},
+
+    -- List of basic wifi rates (optional, required if supported_rates is set)
+    -- Example removes 802.11b compatibility for better performance
+    basic_rate = {6000, 9000, 18000, 36000, 54000},
+
     -- ESSID used for client network.
     ap = {
       ssid = 'entenhausen.freifunk.net',
--- a/package/gluon-core/check_site.lua
+++ b/package/gluon-core/check_site.lua
@@ -28,6 +28,14 @@ for _, config in ipairs({'wifi24', 'wifi5'}) do
     need_string('regdom') -- regdom is only required when wifi24 or wifi5 is configured
 
     need_number(config .. '.channel')
+
+    local rates = {1000, 2000, 5500, 6000, 9000, 11000, 12000, 18000, 24000, 36000, 48000, 54000}
+    local supported_rates = need_array_of(config .. '.supported_rates', rates, false)
+    if supported_rates then
+      need_array_of(config .. '.basic_rate', supported_rates, true)
+    else
+      need_array_of(config .. '.basic_rate', rates, false)
+    end
   end
 end
 
--- a/scripts/check_site_lib.lua
+++ b/scripts/check_site_lib.lua
@@ -7,6 +7,27 @@ local function loadvar(varname)
    end
 end
 
+local function array_to_string(array)
+   local string = ''
+   for _, v in ipairs(array) do
+      if #string >= 1 then
+         string = string .. ', '
+      end
+      string = string .. v
+   end
+   return '[' .. string .. ']'
+end
+
+local function assert_one_of(var, array, msg)
+   for _, v in ipairs(array) do
+      if v == var then
+         return true
+      end
+   end
+
+   error(msg)
+end
+
 local function assert_type(var, t, msg)
    assert(type(var) == t, msg)
 end
@@ -99,7 +120,25 @@ function need_table(varname, subcheck, required)
    return var
 end
 
+function need_one_of(varname, array, required)
+   local var = loadvar(varname)
+
+   if required == false and var == nil then
+      return nil
+   end
+
+   assert_one_of(var, array, "site.conf error: expected `" .. varname .. "' to be one of given array: " .. array_to_string(array))
+
+   return var
+end
+
 function need_string_array(varname, required)
    return assert(pcall(need_array, varname, function(e) assert_type(e, 'string') end, required),
 		 "site.conf error: expected `" .. varname .. "' to be a string array")
+end
+
+function need_array_of(varname, array, required)
+   local ok, var = pcall(need_array, varname, function(e) assert_one_of(e, array) end,required)
+   assert(ok, "site.conf error: expected `" .. varname .. "' to be a subset of given array: " .. array_to_string(array))
+   return var
 end
--- a/scripts/check_site_lib.lua
+++ b/scripts/check_site_lib.lua
@@ -133,8 +133,9 @@ function need_one_of(varname, array, required)
 end
 
 function need_string_array(varname, required)
-   return assert(pcall(need_array, varname, function(e) assert_type(e, 'string') end, required),
-		 "site.conf error: expected `" .. varname .. "' to be a string array")
+   local ok, var = pcall(need_array, varname, function(e) assert_type(e, 'string') end, required)
+   assert(ok, "site.conf error: expected `" .. varname .. "' to be a string array")
+   return var
 end
 
 function need_array_of(varname, array, required)
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -98,6 +98,12 @@ wifi24 \: optional
     This will only affect new installations.
     Upgrades will not changed the disabled state.
 
+    Additionally it is possible to configure the ``supported_rates`` and ``basic_rate``
+    of each radio. Both are optional, by default hostapd/driver dictate the rates.
+    If ``supported_rates`` is set, ``basic_rate`` is required, because ``basic_rate``
+    has to be a subset of ``supported_rates``.
+    The example below disables 802.11b rates.
+
     ``ap`` requires a single parameter, a string, named ``ssid`` which sets the
     interface's ESSID.
 
@@ -112,6 +118,8 @@ wifi24 \: optional
 
        wifi24 = {
          channel = 11,
+         supported_rates = {6000, 9000, 12000, 18000, 24000, 36000, 48000, 54000},
+         basic_rate = {6000, 9000, 18000, 36000, 54000},
          ap = {
            ssid = 'entenhausen.freifunk.net',
          },
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
@@ -27,6 +27,18 @@ local function configure_radio(radio, index, config)
     uci:set('wireless', radio, 'channel', channel)
     uci:set('wireless', radio, 'htmode', 'HT20')
     uci:set('wireless', radio, 'country', site.regdom)
+
+    if config.supported_rates then
+      uci:set_list('wireless', radio, 'supported_rates', config.supported_rates)
+    else
+      uci:delete('wireless', radio, 'supported_rates')
+    end
+
+    if config.basic_rate then
+      uci:set_list('wireless', radio, 'basic_rate', config.basic_rate)
+    else
+      uci:delete('wireless', radio, 'basic_rate')
+    end
   end
 end
 
--- a/package/gluon-mesh-batman-adv-core/check_site.lua
+++ b/package/gluon-mesh-batman-adv-core/check_site.lua
@@ -1,15 +1,18 @@
 for _, config in ipairs({'wifi24', 'wifi5'}) do
+   local rates = {1000, 2000, 5500, 6000, 9000, 11000, 12000, 18000, 24000, 36000, 48000, 54000}
+   rates = need_array_of(config .. '.supported_rates', rates, false) or rates
+
    if need_table(config .. '.ibss', nil, false) then
       need_string(config .. '.ibss.ssid')
       need_string_match(config .. '.ibss.bssid', '^%x[02468aAcCeE]:%x%x:%x%x:%x%x:%x%x:%x%x$')
-      need_number(config .. '.ibss.mcast_rate', false)
+      need_one_of(config .. '.ibss.mcast_rate', rates, false)
       need_number(config .. '.ibss.vlan', false)
       need_boolean(config .. '.ibss.disabled', false)
    end
 
    if need_table(config .. '.mesh', nil, false) then
       need_string(config .. '.mesh.id')
-      need_number(config .. '.mesh.mcast_rate', false)
+      need_one_of(config .. '.mesh.mcast_rate', rates, false)
       need_boolean(config .. '.mesh.disabled', false)
    end
 end


T-X:
Just like the other pull requests, the idea looks very interesting and tempting :-). I'd be very excited to see some statistics showing how much things get improved by this (for instance reduced airtime or increased overall throughput on a couple of real-world example nodes, vs. number of clients connected to a node).

Also, what kind of devices (including operating system + wifi driver) have you tested this with? Just to make sure it does not introduce any compatibility issues.

florianjacob:
+1 for having at least the possibility to turn 802.11b off, we discussed that today at the Freifunk Karlsruhe meeting and decided to investigate how this could be achieved.

For reference, some numbers can be found in this article, Cisco recommends turning of 802.11b since 2014:

> In addition to the overhead created by the SSID’s you also have 802.11g protection mechanism that requires sending of an 802.11b packet reserving the airtime to then send the 802.11g or 802.11n packet – that’s two packets for every single user data packet – and this translates to as much as a 50% reduction of available bandwidth.
>
> See – a very high tax indeed.  Stop the madness – turn it off.  Disabling 1,2,5.5, and 11 Mbps data rates will dramatically improve 2.4 GHz availability availability for usable traffic and throughput, which can translate to a savings of 30-50%.
> http://blogs.cisco.com/wireless/wi-fi-taxes-digging-into-the-802-11b-penalty

scripts/check_site_lib.lua
```diff
@@ -103,3 +103,26 @@ function need_string_array(varname, required)
...
 end
+
+function need_var_in_array(varname, array, required)
+   local var = loadvar(varname)
+
+   if required == false and var == nil then
+      return nil
+   end
+
+   function var_in_array(var, array)
```
NeoRaider: This function should be local.

scripts/check_site_lib.lua
```diff
@@ -103,3 +103,26 @@ function need_string_array(varname, required)
    return assert(pcall(need_array, varname, function(e) assert_type(e, 'string') end, required),
 		 "site.conf error: expected `" .. varname .. "' to be a string array")
 end
+
+function need_var_in_array(varname, array, required)
```
NeoRaider: Most of this function could be replaced by a call to `need_array`, similar to the way `need_string_array` is defined. I think it would make sense to split this function into two - one that just checks if a value is one of list of possible values (maybe `need_one_of`), and one that works like `need_string_array` and calls `need_array` with `need_one_of` (could be called `need_array_of`).

docs/site-example/site.conf
```diff
@@ -35,6 +35,14 @@
     -- Wireless channel.
     channel = 1,
 
+    -- List of supported wifi rates (optional, implies basic_rate)
```
NeoRaider: "implies basic_rate" is a bit unclear. Maybe add "required if supported_rates" is set to the "basic_rate" comment?

kgbvax:
The rate given are are missing 802.11n rates and thus I understand that this limited speed to ancient 54Mbit. (I admit that this may cover most of the freifunk use cases). Anyway I think if something like that is included, this should also include the 802.11n rates / MCS indices. OpenWrt "bitrate" knows mixed notation for this 

> bitrates [legacy-<2.4|5> <legacy rate in Mbps>*] [ht-mcs-<2.4|5> <MCS index>*] [vht-mcs-<2.4|5> <NSS:MCSx,MCSy... | NSS:MCSx-MCSy>*] [sgi-2.4|lgi-2.4] [sgi-5|lgi-5]

However I am not qualified to assess whether this works for basic_rate / supported_rates as well. The documentation does not say so.

NeoRaider:
@kgbvax, I was confused by this at first, too, but it was explained to me that basic_rate / supported_rates doesn't affect 11n rates (which are so-called "extended rates").

mweinelt:
How does the availability of `basic` or `supported` rates relate to the `mcast_rate`?

RubenKelevra:
mcast_rate is just a selection, as far as I can tell, the hotspot uses still 1 MBit for this data. The ad-hoc is set to 6 MBit for mcast_rate and this is our lowest basic and supported.

scripts/check_site_lib.lua
```diff
@@ -99,7 +101,21 @@ function need_table(varname, subcheck, required)
...
+end
+
 function need_string_array(varname, required)
    return assert(pcall(need_array, varname, function(e) assert_type(e, 'string') end, required),
 		 "site.conf error: expected `" .. varname .. "' to be a string array")
 end
+
+function need_array_of(varname, array, required)
+   return assert(pcall(need_array, varname, need_one_of(e, array),required),
+                 "site.conf error: expacted `" .. varname .. "' to be a subset of given array")
```
rotanid: typo, should be "expected" instead of "expacted"

docs/site-example/site.conf
```diff
@@ -35,6 +35,14 @@
     -- Wireless channel.
     channel = 1,
 
+    -- List of supported wifi rates (optional)
+    -- Example removes 802.11b compatibility for better performance
+    supported_rates = {6000, 9000, 12000, 18000, 24000, 36000, 48000, 54000},
+
+    -- List of basic wifi rates (optional, recommend if supported_rates is set)
```
mweinelt: `recommend` should be `recommended`

scripts/check_site_lib.lua
```diff
@@ -74,8 +74,10 @@ function need_array(varname, subcheck, required)
 
    assert_type(var, 'table', "site.conf error: expected `" .. varname .. "' to be an array")
 
-   for _, e in ipairs(var) do
-      subcheck(e)
+   if subcheck then
+      for _, e in ipairs(var) do
+         subcheck(e)
+      end
```
NeoRaider: This diff chunk seems unneeded? All array checks should have a subcheck.

scripts/check_site_lib.lua
```diff
@@ -99,7 +101,21 @@ function need_table(varname, subcheck, required)
...
+   return false
+end
+
 function need_string_array(varname, required)
    return assert(pcall(need_array, varname, function(e) assert_type(e, 'string') end, required),
 		 "site.conf error: expected `" .. varname .. "' to be a string array")
 end
+
+function need_array_of(varname, array, required)
+   return assert(pcall(need_array, varname, need_one_of(e, array),required),
```
NeoRaider: This call is wrong - you need to wrap the `need_one_of` into an anonymous function like it is done in `need_string_array`

scripts/check_site_lib.lua
kb-light: same as need_table()

kb-light: @NeoRaider: I currently left this check, because I use need_array() to get the array of supported_rates to compare against mcast_rate and basic_rate. Maybe there is another solution to solve this.

rotanid:
_just adding a note, so @neoraider knows something was done here and his label "changes requested" is obsolete_

NeoRaider:
Thanks, already noticed, I'm a bit busy this week though...

viisauksena:
just a sidenode, 
this only maintain 80211 **a** , blocking **b** and not touching **n** at all ... 
this may be fine,  maybe we want to add beacon_int (default 100 - integer between 15 and 65k) also
i did some testing, the result is that there is still a bunch of 1Mbit Beacons .. every 100ms 
.. see here https://forum.freifunk.net/t/client-basic-rate-setzen-um-1mbit-verbindungen-zu-blocken/13286

NeoRaider:
1. Beacons don't really matter, they are small, and one every 100ms shouldn't have any noticable effect on the airtime anyways.
2. Beacons are sent using the multicast rate, which is configured independently of the basic/supported rates.

RubenKelevra:
@viisauksena first of all the "usage of channel airtime" stats of horst cannot be trusted:

https://github.com/br101/horst/issues/45

Secondly if you used the following:
> uci set wireless.client_radio0.basic_rate="5500 6000 9000 11000 12000 18000 20000 24000 30000 36000 40000 42000 44000 48000 50000 51000"

... it won't work because it's just wrong. There are still some B-Rates and there are some non-existing ones.

I guess it will just ignored by the driver.

RubenKelevra:
@viisauksena in the End, your conclusion is wrong, becons are sent with 6 mbits after this patch:

horst_wifi_beacon_ff

viisauksena:
@NeoRaider based on my observation youre right for beacon rate in **ibss0** - this seems based on mcast rate, but against @RubenKelevra mentioning i observe 1 Mbit beacons for **client0 AP** half the time ... see esp. here 
https://forum.freifunk.net/uploads/default/original/2X/2/2c7aa7d804d458666bfd6c3a160d1f70aafe1804.png
bildschirmfoto von 2016-08-14 03-38-43

i should also have mentioned that i use ure supported list rates and not my blind guessings in the beginning.
```
# this is the used setup on all cpe and 841 Routers 
uci set wireless.ibss_radio0.mcast_rate='24000'
uci set wireless.client_radio0.supported_rate='6000 9000 12000 18000 24000 36000 48000 54000'
uci set wireless.client_radio0.basic_rate='6000 9000 18000 36000 48000 54000'
uci commit 
wifi
```

RubenKelevra:
@viisauksena

mcast_rate 24000 is insane. 12000 like used in luebeck is somewhat high, I guess. I would recommend 6 mbit for this.

Else, your setup is broken, since this works as I've said:

horst_wifi_beacon_ff2

viisauksena:
@RubenKelevra this is somehow offtopic , and should be discussed somewhere else - i guess,
but to answer your questions .. see this **insane** setup here http://openfreiburg.de/freifunk/meshviewer/#!v:m;n:60e327df06b0 there my mentioned setup is running as described. there a no cables involved, on the left there are three different uplinks on cpe routers - all links seem pretty stable - the build is 2016.1.5 ... more detailed Linux version 3.18.29 (fffr@build) (gcc version 4.8.3 (OpenWrt/Linaro GCC 4.8-2014.04 r49261) ) #5 Mon Jun 27 01:57:57 CEST 2016 ... so this is build on that day with latest/master

rotanid:
could you please take this discussion elsewhere?
you are not contributing to this pull request but rather making it confusing to find the relevant posts here.
feel free to discuss this in a new issue, on IRC or on the mailing list :)

RubenKelevra:
@rotanid yep. Just didn't want to let false information without reply beeing written here, delaying the patch merge further. :)

kb-light:
just trying to close this discussion:
@viisauksena: your settings are wrong and should have no effect. Instead of `wireless.client_radio0.supported_rate` you should set `wireless.radio0.supported_rates` and `wireless.radio0.basic_rate` instead of `wireless.client_radio0.basic_rate`. (It is `radio0`, not `client_radio0` and `supported_rates`, not `supported_rate`)

NeoRaider:
I'd expect `basic_rate` and `supported_rates` to work for `client_radio0` as well (I haven't checked though).

If it does work for the client interface, we should set those values there and not for the whole device.

RubenKelevra:
@neoraider it should be applied for the whole device, we try to remove the b-behavior and we don't gain anything if the mesh still remains on a b mixed mode.

Else I think it works just in the device-level, but I'm not sure though.

viisauksena:
@kb-light : thx for error finding, @NeoRaider it seems that client0 as i've done does not work, the correction by @kb-light work like a charme, so thx all

scripts/check_site_lib.lua
```diff
@@ -7,6 +7,28 @@ local function loadvar(varname)
...
+      string = string .. v
+   end
+   return '[' .. string .. ']'
+end
+
+local function assert_one_of(var, array, msg)
+   local pass = false
+   for _, v in ipairs(array) do
+      if v == var then
+         pass = true
```
NeoRaider: You can just `return` here, then you don't need the `pass` variable. `assert(pass, msg)` could be replaced by `error(msg)` then.

package/gluon-core/check_site.lua
```diff
@@ -28,6 +28,15 @@ for _, config in ipairs({'wifi24', 'wifi5'}) do
     need_string('regdom') -- regdom is only required when wifi24 or wifi5 is configured
 
     need_number(config .. '.channel')
+
+    local rates={1000, 2000, 5500, 6000, 9000, 11000, 12000, 18000, 24000, 36000, 48000, 54000}
+    local supported_rates =  need_array(config .. '.supported_rates', nil, false)
```
NeoRaider: There is some inconsistent whitespace here, `rates` doesn't have any spaces around `=`, 'supported_rates` has one to many.

package/gluon-core/check_site.lua
```diff
@@ -28,6 +28,15 @@ for _, config in ipairs({'wifi24', 'wifi5'}) do
...
+    if supported_rates then
+      need_array_of(config .. '.supported_rates', rates, true)
```
NeoRaider: It doesn't make sense to check supported_rates twice. Just do `local supported_rates = need_array_of(config .. '.supported_rates', rates, false)` above, and only check basic_rate in the if/else branches.

NeoRaider:
Except for the minor issues I pointed out, this is looking fine now.

scripts/check_site_lib.lua
```diff
@@ -133,8 +133,9 @@ function need_one_of(varname, array, required)
 end
 
 function need_string_array(varname, required)
-   return assert(pcall(need_array, varname, function(e) assert_type(e, 'string') end, required),
-		 "site.conf error: expected `" .. varname .. "' to be a string array")
+   assert(pcall(need_array, varname, function(e) assert_type(e, 'string') end, required),
```
NeoRaider: Hmm, the `pcall` should return two values, a true/false value and the original return value of `need_array`, we're just discarding the second one by passing it to assert. I think we could do without the extra `loadvar` like this (untested):

```
local ok, var = pcall(need_array, varname, function(e) assert_type(e, 'string') end, required)
assert(ok, "site.conf error: expected `" .. varname .. "' to be a string array")
return var
```

kb-light: that works

docs/site-example/site.conf
```diff
@@ -35,6 +35,14 @@
     -- Wireless channel.
     channel = 1,
 
+    -- List of supported wifi rates (optional)
+    -- Example removes 802.11b compatibility for better performance
+    supported_rates = {6000, 9000, 12000, 18000, 24000, 36000, 48000, 54000},
+
+    -- List of basic wifi rates (optional, recommended if supported_rates is set)
```
NeoRaider: This comment doesn't match the check_site script: the script makes `basic_rate` mandatory if `supported_rates` is set, not just recommended.

kb-light: That should be required. I updated `docs/user/site.rst`, too, to sound the same.

package/gluon-core/check_site.lua
```diff
@@ -28,6 +28,14 @@ for _, config in ipairs({'wifi24', 'wifi5'}) do
     need_string('regdom') -- regdom is only required when wifi24 or wifi5 is configured
 
     need_number(config .. '.channel')
+
+    local rates = {1000, 2000, 5500, 6000, 9000, 11000, 12000, 18000, 24000, 36000, 48000, 54000}
+    local supported_rates = need_array_of(config .. '.supported_rates', rates, false)
+    if supported_rates then
```
kb-light: @NeoRaider: I don't like this `if`, but I think it is more clear than:
`need_array_of(config .. '.basic_rate', supported_rates or rates, supported_rates and true or false)`

rubo77:
for others searching for this, I add the examples from http://gluon.readthedocs.io/en/v2016.2/user/site.html?highlight=basic_rate

```
    -- Example removes 802.11b compatibility for better performance
    supported_rates = {6000, 9000, 12000, 18000, 24000, 36000, 48000, 54000},

    -- List of basic wifi rates (optional, required if supported_rates is set)
    -- Example removes 802.11b compatibility for better performance
    basic_rate = {6000, 9000, 18000, 36000, 54000},

```

the question is how to configure this if you DONT want to disable 802.11b?

mweinelt:
I'd assume adding `5500` (and optionally `1000` and `2000`) would do that.

viisauksena:
or dont configure this optional feature at all

rubo77:
And does this interact with the `mcast_rate` setting? and how?

kb-light:
@rubo77: mcast_rate should be part of supported_rates (if supported_rates is set)

rubo77:
As far as I understand, if you set `mcast_rate` to for example 12000, then **no** links slower or faster than 12MBit are created, so you are not able to send faster than 12MBit either.

I would like to configure it, so our mesh sends minimum 12000 but also faster, if the link is good.

Is that possible with those three options `supported_rates`, `basic_rate` and `mcast_rate`?

see also: https://forum.freifunk.net/t/moegliche-werte-fuer-mcast-rate/11785/13

NeoRaider:
`mcast_rate` is used for multicast packets only. The multicast rate is always fixed and defaults to 1MBit/s. We recommend setting it to a higher value like 12MBit/s to reduce the airtime used by the background noise. This setting affects the mesh links only, not the clients links (as it is unknown if all clients work with multicast rates other than 1MBit/s).

The unicast rate is completely independent of this and is negotiated for each peer and client independently. It may be higher or lower than the multicast rate.

`supported_rates` and `basic_rate` control the allowed unicast rates for both mesh and client links. They are mainly useful to prevent the use of legacy 802.11b rates, which may impair the overall WLAN performance. Also, these settings don't affect the "extended rates" (802.11n, everything higher than 54MBit/s), these are always allowed additionally.

rubo77:
This should be added in the docs here: http://gluon.readthedocs.io/en/latest/user/site.html

rubo77:
This should be added in the docs here: http://gluon.readthedocs.io/en/latest/user/site.html

viisauksena:
it is not "limited" - it is set fixed

rubo77:
Why is this set fixed? Isn't there a way to get the maximum speed possible of each mesh connection?

RubenKelevra:
@rubo77 it's fixed because this is how wifi works, so no.

rubo77:
It is not set fixed: 

>The mcast_rate is only used for multicast traffic, which is management traffic (WLAN beacons, batman-adv OGMs) and multicast/broadcast payloads (ARP, ICMPv6, ...). It is fixed as multicast packets are not directed to any specific recipient, but sent only once for all peers, so there is no negotiation.
>
Unicast packets are sent only to a single peer and are thus sent with higher or lower bitrate, which is negotiated with the peer individually

https://github.com/freifunk-gluon/gluon/commit/ce53ed209b6d78aabe6a90d4689734618db68d62

RubenKelevra:
The setting is called mcast rate, why should unicast be fixed than? Your concerns makes absolutely no sense.

rubo77:
Look in the forum. Many Admins think this would Set Both fixed

T-X:
Because of the lack of statistics I'm still a little sceptical regarding the general applicability of this feature. For instance while in your 300 nodes mesh there might be ten nodes with too many clients, where this feature is needed to make the node usable at least for the clients in closer range. But there might also be ten other nodes with less clients where clients a little more far away are kicked even though they do no harm.

However, why I'm writing now is actually this patch which I think could put this feature on steroids, if someone were adding another patch to have it available not only for AP, but also IBSS/802.11s interfaces: https://patchwork.ozlabs.org/patch/697637/

For instance we could have a required basic rate of maybe 6 or 12MBit/s for all interfaces. And then safely disable wifi broadcasts all together with this mac80211 multicast-to-unicast patch.

rubo77:
Couldn't we make the configuration depending on the nr of connected clients somehow?

PR #814, opened 2016-07-03
use http://www.candelatech.com/downloads/firmware-2-ct-full-community-16.bin

### LINKED ISSUE #690: Extensive Memory Leaks on Archer C5/C7 devices
Archer C5/C7 devices (ath10k based) there are extensive memory leaks leading to a reboot every 0,5-1,5 days.

This happens with gluon versions at least since 2016.1, and it can be seen on the current master branch with firmwares built as recent as 2016/03/13.

@andir has looked at the problem with kmemleak, since the problem does not seem to be in the userspace. Those logs however looked inconclusive afair.

mmalte:
This patch by @jannic fixes #690 by using a stable Version of the Candeltec Firmware:

> This mail by Ben Greear (candelatech) suggests that the 10.2 firmware is known to be worse than the 10.1 one (firmware-2): http://lists.infradead.org/pipermail/ath10k/2016-June/007743.html
> So perhaps going 'back' to firmware-2 is the way to go, independently of the question if it solves the Archer issue?

It is running on 87 Archer nodes in Aachen with 2016.1.5, only 6 with an uptime less than one day. Most of them without a reboot since the Upgrade, see:
https://map.aachen.freifunk.net/nodelist/

### DIFF
--- a/patches/openwrt/0029-use-http-www.candelatech.com-downloads-firmware-2-ct-full-community-16.bin.patch
+++ b/patches/openwrt/0029-use-http-www.candelatech.com-downloads-firmware-2-ct-full-community-16.bin.patch
@@ -1,0 +1,36 @@
+From: Jan Niehusmann <jan@gondor.com>
+Date: Wed, 15 Jun 2016 18:01:53 +0200
+Subject: use http://www.candelatech.com/downloads/firmware-2-ct-full-community-16.bin
+
+diff --git a/package/firmware/ath10k-firmware/Makefile b/package/firmware/ath10k-firmware/Makefile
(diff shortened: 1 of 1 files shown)


NeoRaider:
fd237f6f43d4ad36e1987bf7461a16647287db18 updates the firmware to "firmware-2-ct-full-community-16.bin-lede", please test if this also fixes your issues.

mweinelt:
Looks good! https://www.linuxlounge.net/~martin/tmp/2016-07-08-012545_918x524_scrot.png

NeoRaider:
Okay, great :)

PR #824, opened 2016-07-16
add patch for a additional download location for busybox

RubenKelevra:
Currently all download-locations are outdated or down, I've added another one to get gluon successfully build


NeoRaider:
busybox.net is back, so I'll close this PR.

For the future, please always open PRs against the Gluon master, I'll backport appropriate patches to the stable branch then. A patch against current LEDE would be even better, so we don't have to carry such patches in Gluon forever.

PR #825, opened 2016-07-17
add missing apostrophe

RubenKelevra:


NeoRaider:
No, "its" is correct. "it's" is a reduction of "it is", while "its" is the possessive pronoun for "it".

Also, wrong branch again, all patches must go into master first.

NeoRaider:
Further issues of this PR would be:

* Missing package name prefix in the commit message
* The patch changes the translation string without adjusting the PO files.

PR #826, opened 2016-07-18
Add package gluon-airtime

corny:
This package adds information about the channel usage to respondd:
- `frequency`: the current frequency
- `active`: time on channel
- `rx`: time spent on receiving
- `tx`: time spent on transmitting
- `busy`: time spent on waiting for transmitting

It works fine ath9k, ath10k does not return the information.

We used this package at our Breminale event Bremen on ~80 nodes (> 2k concurrent clients) to calculate the channel utilization with our [respond-collector](https://github.com/FreifunkBremen/respond-collector/commit/92aac7b7ca55da66e212ce5f71557f6d5d0b2c49#diff-76df96424dfe13846cb6cdcb2a5c7ef5R35).

One strange observation: In some intervals it happens that `Δ(busy) + Δ(rx) + Δ(tx) > Δ(active)`.

### DIFF
--- a/package/gluon-airtime/src/airtime.h
+++ b/package/gluon-airtime/src/airtime.h
@@ -11,9 +11,4 @@ struct airtime_result {
 	uint8_t  noise;
 };
 
-struct airtime {
-	struct airtime_result radio0;
-	struct airtime_result radio1;
-};
-
-struct airtime* get_airtime(const char *radio0, const char *radio1);
+int get_airtime(struct airtime_result *result, int ifx);
--- a/package/gluon-airtime/src/airtime.h
+++ b/package/gluon-airtime/src/airtime.h
@@ -2,18 +2,13 @@
 
 #include <stdint.h>
 
-struct airtime_time {
-	uint64_t current;
-	uint64_t offset;
-};
-
 struct airtime_result {
+	uint64_t active_time;
+	uint64_t busy_time;
+	uint64_t rx_time;
+	uint64_t tx_time;
 	uint32_t frequency;
 	uint8_t  noise;
-	struct airtime_time active_time;
-	struct airtime_time busy_time;
-	struct airtime_time rx_time;
-	struct airtime_time tx_time;
 };
 
 struct airtime {
--- a/package/gluon-airtime/src/airtime.h
+++ b/package/gluon-airtime/src/airtime.h
@@ -1,0 +1,24 @@
+#pragma once
+
+#include <stdint.h>
+
+struct airtime_time {
+	uint64_t current;
+	uint64_t offset;
+};
+
+struct airtime_result {
+	uint32_t frequency;
+	uint8_t  noise;
+	struct airtime_time active_time;
+	struct airtime_time busy_time;
+	struct airtime_time rx_time;
+	struct airtime_time tx_time;
+};
+
+struct airtime {
+	struct airtime_result radio0;
+	struct airtime_result radio1;
+};
+
+struct airtime* get_airtime(const char *radio0, const char *radio1);
--- a/package/gluon-airtime/Makefile
+++ b/package/gluon-airtime/Makefile
@@ -1,0 +1,40 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-airtime
+PKG_VERSION:=1
+PKG_RELEASE:=1
+
+PKG_BUILD_DIR := $(BUILD_DIR)/$(PKG_NAME)
+PKG_BUILD_DEPENDS := respondd
+
+include $(GLUONDIR)/include/package.mk
+
+define Package/gluon-airtime
+  SECTION:=gluon
+  CATEGORY:=Gluon
+  TITLE:=Add airtime to respondd
+  DEPENDS:=+respondd +libnl-tiny
+endef
+
+define Build/Prepare
+	mkdir -p $(PKG_BUILD_DIR)
+	$(CP) ./src/* $(PKG_BUILD_DIR)/
+endef
+
+define Build/Configure
+endef
+
+
+TARGET_CFLAGS += -I$(STAGING_DIR)/usr/include/libnl-tiny
+
+define Build/Compile
+	CFLAGS="$(TARGET_CFLAGS)" CPPFLAGS="$(TARGET_CPPFLAGS)" $(MAKE) -C $(PKG_BUILD_DIR) $(TARGET_CONFIGURE_OPTS)
+endef
+
+
+define Package/gluon-airtime/install
+	$(INSTALL_DIR) $(1)/lib/gluon/respondd
+	$(CP) $(PKG_BUILD_DIR)/respondd.so $(1)/lib/gluon/respondd/airtime.so
+endef
+
+$(eval $(call BuildPackage,gluon-airtime))
--- a/package/gluon-airtime/README.md
+++ b/package/gluon-airtime/README.md
@@ -1,0 +1,36 @@
+This module adds a respondd airtime usage statistics provider.
+The format is the following:
+
+```json
+{
+  "statistics": {
+    "wireless": [
+      {
+        "frequency": 5220,
+        "active": 366561161,
+        "busy": 46496566,
+        "rx": 808415,
+        "tx": 41711344,
+        "noise": 162
+      },
+      {
+        "frequency": 2437,
+        "active": 366649704,
+        "busy": 205221222,
+        "rx": 108121446,
+        "tx": 85453679,
+        "noise": 161
+      }
+    ]
+  }
+}
+```
+
+The numbers `active`, `busy`, `rx` and `tx` are times in milliseconds, where
+`busy`, `rx` and `tx` have to be interpreted by taking the quotient with
+`active`.
+
+The motivation for having a list with the frequency as a value in the objects
+instead of having an object with the frequency as keys is that multiple wifi
+devices might be present, in which case the same frequency can appear multiple
+times (because the statistics are reported once for every phy).
--- a/package/gluon-airtime/src/Makefile
+++ b/package/gluon-airtime/src/Makefile
@@ -11,12 +11,14 @@ CFLAGS         += -pedantic
 
 all: respondd.so
 
-# sudo apt install libnl-3-dev
-airtime-test:	airtime-test.c airtime.c airtime.h
-	$(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) -fPIC -D_GNU_SOURCE -lnl-tiny -o $@ airtime.c airtime-test.c $(LDLIBS)
+%.c: %.h
 
-respondd.so: respondd.c airtime.c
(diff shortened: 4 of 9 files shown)


corny:
It might be more useful to calculate the channel usage directly on the node. This should be done in a seperate thread to avoid side effects by requesting the statistics. Otherwise the respond-client has to know the previous values to do the calculation.

T-X:
That sounds great! Have you thought about adding that information to a node's webpage, too? Having a live graph for airtime similar to the RSSI graphs would be awesome and super helpful :-).

jplitza:
This might be more appropriately put into the gluon-packages repository, as there is not really any special ties to gluon (only to respondd, which is also over there).

corny:
New pull request: https://github.com/freifunk-gluon/packages/pull/141

corny:
@T-X That would be nice. I don't think that I have enough time to implement it. Maybe someone else?

NeoRaider:
But the package *is* Gluon-specific! Even when it's only the naming of the interfaces, moving this package to the OpenWrt/LEDE upstream repos doesn't make sense (and the packages repo is for packages that are support to go there at some point).

So the Gluon main repo is the correct place, not the packages repo.

corny:
ok, everything back  :grimacing:

jplitza:
@NeoRaider To me that means that the package has to autodetect the interface names before it can be moved upstream, but its functionality is generic enough that I would see "moving upstream" as its goal, even though things have to be done first. But if you say so…

corny:
I could also use uci to iterate over the wifi-ifaces and find a wifi-iface with 2.4 GHz and another with 5 GHz. Would that help?

NeoRaider:
Using uci is not necessary IMO. Another option to make this independent of Gluon would be to iterate over all radios and add airtime stats for each of them. Instead of "airtime24" and "airtime5", you could have a field "airtime", which is an (unordered) array of stats for all radios. The "frequency" field can then be used to identify the radios.

If you make this independent of Gluon, then indeed this can be added to the *packages* repo. In that case, the package name should not contain Gluon (maybe just call it "respondd-airtime"?).

corny:
I try to get the airtime by radio name. Currently we do:
```c
ifx = if_nametoindex(interface);
NLA_PUT_U32(msg, NL80211_ATTR_IFINDEX, ifx);
```

NeoRaider:
Will this return the cumulative airtime of all interfaces of a single radio, or just the single VIF?

corny:
The return value for all VIFs of a radio are equal. The values come directly from the radio.

jplitza:
I agree that iterating over all radios is preferable to using uci for anything. And while all VIFs for one radio give you the same numbers (which raises the question why this is an `iw dev` command instead of an `iw phy` one), one might have multiple radios for the same frequency band, in which case the frequency isn't a unique indicator for the values. @NeoRaider, you said "unordered array", so you are proposing a JSON structure like this
```json
"airtime": [
  { "frequency": 5180, "active": ...},
  { "frequency": 5200, "active": ...},
  ...
  { "frequency": 5180, "active": ...},
  { "frequency": 5200, "active": ...},
  ...
]
```
where one frequency can show up multiple times (for different radios)?

NeoRaider:
@jplitza, yes, maybe like this.

The current code looks like it only adds a single entry per radio, but the netlink interface should return a full survey of all channels, this confuses me a bit. If we want a full survey, we might also group the entries by radio, and use the frequencies as object keys:
```json
"airtime": [
  {
    "5180": { "active": ...},
    "5200": { "active": ...}
  },
  ...
  {
    "5180": { "active": ...},
    "5200": { "active": ...}
  },
  ...
]
```
or something like this. But I might also misinterpret the code, I didn't have a closer like at it yet.

rotanid:
it might be nice to have it independent from gluon, but the author already said he won't do that work.
so it would be nice to have it in gluon at least, maybe @NeoRaider could take a look on the code from this point of view if no one wants to volunteer to "upstream" this and not only discuss about it....

RubenKelevra:
> One strange observation: In some intervals it happens that Δ(busy) + Δ(rx) + Δ(tx) > Δ(active)

We discovered that too, some years ago. I think it's happening because the times are rounded to a given value, so it might be "0,89" but 1 is recorded. This sums up to high value for each step. But active seems to be recoreded otherwise, like "we're sending now 103 packets at once, counting them all together as one "step".

corny:
@RubenKelevra Thanks a lot. That helps us to understand the data.

RubenKelevra:
@corny that was just my best guess without knowing any details. So prove me wrong if you found anything better.

Else the chip does output the noise-level of all channels it saws. That might be not especially useful with low dataset-times but on the channel which is in use, this might be helpful to discover a wifi which is some channels off or a different type of signal which still reduces the signal noise ratio significantly.

We just screenread all this values pretty dirty from iw which is not especially a pretty solution. ;)

jplitza:
@NeoRaider I just noticed you asked a question in your last comment. Yes, the package only reports airtime numbers for a single frequency, namely the one currently in use. YMMD for different drivers, but ath9k only returns significant numbers for the current frequency:
```
# iw dev client0 survey dump
Survey data from client0
	frequency:			2412 MHz
	channel active time:		26 ms
	channel busy time:		21 ms
	channel receive time:		16 ms
	channel transmit time:		0 ms
Survey data from client0
	frequency:			2417 MHz
Survey data from client0
	frequency:			2422 MHz
[...]
Survey data from client0
	frequency:			2437 MHz [in use]
	noise:				-95 dBm
	channel active time:		1704490481 ms
	channel busy time:		945335227 ms
	channel receive time:		584372518 ms
	channel transmit time:		310956458 ms
Survey data from client0
	frequency:			2442 MHz
Survey data from client0
	frequency:			2447 MHz
[...]
```
Channel 1 was probably used during initialization of the driver/device, and channel 6 is used in our network. No other channel has any numbers.

I still plan on making this iterate over all radios. But in case that takes too long, I just pushed a version that uses the format in my previous comment (not the one in @NeoRaider's previous comment, for the above reasons: We only report one frequency per device per band). Although, to be honest, I don't consider this mergeable as long as the device names are hardcoded, because client interfaces can be (and often are) turned off and thus inexistent.

RubenKelevra:
@jplitza you've holding your wifi wrong. Only 1, 5, 9 and 13 should be used in Germany.

package/gluon-airtime/src/airtime.h
```diff
@@ -0,0 +1,24 @@
...
+#include <stdint.h>
+
+struct airtime_time {
+  uint64_t current;
+  uint64_t offset;
+};
+
+struct airtime_result {
+  uint32_t frequency;
+	uint8_t  noise;
```
rotanid: tab indention here, 2 spaces everywhere else?

package/gluon-airtime/src/airtime.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+#include <stdint.h>
+
+struct airtime_time {
```
rotanid: indention of this struct is two spaces, but everywhere else its tabs?

package/gluon-airtime/src/airtime.h
```diff
@@ -0,0 +1,24 @@
...
+  uint64_t current;
```
rotanid: indention of this struct is two spaces, but everywhere else its tabs?

package/gluon-airtime/src/airtime.h
```diff
@@ -0,0 +1,24 @@
...
+  uint64_t offset;
+};
+
+struct airtime_result {
+  uint32_t frequency;
```
rotanid: indention of this struct is two spaces, but everywhere else its tabs?

rotanid:
fyi: tried this package using a WDR3500 a few days ago and it didn't work, neither respondd nor the included demo did output the information. did some quick/short debugging with @corny but no result yet.

viisauksena:
we have something similar in freiburg - using respondd and a shell airtime script which also calculate averages 
i summarize shortly
package -> https://github.com/viisauksena/gluon-airtime
only works for 2G - is heavily redundant (tx,rx,avg..) shows uplink-interface and channel and quality (batman ttvn)

airtime example

corny:
I would like to test the package on multiple devices. Unfortunately I forgot how to build a single package. Does anyone know how to build it if I start with a gluon site repo? Maybe @NeoRaider ?

corny:
Thanks to @Brother-Lal I was able to fix the issues. I tested it succesfully with:
* WR841 v10
* WDR3600
* CPE210

jplitza:
@corny so what issues were there to fix? The only issue remaining, in my opinion, is the hardcoded interface problem, which still persists. And by force pushing, you overwrote my changes. Please stop that.

corny:
The checks in the `survey_airtime_handler()` lead to abortion of the function.

jplitza:
I think this is finished. The interfaces are now autodetected by asking the kernel for a list of wifi interfaces (VIFs, not PHYs) and picking (the first) one for each PHY.

The only question that remains: Should this instead go to the gluon-packages repo, because it doesn't depend on Gluon now in any way?

rotanid:
as already controversially discussed on IRC, we would prefer to also include some average over the last n minutes - this could be used on the status page or in external software like hopglass or meshviewer without having to do much coding in every one of these three examples.
we already implemented that a while ago: https://github.com/Brother-Lal/gluon/commits/airtime

corny:
My opinion: Keep out complexity from the C code and calculate the values you need in your map backend.

NeoRaider:
I also think that the respondd provider should always return a momentary value and not average itself.

This should indeed go into the packages repo (might also make sense to rename it in a way that doesn't contain "gluon", but that doesn't really matter).

rotanid:
i don't think the momentary value and the average contradict each other - our version simply returns both.
well, whatever...

Adorfer:
Everybody can pick what is needed. 
If you do not want to poll every minute, an average could come handy.

jplitza:
I fixed a memory leak that caused havoc in my long running respondd (valgrind doesn't report anything now for the test program) and added a readme detailing the format.

As respondd currently searches its providers in `/lib/gluon/respondd/`, moving the package to the packages repository isn't possible yet. A bit OT: As freifunk-gluon/packages#143 won't be mergable in the near future and thus neither #840, I will prepare a PR for gluon-respondd that moves the provider modules (as proposed in #840) and converts the initscript to procd (to have automatic restarts).

lemoer:
Has anyone tested how stable the results are? Maybe will get the same oscillating issues as in #508 (which are still not fixed yet).

corny:
We monitored these values at a public event with 80 nodes and they are fine.
Values are monotonically increasing unless the channel changes.

genofire:
Hey, here are some nostalgie (collected respondd answers with our [respond-collector](https://github.com/FreifunkBremen/respond-collector) of this event (take a look of the ninth diagram)
screencapture-mgmt-ffhb-de-8080-dashboard-db-breminale-1468785816780

jplitza:
Opened pull request for Gluon-independent package, closing this one.

PR #827, opened 2016-07-18
add patch to enable CONFIG_SATA_VIA on x86-generic

flokli:


flokli:
Also sent to lede-dev ML.

https://patchwork.ozlabs.org/patch/649640/

NeoRaider:
Applied with minor changes (you should use `make update-patches` instead of manually creating patch files).

PR #828, opened 2016-07-25
Added support for ALFA Networks Tube2H

Sprinterfreak:
The Tube2H is a cheap Bullet-Clone.
It is supported barely undocumented by openwrt.
So it's a good point to add Gluon support to it.

I am testing it on the small 8M version and runs quiet well and stable.
The only ToDo is enhance the implementation of the rf chip. It's actually not possible to set txpower.
MultiSSID is working and default txpower seems to be normal compared to the possible range.

### DIFF
--- a/targets/ar71xx-generic/profiles.mk
+++ b/targets/ar71xx-generic/profiles.mk
@@ -355,6 +355,13 @@ $(eval $(call GluonModel,HORNETUB,hornet-ub,alfa-hornet-ub))
 $(eval $(call GluonModelAlias,HORNETUB,alfa-hornet-ub,alfa-ap121))
 $(eval $(call GluonModelAlias,HORNETUB,alfa-hornet-ub,alfa-ap121u))
 
+ifneq ($(BROKEN),)
+# Tube2H
+$(eval $(call GluonProfile,TUBE2H))
+$(eval $(call GluonModel,TUBE2H8M,tube2h-8m,alfa-tube2h-8m)) # BROKEN: working, only generic bgn
+$(eval $(call GluonModel,TUBE2H16M,tube2h-16m,alfa-tube2h-16m)) # BROKEN: untested
+endif
+
 ## Meraki
 
 # Meraki MR12/MR62


pbriesch:
I have compiled the images for gluon v2016.2.2

The file gluon-xxx-alfa-network-tube2h.bin turns out to be **3.5** MB in size.
However, the file gluon-xxx-alfa-network-tube2h-**sysupgrade**.bin turns out to be **7.2** MB in size.

Unfortunately, I didn't notice this until after I put the sysupgrade via ssh on the router which seems now to be bricked.

Please advice! Thanks

Sprinterfreak:
Strange. My Tube worked just fine on ffrgb even with autoupdate enabled (sysupgrade).
Ok, the tube2h code was probably never fully developed analog to the cheap bodgy hardware itself. But it ran in my case more stable than wr841... Any compile errors? Or where there recent changes in ImageBuilder or so?
Time for serial. Ive never connected it on it so far. It should be available on the board.

pbriesch:
I learned that there are two different hardware versions:

1.   8 GB flash, 32 GB ram
2. 16 GB flash, 64 GB ram

But I don't know how to distinguish these from just looking at the part.

@Sprinterfreak could you please give me some hints how I could debrick the Tube?
I already found a link with a picture of the board with the description of the JTAG and serial pins.

Is there a way to set the part in recovery mode and then flash the firmware with tftp? Any hints?

NeoRaider:
AFAIK, the 16MB version isn't official, but a hardware mod. We only support the 8MB version.

You can probably get into the bootloader using a 3.3V serial adapter and find out what is wrong.

pbriesch:
OK, then I need to find out how to open the Tube.
How come that the files size of the image for sysupgrade is more than double as big than the one for manufacturing?

heini66:
unscrew the big silber round screw on the antenna plug and slide it out of
the plastic

2016-12-22 21:12 GMT+01:00 Paul <notifications@github.com>:

NeoRaider:
@pbriesch The sysupgrade image is just a linear flash image. The TUBE uses an unusual flash layout that starts with the rootfs, then has a lot of padding (where OpenWrt creates the overlay), and the kernel lies at the end at a fixed address. The factory image is a compressed archive containing the kernel and the rootfs separately. The padding makes the sysupgrade image much bigger.

Sprinterfreak:
Sorry, I had never bricked a Tube so I have no experience debricking them. I would suggest that the bootloader is pretty much the same like UBNTs uboot so I would first try to connect the serial like @NeoRaider mentioned while I stumbled my post together...
I think there should be a boot interrupt option to put the device into tftp mode.
Possibly this could be done by any reset button press combination too. Who knows...
I have found no official documentation to the hardware. I mentioned this in the description too.

Opening is pretty simple like heini66 wrote. But I wouldn't really trust that this device is waterproof at all.
Unscrew the Antenna, then the nut from the N-Plug. Then unscrew the bottom cup and You're in.

pbriesch:
OK, thank you all for you help. This is as far as I am now:
```
U-Boot 1.1.5-g0ad4dedd-dirty (May 18 2012 - 14:18:30)

AP121-8MB (ar9331) U-boot
DRAM:  32 MB
Top of RAM usable for U-Boot at: 82000000
Reserving 248k for U-Boot at: 81fc0000
Reserving 192k for malloc() at: 81f90000
Reserving 44 Bytes for Board Info at: 81f8ffd4
Reserving 36 Bytes for Global Data at: 81f8ffb0
Reserving 128k for boot params() at: 81f6ffb0
Stack Pointer at: 81f6ff98
Now running in RAM - U-Boot at: 81fc0000
id read 0x100000ff
flash size 8388608, sector count = 128
Flash:  8 MB
In:    serial
Out:   serial
Err:   serial
Net:   ag7240_enet_initialize...
Fetching MAC Address from 0x81feb688
Fetching MAC Address from 0x81feb688
: cfg1 0x5 cfg2 0x7114
eth0: 00:c0:ca:8d:5c:45
eth0 up
: cfg1 0xf cfg2 0x7214
eth1: 00:c0:ca:8d:5c:44
athrs26_reg_init_lan
ATHRS26: resetting s26
ATHRS26: s26 reset done
eth1 up
eth0, eth1

Please choose the operation: 
   1: Entr boot command line interface. 
   2: Load system code then write to Flash via TFTP. 
   3: Boot system code via Flash (default).
 3 2 1 0

   
 System Boot system code via Flash.
## Booting image at 9f650000 ...
Bad Magic Number
```

I'm following the instructions to perform a tftp flash from here: [https://wiki.openwrt.org/toh/alfa.network/hornet-ub#serial](url)

However some addresses are different.

pbriesch:
The router is running again! I followed the above mentioned instructions. However, the default IP-adresses were different:
`TFTP from server 192.168.2.222; our IP address is 192.168.2.10`
I flashed a gluon sysupgrade image.
Thanks again!

PR #829, opened 2016-07-25
gluon-neighbour-info: improve verbosity on wrong cli parameters

lemoer:


package/gluon-neighbour-info/src/gluon-neighbour-info.c
```diff
@@ -206,10 +208,20 @@ int main(int argc, char **argv) {
...
+    fprintf(stderr, "No request string supplied\n");
     exit(EXIT_FAILURE);
   }
 
+  if (client_addr.sin6_port == htons(0)) {
+	fprintf(stderr, "No port supplied\n");
+	exit(EXIT_FAILURE);
+  }
+
+  if (!destination_set) {
```
NeoRaider: You don't need to introduce a new variable, you can just keep the ip6addr_any default value and check with IN6_IS_ADDR_UNSPECIFIED().

lemoer:
Thanks, I did not know about ```IN6_IS_ADDR_UNSPECIFIED()``` before.

Assigning ```in6addr_any``` into ```sin6_addr``` is not necessary, since the data structure is initialized with zeros.

PR #839, opened 2016-07-30
gluon-{radvd,respondd}: Use procd in init scripts

jplitza:
The main advantage is that procd is capable of logging stderr to logd,
making errors easier to debug.

### DIFF
--- a/package/gluon-radvd/files/etc/init.d/gluon-radvd
+++ b/package/gluon-radvd/files/etc/init.d/gluon-radvd
@@ -1,17 +1,15 @@
 #!/bin/sh /etc/rc.common
 
+USE_PROCD=1
 START=50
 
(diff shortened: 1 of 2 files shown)


NeoRaider:
No objections regarding gluon-radvd, but I don't think the locking in gluon-respondd can be dropped (it prevents a race condition of manual/autoupdater executions of the init script against the hotplug restart_if_running call). We should finally implement a way to update the interface list without restarting respondd...

jplitza:
I have to admit I'm not sure, but I assumed that procd has its own locking. Will do more research.

jplitza:
Okay, committed the radvd part in d445cb7, expect a PR for respondd that introduces another method of supplying interfaces.

PR #840, opened 2016-07-30
gluon-respondd: Adapt to new respondd version and change provider path

jplitza:
See freifunk-gluon/packages#143

Rationale for the change of the data directory: It is impossible for respondd data providers to be non-Gluon-specific (as striven for and discussed in #826) and at the same time work with Gluon as long as our data directory is under /lib/gluon.


rotanid:
adding the bug label, too, because the related packages-change could fix an issue where respondd is sometimes crashing and then isn't running until the node is rebooted or someone with remote access starts respondd again.
that is bad, because the nodes disappear on a map that doesnt use alfred and on maps that use alfred-data the node has less information. and it affects the status page, too.

rotanid:
@jplitza would be nice to have this, could you tell us the reasons for it being labled "blocked" ?

jplitza:
Well, this particular PR is blocked until freifunk-gluon/packages#143 has been merged (which is why I added the tag – not really a reason to add the tag in hindsight), which in turn seems to be blocked by freifunk-gluon/packages#144 because they conflict.

jplitza:
I just closed the original PR freifunk-gluon/packages#143, so I'm also closing this one.

The move of the data directory is now a separate PR in #971, alternatively there's an effort to have support for multiple directories in freifunk-gluon/packages#153

PR #849, opened 2016-08-07
modules: change url from openwrt.org to github one

RubenKelevra:
This does not work though, but the original git is down and I've no idea how to specify the remote ref correctly so feel free to fix it.

### DIFF
--- a/modules
+++ b/modules
@@ -1,7 +1,8 @@
 GLUON_FEEDS='openwrt gluon routing luci'
 
-OPENWRT_REPO=git://git.openwrt.org/15.05/openwrt.git
+OPENWRT_REPO=git://github.com/openwrt/openwrt.git
 OPENWRT_COMMIT=eadf19c0b43d2f75f196ea8d875a08c7c348530c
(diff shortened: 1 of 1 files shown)


modules
```diff
@@ -1,7 +1,8 @@
 GLUON_FEEDS='openwrt gluon routing luci'
 
-OPENWRT_REPO=git://git.openwrt.org/15.05/openwrt.git
-OPENWRT_COMMIT=eadf19c0b43d2f75f196ea8d875a08c7c348530c
+OPENWRT_REPO=git://github.com/openwrt/openwrt.git
+OPENWRT_COMMIT=e663db7bb1797740c4a29ac0fc96eda1b88f9e6e
+OPENWRT_BRANCH="chaos_calmer:chaos_calmer"
```
NeoRaider: This should be just `OPENWRT_BRANCH=chaos_calmer`.

RubenKelevra:
@NeoRaider I've changed it, thanks for the hint :)

PR #859, opened 2016-08-22
ar71xx-generic: remove broken mark for wr842n-v3

simonwunderlich:
I've bought a couple of those devices from Senetic GmbH.

https://www.senetic.de/product/TL-WR842N

They have 16 MB of Flash and 64 MB of RAM. Platform support works fine,
I've also tested a little with Ethernet (since I saw some regressions on
OpenWRT/LEDE with 841v11), no problems.

Therefore, lets remove the broken mark.

Signed-off-by: Simon Wunderlich <sw@simonwunderlich.de>


rubo77:
Nice, they also have 650 MHz, see https://wiki.openwrt.org/toh/tp-link/tl-wr842nd#info
did you try the throughput speed with fastd?

Maybe you can add some test report here?
Or in the Forum: https://forum.freifunk.net/t/testbericht-tl-wr842nd-v3/13369

rotanid:
nice, thanks for your report and the link to a supplier, couldn't find anyone delivering v3 before!

PR #865, opened 2016-08-28
Remove squashing from contribution guidelines

jplitza:
Github by now allows project maintainers to squash PRs with a few clicks, thus imposing the squashing onto contributors isn't really necessary anymore in my opinion.


NeoRaider:
I think this is fine.

In individual cases, we might like to have a complex change split into meaningful commits, but as this won't matter for most changes (and requires good git skills to fix afterwards), I don't think it needs to be addressed in the contribution guidelines.

Issue #842, opened 2016-08-03
problem with upgrade from 2016.1.5 to master, duplicate wifi mac


rotanid:
tonight our routers on our experimental firmware branch updated from an 2016.1.5-based firmware to an master-based one.
all of these routers have both ibss and 11s meshing enabled.

in one location, two ath9k dual-band routers (WDR3600) mesh over 2,4ghz (too far away for 5ghz to work).
the one which has the vpn-uplink also has private wifi enabled, so 4 wifis on 2 radios makes 8.

it broke a bit due to the update:
the client_radio0 mac address wasn't similar to the client_radio0 as it is on other 2016.1.5-based devices.
it was the same as the one for wlan0-2 and as a consequence, hostapd didn't want to start and the 2,4ghz radio wifi interfaces were down.

i could fix it by copying the mac from client_radio1 in /etc/config/wireless and only change the 3rd octet (compared the macs on other devices to come up with that solution)

what could have caused that issue, are 8 wifi interfaces too much for master branch upgrade scripts?

RubenKelevra:
Would you be so kind to check the wan-mac for gigabit-devices? I think we've discovered a but there too. On our experimental dev build with the old building-system all wan-macs on gigabit-devices were random.

rotanid:
what exactly is "experimental dev build with the old building-system" ? :-D
if you talk about the mac of the "br-wan" interface, it is the same as the mac of the "client1" interface on the two wdr3600 i am able to check at the moment. @Brother-Lal might have another wdr3600 running the same gluon-master-based firmware

NeoRaider:
I'm a bit confused why client0 and wlan0-2 have the same MAC address... client0 is expected to have a completely different address than in v2016.1.x, but I don't see how wlan0-2 gets the same address.

Is there any address set for the private-wifi interface in `/etc/config/wireless`?

rotanid:
no, wan_radio0 and wan_radio1 have no mac address set in that config file

rotanid:
no, wan_radio0 and wan_radio1 have no mac address set in that config file.
i have a few of the wdr3600 left, i could do the upgrade again but i don't think that it will be different...

benerudolph:
I do experience the same behavior stated above on an Archer C7 (it has two wifi interfaces) under the very same circumstances.
My logread prints this, when the system starts the hostapd process that is supposed to manage the 
WiFi interfaces:
```
[...]

Issue #834, opened 2016-07-28
MAC address assignment incompatible with Bridge Loop Avoidance


NeoRaider:
Causes

    br-client: received packet on bat0 with own address as source address

as BLA uses packets with batman-adv's primary address as source.

RubenKelevra:
In case you got a look or no loop on the network?

NeoRaider:
@RubenKelevra, already occurs with a single isolated node.

Byggvir:
This kern.warn is only on our nodes generated after 22nd Jul. The 15th April image is running fine. It also occurs on single nodes without neighbors.

stweil:
I see `br-client: received packet on bat0 with own address as source address` on different router hardware with recent firmware. Is that a critical error? Should it be reported in a new issue?

PR #870, opened 2016-08-30
Introduce generic gluon_mesh netifd proto and reorganize MAC addresses

NeoRaider:
This PR is a new attempt at making the core network config mesh protocol agnostic, while giving us better control over the setup of the mesh.

In contrast to #747, which is replaced by this, the network interfaces and their properties (transitive connectivity, fixed MTU) are stored in UCI now instead of moving this information to a JSON file, so the user can extend the config with additional mesh interfaces as before.

Contains fixes for #834 and #842.


NeoRaider:
Testers welcome.

PR #871, opened 2016-08-30
 ar71xx-generic: add support for ALFA N2 and N5

ambassador86:
+ added profile for these devices
+ added entry in the docs for these devices plus the missing alfa tube


ambassador86:
I tested two of them on a street festival without issues.
https://meshviewer.chemnitz.freifunk.net/#!v:m;n:00c0ca60bcb2
https://meshviewer.chemnitz.freifunk.net/#!v:m;n:00c0ca66eba6

Issue #453, opened 2015-08-29
Change label for altitude


rubo77:
This is a followup of #381:

Since for most people it is not obvious how to find out the correct altitude (above NN or similar) of a certain building,

> we should have a button: "show altitude of streetlevel" in the mesh viewer or a link to a map, that has this easily accessible

Or we just change the label to 

"Altitude above streetlevel"

do9xe:
Is this a request for a change in the packages textfiles or a request to add a feature to the mesh-viewer? If it is for the meshviewer, you are in the wrong repository.
If it is for the package: please create a pullrequest by yourself.

do9xe:
After talking to @jplitza and @tcatm we decided that the visualisation of the altitude gets to complicated, when you use the altitude over ground. So this is going to stay altitude over NN, until we discussed this in a propper way.

rubo77:
Until now its not visible that it should be "altitude over NN" it just said "altitude"

Did you find a map that shows the right altitude?

jplitza:
As mentioned yesterday, we should define more precisely what is meant by "latitude" and "longitude" (as in "according to WGS84 projection") as well.

rubo77:
NN is replaced by [NHN (Normalhöhennull)](https://de.wikipedia.org/wiki/Normalh%C3%B6hennull#Umstellung_von_Normalnull_auf_Normalh.C3.B6hennull)

so the title for https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-config-mode-geo-location/i18n/de.po
should be

    msgstr "Höhe über NHN"

with NHN as an external link to get the height for the entered coordinates in the fields above

bittorf:
rubo77: 8-))) *no*body (except you) knows NHN without explanation

rubo77:
The myin point is, We should use the most commonly used system and add a link to a site te receive that altitude in config mode directly, so everyone uses the same system.

(by the way NHN only differs by some 10cm from the old NN)

bittorf:
rubo77: now i'am unsure: do we talk about "Altitude above streetlevel" or NN? also: better use 'NN' if you really talk about this: it's difficult anyway to measure that exactly...

rubo77:
We could get the information from the Google Elevation API.

By making a GET call and passing in the latitude and longitude in the query, you can get back the altitude from sea level in meters:

    https://maps.googleapis.com/maps/api/elevation/json?locations=40.714728,-73.998672

This will return a json object like this:

[...]

PR #872, opened 2016-08-30
Rename Altitude to Storey

rubo77:
This replaces the unintuitive value "altitude" with Storey, which is either known, or can be aproximated by imaginary stroeys of 3m per storey

TODO: 

- [ ] migtrate script, that fills the field with a third of the old altitude value, (if given)
- [ ] french translation


kb-light:
I see the same problems as @NeoRaider mentioned already in #581. For example at University we have a building, where storey 2 is the ground-floor (others starts with storey 0).
Additionally the router can be placed on the floor or be ceiling-mounted.  And if you have some outdoor installation, it is maybe more complicated to reverence to a storey, than to some altitude.
So I think "storey" is more complicated than altitude and I see no difference to "altitude in meters above streetlevel" if you can "aproximate" it with 3m.

In my opinion the goal should be to get an easy and accurate method for measuring the altitude, for example with a navigation app on your smartphone. And not trying to rename "altitude" every few month.

rubo77:
there is no easy way to measure the altitude with a navigation app on your smartphone, because they are incredibly inaccurate (several 10m inaccurate) and another outdoor option won't be there for a normal user.

So there is either:

 - no altitude information at all
 - an **only informative**, but not physical precise, information to place routers inside a house relative to each other

I think the latter is a very useful information if you place more than one router in the same building.

kb-light:
The first is the way we actually proceed. The altitude field is disabled within the config mode, because we see the problem with actually not well defined altitude, too.
The second might by interesting, when searching the router, but it is nor usable for calculating distances between the routers.

I agree, that 10m is not accurate enough. Currently I am thinking about getting the altitude by some map services similar to lat/lon, which was already mentioned by you in #453. Then the owner has to estimate the altitude above "streetlevel" and add it to the value. If the steetlevel is not well defined in some point, one can chose an other location as reference.

Adorfer:
Nur weil es einige Gebäude an Hanglage gibt muss man das Feature doch nicht für alle ruinieren.
Und da die meisten Gebäude einen postalischen Haupteingang haben (der als EG/L0 zählt), sollte das mit der Etagenzählung "in OSM-Stil" eigentlich kein Problem darstellen. 
Wobei da jetzt bestimmt jemand mit einem Wohnblock mit mehreren Hausnummern und Eingängen am Hang kommt...

Fakt ist, dass sich die Höhe mit bezahlbarem GPS nicht sauber ermitteln lässt, die Abweichung ist etwa um Faktor 4 schlechter als die in lat/lon. Und da freut man sich über 5m Genauigkeit. In der Praxis weist die Höhe eine Abweichung von minimal +/-10 m auf, tendenziell eher 20m, wenn man nicht "runherum" freie Sicht auf den Horizont hat.
Wenn also zwei Leute im gleichen Haus nebeneinander je einen Node aufstellen, dann stehen die mit etwas Pech auf 80m und 120m, bei 20m seitlichem Abstand.

jplitza:
I think two things are obvious:
* There are people unhappy with the current semantics of the height field, wanting to change it to something different.
* There are people happy with the current semantics of the height field, not wanting to change it to something different.

Thus, we won't reach consensus on a Gluon-wide approach. Instead, I propose to rename the field to something neutral (although "altitude" already is kind of neutral, according to Wikipedia, and "height" has some ISO meanings as well…) and move all descriptional messages for this field into site translation files. That way, every community can decide on its own what semantics this field should have and how to best describe it.

As that is basically also what @tcatm and @NeoRaider proposed in #581, I think it is save to say the current PR won't get merged. Thus, I'll close it.

rubo77:
OK, I created #873

PR #883, opened 2016-09-13
gluon-config-mode-geo-location: Make the label for altitude definable…

### LINKED ISSUE #873: Make the label for altitude definable in the site folder
We need to remove the current strings and replace them with a placeholder.

rubo77:
… in the site folder

closes #873

### DIFF
--- a/docs/site-example/i18n/de.po
+++ b/docs/site-example/i18n/de.po
@@ -42,3 +42,11 @@ msgstr ""
 "Betrieb für drei Sekunden den Reset-Button. Das Gerät wird dann im Config "
 "Mode neustarten.</p>"
 "<p>Viel Spaß mit deinem Knoten und der Erkundung von Freifunk!</p>"
+
+msgid "gluon-config-mode:altitude-label"
+msgstr "Höhe"
+
+msgid "gluon-config-mode:altitude-help"
+msgstr ""
+"Die Höhenangabe ist optional und sollte nur gesetzt werden, wenn ein "
+"exakter Wert bekannt ist."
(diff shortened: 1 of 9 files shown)


rubo77:
Successfully tested on a 841 v8. looks for example like this then:

altitude

rubo77:
can this be merged? We would like to customize the altitude message in our community.

rubo77:
I tried it out and it seems like translations added to your sites i18n files is stronger than the default translations defined in the gluon packages.

So you can already achieve a customized wording in your config mode by adding these two translations to your site config:

```
msgid "Altitude"
msgstr "Höhe (3m pro Stock)"

msgid ""
"Specifying the altitude is optional and should only be done if a proper "
"value is known."
msgstr ""
"Die Höhenangabe ist optional und kann geschätzt werden mit 3m pro Stockwerk."
```

I think we should add this information to the documentation, so everyone knows, how to enhance their config mode wordings to their needs

NeoRaider:
As explained in the other thread, there is no way to know which version is used when there are duplicate translation records. Even if it works now, this may change at any time.

Therefore, this PR is still needed.

rubo77:
Renamed to altitude-help.

as mentioned in the other issue #953 maybe we still find a way to add these by replacing the default texts. that would be more flexible over all. I will search the next days for an explanation which file gets precedence and how to force it eventually

NeoRaider:
Looking pretty good now, one more note: I'd place the `altitude-label` definitions before `altitude-help` in the .pot/.po files and the site docs.

rubo77:
done

PR #884, opened 2016-09-13
Firmware upgrade: reload after 3 Minutes

rubo77:
This will redirect to http://192.168.1.1/ from the Firmware Upgrade screen after a safe timeout

### DIFF
--- a/package/gluon-luci-admin/files/usr/lib/lua/luci/view/admin/upgrade_reboot.htm
+++ b/package/gluon-luci-admin/files/usr/lib/lua/luci/view/admin/upgrade_reboot.htm
@@ -5,6 +5,7 @@
     <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
     <title><%:Upgrading firmware%></title>
     <link rel="stylesheet" type="text/css" media="screen" href="<%=media%>/cascade.css" />
+    <meta http-equiv="refresh" content="180; URL=http://192.168.1.1/">
   </head>
   <body>
     <div id="maincontainer">


jplitza:
Why?

viisauksena:
thats the page where fastd pubkey and maybe some instructions are shown ..? .. i would yell at hell if that move away after x seconds ..

rubo77:
No.

That's the page, where it only sais, the Router is restart Afterburner a manual Firmware Upgrade

viisauksena:
ah, i see - and if somebody flashed factory - there is a need to do configmode - so 192.168.1.1 sounds good

NeoRaider:
The router may not boot into config mode after the update, I think this doesn't really make sense.

rubo77:
It is rebooting into config mode when you selected to forget all settings though, so it should only redirect there automatically in that case

rubo77:
Couldn't this be implemented if you upgraded without "keep settings" checked?

Also there could be a spinner on the page where it sais:


> Die Firmware wird jetzt aktualisiert. Unterbrich auf keinen Fall die Stromversorgung! Dieser Vorgang wird einige Minuten dauern. Anschließend startet das Gerät automatisch neu.


this would be a CSS only spinner: https://codepen.io/rubo77/pen/jLpYBB

bobcanthelpyou:
@rubo77 it is also possible to skip config mode via site.conf
```
setup_mode = {
  skip = true,
},
```
http://gluon.readthedocs.io/en/latest/user/site.html

Adorfer:
off course. But the requested change would match likely more than 95 percent of use cases.

bobcanthelpyou:
@Adorfer but the redirect should be set for these 95 percent, only .. 

Maybe a preselected checkbox 'start in config mode' or 'redirect me to config mode start page' or something similar could help here. But it still wouldn't work in some settings ..

Maybe a 'splash page' via DHCP in config mode could be a solution for that case (redirect all request to 192.168.1.1)?

Adorfer:
@bobcanthelpyou  redirection would be very nasty for people who connect via lan to the node and have a second connection already present. 
So i would rather like a config mode where the dhcp neither announces a DNS, nor announces a default route (but that's potentially another issue.)

PR #885, opened 2016-09-13
Docs: Add email subject and body to site-example

rubo77:


jplitza:
I like the idea. However, I think explaining how mailto-Links work doesn't make sense. Either there's a handler configured and people are used to clicking on them and will notice the information is already prefilled, or not, in which case they are probably used to simply copying the address.

Also, for consistency, you used spaces around the pubkey variable but not around the others.

rubo77:
You think, the small text is obsolete?

jplitza:
Exactly.

rubo77:
The small hint looks like this: 

auswahl_045

NeoRaider:
I don't like the small text either.

Also, this is missing escaping (`escape()` in the text, `urlescape()` in the URL).

In addition, the other languages need to be updated, at least English.

rubo77:
I also added `urlescape(sysconfig.primary_mac)` don't know if escaping is needed here?

NeoRaider:
The primary MAC doesn't need to be escaped.

rubo77:
rebased to master

docs/site-example/i18n/de.po
```diff
@@ -23,7 +23,9 @@ msgstr ""
 "kann sich dein Knoten mit dem Mesh-VPN dort verbinden. Bitte "
 "schicke dazu diesen Schlüssel und den Namen deines Knotens "
 "(<em><%=escape(hostname)%></em>) an "
-"<a href=\"mailto:keys@alpha-centauri.freifunk.net\">keys@alpha-centauri.freifunk.net</a>."
+"<a href=\"mailto:keys@alpha-centauri.freifunk.net?subject=Anmeldung:%20<%= urlescape(hostname) %>&amp;"
+"body=%23%20<%= urlescape(hostname) %>%0A%23%20<%= usysconfig.primary_mac %>%0Akey%20%22"
```
NeoRaider: There's a typo here (usysconfig instead of sysconfig), in all three versions.

rubo77: corrected.

NeoRaider:
Another idea: how about enclosing the whole body of the mail in a single *urlescape* call? That would avoid all that unreadable percent-encoded text. The same could be done with the subject, but it's not as important there.

rubo77:
that would look much cleaner, you can add that while merging? I have no way to test this before friday

NeoRaider:
Merged. I changed it to use a single *urlencode*, and made the body match the other part of the example.

PR #900, opened 2016-10-04
ar71xx-generic: Add support for TP-Link Archer C7 v1

ecsv:
Tested-by: Maximilian Hahn <multimediahahn@gmail.com>

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -156,7 +156,7 @@ ar71xx-generic
 * TP-Link
 
   - Archer C5 (v1) [#ath10k]_
-  - Archer C7 (v2) [#ath10k]_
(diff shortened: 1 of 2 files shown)


NeoRaider:
I'm surprised, I thought the QCA9880-AR1A was not supported by ath10k at all and no support was planned. Does the 5GHz WLAN really work on this device?

ecsv:
I never had this device. But @simonwunderlich  currently has the device from Maximilian Hahn. Maybe he can check

ecsv:
Just checked the OpenWrt wiki and forums and it looks like this device really is a QCA988x hw1.x. So I will now close this pull request because I have doubts that it was actually tested.

PR #915, opened 2016-10-25
Make fastd syslog level configurable

### LINKED ISSUE #896: wanted : fastd syslog level in site.conf (or stop flooding syslog)
i assumed that that i could set syslog level in fastd - 
i'd love to set it in site conf - without a hacky way 

while we have plenty Gateway Servers, our syslog become pretty unreadable with all the fastdinfo - it is limited in (ring)size - and by hunting down other problems the 24/7 running fastd even in mesh only routers spam the syslog a lot 
for us 
uci set fastd.mesh_vpn.syslog_level='warn'
or in site conf under fastd mesh-vpn ... a syslog_level = 'warn' 
is enough, if we need more info in detail we could do it seperatly and easily on specific routers

viisauksena:
it is easy to set this to a higher value again if needed. This will prevent fastd from flooding the log ringbuffer, even when it is not needed/no uplink via fastd. 
This log - flooding raise with every peer known from site.conf.

Fixes #896

### DIFF
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -183,7 +183,7 @@ fastd_mesh_vpn
     if it is supported at all. You should only set `configurable` to `true` if the
     configured peers support both the ``null`` method and methods with encryption.
     
-    You can set syslog_level from verbose(default) to warn to reduce syslog output.
+    You can set syslog_level from verbose (default) to warn to reduce syslog output.
     ::
 
       fastd_mesh_vpn = {
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -182,12 +182,15 @@ fastd_mesh_vpn
     In any case, the ``null`` method should always be the first method in the list
     if it is supported at all. You should only set `configurable` to `true` if the
     configured peers support both the ``null`` method and methods with encryption.
+    
+    You can set syslog_level from verbose(default) to warn to reduce syslog output.
     ::
 
       fastd_mesh_vpn = {
         methods = {'salsa2012+umac'},
       	-- enabled = true,
       	-- configurable = true,
+	-- syslog_level = 'warn',
         mtu = 1280,
         groups = {
           backbone = {
--- a/package/gluon-mesh-vpn-fastd/luasrc/lib/gluon/upgrade/400-mesh-vpn-fastd
+++ b/package/gluon-mesh-vpn-fastd/luasrc/lib/gluon/upgrade/400-mesh-vpn-fastd
@@ -21,6 +21,12 @@ if not enabled then
 end
 
 
+local syslog_level = uci:get('fastd', 'mesh_vpn', 'syslog_level')
+if not syslog_level then
+  syslog_level = 'verbose'
+end
+
+
 local methods
 
 if site.fastd_mesh_vpn.configurable then
@@ -52,7 +58,7 @@ uci:section('fastd', 'fastd', 'mesh_vpn',
 	  {
 		  enabled = enabled,
 		  group = 'gluon-fastd',
-		  syslog_level = 'warn',
+		  syslog_level = syslog_level,
 		  interface = 'mesh-vpn',
 		  mode = 'tap',
 		  mtu = site.fastd_mesh_vpn.mtu,
--- a/package/gluon-mesh-vpn-fastd/luasrc/lib/gluon/upgrade/400-mesh-vpn-fastd
+++ b/package/gluon-mesh-vpn-fastd/luasrc/lib/gluon/upgrade/400-mesh-vpn-fastd
@@ -52,7 +52,7 @@ uci:section('fastd', 'fastd', 'mesh_vpn',
 	  {
 		  enabled = enabled,
 		  group = 'gluon-fastd',
-		  syslog_level = 'verbose',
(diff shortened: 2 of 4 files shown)


viisauksena:
if i check if syslog_level is set in mesh_vpn similar to the enabled stuff in line 18
```local syslog_level = uci:get('fastd', 'mesh_vpn', 'syslog_level')
if not syslog_level then
  syslog_level = 'warn'
end
```
then it is sufficient to set line 55 from 
	syslog_level = 'warn',
to 
	syslog_level = site.fastd_mesh_vpn.syslog_level,
am i right with this? and if so, i need to add some lines to the doku right?

NeoRaider:
Hmm, I'd rather increase the ring buffer size, as fastd connection issues are a common problem, so having the connection attempts in the log is helpful.

viisauksena:
you can guess yourself, having 10 gw , with limit 2 will fill this buffer very fast (even if bigger) - since the connection attempt appear even if limit is fullfilled or no fastd/uplink is available at all
an event which occurs some hours ago is then lost

viisauksena:
if my tests were right - this should do the trick to have it configurable while keeping your default (verbose) if not set explicitly in site.conf

package/gluon-mesh-vpn-fastd/luasrc/lib/gluon/upgrade/400-mesh-vpn-fastd
```diff
@@ -21,6 +21,12 @@ if not enabled then
 end
 
 
+local syslog_level = uci:get('fastd', 'mesh_vpn', 'syslog_level')
+if not syslog_level then
+  syslog_level = 'verbose'
+end
```
NeoRaider: This could be simplified to
```lua
local syslog_level = uci:get('fastd', 'mesh_vpn', 'syslog_level') or 'verbose'
```

docs/user/site.rst
```diff
@@ -182,12 +182,15 @@ fastd_mesh_vpn
     In any case, the ``null`` method should always be the first method in the list
     if it is supported at all. You should only set `configurable` to `true` if the
     configured peers support both the ``null`` method and methods with encryption.
+    
+    You can set syslog_level from verbose(default) to warn to reduce syslog output.
```
NeoRaider: Please add a space before the parentheses.

NeoRaider:
This looks mostly okay, please also add the new optoin to the site check script to ensure only valid values can be set.

viisauksena:
i am not sure in all allowed values while fastd seem not to allow the full scale of possible debug values - i remember having some trouble with them by manualy setting them. 
I guess you mean https://github.com/freifunk-gluon/gluon/blob/master/scripts/check_site_lib.lua -  i didnt get where/how to ensure valid values there

NeoRaider:
You'd add a line like the following to https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-mesh-vpn-fastd/check_site.lua
```lua
need_array_of('fastd_mesh_vpn.syslog_level', {'error', 'warn', 'info', 'verbose', 'debug', 'debug2'}, false)
```

NeoRaider:
Eeeeh, not `need_array_of`, I mean `need_one_of`

viisauksena:
ich mach das heut nacht mal fertig

viisauksena:
done ! - 
thx for help

package/gluon-mesh-vpn-fastd/luasrc/lib/gluon/upgrade/400-mesh-vpn-fastd
viisauksena: if you prefer your simplification , i could change it to your suggestion

NeoRaider: That would be great, it would also make the coding style match other scripts.

package/gluon-mesh-vpn-fastd/luasrc/lib/gluon/upgrade/400-mesh-vpn-fastd
```diff
@@ -15,11 +15,9 @@ users.remove_user('gluon-fastd')
 users.add_group('gluon-fastd', 800)
 
 
-local enabled = uci:get('fastd', 'mesh_vpn', 'enabled')
-if not enabled then
-  enabled = site.fastd_mesh_vpn.enabled and 1 or 0
-end
+local enabled = uci:get('fastd', 'mesh_vpn', 'enabled') and 1 or 0
```
NeoRaider: Please don't change unrelated code - in this case, you've even broken it by dropping `site.fastd_mesh_vpn.enabled`, which is what the whole `and 1 or 0` is about.

viisauksena: mhm, tested on a random node - seemed to work fine - but i can delete this part

viisauksena: mhm, i didnt get the problem - but ok so far - i reverted it.

Issue #753, opened 2016-05-04
System instability above specific point


nazco:
As I've reported on IRC we at FFRN can observe a higher load and frequent reboots on nearly 30% of our nodes.This happens if we reach a specific number of nodes and clients in our network. We have debugged this issue for over a month now and we think we've limited the possible sources.

Let's start with our observations. The first time we became really aware of this problem was the moment we reached the number of 1500 clients in our network spread over nearly 700 nodes. This happened around the first of April this year. But after analyzing the problem we think it started even earlier with some "random" reboots we were analyzing too.

The first thing we observed was, that the majority of the affected nodes are the small tl-wr841 devices. This does not mean, that the bigger nodes like a tl-wr1043 are not affected, the problem just hasn't a big enough impact on them. But interestingly not all of these nodes where affected, only a portion of 30% is showing all the characteristics of this problem. All other nodes are running without any interruptions.

On an effected node we can see the following: If we reach more that 3000 entries in the transglobal table the problems start and if the number is falling under this mark the problems are mainly gone. Such a node shows a increased average load of around 0.45  to 0.9 compared with 0.2 to 0.25 of an not effected node. But the load also starts to peak to values of 2-4 in the time we are above the mark on the problematic nodes. And sometimes every few hours or every few minutes the node reboots.
Another interesting observation is, that affected nodes get alot of more free RAM when the problems start. The RAM usage decreases from the healthy default of around 85% (on a 841) to 75%-80%.

On a TL-1043 it looks like this: 

![image](https://cloud.githubusercontent.com/assets/1292551/15027466/3e56eff6-1244-11e6-951c-fca632d1f8ef.png)

On a TL-841v9 it looks like this:
![image](https://cloud.githubusercontent.com/assets/1292551/15027728/76d2f8ce-1245-11e6-9642-ae721f9816cf.png)

At all times we can't see a single process making problems or using more RAM than usual, only the load and the system CPU utilization showed that something went wrong. So we though that the problem has to be in the kernel or in combination with the RAM.

So we started to debug the problems and first we tried to locate a pattern in our statistics to limit the number of possible sources for the problem. There were alot of other ideas we tested but all with nearly no effect. So the most promising was the TG table. But it's not the number of entries because we can't find any limits near this number in the sources and also some other things are speaking against it. So the problem has to be in the processing of the entries or somewhere else.

[...]

PR #838, opened 2016-07-29
Add package gluon-radv-filterd

jplitza:
This package drops all incoming router advertisements except for the default router with the best metric according to B.A.T.M.A.N. advanced.

Note that advertisements originating from the node itself (for example via gluon-radvd) are not affected.

This PR supersedes #718.

Caveat: Router solicitations sent by clients are still broadcast, and all but one response are dropped. Making them unicast would require fiddling with the packet, which isn't easily possible from userspace as we've seen in #718.

PR #916, opened 2016-10-28
RFC: Switch from batman-adv debugfs to netlink

ecsv:
The batadv debugfs requires large memory blocks to write the text debug tables. This is inefficient for large tables like the global translation table or the originators table.

The memory requirement can be reduced by using netlink. It copies smaller packets in a binary format to the userspace program. gluon can therefore parse larger originator tables without causing an OOM on systems which are tight on memory.

This is only an RFC because it may collide with the LEDE port, batman-adv-legacy support for netlink is missing and it would actually be nicer to update the openwrt-routing feed instead of having to use patches for it. The code is an updated version of what was already posted in #753


NeoRaider:
For the update to batman-adv 2016.4, I would update the *openwrt-routing* feed to the current master (as soon as the 2016.4-maint PR is in).

ecsv:
Ok, sounds good. I have now updated the RFC to only include the patches which are currently missing in openwrt-routing

belzebub40k:
Just for completeness. gluon-radv-filterd #838 is also using debugfs at the moment.

jplitza:
Having a set of three files (netlink.c, netlink.h, batman_adv.h) in every package (currently: 3 with #838) that queries batman-adv seems suboptimal. I didn't have a closer look yet, how difficult would it be to have a library that handles querying the netlink interface and package it separately?

ecsv:
It is basically a single function (netlink_query_common) and a package specific definition for batadv_netlink_policy. The batman_adv.h is a linux kernel uapi header which is just not yet available in the kernel from gluon. I personally will not create a library for that.

jplitza:
I see. Wouldn't it make sense to include the batman_adv.h as a patch against the kernel then?

ecsv:
The requested changes were pushed. Now it is not enough to do a "make clean" but instead a complete toolchain rebuild is required to get linux/batman_adv.h (basically to recreate toolchain/kernel-headers).

NeoRaider:
Hmm, I'd prefer to add a batman-adv-provided include path to the TARGET_CFLAGS of the relevant packages rather than patching it into the kernel, otherwise it will be another patch to carry over when switching to LEDE.

ecsv:
I personally will not do that.

NeoRaider:
@ecsv, do you still have the previous version of this branch, with each package including the *batman_adv.h* itself? I'd be willing to merge it like that (after all, including a copy of required UAPI headers that can't be taken for given is quite common).

ecsv:
@NeoRaider I have it in https://github.com/FreifunkVogtland/gluon/tree/batadv-netlink-perpackage . But there is still the batman-adv-legacy change missing in https://github.com/freifunk-gluon/packages/commits . I personally would recommend not to merge it until it is rebased on a patch which imports the gluon-packages with the batman-adv-legacy import.

I don't know what your plans are for upgrading the batman-adv-legacy in gluon-packages.

NeoRaider:
Yes, that makes sense. I have the update staged locally, I'll see that I can get it tested and pushed soon.

NeoRaider:
Both batman-adv versions have been updated to a version with netlink support.

ecsv:
Looks like you've broken reference counting in batadv_send_outstanding_bcast_packet. See my original patch for updating to 2016.4 for details

PR #921, opened 2016-11-06
Update getting_started.rst

A-Kasper:
just changed to an existing example


NeoRaider:
v2016.2.x is not a release, it's a branch. The next release will be v2016.2.1.

rubo77:
But explain to still Use the broken Tag is also not Wise.

Are there any problems in the x-branch?

A-Kasper:
That was my intention. If you use the documented example you're using
master.

Am 07.11.2016 06:37 schrieb "Ruben Barkow" <notifications@github.com>:

PR #927, opened 2016-11-10
pata and mmc drivers for x86 targets

rotanid:
first, this PR adds the PATA driver modules in x86-generic config also to the x86-64 target
second, it adds support for booting from a sdcard to x86-generic and x86-64


NeoRaider:
Are these changes in LEDE already?

rotanid:
difficult situation.

PATA drivers:
are in x86-generic for gluon and openwrt, but not anymore in LEDE, removed in febc6cc5bf4b9523842097e917e0ba9addf18193 without a reason?!?
and in both cases not in x86-64, except of one drive which is active in LEDE (the one used for Futros)

MMC drivers:
are in x86-64 for LEDE, but not openwrt
aren't in x86-generich for neither

as i already wrote on IRC, i don't see any clear sense in what's included in x86-generic or not and what's included in x86-64 or not - and even stranger that some part was removed without a comment, see above.

PR #937, opened 2016-11-16
Add site commit-ID and dirty-flag to gluon and site commit descriptions

rubo77:
First test:

```
cat build/ar71xx-generic/profiles/TLWR841/root/lib/gluon/site-remote 
https://github.com/rubo77/site-ffki/

cat build/ar71xx-generic/profiles/TLWR841/root/lib/gluon/site-version 
v0.4-140-g6de441e
```
which looks fine. you can construct this URL out of it: [https://github.com/rubo77/site-ffki/tree/**6de441e**](https://github.com/rubo77/site-ffki/tree/6de441e)

for the gluon repo:
```
cat build/ar71xx-generic/profiles/TLWR841/root/lib/gluon/gluon-remote 
https://github.com/freifunk-gluon/gluon.git
```
this could be improved, cause I really used my own repo: https://github.com/rubo77/gluon.git

the url shows my origin, but not the branch, I actually built upon. but the commit Id is correct anyway:

```
cat build/ar71xx-generic/profiles/TLWR841/root/lib/gluon/gluon-version 
v2016.2.1-4-g7a7b1a7
```
the result shows the right commit, but not the right organization:

https://github.com/freifunk-gluon/gluon/tree/7a7b1a7 (here  you had to cutoff the `.git` at the end)

### DIFF
--- a/include/gluon.mk
+++ b/include/gluon.mk
@@ -35,8 +35,11 @@ $(GLUON_SITEDIR)/site.mk:
 -include $(GLUON_SITEDIR)/site.mk
 
 
-GLUON_VERSION := $(shell cd $(GLUONDIR) && git describe --always 2>/dev/null || echo unknown)
+GLUON_VERSION := $(shell cd $(GLUONDIR) && git --git-dir=.git describe --always --dirty=+ 2>/dev/null || echo unknown)
 export GLUON_VERSION
+
+GLUON_SITE_VERSION := $(shell cd $(GLUON_SITEDIR) && git --git-dir=.git describe --always --dirty=+ 2>/dev/null || echo unknown)
+export GLUON_SITE_VERSION
 
 GLUON_LANGS ?= en
 export GLUON_LANGS
--- a/package/gluon-core/Makefile
+++ b/package/gluon-core/Makefile
@@ -38,6 +38,7 @@ define Package/gluon-core/install
 
 	$(INSTALL_DIR) $(1)/lib/gluon
 	echo "$(GLUON_VERSION)" > $(1)/lib/gluon/gluon-version
+	echo "$(GLUON_SITE_VERSION)" > $(1)/lib/gluon/site-version
 endef
 
 define Package/gluon-core/postinst


rubo77:
To get the real url you can use this:

    git remote get-url $(git for-each-ref --format='%(upstream:short)' $(git symbolic-ref -q HEAD)|cut -d/ -f1)

but that must get somewhat shorter somehow???

NeoRaider:
I already rejected this idea in #917, and I mean it. I really don't want this, it's too fragile and of questionable benefit.

rubo77:
This PR is about something completely different (I don't want any links in config mode or anything like that in the other issue), I just showed, that the information would be there to find the right Repository that was user during build

I think this is valuable information, that should be kept in the config files, so you can later reconstruct, how the firmware was built.

the three details I added here were missing.

One more thing that should be stored should be if the git status is " working directory clean" somehow, so you can see if any more uncommitted changes would be in the build.

NeoRaider:
Adding the site commit ID would be okay, as long as it doesn't break if the site dir is no git repo. Please note that your current code would just print the `git describe` output of the Gluon repo when site is no repository.

All other information (used remotes etc.) is internal to the build machine and not necessarily of any value for others. URLs may point to the local machine, internal repositories, ... Let's not go there, there is no sane way for the build system to know if such an URL is useful or not. The URLs should just be published on the download pages together with the images.

rubo77:
OK, I agree. If you know the exact commit id and you know the repository from the website, you can work it all out.

What do you think about adding some output about if the git status is " working directory clean" ? That would be valuable information if you want to check if the firmware was manipulated (without bad intends)

We could add something like `lib/gluon/gluon-git-status` that contains "clean" or an md5checkksum of `git diff`.

can you reopen this PR, so I can work on it?

I will just add the site commit-id then.

NeoRaider:
I'd rather add `--dirty=+` to all `git describe` calls, so a `+` is added to the output when there are local changes.

NeoRaider:
Note that all this information is worthless to track willfull manipulation, it will only catch accidental changes. The long-term goal would be to make builds reproducible (which is an ongoing project in OpenWrt/LEDE).

rubo77:
sure, this is not a proop, just more info for honest admins and users ;)

I added the changes, ok like this?

include/gluon.mk
```diff
@@ -35,9 +35,12 @@ $(GLUON_SITEDIR)/site.mk:
 -include $(GLUON_SITEDIR)/site.mk
 
 
-GLUON_VERSION := $(shell cd $(GLUONDIR) && git describe --always 2>/dev/null || echo unknown)
+GLUON_VERSION := $(shell cd $(GLUONDIR) && git describe --always --dirty=+ 2>/dev/null || echo unknown)
 export GLUON_VERSION
 
+GLUON_SITE_VERSION := $(shell cd $(GLUON_SITEDIR) && git describe --always --dirty=+ 2>/dev/null || echo unknown)
```
NeoRaider: Add `--git-dir=.git` between `git` and `describe` to ensure the command fails when GLUON_SITEDIR is no git repository.

NeoRaider:
Manually merged with some adjustments.

rubo77:
for info: this is merged in https://github.com/freifunk-gluon/gluon/commit/2a101ddd561057c4b82bc29dbc1d3028e848d1bf

The original Idea also include the used repository for the site was abandoned. So you have to manually communicate on your community site, which site repositories you use (gluon and site)

rubo77:
To communicate the site repository, I started this PR: https://github.com/freifunk-gluon/gluon/pull/1482
that adds the content of new `/lib/gluon/site-version ` to the info pyge in config mode.

NeoRaider:
@rubo77 Don't comment on closed issues/PRs unless an issue must be reopened.

PR #934, opened 2016-11-12
add support for babel routing protocol

christf:
this PR contains first support for a network as described in the l3 master plan.
this requires preliminary packages for l3roamd, babel (git version), mmfd which are all contained in the frankfurt main package repository on https://github.com/freifunk-ffm/packages

This PR contains init scripts for mmfd, l3roamd, babel, a firewall and the necessary integration of these components. 

This set now (2017-06-17) includes a proof of concept implementation of prefixd including a gui.

https://wiki.ffm.freifunk.net/firmware:babel

this replaces #732

PR #938, opened 2016-11-18
add zones having proto = gluon_mesh to firewall mesh-zone

christf:
as discussed on irc, this stunt allows for dynamically adding interfaces to firewall zones as they come up. This is needed when mesh-interfaces change.
This has been tested on a babel firmware.

### DIFF
--- a/package/gluon-core/files/lib/netifd/proto/gluon_mesh.sh
+++ b/package/gluon-core/files/lib/netifd/proto/gluon_mesh.sh
@@ -25,10 +25,14 @@ proto_gluon_mesh_setup() {
 
 	proto_init_update "$IFNAME" 1
 
-	proto_add_data
-	json_add_boolean fixed_mtu "$FIXED_MTU"
-	json_add_boolean transitive "$TRANSITIVE"
-	proto_close_data
+	if [ "$IFNAME" != "br-wan" ]
+	then
+		proto_add_data
+		json_add_string zone mesh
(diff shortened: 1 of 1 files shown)


NeoRaider:
Hmm. I wonder if this will break WAN firewalling when mesh-on-WAN is enabled.

NeoRaider:
Then again, Mesh-on-WAN will need redesign anyways if it is supposed to work with Babel.

It would be good to encapsulate Mesh-on-WAN/LAN using VXLANs, so WAN traffic can be properly handled and firewalled separately. Also, VXLANs would prevent accidential bridging of mesh domains.

christf:
so do you think kmod-vxlan should be part of this PR as well?

christf:
@rotanid does this hurt when applied to gluon running batman?

christf:
I have tested this change on a batman network.

Adding interfaces to non-existing firewall-zones yields no visible result, the calls will probably fail but I was not able to catch this in the terminal.
The transitive-flag which is set after the firewall zone actually makes it to the configuration.

There is no negative impact when running a current gluon with batman mesh protocol.

christf:
@NeoRaider so is this blocked then by the vxlan-extension of netifd?

christf:
Mentioning #934 to have this status shown over there

NeoRaider:
I've implemented the change a bit differently, but it's in the master now.

PR #942, opened 2016-11-20
docs/wired make uci commit more specific

viisauksena:
propose to make uci commit more specific to minimize unwanted changes in other temporary written config files like wireless ssid (in some offline changers)

Signed-off-by: fuzzle <gluon@viisauksena.de>


NeoRaider:
Thanks, merged.

One nitpick for other projects: we don't use Signed-offs in Gluon; but in projects that require a sign-off, you need to use your real name, as it is a legal statement (the Developer's Certificate of Origin, see http://developercertificate.org/ )

viisauksena:
wie mans macht .. :) ... les da nix von klarnamen

PR #944, opened 2016-11-25
docs: add escape() for pubkey

kb-light:


kb-light:
It seems, that this is not needed, because the fastd-key is alphanumeric.

PR #947, opened 2016-11-25
Add tunneldigger / l2tp support

lcb01a:
Hi,

this push request adds the openwrt/lede tunneldigger package and necessary gluon packages.

Package names:
gluon-config-mode-tunneldigger
gluon-mesh-vpn-tunneldigger
gluon-migrate-vpn
gluon-tunneldigger-watchdog
tunneldigger

Is this sufficient or do you guys think something might be missing?
Respondd support has not been added as we currently can only define whether tunneldigger is installed or not. It doesn't offer any statistics to be used.

Cheers
Cyrus


mweinelt:
Hi,

glad to see this.

I'm not sure @NeoRaider would be eager to maintain tunneldigger itself in gluon. Is there a chance we can find a maintainer that gets it into openwrt/packages instead?

lcb01a:
> I'm not sure @NeoRaider would be eager to maintain tunneldigger itself in gluon. Is there a chance we can find a maintainer that gets it into openwrt/packages instead?

Unfortunately this is up to the tunneldigger developers, so far they haven't done this. I didn't want to do this without their approval.

This can also be easily removed from gluon again when it gets added upstream.

mweinelt:
They maintain their own opkg repository, so there probably will be no incentive to get it upstream. Does it make sense to rely on their repository instead?

https://github.com/wlanslovenija/firmware-packages-opkg

lcb01a:
Of course this is also possible, but it would somehow make this push request senseless as you would have to include the foreign repository anyway every time you want to use these packages.

christf:
won't there still be a need for some gluon infrastructure if tunneldigger is used as mesh-vpn?
just thinking of what fastd and gluon-mesh-vpn-fastd is doing.

NeoRaider:
Including https://github.com/wlanslovenija/firmware-packages-opkg in our `modules` would be fine, so this PR would just contain the Gluon-specific parts, but it seems that the FFRL tunneldigger has some nice features that aren't supported by the upstream version...

So I guess our best option would be to add the tunneldigger package to https://github.com/freifunk-gluon/packages as long as it is not upstream, can you open a PR to do this?

Do you know if there has been any contact with the wlanslovenija maintainers, e.g. to get the usage broker upstream?

lcb01a:
Sounds like good Idea, I will prepare a PR.

mitar:
> Unfortunately this is up to the tunneldigger developers, so far they haven't done this. I didn't want to do this without their approval.
> They maintain their own opkg repository, so there probably will be no incentive to get it upstream. Does it make sense to rely on their repository instead?

Feel free to include Tunneldigger in OpenWrt/LEDE as well. The issue is that we never managed to get that to happen. If somebody has commit flag there and can maintain a copy of the package metadata there, that would be great. Feel free to just use package information from our feed.

> Do you know if there has been any contact with the wlanslovenija maintainers, e.g. to get the usage broker upstream?

Pull requests are welcome. I do not know why you have some commits [here](https://github.com/wlanslovenija/tunneldigger/compare/master...ffrl:master) you didn't push upstream. :-)

NeoRaider:
@mitar Getting tunneldigger into OpenWrt/LEDE is as simple as opening a PR at https://github.com/openwrt/packages . I can push the package there, but it would be great for someone from the wlanslovenija team (or, actually, anybody who is actively using tunneldigger) to step up as package maintainer, so the package is kept up-to-date.

PR #953, opened 2016-11-28
DOCS site: hint to change default texts in PO file

rubo77:


rubo77:
Ok.
Is there really "no way" to find out the order?

Maybe we just have to look through the lua code to find out, which translations from which files have privilege?

rubo77:
This would be such a helpful feature.

Maybe we can force somehow, that translations from the site. Conf have preference always.

Maybe it is the file timestamp, that produces the preference? Or the alphabetical filename, them we could suggest filenames for the po files with numbers

PR #958, opened 2016-12-01
config mode: Move Autoupdater at the end

rubo77:
### 1. 
The message is not that important, that it must be the first line. The first Block can be better used to explain the following Form. That Info-Block is in the way at the top to stay in the context.

At the bottom before the Submit Button it has a better Place and is evenly pregnant.

### 2.
If the Auto Updater is disabled, this PR adds a message 
> Automatic updates are disabled. They can be enabled in <em>Advanced Settings</em>

### 3.
Also I think the Menu-point "Advanced Settings" should be written with a capital S.


NeoRaider:
1. We've had enough discussion if an automatic updater is even acceptable at all for Freifunk. I think that makes this information important enough to keep the message at the top.
2. Hmm, I don't care about this. Would accept it.
3. I took these strings from Chromium, where it's written "Advanced settings". Let's keep it like this.

rubo77:
Sure, In a text flow "advanced settings" wout be lowercase, but I think it is quite common to capitalize all important words in a heading:

https://ux.stackexchange.com/questions/75454/are-there-guidelines-for-when-to-use-uppercase-capitalize-each-word-or-only-th

So is it ok, to put the autoupdater warning at the bottom?

rubo77:
I created a separate PR for 2.) #974

PR #971, opened 2016-12-08
Rename /lib/gluon/respondd/ to /lib/respondd/

jplitza:
respondd is supposed to be non-Gluon-specific (it is in the packages repository), thus having its data direectory in /lib/gluon/ doesn't make sense.

This is the prerequisite for putting non-Gluon respondd providers in the packages repo, as is planned in #826.

Also, maybe this should be mentioned in the release notes?


jplitza:
@NeoRaider indicated that he would rather have multiple directories passed to respondd, one for generic modules and one for Gluon-specific ones. That is being pursued in freifunk-gluon/packages#153, if that gets merged this can be closed.

NeoRaider:
freifunk-gluon/packages#153 has been merged.

PR #972, opened 2016-12-08
gluon-respondd: Use procd in initscript and to restart on network changes

jplitza:
procd's (trigger) reloads work as follows: When the reload is triggered (or also manually called), procd builds the command line, environment etc. to start the program. If they differ from the last/running invocation of the program, the program is killed and restarted with the new command line. This is exactly what we want!

Also, if a program is not currently running, the triggers aren't active. So respondd is restarted if and only if it is running and the interfaces relevant for the command line changed.

TBH, I didn't test this yet. But I will in the near future.


NeoRaider:
I found an issue with this change: at the moment, we unconditionally add the contents `/lib/gluon/respondd/client.dev` to the command line. This means that the command line doesn't change when `br-client` comes up, thus not restarting and joining the multicast group on this interface if it becomes available last.

I think it would make sense to use ifstatus on this interface (and thus change the actual interface name to the UCI identifier in `/lib/gluon/respondd/client.dev`) and use a similar jsonfilter expression as is used for the mesh interfaces.

jplitza:
Fixed in fcf3abd9c42a777804bd1f3512363abc95164f3b

PR #976, opened 2016-12-11
Switch from batman-adv v14/v15 debugfs to netlink

ecsv:
This is a follow up pull request of RFC #916 

----

The batadv debugfs requires large memory blocks to write the text debug tables. This is inefficient for large tables like the global translation table or the originators table.

The memory requirement can be reduced by using netlink. It copies smaller packets in a binary format to the userspace program. gluon can therefore parse larger originator tables without causing an OOM on systems which are tight on memory.

----

This pull request is not about providing a generic way to access batman-adv netlink. This could be done by an interested third party. Important for a library which basically only provides `batadv_nl_query_common` and `batadv_netlink_policy` would be:

 * batadv-netlink.c has to be integrated in a new/existing library
 * `batadv_nl_query_common` and `batadv_netlink_policy` has to be marked as `__attribute__ ((visibility ("default")))`
 * `batadv_netlink_policy` must be extended to include all batman-adv netlink attributes (it currently only contains what gluon-mesh-batman-adv-core/gluon-status-page-api needs)
    - the current list of attributes can be found at https://git.open-mesh.org/batctl.git/blob/HEAD:/netlink.c#l67
 * batman_adv.h and batadv-netlink.h has to be installed by `Build/InstallDev` to `$(1)/usr/include/`
 * batman_adv.h/batadv-netlink.* has to be removed from gluon-mesh-batman-adv-core/gluon-status-page-api
 * dependency to new/existing library has to be added to gluon-mesh-batman-adv-core/gluon-status-page-api
 *  `#include "batadv-netlink.h"` in gluon-mesh-batman-adv-core/gluon-status-page-api has to be changed to `#include <batadv-netlink.h>`.
 * `-l<name_of_new_library>` has to be added to gluon-status-page-api (for neighbours-batadv)/gluon-mesh-batman-adv-core

It might also be interesting to get in contact with @jjsarton because he currently has a similar ticket assigned to him on open-mesh.org (https://www.open-mesh.org/issues/302)

### DIFF
--- a/package/gluon-mesh-batman-adv-core/Makefile
+++ b/package/gluon-mesh-batman-adv-core/Makefile
@@ -13,7 +13,7 @@ define Package/gluon-mesh-batman-adv-core
   SECTION:=gluon
   CATEGORY:=Gluon
   TITLE:=Support for batman-adv meshing (core)
-  DEPENDS:=+gluon-core +libgluonutil +gluon-client-bridge +gluon-ebtables +firewall +libiwinfo +kmod-dummy +kmod-macvlan
+  DEPENDS:=+gluon-core +libgluonutil +gluon-client-bridge +gluon-ebtables +firewall +libiwinfo +kmod-dummy +kmod-macvlan +libnl-tiny
 endef
 
 define Build/Prepare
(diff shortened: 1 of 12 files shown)


NeoRaider:
I can't find any issues with this, everything is working as expected :)

PR #978, opened 2016-12-12
Added Gluon tunneldigger/L2TP related packages

lcb01a:
Hi,

Recreated the PR as there were issues with empty commits.

To answer a previous question: 
Tunneldigger is an alternative for Fastd without any encryption which uses Kernelspace L2TP and thus is way faster than Fastd.

gluon-config-mode-tunneldigger
gluon-mesh-vpn-tunneldigger
gluon-migrate-vpn
gluon-tunneldigger-watchdog

Cheers
Cyrus


package/gluon-migrate-vpn/files/lib/gluon/upgrade/499-migrate-vpn
```diff
@@ -0,0 +1,31 @@
...
+local util = require 'gluon.util'
+
+local uci = require('luci.model.uci').cursor()
+local lutil = require 'luci.util'
+
+local tunneldigger_enabled = uci:get_first('tunneldigger', 'broker', 'enabled')
+local fastd_enabled = uci:get('fastd', 'mesh_vpn', 'enabled')
+
+local tunneldigger_installed = util.exec('sh' , '-c', 'opkg list-installed | grep -e \'^tunneldigger\'')
+local fastd_installed = util.exec('sh' , '-c', 'opkg list-installed | grep -e \'^fastd\'')
```
NeoRaider: This will break on systems without opkg (for example the *ar71xx-tiny* target of the Gluon *lede* branch). I think testing for the existence of `/usr/bin/fastd` and `/usr/bin/tunneldigger` would be more robust.

package/gluon-mesh-vpn-tunneldigger/files/lib/gluon/upgrade/400-mesh-vpn-tunneldigger
```diff
@@ -0,0 +1,61 @@
...
+	  }
+)
+
+uci:save('tunneldigger')
+uci:commit('tunneldigger')
+
+uci:section('network', 'interface', 'mesh_vpn',
+	  {
+		  ifname = 'mesh-vpn',
+		  proto = 'batadv',
```
NeoRaider: This is outdated, see https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-mesh-vpn-fastd/luasrc/lib/gluon/upgrade/400-mesh-vpn-fastd#L124 for the new mesh protocol agnostic configuration using proto *gluon_mesh*.

package/gluon-mesh-vpn-tunneldigger/files/lib/gluon/upgrade/400-mesh-vpn-tunneldigger
```diff
@@ -0,0 +1,61 @@
...
+		  mesh = 'bat0',
+		  mesh_no_rebroadcast = 1,
+		  mtu = site.tunneldigger_mesh_vpn.mtu,
```
NeoRaider: How is the MAC address assigned for the tunneldigger interface?

As it can't be enabled at the same time as the fastd mesh VPN (as both use the interface name `mesh-vpn`), it could use the same MAC address, I think.

package/gluon-tunneldigger-watchdog/files/usr/bin/tunneldigger-watchdog
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+PIDFILE=/var/run/tunneldigger.mesh-vpn.pid
+
+if [ "$(uci get tunneldigger.@broker[0].enabled)" == "1" ]; then
```
NeoRaider: `==` is invalid in `[]` expressions (but accepted by many implementations, sometimes with a warning), this should be `=`.

package/gluon-tunneldigger-watchdog/files/usr/bin/tunneldigger-watchdog
```diff
@@ -0,0 +1,12 @@
...
+    if [ "$(pgrep tunneldigger | head -n 1)" != "$(cat $PIDFILE)" ]; then
+        /etc/init.d/tunneldigger restart
+        logger -t tunneldiger-watchdog "Daemon not running, restarted tunneldigger."
+    elif [ "$(batctl o |grep mesh-vpn |wc -l)" == "0" ]; then
```
NeoRaider: Quite ugly, I wonder if we can find a nicer solution...

package/gluon-mesh-vpn-tunneldigger/files/lib/gluon/upgrade/400-mesh-vpn-tunneldigger
```diff
@@ -0,0 +1,61 @@
...
+if (tonumber(enabled) == 0) then
+  enabled = site.tunneldigger_mesh_vpn.enabled and 1 or 0
+end
+
+-- Delete old broker config section       
+uci:delete_all('tunneldigger', 'broker')
+
+section = uci:add('tunneldigger', 'broker')
+
+uci:section('tunneldigger', 'broker', section,
```
NeoRaider: It would be nicer to use a static section name like `mesh_vpn` for the broker instead of referring to it using `@broker[0]`, so this can't collide with other tunneldigger configurations (that is, if the tunneldigger init script can handle multiple instances, which I didn't check)

package/gluon-tunneldigger-watchdog/files/usr/bin/tunneldigger-watchdog
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+PIDFILE=/var/run/tunneldigger.mesh-vpn.pid
+
+if [ "$(uci get tunneldigger.@broker[0].enabled)" == "1" ]; then
+    if [ "$(pgrep tunneldigger | head -n 1)" != "$(cat $PIDFILE)" ]; then
```
NeoRaider: I guess this could be `pidof tunneldigger`? Again, this should probably learn to cope with multiple instances (but this can be done later).

package/gluon-mesh-vpn-tunneldigger/files/lib/gluon/upgrade/400-mesh-vpn-tunneldigger
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/lua
```
NeoRaider: All Lua scripts should be moved from `files` to `luasrc`, and `GluonSrcDiet` (and luadest CP) calls be added to the corresponding Makefiles, as seen in most Gluon packages.

package/gluon-mesh-vpn-tunneldigger/files/lib/gluon/upgrade/400-mesh-vpn-tunneldigger
lcb01a: The init script only supports one instance at the time, I guess we could add the support later on but at the time I didn't want to derive too far from the original package.

package/gluon-tunneldigger-watchdog/files/usr/bin/tunneldigger-watchdog
lcb01a: I commented on a few items, I will cover the rest as soon as possible :)

lcb01a:
I commented on a few items, I will fix the rest of the issues as soon as possible.

lcb01a:
I've updated the commit to address some of the issues. 
I guess the most important issue left is regarding the new way to configure the interface.

jplitza:
I probably missed something, but why do we need a watchdog? Is Tunneldigger that unstable?

lcb01a:
@jplitza 
The watchdog is there just in case the daemon dies, its an optional package.

Regarding the remaing issues:
- Ugly watchdog: I could remove the package for now and people who need this can still grab it from the ffrl / düsseldorf repositories
- Use the mesh protocol agnostic  "gluon_mesh" in uci network config: This is not possible at this moment as the gluon_mesh protocol does not support to set the MTU of the unterlying interface.
e.g. setting mtu = "1360" is completely ignored, this should be added to the gluon_mesh protocol as tunneldigger is not responsible for the initial MTU configuration and we don't use the path MTU feature of tunneldigger (It doesn't work well with batman).

corny:
@lcb01a Just use procd, which is able to restart a died process.

lcb01a:
@corny 
Thanks for the suggestion, this looks promising :)

lcb01a:
Hi,

The PR has been updated:

- Removed tunneldigger-watchdog, this can still be included via repo if needed
- gluon-mesh-von-tunneldigger now uses the gluon_mesh proto correctly

I guess this should address all issues that have emerged.

lcb01a:
Hi, any update on this? :)

package/gluon-config-mode-tunneldigger/luasrc/lib/gluon/config-mode/wizard/0300-mesh-vpn.lua
```diff
@@ -0,0 +1,64 @@
...
+    if data._limit_egress ~= nil then
+      uci:set("simple-tc", "mesh_vpn", "limit_egress", data._limit_egress)
+    end
+
+    uci:save("simple-tc")
+    uci:commit("simple-tc")
+  end
+end
+
+return M
```
NeoRaider: This file will need to be updated for gluon-web. Doing so should be straightforward, as it is very similar to the fastd-mesh-vpn one.

package/gluon-config-mode-tunneldigger/luasrc/lib/gluon/config-mode/wizard/0300-mesh-vpn.lua
```diff
@@ -0,0 +1,64 @@
+local cbi = require "luci.cbi"
+local i18n = require "luci.i18n"
+local uci = require("simple-uci").cursor()
+
+local M = {}
+
+function M.section(form)
+  local msg = i18n.translate('Your internet connection can be used to establish a ' ..
+                             'L2TP VPN connection with other nodes. ' ..
```
NeoRaider: I'm not sure if we want to explicitly mention "L2TP" here, as we're trying to avoid technical terms in the wizard. Unfortunately, I have no idea how to distinguish this from the fastd mesh VPN otherwise...

package/gluon-mesh-vpn-tunneldigger/Makefile
```diff
@@ -0,0 +1,42 @@
...
+
+PKG_BUILD_DIR := $(BUILD_DIR)/$(PKG_NAME)
+
+include ../gluon.mk
+
+define Package/gluon-mesh-vpn-tunneldigger
+  SECTION:=gluon
+  CATEGORY:=Gluon
+  TITLE:=Support for connecting batman-adv meshes via tunneltigger/l2tpv3 pseudowire
+  DEPENDS:=+gluon-core gluon-mesh-batman-adv +gluon-wan-dnsmasq +tunneldigger +iptables-mod-extra +simple-tc
```
NeoRaider: Remove gluon-mesh-batman-adv from the depends, so the package is usable with Babel as well (as soon as we merge that).

package/gluon-mesh-vpn-tunneldigger/check_site.lua
```diff
@@ -0,0 +1,9 @@
+need_number('tunneldigger_mesh_vpn.mtu')
+need_boolean('tunneldigger_mesh_vpn.enabled', false)
+need_string_array('tunneldigger_mesh_vpn.brokers')
+
+if need_table('tunneldigger_mesh_vpn.bandwidth_limit', nil, false) then
+  need_boolean('tunneldigger_mesh_vpn.bandwidth_limit.enabled', false)
+  need_number('tunneldigger_mesh_vpn.bandwidth_limit.ingress', false)
+  need_number('tunneldigger_mesh_vpn.bandwidth_limit.egress', false)
```
NeoRaider: Please always use tabs for indentation (in the other files as well).

package/gluon-mesh-vpn-tunneldigger/luasrc/lib/gluon/upgrade/400-mesh-vpn-tunneldigger
```diff
@@ -0,0 +1,61 @@
...
+if (tonumber(enabled) == 0) then
+  enabled = site.tunneldigger_mesh_vpn.enabled and 1 or 0
+end
+
+-- Delete old broker config section       
+uci:delete_all('tunneldigger', 'broker')
+
+section = uci:add('tunneldigger', 'broker')
+
+uci:section('tunneldigger', 'broker', section,
```
NeoRaider: It would be nicer to just explicitly name the section instead of using an anonymous section. You could just 
```lua
uci:delete('tunneldigger', 'mesh_vpn')
uci:section('tunneldigger', 'broker', 'mesh_vpn', ...)
```
instead of deleting all broker sections. Instead of `uci:get_first` you could use `uci:get_bool`, like we to in other scripts.

For migration from an older configuration using an anonymous section, you could use:
```lua
if not uci:get('tunneldigger', 'mesh_vpn') then
	uci:delete_all('tunneldigger', 'broker')
end
```

package/gluon-mesh-vpn-tunneldigger/luasrc/lib/gluon/upgrade/400-mesh-vpn-tunneldigger
```diff
@@ -0,0 +1,61 @@
...
+local uci = require('simple-uci').cursor()
+
+
+-- Group for iptables rule
+users.add_group('gluon-tunneldigger', 900)
+
+local enabled = uci:get_first('tunneldigger', 'broker', 'enabled')
+if (tonumber(enabled) == 0) then
+  enabled = site.tunneldigger_mesh_vpn.enabled and 1 or 0
+end
```
NeoRaider: This condition looks wrong. Together with the changes below, I'd probably write something like:
```lua
local enabled = uci:get('tunneldigger', 'mesh_vpn', 'enabled')
if enabled == nil then
	enabled = uci:get_first('tunneldigger', 'broker', 'enabled')
end
if enabled == nil then
	enabled = site.tunneldigger_mesh_vpn.enabled or false
end
```
Note that the new *uci-simple* module can properly handle boolean values in set/section commands, you just need to make sure that the value is not *nil* (which is done here with the `or false`).

package/gluon-mesh-vpn-tunneldigger/luasrc/lib/gluon/upgrade/400-mesh-vpn-tunneldigger
```diff
@@ -0,0 +1,61 @@
...
+
+uci:section('network', 'interface', 'mesh_vpn',
+	  {
+		  ifname = 'mesh-vpn',
+		  proto = 'gluon_mesh',
+		  fixed_mtu = '1',
+		  transitive = '1',
+		  mtu = site.tunneldigger_mesh_vpn.mtu,
+	  }
+)
```
NeoRaider: The indentation here uses tabs and spaces mixed. For *uci:section* calls, I prefer indentation like this:
```lua
	-- lots of values...
})
```

package/gluon-mesh-vpn-tunneldigger/luasrc/lib/gluon/upgrade/400-mesh-vpn-tunneldigger
```diff
@@ -0,0 +1,61 @@
...
+uci:section('firewall', 'include', 'mesh_vpn_dns',
+	  {
+	    type = 'restore',
+	    path = '/lib/gluon/mesh-vpn-tunneldigger/iptables.rules',
+	    family = 'ipv4',
+	  }
+)
+
+uci:save('firewall')
+uci:commit('firewall')
```
NeoRaider: Never commit UCI config in upgrade scripts, the last upgrade script will commit everything.

package/gluon-mesh-vpn-tunneldigger/luasrc/lib/gluon/upgrade/420-mesh-vpn-tunneldigger-simple-tc
```diff
@@ -0,0 +1,32 @@
...
+
+if fs.access('/etc/config/gluon-simple-tc') then
+  os.rename('/etc/config/gluon-simple-tc', '/etc/config/simple-tc')
+end
+
+
+if not uci:get('simple-tc', 'mesh_vpn') then
+  local config = {
+    ifname = 'mesh-vpn',
+    enabled = 0,
```
NeoRaider: These 0 and 1 can be changed to *false* and *true* as well with *simple-uci*.

package/gluon-migrate-vpn/luasrc/lib/gluon/upgrade/499-migrate-vpn
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/lua
+
+local site = require 'gluon.site_config'
+local users = require 'gluon.users'
+local util = require 'gluon.util'
+
+local uci = require("simple-uci").cursor()
+
+
+function file_exists(name)
```
NeoRaider: Please use *nixio.fs.access* as in the other files instead of defining your own function.

package/gluon-migrate-vpn/luasrc/lib/gluon/upgrade/499-migrate-vpn
```diff
@@ -0,0 +1,33 @@
...
+
+if (enabled == 1 and file_exists("/usr/bin/tunneldigger") and not file_exists("/usr/bin/fastd")) then
+  uci:set("tunneldigger", uci:get_first("tunneldigger", "broker"), "enabled", enabled)
+  uci:save("tunneldigger")
+  uci:commit("tunneldigger")
+elseif (enabled == 1 and not file_exists("/usr/bin/tunneldigger") and file_exists("/usr/bin/fastd")) then
+  uci:set("fastd", "mesh_vpn", "enabled", enabled)
+  uci:save("fastd")
+  uci:commit("fastd")
+end
```
NeoRaider: Again, you can change *enabled* to use booleans, and don't commit UCI here.

package/gluon-migrate-vpn/luasrc/lib/gluon/upgrade/499-migrate-vpn
```diff
@@ -0,0 +1,33 @@
...
+
```
NeoRaider: One empty line too much here.

package/gluon-config-mode-tunneldigger/luasrc/lib/gluon/config-mode/wizard/0300-mesh-vpn.lua
lcb01a: I guess everyone in the freifunk community knows it by the term L2TP so sticking with it for now should be the best way. We could also use the daemon name instead but I don't know if that is any more meaningful

rotanid: i'm not sure whether there will be two vpn options or if it's even possible to build with both (fastd/l2tp)
if not, we gain nothing from mentioning the technical term L2TP (or fastd)

lcb01a: thats true, but I rather not use the term "unencrypted" VPN since people falsely assume this means they are not safe from the receiving a law suit.
I can leave out the term completely and just call it VPN :)

lcb01a:
Hi

The issues you mentioned have been adressed and verified by me. Only the terminology for the config mode isn't clear (Whether to use L2TP in the text or not). 

Cheers

NeoRaider:
As we don't plan to support fastd and L2TP VPN in the same image (yet?), I think it's preferable to keep the texts the same and not to mention L2TP explcitly.

lcb01a:
I removed the mentioning of L2TP from the config mode package and updated the commit :)

NeoRaider:
Thanks, I've just merged your changes.

I ended up doing some refactoring to decrease duplication in the fastd and tunneldigger mesh VPN packages, and added the new site.conf options to the docs. I integerated the *migrate-vpn* script into a new *gluon-mesh-vpn-core* package and added support for tunneldigger to the existing *gluon-config-mode-mesh-vpn* package after adding your translation updates.

It would be great if you could check that I didn't break anything when refactoring.

PR #979, opened 2016-12-12
mvebu: add missing package & correct wrong comment

belzebub40k:
This PR adds the missing package uboot-envtools which is needed for a successful sysupgrade.
It also corrects a wrong comment in targets.mk.


NeoRaider:
Thanks, I've applied the uboot-envtools commit and fixed the comment myself to make it even clearer.

PR #962, opened 2016-12-04
Add support for ZBT-WG3526-32M

jplitza:
These are (almost) the devices that @corny has shipped in from China.

While those have 32 MB Flash, the image only works with 16 MB currently. They have 512 MB RAM and support 802.11ac.

The image currently has a [problem with rebooting because of the flash chip](http://lists.infradead.org/pipermail/linux-mtd/2016-November/070368.html), thus it would even be broken if the target wouldn't be as a whole. Nonetheless, I would like to challenge the note by @NeoRaider left in `targets/targets.mk`:
```
$(eval $(call GluonTarget,ramips,mt7621)) # BROKEN: No AP+IBSS support, 11s has high packet loss
```
I do currently operate the device in AP+IBSS mode. Thus, this either seems to be fixed or a restriction of the only other device in that subtarget (d-link-dir-860l-b1). However, throughput through IBSS is quite low (11 MBit/s net).

Another caveat is that seemingly random, there are stack traces like these in the log:
```
[ 2739.120000] ------------[ cut here ]------------
[ 2739.120000] WARNING: CPU: 0 PID: 0 at /home/jplitza/Code/freifunk/site/gluon/build/ramips-mt7621/openwrt/build_dir/target-mipsel_1004kc+dsp_uClibc-0.9.33.2_gluon-ramips-mt7621/linux-ramips_mt7621/mt76-2016-08-25/mt7603_mac.c:403 mt7603_mac_fill_rx+0x2c4/0x3cc [mt7603e]()
[...]

PR #981, opened 2016-12-17
ramips: Add patches to support devices with > 128Mib flash

corny:
Patch extracted from:
- http://lists.infradead.org/pipermail/linux-mtd/2016-December/070890.html
- http://lists.infradead.org/pipermail/linux-mtd/2016-December/070891.html

### DIFF
--- a/patches/openwrt/0086-mtd-patch.patch
+++ b/patches/openwrt/0086-mtd-patch.patch
@@ -1,0 +1,303 @@
+From: Julian Kornberger <jk+github@digineo.de>
+Date: Sat, 17 Dec 2016 15:56:53 +0100
+Subject: mtd patch
+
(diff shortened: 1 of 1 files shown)


NeoRaider:
These patches aren't applicable. You're trying to apply kernel patches to the OpenWrt repository.

As explained in our docs, patches in Gluon are to be created by making commits in the sub-repositories like `/openwrt` and then running `make update-patches` to convert commits into patch files. In this case, you'll need to create a commit that adds the patches to `target/linux/ramips/patches-3.18` or similar and make a patch from from this commit.

corny:
Thanks for the hint, I updated the PR.

corny:
I moved this patch to #962.

Issue #945, opened 2016-11-25
status-page: preparing for babel


lemoer:
Current working branch: https://github.com/lemoer/gluon/tree/rebase_statuspage

TODOs
----

**gluon-mesh-babel:**

- [x] adapt a stub makefile out of [the babel pr](https://github.com/freifunk-ffm/gluon/blob/98ee489ce9a7e22d8c4cc27c1af5cba054aa195a/package/gluon-mesh-babel/Makefile)
- [x] create a respondd module
- [x] respondd-module: provide nodeinfo.network.mac
- [x] respondd-module: provide nodeinfo.network.mesh
- [x] respondd-module: provide nodeinfo.network.addresses (adresses on ```br-client```, see question 4)
- [x] create an issue to extend the respondd module for gluon-mesh-babel (see notice 1)

**status-page-api:**

- [x] add "protocol" field to /cgi-bin/dyn/neighbours-batadv
- [x] provide information about babel neighbours in /cgi-bin/dyn/routing-metrics
- [x] rename to /cgi-bin/dyn/routing-metrics
- [x] legacy symlink to /cgi-bin/dyn/neighbours-batadv
- [x] port /cgi-bin/dyn/{stations,neighbours-nodeinfo} to lua (adapt work done in #745)
- [x] use generic function to get mesh ifaces in /cgi-bin/interfaces (see question 2)
- [x] use generic function to get mesh ifaces in /cgi-bin/dyn/stations?iface (see question 2)
- [x] use generic function to get mesh ifaces in /cgi-bin/dyn/neighbours-nodeinfo?iface (see question 2)

**status-page:**

- [x] rename /cgi-bin/dyn/neighbours-batadv to /cgi-bin/dyn/routing-metrics (see [here](https://github.com/lemoer/gluon/commit/d6933def68a3eac6d8fb5248d4be338bba30182f))
- [x] differentiate between batman and babel metrics and show them differently (see [here](https://github.com/lemoer/gluon/commit/e6334488ba34e23efee7f03b27efcb174db410ec))

**general:**

- [x] test in babel
- [x] test in batman-adv
- [ ] docs: add documentation about gluon-status-page (see [here](https://github.com/lemoer/gluon/commit/39b8a5cf52006184b83d3d264374c156798dd85e))

Notices:
--------

1. these resources remain unimplemented in the gluon-mesh-babel respondd provider:
    - nodeinfo.software.babeld
    - statistics.clients
    - statistics.traffic
[...]

PR #990, opened 2016-12-21
gluon-core: add function to gluon.util that returns all active mesh-devices

christf:
this can then be used by the statuspage et al.

### DIFF
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
@@ -1,25 +1,25 @@
 -- Writes all lines from the file input to the file output except those starting with prefix
 -- Doesn't close the output file, but returns the file object
 local function do_filter_prefix(input, output, prefix)
-  local f = io.open(output, 'w+')
-  local l = prefix:len()
+	local f = io.open(output, 'w+')
+	local l = prefix:len()
 
-  for line in io.lines(input) do
-    if line:sub(1, l) ~= prefix then
-      f:write(line, '\n')
-    end
-  end
+	for line in io.lines(input) do
+		if line:sub(1, l) ~= prefix then
+			f:write(line, '\n')
+		end
+	end
 
-  return f
+	return f
 end
 
 
 local function escape_args(ret, arg0, ...)
-  if not arg0 then
-    return ret
-  end
+	if not arg0 then
+		return ret
+	end
 
-  return escape_args(ret .. "'" .. string.gsub(arg0, "'", "'\\''") .. "' ", ...)
+	return escape_args(ret .. "'" .. string.gsub(arg0, "'", "'\\''") .. "' ", ...)
 end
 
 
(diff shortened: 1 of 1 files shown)


package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -37,80 +37,95 @@ local site = require 'gluon.site_config'
...
+	return string.gsub(sysconfig.primary_mac, ':', '')
 end
 
+function find_mesh_devices()
+	local ubus = require 'ubus'
+	local uconn = ubus.connect()
+	if not uconn then
+	    error("Failed to connect to ubusd")
+	end
+	local dump =uconn:call("network.interface", "dump", {})
```
mweinelt: missing whitespace after `=`

christf:
@lemoer fyi

rotanid:
most of the changes in this PR only change the indention of some lines, should this really be done in this PR? a separate one would be better i guess.

christf:
I was not going to create 2 PRs for the same file at the same time to avoid conflicts.

NeoRaider:
Doing this in the same PR is fine as long as the changes are separate commits. The commits will need better titles though (package prefix etc.)

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -73,6 +74,20 @@ function node_id()
...
 end
 
+function find_mesh_devices()
+	local uconn = ubus.connect()
+	if not uconn then
+	    error("Failed to connect to ubusd")
+	end
+	local dump = uconn:call("network.interface", "dump", {})
+	local devices = {}
+	for index, interface in pairs(dump.interface) do
```
NeoRaider: `ipairs` should be used instead of pairs for integer-indexed tables ("arrays"). It is a convention in our code to call the index variable `_` when it is not used.

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -73,6 +74,20 @@ function node_id()
 	return string.gsub(sysconfig.primary_mac, ':', '')
 end
 
+function find_mesh_devices()
```
NeoRaider: Please call this `get_mesh_devices()`, as it doesn't search for some specific entry like the `find_*` functions in this file do.

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -73,6 +74,20 @@ function node_id()
...
+	local uconn = ubus.connect()
+	if not uconn then
+	    error("Failed to connect to ubusd")
+	end
+	local dump = uconn:call("network.interface", "dump", {})
+	local devices = {}
+	for index, interface in pairs(dump.interface) do
+	    if ( (interface.proto == "gluon_mesh") and (interface.up == true)) then
```
NeoRaider: You can just write `interface.up` instead of `(interface.up == true)`.

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -73,6 +74,20 @@ function node_id()
...
+			table.insert(devices, interface.device)
+	    end
+	end
+	return(devices)
```
NeoRaider: As in C, it is more common to write this as
```lua
return devices
```

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -73,6 +74,20 @@ function node_id()
 	return string.gsub(sysconfig.primary_mac, ':', '')
 end
 
+function find_mesh_devices()
+	local uconn = ubus.connect()
+	if not uconn then
+	    error("Failed to connect to ubusd")
+	end
```
NeoRaider: I'd prefer the connection to be provided as an argument instead of creating a new one for each call.

christf:
@NeoRaider danke für das feedback, es ist umgesetzt.

christf:
referencing #945 so the status of this PR is visible over there....

PR #991, opened 2016-12-21
ar71xx-generic: backport support for TP-LINK TL-WR1043ND v4

### LINKED ISSUE #961: Add support for TL-WR1043nd-v4
One of our community member told me that he bought a TL-WR1043nd-v4.
But the v4 is not supported by gluon and, as I see, not by OpenWRT.

So you can add support for the TL-WR1043nd-v4?

NeoRaider:
Fixes #961


rotanid:
maybe also add the device to the hardware list with this PR

NeoRaider:
@rotanid Good point, updated.

PR #996, opened 2016-12-28
add basic danish support in Luci

rubo77:
so danish language files can be added in gluon


rubo77:
This patch is needed so we can start translating gluon to danish. Without the basic danish support, it is not possible to add daish .po files.

mweinelt:
Isn't this something that is more appropriately tracked upstream? I don't see the need and commitment do maintain danish translations downstream.

https://github.com/openwrt/luci

rubo77:
Some people up here in the North would like to start translate but it is not possible without the Danish enabled in Gluon

rubo77:
And Upstream would be very much work, although we don't need anything of it in gluing.

We need only the base from this patch

NeoRaider:
We're currently removing LuCI from Gluon, obsoleting this patch.

PR #997, opened 2016-12-29
Hoodselector

jplitza:
(copy&pasted from @2tata's mail thread) 

This patch set contains the hoodselector and a hoodfile as exsample.

The hoodselector is a software that creates decentralized, semi automated ISO OSI layer 2 network segmentation for batman-adv layer 2 routing networks. This program reads the geobased sub-networks called hoods from the above mentioned hoodfile. The decision of choosing the right hood is made on following points: first, the hoodselector checks, if the router has a VPN connection. If it has, the hoodselector then checks, if a geoposition was set on the router. Knowing the position of the router the hoodselector can find the right hood, because each hood is defined with geocoordinates. If the Router doesn't have a VPN connection e.g. as a mesh only router, the hoodselector triggers a WIFI scan and searches for neighboured mesh routers in other hoods. If there is an other router with a different BSSID but with the same mesh SSID, the router chooses it’s hood based on the neighboured BSSID. Open issues can be found [here](https://git.nordwest.freifunk.net/ffnw-firmware/packages/issues?label_name%5B%5D=hoodselector)

### DIFF
--- a/package/ffnw-hoods/files/lib/ffnw/hoods/hoods.json
+++ b/package/ffnw-hoods/files/lib/ffnw/hoods/hoods.json
@@ -1,0 +1,64 @@
+[
+  {
+    "name": "default",
+    "bssid": "02:CA:FF:AA:BA:BF",
+    "defaulthood": true,
+    "servers": [
+      {
+        "host": "default02.entenhausen.de",
+        "port": "11111",
+        "publickey": "32a02dd5e6b9454734tt3r3ad30bcf269564546ewtrew5442533tgr258cb"
+      },
+      {
+        "host": "default03.entenhausen.de",
+        "port": "11111",
+        "publickey": "973732543523frea17cfa53d2b9e025202trwrtefacf9503737tf5a23204"
+      },
+      {
+        "host": "default04.entenhausen.de",
+        "port": "11111",
+        "publickey": "432534t2ewe9a10620d09402ca2frewfwdf364e6fwqfer4re669a2eb5e6f"
+      },
+      {
+        "host": "default05.entenhausen.de",
+        "port": "11111",
+        "publickey": "34rfewf4754tg3f54538236bae815ef2r4245f20c63aeregt452b183045d"
+      },
+      {
+        "host": "default06.entenhausen.de",
+        "port": "11111",
+        "publickey": "2245t42gvre0ff93a0rewferewfrewb00dd215arefewf4355e5eb16f0b69"
+      }
+    ],
+    "boxes": []
+  },
+  {
+    "name": "quackenbruek",
+    "bssid": "02:AA:0A:12:80:43",
+    "defaulthood": false,
+    "servers": [
+      {
+        "host": "quackenbruek01.sn.entenhausen.de",
+        "port": "10000",
+        "publickey": "c6485dsaqdwe33dbebdbweqrfqca4aewfc3453453224532434fr4f3d629d"
+      },
+      {
+        "host": "quackenbruek02.sn.entenhausen.de",
+        "port": "10001",
+        "publickey": "887c6c4e1ff435t4t42gt4aac03cd9494bd13d1221ea6cewqdq3431r4194"
+      }
+    ],
+    "boxes": [
+      [
+        [
+          52.18,
+          7.41
+        ],
+        [
+          52.35,
+          7.9
+        ]
+      ]
+    ]
+  }
+]
--- a/package/ffnw-hoodselector/files/usr/lib/micron.d/hoodselector
+++ b/package/ffnw-hoodselector/files/usr/lib/micron.d/hoodselector
@@ -1,0 +1,1 @@
+*/2 * * * * /usr/sbin/hoodselector
--- a/package/ffnw-hoodselector/luasrc/hoodselector
+++ b/package/ffnw-hoodselector/luasrc/hoodselector
@@ -1,0 +1,722 @@
+#!/usr/bin/lua
+
+-- This is the hoodselector. The hoodselector is one of the main components for
+-- splitting a layer 2 mesh network into seperated network segments (hoods).
+-- The job of the hoodselector is to automatically detect in which hood
+-- the router is located based on geo settings or by scanning its environment.
+-- Based on these informations the hoodselector should select a hood from a
+-- list of known hoods (hoodlist) and adjust vpn, wireless and mesh on lan
+-- configuration based on the settings given for the selected hood.
+--
+-- The hoodlist containing all hood settings is located in a seperate hoodfile
+-- in the hoods package.
+--
+-- The hoodselector depends on the folowing additional software:
+--  * fastd (vpn configuration) see getCurrentPeers(), setHoodVPN()
+--  * iw (wireless network scanning) see getNeigbourBssid()
+--  * batman-adv (mesh protocol) see directVPN(), getGwRange()
+--  * respondd (molwm) see molwm()
+--
+--  To detect the current hood the hoodselector knows 2 modes containing
+--  * 1. Default mode (VPN Router)
+--    - set real hood dependent on geo position.
+--    - set default hood dependent on geo position.
+--  * 2. Scan modes
+--   - Set wifi conf on scanned BSSID
+--   - Set vpn conf getting by BSSID (if no VPN conf exsist disable fastd)
+-- When selecting a hood, the hoodselector has the following priorities:
+--   1. Selecting a hood by geo position depending on direct VPN connection.
+--   2. force creating one mesh cloud with neigbour mesh routers
+--   3. if routers had only mesh setting vpn config depends on the BSSID
+--
+-- Resources
+--  * https://wireless.wiki.kernel.org/en/users/documentation/iw
+
+-- MOLWM respondd file
+local molwmFile="/tmp/.hoodselector"
+
+local molwmtable = {}
+molwmtable["md5hash"] = ""
+molwmtable["vpnrouter"] = ""
+molwmtable["hoodname"] = ""
+
+-- PID file to ensure the hoodselector isn't running parallel
+local pidPath="/var/run/hoodselector.pid"
+
+if io.open(pidPath, "r") ~=nil then
+  io.stderr:write("The hoodselector is still running.\n")
+  os.exit(1)
+else
+  io.close(io.open(pidPath, "w"))
+end
+
+local json = require ("luci.jsonc")
+local uci = require('luci.model.uci').cursor()
+local file = '/lib/ffnw/hoods/hoods.json'
+-- initialization done
+
+-- Read the full hoodfile. Return nil for wrong format or no such file
+local function readHoodfile(file)
+  local jhood = io.open(file, 'r')
+  if not jhood then return nil end
+  local obj, pos, err = json.parse (jhood:read('*a'), 1, nil)
+  if err then
+    return nil
+  else
+    return obj
+  end
+end
+
+local function mesh_on_wan_disable()
+  os.execute('ifdown mesh_wan')
+  io.stderr:write('Interface mesh_wan disabled.\n')
+end
+
+local function mesh_on_wan_enable()
+  os.execute('ifup mesh_wan')
+  io.stderr:write('Interface mesh_wan enabled.\n')
+end
+
+local function mesh_on_lan_disable()
+  os.execute('ifdown mesh_lan')
+  io.stderr:write('Interface mesh_lan disabled.\n')
+end
+
+local function mesh_on_lan_enable()
+  os.execute('ifup mesh_lan')
+  io.stderr:write('Interface mesh_lan enabled.\n')
+end
+
+local function molwm()
+  local mesh_en = true
+  local respondd = string.format("gluon-neighbour-info -i bat0 -p 1001 -d ff02::2 -r hoodselector -t 0.5")
+  for line in io.popen(respondd, 'r'):lines() do
+    local obj, pos, err = json.parse (line, 1, nil)
+    if err then
+      io.stderr:write("json parse error!\n")
+      mesh_en = false
+      break
+    else
+      if obj["hoodinfo"] ~= nil then
+        if not ( obj["hoodinfo"]["md5hash"] == molwmtable["md5hash"]:gsub('\"', '') ) then
+          io.stderr:write("hashes are not equals!\n")
+	  mesh_en = false
+	  break
+        end
+      end
+    end
+  end
+  if uci:get('network', 'mesh_wan') and not mesh_en then
+    mesh_on_wan_disable()
+  end
+  if uci:get('network', 'mesh_lan') and not mesh_en then
+    mesh_on_lan_disable()
+  end
+  if uci:get('network', 'mesh_wan') and mesh_en then
+    mesh_on_wan_enable()
+  end
+  if uci:get('network', 'mesh_lan') and mesh_en then
+    mesh_on_lan_enable()
+  end
+end
+
+-- Create md5 hash from currend hood
+local function molwm_md5hash(hood)
+  local file = io.open("/tmp/.hoodhash", "w")
+  if not file then
+    io.stderr:write('\"/tmp/.hoodhash\" can not created\n')
+  else
+    file:write(json.stringify(hood, { indent = true }))
+    file:close()
+    --part to create md5 hash of this file
+    for line in io.popen(string.format( "md5sum /tmp/.hoodhash")):lines() do
+      for i in string.gmatch(line, "%S+") do
+	if (string.len(i) == 32) then
+	  molwmtable["md5hash"] = "\"" .. string.format(i) .. "\""
+	  break
+	end
+      end
+    end
+    os.remove("/tmp/.hoodhash")
+  end
+end
+
+-- Write MOLWM content into file
+local function write_molwm(hood)
+  if hood ~= nil then
+    molwm_md5hash(hood)
+    molwmtable["hoodname"] = "\"" .. hood["name"] .. "\""
+  end
+  molwm()
+  local file = io.open(molwmFile, "w")
+  if not file then
+    io.stderr:write(molwmFile ..' not found or not createble!\n')
+  else
+      file:write("\"md5hash\": " .. molwmtable["md5hash"] .. "\n")
+      file:write("\"vpnrouter\": " .. molwmtable["vpnrouter"] .. "\n")
+      file:write("\"hoodname\": " .. molwmtable["hoodname"] .. "\n")
+    file:close()
+  end
+end
+
+-- Program terminating function including removing of PID file
+local function exit()
+  if io.open(pidPath, "r") ~=nil then
+    os.remove(pidPath)
+  end
+  os.exit(0)
+end
+
+local function trim(s)
+  -- from PiL2 20.4
+  return (s:gsub("^%s*(.-)%s*$", "%1"))
+end
+
+local function sleep(n)
+  os.execute("sleep " .. tonumber(n))
+end
+
+local function brclient_restart()
+  os.execute('ifconfig br-client down')
+  os.execute('ifconfig br-client up')
+  io.stderr:write('Interface br-client restarted.\n')
+end
+
+local function vpn_stop()
+  os.execute('/etc/init.d/fastd stop')
+  io.stderr:write('VPN stopped.\n')
+end
+
+local function vpn_start()
+  os.execute('/etc/init.d/fastd start')
+  io.stderr:write('VPN started.\n')
+  brclient_restart()
+end
+
+local function vpn_disable()
+  -- disable VPN if not already disabled
+  os.execute('/etc/init.d/fastd disable')
+  io.stderr:write('VPN disabled.\n')
+end
+
+local function vpn_enable()
+  -- enable VPN if not already enabled
+  os.execute('/etc/init.d/fastd enable')
+  io.stderr:write('VPN enable.\n')
+end
+
+local function wireless_restart()
+  os.execute('wifi')
+  io.stderr:write('Wireless restarted.\n')
+end
+
+-- Get a list of wifi devices return an emty table for no divices
+local function getWifiDevices()
+  local radios = {}
+  uci:foreach('wireless', 'wifi-device',
+    function(s)
+      table.insert(radios, s['.name'])
+    end
+  )
+  return radios
+end
+
+-- Scans for wireless networks and returns a two dimensional array containing
+-- wireless mesh neigbour networks and their properties.
+-- The array is sorted descending by signal strength (strongest signal
+-- first, usually the local signal of the wireless chip of the router)
+local function wlan_list_sorted(radios)
+  local networks = {}
+  for index, radio in ipairs(radios) do
+    local ifname = uci:get('wireless', 'ibss_' .. radio, 'ifname')
+    local ssid = uci:get('wireless', 'ibss_' .. radio, 'ssid')
+    if (ifname ~= nil and ssid ~= nil) then
+      local wireless_scan = string.format( "iw %s scan", ifname)
+      local row = {}
+      row["radio"] = radio
+      -- loop through each line in the output of iw
+      for wifiscan in io.popen(wireless_scan, 'r'):lines() do
+        -- the following line matches a new network in the output of iw
+        if wifiscan:match("BSS (%w+:%w+:%w+:%w+:%w+:%w+)") then
+          if(row["bssid"] ~= nil and row["quality"] ~= nil
+            and row["ssid"] == ssid) then
+            table.insert(networks, row)
+            row = {}
+            row["radio"] = radio
+          end
+        end
+
+        -- get ssid
+        if wifiscan:match("SSID:") then
+          row["ssid"] = wifiscan:split(":")
+          row["ssid"] = row["ssid"][2]
+          if(row["ssid"] ~= nil) then
+            row["ssid"] = trim(row["ssid"])
+          end
+        end
+
+        -- get frequency
+        if wifiscan:match("freq:") then
+          row["frequency"] = wifiscan:split(":")
+          row["frequency"] = row["frequency"][2]
+          if(row["frequency"] ~= nil) then
+            row["frequency"] = trim(row["frequency"])
+          end
+        end
+
+        -- get bssid
+        if wifiscan:match("(%w+:%w+:%w+:%w+:%w+:%w+)") then
+          row["bssid"] = wifiscan:match("(%w+:%w+:%w+:%w+:%w+:%w+)"):upper()
+        end
+
+        -- get signal strength
+        if wifiscan:match("signal:") then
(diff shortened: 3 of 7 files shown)


lemoer:
I also think that a flow chart or a graph of the state machine in this package would be very nice.

lemoer:
I have not read the implementation of the hoodselector yet, but I already have some comments on the discussed algorithm/state machine:
1. You are using rectangles for the "geo decision". This will only work for your actual split-task in ffnw. Other communities will need much more differentiated geo decisions, since they e.g. want to split at rivers.
2. I think it would be very useful, if the algorithm would rely on the "scan decision" if the geo position of the router is at the **border** of a hood. Now the question comes up, how thick this border should be. I think we can handle this question by the following idea:
    - I think (by now) you assert, that the geo hoods are defined geometrically disjunct.
    - Drop this assertion.
    - Use the following algorithm:
        ```
        1. If geo position set:
            1. If multiple hoods match geo: use the "scan decision":
                1. sucessful: join the resulting hood
                2. no neighbours: use one of the geo matches
            2. If one hood match geo: use this hood
            3. Else: Use the default hood
         2. Else: use the "scan decision"
         ```
    - The result of this algorithm would be, that the above described **borders** are defined by the overlap of hoods.
        - If one do not want to have this "border scan"-behavior, he could simply define the hoods geomatrically disjunct.
    - This idea would also remove the need for validating if the hoods are defined geomatrically disjunct, which I think is needed by now.
              
3. A manual overwrite in the config mode would be nice to have.

package/ffnw-hoodselector/luasrc/hoodselector
```diff
@@ -0,0 +1,722 @@
...
+-- Create md5 hash from currend hood
+local function molwm_md5hash(hood)
+  local file = io.open("/tmp/.hoodhash", "w")
+  if not file then
+    io.stderr:write('\"/tmp/.hoodhash\" can not created\n')
+  else
+    file:write(json.stringify(hood, { indent = true }))
+    file:close()
+    --part to create md5 hash of this file
+    for line in io.popen(string.format( "md5sum /tmp/.hoodhash")):lines() do
```
lemoer: We have a buildin function to do md5 hashes:
- https://github.com/freifunk-gluon/packages/blob/master/libs/lua-hash/src/hash.c

rotanid:
@lemoer please note that the "real" author of this pull request isn't active on github, so it might be helpful also answer him with your improvement suggestions on the mailing list.

PowerPan:
Depending on the comment of @lemoer 

> I think it would be very useful, if the algorithm would rely on the "scan decision" if the geo position of the router is at the border of a hood. Now the question comes up, how thick this border should be. I think we can handle this question by the following idea:

If you take a look to the Sourcecode than the border is really thin and not thick. There is no Tri-State Possible. The Router can only be in one Area and not in Two Areas to the same Time

https://git.ffnw.de/ffnw-firmware/packages/blob/master/hoodselector/luasrc/usr/sbin/hoodselector#L428-438

```
-- Return hood from the hood file based on geo position or nil, no real hood could be determined
local function getHoodByGeo(jhood,geo)
  for _, h in pairs(jhood) do
    for _, box in pairs(h.boxes) do
      if ( geo.lat >= box[1][1] and geo.lat < box[2][1] and geo.lon >= box[1][2] and geo.lon < box[2][2] ) then
        return h
      end
    end
  end
  return nil
end
```

All in all, there can not be (for the moment)

> If multiple hoods match geo

package/ffnw-hoodselector/luasrc/hoodselector
2tata: In the current version it is already solved by hash.c

2tata:
I`ve created a rough state diagrame from the hoodselector:

but notice the version here ist not the newest.

The diagrame for the newest is:
https://mw.ffnw.de/images/2/2b/Hoodselector.jpeg

vg
Tarek

Adorfer:
So if you have radios, but no radio neighbors, then the router would not look for cable neighbors hood?

2tata:
On 02/18/17 15:27, Adorfer wrote:
> So if you have radios, but no radio neighbors, then the router would not look for cable neighbors hood? 
> 

Oh, I´ll change it in the diagram.

Inside the code the hoodselector will check the cable neighbors if no radio neighbors founded.

vg
Tarek

Adorfer:
@2tata 

grafik
In case a node has VPN and no geoposition, it will go into the default hood, even if it sees radio- or lan neighbors in local hoods?

2tata:
On 02/18/17 22:14, Adorfer wrote:
> @2tata 
> 
> grafik
> In case a node has VPN and no geoposition, it will go into the default hood, even if it sees radio- or lan neighbors in local hoods?

Right, because a vpn router has a high priorety then others. A VPN router have a connection to internet an can get its position
for exsample wifi based location or geo IP location.

even switching of too many routers from one hood into an another from the VPN server view is not possebile.

vg
Tarek

Adorfer:
I would suggest to avoid routers from different hoods on the same ethernet (even on different vlan/vxlan). 
So i would always priorize any ethernet-neighbors's decision, even if there is other location "via wifiscanning" is available,

2tata:
On 02/19/17 02:03, Adorfer wrote:
> I would suggest to avoid routers from different hoods on the same ethernet (even on different vlan/vxlan). 
> So i would always priorize any ethernet-neighbors's decision, even if there is other location "via wifiscanning" is available, 
>

right that wars the plan but vxlan is at themoment not avalebile.


The developmend based on v2016.1.x

vg
Tarek

2tata:
On 02/19/17 02:03, Adorfer wrote:
> I would suggest to avoid routers from different hoods on the same ethernet (even on different vlan/vxlan). 
> So i would always priorize any ethernet-neighbors's decision, even if there is other location "via wifiscanning" is available,

The main plan is to avoid collisions by lan/wan with VXLAN but currently VXLAN wars not available. Actually there is a
alterative solution called molwm (mesh on lan wan management) that routers checking if there have mesh neighbours over
lan/wan from other hoods they will disable mesh on the collision interface and automatically enable it if the collision
solved. This Workaround will be replaced in future with VXLAN.

vg
Tarek

package/ffnw-hoodselector/luasrc/hoodselector
```diff
@@ -0,0 +1,722 @@
...
+  for i,hood_server in pairs(hood_serverlist) do
+    uci:section('fastd', 'peer', group .. '_peer_' .. hood_server.host:split('.')[1]:gsub("%-", "%_"),
+      {
+        enabled = 1,
+        net = 'mesh_vpn',
+        group = group,
+        key = hood_server.publickey,
+        remote = {'\"'..hood_server.host..'\"'..' port '..hood_server.port}
+      }
+    )
```
mweinelt: Having such large config blocks in here makes maintenance had, since every change to the fastd uci configuration block needs to be mirrored here :(

package/ffnw-hoods/files/lib/ffnw/hoods/hoods.json
```diff
@@ -0,0 +1,64 @@
+[
+  {
+    "name": "default",
+    "bssid": "02:CA:FF:AA:BA:BF",
+    "defaulthood": true,
+    "servers": [
+      {
+        "host": "default02.entenhausen.de",
```
kb-light: Why are you using a different structure for the fastd configuration than the site.conf?

package/ffnw-hoodselector/files/usr/lib/micron.d/hoodselector
```diff
@@ -0,0 +1 @@
+*/2 * * * * /usr/sbin/hoodselector
```
kb-light: So the hood may change every two minutes? That seems to be rather often.

package/ffnw-hoodselector/luasrc/hoodselector
```diff
@@ -0,0 +1,722 @@
...
+  if uci:get('network', 'mesh_wan') and mesh_en then
+    mesh_on_wan_enable()
+  end
+  if uci:get('network', 'mesh_lan') and mesh_en then
+    mesh_on_lan_enable()
+  end
+end
+
+-- Create md5 hash from currend hood
+local function molwm_md5hash(hood)
```
kb-light: What are the improvements in using the md5hash instead of the hoodname?

package/ffnw-hoodselector/luasrc/hoodselector
```diff
@@ -0,0 +1,722 @@
...
+-- wireless mesh neigbour networks and their properties.
+-- The array is sorted descending by signal strength (strongest signal
+-- first, usually the local signal of the wireless chip of the router)
+local function wlan_list_sorted(radios)
+  local networks = {}
+  for index, radio in ipairs(radios) do
+    local ifname = uci:get('wireless', 'ibss_' .. radio, 'ifname')
+    local ssid = uci:get('wireless', 'ibss_' .. radio, 'ssid')
+    if (ifname ~= nil and ssid ~= nil) then
+      local wireless_scan = string.format( "iw %s scan", ifname)
```
kb-light: You may use something like:

    local iwinfo = require "iwinfo"
    local wifitype = iwinfo.type(radio)
    local iw = iwinfo[wifitype]
    if iw then
      local tmplist = iw.scanlist(radio)
      for _, net in ipairs(tmplist) do
      ...

where `tmplist` is some kind of lua array, so there is no need to parse the output of `iw $ifname scan`

package/ffnw-hoodselector/luasrc/hoodselector
```diff
@@ -0,0 +1,722 @@
...
+  table.insert(ret, tonumber(uci:get('gluon-node-info', uci:get_first('gluon-node-info', 'location'), 'latitude')))
+  table.insert(ret, tonumber(uci:get('gluon-node-info', uci:get_first('gluon-node-info', 'location'), 'longitude')))
+  return ret
+end
+
+-- Return hood from the hood file based on geo position or nil, no real hood could be determined
+local function getHoodByGeo(jhood,geo)
+  for n, h in pairs(jhood) do
+    for n, box in pairs(h.boxes) do
+      if ( geo[1] >= box[1][1] and geo[1] < box[2][1] and geo[2] >= box[1][2] and geo[2] < box[2][2] ) then
```
kb-light: This only allows some kind of rectangular shape, which seems not to be suitable for practical use.

package/ffnw-hoodselector/luasrc/hoodselector
```diff
@@ -0,0 +1,722 @@
...
+  io.stderr:write('VPN stopped.\n')
+end
+
+local function vpn_start()
+  os.execute('/etc/init.d/fastd start')
+  io.stderr:write('VPN started.\n')
+  brclient_restart()
+end
+
+local function vpn_disable()
```
kb-light: This script should not disable or enable the fastd autostart.

package/ffnw-hoodselector/src/respondd.c
```diff
@@ -0,0 +1,119 @@
...
+
+		if (strncmp(e->name, "ibss_", 5))
+			continue;
+
+		const char *bssid = uci_lookup_option_string(ctx, s, "bssid");
+		if (!bssid)
+			continue;
+
+		struct json_object *ret = json_object_new_object();
+		json_object_object_add(ret, "bssid", gluonutil_wrap_string(bssid));
```
kb-light: I may agree with adding some `hoodname` and/or `md5hash`, but I do not understand, why the `bssid` is needed.

2tata:
Hi,

On 03/24/17 12:44, kb-light wrote:
> kb-light commented on this pull request.
> 
> - no manual configuration possible

Right the hoodselector sould do it by it selfe.

> - no change auf the SSID

The SSID should not change by switching a hood.

> - no change of next-node IP and prefixes

This is a special v6 address to just get datas from directly local nodes via linklocal.
This address is fix.

> In my opinion it makes more sense to store the hoods.json besides the site.conf, so there is no need for an extra package and the package should be renamed to something like `gluon-*`.

Right, I am working on it.

Today I had finished support for IEEE802.11s
At the moment I am working on l2tp support in the hoodselector
Next should be the hoodfile package.

cheers
Tarek

2tata:
Sorry, I diddet saw the inlines.

> 
>> +local function vpn_reconfigure(hood_serverlist,local_serverlist)
> +  -- remove all servers
> +  for config_index, local_server in pairs(local_serverlist) do
> +    uci:delete('fastd',config_index)
> +  end
> +
> +  -- add servers from hoodfile
> +  local group = 'mesh_vpn_backbone'
> +  for i,hood_server in pairs(hood_serverlist) do
> +    uci:section('fastd', 'peer', group .. '_peer_' .. hood_server.host:split('.')[1]:gsub("%-", "%_"),
> +      {
> +        enabled = 1,
> +        net = 'mesh_vpn',
> +        group = group,
> +        key = hood_server.publickey,
> +        remote = {'\"'..hood_server.host..'\"'..' port '..hood_server.port}
> 
> It is only possible to add just one remote per peer, but we have three in our site.conf (dns, IPv4, IPv6)

correkt... I have to check this. Please can you send me a link of your site.conf?

> 
>> +        key = hood_server.publickey,
> +        remote = {'\"'..hood_server.host..'\"'..' port '..hood_server.port}
> +      }
> +    )
> +  end
> +
> +  uci:save('fastd')
> +  uci:commit('fastd')
> +  io.stderr:write('Fastd needed reconfiguration. Stopped and applied new settings.\n')
> +end
> +
> +-- Checks if wireless needs a reconfiguration. Returns true if any of the checks
> +-- passes. Otherwise the method returns false.
> +local function wireless_reconfiguration_needed(radios, hood_bssid)
> +  for index, radio in ipairs(radios) do
> +    if ( uci:get('wireless', 'ibss_' .. radio, 'bssid') ~= hood_bssid ) then
> 
> This only refers to IBSS mesh networks, but there is also the 802.11s mesh.

In a new version this is already solved.

> 
>> +
> +-- Checks if wireless needs a reconfiguration. Returns true if any of the checks
> +-- passes. Otherwise the method returns false.
> +local function wireless_reconfiguration_needed(radios, hood_bssid)
> +  for index, radio in ipairs(radios) do
> +    if ( uci:get('wireless', 'ibss_' .. radio, 'bssid') ~= hood_bssid ) then
> +      return true
> +    end
> +  end
> +  return false
> +end
> +
> +-- Reconfigure wireless
> +local function wireless_reconfigure(radios, hood_bssid)
> +  for index, radio in ipairs(radios) do
> +    if not ( uci:get('wireless', 'ibss_' .. radio, 'bssid') == hood_bssid ) then
> 
> Same es line 464
> 
>> @@ -0,0 +1,64 @@
> +[
> +  {
> +    "name": "default",
> +    "bssid": "02:CA:FF:AA:BA:BF",
> +    "defaulthood": true,
> +    "servers": [
> +      {
> +        "host": "default02.entenhausen.de",
> 
> Why are you using a different structure for the fastd configuration than the site.conf?

Because more informations wars not necessery.

> 
>> @@ -0,0 +1 @@
> +*/2 * * * * /usr/sbin/hoodselector
> 
>  So the hood may change every two minutes? That seems to be rather often.
> 
>> +  if uci:get('network', 'mesh_wan') and not mesh_en then
> +    mesh_on_wan_disable()
> +  end
> +  if uci:get('network', 'mesh_lan') and not mesh_en then
> +    mesh_on_lan_disable()
> +  end
> +  if uci:get('network', 'mesh_wan') and mesh_en then
> +    mesh_on_wan_enable()
> +  end
> +  if uci:get('network', 'mesh_lan') and mesh_en then
> +    mesh_on_lan_enable()
> +  end
> +end
> +
> +-- Create md5 hash from currend hood
> +local function molwm_md5hash(hood)
> 
> What are the improvements in using the md5hash instead of the hoodname?

There ist a special case inside the scan mode if the hood not known. Were a
hoodconf will aplay with changed neigbour bssid. This will change the hash.

> 
>> +    end
> +  )
> +  return radios
> +end
> +
> +-- Scans for wireless networks and returns a two dimensional array containing
> +-- wireless mesh neigbour networks and their properties.
> +-- The array is sorted descending by signal strength (strongest signal
> +-- first, usually the local signal of the wireless chip of the router)
> +local function wlan_list_sorted(radios)
> +  local networks = {}
> +  for index, radio in ipairs(radios) do
> +    local ifname = uci:get('wireless', 'ibss_' .. radio, 'ifname')
> +    local ssid = uci:get('wireless', 'ibss_' .. radio, 'ssid')
> +    if (ifname ~= nil and ssid ~= nil) then
> +      local wireless_scan = string.format( "iw %s scan", ifname)
> 
> You may use something like:
> 
>     local iwinfo = require "iwinfo"
>     local wifitype = iwinfo.type(radio)
>     local iw = iwinfo[wifitype]
>     if iw then
>       local tmplist = iw.scanlist(radio)
>       for _, net in ipairs(tmplist) do
>       ...
> 
> where `tmplist` is some kind of lua array, so there is no need to parse the output of `iw $ifname scan`

Oh, thanks that a good point :)
> 
>> +end
> +
> +
> +-- Get Geoposition. Return nil for no position
> +local function getGeolocation()
> +  local ret = {}
> +  table.insert(ret, tonumber(uci:get('gluon-node-info', uci:get_first('gluon-node-info', 'location'), 'latitude')))
> +  table.insert(ret, tonumber(uci:get('gluon-node-info', uci:get_first('gluon-node-info', 'location'), 'longitude')))
> +  return ret
> +end
> +
> +-- Return hood from the hood file based on geo position or nil, no real hood could be determined
> +local function getHoodByGeo(jhood,geo)
> +  for n, h in pairs(jhood) do
> +    for n, box in pairs(h.boxes) do
> +      if ( geo[1] >= box[1][1] and geo[1] < box[2][1] and geo[2] >= box[1][2] and geo[2] < box[2][2] ) then
> 
> This only allows some kind of rectangular shape, which seems not to be suitable for practical use.

This is realy practical for use. See https://srv11.ffnw.de/ and enable hood layer. (cert is wrong sorry for that)

>> +  os.execute('ifconfig br-client up')
> +  io.stderr:write('Interface br-client restarted.\n')
> +end
> +
> +local function vpn_stop()
> +  os.execute('/etc/init.d/fastd stop')
> +  io.stderr:write('VPN stopped.\n')
> +end
> +
> +local function vpn_start()
> +  os.execute('/etc/init.d/fastd start')
> +  io.stderr:write('VPN started.\n')
> +  brclient_restart()
> +end
> +
> +local function vpn_disable()
> 
> This script should not disable or enable the fastd autostart.

It shoult in a special case. To avoide damaging the network.

> 
>> +
> +	struct uci_element *e;
> +	uci_foreach_element(&p->sections, e) {
> +		struct uci_section *s = uci_to_section(e);
> +		if (strcmp(s->type, "wifi-iface"))
> +			continue;
> +
> +		if (strncmp(e->name, "ibss_", 5))
> +			continue;
> +
> +		const char *bssid = uci_lookup_option_string(ctx, s, "bssid");
> +		if (!bssid)
> +			continue;
> +
> +		struct json_object *ret = json_object_new_object();
> +		json_object_object_add(ret, "bssid", gluonutil_wrap_string(bssid));
> 
> I may agree with adding some `hoodname` and/or `md5hash`, but I do not understand, why the `bssid` is needed.

The hoodselector in a newer version work with this bssid in mesh lan/wan constructed.

An new pull request will follow soon.

cheers
Tarek

kb-light:
@2tata does every hood has its own IP space?

2tata:
On 03/25/17 11:55, kb-light wrote:
> @2tata does every hood has its own IP space?
> 

Right. Encoded in the BSSID

cheers
Tarek

2tata:
On 03/25/17 11:55, kb-light wrote:
> @2tata does every hood has its own IP space?
> 
You can read folliwing:

https://blog.freifunk.net/2016/05/23/monitoring-and-quality-assurance-open-wifi-networks-out-client-view/
https://blog.freifunk.net/2016/06/19/monitoring-and-quality-assurance-open-wifi-networks-out-client-view-midterm-evaluation/
https://blog.freifunk.net/2016/08/22/monitoring-and-quality-assurance-open-wifi-networks-out-client-view-final-evaluation/

cheers
Tarek

kb-light:
> Right the hoodselector sould do it by it selfe.

In case of discussions about the autoupdater it would be nice to have the option to set the hood manually. Additional there nodes where the owner does not want to set the geolocation, but the hood should be selected correctly. Even if there is no local mesh.

> The SSID should not change by switching a hood.

But each hood has it own IP space, so not changing the SSID may break roaming.

> This is a special v6 address to just get datas from directly local nodes via linklocal. This address is fix.

But changing the IP space, because of changing the hood, would break the next node feature, so this address should be changed to, to match the new IP space.

> correkt... I have to check this. Please can you send me a link of your site.conf?

An example was added to gluon with adf3b91.

> Because more informations wars not necessery.

I do not meant to add the whole fastd section but adding the remotes as it is done within the site.conf, so hostname and port are just one string.

> This is realy practical for use. See https://srv11.ffnw.de/ and enable hood layer. (cert is wrong sorry for that)

We have selected some shapes for our regions/hoods that are not even foursquar, see: http://umap.openstreetmap.fr/de/map/freifunk-hochstift-sites-karte_60146#10/51.7011/8.9243

> It should in a special case. To avoid damaging the network.

Is stopping the fastd not enough in that case?

T-X:
Hi @2tata,

Great work so far! And thanks for the flow chart, that definitely helps to get a quick understanding.

Two technical question: "geo pos. exists", is it determined from the user geo position manually configured by the user? Why is that branch depending on a VPN connection?

A few conceptual questions:

_Scenario A):_

hoodselector

[hoodselector.dia.its-not-a.txt](https://github.com/freifunk-gluon/gluon/files/880915/hoodselector.dia.its-not-a.txt)

Node A and B are in different hoods and are the only VPN uplinks for this local mesh cloud. Node C can reach A and B about equally well. Sometimes A is "better", sometimes B.

Questions: Can you describe what would usually happen for the hood selection for nodes in Q2? What could happen in the worst case?

_Scenario B):_

Just like scenario A), but node D sometimes gains and sometimes looses a VPN uplink.

Questions: Same above.

_Scenario C):_

hoodselector2

[hoodselector2.dia.its-not-a.txt](https://github.com/freifunk-gluon/gluon/files/880917/hoodselector2.dia.its-not-a.txt)

Compared to scenario A), now E has the wifi connection to node A.

Questions: Would anything change concerning the hood selection within Q2 compared to scenario A) or B)? What would happen if the topology switches between scenario A) and C) from time to time?

---

Also, the roaming concerns @kb-light mentioned with using the same ESSID with different IP spaces is something I'd be worried about, too. At least along hood borders. Would love to hear some clarifications for that too.

T-X:
And one more question: Are there any recommendations on how to use this package together with wifi backbone setups?

2tata:
@T-X sorry I didnt saw the Mail.

> Great work so far! And thanks for the flow chart, that definitely helps to get a quick understanding.

> Two technical question: "geo pos. exists", is it determined from the user geo position manually
> configured by the user? Why is that branch depending on a VPN connection?

Because A VPN router have a function like "gate to the rest on net"  each hood has other VPN peers determinate by geopositions. Other Router without VPN conections trys to stay by a VPN router. If we just say on a mesh only router set hood by geo that would not work.

Scenario A):
The mesh Only Routers try surrounded BSSIDs sort by signal quallity if a Mesh Only Router find batman gateways the the Mesh only will stay there.

Scenario B):
The If the VPN connection ist lost the routers will handle same like Senario A) and C). If the VPN connection comes up all routers except , C and the between E,C and D will stay in the other hoods.
D and the routers behind D goes back into Q2

Scenario C):
Same like A) chosing one of the router depending on Link quality and stay there.

In Freifunk Nordwest we defined the hoods that hood borders will not across a city.
On that point we need more work. The hoodselector automated many scenarios but needs in some cases still work.

Few weeks ago I have already finished the IEEE802.11s support. If you are on the WCW2017 I will hold a presentation and we can discuss about that.

2tata:
@T-X sorry I didnt saw the Mail.

> Great work so far! And thanks for the flow chart, that definitely helps to get a quick understanding.

> Two technical question: "geo pos. exists", is it determined from the user geo position manually
> configured by the user? Why is that branch depending on a VPN connection?

Because A VPN router have a function like "gate to the rest on net"  each hood has other VPN peers determinate by geopositions. Other Router without VPN conections trys to stay by a VPN router. If we just say on a mesh only router set hood by geo that would not work.

Scenario A):
The mesh Only Routers try surrounded BSSIDs sort by signal quallity if a Mesh Only Router find batman gateways the the Mesh only will stay there.

Scenario B):
The If the VPN connection ist lost the routers will handle same like Senario A) and C). If the VPN connection comes up all routers except , C and the between E,C and D will stay in the other hoods.
D and the routers behind D goes back into Q2

Scenario C):
Same like A) chosing one of the router depending on Link quality and stay there.

In Freifunk Nordwest we defined the hoods that hood borders will not across a city.
On that point we need more work. The hoodselector automated many scenarios but needs in some cases still work.

Few weeks ago I have already finished the IEEE802.11s support. If you are on the WCW2017 I will hold a presentation and we can discuss about that.

2tata:
hoodselector
Here is a new state diagram.
Some known issues can be found here:
https://git.nordwest.freifunk.net/ffnw-firmware/packages/issues?label_name%5B%5D=hoodselector

lemoer:
@2tata Are there any simplifications in the mechanism since vxlan is supported?

2tata:
Hi @lemoer,

> @2tata Are there any simplifications in the mechanism since vxlan is supported?

currently I am working on #1201 and #1211.
VXLAN will may replace fully the MOWLM protocol and maybe also the bssid switching part.
Currently I was not able to work with VXLAN in the last past weeks.

Hopefully I can force working on this MR this month. 

vg
Tarek

lemoer:
@2tata Do you think, you see a chance to drop your "hoodfiles" in favor of the domain configs proposed in #1216 ?

- **disadvantage:** hoods could not be changed without rolling out a new firmware anymore.
- **advantage:** checking the domain config would be well integrated in the gluon framework and so there will be no more than one (code) location, where configs are verified.

2tata:
On 09/04/17 22:45, lemoer wrote:
> @2tata Do you think, you see a chance to drop your "hoodfiles" in favor of the domain configs proposed in #1216 ?
> 
> - **disadvantage:** hoods could not be changed without rolling out a new firmware anymore.

That was my plan, to share the hoodfile with other routers because for thous configurations it is not necessary to roll out a new Firmware over the whole network.

> - **advantage:** checking the domain config would be well integrated in the gluon framework and so there will be no more than one (code) location, where configs are verified.

Maybe we can split up the site.conf into a dynamic and a static part?
So the static part will be only change be Firmware upgrades and the
dynamic part can be replace on runtime in future?

But I am flexible and can also drop the hoodfile if it wished.

cheers,
Tarek

2tata:
**This MR can be close.**
@lemoer, @Adorfer, @mweinelt, @kb-light, @T-X: Please take a look at #1226. If you have some comments still open here please take them to the new merge request. :)

PR #999, opened 2016-12-30
gluon-dns-config: add package that allows the next node to be used as dns cache

christf:
this reconfigures the dnsmasq-instance to accept dns queries including the forwards.


christf:
brml. git fuckup.

PR #1003, opened 2017-01-11
[RFC] add packages for multi site support

### LINKED ISSUE #789: include new packages: hoods/ hoodselector
Hookselector sounds nice!
there should be (3) packages inside this repo
https://blog.nordwest.freifunk.net/2016/05/20/neue-firmware-0-9-hoodselector-und-support-fuer-wr841ndv11/

code atm near:
https://git.nordwest.freifunk.net/ffnw-firmware/packages/blob/a1b06b6646c4b51129cb16575776b65f5b278b15/hoodselector/luasrc/hoodselector

kb-light:
As an alternative to the hoodselector (#997) I ported the Freifunk Hochstift packages for multi site support.

fixes #789

ToDo:

- [ ] The selected site is currently storred as `currentsite.current.name`, but there is mayby a better place. Any suggestions?
- [ ] Same applies to the config files in `${SITEDIR}/extra/`
- [ ] Update for Lede based master
- [ ] We may should add some checks for the additional config files
- [ ] Embed generation of default site.conf to gluon build environment
- [ ] Remove `subst['%%GS'] = site.site_code` from opkg upgrade script, because modules/packages path is only firmware dependent, not site dependent.
- [ ] France translation

The original packages can be found here:

- https://git.c3pb.de/freifunk-pb/ffho-packages
- https://github.com/FreifunkHochstift/ffho-packages

There is also a package available, which supports automatic site selection based on coordinates and gluon-neighbour-info.

EDIT:
An example configuration can be found [here](https://git.c3pb.de/freifunk-pb/site-ffho/tree/site-split)

EDIT2:
Add ToDo list


H4ndl3:
I like this approach, because it's simple yet effective.

The main advantages I see over #997 are:

- works without coordinates
- transparent to the user
- no issues with defining hoods, which could be difficult in some cases (e. g. a site within a site)
- less prone to errors, because there are no corner cases that make automatic selection difficult
- simpler and therefore easier to configure and maintain

docs/package/gluon-site-generate.rst
```diff
@@ -0,0 +1,75 @@
+gluon-site-generate
+===================
+
+This package generates the needed `site.json` directly on the node after firmware
+upgrad has been performed. This can be used, to support different sites/regions
```
mweinelt: typo: upgrade

docs/package/gluon-site-generate.rst
```diff
@@ -0,0 +1,75 @@
...
+within one firmware image. The default `site.json` will be replaced by
+`extra/template.conf` which is patly replacte by values defined in
```
mweinelt: typo: partially replaced

kb-light:
I added a ToDo list with some open tasks to my first post

lemoer:
I think I wrote enough documentation, so I'll want to mention this here...

We are currently planning our "segregation" in Hannover and did some major adjustments to the `site-select` package. I think, we branched at an earlier version, so I'm not hundred percent up to date with your state.

The main difference of our branch is, that we are "patching" one `site.conf`, instead of using multiple `site.conf`, because we want to have around 50-60 potentially different sites. Each city district and each subcommunity outside of hannover should get it's own field in the dropdown, so we can later decide (and redecide) which of them should be in the same l2 network.

I wrote a lot of documentation for our [package](https://github.com/freifunkh/ffh-packages/blob/master/ffh-district-core/README.md). Also keeping in mind, that maybe some other communities may want to use this?

Some comments:
- It's not yet in use. Our server infrastructure for this is not ready yet.
- There are some ugly things left, which should fixed later (sites are not verified correctly yet, config should be completely in `site/` repo, ...)

What do you think about our approach?

NeoRaider:
Another idea I had was to split the site.conf into two parts (might be separate files, or just a nested layer inside a single site.conf): a "site" and a "domain" part (where the site options are valid for all nodes using the same images, while domain options belong to a single mesh domain). Some options like autoupdater configuration (especially signer keys) would be redundant when each domain using the same images would have to include them, so they would stay in the site part, while others (SSIDs, VPN configuration, ...) would be moved to the domain configuration.

It will need some adjustment in packages, but I think it is cleaner than the existing approaches (either using multiple site.conf or patching it in some way).

It would be nice to have some basic domain switching support in the Gluon core, so both automatic solutions like the Hood Selector and manual selection packages can build on it.

kb-light:
I will close this in favour of #1216 and #1250

PR #1006, opened 2017-01-18
fixing status page in Microsoft Edge browser

altima:
this will fix the problem for displaying status page on Microsoft Edge (maybe older IE's) browsers.

### DIFF
--- a/package/gluon-status-page/Makefile
+++ b/package/gluon-status-page/Makefile
@@ -28,6 +28,12 @@ define Download/almond
 	MD5SUM:=aa66c0c0cb55a4627bb706df73f3aff5
 endef
 
+define Download/EventSource
+	FILE:=eventsource.js
+	URL:=https://raw.githubusercontent.com/altima/EventSource/master/eventsource.js
+	MD5SUM:=9b5c81ca6ba306ac6aa11a61c9ab9891
+endef
+
 $(eval $(call Download,rjs))
 $(eval $(call Download,Bacon))
 $(eval $(call Download,almond))
@@ -46,7 +52,7 @@ endef
 
 define Build/Prepare
 	mkdir -p $(PKG_BUILD_DIR)
-	$(CP) -t $(PKG_BUILD_DIR) $(DL_DIR)/r.js $(DL_DIR)/Bacon.js $(DL_DIR)/almond.js
+	$(CP) -t $(PKG_BUILD_DIR) $(DL_DIR)/r.js $(DL_DIR)/Bacon.js $(DL_DIR)/almond.js $(DL_DIR)/eventsource.js
 endef
 
 define Build/Configure
--- a/package/gluon-status-page/src/build.js
+++ b/package/gluon-status-page/src/build.js
@@ -1,6 +1,7 @@
 ({
   paths: {
-    "bacon": "../Bacon"
+    "bacon": "../Bacon",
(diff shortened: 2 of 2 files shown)


altima:
see https://github.com/FreifunkVogtland/gluon/pull/2 for first try and https://github.com/FreifunkVogtland/gluon/pull/3 (same as this request) for better workaround

package/gluon-status-page/Makefile
```diff
@@ -28,6 +28,12 @@ define Download/almond
 	MD5SUM:=aa66c0c0cb55a4627bb706df73f3aff5
 endef
 
+define Download/EventSource
+	FILE:=eventsource.js
+	URL:=https://raw.githubusercontent.com/altima/EventSource/master/eventsource.js
```
NeoRaider: The URL should be (pin it to a tag, use upstream repo URL, don't include filename):

    https://raw.githubusercontent.com/Yaffle/EventSource/0.0.8

Also, you need to add `$(eval $(call Download,EventSource))` below.

package/gluon-status-page/src/build.js
```diff
@@ -1,6 +1,7 @@
 ({
   paths: {
-    "bacon": "../Bacon"
+    "bacon": "../Bacon",
+    "eventsource": "../EventSource"
```
NeoRaider: Referencing it here won't help, require.js will only include code that is referenced. I think it would make sense to just download the pre-minified version from the git repo, copy that into the output dir in Package/gluon-status-page/install, and reference it in the HTML code like you did before.

altima: thanks for the info. I'm not familiar with the build scripts of gluon. @ecsv told me to introduce a patch in main gluon repo. I will change my request to the hints you made. We should support all browsers with or without a polyfill. I use the Chrome browser mainly but Edge is the common browser on Windows 10 and I think a handful of users will use this platform. I don't want to start a discussion about used platforms or browsers, but Edge is the main browser for Windows.

just wait for a new change from me before accept the pull request. thx

rotanid:
maybe we should wait with this if multiple communities complain after the lede-migration about the lack of space and only proceed here if they don't ?
according to statista, edge was at only 4% market share on November 2016

jplitza:
I don't like adding KBs of JS just for one browser with a low market share either.

Although I usually like including external stuff just as little, the approach in  d3475f81b3f2160b9185552451f28affd235364f appeals more to me *if* it could somehow be made that the external file is included if needed (I understand Edge doesn't support conditional comments anymore).

altima:
Ok, so I change the html to use an cdn source for the polyfill. But for communities that are not connected to the normal internet will get an error. Any suggestions for that? Ignoring users with Microsoft enabled devices or is there any other approach, maybe websocket or polling?

NeoRaider:
After thinking a bit more about this, I think 3KB are okay; I really don't like adding dependencies on external resources.

altima:
thanks for the positiv anwser. I will change the suggested stuff tonight or tomorrow.

NeoRaider:
I just had a look at this, and the "polyfill" isn't actually a proper polyfill, but a replacement with compatible client-side API; it is not compatible on the server side, but requires adjustments to the binaries generating the event streams. This would further complicate our implementation, so I'll have to reject this.

PR #1007, opened 2017-01-19
Remove dependency on luci-base, introduce gluon-web as new base for Config Mode

NeoRaider:
luci-base is huge (> 100KB compressed), and we only use a small subset of it. LuCI base translations take up >20KB / language.

Gluon packages are converted from using *luci.model.uci* to *uci-simple*, remaining uses *luci-base* libraries outside of the config mode are removed.

LuCI is replaced with a stripped-down version "gluon-web" without features we don't need; in addition, it has been refactored significantly to make it more maintainable.

gluon-web has a base size of about 19KB. The total per-language overhead for all packages in the Gluon repository is less than 4KB. In total, this PR reduces the rootfs size for a typical build with *en* and *de* languages by about 140KB.

To accommodate the removal of LuCI, all *luci-admin-* packages have been renamed:

* gluon-luci-admin -> gluon-web-admin
* gluon-luci-autoupdater -> gluon-web-autoupdater
* gluon-luci-mesh-vpn-fastd -> gluon-web-mesh-vpn-fastd
* gluon-luci-node-role -> gluon-web-node-role
* gluon-luci-portconfig -> gluon-web-network
* gluon-luci-private-wifi -> gluon-web-private-wifi
* gluon-luci-theme -> gluon-web-theme
* gluon-luci-wifi-config -> gluon-web-wifi-config

Testing of the new config wizard/advanced settings is very welcome. The following points should be completed before merging:
* [x] Documentation
* [ ] CSRF protection (the newest version of LuCI has CSRF support, but relied on rpcd for its session management; therefore this implementation has not been included in gluon-web)


NeoRaider:
I've decided to postpone the CSRF protection and get this merged. No released version of Gluon has CSRF protection, so this isn't a regression (I'll still try to add this before Gluon 2017.1).

PR #1013, opened 2017-01-25
gluon-dns-config: update documentation

rotanid:


rotanid:
@christf do you object with any of my suggested changes to your documentation? if not, we can merge it without the need for spare time of another maintainer.

christf:
not at all, thank your for those improvements! :)

rotanid:
hm, maybe someone would object to the prefix in the commit message, maybe it has to be gluon-core... sorry, too late...

PR #1015, opened 2017-01-26
[RFC] add package gluon-ebtables-source-filter

### LINKED ISSUE #541: "Foreign" ARP table entries
We observe, that there is an extreme number of ARP requests flooding our network from many sources. These are either by foreign networks or for foreign networks and by "foreign" I mean non-Freifunk networks, mostly private IP ranges such as the ones found in many home networks. 

I suggest creating an ebtables filter package that prevents such ARP packets (src IP does not match prefix4 from site.conf OR dst IP does not match site.conf) from being forwarded via batman. It might be useful to have them forwarded to other local devices, so the rule should probably not be applied at br-client level, but rather really at bat-device level.

kb-light:
First draft to port ffho-ebtables-net-rules to Gluon

depends on #1016
fixes #541

decision needed:

- [x] the packagename is verry generic, so it should be changed to something that fits better, e.g. gluon-ebtables-strict-filter/-strict-net/-strict-source/-source-guard/-source-filter/...
  **=> gluon-ebtables-source-filter it is**
- [x] if no prefix4 is given within `site.conf` this package blocks all IPv4 traffic. Is that a reasonable default? **=> yes**
- [x] should we allow link-local-IPv4/APIPA/`169.254.0.0/16`? **=> no**
- [x] for documentation and statistics it may be intresting to use:
  `rule 'SOURCE -p IPv6 --ip6-src ::/128 --ip6-protocol ipv6-icmp --ip6-icmp-type 131 -j RETURN'`
  `rule 'SOURCE -p IPv6 --ip6-src ::/128 --ip6-protocol ipv6-icmp --ip6-icmp-type 135 -j RETURN'`
  `rule 'SOURCE -p IPv6 --ip6-src ::/128 --ip6-protocol ipv6-icmp --ip6-icmp-type 143 -j RETURN'`
  instead of: `rule 'SOURCE -p IPv6 --ip6-src ::/128 -j RETURN'` **=> keep it as it is**

One can find the original package under the following links:
https://git.c3pb.de/freifunk-pb/ffho-packages/tree/master/ffho/ffho-ebtables-net-rules
https://github.com/FreifunkHochstift/ffho-packages/tree/master/ffho/ffho-ebtables-net-rules

### DIFF
--- a/package/gluon-ebtables-source-filter/check_site.lua
+++ b/package/gluon-ebtables-source-filter/check_site.lua
@@ -1,0 +1,2 @@
+need_string_match('prefix4', '^%d+.%d+.%d+.%d+/%d+$', false)
+need_string_array_match('extra_prefixes6', '^[%x:]+/%d+$', false)
--- a/package/gluon-ebtables-source-filter/files/lib/gluon/ebtables/110-local-forward-allow-arp
+++ b/package/gluon-ebtables-source-filter/files/lib/gluon/ebtables/110-local-forward-allow-arp
@@ -1,0 +1,6 @@
+prefix4 = require('gluon.site_config').prefix4
+
+if prefix4 then
+	rule('LOCAL_FORWARD -p ARP --arp-ip-src ' .. prefix4 .. ' --arp-ip-dst ' .. prefix4 .. ' -j RETURN')
+	rule('LOCAL_FORWARD -p ARP --arp-ip-src 0.0.0.0 --arp-ip-dst ' .. prefix4 .. ' -j RETURN')
+end
--- a/docs/package/gluon-ebtables-source-filter.rst
+++ b/docs/package/gluon-ebtables-source-filter.rst
@@ -1,0 +1,30 @@
+gluon-ebtables-source-filter
+============================
+
+The *gluon-ebtables-source-filter* package adds an additional layer-2 filter
+ruleset to prevent unreasonable traffic entering the network via the nodes.
+Unreasonable means traffic entering the mesh via a node which source IP does
+not belong to the configured IP space.
+
+One may first check if there is a certain proportion of unreasonable traffic,
+before adding this package to the firmware image. Additional one should not
+use this package if some kind of gateway or upstream network is provided by
+a device connected to the client port.
+
+site.conf
+---------
+
+prefix4 : optional
+    - IPv4 subnet
+
+prefix6 :
+    - IPv6 subnet
+
+extra_prefixes6 : optional
+    - list of additional IPv6 subnets
+
+Example::
+
+  prefix4 = '198.51.100.0/21',
+  prefix6 = '2001:db8:8::/64',
+  extra_prefixes6 = { '2001:db8:9::/64', '2001:db8:100::/60' },
--- a/package/gluon-ebtables-source-filter/Makefile
+++ b/package/gluon-ebtables-source-filter/Makefile
@@ -1,0 +1,43 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-ebtables-source-filter
+PKG_VERSION:=1
+PKG_RELEASE:=1
+
+PKG_BUILD_DIR := $(BUILD_DIR)/$(PKG_NAME)
+
+include ../gluon.mk
+
+
+define Package/gluon-ebtables-source-filter
+  SECTION:=gluon
+  CATEGORY:=Gluon
+  TITLE:=Ebtables rules to filter unreasonable L2 traffic.
+  DEPENDS:=+gluon-core +gluon-ebtables
+endef
+
+define Package/gluon-ebtables-source-filter/description
+	This package adds an additional layer-2 filter-ruleset to prevent unreasonable
+	traffic entering the network via the nodes.
+endef
+
+define Build/Prepare
+	mkdir -p $(PKG_BUILD_DIR)
+endef
+
+define Build/Configure
+endef
+
+define Build/Compile
+endef
+
+define Package/gluon-ebtables-source-filter/install
+	$(CP) ./files/* $(1)/
+endef
+
+define Package/gluon-ebtables-source-filter/postinst
+#!/bin/sh
+$(call GluonCheckSite,check_site.lua)
+endef
+
+$(eval $(call BuildPackage,gluon-ebtables-source-filter))
--- a/package/gluon-ebtables-source-filter/files/lib/gluon/ebtables/100-local-forward-chain
+++ b/package/gluon-ebtables-source-filter/files/lib/gluon/ebtables/100-local-forward-chain
@@ -1,0 +1,1 @@
+chain('LOCAL_FORWARD', 'DROP')
--- a/package/gluon-ebtables-source-filter/files/lib/gluon/ebtables/110-local-forward-allow-ipv4
+++ b/package/gluon-ebtables-source-filter/files/lib/gluon/ebtables/110-local-forward-allow-ipv4
@@ -1,0 +1,6 @@
+prefix4 = require('gluon.site_config').prefix4
+
+if prefix4 then
+	rule('LOCAL_FORWARD -p IPv4 --ip-protocol udp --ip-destination-port 67 -j RETURN')
+	rule('LOCAL_FORWARD -p IPv4 --ip-src ' .. prefix4 .. ' -j RETURN')
+end
--- a/package/gluon-ebtables-source-filter/files/lib/gluon/ebtables/110-local-forward-allow-ipv6
+++ b/package/gluon-ebtables-source-filter/files/lib/gluon/ebtables/110-local-forward-allow-ipv6
@@ -1,0 +1,9 @@
+site = require('gluon.site_config')
+
+rule('LOCAL_FORWARD -p IPv6 --ip6-src fe80::/64 -j RETURN')
+rule('LOCAL_FORWARD -p IPv6 --ip6-src ::/128 --ip6-proto ipv6-icmp -j RETURN')
+rule('LOCAL_FORWARD -p IPv6 --ip6-src ' .. site.prefix6 .. ' -j RETURN')
+
+for _, prefix in ipairs(site.extra_prefixes6 or {}) do
+	rule('LOCAL_FORWARD -p IPv6 --ip6-src ' .. prefix .. ' -j RETURN')
+end
--- a/package/gluon-ebtables-source-filter/files/lib/gluon/ebtables/300-local-forward-rules
+++ b/package/gluon-ebtables-source-filter/files/lib/gluon/ebtables/300-local-forward-rules
@@ -1,0 +1,1 @@
+rule('FORWARD --logical-in br-client -i ! bat0 -j LOCAL_FORWARD')


package/gluon-ebtables-net-rules/files/lib/gluon/ebtables/110-net-allow-ipv6
```diff
@@ -0,0 +1,8 @@
+siteConfig = require("gluon.site_config")
+
+rule ('NET_SRC -p IPv6 --ip6-src fe80::/10 -j RETURN')
```
T-X: For IPv6 Duplicate Address Detection "::" is used as the source address of an ICMPv6 Neighbor Solicitation message. This is mandatory by the IPv6 standard.

(Could it be that this is where your 30% overhead reduction came from?)

kb-light: Now the additional multicast rule within our package makes sense. I have added it to this PR.

package/gluon-ebtables-net-rules/files/lib/gluon/ebtables/110-net-allow-ipv6
```diff
@@ -0,0 +1,9 @@
...
+rule ('NET_SRC -p IPv6 --ip6-dst ff00::/8 -j RETURN')
```
kb-light: maybe reduce this to `--ip6-protocol ipv6-icmp` or `--ip6-protocol ipv6-icmp --ip6-icmp-type 135`?

T-X: Hm, yes, I think the latter makes sense. I could not think of any other protocol requiring "::" yet.

If this package is supposed to be one for strict filtering, then maybe also add something like "--ip6-src ::/128" and change the destination to "ff02::/16"?

T-X: Ah, reports should be excluded for "::", too, they will happen just before the IPv6 DAD Neighbor Solicitations. So icmpv6 types 131 and 143.

(See https://tools.ietf.org/html/rfc3810#section-5.2.13., for instance)

kb-light: It seems, that there are the following possible solutions:

- `rule ('NET_SRC -p IPv6 --ip6-src ::/128 -j RETURN')`
- `rule ('NET_SRC -p IPv6 --ip6-src ::/128 --ip-dst ff00::/8 -j RETURN')`
- `rule ('NET_SRC -p IPv6 --ip6-src ::/128 --ip6-protocol ipv6-icmp -j RETURN')`

Just allowing each possible icmp-type individually seemes unreasonable.

T-X: It's just three, so it'd be three rules :-). One for 131, one for 135 and one for 143.

kb-light: Thats right, but does it makes any difference compared to globally allow icmp from `::`? (If only 131, 135 and 143 are expected from source `::`)

T-X: Hm, no, probably not. The advantage would be to have individual counters/statistics, which you could look at at any time with ebtables, and some sort of documentation regarding what exactly needs "::".

mweinelt:
The documentation states:
```
prefix4 : optional
The IPv4 Subnet of your community mesh network in CIDR notation, e.g.
```

If no prefix4 is given it is within reason not to expect anyy IPv4 traffic in my opinion.

T-X:
Just finished some measurements regarding these new rules in the Freifunk Lübeck network. FFHL currently has about 370 nodes.

The test data was captured for about 29 hours (2017-02-07 00:16 -> 2017-02-08 05:01) with tcpdump on a VPN interface. Therefore the original capture includes OGMs and batman-adv headers.

I then simulated the rules provided by this pull-request via tshark, by splitting the capture into three files:

0. original
1. bcast:
 <code>batadv.batman.packet_type == 4</code>
2. bcast + src filter:
 <code>batadv.batman.packet_type == 4 && (ipv6.src == fe80::/10 || (ipv6.src == ::/128 && icmpv6) || ipv6.src == fdef:ffc0:3dd7::/64 || ipv6.src == 2001:67c:2d50::/48 || ip.src == 10.130.0.0/20 || (arp.src.proto_ipv4 == 10.130.0.0/20 && arp.dst.proto_ipv4 ==10.130.0.0/20) || (arp.src.proto_ipv4 == 0.0.0.0 && arp.dst.proto_ipv4 == 10.130.0.0/20))</code>

### Results:

With "capinfos" the result is the following:

(0): original
<pre>
File name:           ffhl-stats-2017-02-07-original.cap
File type:           Wireshark/... - pcapng
File encapsulation:  Ethernet
File timestamp precision:  microseconds (6)
Packet size limit:   file hdr: (not set)
Number of packets:   7226 k
File size:           890 MB
Data size:           649 MB
Capture duration:    103514,488168 second
First packet time:   2017-02-07 00:16:11,875633
Last packet time:    2017-02-08 05:01:26,363801
Data byte rate:      6277 bytes/s
Data bit rate:       50 kbps
Average packet size: 89,92 bytes
Average packet rate: 69 packets/s
SHA1:                bec66c674e59e2ae2dc8cd32e8ddb8321a7a2864
RIPEMD160:           5e2755ac5dd669e8149c9d05cc6e046387455a47
MD5:                 48aeec621645be2beaf08667aae62730
Strict time order:   True
Capture application: TShark (Wireshark) 2.2.1 (Git Rev Unknown from unknown)
Number of interfaces in file: 1
Interface #0 info:
                     Encapsulation = Ethernet (1/1 - ether)
                     Capture length = 262144
                     Time precision = microseconds (6)
                     Time ticks per second = 1000000
                     Number of stat entries = 0
                     Number of packets = 7226441
</pre>

(1): bcast
<pre>
File name:           ffhl-stats-2017-02-07-bcast.cap
File type:           Wireshark/... - pcapng
File encapsulation:  Ethernet
File timestamp precision:  microseconds (6)
Packet size limit:   file hdr: (not set)
Number of packets:   3690 k
File size:           509 MB
Data size:           384 MB
Capture duration:    103514,338195 second
First packet time:   2017-02-07 00:16:12,025606
Last packet time:    2017-02-08 05:01:26,363801
Data byte rate:      3712 bytes/s
Data bit rate:       29 kbps
Average packet size: 104,12 bytes
Average packet rate: 35 packets/s
SHA1:                a05c5a32c64d6bec13af7c029d10cbeb02304f57
RIPEMD160:           63f93979320de6200797abc20084dc47b2575850
MD5:                 c2622b96afddf7fb5410667315611ade
Strict time order:   True
Capture application: TShark (Wireshark) 2.2.1 (Git Rev Unknown from unknown)
Number of interfaces in file: 1
Interface #0 info:
                     Encapsulation = Ethernet (1/1 - ether)
                     Capture length = 262144
                     Time precision = microseconds (6)
                     Time ticks per second = 1000000
                     Number of stat entries = 0
                     Number of packets = 3690678
</pre>

(2): bcast + srcfilter
<pre>
File name:           ffhl-stats-2017-02-07-bcast+srcfilter.cap
File type:           Wireshark/... - pcapng
File encapsulation:  Ethernet
File timestamp precision:  microseconds (6)
Packet size limit:   file hdr: (not set)
Number of packets:   3611 k
File size:           501 MB
Data size:           378 MB
Capture duration:    103514,338195 second
First packet time:   2017-02-07 00:16:12,025606
Last packet time:    2017-02-08 05:01:26,363801
Data byte rate:      3658 bytes/s
Data bit rate:       29 kbps
Average packet size: 104,86 bytes
Average packet rate: 34 packets/s
SHA1:                11e5aa40a81853d3a97a3e4b90e027488504874d
RIPEMD160:           f00a574792b0e5cca36872589be319dba453672a
MD5:                 7e00cae450f9b0d65b671ef10d122b5e
Strict time order:   True
Capture application: TShark (Wireshark) 2.2.1 (Git Rev Unknown from unknown)
Number of interfaces in file: 1
Interface #0 info:
                     Encapsulation = Ethernet (1/1 - ether)
                     Capture length = 262144
                     Time precision = microseconds (6)
                     Time ticks per second = 1000000
                     Number of stat entries = 0
                     Number of packets = 3611161
</pre>

### Conclusion:

This package seems to reduce the broadcast overhead by about **2.15% / 1.45%** (packet count / byte count), excluding VPN header considerations.

So, unfortunately, I do not seem to be able to reproduce the proclaimed 30% :-(. Any thoughts regarding my method?

BarbarossaTM:
Hi,

A short clarification: The 30% reduction was calculated against the control traffic measured before putting the filters in place ("Grundrauschen": ARP, ICMP(6), STP, ...).

As it's quite a while ago I've done that, I'm not sure about the duration of my control traffic sample and the like. What I can tell you is that I used filters in Wireshark on a pcap file of caputred control traffic and filtered out the stuff which I put into ebtables rules to be blocked back then. The capture(s) I took were done on the bridge interfaces (like "inside" batman).

We were surprised too back then that there was that much illegitimate stuff flying around :-)

I hope this helps to better relate your findings.

Best
Max

rotanid:
maybe it depends massively on the specific setups people use in the networks, maybe Barbarossa had some really bad behaving nodes (or clients on these nodes) and T-X doesn't have such

T-X:
Ok, thanks for the clarification, @BarbarossaTM! Looks similar to the method I used.

Hm, right, maybe with the dawn of the "Internet of (Sh..ty) Things" we might need rules like that... as there seem to be some networks having weird devices, it seems to make sense to have such a package available.

My current thought/recommendation would be for people to install this package after having performed some measurements. Just to avoid firmwares accumulating "placebo" packages which do nothing for their network. Would it make sense to try to clarify the pros and cons of this package a little more in the package description?

One more thought came to my mind after reading #541: Could it be, that foreign address ranges are only a problem for the IPv4 world, with "smart" devices scanning popular /24 subnets? Would it be possible for someone needing source filtering to gather some "ebtables -L --Lc" statistics so we could check whether the 30% were just IPv4 and not IPv6? @BarbarossaTM, maybe?

Also, with the integration of batman-adv v2016.3 and its multicast optimizations, devices may scan for IPv6 address as much as they want to without causing harm. For IPv4 scans, there is no solution yet, though. So my thoughts are currently, whether the IPv6 source filters should be removed, maybe?

---

I also started testing this patchset in a firmware, a few more notes:

- Looks like it needs an update for the move to LEDE for "include ../gluon.mk" -> "include $(INCLUDE_DIR)/package.mk"
- The source chain remains emtpy for me, I get the following errors when trying to "/etc/init.d/gluon-ebtables restart":
<code>
/usr/bin/lua: /lib/gluon/ebtables/110-source-allow-ipv6:5: unexpected symbol near '..'
/usr/bin/lua: /lib/gluon/ebtables/110-source-allow-ipv4:5: unexpected symbol near '..'
/usr/bin/lua: /lib/gluon/ebtables/110-source-allow-arp:4: unexpected symbol near '..'
/usr/bin/lua: /lib/gluon/ebtables/110-source-allow-arp:4: unexpected symbol near '..'
/usr/bin/lua: /lib/gluon/ebtables/110-source-allow-ipv4:5: unexpected symbol near '..'
/usr/bin/lua: /lib/gluon/ebtables/110-source-allow-ipv6:5: unexpected symbol near '..'
</code>

- There is an ACCEPT rule in MULTICAST_OUT_ICMPV6, causing accepted ICMPv6 multicast traffic to never be dropped by the source filter
  - If we were to keep the IPv6 rules then the easiest fix would be to move the source filter rules in front of the multicast ones

mweinelt:
We were just planning to resize our IPv4 prefix, but now we need to update all nodes first or else we won't be able to make use of the additional dhcp pools we allocated, as clients would end up with unusable addresses. :)

So maybe add this as an additional warning of real-world-fuckups to the package documentation.

kb-light:
@mweinelt but that would be the same for the gluon-next-node-IP (like the status-page and the DNS-cache)

rotanid:
what's the status here? would be nice to get this into the next release!

SimJoSt:
From reading the doc I don't understand what the package is doing exactly.
What kind of traffic is filtered? In what circumstances? Examples?
I like the doc for gluon-ebtables-filter-multicast. It has a clear explanation and exactly says what the package does.

kb-light:
Updates:

- filter rules are moved in front of the multicast rules 
- documentation improvements to specify the use case more clearly

package/gluon-ebtables-source-filter/files/lib/gluon/ebtables/100-source-chain
```diff
@@ -0,0 +1 @@
+chain('SOURCE', 'DROP')
```
NeoRaider: The chain name SOURCE is very inexpressive, it should in some way describe the packets that are passed to the chain. Maybe something like LOCAL_FORWARD.

package/gluon-ebtables-source-filter/files/lib/gluon/ebtables/110-source-allow-ipv6
```diff
@@ -0,0 +1,9 @@
+site = require('gluon.site_config')
+
+rule('SOURCE -p IPv6 --ip6-src fe80::/10 -j RETURN')
```
NeoRaider: I guess this can be changed to fe80::/64, no other link-local addresses should be used on Ethernet-like links like the mesh.

package/gluon-ebtables-source-filter/check_site.lua
```diff
@@ -0,0 +1,2 @@
+need_string_match('prefix4', '^%d+.%d+.%d+.%d+/%d+$', false)
+need_string_array_match('additional_prefix6', '^[%x:]+/%d+$', false)
```
NeoRaider: I think this should be called "additional_prefixes6", and we should have the same for IPv4 for consistency.

T-X: What would you guys think about "extra_prefix6" or "extra_prefixes6"? (so similar to EXTRA_CFLAGS/LDFLAGS, for instance)

rotanid:
hm, changes applied. but what happened to neoraiders idea of also having an "extra_prefixes4" ?

kb-light:
@rotanid we discussed that via IRC and have not find some reasonable usecase.

rotanid:
@kb-light maybe a community that wants to change their ipv4 subnet without disruption?

package/gluon-ebtables-source-filter/files/lib/gluon/ebtables/110-local-forward-allow-arp
```diff
@@ -0,0 +1,6 @@
+prefix4 = require('gluon.site_config').prefix4
+
+if prefix4 then
+   rule('LOCAL_FORWARD -p ARP --arp-ip-src ' .. prefix4 .. ' --arp-ip-dst ' .. prefix4 .. ' -j RETURN')
+   rule('LOCAL_FORWARD -p ARP --arp-ip-src 0.0.0.0 --arp-ip-dst ' .. prefix4 .. ' -j RETURN')
```
NeoRaider: One more thing I noticed: please use tabs for indentation here, too.

kb-light: done

PR #1018, opened 2017-01-29
[RFC] gluon-mesh-vpn-fastd: improve check_site

kb-light:
lua can not match `(ipv4|ipv6)?` or similar, so I decided to match `[ipv46 ]*`, one can also match `i?p?v?[46]?%s?` and I would change this PR accordingly if requested.

depends on #1016


NeoRaider:
This is just too weird and fuzzy. Either we add something like alternative matches to the *check_site_lib*, or we just don't change this. (Also, the match expression has some mistakes, a literal `.` must be escaped as `%.`, not `\.`)

kb-light:
- weird check for `'remotes'` removed
- rebased

PR #1026, opened 2017-02-06
Docs: site.conf - add Ad-Hoc ibss info as comments

rubo77:


rubo77:
I just found, that it is a bit documented at https://gluon.readthedocs.io/en/v2016.2.2/user/site.html

> Each interface may be disabled by setting disabled to true. This will only affect new installations. Upgrades will not changed the disabled state.

(although there is a typo "changed" should be "change")

So maybe this PR is not needed, although maybe it should be more visible, that you can add the disabled option everywhere.

maybe an extra commented line in https://gluon.readthedocs.io/en/v2016.2.2/user/site.html instead?

NeoRaider:
We've removed this section intentionally. New configurations should never use IBSS.

PR #1027, opened 2017-02-06
ramips-mt7628: add new platform and vocore2

ambassador86:
This change adds the VoCore2 to the devices which are gluon ready. The VoCore2 was also able to handle ibss, 802.11s and the client net at the same time. Unfortunately the stability is not so good. 
Test device: https://meshviewer.chemnitz.freifunk.net/#!v:m;n:b8d81267052b

### DIFF
--- a/targets/ramips-mt7628
+++ b/targets/ramips-mt7628
@@ -1,0 +1,4 @@
+# VoCore 2
+
+device vocore2 vocore2
+factory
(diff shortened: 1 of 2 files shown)


rotanid:
i'm puzzled, in your description you say that it was able to handle ibss, 11s and clients - but your commit adds it as BROKEN with a comment saying that ibss wasn't possible?
shouldn't the comment be like "poor stability" ?

ambassador86:
During my test this ibss0 interface came up together with mesh0 and client0 and after few seconds it disappeared. The batman chose the mesh0 for connecting to the other nodes.

rotanid:
did the mesh via mesh0 work good, performance-wise?
i'm asking because on mt7621 chips we saw packetloss

ambassador86:
My test was not long enough to get reliable data. But I will do some additional testing. At the moment the platform is marked as BROKEN so everyone who is trying to use it should be aware of the status. I simply created it because I have one of these little routers. At the moment I would not recommend to buy additional VoCore2 devices.

PR #1028, opened 2017-02-06
add debug-flag to build system, if enabled set CONFIG_DEBUG and do no…

christf:
…t strip binaries


christf:
this will make it much easier to build images containing debugging symbols to use gdb / valgrind in the environment where the programs work in.

For actually being able to use valgrind, glibc instead of musl should be selected, however I did not include this part in here. If that is interesting I will prepare a commit for this as well.

jplitza:
If I'm not mistaken, this PR only applies to the x86-64 target. I think it would be useful to have this option for all targets. (Personally, I don't test in a VM but on a WDR4300)

christf:
Yes this only applies to x86-64 at the moment because I did not build images for other targets and I was not sure if they would fit. But you are right that should not matter when building an option like this... The devices will contain more flash in the future. I will rework the patch in this spirit. 

Am 8. Februar 2017 23:01:24 MEZ schrieb Jan-Philipp Litza <notifications@github.com>:
>If I'm not mistaken, this PR only applies to the x86-64 target. I think
>it would be useful to have this option for all targets. (Personally, I
>don't test in a VM but on a WDR4300)

Viele Grüße

Christof
-- 
Diese Nachricht wurde mit K-9 Mail gesendet.

christf:
@jplitza how about that?

Makefile
```diff
@@ -99,6 +99,11 @@ config: FORCE
 		&& echo 'CONFIG_GLUON_RELEASE="$(GLUON_RELEASE)"' \
 		&& echo 'CONFIG_GLUON_SITEDIR="$(GLUON_SITEDIR)"' \
 		&& echo 'CONFIG_GLUON_BRANCH="$(GLUON_BRANCH)"' \
+		$(if $(DEBUG),&& echo 'CONFIG_DEBUG=y' \
+			&& echo 'CONFIG_NO_STRIP=y' \
+			&& echo '# CONFIG_USE_STRIP is not set' \
+			&& echo '# CONFIG_USE_SSTRIP is not set' \
+			&& echo '# CONFIG_STRIP_KERNEL_EXPORTS is not set') \
```
NeoRaider: Just add this stuff to `targets/generic` instead of cluttering the Makefile.

NeoRaider: CONFIG_STRIP_KERNEL_EXPORTS is never set, so you don't need to disable it.

Makefile
```diff
@@ -99,6 +99,11 @@ config: FORCE
 		&& echo 'CONFIG_GLUON_RELEASE="$(GLUON_RELEASE)"' \
 		&& echo 'CONFIG_GLUON_SITEDIR="$(GLUON_SITEDIR)"' \
 		&& echo 'CONFIG_GLUON_BRANCH="$(GLUON_BRANCH)"' \
+		$(if $(DEBUG),&& echo 'CONFIG_DEBUG=y' \
+			&& echo 'CONFIG_NO_STRIP=y' \
+			&& echo '# CONFIG_USE_STRIP is not set' \
+			&& echo '# CONFIG_USE_SSTRIP is not set' \
```
NeoRaider: Have you checked that all these 3 lines are necessary, and just setting `CONFIG_NO_STRIP=y` doesn't have the same effect?

NeoRaider:
I seems LEDE uses a variable *DEBUG* for some internal things, let's rename it to *GLUON_DEBUG* to avoid conflicts, and then add it to the export line in the main Makefile together with *GLUON_RELEASE* etc., so it doesn't need to be exported explicitly.

christf:
done.

Issue #894, opened 2016-09-29
Add support for TP-Link  TL-WA7210N


J0WI:
Support for the 5GHz model TP-Link TL-WA7510N has already been added in #681, but the 2.4 GHz model is not yet officially supported.
Support to OpenWRT has been added here: https://dev.openwrt.org/changeset/44526

NeoRaider:
I will add a profile for this device after the conversion from OpenWrt CC to LEDE is finished.

PR #1029, opened 2017-02-07
add support for TP-Link TL-WA7210N

txt-file:
fixes https://github.com/freifunk-gluon/gluon/issues/894

As far as I know the sysupgrade is broken on 7510N. I expect that it is broken on 7210N, too.

The patch has not been tested on a device.

### DIFF
--- a/targets/ar71xx-tiny
+++ b/targets/ar71xx-tiny
@@ -11,6 +11,8 @@ device d-link-dir-615-rev-c1 dir-615-c1 DIR615C1
 # TP-Link
 
 device tp-link-tl-wa701n-nd-v1 tl-wa701nd-v1
+
+device tp-link-tl-wa7210n-v2 tl-wa7210n-v2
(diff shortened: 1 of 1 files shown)


NeoRaider:
Do you have a reference for the broken sysupgrade, or was it the result of your own testing?

NeoRaider:
Looking at the relevant files (target/linux/ar71xx/base-files/lib/ar71xx.sh, target/linux/ar71xx/base-files/lib/upgrade/platform.sh), sysupgrade should work fine for both devices. Do you have any more specific information on this issue than "it is broken"?

txt-file:
sadly no. I have a TL-WA7510N at home but did not find the time to try
sysupgrade yet.

From a conservative point I would mark it as broken until somebody
proves to be working.

txt-file:
We tested a TL-WA7510N and it was able to autoupgrade from August 2016 to gluon-2016.2.3

I changed the pull request to only add TL-WA7210N without marking it as broken.

Issue #984, opened 2016-12-18
respondd should make use of address ff05::2:1001 for mesh-wide broadcasts and use ff02::2:1001 for talking to its neighbors


christf:
this will work for both batman and babel networks but enable respondd-usage in babel networks.

this affects https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-respondd/files/etc/init.d/gluon-respondd#L22

also it may be reasonable to extend respondd to handle teh distinction between ff02 and ff05

T-X:
For batman-adv meshes I would prefer sticking to ff02 for now. At least until MRD (multicast router discovery) snooping is implemented in the bridge + batman-adv.

christf:
does this mean we should split the package for now?

NeoRaider:
Please don't split the package. I'd really like to keep this consistent and switch the address to ff05::2:1001 for batman-adv as well.

I think the batman-adv multicast optimizations won't help here much; as long as respondd runs on all nodes, the package will need to reach every node anyways.

christf:
ok. I am using an upgrade script for now that modifies the initscript form respondd to obtain a testable babel firmware. This script can be removed easily once the address is changed globally

jplitza:
Can't we simply let respondd join both multicast groups?

christf:
>Can't we simply let respondd join both multicast groups?
At the moment it seems respondd does not support that.

jplitza:
No it doesn't, and implementing it doesn't seem to be as easy as I first thought. But that seems to be the cleanest solution to me.

christf:
so. what shall it be?
a) extend respondd to join multiple multicast-groups

christf:
b) use ff05 for batman and babel-networks?

christf:
... what is actually to be done before batman-based networks can switch to ff05?

T-X:
Ideally, the following would be done:

a) make the bridge join the "All-Snoopers-Multicast" address
b) make the bridge parse Multicast Router Advertisement and Multicast Router Termination messages and update its internal router list accordingly
c) make batman-adv parse the same messages to update its "WANT_ALL_IPV4" and "WANT_ALL_IPV6" multicast flags accordingly
d) then change a line in batman-adv to apply multicast optimizations for a multicast scope greater than link-local, too

Optionally, but recommended:

e) Make the bridge and/or batman-adv send Multicast Router Solicitations in some useful way.
[...]

PR #1030, opened 2017-02-08
have respondd work on ff05 in addition to ff02

christf:
This provides one possible solution for #984 

This implies that the hopglass configuration is also changed in the meshes to use ff05:... instead of ff02:..
This is already working with a hopglass instance in FFM

### DIFF
--- a/package/gluon-respondd/files/etc/init.d/gluon-respondd
+++ b/package/gluon-respondd/files/etc/init.d/gluon-respondd
@@ -13,7 +13,7 @@ start_service() {
 	local clientdevs=$(for dev in $(echo "$ifdump" | jsonfilter -e "@.interface[@.interface='$(cat /lib/gluon/respondd/client.dev 2>/dev/null)' && @.up=true].device"); do echo " -i $dev -t $MAXDELAY";done;)
 
 	procd_open_instance
-	procd_set_param command $DAEMON -g ff02::2:1001 $meshdevs $clientdevs -p 1001 -d /usr/lib/respondd -d /lib/gluon/respondd -g ff05::2:1001 $clientdevs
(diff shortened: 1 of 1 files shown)


rotanid:
@christf i guess you meant "hopglass-server", not "hopglass"?
nevertheless, this change will lead to problems depending on the way you plan to introduce it to hopglass-server - it has to work with both configurations because no one will update everything at once (gluon-respondd & hopglass-server) on all nodes.
... forgive me if i don't understand the suggested change and talk nonsense

NeoRaider:
We must only change the meshwide respondd listen address to ff05::..., not the individual mesh interfaces. neighbours-nodeinfo must not be changed.

christf:
if it is not changed, how will the respondd-node on the other node reply if it is subscribed to a different address?

jplitza:
Well, as discussed in #984, respondd has to be taught how to listen to multiple multicast addresses.

christf:
I did not think this was the outcome of the discussion. Hence this pr. 
Not changing neighbours-nodeinfo will result in missing data. 

Am 9. Februar 2017 19:55:51 MEZ schrieb Jan-Philipp Litza <notifications@github.com>:
>Well, as discussed in #984, respondd has to be taught how to listen to
>multiple multicast addresses.

Viele Grüße

Christof
-- 
Diese Nachricht wurde mit K-9 Mail gesendet.

NeoRaider:
neighbours-nodeinfo queries direct neighbours, not mesh-wide. The respondd multicast group on the individual mesh interfaces is supposed to stay ff02::.., so neighbours-nodeinfo must not be changed.

christf:
apparently repsondd did not need to be taught anything and it was just a matter of initialisation.

PR #1031, opened 2017-02-09
docs: fix typos in 2016.1 release notes

SimJoSt:
adds missing trailing `,` in simple_tc section and changes the order of ingress and egress to default format used by Gluon

### DIFF
--- a/docs/releases/v2016.1.rst
+++ b/docs/releases/v2016.1.rst
@@ -175,9 +175,9 @@ Site changes
         mesh_vpn = {
           ifname = 'mesh-vpn',
           enabled = false,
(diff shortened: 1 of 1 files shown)


NeoRaider:
I don't think fixing up old release notes makes much sense (the release announcements and Github releases link the tagged versions anyways), and none of these changes fix actual errors... I'll merge this because it doesn't hurt either, but it's not something to invest more time into.

SimJoSt:
I knew this from the last time I fixed something in a release doc, but there might be people like me searching for something and open old release files. With the next release you tag, it will be accessible.

PR #1004, opened 2017-01-14
ar71xx-generic: Fix flash layout and MAC addresses of Meraki MR12/MR16

oleeander:
Fixes #993. This has so far only been verified to work with MR16.

Flashing:
```
# Upgrade from LEDE with manually set MAC address (remove Xs)
# eXrXaXsXe 0xbf080000 +0x20000; cXpX.Xb 0xbffd0000 0x80010066 0x20000; cXpX.Xb 0x80010000 0xbf080000 0x20000

tftpboot 0x80010000 gluon-mr16-squashfs-sysupgrade.bin; erase 0xbf0a0000 +0xf40000; cp.b 0x80010000 0xbf0a0000 0xf40000
setenv bootcmd bootm 0xbfda0000; saveenv; boot
```

PR #1033, opened 2017-02-10
docs: add release notes for Gluon v2016.2.3

NeoRaider:


viisauksena:
sound bit strange not to make it 2017.1

NeoRaider:
@viisauksena, this is a minor release based on the v2016.2.x branch. v2017.1 will be the next major release.

rotanid:
i'm curious why you remove the support for meraki devices even though somebody created a PR #1004 that fixes the issues - i'm not using any meraki device, but fixing should be preferred.

NeoRaider:
@rotanid The problem is that we don't yet have a way to migrate from a broken Meraki installation to a working one - a sysupgrade will not lead to bootable system because of the changed flash layout.

PR #1036, opened 2017-02-11
gluon-radvd init remove netdev

christf:
remove netdev from initiscript as it is not necessary.

### DIFF
--- a/package/gluon-radvd/files/etc/init.d/gluon-radvd
+++ b/package/gluon-radvd/files/etc/init.d/gluon-radvd
@@ -8,7 +8,6 @@ start_service() {
     procd_open_instance
     procd_set_param command /usr/sbin/uradvd $(/lib/gluon/radvd/arguments)
     procd_set_param respawn ${respawn_threshold:-3600} ${respawn_timeout:-5} ${respawn_retry:-5}
-    procd_set_param netdev br-client
     procd_set_param stderr 1
     procd_close_instance
   }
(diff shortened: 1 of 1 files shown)


package/gluon-radvd/files/etc/init.d/gluon-radvd
```diff
@@ -4,12 +4,12 @@ USE_PROCD=1
...
 start_service() {
-  [ -x /lib/gluon/radvd/arguments ] && {
-    procd_open_instance
-    procd_set_param command /usr/sbin/uradvd $(/lib/gluon/radvd/arguments)
-    procd_set_param respawn ${respawn_threshold:-3600} ${respawn_timeout:-5} ${respawn_retry:-5}
-    procd_set_param netdev br-client
-    procd_set_param stderr 1
-    procd_close_instance
-  }
+	[ -x /lib/gluon/radvd/arguments ] &&
```
NeoRaider: I think it would be even nicer to change this into

    [ -x /lib/gluon/radvd/arguments ] || return 1

NeoRaider:
Merged manually.

jplitza:
FTR: I can't observe these restarts on 2016.2.3

NeoRaider:
@jplitza, yes, we concluded in IRC that the issue (frequent calls of respondd reload) was probably caused by something else flooding the network with router advertisements after all...

PR #1042, opened 2017-02-13
gluon-l3roamd: add package containing basic l3roamd configuration

christf:
this adds a package for configuring l3roamd for gluon.

### DIFF
--- a/package/gluon-l3roamd/luasrc/lib/gluon/upgrade/380-gluon-l3roamd-route
+++ b/package/gluon-l3roamd/luasrc/lib/gluon/upgrade/380-gluon-l3roamd-route
@@ -1,6 +1,6 @@
 #!/usr/bin/lua
 
-local uci = require('luci.model.uci').cursor()
+local uci = require('simple-uci').cursor()
 local site = require 'gluon.site_config'
 
 uci:delete('network', 'l3roam')
--- a/package/gluon-l3roamd/luasrc/lib/gluon/upgrade/380-gluon-l3roamd-route
+++ b/package/gluon-l3roamd/luasrc/lib/gluon/upgrade/380-gluon-l3roamd-route
@@ -4,21 +4,17 @@ local uci = require('simple-uci').cursor()
 local site = require 'gluon.site_config'
 
 uci:delete('network', 'l3roam')
-uci:section('network', 'interface', 'l3roam',
-{
+uci:section('network', 'interface', 'l3roam', {
 	ifname = 'l3roam0',
 	proto = 'none',
-}
-)
+})
 
 uci:delete('network', 'l3roamd_client')
-uci:section('network', 'route6', 'l3roamd_client',
-{
+uci:section('network', 'route6', 'l3roamd_client', {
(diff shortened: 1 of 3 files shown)


rotanid:
typo in commit message and title ;)

christf:
of course I will fix that typo. is there other input I should consider when working on this?

package/gluon-l3roamd/files/etc/modules.d/gluon-l3roamd
```diff
@@ -0,0 +1 @@
+cfg80211
```
NeoRaider: If mmfd does not work without *cfg80211* loaded, such a file should rather be part of the l3raomd package.

Does mmfd do anything useful with *cfg80211* when there are no WLAN adapters? I think it would be much better to fix mmfd to work correctly without *cfg80211* adding a hack to load the module.

NeoRaider: Ugh, replace "mmfd" by "l3roamd" or vice-versa in the above message, whatever actually depends on *cfg80211* at the moment...

christf: l3roamd soon will work without cfg80211 being loaded https://github.com/tcatm/l3roamd/pull/31
unfortunately it cannot know if cfg80211 is being loaded after it was started.
so either we should have l3roamd started by netifd or keep this hack... not sure which one is better...

NeoRaider: All kernel modules will be loaded very early in the boot process, we don't need any hacks.

christf: allright then, I will remove this line from my image an report back. I know I added when I was already well into the project but cannot remember the initial motivation.

christf: now I remember. I included this file to have l3roamd start up even in the x86 image where this module is not automatically loaded. Also l3roamd will not work without this module yet.
That means: For now I will move the file into the l3roamd-package. Once the PR is confirmed working and merged I will remove the modules-file.

christf: well it seems I removed it altogether already :)

package/gluon-l3roamd/luasrc/lib/gluon/upgrade/380-gluon-l3roamd-route
```diff
@@ -0,0 +1,24 @@
...
+local uci = require('simple-uci').cursor()
+local site = require 'gluon.site_config'
+
+uci:delete('network', 'l3roam')
+uci:section('network', 'interface', 'l3roam',
+{
+	ifname = 'l3roam0',
+	proto = 'none',
+}
+)
```
NeoRaider: Please adjust the coding style to match our other scripts: opening `{` on the `uci:section` line, closing `})` on a single line.

Issue #1032, opened 2017-02-09
uhttpd: identify if a limit of max_connections lower than 32 is a good idea


christf:
as seen today on #gluon having 32 max_connections leads quickly into oom-situations when multiple people are viewing the statuspage.

with 16 it was still possible to overload the test device (WR740v4, 32MB RAM), however it felt much more responsive until after a couple of minutes of runtime suddenly the load increased from 0.4 to above 7 and the device rebootet. A test with longer standing connections may show if there is a memory issue with that scenario.

christf:
T_X| wäre mal interessant, wieviele connections du hinbekommst, wenn du folgendes machst: $root@[host]:/sys/kernel/debug/ieee80211/phy1# echo "fq_memory_limit 1048576" > aqm


Ja, wäre es. setzt man das auf 1MB und gleichzeitig max_connections von uhttpd auf 16 ist der WR740 
durch Statuspagebetrachtung nicht mehr in den OOM zu treiben.


Was machen wir damit? Ein neues Paket fürs tiny-target (Geräte mit 32MB RAM), dass Speichereinstellungen dieser Art setzt?

christf:
on irc it was discussed that fq_memory_limit should be set as part of gluon-core on devices with little RAM.

PR #1045, opened 2017-02-15
gluon-status-page-api: limit uhttp max_requests to 16 on devices with <48MB RAM (part 1/2  #1032)

christf:
this will limit the uhttp max_requests to 16 saving even  slow tl-wr740 from being overloaded by statuspage-traffic.

This addresses one consideration in #1032

### DIFF
--- a/package/gluon-status-page-api/files/lib/gluon/upgrade/500-status-page-api
+++ b/package/gluon-status-page-api/files/lib/gluon/upgrade/500-status-page-api
@@ -8,6 +8,15 @@ uci -q batch <<-EOF
 	delete uhttpd.main.listen_https
 
 	set uhttpd.main.home=/lib/gluon/status-page/www
-
-	set uhttpd.main.max_requests=32
 EOF
+
+RAM=$(grep MemTotal /proc/meminfo |awk '{print $2}')
+if [ $RAM -lt $((48*1024)) ]; then
+	echo "set uhttpd.main.max_requests=16"
+else
+	echo "set uhttpd.main.max_requests=32"
+fi |uci -q batch
+
+if [ -x /etc/init.d/rpcd ]; then
+	/etc/init.d/rpcd disable
+fi
--- a/package/gluon-status-page-api/files/lib/gluon/upgrade/500-status-page-api
+++ b/package/gluon-status-page-api/files/lib/gluon/upgrade/500-status-page-api
@@ -10,12 +10,12 @@ uci -q batch <<-EOF
 	set uhttpd.main.home=/lib/gluon/status-page/www
 EOF
 
-RAM=$(grep MemTotal /proc/meminfo |awk '{print $2}')
+RAM=$(awk '/MemTotal/ {print $2}' /proc/meminfo)
 if [ $RAM -lt $((48*1024)) ]; then
-	echo "set uhttpd.main.max_requests=16"
+	uci set uhttpd.main.max_requests=16
 else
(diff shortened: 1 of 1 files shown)


mweinelt:
This should only be applied to the `ar71xx-tiny` target in my opinion.

christf:
strictly speaking the tiny-target refers to devices with little flash. While these devices usually have little ram as well there is no direct dependency.

How about checking the amount of RAM and applying this setting only if less than 50MB is found the way I am preparing this with fq_maxmem patches?
That being said, RAM is not the only consideraton to take into account, CPU may become a bottleneck as well.

@mweinelt  could you try overloading the statuspage of a device with enough ram to run 32 requests of uhttpd?

christf:
a 1043v2 was tried with max_connections=32 it had a peak load of .4 so it seems that considering RAM is sufficient. The patch was adapted accordingly.

SimJoSt:
typo in the title and commit message: guon -> gluon

rotanid:
@christf regarding -tiny i agree with you, there are devices with >4mb flash but only 32mb ram.

package/gluon-status-page-api/files/lib/gluon/upgrade/500-status-page-api
```diff
@@ -8,10 +8,15 @@ uci -q batch <<-EOF
 	delete uhttpd.main.listen_https
 
 	set uhttpd.main.home=/lib/gluon/status-page/www
-
-	set uhttpd.main.max_requests=32
 EOF
 
+RAM=$(grep MemTotal /proc/meminfo |awk '{print $2}')
```
NeoRaider: awk '/MemTotal/ {print $2}' /proc/meminfo

package/gluon-status-page-api/files/lib/gluon/upgrade/500-status-page-api
```diff
@@ -8,10 +8,15 @@ uci -q batch <<-EOF
...
+if [ $RAM -lt $((48*1024)) ]; then
+	echo "set uhttpd.main.max_requests=16"
+else
+	echo "set uhttpd.main.max_requests=32"
+fi |uci -q batch
```
NeoRaider: Using `uci batch` for a single set command doesn't make much sense...

christf: done.

package/gluon-status-page-api/files/lib/gluon/upgrade/500-status-page-api
christf: that is much better. thanks. done.

package/gluon-status-page-api/files/lib/gluon/upgrade/500-status-page-api
```diff
@@ -8,6 +8,15 @@ uci -q batch <<-EOF
...
+RAM=$(awk '/MemTotal/ {print $2}' /proc/meminfo)
+if [ $RAM -lt $((48*1024)) ]; then
+	uci set uhttpd.main.max_requests=16
+else
+	uci set uhttpd.main.max_requests=32
+fi
+
+if [ -x /etc/init.d/rpcd ]; then
+	/etc/init.d/rpcd disable
+fi
```
NeoRaider: You accidentially added these removed lines again when rebasing.

christf: done.

PR #1046, opened 2017-02-15
gluon-core: limit memory that may be used by fq_codel to 1MB per AP o…

christf:
…n devices with less than 48MB RAM to avoid OOM

### DIFF
--- a/package/gluon-core/files/etc/hotplug.d/ieee80211/01-gluon-core-codel-memusage
+++ b/package/gluon-core/files/etc/hotplug.d/ieee80211/01-gluon-core-codel-memusage
@@ -1,7 +1,7 @@
 #!/bin/sh
 if [ "${ACTION}" = "add" ]; then
 
-	RAM=$(grep MemTotal /proc/meminfo |awk '{print $2}')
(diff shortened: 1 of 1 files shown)


christf:
this addresses the second concern of #1032

mweinelt:
Can you possibly reference the discussion back when it was lowered to 4MB in OpenWrt? According to @T-X the fq_codel upstream developers had many concerns.

christf:
@mweinelt I was not able to find it. @T-X can you help here?

package/gluon-core/files/etc/hotplug.d/ieee80211/01-gluon-core-codel-memusage
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+if [ "${ACTION}" = "add" ]; then
+
+	RAM=$(grep MemTotal /proc/meminfo |awk '{print $2}')
```
NeoRaider: Another grep+awk that can be simplified.

christf: right you are. done.

PR #1050, opened 2017-02-21
move lua files into luasrc dirs and add GluonSrcDiet to Makefile

2tata:
I moved forgotten luafiles into luasrc dirs and prepare affected Makefiles with GluonSrcDiet. (for master)


NeoRaider:
gluon-node-info has been kept like this because it will probably be moved out of Gluon into packages as soon as we have an idea how to name it. The gluon-mesh-batman-adv-core change looks okay.

NeoRaider:
I applied the gluon-mesh-batman-adv-core patch, thanks.

PR #1055, opened 2017-02-24
docs: fixed some grammar

rubo77:


NeoRaider:
Thanks, I've applied the first change, the rest doesn't really seem wrong to me (the list points before the colons aren't sentences, and many of the parts after the colons aren't either, so there aren't any definite rules anyways). Also, you added an additional typo in one of your fixes...

Issue #863, opened 2016-08-27
respondd suddenly not running anymore


rotanid:
we observed some cases, where respondd wasn't running anymore suddenly.
on the nodes where we have remote access, there doesn't seem to be any log entry regarding it's stop or crash in dmesg or logread.

if you have remote access, the issue can be fixed by simply starting gluon-respondd again.

how do you know that respondd is down without remote access:
- if you use meshviewer, the node still appears as online but with less information, for example client cound and uptime are missing.
- if you use respondd data (like with hopglass), the node appears as offline.

the changes in https://github.com/freifunk-gluon/packages/pull/143 regarding the init file may workaround this issue as the new initfile contains the "respawn" option.

rotanid:
we created a workaround-package to make sure respondd doesn't stay not-running: https://github.com/tecff/gluon-packages/tree/master/tecff-respondd-watchdog

AKA-47:
I found the same problem today with 2016.2 but was not able to debug it. Will try your package

jplitza:
#972 improved the initscript so that stderr is logged and respondd automatically restarted. It would be great if anybody experiencing unexpected stops could try that new initscript and look out for messages in logread.

rotanid:
not easy to test, i will have to backport it to v2016.2.x, remove my workaround, deploy it to the experimental-branch nodes and wait at least a week, the longer the better, if one of them will have the issue again.
as a late comment to @AKA-47 , in our network it appears with v2016.1.x, too (assuming you're talking about gluon version numer, not badman-adv)

rotanid:
@jplitza i backported your commit d8bb97831b197c543d9727f1c81539bb6fba127a together with ca57cdfe77b73351708f3b18235f508a42bccac9 (made it very easy, porting both) to v2016.2.x and removed my cronjob/watchdog.
running it on one node atm, will be running on ~60 nodes soon.
we'll have to wait a few weeks to make a somewhat strong statement about the effect.

kpanic23:
By the way, the same seems to happen with alfred. (Yes, I know, we're still using alfred...)

rotanid:
@jplitza so far (3 weeks) it didn't appear again on these ~60 nodes running with the v2016.2.x branch patched with the two mentioned commits. maybe backporting this @NeoRaider ?

NeoRaider:
I backported the relevant patches to v2016.2.x.

bobcanthelpyou:
[...]

PR #1062, opened 2017-03-07
gluon-respondd: Use hotplug script for reload

jplitza:
Before, we used a procd netdev trigger, but that fired *way* too often (maybe once per router announcement received?), causing high load for nothing.

This basically reverts to the behaviour before commit d8bb978, only that the init script still handles argument collection.

### DIFF
--- a/package/gluon-respondd/files/etc/hotplug.d/iface/10-gluon-respondd
+++ b/package/gluon-respondd/files/etc/hotplug.d/iface/10-gluon-respondd
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+[ "$ACTION" = "ifupdate" ] && exit
+
+exec /etc/init.d/gluon-respondd reload
--- a/package/gluon-respondd/files/etc/init.d/gluon-respondd
+++ b/package/gluon-respondd/files/etc/init.d/gluon-respondd
@@ -18,12 +18,3 @@ start_service() {
 	procd_set_param stderr 1
 	procd_close_instance
 }
-
-service_triggers() {
-	local script=$(readlink "$initscript")
-	local name=$(basename ${script:-$initscript})
-
-	procd_open_trigger
-	procd_add_raw_trigger "interface.*" 0 "/etc/init.d/$name" reload
-	procd_close_trigger
-}


package/gluon-respondd/files/etc/hotplug.d/iface/10-gluon-respondd
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+exec /etc/init.d/gluon-respondd reload
```
NeoRaider: This exec will break things, hotplug scripts are sourced, not executed.

jplitza: It won't, trust me.

1. I tested this exact thing by creating scripts that all echo something and then `exec true`, they were all executed in order.
2. `/etc/hotplug.d/net/10-ar922x-led-fix` ends with `exit 0`

I have no idea what magic hotplug-call does with these scripts, as they don't need to be executable either, but they surely are not sourced.

NeoRaider: Ah, I just had a closer look at `base-files/files/sbin/hotplug-call`, they are sourced in a sub-shell.

rotanid:
and this change doesn't reintroduce #863 , if i'm reading the changes correctly?

jplitza:
@rotanid Correct, because procd still takes care of restarting the process if it dies. It just doesn't automatically reload it on update from netifd.

NeoRaider:
@jplitza Btw., have you checked how often the new script is called? I'd expect netifd calling the hotplug handlers for each RA, too (with `ACTION=ifupdate`), so we might want to limit this to *ifup* and *ifdown* actions (see https://git.lede-project.org/?p=project/netifd.git;a=blob;f=interface-event.c;h=4976c2cf032d6b3bc25069f03398aefd454f3d50;hb=HEAD#l33 for a list of possible actions). I'm not sure when netifd triggers *ifupdate* events though, and if we also need to check for *iflink* events...

jplitza:
Welp, you got me. I only had tested an earlier version where the script was in `hotplug.d/net/`. Yes, the ifupdate hotplug also fires way too often, so I blacklisted it.

NeoRaider:
New issue: the reload command will start respondd even when it was not running.

jplitza:
Fixed instead by backporting a change to netifd in 9a0aeb9b7482df4e4515e61356b9d393e3a7eacb

NeoRaider:
Superseded by 9a0aeb9b7482df4e4515e61356b9d393e3a7eacb.

PR #1071, opened 2017-03-16
fixed: image file name creation for tp-link-tl-wr940n

Tarnatos:
The WR940n doesn't has detachable antennas.


kb-light:
Same as already mentioned in #1072: I would expect this commit to break the autoupdater.

NeoRaider:
In this case, the autoupdater is unaffected, as the 940v1-v3 identify as 941v4-v6, so these image names are never used for updates.

PR #1072, opened 2017-03-16
fixed: image file name creation for tp-link-tl-wr941nd

Tarnatos:
The WR941 only has detachable antennas.

### DIFF
--- a/targets/ar71xx-tiny
+++ b/targets/ar71xx-tiny
@@ -47,16 +47,16 @@ factory -squashfs-factory${GLUON_REGION:+-${GLUON_REGION}} .bin
 
 device tp-link-tl-wr843n-nd-v1 tl-wr843nd-v1
 
-device tp-link-tl-wr941n-nd-v2 tl-wr941nd-v2
-device tp-link-tl-wr941n-nd-v3 tl-wr941nd-v3
+device tp-link-tl-wr941-nd-v2 tl-wr941nd-v2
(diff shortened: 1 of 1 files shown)


targets/ar71xx-tiny
```diff
@@ -47,16 +47,16 @@ factory -squashfs-factory${GLUON_REGION:+-${GLUON_REGION}} .bin
 
 device tp-link-tl-wr843n-nd-v1 tl-wr843nd-v1
 
-device tp-link-tl-wr941n-nd-v2 tl-wr941nd-v2
-device tp-link-tl-wr941n-nd-v3 tl-wr941nd-v3
+device tp-link-tl-wr941-nd-v2 tl-wr941nd-v2
```
mweinelt: 941-nd? 941nd?

kb-light:
@Tarnatos did you have some autoupdater issues on these devices that are fixed by this commit?

You know, that it does not depend on what TP-Link has build but on the return value of: `lua -e 'print(require("platform_info").get_image_name())'`. And for a TL-WR941ND it says `tp-link-tl-wr941n-nd-v6`, additional we had no autoupdater issues on v2016.2.x with TL-WR94xN/ND.

In my opinion this PR should be closed because it would break the autoupdater.

NeoRaider:
Indeed, we can't change these image names. They are generated from the strings defined in *tplink_board_detect* in `/lib/ar71xx.sh`.

PR #1074, opened 2017-03-17
Docs: fix missing files for make in 2016.2.x

rubo77:
This is just a first stub, I think we should host the needed files in the gluon organisation and it seems there are a lot of those files missing: https://github.com/Freifunk-Nord/gluon-build-dependencies/tree/master/openwrt/dl


NeoRaider:
This issue has been fixed in v2016.2.4.

NeoRaider:
If there are other missing files, we should fix the code and not document workarounds.

rubo77:
OK, I found out, that we didn't checkout the right tag 2016.2.4. the errors only occurred 2016.2.1 tag, so there doesn't seem to be needed anything done here

PR #1075, opened 2017-03-18
Gluon-mesh-vpn-*: Fixed bugs in new implementation

lcb01a:
- Gluon-mesh-vpn-core: Moved migration code to its own script as the initial VPN daemon config needs to be in place first when migrating
- gluon-mesh-vpn-tunneldigger: Added check to ignore default config when checking for legacy configs

The migration code ran before the initial config file for either fastd or tunneldigger was written, thus disabling the daemon completely when upgrading.

### DIFF
--- a/modules
+++ b/modules
@@ -2,7 +2,7 @@ GLUON_FEEDS='openwrt gluon routing luci'
 
 LEDE_REPO=git://git.lede-project.org/source.git
 LEDE_BRANCH=lede-17.01
-LEDE_COMMIT=4a405ac8f956b725de037744d62497f93bba6080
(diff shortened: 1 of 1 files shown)


NeoRaider:
I've pushed the first fix, I'm not so sure about the second one.

I've changed the migration code to enable the VPN when either of the VPNs was enabled before; this should be set before the rest of the configuration is created, so the defaults from *site.conf* aren't applied. So, either the migration code must run first and create skeleton sections with the *enabled* value itself, or we must move the application of the *site.conf* setting to the migration script as well. I think I'd prefer the latter solution.

PR #1078, opened 2017-03-23
patch for iw to enable scan output for IEEE802.11s

2tata:


NeoRaider:
We don't really need the mesh cabapilities, do we? The mesh ID should be enough AFAICT.

2tata:
On 03/23/17 12:26, Matthias Schiffer wrote:
> We don't really need the mesh cabapilities, do we? The mesh ID should be enough AFAICT.
> 

Right, the cabapilities can be dropt. For the hoodselector, ID is enough.

cheers
Tarek

2tata:
done

cheers
Tarek

NeoRaider:
Thanks, I've pushed a similar change to LEDE and updated *modules*. We really want to avoid adding patches in Gluon when we can.

PR #1080, opened 2017-03-25
gluon-mesh-vpn-core: Fixed conditions for migration code

lcb01a:
VPN daemons were always disabled

This fixes the issue that was left in the previous pull request without moving the code to a new file or creating skeleton configs.

has_fastd and has_tunneldigger were incorrectly  used twice in defining the enabled variables and when setting the final value in UCI.


NeoRaider:
Thanks, but this fix is not enough. The uci set won't work unless the config sections already exist; I described two possible solutions in #1075. I might find time to write a proper fix myself today though.

lcb01a:
hmm in my tests the uci setting was ignored if it was not in place yet, basically using uci:set without adding the section first will just do nothing.

PR #1082, opened 2017-03-27
modules: use newest package-repository to include the latest version …

christf:
…of l3roamd which supports an RO unix-socket


NeoRaider:
Please update to the correct hash, you couldn't know it before the the package PR was merged...

christf:
I wasn't aware that the commit hash would change when merging a commit from a PR.

Issue #1088, opened 2017-04-07
Gluon (Lede) ipq806x and Archer C2600


seligenthaler:
@NeoRaider 
At your request

Hardware: TP-Link ARCHER C2600 v1.1
Gluon: v2016.2-244-g13c61d9

If you reboot the router there is a shutdown with all LED going off.
You must use the power switch to bring it back.

Error in config- mode wifi:
>500 Interner Serverfehler
>Entschuldigung, auf dem Server ist ein unerwarteter Fehler aufgetreten.
>Failed to execute dispatcher target for entry '/admin/wifi-config'.
>The called action terminated with an exception:
>/lib/gluon/web/model/admin/wifi-config.lua:19: bad argument #1 to 'txpwrlist' (string expected, got nil)

After the config is done and after rebooting (manualy s.a.) the router comes up and is accessible by SSH.
Alfred and batman did not work.  (lack of bat0)

The client0 (5GHz) is on purpose manualy disabled.
In (/sys/class/net) the virtual bat0-device and primary0-device are missing.
The  bat0-device section in /etc/config/network is available.

Just for testing and information:
You can use the command "batctl if add client1" (does this make sense?)
This will create a temporary virtual device bat0
Then you can start alfred
/etc/init.d/alfred start
>/etc/init.d/alfred: waiting 30 secs for br-client address...
>/etc/init.d/alfred: starting alfred
>/etc/init.d/alfred: starting batadv-vis

any suggestions?

documents attached:  network wireless logdata (UPDATE v2016.2-244-g13c61d9)
[c2600.zip](https://github.com/freifunk-gluon/gluon/files/906989/c2600.zip)

belzebub40k:
I also recognized this with the mvebu target. I assume it is because the device does not support IBSS or 802.11s but I didn't had time to look deeper into the issue yet.

belzebub40k:
Found the problem for the Linksys WRT1200AC (mvebu). It addresses the radios by its PCIe path in `wireless.radio(0|1).path`

```
# uci show wireless | grep .path
wireless.radio0.path='soc/soc:pcie-controller/pci0000:00/0000:00:01.0/0000:01:00.0'
wireless.radio1.path='soc/soc:pcie-controller/pci0000:00/0000:00:02.0/0000:02:00.0'
```

[...]

PR #1089, opened 2017-04-09
gluon-web-wifi-config: Fix PHY detection

belzebub40k:
Fix for  #1088. Tested on Linksys WRT1200AC and TP-Link TL-WR841N-v11.

### DIFF
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
@@ -138,6 +138,10 @@ local function find_phy_by_path(path)
 	for phy in fs.glob('/sys/devices/' .. path .. '/ieee80211/phy*') do
 		return phy:match('([^/]+)$')
 	end
+
+	for phy in fs.glob('/sys/devices/platform/' .. path .. '/ieee80211/phy*') do
+		return phy:match('([^/]+)$')
+	end
 end
 
 local function find_phy_by_macaddr(macaddr)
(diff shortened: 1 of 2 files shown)


NeoRaider:
Hmm, I think what LuCI and some other scripts do is to simply try the path once with "platform", and fall back to the path without if that does not work. We should probably do the same.

belzebub40k:
I've updated the PR as requested. This was my initial solution but I thought you prefer code with less duplicates ...

NeoRaider:
Speaking of duplicate code, it would be a good idea to make *find_phy_by_path* and *find_phy_by_macaddr* non-local in gluon.util and remove those functions from the gluon-web-wifi-config model.

belzebub40k:
Updated PR to use `find_phy()` from _gluon-core_.

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -134,10 +134,14 @@ function get_mesh_devices(uconn)
 	return devices
 end
 
-local function find_phy_by_path(path)
+function find_phy_by_path(path)
```
NeoRaider: Wrong function: you use *find_phy*, but removed the `local` for *find_phy_by_path*.

belzebub40k: Fixed

NeoRaider:
Thanks, I've pushed your changes with some additional refactoring.

belzebub40k:
Thanks. Didn't had time to make the requested changes yesterday.

PR #1090, opened 2017-04-10
docs: improve sidebar structure

belzebub40k:
The current structure of the _gluon-web_ documents creates a very unintuitive menu on the left (see screenshot). And as "Config Mode" already exists we now have to sections with the same name. To solve this I merged it in one file.

I also added a missing link to the _gluon-ebtables-source-filter_ package which was only discoverable via the search function.

## Before
docs

## After
docs_new


NeoRaider:
Meh... the menu is much better with this patch, but I don't really like having excessively long pages just to fix the menu. Maybe Sphinx has another way to structure the sidebar page list?

belzebub40k:
You are right. Now it looks like this.

docs

PR #1094, opened 2017-04-13
gluon-alt-esc: add client + provider package

T-X:
The client package allows to create a new wifi interface with direct access
to the local Freifunk network but with alternative exit gateways for internet
connectivity. A custom ESSID can be chosen.

The provider package is the counterpart to the client package and
configures the firewall of the selected Gluon node to grant permission to
route packets between the client and wan zone.

Note that the Gluon Alt-ESC provider package is not mandatory for the
Gluon Alt-ESC client package. In fact, any client device in the mesh network
can be chosen and configured to provide internet access for the Alt-ESC
client package.


T-X:
**Pictures for provider side:**

01-altesc-provider-off

02-altesc-provider-stilloff

03-altesc-provider-on

**Pictures for client side:**

11-altesc-client-off

12-altesc-client-on

13-altesc-client-onv4only

T-X:
I guess the question will arise, so trying to answer it beforehand:

**Why an extra ESSID?**

Two reasons:

1. Friendly co-existence with neighboring nodes: Avoiding roaming issues, not everyone one wants to use $your potentially broken gateway.
-> The idea is to keep the option to always choose the community maintained gateways
2. But also de-etanglement concerning the (partial) commercial interests which arose in the past years. This would allow hotels, cafes, backeries etc. to choose an alternative exit independent of the community infrastructure (as long as they do not use the word "Freifunk" in the new, extra ESSID) while still contributing to it. Such a cleaner separation might become necessary if the charitable nature of Freifunk were to be accepted by the German government?

package/gluon-alt-esc-client/Makefile
```diff
@@ -0,0 +1,38 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-alt-esc-client
+PKG_VERSION:=1
+PKG_RELEASE:=1
+
+PKG_BUILD_DIR := $(BUILD_DIR)/$(PKG_NAME)
+
+include $(INCLUDE_DIR)/package.mk
```
belzebub40k: This needs to be `include ../gluon.mk` with LEDE based builds. Same applies to the provider package.

belzebub40k:
_/**var**/gluon/firewall.wan-nat6_ should be located in _/**lib**/gluon/firewall.wan-nat6_ or even better _/**lib**/gluon/alt-esc/iptables.rules_ like in [gluon-mesh-vpn-core](https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-mesh-vpn-core/files/lib/gluon/mesh-vpn/iptables.rules)

https://github.com/freifunk-gluon/gluon/pull/1094/files#diff-d015fd309cc74651d3e87ac021f94017R66

package/gluon-alt-esc-client/luasrc/lib/gluon/web/model/admin/altesc-client.lua
```diff
@@ -0,0 +1,130 @@
...
+
+  uci:commit('gluon-alt-esc-client')
+
+  i=0
+  util.iterate_radios(
+    function(radio, index)
+      local name = "altesc_" .. radio
+
+      if enabled.data then
+        local macaddr = util.get_wlan_mac(radio, index, 4)
```
belzebub40k: `get_wlan_mac()` changed and now needs the UCI config as first parameter https://github.com/freifunk-gluon/gluon/commit/c4613c4e8d6b888d6402818aeb4e8656959e5fcf

package/gluon-alt-esc-client/luasrc/lib/gluon/web/model/admin/altesc-client.lua
```diff
@@ -0,0 +1,130 @@
...
+
+  uci:set('gluon-alt-esc-client', client, 'disabled', enabled.data and '0' or '1')
+  uci:set('gluon-alt-esc-client', client, 'exit4', exit4.data or '')
+  uci:set('gluon-alt-esc-client', client, 'exit6', exit6.data or '')
+  uci:set('gluon-alt-esc-client', client, 'altesc_on_lan', altesc_on_lan.data and '1' or '0')
+
+  uci:commit('gluon-alt-esc-client')
+
+  i=0
+  util.iterate_radios(
```
belzebub40k: `iterate_radios()` changed and now needs the UCI config as additional parameter https://github.com/freifunk-gluon/gluon/commit/f48d10bdde46bf67f982115d93ae487b5da7a3e0

jplitza:
Wow. That's huge, both technically (haven't had a closer look yet but am baffled how this works) and politically. Thank you so much!

Politically, I partially disagree with your reasoning for a (forced) separate ESSID. I would vote for the option to be a regular gateway, unset by default, as this still satisfies argument 2 I think, and doesn't blindly force your gateway onto others by default. That first argument actually is quite weak IMO, because it is technically possible to set up a gateway that will be chosen by neighbors (and without radv-filterd even by everybode in the mesh in the case of IPv6).

package/gluon-alt-esc-client/files/usr/sbin/altesc-update
```diff
@@ -0,0 +1,34 @@
...
+disabled=`uci get gluon-alt-esc-client.@client[0].disabled`
+
+if [ ! "$disabled" = "0" ]; then
+	ebtables -t nat -F ALT_ESC_STDGW
+	exit 0
+fi
+
+
+get_gw_hosts() {
+	for gw in `batctl gwl | tail -n +2 | sed "s/.* \(..:..:..:..:..:..\) (.*/\1/"`; do
```
belzebub40k: Since batman-adv 2017.0 the header of `batctl gwl` consists of 3 lines.

SimJoSt:
As the English language, contrary to the German language, likes to repeat itself, I would recommend:
> I am brave and *I* know what I am doing.

package/gluon-alt-esc-client/files/usr/sbin/altesc-update
belzebub40k: Also the output of `batctl tg` changed completely. The following works for my.

```
get_gw_hosts() {
	for gw in `batctl gwl | tail -n +3 | sed "s/.* \(..:..:..:..:..:..\) (.*/\1/"`; do
		batctl tg | egrep "^ \* .*$gw"
	done | tr -s ' ' | sed -r -e 's/\( +/\(/g' | cut -d ' ' -f 3 | sort | uniq
}
```

Adorfer:
i would suggest a renaming of this package to BoSE (S stands for steel, E for Exit).
(no, this was not completely serious.)

T-X:
Thanks for all the feedback and eager bug hunting so far! Found one more and I hope they are fixed now (@belzebub40k, can you check that the new regex-ing works for you and your compat15 setup, too?).

T-X:
@jplitza: Thanks for your valuable input! I like the idea of extending Alt-ESC with an option to provide an exit mesh wide for the standard ESSID. Yes, that should satisfy argument 2.

However, so far I couldn't think of a way to use the DNAT-ing trick for that without causing potential roaming issues for the standard ESSID. If you have an idea, maybe we can discuss that on IRC or the Gluon mailinglist?

Adorfer:
If i understand correct, this will only affect (ap-)clients of the local node, correct?
In other words: clients in my local mesh network will still use the "domain wide" gateways. 

If so, i would preferr an option, to signalize this exit with a very low bandwidth, in plus putting some additional hop penalty on the meshvpn link, so that clients in the local mesh use this local exit. 
(if somebody want's to make it more complex, a 2nd batman mesh network using v(x)lan on wifi and ethernet may be a solution in "keep up the roadming", in order to have 2 different client ssids entirely in the local mesh cloud.

belzebub40k:
@T-X the sed statement to get the MACs from transglobal still does not work. With `sed "s/^ \* \(..:..:..:..:..:..\) .*/\1/"` I get the correct value.

rubo77:
Since the Störerhaftung for Freifunk seems to be gone now, how about this brilliant PR?

Adorfer:
does this exist as an "optional package" somewhere for simply integration via site.mk?

rubo77:
Is there an easy way to use this as external package, that we could add to our site.mk and modules?

rubo77:
Ich hab das mal einfach in ein git kopiert, mal sehen ob man das so bauen kann:

https://github.com/rubo77/gluon-alt-esc

package/gluon-client-bridge/luasrc/lib/gluon/upgrade/310-gluon-client-bridge-local-node
```diff
@@ -19,8 +19,7 @@ uci:section('network', 'device', 'local_node_dev', {
 local ip4, ip6
 
 if site.next_node.ip4 then
-	local plen = site.prefix4:match('/%d+$')
-	ip4 = site.next_node.ip4 .. plen
+	ip4 = site.next_node.ip4 .. '/32'
```
rubo77: Why did you remove `site.prefix4:match('/%d+$')` ?

rubo77:
ich habe einfach die upgrade Pakete umbenannt: die Nummer einen höher jeweils, ist zwar dann doppelt drauf, aber so kann man die upgrade scripte in dem externen Paket aufrufen

Jetzt baut die Firmware auch...

package/gluon-client-bridge/luasrc/lib/gluon/upgrade/310-gluon-client-bridge-local-node
T-X: The br-client needs a more specific route than the local-node one (/24 vs. /32) as otherwise the kernel would select the wrong interface for outputting de-nat'ed packets coming back.

rubo77:
I changed the texts, so they are more intuitive for not so advanced users: https://github.com/rubo77/gluon-alt-esc/commit/e6c52469fa69811782b29d2b85deb89dba1a205c#diff-68586350a587e33d8e4ec7b3ca599fbdR43

Maybe you can take these changes into account in your PR?

jerix777:
This is a great job ! how to load it into gluon ?

rubo77:
@jerix777 your paste is not accessible. what is your question?

rubo77:
@T-X: Good Idea! You pointed out in IRC

1. that we should use the long texts "Mac of the node, that serves as..." should be placed as hints below the fields. the same style like "(unchangeable,...)"

2. the wording "bypassing the Freifunk-Gateeways" must be replaced with "bypassing the community gateways"

3. "via this node via Alt-ESC-Client" add a "for instance"

rubo77:
I updated my standalone package at https://github.com/rubo77/gluon-alt-esc

New screenshots:

auswahl_127


auswahl_126

Tarnatos:
Nice +1

T-X:
- Rebased to master
- Added German translation
- Updated texts with @rubo77's suggestions (with some small adjustments)

package/gluon-alt-esc-client/i18n/de.po
```diff
@@ -0,0 +1,49 @@
...
+"Hier kannst du ein WiFi Interface mit einem alternativen Gateway für "
+"dessen Internetzugang hinzufügen. Normalerweise verbindest du dich mit "
+"einem Knoten, welcher das Alt-ESC-Provider Paket aktiviert hat (wobei "
+"auch andere Systeme Zugang bieten können)."
+
+msgid "Name (SSID)"
+msgstr "Name (SSID)"
+
+msgid "Example: \"Lisa's Garden Gate\""
+msgstr "Beispiel: \"Lisas Gartentor\""
```
rubo77: shouldn't we take an SSID that sounds like Freifunk as example? e.g. "Lisas Garten Freifunk"

T-X:
@rubo77 fair question. I'm just a little paranoid that people might abuse the word "Freifunk" in an inappropriate way if having "Freifunk" in the suggestion. I was actually thinking about implementing a blacklist option for words in site.conf to be able to forbid "Freifunk" in this ESSID all together.

Currently, I would be OK-ish if people were using the ESSID for businesses. For instance a cafe, backery, hotel or hostel.  Even if they were inserting a splash page in between. Even if they were handing out access tokens. I also expect that many people will (need to, "thanks" to the new German laws,) use it with filter mechanisms provided by the guest access of their FritzBox. But in all these cases while acknowledging these use-cases, I would not like to see the word "Freifunk" in the ESSID then.

Kind of what I tried to say [in my previous comment](https://github.com/freifunk-gluon/gluon/pull/1094#issuecomment-293881395). Does that make sense?

rubo77:
But what if you just enable it without any filters and splash pages. That's what I will do at home. And then it is still "real" Freifunk.

Majbe we should just give an explanation how Freifunk should stay free of such restricted access methods if someone tries to name the Ssid something with"Freifunk"

T-X:
@rubo77: But maybe even then that's not Freifunk - the Freifunk memorandum could be interpreted in a way that "real Freifunk" may only refer to the intranet, the mesh, as the internet is mostly unfree these days. You see, that's why I wanted to avoid suggesting "Freifunk", because that might spark quite some discussions ;-).

I nevertheless like your idea of a pop-up regarding the naming, informing the user what Freifunk is about, and when not using that word. I've never implemented such a pop-up for LuCI-ish pages, don't know whether that's easily doable right now. Therefore maybe a nice improvement for later? (maybe you would like to have a shot at it?)

T-X:
Changelog for the last push:
* Rebased to master
* gluon-alt-esc: Remove some commented out, unused code
* gluon-alt-esc: swap order of Alt-ESC-provider and -client in menu
* gluon-alt-esc-client: add password option (WPA)

T-X:
Changelog:
* Rebased to master:
  * Switched to using ebtables --concurrent
* Squashed fixups to ease reviewing

Also made a separate PR for the gluon-client-bridge patch in #1094 (it is identical to the patch in here).

MPW1412:
This looks very interesting.

Questions:
 - package/gluon-alt-esc-provider/files/lib/gluon/alt-esc-provider/iptables.rules
I would suggest to filter all ip packages that might go into the private network of the node owner like 192.168.0.0/16 or 10.0.0.0/8. The NAT alone would not prevent, that you can access the private network if you guess the IP block correctly, as far as I know. Correct me if I am wrong here. Only IP packages to public IP addresses and only frames to the default route, e. g. the FritzBox like 192.168.178.1, should be allowed to go though. but not 192.168.178.99 for example.
 - As far as I understand it, this PR does only provide a repeater like distribution of the locally exited second network. So there would be no smart mesh routing like Batman does, right? Or did I miss that in the code. Maybe a second Batman interface could be created for this, and a real mesh network separated by VLAN from the normal one could be set up. That would save airtime and put less stress on the wifi driver, as ff-mesh, ff-client, local-exit-ff, local-exit-mesh, privat wifi would result in five wifi networks. That is barely manageable by one radio device.

In general, I'd prefer a non-separated approach. Batman should be declared a gateway locally and exit IPV4 packages to the local ISP. Roaming would still work as long as you stay in the local mesh cloud. IPV6 should be given as second router advertisement. 

Batman would have to be patched to not exit incomming connections over VPN.

T-X:
Hi @MPW1412,

Thanks for your comments!

Regarding your first point, you're absolutely right. I had briefly been wondering about that, too. But got stuck when wondering about what to do with IPv6 then. I somehow got to the conclusion then that we live an IPv6 world now where every device is accessible directly via IP(v6) anyways. And if you want to opt-out, creating a DMZ is the way to go. Does that make sense? Or do you have a threat scenario in mind where this assumption does not hold?

Concerning your second question: No, it's not just a "repeater-mode". gluon-alt-esc actually uses and works on top of the batman-adv mesh network. So batman-adv figures out the best route from the gluon-alt-esc-client node to the gluon-alt-esc-provider node (your alternative internet gateway). Also, the batman-adv gateway feature is not used for that, the alternative exit is selected via the statically configured MAC address (see screenshots above).

(The batman-adv gateway server list is only used to determine whether the to be forwarded frame is local or internet traffic, by comparing the frame's destination MAC. If it is in the batman-adv gateway list, then and only then the destination MAC address of the frame is replaced by the configured, alternative gateway MAC address).

rubo77:
Did someone create a standalone package for this for gluon 2018.1?

CodeFetch:
@T-X You need to filter 172.16.0.0/12, 10.0.0.0/8, 192.168.0.0/16 and maybe also compatibility/dummy addresses 192.0.0.8/32 and 192.0.0.0/29.

T-X:
@CodeFetch: See @MPW1412's question here:

https://github.com/freifunk-gluon/gluon/pull/1094#issuecomment-364127491

And my response here:

https://github.com/freifunk-gluon/gluon/pull/1094#issuecomment-364167682

Anything to add to that?

CodeFetch:
@T-X Yes, you need to filter 172.16.0.0/12, 10.0.0.0/8, 192.168.0.0/16 and maybe also compatibility/dummy addresses 192.0.0.8/32 and 192.0.0.0/29.

christf:
should we mark this as needs work as per discussin in #1472?

T-X:
Changelog v4:
* Rebase to master:
  * "local site = require 'gluon.site_config'" -> "local site = require 'gluon.site'"
      => not available anymore since f1f2bae94f684
         ("gluon-core: remove obsolete gluon.site_config library")
  * "... = site.site_code" -> "... = site.site_code()"
  * firewall: renamed zone "client" to "mesh"
      => renaming introduced by 1c1c9f8fc70f
         ("gluon-core: firewall rework, make base policy more restrictive")
* moved dependency for KERNEL_BRIDGE_EBT_DNAT and KERNEL_BRIDGE_EBT_REDIRECT from gluon-ebtables to gluon-alt-esc-client as requested in #1472

Note, that this pull-request should now only be merged after #1472 and #1568 were merged. This pull-request was retested together with #1472 and #1568 on top of current master.

rotanid:
thanks. i removed the "needs work" label and added the "blocked" label (as both of the other PRs have to be merged before this one) and the "needs testing" label

yanosz:
I like your idea <3 - unfortunatly, I won't be able to test it in the next weeks. 

At a quick glance, the UX appears a little bit complicated, but that won't be a problem for a first try.

mweinelt:
This package is intended for the community repository, which is soon to be created.

2tata:
@mweinelt sharing of the local internet connection could be a nice core feature! In combination with ddhcp or xlat. So people can be able to easy turn a gluon node into a supernode.

mweinelt:
It could be, what we're not quite there yet. And alt-escape being in community packages will not make that impossible either.

If this use case turns out to be practical there is not much that would prevent a package to migrate from community repos to the main repo.

A current issue we have with this package is that it makes use of L2 NAT via ebt_dnat, which we consider to be a hack.

rubo77:
Are all changes in Gluon accepted, so this can become a stand-alone package?

mweinelt:
It looks like they will be, we've just discussed what is necessary and what needs to be reviewed again.

Blockers are here:
- https://github.com/freifunk-gluon/gluon/pull/1568#discussion_r342269599
- https://github.com/freifunk-gluon/packages/pull/191#issuecomment-549545642

mweinelt:
Closing as we will not accept this into the core repository, as stated above.

rubo77:
That really is a pitty.

Is there another way, how we could re-implement a local exit then this into gluon?

smoe:
I followed @AlylonPrime's link in response to my feature request to this module and, well, what can I say, it reads fantastic. And I wish this thread had a Hollywood ending.

So, I can only second @rubo77 and ask for just anything, really :o/

Adorfer:
> Is there another way, how we could re-implement a local exit then this into gluon?

the blockers statet by mweinelt in https://github.com/freifunk-gluon/gluon/pull/1094#issuecomment-549579819 were quite clear. 
Adressing them might have helped. 
If there is an alternative route? Who knows? Probably there might be issues too, but perhaps less.

PR #1095, opened 2017-04-13
docs, site-example/site.mk: improve comment about variables

SimJoSt:
The comment had a typo, it was unclear to what it was referring and was positioned poorly.  
The new position is just slightly better, but everything else is fixed.

### DIFF
--- a/docs/site-example/site.mk
+++ b/docs/site-example/site.mk
@@ -36,7 +36,7 @@ GLUON_SITE_PACKAGES := \
 
 DEFAULT_GLUON_RELEASE := 0.6+exp$(shell date '+%Y%m%d')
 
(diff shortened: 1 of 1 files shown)


docs/site-example/site.mk
```diff
@@ -36,6 +36,7 @@ GLUON_SITE_PACKAGES := \
 
 DEFAULT_GLUON_RELEASE := 0.6+exp$(shell date '+%Y%m%d')
 
+# Variables set with =? can be overwritten from the command line
```
NeoRaider: Should be `?=` instead of `=?`.

SimJoSt: Oh... fixed.

PR #1097, opened 2017-04-14
CVE-2016-10229: udp: properly support MSG_PEEK with truncated buffers

jannic:
Add upstream patch to fix CVE-2016-10229

This just adds a commit from upstream linux v3.18.45, which should fix CVE-2016-10229 (https://nvd.nist.gov/vuln/detail/CVE-2016-10229).

Before merging, this patch should get some testing. By now, I only checked that ar71xx-generic still builds - not even tried to flash the resulting firmware.

This patch _may_ also be necessary for master (LEDE), if there are any relevant architectures still using 3.18. Architectures using 4.4 should be fine, as it contains the fix since v4.4.21.


NeoRaider:
Thanks, will merge after a quick build test.

Fortunately, usual Gluon setups aren't affected, as none of the default packages use MSG_PEEK (the dnsmasq code contains one call, but it is in the DHCP component which is not active during the normal operation of a node).

PR #1106, opened 2017-04-25
better upstream server description

Tarnatos:
Now its clear that upstreams severs normally means "community gateways".

### DIFF
--- a/docs/features/dns-cache.rst
+++ b/docs/features/dns-cache.rst
@@ -6,7 +6,8 @@ between node and gateway as small as possible. In order to do this, a
 dns-cache may be used on a node. The dnsmasq instance listening on port
 53 on the node will be reconfigured to answer requests, use a list of
 upstream servers and a specific cache size if the options listed below are
-added to site.conf. Upstream servers normally are the community's gateways.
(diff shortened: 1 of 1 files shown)


rubo77:
Some Call them"super nodes"

christf:
the description is now wrong. the node will not cache every request but it will cache a list with the size of of <cacheentries> results. 

Upstream servers may not be the communities gateways. The ffm dns is sitting on an entirely different infrastructure although it is possible to set the gateway. 

It is wrong that if no server is set that no setting will be changed. In any case site.nextnode and site.nextnode6 if they do exists are added to dnsmasq.

Tarnatos:
> the description is now wrong. the node will not cache every request but it will cache a list with the size of of cacheentries results.


Fixed.

> Upstream servers may not be the communities gateways.

Is more general now.
https://github.com/freifunk-gluon/gluon/pull/1106/files#diff-9755f535a05bebedf4fecc9ba425d995

> It is wrong that if no server is set that no setting will be changed. In any case site.nextnode and site.nextnode6 if they do exists are added to dnsmasq.

Was allready in the text an not written by me.

PR #1108, opened 2017-04-26
gluon-mesh-vpn-fastd: allow outbound connections on fastd-ports

christf:
This will allow a stricter fireall in general like one is used in the babel networks.

### DIFF
--- a/package/gluon-mesh-vpn-fastd/luasrc/lib/gluon/upgrade/400-mesh-vpn-fastd
+++ b/package/gluon-mesh-vpn-fastd/luasrc/lib/gluon/upgrade/400-mesh-vpn-fastd
@@ -55,6 +55,7 @@ uci:delete('fastd', 'mesh_vpn', 'user')
 
 
 local add_groups
+local ports = {}
 
 local function add_peer(group, name, config)
 	uci:section('fastd', 'peer', group .. '_peer_' .. name, {
@@ -64,6 +65,20 @@ local function add_peer(group, name, config)
 		key = config.key,
 		remote = config.remotes,
 	})
+
+	for k,r in ipairs(config.remotes)
+	do
+		for i in string.gmatch(r, "%S+")
+		do
+			if (tonumber(i))
+			then
+				if not isinset(tonumber(i),ports)
+				then
+					table.insert(ports, i)
+				end
+			end
+		end
+	end
 end
 
 local function add_group(name, config, parent)
@@ -98,7 +113,31 @@ function add_groups(prefix, groups, parent)
 	end
 end
 
+function isinset(needle, haystack)
+	for i,v in ipairs(haystack)
+	do
+		if (tonumber(v) == tonumber(needle))
+		then
+			return true
+		end
+	end
+	return false
+end
+
+
 add_groups('mesh_vpn', site.mesh_vpn.fastd.groups)
+
+for _,p in ipairs(ports)
+do
+	uci:section('firewall', 'rule', 'wan_fastd_' .. p, {
+		name = 'wan_fastd_' .. p,
+		dest = 'wan',
+		dest_port = p,
+		proto = 'udp',
+		target = 'ACCEPT',
+	})
+end
 
 
 uci:save('fastd')
+uci:save('firewall')


NeoRaider:
Isolated like this, this commit doesn't make much sense. Actually restricting the firewall on the WAN port should be done at the same time.

I don't want different WAN firewalls for batman-adv and Babel: either we restrict it for both, or for neither. I can see arguments for both settings, so at the moment I'm undecided which I like better.

christf:
This never really popped up but since in an l3 network each node is a router it feels better if there is a firewall restricting wan traffic. For Batman it likely would not hurt to have a firewall on the wan port.

If the firewall is used by both protocols then I feel that we should have a basic policy set somewhere in a central package which is the same and each package that needs to drill holes into the firewall should do this itself. ie fastd or prefixd once it is ready etc.

NeoRaider:
Exactly, the policy should be something basic.

In any case, I don't think there is a significant difference between L2 and L3 meshes in this regard, at least not for the OUTPUT policy.

christf:
so where do you think should the base policy go? gluon-core?
or rather gluon-firewall-base-policy

NeoRaider:
I think it should go into gluon-core, but I'm still not sure if we want a restrictive policy, and if we maybe want to make this configurable via site.conf.

Basically, I see the following arguments right now:

* \+ The autoupdater won't accidentially fall back to IPv4 via WAN, so a node with WAN connection will behave the same as a pure mesh node
* \- The autoupdater won't fall back to IPv4 via WAN, even if desired for some reason

No other software except the mesh VPN (fastd/tunneldigger, DNS) should try to open outgoing connections via IPv4 anyways, so the setting pretty much boils down to this one point. While I don't think there is a good reason to allow this fallback, I expect some backlash if we were to block it... Also, I'd like to keep the firewall as simple as possible; if the policy is just to allow outgoing connections, we don't need any exceptions for the mesh VPN.

christf:
in any case having a firewall in the grand scheme of things does not only concert outgoing connections. For babel we made a small table of what to support: http://l3-freifunk.readthedocs.io/de/latest/firmware.html#gluon-mesh-babel
obviously some programs are not available on batman-based nodes.
What do you think of the policies?

NeoRaider:
The policies make sense at first glance. I'd still argue that the outgoing policy for the WAN interface is independent of the other rules, so we can discuss it without looking at the whole picture.

I have some more points against a restrictive policy on the WAN port: it would also block connections explicitly opened by an admin working on the node via SSH. It might be desirable to be able to use SSH to jump to another system on the WAN, or some other protocol.

I'm currently leaning towards making the WAN policy restrictive by default, but finding more legitimate usecases for an open policy still adds to my undecidedness...

christf:
indeed, being able to open connections as admin is a strong argument for a less restrictive policy.
At the moment I feel a restrictive policy is in line with the dns-configuration that allows only fastd-users to use the wan-assigned dns. OTOH using separate dns-servers for fastd is not mainly a security measure but a necessity to make it work in the first place. 
Frankly I turn off the firewall to do  maintenance work that requires any outbound traffic because the dns settings gets in the way anyways. Yes, this is horrible but I feel when using an open outbound policy, this should be taken care of as well.

jplitza:
I don't see any reason for restricting outbound connections. It can hardly count as a security feature, as almost everything on the node is run as root and thus could easily change the firewall if it was pwned.

christf:
@NeoRaider when using ssh outbound connections on wan, how do you handle the dns-configuration?

NeoRaider:
I mostly expect SSH connections to other devices on the WAN network, for example over IPv6 link-local.

Resetting our rules from scratch on each upgrade should be okay, as long as additional custom rule sections are not touched.

christf:
not touching additional custom rules is a tough one as it means the firewall must not be reset. How can this be achieved?

Wouldn't it be an option to also allow outbound ssh connections and have a restrictive policy?

NeoRaider:
Our code never deletes UCI section it doesn't know, so additional rules are always retained; i'd like to keep it that way.

For sections that require the removal of many options, I usually prefer deleting the whole section and re-creating it, so we don't need to list options to delete. I don't think we have many sections of this kind in the firewall, so I guess it doesn't make much sense to reset the firewall configuration on upgrades after all...

christf:
this depends on #1124

christf:
this can be closed, since we are now using ACCEPT as outbound policy

PR #1109, opened 2017-04-27
Christf s/gluon core/docs

christf:
this is an attempt to fix the commit message.


NeoRaider:
Commit messages can't be changed, and reverting it doesn't make the history any more readable...

PR #1125, opened 2017-05-15
DOCS: site-example with cryptic SSID for mesh

rubo77:
The SSID for the mesh network should be cryptic, otherwise some users will try to connect to that network and get a bad user experience.

### DIFF
--- a/docs/site-example/site.conf
+++ b/docs/site-example/site.conf
@@ -53,7 +53,7 @@
 
     mesh = {
       -- Adjust these values!
-      id = 'ffxx-mesh',
+      id = 'ueH3uXjdp', -- usually you don't want users to connect to this mesh-SSID, so use a cryptic id that no one will accidentally mistake for the client WiFi
       mcast_rate = 12000,
       -- disabled = true, (optional)
     },
@@ -68,7 +68,8 @@
       ssid = 'alpha-centauri.freifunk.net',
     },
     mesh = {
-      id = 'ffxx-mesh',
+      -- Adjust these values!
+      id = 'ueH3uXjdp',
       mcast_rate = 12000,
     },
   },
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -111,9 +111,12 @@ wifi24 \: optional
     The example below disables 802.11b rates.
 
     ``ap`` requires a single parameter, a string, named ``ssid`` which sets the
-    interface's ESSID.
+    interface's ESSID. This is the WiFi the clients connect to.
 
-    ``mesh`` requires a single parameter, a string, named ``id`` which sets the mesh id.
+    ``mesh`` requires a single parameter, a string, named ``id`` which sets the
+    mesh id, also visible as an open WiFi in some network managers. Usually you 
+    don't want users to connect to this mesh-SSID, so use a cryptic id that no
+    one will accidentally mistake for the client WiFi.
 
     ``ibss`` requires two parametersr: ``ssid`` (a string) and ``bssid`` (a MAC).
     An optional parameter ``vlan`` (integer) is supported.
@@ -131,7 +134,7 @@ wifi24 \: optional
            ssid = 'alpha-centauri.freifunk.net',
          },
          mesh = {
-           id = 'alpha-centauri-mesh',
+           id = 'ueH3uXjdp',
            mcast_rate = 12000,
          },
          ibss = {


rubo77:
It is not that bad, like it was with the SSID for adhoc, because at least Android and Windows donÄt show the mesh-SSID in the networkmanager, but Linux does and I guess it will be more os's in the future.

Tarnatos:
At least Android 7 shows the mesh ssid on 11s, too.

NeoRaider:
Using something like a MAC address seems weird, too. In Lübeck, we just use a random string, created using `pwgen` (or plan to use it when we switch to 11s).

Also, we could add this as a comment to the file instead of changing the example itself. From the Freifunk perspective, using a cryptic mesh ID makes sense I guess; for other uses of Gluon, I'd rather choose a more descriptive value.

kpanic23:
Actually I think most users would automagically use the least cryptic SSID to connect to. So I would rather keep the normal client SSID as simple as possible. 
We for example are using "Freifunk" for all client networks and, depending on the sub-community, for example "ff3l-hoho-mesh", "ff3l-wtk-mesh", "ff3l-wald-mesh", ... for the respective mesh net.

SimJoSt:
In Bremen we use `mesh.ffhb.de` which redirects to our website if you type it in.

Adorfer:
@kpanic23  
> So I would rather keep the normal client SSID as simple as possible.

People seem to connect to any uncrypted network systematically. Perhaps it's some devices which do that automatically, but they just try. And at least i would suggest not to create any obvious link between that network and the brand "freifunk", otherwise it creates just the notion of "freifunk" being unreliable and disfunctional al the time. (and no, the link been an SSID and fqdn hostname to connecti via port 80/443 is none which a normal user makes, especially if it does not start by 'www'.)

rotanid:
@rubo77 it was suggested by neoraider to use a comment, please read his comment containing the suggestion and answer and/or update your PR accordingly.

rubo77:
I changed it to a random string and added a hint, that you useually don't want clients to connect to the mesh wifi

Adorfer:
this would have saved me the discussions about the 
_"non working 'freifunk batbone' which has a spelling mistake in it too!"_
(and i can not imagine how many people did not dare to ask and just considered Freifunk as at least partly defective.)

rubo77:
This can be merged. It is only docs

docs/site-example/site.conf
```diff
@@ -53,7 +53,7 @@
 
     mesh = {
       -- Adjust these values!
-      id = 'ffxx-mesh',
+      id = 'ueH3uXjdp', -- usually you don't want users to connect to this mesh-SSID, so use something cryptic, that no one accidentally mistakes this for the client WiFi
```
rotanid: "cryptic, that no one" isn't correct english

docs/user/site.rst
```diff
@@ -111,9 +111,12 @@ wifi24 \: optional
...
 
     ``ap`` requires a single parameter, a string, named ``ssid`` which sets the
-    interface's ESSID.
+    interface's ESSID. this is the WiFi the clients connect to.
 
-    ``mesh`` requires a single parameter, a string, named ``id`` which sets the mesh id.
+    ``mesh`` requires a single parameter, a string, named ``id`` which sets the
+    mesh id, also visible as open WiFi in some network managers. Usually you 
+    don't want users to connect to this mesh-SSID, so use something cryptic,
+    that no one accidentally mistakes this for the client WiFi.
```
rotanid: same as above, "cryptic, that no one"

docs/user/site.rst
```diff
@@ -111,9 +111,12 @@ wifi24 \: optional
     The example below disables 802.11b rates.
 
     ``ap`` requires a single parameter, a string, named ``ssid`` which sets the
-    interface's ESSID.
+    interface's ESSID. this is the WiFi the clients connect to.
 
-    ``mesh`` requires a single parameter, a string, named ``id`` which sets the mesh id.
+    ``mesh`` requires a single parameter, a string, named ``id`` which sets the
+    mesh id, also visible as open WiFi in some network managers. Usually you
```
rotanid: insert "an" after "visible as"

rubo77:
done:

>Usually you don't want users to connect to this mesh-SSID, so use a cryptic id that no
    one will accidentally mistake for the client WiFi

rotanid:
very small issue: please don't use capital "d" in the commit message ("Docs")

PR #1138, opened 2017-06-12
add gluonShellDiet to remove all comment lines in shell files

rubo77:
We can save some space on the nodes deleting all comment lines there, for example:

    $ cat /etc/init.d/*|grep '#'|grep -v '#!'|wc
       90       617      3850

### DIFF
--- a/package/gluon-ebtables/Makefile
+++ b/package/gluon-ebtables/Makefile
@@ -32,6 +32,7 @@ endef
 
 define Package/gluon-ebtables/install
 	$(CP) ./files/* $(1)/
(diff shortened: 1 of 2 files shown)


rubo77:
Maybe we can better put this in the main makefile at the end? Where would that be?

NeoRaider:
1. Most shell scripts are upstream OpenWrt/LEDE stuff. We don't want to touch upstream packages in Gluon.
2. Upstream OpenWrt/LEDE doesn't want script minifiers.

Adorfer:
what would have been the overall gain? (in the compressed image size, not in the uncompressed files.)

PR #1144, opened 2017-06-17
gluon-client-bridge: change local_node interface to be part of client…

christf:
… zone

otherwise traffoc to local_node will not be considered coming from a client. this would break l3-networks


rotanid:
you added firewall changes without updating the subject or text of your PR, please clarify :)

NeoRaider:
br-client and local_node must be handled separately. For L3 routing protocols, br-client should be kept on DROP (the bridge is only used to tie together interfaces); all traffic must go though local_node.

batman-adv already overrides the policy for br-client, but this setting only makes sense for a L2 routing protocol.

PR #1149, opened 2017-06-22
gluon-web-admin: add option to show/hide password-login and add minimum password length

rubo77:
Or maybe better min 12 characters?

It would be nice, if you get a verbose error message, if the pass is too short, but that needs some more work on gluon-web, i guess.

pw

### DIFF
--- a/docs/package/gluon-web-admin.rst
+++ b/docs/package/gluon-web-admin.rst
@@ -1,0 +1,27 @@
+gluon-web-admin
+===============
+
+This package allows the user to set options like the password for ssh access
+within config mode. You can define in your ``site.conf`` whether it should be
+possible to access the nodes via ssh with a password or not and what the mimimum
+password length must be.
+
+site.conf
+---------
+
+config_mode.remote_login.show_password_form \: optional (defaults to ``false``)
+  If ``show_password_form`` is set to ``true``, the password section in
+  config mode is shown.
+  
+config_mode.remote_login.min_password_length \: optional (defaults to ``12``)
+  This sets the minimum allowed password length. Set this to ``1`` to
+  disable the length check.
+
+If you want to enable the password login you can use this example::
+
+  config_mode = {
+    remote_login = {
+      show_password_form = true, -- default false
+      min_password_length = 12
+    }
+  }
--- a/package/gluon-web-admin/i18n/de.po
+++ b/package/gluon-web-admin/i18n/de.po
@@ -32,6 +32,9 @@ msgstr "Abbrechen"
 msgid "Confirmation"
 msgstr "Bestätigung"
 
+msgid "%u characters min."
+msgstr "Mindestens %u Zeichen"
+
 msgid "Continue"
 msgstr "Fortfahren"
 
--- a/package/gluon-web-admin/luasrc/lib/gluon/web/model/admin/remote.lua
+++ b/package/gluon-web-admin/luasrc/lib/gluon/web/model/admin/remote.lua
@@ -13,12 +13,13 @@ You may obtain a copy of the License at
 local nixio = require "nixio"
 local fs = require "nixio.fs"
 local util = require "gluon.util"
+local site = require 'gluon.site_config'
 
 local f_keys = Form(translate("SSH keys"), translate("You can provide your SSH keys here (one per line):"), 'keys')
 local s = f_keys:section(Section)
 local keys = s:option(TextValue, "keys")
-keys.wrap    = "off"
-keys.rows    = 5
+keys.wrap = "off"
+keys.rows = 5
 keys.default = fs.readfile("/etc/dropbear/authorized_keys") or ""
 
 function keys:write(value)
@@ -30,11 +31,19 @@ function keys:write(value)
 	end
 end
 
+local config=((site.config_mode or {}).remote_login or {})
+local min_password_length = config.min_password_length or 12
+local mintype = 'minlength(' .. min_password_length .. ')'
 
-local f_password = Form(translate("Password"),
-	translate(
-                "Alternatively, you can set a password to access your node. Please choose a secure password you don't use anywhere else.<br /><br />"
-                .. "If you set an empty password, login via password will be disabled. This is the default."
+if not config.show_password_form then
+	-- password login is disabled in site.conf
+	return f_keys	
+end
+
+local f_password = Form(translate("Password"), translate(
+	"Alternatively, you can set a password to access your node. Please choose a "
+	.. "secure password you don't use anywhere else.<br /><br />If you set an empty "
+	.. "password, login via password will be disabled. This is the default."
 	), 'password'
 )
 f_password.reset = false
@@ -43,12 +52,18 @@ local s = f_password:section(Section)
 
 local pw1 = s:option(Value, "pw1", translate("Password"))
 pw1.password = true
+pw1.optional = true
+pw1.datatype = mintype
 function pw1.cfgvalue()
 	return ''
 end
 
-local pw2 = s:option(Value, "pw2", translate("Confirmation"))
+local pw2 = s:option(Value, "pw2", translate("Confirmation"), 
+	min_password_length <= 1 and nil or translatef("%u characters min.", min_password_length)
+)
 pw2.password = true
+pw2.optional = true
+pw2.datatype = mintype
 function pw2.cfgvalue()
 	return ''
 end
@@ -93,7 +108,7 @@ function f_password:write()
 
 	local pw = pw1.data
 
-	if #pw > 0 then
+	if pw ~= nil and #pw > 0 then
 		if set_password(pw) then
 			f_password.message = translate("Password changed.")
 		else
--- a/package/gluon-web-admin/Makefile
+++ b/package/gluon-web-admin/Makefile
@@ -39,4 +39,9 @@ define Package/gluon-web-admin/install
 	$(call GluonInstallI18N,gluon-web-admin,$(1))
 endef
 
+define Package/gluon-web-admin/postinst
+#!/bin/sh
+$(call GluonCheckSite,check_site.lua)
+endef
+
 $(eval $(call BuildPackage,gluon-web-admin))
--- a/package/gluon-web-admin/check_site.lua
+++ b/package/gluon-web-admin/check_site.lua
@@ -1,0 +1,4 @@
+if need_table('config_mode', nil, false) and need_table('config_mode.remote_login', nil, false) then
+  need_boolean('config_mode.remote_login.show_password_form', false)
+  need_number('config_mode.remote_login.min_password_length', false)
+end
--- a/package/gluon-web-admin/i18n/fr.po
+++ b/package/gluon-web-admin/i18n/fr.po
@@ -33,6 +33,9 @@ msgstr "Annuler"
 msgid "Confirmation"
 msgstr "Confirmation"
 
+msgid "%u characters min."
+msgstr "Au moins %u caractères"
+
 msgid "Continue"
 msgstr "Continuer"
 
--- a/package/gluon-web-admin/i18n/gluon-web-admin.pot
+++ b/package/gluon-web-admin/i18n/gluon-web-admin.pot
@@ -19,6 +19,9 @@ msgstr ""
 msgid "Confirmation"
 msgstr ""
 
+msgid "%u characters min."
+msgstr ""
+
 msgid "Continue"
 msgstr ""
 


NeoRaider:
How about adding a hint about the minimum length to one the texts, for example the label of the text field?

Sunz3r:
pls let the user decide how strong the password should be

rubo77:
If you want a weak password, you can set that ok the console.

If we allow weak passwords​ by default, we'll get complaints, that some users home net was hacked by Freifunk.

We could add a site. Conf option"allow weak passwords"

rubo77:
We could add a site. Conf option"allow weak passwords"

rubo77:
We could add a check for the common mistakes like repeating string, no capital chat, no number,....

Sunz3r:
you are right, hacked by freifunk are very bad.
how about limiting brute force attacks on ssh with iptables?

rubo77:
I added to the text " (min. 8 characters)"

rubo77:
tested and working. I added a screenshot to the first post.

the empty password field is now always red on load, which looks a bit strange.

NeoRaider:
I don't see a reason to allow passwords < 8 characters, that's a very weak limilation.

If you really want a weaker password, you can still set one via SSH.

rotanid:
good idea @rubo77 .
we removed the password option completely in our firmware - is just don't have time to upstream this using a site.conf-option to enable/disable this...

package/gluon-web-admin/luasrc/lib/gluon/web/model/admin/remote.lua
```diff
@@ -33,8 +33,9 @@ end
 
 local f_password = Form(translate("Password"),
 	translate(
-                "Alternatively, you can set a password to access you node. Please choose a secure password you don't use anywhere else.<br /><br />"
-                .. "If you set an empty password, login via password will be disabled. This is the default."
+		"Alternatively, you can set a password to access you node. Please choose a "
```
rotanid: although this typo was already there, please fix "you node" -> "your node"

rubo77:
What would you want to control via site conf?

Maybe I find the time

rotanid:
@rubo77 control if the password-ssh-option is in the config mode or not.
we removed it in our gluon fork, but did it by removing the code, as that was easier than making it optional via site.conf
ask me on IRC if something isn't clear

rubo77:
OK, i will add two options to site.conf:

remote = {
  allow_password_login = true, -- optional; default: true
  min_passwordlength = '8'     -- optional; default: 0; set this to 0 to disable a length check
},

package/gluon-web-admin/luasrc/lib/gluon/web/model/admin/remote.lua
```diff
@@ -30,80 +31,88 @@ function keys:write(value)
 	end
 end
 
+if ((site.config_mode or {}).remote_login or {}).allow_password_login then
```
rubo77: not sure, how to formulate this, so it defaults to true

rubo77:
it is configurable now in site.conf.

rubo77:
What do you thhink of an option to check the password-strength also?

I could add a codeblock that checks for existance of capital letters, numbers and special chars

docs/package/gluon-web-admin.rst
```diff
@@ -0,0 +1,27 @@
...
+
+site.conf
+---------
+
+config_mode.remote_login.allow_password_login \: optional (defaults to true)
+  If ``allow_password_login`` is set to ``false``, the password section in
+  config mode is hidden
+  
+config_mode.remote_login.min_password_length \: optional (defaults to '8')
+  This stes a minimum limit to the allowed password length. Set this to 0 to
```
rotanid: "stes" is wrong. also, the sentence is weird... maybe "This sets the minimum allowed password length."

rotanid:
checking the password strength should be optional though, if you add it.
suggestion would be to add some levels like "0 = no check", "1 = at least two of four types of chars" , "2 = at least three of four types" etc. - where types are lowercase letters, uppercase letters, digits, special chars

please also update the pull request title and text and the commit message according to your additional changes

NeoRaider:
Having an option to disable the password form completely make sense in my opinion; the other options just add complexity. We should define simple, sane defaults and keep it at that.

rubo77:
But I want to set the minimum length to 12 characters in Kiel

NeoRaider:
Is this an actual issue? Passwords are only for people who need SSH access. If you know how to use the command line, you should be able to choose a good password...

If you're concerned about stupid users, you really shoud just hide the password form. It will still be possible to set passwords over SSH in config mode (and users who're annoyed by arbitrary password restrictions in the web interface will simply go that route...)

rubo77:
OK, If I remove the length-definition, I will set the minimum Length to 12 characters

rubo77:
OK, I removed the length-definition and set the minimum Length to 12 characters

package/gluon-web-admin/luasrc/lib/gluon/web/model/admin/remote.lua
```diff
@@ -30,80 +31,87 @@ function keys:write(value)
 	end
 end
 
+if ((site.config_mode or {}).remote_login or {}).allow_password_login ~= false then
```
kb-light: maybe:
```
if ((site.config_mode or {}).remote_login or {}).allow_password_login == false then
 	-- password login is disabled in site.conf
 	return f_keys
end
[...]
return f_keys, f_password
```
to not indent the whole script

rubo77: I thought about this earlyer, but then I thought it wouldn't be good readeable to exit the script so early with return. But fine ;=

done

rubo77:
I built it again, and it looks fine.

package/gluon-web-admin/luasrc/lib/gluon/web/model/admin/remote.lua
```diff
@@ -43,12 +48,15 @@ local s = f_password:section(Section)
...
 local pw1 = s:option(Value, "pw1", translate("Password"))
 pw1.password = true
+pw1.datatype = 'minlength(12)'
 function pw1.cfgvalue()
 	return ''
 end
 
-local pw2 = s:option(Value, "pw2", translate("Confirmation"))
+local pw2 = s:option(Value, "pw2", translate("Confirmation"),
+	translate("Minimum") .. " " .. 12 .. " " .. translate("characters"))
```
NeoRaider: Always use `translatef` with format strings instead of splitting up sentences into multiple translation strings (sentences aren't constructed the same way in all languages). The English string should be "%u characters min.".

docs/package/gluon-web-admin.rst
```diff
@@ -0,0 +1,21 @@
...
+
+config_mode.remote_login.allow_password_login \: optional (defaults to true)
+  If ``allow_password_login`` is set to ``false``, the password section in
+  config mode is hidden
+
+Example::
+
+  config_mode = {
+    remote_login = {
+      allow_password_login = true
```
NeoRaider: Maybe "show_password_form"? Password login is still allowed when the form is disabled.

package/gluon-web-admin/luasrc/lib/gluon/web/model/admin/remote.lua
```diff
@@ -43,12 +48,15 @@ local s = f_password:section(Section)
 
 local pw1 = s:option(Value, "pw1", translate("Password"))
 pw1.password = true
+pw1.datatype = 'minlength(12)'
```
NeoRaider: After thinking about this a bit more, I'm okay with making the limit configurable in site.conf after all.

package/gluon-web-admin/luasrc/lib/gluon/web/model/admin/remote.lua
```diff
@@ -30,11 +31,15 @@ function keys:write(value)
 	end
 end
 
+if ((site.config_mode or {}).remote_login or {}).allow_password_login == false then
```
NeoRaider: site.conf check for the new setting is missing

package/gluon-web-admin/luasrc/lib/gluon/web/model/admin/remote.lua
NeoRaider: You'll also need to set *optional* on these fields to allow unsetting the password again. The *write* method must then be adjusted to check for *nil* values.

package/gluon-web-admin/luasrc/lib/gluon/web/model/admin/remote.lua
rubo77: I'll check that if I built again. what exactly do you have in mind here?

rubo77:
I added everything and tested it again. build complete and it works fine:

auswahl_100

package/gluon-web-admin/luasrc/lib/gluon/web/model/admin/remote.lua
```diff
@@ -86,6 +94,11 @@ local function set_password(password)
 end
 
 function f_password:write()
+	if pw1.data:len() < min_password_length then
```
NeoRaider: Using the datatype is much better than checking manually, as the fields will give direct feedback (background color) then. Nicer textual error messages are a TODO for future improvements in gluon-web.

We're only using the ugly errmessage because we can't express the "same value" constraint using datatypes, ideally, we'd change that to use a similar validation in the future, too.

package/gluon-web-admin/luasrc/lib/gluon/web/model/admin/remote.lua
```diff
@@ -30,11 +31,17 @@ function keys:write(value)
 	end
 end
 
+local min_password_length = ((site.config_mode or {}).remote_login or {}).min_password_length or 12
```
NeoRaider: I'd like to keep the default unchanged from the current situation, so make the default 1.

package/gluon-web-admin/i18n/de.po
```diff
@@ -17,7 +17,7 @@ msgid "Advanced settings"
 msgstr "Erweiterte Einstellungen"
 
 msgid ""
-"Alternatively, you can set a password to access you node. Please choose a "
+"Alternatively, you can set a password to access your node. Please choose a "
```
NeoRaider: This is an unrelated change that affects multiple files. Please split it out into a second commit.

package/gluon-web-admin/luasrc/lib/gluon/web/model/admin/remote.lua
```diff
@@ -47,7 +54,8 @@ function pw1.cfgvalue()
 	return ''
 end
 
-local pw2 = s:option(Value, "pw2", translate("Confirmation"))
+local pw2 = s:option(Value, "pw2", translate("Confirmation"),
+	translatef("%u characters min.", min_password_length))
```
NeoRaider: The length hint should only be shown when min_password_length > 1.

package/gluon-web-admin/luasrc/lib/gluon/web/model/admin/remote.lua
rubo77: But we try to fix an existing serious deficit here, so adding a minimum length is cruicial to security in Freifunk.

And there is no drawback for users, that used a short password before, because there cannot be no check of en existing password that is too short.

package/gluon-web-admin/luasrc/lib/gluon/web/model/admin/remote.lua
rubo77: I had this before, but Then the password field was always red, until you typed something in.

The solution now is much more intuitive and uses the correct warning colors.

package/gluon-web-admin/i18n/de.po
rubo77: Sure, this is unrelated, but it only affects files, that we change in this PR anyway. Should I still split it off?

package/gluon-web-admin/luasrc/lib/gluon/web/model/admin/remote.lua
NeoRaider: It was red because you didn't define the fields as optional.

package/gluon-web-admin/i18n/de.po
NeoRaider: Yes.

package/gluon-web-admin/luasrc/lib/gluon/web/model/admin/remote.lua
rubo77: Usually I would agree, but here we try to fix an existing serious security whole. So adding a minimum length is cruicial for the security of the nodes. I cannot think of a case, where a short password is needed.

Also here there is no drawback for users, that used a short password before, because there cannot be no check of en existing password that is too short.

package/gluon-web-admin/luasrc/lib/gluon/web/model/admin/remote.lua
rubo77: done

package/gluon-web-admin/i18n/de.po
rubo77: done

package/gluon-web-admin/luasrc/lib/gluon/web/model/admin/remote.lua
rubo77: done

PowerPan:
I think every user should have the chance to select his own password.

Maybe it could be better to make a informative message, that the password with the length < $minlength is a weak password but if someone want to use it he should do this. 

It is not our responsibility to enforce security, but we should point it out

rubo77:
As said before: If you want a weak password, you can set it on the console via SSH. And if you want to allow it in your Community in config mode, you can set the min-length to 1 in your `site.conf`.

We are discussing at the moment the default settings if nothing is set in `site.conf`. Those should be set on maximum security, which would be one of the following:

rubo77:
We could set as default **password-login disabled**

rubo77:
We could set a long password as default (10 or 12 chaaracters)

rotanid:
@rubo77 the last time we talked about this in IRC, as far as i remember we said "default disabled, default 12 chars", because we said existing setups aren't touched and everyone can change it in the site.conf if they really want

rubo77:
Ok, I changed the default for `show_password_form` to false

rotanid:
maybe the example in the documentation should reflect the default values, not something random.
also, please update the commit message to reflect the password show/hide

rubo77:
I changed the example, so it shows how to enable it with 12 chars. and changed the text afccordingly

commit message is now "gluon-web-admin: add option to show/hide password-login and add minimum password length"

package/gluon-web-admin/luasrc/lib/gluon/web/model/admin/remote.lua
```diff
@@ -30,11 +31,18 @@ function keys:write(value)
 	end
 end
 
+local min_password_length = ((site.config_mode or {}).remote_login or {}).min_password_length or 12
```
NeoRaider: You could reduce the code duplication a bit by using a variable for the remote login config:
```lua
local config = (site.config_mode or {}).remote_login or {}
```

package/gluon-web-admin/luasrc/lib/gluon/web/model/admin/remote.lua
```diff
@@ -43,12 +51,20 @@ local s = f_password:section(Section)
...
 local pw1 = s:option(Value, "pw1", translate("Password"))
 pw1.password = true
+pw1.optional = true
+pw1.datatype = mintype
 function pw1.cfgvalue()
 	return ''
 end
 
-local pw2 = s:option(Value, "pw2", translate("Confirmation"))
+local hint = nil;
```
NeoRaider: Unneeded semicolon and initialization (a variable's value will default to nil when not specified)

NeoRaider:
I'm not sure what happened to the *hint* variable, and the coding style of the latest changes is bad, but I'll just fix it up myself...

rubo77:
the variable hint was just used once, so I thought it would be better to directly add it in place. But ok, I'll keep in mind, that the readeablility is more important than saving a few characters in the code.

NeoRaider:
Merged with minor adjustments.

rotanid:
commit that was merged, for reference:
84d117ff73ac3a2099f851d0c4748cbfd1f41626

PR #1153, opened 2017-06-24
add new package gluon-web-logging

kb-light:


NeoRaider:
Nice! Maybe simply call the package "gluon-web-logging" and also change the page name to "Logging"? Other options like the logread buffer size could be added later.

rotanid:
the idea is nice, although i have a problem with logs being transported without encryption, as there may be information inside that you dont want to make "public".
at least, there should be a very clear warning about this...

docs/package/gluon-web-logging.rst
```diff
@@ -0,0 +1,5 @@
+gluon-web-logging
+=================
+
+The *gluon-web-logging* package add a new section to advanced settings,
+to allow GUI based configuration of an remote syslog server.
```
rubo77: "The *gluon-web-logging* package adds a new section to the advanced settings to allow GUI based configuration of a remote syslog server"

kb-light:
@rotanid is the warning clear enough?

package/gluon-web-logging/luasrc/lib/gluon/web/model/admin/logging.lua
```diff
@@ -0,0 +1,49 @@
+local uci = require('simple-uci').cursor()
+local system = uci:get_first('system', 'system')
+
+local f = Form(translate('Logging'), translate(
+	'If you want to use a remote syslog server, you can set it up here. '
+	.. 'Please keep in mind, that the data is transmitted unencrypted.'
```
NeoRaider: There shouldn't be a comma before "that" here.

package/gluon-web-logging/luasrc/lib/gluon/web/model/admin/logging.lua
```diff
@@ -0,0 +1,49 @@
...
+))
+local s = f:section(Section)
+
+local enable = s:option(Flag, 'log_remote', translate('Enable'))
+enable.default = uci:get_bool('system', system, 'log_remote')
+enable.optional = true
```
NeoRaider: Setting *optional* on a *Flag* doesn't make sense, as it is always true or false, but never undefined.

package/gluon-web-logging/luasrc/lib/gluon/web/model/admin/logging.lua
```diff
@@ -0,0 +1,49 @@
...
+end
+
+local ip = s:option(Value, 'log_ip', translate('IP'))
+ip.default = uci:get('system', system, 'log_ip')
+ip:depends(enable, true)
+ip.optional = false
+ip.placeholder = '0.0.0.0'
+ip.datatype = 'ipaddr'
+function ip:write(data)
+	if data ~= nil then
```
NeoRaider: You set *optional* to false, so *data* can never be nil. (And if it could be nil, `uci:set('system', system, 'log_ip', data)` would delete it with simple-uci, so the if/else would not be necessary anyways.)

package/gluon-web-logging/luasrc/lib/gluon/web/model/admin/logging.lua
```diff
@@ -0,0 +1,49 @@
...
+		uci:delete('system', system, 'log_ip')
+	end
+end
+
+local port = s:option(Value, 'log_port', translate('Port'))
+port.default = uci:get('system', system, 'log_port')
+port:depends(enable, true)
+port.optional = true
+port.placeholder = 514
+port.datatype = 'uinteger'
```
NeoRaider: `irange(1, 65535)` would be a better check.

package/gluon-web-logging/luasrc/lib/gluon/web/model/admin/logging.lua
```diff
@@ -0,0 +1,49 @@
...
+function port:write(data)
+	if data ~= nil then
```
NeoRaider: Again, the if/else is not necessary with simple-uci.

rotanid:
first i thought "yes", but maybe we should add why this might be a problem?
not everyone knows or thinks of when reading/enabling this, what might be inside the logs.
in case of smartphones, you could say the mac address is "personenbezogen" and therefore say "... dass diese Daten, die teils auch personenbezogen sind, unverschlüsselt ..."

docs/package/gluon-web-logging.rst
```diff
@@ -0,0 +1,5 @@
+gluon-web-logging
+=================
+
+The *gluon-web-logging* package add a new section to advanced settings,
```
NeoRaider: Typo: adds

I'd also remove the comma at the end of the line.

docs/package/gluon-web-logging.rst
```diff
@@ -0,0 +1,5 @@
...
+to allow GUI based configuration of a remote syslog server.
```
NeoRaider: I'd write "GUI-based" with a dash. Or maybe drop it completely, and just write "allow configuration"?

package/gluon-web-logging/luasrc/lib/gluon/web/model/admin/logging.lua
kb-light: without the if/else here I got some trouble while testing, because the whole section was renamed to log_port if data was nil.

before:
```
system.@system[0]=system
system.@system[0].timezone='CET-1CEST,M3.5.0,M10.5.0/3'
system.@system[0].hostname='xxx'
system.@system[0].log_ip='192.168.1.8'
system.@system[0].log_remote='0'
```

afterwards:
```
system.@log_port[0]=log_port
system.@log_port[0].timezone='CET-1CEST,M3.5.0,M10.5.0/3'
system.@log_port[0].hostname='xxx'
system.@log_port[0].log_ip='192.168.1.8'
system.@log_port[0].log_remote='1'
```

and that seems expactably to me after reviewing: https://github.com/freifunk-gluon/packages/blob/master/libs/lua-simple-uci/src/simple-uci.lua#L34

NeoRaider: Ah, you're right, the set() interface is still confusing in simple-uci... I'll need to fix that in simple-uci, but we can keep the if/else here for now.

package/gluon-web-logging/i18n/de.po
```diff
@@ -0,0 +1,15 @@
...
+msgstr "Content-Type: text/plain; charset=UTF-8"
+
+msgid ""
+"If you want to use a remote syslog server, you can set it up here. "
+"Please keep in mind that the data is not encrypted, which may cause "
+"individual-related data to be transmitted unencrypted over the internet."
+msgstr ""
+"Wenn du einen Remote-Syslog-Server nutzen möchtest, dann kannst du ihn hier "
+"eintragen. Bitte beachte, dass diese Daten nicht verschlüsselt werden. Dies "
+"kann dazu führen, dass personenbezogene daten unverschlüsselt über das Internet "
```
Brother-Lal: "daten" -> "Daten"

lrnzo:
what would i have to write into my site.conf in order to replace the `ip.placeholder` with e.g. `1.2.3.4`? i tried this:
```
       config_mode = {
                gluon_web_logging = {
                        ip = {
                                placeholder = '1.2.3.4',
                        },
                },
        },

```
but it does not work. there is still `0.0.0.0` in the config mode.

NeoRaider:
@lrnzo 
- The placeholder is hardcoded in Lua code, it can't be configured at the moment (in fact, the gluon-web-logging package doesn't use any site.conf settings)
- A placeholder is just that: A placeholder that is shown to the user when the field is empty to describe how the expected input might look. It isn't a default value and doesn't have any other effect.

What you're looking for would be a default value, which seems like a useful addition to me. Especially in private (non-Freifunk) meshes, an admin might even want to have all nodes send logs to a central server by default. A closed ticket is not the right place to ask questions about configuration, please open a new ticket with the feature request.

lrnzo:
@NeoRaider thanks! i will follow your advice

PR #1159, opened 2017-06-26
build: add gluon-release make target

kb-light:
`make gluon-release` can be used to discover the current release in scripts.


NeoRaider:
How about calling it `show-release`?

kb-light:
show-release it is

PR #1161, opened 2017-06-26
gcc: add support for gcc 7.1+

RubenKelevra:


NeoRaider:
Why would we want this...?

* There are serious issues in core components when compiled with GCC7 (dropbear crypto breaks, not allowing SSH logins)
* I'd like to avoid adding patches that don't fix bugs or are directly relevant for features we want in Gluon

RubenKelevra:
We use an ArchLinux build-server, so this might be mandatory for some communities as well for building gluon on their build-servers as well - for Fedora for example.

NeoRaider:
All the patches you add are for switching the *target* GCC to 7.1, this has nothing to do with host compiler support. Gluon is already building fine using GCC 7.1 as host compiler.

Issue #1123, opened 2017-05-11
support for TP-Link Archer C25


rotanid:
the new Archer C25 from TP-Link has dual-radio, 775MHz Qualcomm Atheros CPU, 64MB RAM, 8MB Flash, 5* 10/100 Mbps Ethernet.

@Brother-Lal did already work on LEDE support:
https://github.com/lede-project/source/pull/1069

jannic:
I made a build of gluon based on the LEDE patches:
https://github.com/jannic/gluon/tree/c25
This gives a bootable firmware, but is not ready for merging because of a few issues:

- There is some major memory leak, probably related to ath10k. The router reboots after ~1 minute with an oom error. If I disable the 5GHz wifi, the leak seems to go away.
- LAN/WAN ports are swapped
- LED support seems to be broken (all LEDs are always on)
- not much testing done, yet - just 'it boots and gets mesh connectivity'

This is a purely mechanical backport of the LEDE C25 patches  to LEDE 17.1, adding a bare minimum of gluon integration. Take it as a proof of concept and a starting point for further development, it's not yet finished.

NeoRaider:
The swapped LAN/WAN ports are weird, does this work on plain LEDE?

jannic:
Didn't try plain LEDE, so I don't know. The C25 support of LEDE is still in development, so it's quite possible that there is still a bug.
On the other hand, the LEDE patch lists eth0 as the WAN port, which seems to be correct (I'm currently connected to the WAN port and eth0 is UP.)
But gluon assigns eth0 to the br-client bridge.

Brother-Lal:
Please try my patches with a bare lede and see if you can reproduce the issues!
I had no issues in that regard, leds, wan/lan wifi should all be working, so it would be nice to have feedback if it's broken for you.

jannic:
@Brother-Lal sure, I'll do if I find the time.
There are actually two versions of LEDE to test: Your tree, based on LEDE master, and the 'backport' (just fixing merge conflicts, nothing fancy) to 17.01. Perhaps I missed some patch in LEDE master which should be backported, as well.

jannic:
Ok, LEDE (both 17.1 and master) properly assigns eth1 to br-lan, so it looks like it's a gluon-only issue.
The LEDs don't work with 17.1, but they do with master. Perhaps I broke something while backporting, or they just need some feature not yet available in 17.1.
[...]

PR #1162, opened 2017-06-26
ar71xx: Archer c25/c59/c60 support

RubenKelevra:


rubo77:
How much of https://github.com/freifunk-gluon/gluon/issues/1123 is taken from https://github.com/freifunk-gluon/gluon/issues/1123 ?

NeoRaider:
There's quite some chaos here:

* Many patches don't do what their commit messages claim
* The patches have not been created using `make update-patches`
* The patches are out of order
* Some patches are entirely unrelated to the backport
* Hardware backports should rather happen in the LEDE 17.01 upstream than in Gluon, at least when they're possible without too many changes affecting other devices (which should be possible if all the unrelated patches are removed)

RubenKelevra:
Hey NeoRaider,

I'm sorry if this looks like chaos, to you. I'm not familiar with the exact procedure you would recommend, do you have any link to a guideline?

* Many patches don't do what their commit messages claim

The patches are not from me, they are from the lede-master branch, so if you dislike the summaries there, I would recommend contacting the authors.

* The patches have not been created using make update-patches

Yes, they have been created with git format-patch.

* The patches are out of order.

Yes, since some interim changes have been committed to lede 17.01 too and some not I had serious trouble to identify anything needed. And the commit has been modified to apply in the order I provided the idea of minimal changes to them as possible.

* Some patches are entirely unrelated to the backport

All changes are needed to apply the following patches because they contain changes necessary. I tried to not change the commit itself.

* Hardware backports should rather happen in the LEDE 17.01 upstream than in Gluon, at least when they're possible without too many changes affecting other devices (which should be possible if all the unrelated patches are removed)

Feel free to do this instead.

NeoRaider:
Usually, you just want to cherry-pick commits from the LEDE master into lede-17.01 (or in the case of Gluon-specific changes, to the *lede* directory in the Gluon tree and create patch files using `make update-patches`. Using `make update-patches` is not optional; no other ways to create patch files are acceptable). You should always ensure the patch order makes sense using `git rebase -i`.

Many commits you backported contain changes that don't make sense, and are not part of the upstream commit. One example would be the commit "ar71xx: image: drop redundant kmod-usb-ohci from DEVICE_PACKAGES", which somehow adds new device sections to the file, which is not the case in the original patch. Other patches have the same issue.

Rather than not changing the patches, you should try to minimize the amount of changes necessary.

RubenKelevra:
Since you're not willing to accept a patch for hardware-support here anyway I'll wait until you fixed that upstream. :)

rubo77:
It would be great, if we get a running version out of the hints here in this pr and https://github.com/freifunk-gluon/gluon/issues/1123

rotanid:
many things would be great ;-)
but there's money and/or time blocking the way :-(

RubenKelevra:
I guess only LAN/WAN-Ports swap is missing here.

rubo77:
@RubenKelevra where can we test your branch? Here at the top it only sais "from unknown repository" do you have your gluon-repo on github?

I would like to build a firmware for our C25 too

RubenKelevra:
@rubo77 we use a custom build-tool for merging all of our patches of lede/gluon/etc to one firmware, gluon10 is our branch for 2017.1.x.

https://github.com/vfn-nrw/gluon-firmware-builder

rubo77:
I built RubenKelevra's branch https://github.com/rubo77/gluon/tree/rubo-c25 and this is the test so far:

- mesh and ibss both work.
- LAN/WAN switch didn't work, I only get into config mode on the blue port and DHCP doesn't work, I had to give an IP in 192.168.1.x range to access the node.
- also In normal mode, I only get a freifunk LAN ip on the blue port
- 5GHz client WiFi works fine

RubenKelevra:
Thanks for your testing. I’ve bought a C25 as well as a C59 and a C60 for the development to do the testing but they are not yet shipped. I’ll create a new branch/PR for the necessary changes on Gluon to support them if the LEDE support is done by NeoRaider.

I’m currently testing 2017.1 plus some master-changes as well as the C2600. When I’m done the C25 etc. will be the next stuff for us.

NeoRaider:
I don't have a test device and don't plan to work on any hardware support at the moment, so someone else would need to provide patches for lede-17.01.

RubenKelevra:
Merging patches from the maater to the stable branch on LEDE requires a lot more testing than just the small Gluon usecase.

I would recommend that we just backport the master patches onto 17.01 locally than doing the testing for LEDE - since stabilization of lede patches isn’t really our business.

NeoRaider:
I don't understand what "the small Gluon usecase" is supposed to be. Hardware support has pretty much nothing that is specific to Gluon.

If a backport works correctly with Gluon and doesn't touch any code that affects other hardware, it can also be pushed to lede-17.01 without problems.

RubenKelevra:
On the C2600 port for example the Wifi-LEDs still don’t work as expected. We also doesn’t use NAT on the hardware, doesn’t split the LAN-Switch to different ports etc. I would expect that those stuff is tested before merging it to a stable branch, wouldn’t you agree?

NeoRaider:
Not really. NAT is a software feature, that's not in any way hardware-specific. Good VLAN switch support is nice, but hardly something that is done by common users.

Generally, nobody explicitly tests such stuff before making a new stable release. Having some hardware support in LEDE only means that someone contributed patches. We (the people with commit access) don't have any way to verify what is working and what is not, so it's usually up to the contributors to decide what is tested and what not.

Even partial hardware support will end up in a stable release at some point, at least when the master becomes "stable" the next time; sometimes people notice that some features are broken, and sometimes even contribute fixes, but support for a device being in a stable release doesn't really mean that it is "working". Having a nice hardware test database that would give some information on the support state of different hardware is something people in the LEDE project have been working on, but it hasn't been realized yet.

Stable release backports shouldn't break things that have been working before; therefore we're very careful when doing backports that affect lots of hardware. If the new hardware doesn't work 100%, that's fine. If someone notices an issue, it can be reported, and if not fixed, at least be documented in the Wiki.

RubenKelevra:
Okay. I thought this would be handled differently on LEDE now, since this was a huge weakpoint on OpenWRT.

A split of the switch-chip start to be very commonly done to support devices like the Telekom Entertain devices which do IGMP on a different VLAN of the DSL connection to one device plugged into the home net router ... but yeah, I understand your point there.

I guess it would be nice to get some kind of test checklist in the future, so everybody can check a detail and mark the version where it worked. This would help to find new bugs by e.g. running a bisect much easier.

I still got no hardware, maybe at the weekend, I gonna try to merge the important patches and backport them. I will also do the (I find) necessary testing to add those parches to the stable branch.


Best regards 

Ruben

PR #1163, opened 2017-06-27
Add support for GL.INet GL-AR300M

H4ndl3:
This PR adds support vor the GL-AR300M. The device has two flash chips, a 16MB NOR flash and a 128MB NAND flash. Currently only the NOR flash can be used with LEDE (and therefore Gluon) as support for the NAND flash is still missing.

After I applied the settings in the config mode I got these errors:
error

However, everything seemed to be working fine after the node rebooted itself. It meshed instantly with my other node.

### DIFF
--- a/targets/ar71xx-generic
+++ b/targets/ar71xx-generic
@@ -60,6 +60,8 @@ device gl-inet-6416a-v1 gl-inet-6416A-v1
 
 device gl-ar150 gl-ar150
 factory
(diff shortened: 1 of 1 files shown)


NeoRaider:
0100-l2tp-vpn.lua is not part of Gluon, so the issue is in a site-specific package.

rotanid:
this has been backported to the v2017.1.x branch now

PR #1168, opened 2017-06-29
pkg gluon-node-info prepare for GluonSrcDiet

2tata:
move lua files to luasrc dir


NeoRaider:
You can't do this in two commits, each commit must result in a working state. I'll squash them when merging.

PR #1171, opened 2017-07-04
Apply site defined patches

PowerPan:
with this changes it is possible to apply site defined patches during the build process when calling "make update"

### DIFF
--- a/scripts/patch.sh
+++ b/scripts/patch.sh
@@ -13,6 +13,10 @@ GLUONDIR="$(pwd)"
 PATCHDIR="$GLUON_TMPDIR"/patching
 trap 'rm -rf "$PATCHDIR"' EXIT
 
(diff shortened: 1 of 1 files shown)


NeoRaider:
There are both conceptual and technical issues with this approach, so we can't accept this.

* The script is modifying files that are not in .gitignore during build
* A site directory should not need to be updated when updating inside a stable Gluon branch; patches are too fragile for this and often need to be adjusted when we update LEDE and package feeds
* With patches, you can make arbitrary changes to LEDE and the package feeds; I'd argue that the resulting version string shouldn't show an upstream commit ID anymore, as it isn't the same Gluon version anymore

So I see two options here:
* Submit useful patches to Gluon or LEDE upstream
* Add these patches to your local Gluon fork (and deal with patch rebase issues yourself)

PR #1173, opened 2017-07-05
[RFC] gluon-web-network: make poe_passthrough more generic

kb-light:
Naming of POE passthroug section changed from `gpio_switch_poe_passthrough` (in OpenWRT) to `poe_passthrough` (in Lede). Therefore it is not possible to enable POE passthroug on a node which was initaly setup with Lede based gluon. (This is a bug in v2017.1.x)

This pull request fixes that issue and prepares `gluon-web-network` for future devices like the ubnt-erx-sfp, which has multiple POE ports.

### DIFF
--- a/package/gluon-web-network/i18n/de.po
+++ b/package/gluon-web-network/i18n/de.po
@@ -22,6 +22,9 @@ msgstr "Deaktiviert"
 msgid "Enable PoE Passthrough"
 msgstr "PoE-Passthrough aktivieren"
 
+msgid "Enable PoE Power Port %s"
+msgstr "PoE-Ausgabe auf Port %s aktivieren"
+
 msgid "Enable meshing on the LAN interface"
 msgstr "Mesh auf dem LAN-Port aktivieren"
 
--- a/package/gluon-web-network/i18n/fr.po
+++ b/package/gluon-web-network/i18n/fr.po
@@ -22,6 +22,9 @@ msgstr "Désactivé"
 msgid "Enable PoE Passthrough"
 msgstr ""
 
+msgid "Enable PoE Power Port %s"
+msgstr ""
+
 msgid "Enable meshing on the LAN interface"
 msgstr "Activer le réseau MESH sur le port LAN"
 
--- a/package/gluon-web-network/luasrc/lib/gluon/web/model/admin/network.lua
+++ b/package/gluon-web-network/luasrc/lib/gluon/web/model/admin/network.lua
@@ -128,7 +128,15 @@ uci:foreach("system", "gpio_switch", function(s)
 		if not section then
 			section = f:section(Section)
 		end
-		local poe = section:option(Flag, s[".name"], translate("Enable " .. s.name))
+
+		local port = s.name:match("^PoE Power Port(%d*)$")
+		local name
+		if port then
+			name = translatef("Enable PoE Power Port %s", port)
+		else
+			name = translate("Enable " .. s.name)
+		end
+		local poe = section:option(Flag, s[".name"], name)
 		poe.default = uci:get_bool("system", s[".name"], "value")
 
 		function poe:write(data)
--- a/package/gluon-web-network/i18n/gluon-web-network.pot
+++ b/package/gluon-web-network/i18n/gluon-web-network.pot
@@ -13,6 +13,9 @@ msgstr ""
 msgid "Enable PoE Passthrough"
 msgstr ""
 
+msgid "Enable PoE Power Port %s"
+msgstr ""
+
 msgid "Enable meshing on the LAN interface"
 msgstr ""
 


NeoRaider:
Let's just rename the section in an upgrade script instead of changing the web interface. `package/gluon-core/luasrc/lib/gluon/upgrade/150-poe-passthrough` would be perfect for this, it needs to be adjusted for the new name anyways.

kb-light:
The idea of this PR was to additionally support PoE output on erx-sfp which has 5x PoE output.

I am going to adjust `package/gluon-core/luasrc/lib/gluon/upgrade/150-poe-passthrough` anyways.

NeoRaider:
Ah, that makes sense. I wonder what `package/gluon-core/luasrc/lib/gluon/upgrade/150-poe-passthrough` should do with multiple PoE outputs...

kb-light:
I added patches for `package/gluon-core/luasrc/lib/gluon/upgrade/150-poe-passthrough`.
As discussed on IRC, I won't rename the existing config section for PoE Passthrough.
In case, that we do not use the `poe_passthrough` flag within `site.conf`, I don't know what to expect as default. Maybe just keep it as it is, because it is PoE Power output and not passthrough on erx-sfp?

package/gluon-web-network/i18n/fr.po
```diff
@@ -19,8 +19,8 @@ msgstr "Automatique (RA/DHCPv6)"
 msgid "Disabled"
 msgstr "Désactivé"
 
-msgid "Enable PoE passthrough"
-msgstr ""
+msgid "Enable %s"
+msgstr "%s"
```
NeoRaider: This string must stay empty, as there is no translation.

package/gluon-web-network/luasrc/lib/gluon/web/model/admin/network.lua
```diff
@@ -122,15 +122,17 @@ if sysconfig.lan_ifname then
...
-if uci:get('system', 'gpio_switch_poe_passthrough') then
-	local s = f:section(Section)
-	local poe_passthrough = s:option(Flag, "poe_passthrough", translate("Enable PoE passthrough"))
-	poe_passthrough.default = uci:get_bool("system", "gpio_switch_poe_passthrough", "value")
-
-	function poe_passthrough:write(data)
-		uci:set('system', 'gpio_switch_poe_passthrough', 'value', data)
+uci:foreach("system", "gpio_switch", function(s)
+	if s[".name"]:match("poe") then
+		local section = f:section(Section)
```
NeoRaider: I don't think it makes much sense to put each option into a separate Section.

Note that it will be necessary to give each option a unique identifier instead of "poe_passthrough" (for example `s[".name"]`) when they're added to the same Section.

package/gluon-web-network/i18n/de.po
```diff
@@ -19,8 +19,8 @@ msgstr "Automatisch (RA/DHCPv6)"
 msgid "Disabled"
 msgstr "Deaktiviert"
 
-msgid "Enable PoE passthrough"
-msgstr "PoE-Passthrough aktivieren"
+msgid "Enable %s"
+msgstr "%s aktivieren"
```
NeoRaider: I don't really like that "PoE-Passthrough" won't be formatted correctly in German anymore. It would be nice to make the GPIO names translatable as well.

NeoRaider:
I have applied the first two patches.

I'm currently doing some treewide changes based on 57adb49de2c7fc4ca311a569c69e07ed20cefc9a, so the third patch will have some rebase conflicts, but nothing too bad I presume.

kb-light:
new version, not tested

rotanid:
@kb-light , has it been tested in the meantime? i think that would be good before the final review.

kb-light:
Currently it looks like this on an er-x-sfp.
english version:
image

german version:
image

NeoRaider:
I've applied the first commit for now, so we have this working at all with current Gluon (I'll also backport it to v2017.1.x).

NeoRaider:
Will need rebasing before we can get the rest merged.

package/gluon-web-network/i18n/de.po
```diff
@@ -22,6 +22,9 @@ msgstr "Deaktiviert"
 msgid "Enable PoE Passthrough"
 msgstr "PoE-Passthrough aktivieren"
 
+msgid "Enable PoE Power Port%s"
+msgstr "PoE-Ausgabe Port%s aktivieren"
```
kb-light: Maybe "PoE-Ausgabe auf Port%s aktivieren"?

kb-light:
Rebased

kb-light:
Updated:
image
(english)

image
(german)

rotanid:
@kb-light if you're finished and you have build & run-tested this, i suggest you merge the two commits into one and we'll merge it.

package/gluon-web-network/luasrc/lib/gluon/web/model/admin/network.lua
```diff
@@ -128,7 +128,14 @@ uci:foreach("system", "gpio_switch", function(s)
 		if not section then
 			section = f:section(Section)
 		end
-		local poe = section:option(Flag, s[".name"], translate("Enable " .. s.name))
+
+		local name, port = s.name:match("^(.+[^%d])(%d*)$")
```
NeoRaider: Please change this very generic match to one that explicitly matches `PoE Power Port(%d*)`.

package/gluon-web-network/luasrc/lib/gluon/web/model/admin/network.lua
```diff
@@ -128,7 +128,14 @@ uci:foreach("system", "gpio_switch", function(s)
...
+		local poe
+		if port then
+			poe = section:option(Flag, s[".name"], translatef("Enable " .. name .. " %s", port))
+		else
+			poe = section:option(Flag, s[".name"], translatef("Enable " .. name))
```
NeoRaider: This can use *translate* rather than *translatef*.

package/gluon-web-network/luasrc/lib/gluon/web/model/admin/network.lua
```diff
@@ -128,7 +128,14 @@ uci:foreach("system", "gpio_switch", function(s)
...
+		end
```
NeoRaider: Rather than duplicating the whole `section:option(...)`, only the label needs to be set inside the if/else.

kb-light:
updated, rebased, squashed and tested

PR #1175, opened 2017-07-06
add travis check

rubo77:
This script adds a travis check for the site.conf in docs/site-example/ and checks all shell scripts for validity

Travis has to be added to the gluon repository in Settings->Integrations & services->Installed GitHub Apps

You wil see a green hook or a red X in the commit history. example: https://github.com/rubo77/gluon/commits/travis

(Klick the green hook for details)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,14 @@
+language: bash
+
+addons:
+  apt:
+    packages:
+    - lua5.1
+    - luarocks
+  
+install:
+  - eval $(luarocks path --bin)
+  - luarocks install --local lua-cjson
+
+script:
+  - bash tests/validate_site.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,19 @@ addons:
     packages:
     - lua5.1
     - luarocks
-  
+    - git
+    - make
+    - gcc
+    - g++
+    - unzip
+    - libncurses5-dev
+    - zlib1g-dev
+    - subversion
+    - gawk
+    - bzip2
+    - libssl-dev
+    - wget
+
 install:
   - eval $(luarocks path --bin)
   - luarocks install --local lua-cjson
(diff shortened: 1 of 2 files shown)


NeoRaider:
* The site example isn't a valid configuration
* The build will take too long and use too much space
* We just spoke in IRC and the general opinion was that Travis is not very useful for Gluon

rubo77:
I added the gluon build in the travis sandbox, but it has currently only 3GB. Plus, there are two cores available.

So at the moment travis  stops after some time now without success.

Is there a way to just start the make command to just simulate a make without having to setup the whole lede build environment?

rubo77:
It works fine with only doing `make update` and `make clean`. And now ich checks all shell scripts in the whole gluon branch for validity.

I think, we get a lot of debugging information this way immediately after a commit. have a look:

https://travis-ci.org/rubo77/gluon/builds/251055441?utm_source=github_status&utm_medium=notification

rubo77:
I amended a lot in the travis branch at https://github.com/rubo77/gluon/commits/travis

can you reopen this pull request, so I can amend the changes here?

rotanid:
reopening isnt possible, as your branch isnt the original one anymore, according to github.

NeoRaider:
I don't see a reason to check the things you check. `make update` will only fail after *modules* updates, which I never update though Github PRs. Other checks rather validate the site than Gluon itself. 90% of commits/PRs don't change anything relevant to the build, so trying to build is just wasting CPU time.

All in all: Travis is not very useful for Gluon. Please just accept the "no", I have no intention to reopen this.

rubo77:
Ok, I thought, that if this would be in the main repo, then all forks would also get the travis functionality, we could provide checks here, that would show errors in new packages before you add them as PR already in your forked repository.

What kind of checks would be useful? I could think of **validity check for all lua files** as very useful.

PR #1179, opened 2017-07-07
Docs: standardize package descriptions and add missing example in site-example

rubo77:


docs/site-example/site.conf
```diff
@@ -211,6 +211,11 @@
     -- owner = {
       -- obligatory = true
     -- },
+    -- Show/hide the password login
+    -- remote_login = {
+      -- show_password_form = true,
+      -- min_password_length = 12
+    -- }
```
NeoRaider: Please terminate all lines using a comma, as it will reduce copy-and-paste induced syntax errors.

docs/package/gluon-ebtables-source-filter.rst
```diff
@@ -27,4 +27,6 @@ Example::
 
   prefix4 = '198.51.100.0/21',
   prefix6 = '2001:db8:8::/64',
-  extra_prefixes6 = { '2001:db8:9::/64', '2001:db8:100::/60' },
+  extra_prefixes6 = {
+    '2001:db8:9::/64', '2001:db8:100::/60'
```
NeoRaider: I think it would be nice to wrap this to have one prefix per line, and a comma after the last prefix for consistency.

docs/package/gluon-config-mode-geo-location.rst
```diff
@@ -12,3 +12,11 @@ config_mode.geo_location.show_altitude : optional
     - ``true`` enables the altitude field
     - ``false`` disables the altitude field if altitude has not yet been set
     - defaults to ``true``
+
+Example::
+
+  config_mode = {
+    geo_location = {
+      show_altitude = true
```
rotanid: we should be consistent in whether we add a comma after the last item or not.
i suggest you add one here too, like it is in most places in the docs

docs/package/gluon-ebtables-source-filter.rst
```diff
@@ -6,8 +6,8 @@ ruleset to prevent unreasonable traffic entering the network via the nodes.
 Unreasonable means traffic entering the mesh via a node which source IP does
 not belong to the configured IP space.
 
-One may first check if there is a certain proportion of unreasonable traffic,
-before adding this package to the firmware image. Additional one should not
+You may first check if there is a certain proportion of unreasonable traffic,
+before adding this package to the firmware image. Additional you should not
```
rotanid: there should be an "y" - "Additionally"
even better: "Furthermore, you should [...]"

docs/package/gluon-web-admin.rst
```diff
@@ -9,15 +9,17 @@ password length must be.
...
+  - sets the minimum allowed password length. Set this to ``1`` to disable the
+    length check.
+  - defaults to ``12``
 
-config_mode.remote_login.min_password_length \: optional (defaults to ``12``)
-  This sets the minimum allowed password length. Set this to ``1`` to
-  disable the length check.
-
-If you want to enable the password login you can use this example::
+Example::
```
rotanid: only one colon please ;-)

docs/site-example/site.conf
```diff
@@ -209,8 +209,13 @@
...
     -- define if the contact field is obligatory (optional)
     -- owner = {
-      -- obligatory = true
+      -- obligatory = true,
     -- },
+    -- Show/hide the password login
+    -- remote_login = {
+      -- show_password_form = true,
+      -- min_password_length = 12,
+    -- }
```
rotanid: same issue as in another comment: do we want to add a comma after the last entry?

docs/site-example/site.conf
```diff
@@ -209,8 +209,13 @@
     -- },
     -- define if the contact field is obligatory (optional)
     -- owner = {
-      -- obligatory = true
+      -- obligatory = true,
```
kb-light: Style of comments should be unified to. (see `roles`)

docs/site-example/site.conf
```diff
@@ -209,8 +209,13 @@
...
     -- },
+    -- Show/hide the password login
+    -- remote_login = {
```
kb-light: Did we really want to add all possible values to the `site-example`?

In my opinion there is no need for package dependent configuration in the `site-example`, when there is already an example in `docs/packages`

docs/package/gluon-web-admin.rst
rubo77: not sure, but i think, this is rst syntax and results in only one colon. it is like this in many places.

docs/site-example/site.conf
rubo77: Only examples that are interesting should be listed in the site-example. I added this, because many communities will wonder where the password login went and want it back.

docs/site-example/site.conf
rubo77: so this one is correct.

docs/site-example/site.conf
rubo77: is correct too

docs/site-example/site.conf
rotanid: no, there has to be a comma after the closing curly bracket to be consistent

rubo77:
Are all requests cleared now?

PR #1181, opened 2017-07-07
Update dns-cache.rst: name should contain a dot.

A-Kasper:
The next_node name should be a name with a dot in it to avoid (mobile) browsers interpreting "nextnode" as searchphrase. (You need to enter http://manually.. nobody wants this ;)
Another suggestion was to change it to gluon.nextnode, but that would be longer,

### DIFF
--- a/docs/features/dns-cache.rst
+++ b/docs/features/dns-cache.rst
@@ -31,7 +31,7 @@ of next_node.name may be resolved even without upstream connectivity.
   },
 
   next_node = {
-    name = 'nextnode',
+    name = 'next.node',
     ip6 = '2001:db8:8::1',
     ip4 = '198.51.100.1',
   }
--- a/docs/features/dns-cache.rst
+++ b/docs/features/dns-cache.rst
@@ -21,9 +21,9 @@ If these settings do not exist, the cache is not intialized and RAM usage will n
 
 When next_node.name is set, an A record and an AAAA record for the
 next-node IP address are placed in the dnsmasq configuration. This means that the content
-of next_node.name may be resolved even without upstream connectivity. It's suggested to use
+of next_node.name may be resolved even without upstream connectivity. It is suggested to use
 the same name as the DNS server provides: e.g nextnode.yourdomain.net (This way the name also 
-works if client uses static DNS Servers). Hint: If next_node.name doesn't contain a dot some 
+works if client uses static DNS Servers). Hint: If next_node.name does not contain a dot some 
 browsers would open the searchpage instead.
 
 ::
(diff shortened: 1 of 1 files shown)


NeoRaider:
Well, it's only an example. Usually, you'd use the same domain name your other DNS servers provide (in Lübeck that would be *node.luebeck.freifunk.net*).

christf:
I always wondered how I'd stop my browser from doing crazy things when entering "nextnode".
The example is not very clear. That being said I do not think this patch helps.

nextnode.domain or node.domain or nextnode.site_id or even nextnode.internaldomain could be considered to make this clearer. Maybe this behavior even justifies a comment in the site.conf

A-Kasper:
I've updated the PR and merged your feedback. I hope things are clearer now.

kpanic23:
Actually, in order to keep the browser from googling, all you have to do is **type a slash after the hostname**. Works for both (in our case) nextnode/ and nextnode.ff3l/

A-Kasper:
but in fact nearly nobody knows it and get to the searchpage. best way is
to use an internetwide dns entry for nextnode which is hardcoded in
dns-cache. so the same name is reachable independet from uplink or static
dns server.

Am 09.07.2017 10:29 schrieb "David Lutz" <notifications@github.com>:

docs/features/dns-cache.rst
```diff
@@ -31,7 +34,7 @@ of next_node.name may be resolved even without upstream connectivity.
   },
 
   next_node = {
-    name = 'nextnode',
+    name = 'nextnode.yourdomain.net',
```
skorpy2009: maybe, your search domain?

A-Kasper: Not every community uses a searchdomain. My intention was to keep it as universal as possible. For me the doc should contain examples which are working for the most of the users. searchdomain would confuse some....

christf: I quite like it now. one minor style-thing: short forms "it's", "doesn't" usually is not used in written language

rotanid: @A-Kasper do you want to implement that last little change request?

rotanid:
@christf please check it a last time and if it is OK for you, merge it as a single commit with an optimized commit message

A-Kasper:
I've put it in a single commit as requested: #1207

rotanid:
thanks, but that doesn't need a new pull request, you can simply "rebase" your commits and force push them here into this branch.
also, we can do that during the merge, too, so you don't necessarily have to do it yourself and that's why i asked @christf to do it ;-)

A-Kasper:
sorry... i'm not very familiar with git... thank you

christf:
somehow I need to re-check my email notification settings. This has been sitting for 9 days now without me noticing it. I apologize.

PR #1191, opened 2017-07-15
[WIP][BUGFIX] fix PoE passthrough

kb-light:
this splitts #1173 in two PRs to first fix the current bug in `master` and `v2017.1.x` and then may add options for future devices.

This is not yet tested.


NeoRaider:
Systems upgraded from older releases will still use the old section name.

I like the support for multiple PoE GPIOs in #1173, and would backport it to v2017.1.x as a whole as long as translations are added, so the displayed string looks the same as before.

NeoRaider:
I'll close this in favor of #1173.

PR #1192, opened 2017-07-16
Gluon-mesh-vpn-core: Moved migration code to own script

### LINKED ISSUE #1187: fastd -> tunneldigger upgrade-problem
Hi :)

If an old version of tunneldigger was installed and fastd enabled the upgrade wont enable tunneldigger with removed fastd.

The Problem is that "400-mesh-vpn" comes before "400-mesh-vpn-tunneldigger" in upgrade-process.
I solve this by rename "400-mesh-vpn-tunneldigger" to "399-mesh-vpn-tunneldigger".

---

https://www.freifunk-gera-greiz.de/forum/-/message_boards/message/115277#_19_message_115050
https://www.freifunk-gera-greiz.de/forum/-/message_boards/message/115277#_19_message_115210

lcb01a:
It is mandatory that it runs after the initial configuration is in place.

This fixes #1187. Basically we are back to having the VPN migration code in its own script file as it has to run after the daemon configuation has been written. Otherwise we run into the problem that a section we want to enable is not yet existing when migrating from fastd <-> tunneldigger.

### DIFF
--- a/package/gluon-mesh-vpn-core/luasrc/lib/gluon/upgrade/400-mesh-vpn
+++ b/package/gluon-mesh-vpn-core/luasrc/lib/gluon/upgrade/400-mesh-vpn
@@ -58,19 +58,3 @@ uci:section('firewall', 'include', 'mesh_vpn_dns', {
 })
 
 uci:save('firewall')
-
-
--- VPN migration
-local has_fastd = fs.access('/lib/gluon/mesh-vpn/fastd')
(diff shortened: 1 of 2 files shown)


NeoRaider:
Thank you for analyzing this issue, and sorry for taking so long to take a look at your solution. I eventually went for a more involved fix (9524398), it would be great if you could give it some more testing (I checked upgrades between the official Gluon fastd and tunneldigger packages, but not from the old tunneldigger.)

PR #1199, opened 2017-07-29
batman-adv: fix TT sync flag inconsistencies

T-X:
This patch did not make it into the upstream batman-adv repository yet. However, Antonio has had a look at it too and Ack'ed it a few days ago. Therefore, I think this fix should be suitable at least for Gluon master for now.

For a Gluon stable backport, it should get a little more field testing first, though. (Testers very welcome!)

Upstream reference: https://patchwork.open-mesh.org/patch/17072/


mweinelt:
The patch is now part of the maint branch, Simon has merged it.

A-Kasper:
to test it  is necessary that also the supernode batman-adv is fixed. Can't
test it until they release the fixed batman version

maybe we could change to the fixed version in gluon but keep mo
deactivated. so we would be able to enable MO on supernode level and have a
look if the nodes are getting unstable. If not we could try to enable MO in
Gluon again an test it.



Am 31.07.2017 15:23 schrieb "hexa-" <notifications@github.com>:

mweinelt:
Our gateways are upgraded and the experimental firmware is rolling out tonight.

NeoRaider:
First patch obsoleted by c66b18ca9e773e43b3ec83f5b9a2ba4605d7d506, second patch applied.

PR #1200, opened 2017-07-29
RFC: batman-adv: fix TT sync flag inconsistencies

T-X:
This  pull request is identical to the one for the master branch (#1199). However I thought of leaving this here as a minimal changeset on top of a stable branch for people who would like to test it maybe on a larger scale, without dragging in all other potential changes from the master branch.

Upstream reference: https://patchwork.open-mesh.org/patch/17072/


A-Kasper:
I'm not able to test it until I know how to instaall it at supernode level.
If Supernodes have broken MO we  fon't need to start....

Am 29.07.2017 05:51 schrieb "T-X" <notifications@github.com>:

NeoRaider:
First patch obsoleted by b0a6cb72269df7105d6d843304910fd001711594. I'd like to keep the mcast optimizations disabled for the upcoming v2017.1.2 release, let's consider it again for v2017.1.3.

A-Kasper:
feedback:
I've updated my supernodes (gateways) to the fixed batman version and enabled mm. The rebooting and load issue seem to start again. 

but:
1. I have an heterogen network, where are also nodes without the fix (but with disabled mm)
2. I had found an mistake in my vm host, where nf_conntrack was to short, so it dropped packets of the gateways.... 

--> can't say something reliable.. will report again later... 
Has someone else tried to enable mm at gateway level after upgrading to latest gluon?

PR #1201, opened 2017-08-04
Add pkg gluon-geolocator from google summer of code 2017 

2tata:
This package makes able to receive positions base on surrounding WIFIs.

the geolocator program takes surrounding BSSIDs and send this to the openwifi project.
As an answer the router will get a position base on the above sent BSSIDs. This will be done
in a defined interval, which can be set by site.conf or by configMode.

Following site.conf parameters a definable for community's:
* `autolocation` enable flag. Can be 1 or 0 means enabled or disabled. Default is disabled.
* `interval` in minutes, means how often should the geolocator refresh its position. Default is every 12 hours.

Here an example site.conf part:
```
geolocator = {
                autolocation = true,
                interval = 1000,
             },
```

This package is over a year tested in the Nordwest Freifunk network. 

This is one of 4 other packages there should follow in the next few weeks.
2 of the other packages are responsible for the configuration over the configMode.
https://blog.freifunk.net/2017/07/26/geolocator-software-defined-gps-second-evaluation/


rubo77:
Is it possible for the user to select, that he/she doesn't want to be discovered?

So you have some screenshots of the admin interface?

2tata:
This pkg contains only the geolocator. The admin interface comes in the 3 other above mentioned pkgs.
Users can set share_location [0] to discover there positions.

As well users have to be enable the service by setting auto_location [1] otherwise the geolocator will be disabled.

[0] https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-node-info/files/etc/config/gluon-node-info
[1] https://github.com/freifunk-gluon/gluon/pull/1201/files#diff-ca9821c5dfe71d5edd884409cceef57bR30

package/gluon-geolocator/check_site.lua
```diff
@@ -0,0 +1,5 @@
+if need_table('geolocator', nil, false) then
+  need_number('geolocator.autolocation', false)
+  need_number('geolocator.interval', false)
+  need_string_array_match('geolocator.blacklist', '^%w+:%w+:%w+:%w+:%w+:%w+$', false)
```
NeoRaider: Please use tabs instead of spaces.

package/gluon-geolocator/files/lib/gluon/geolocator/geolocator.sh
```diff
@@ -0,0 +1,65 @@
...
+  [ $? -eq 1 ] && Clean_pid
+  #ceck if static location true or not
+  [ "$(uci get "${GLC}".static_location)" -eq 0 ] && {
+    uci set "${LOC}".latitude="$LAT"
+    uci set "${LOC}".longitude="$LON"
+    uci commit gluon-node-info
+  }
+  date +%s > $TIME_STAMP
+fi
+Clean_pid
```
NeoRaider: I have not competely reviewed this script yet, but
* it should use tabs instead of spaces
* please add blank lines to group related commands
* there are some unneeded semicolons
* don't use {} instead of if/fi that much

Also, parsing the output of *iw* is generally a bad idea, and that part of the script is very complex. Using libiwinfo-lua would probably result in more maintainable code...

package/gluon-geolocator/files/usr/lib/micron.d/geolocator
```diff
@@ -0,0 +1 @@
+*/5 * * * *     sh /lib/gluon/geolocator/geolocator.sh;
```
NeoRaider: * Unneded semicolon
* Remove the `sh`, the script is executable and has a shebang line

package/gluon-geolocator/luasrc/lib/gluon/upgrade/540-geolocator-settings
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/lua
+
+local site = require 'gluon.site'
```
NeoRaider: You switched to the new *gluon.site* library, but didn't adjust the rest of the code accordingly... see other comments below.

package/gluon-geolocator/luasrc/lib/gluon/upgrade/540-geolocator-settings
```diff
@@ -0,0 +1,33 @@
...
+local uci = require('simple-uci').cursor()
+
+local config = 'geolocator'
+
+local static_location = uci:get(config, 'settings', 'static_location')
+
+local auto_location = uci:get(config, 'settings', 'auto_location')
+if not auto_location then
+  auto_location = 0
```
NeoRaider: `auto_location = site.geolocator.autolocation(false)`

package/gluon-geolocator/luasrc/lib/gluon/upgrade/540-geolocator-settings
```diff
@@ -0,0 +1,33 @@
...
+  if site.geolocator.autolocation then
+    auto_location = tonumber(site.geolocator.autolocation)
+  end
+end
+
+local refresh_interval = uci:get(config, 'settings', 'refresh_interval')
+if not refresh_interval then
+  refresh_interval = 43200 --12h
```
NeoRaider: `refresh_interval = site.geolocator.interval(43200) -- default: 12h`

package/gluon-geolocator/luasrc/lib/gluon/upgrade/540-geolocator-settings
```diff
@@ -0,0 +1,33 @@
...
+    refresh_interval = tonumber(site.geolocator.interval)
+  end
+end
+
+uci:delete(config, 'settings')
+uci:section(config, config, 'settings', {
+  static_location = static_location,
+  refresh_interval = refresh_interval,
+  auto_location = auto_location,
+  blacklist = site.geolocator.blacklist,
```
NeoRaider: `blacklist = site.geolocator.blacklist(),`

package/gluon-geolocator/check_site.lua
```diff
@@ -0,0 +1,5 @@
+if need_table('geolocator', nil, false) then
+  need_number('geolocator.autolocation', false)
```
NeoRaider: Should be *need_boolean*

package/gluon-geolocator/luasrc/lib/gluon/upgrade/540-geolocator-settings
2tata: done in 1acc557

rubo77:
Why is the text below the dropdown with such a big padding?

2tata:
On 08/21/17 07:56, Ruben Barkow wrote:
> Why is the text below the dropdown with such a big padding?
> 

I dont know which text do you mean?

vg
Tarek

2tata:
@NeoRaider now every requested changes should be done. :-)

rotanid:
status update from dev meeting:
- needs rebase for check_site
- awaiting review

NeoRaider:
I have started some code cleanup, unfortunately I can't test it at the moment, as `http://openwifi.su/api/v1/bssids/...` only returns "502 Bad Gateway"...

NeoRaider:
Btw, is the quality of the openwifi.su data high enough so that the geolocator package will return even marginally useful data? I just tested this with a node in Leipzig, and I got a location in Braunschweig (the scan returned 46 BSSIDs on 2.4GHz, so there should be sufficient data to avoid getting the location *that* wrong?)

Anyways, I've just pushed the first part of my rework (which has turned out to be mostly a rewrite...) as https://github.com/freifunk-gluon/packages/commit/53a659abf8809f36f2da130bb920a41ac6b698cb, a reusable Lua library for geolocation. In addition to code cleanup, I changed the following:
- use ubus to get the interface list from netifd instead of using UCI (nicer interface, and we don't depend on the optional `ifname` UCI option)
- only query one interface per radio

2tata:
> Btw, is the quality of the openwifi.su data high enough so that the geolocator package will return even marginally useful data? I just tested this with a node in Leipzig, and I got a location in Braunschweig (the scan returned 46 BSSIDs on 2.4GHz, so there should be sufficient data to avoid getting the location that wrong?)

You need to add the mesh BSSIDs to the blacklist as i wrote above. The openwifi.su backend needs atleast one known WiFi network in the database. For places where nearly no wifi is known in the database, you can just use the openWLANmap wardriving App for Android [0]. The wardriving App will may replace in the coming months because this is in the GSoC18 as a Project for Freifunk. I started work on a new backend witch will also use Mozilla Location service in addition to the own database. This was my last Google Summer of Code project last year. I just got the idea if I change the the request url to a siteconf variable local communitis without v6 can use my GSoC17 project as a MLS proxy [1]. So I will make the request url variable.

[0] http://openwifi.su/owlmapatandroid.php?lang=en
[1] https://github.com/openwifi-su/sgps-core

> Anyways, I've just pushed the first part of my rework (which has turned out to be mostly a rewrite...) as freifunk-gluon/packages@53a659a, a reusable Lua library for geolocation. In addition to code cleanup,

I explicitly change it from a C-Lib into the geolocator corn to avoid airtime wasting by multiple access on scanning. For example before the geolocator, the hoodselector and an autolocator (witch place the location in the uci config) where using the C-Lib that means scanns twice for the same result. The geolocator includes the idea to update location position in an intervall where other programs can access. So maybe we should keep the lib out to force using the uci interface. 

> I changed the following:

> use ubus to get the interface list from netifd instead of using UCI (nicer interface, and we don't depend on the optional ifname UCI option)

ubus was also my plan but I drop this idea while I was discussing with people from the Mozilla Location Service project and Zeeshan Ali who already develop a software calls gps-share to provide positions via ubus by using GPS HW in the background.

2tata:
>>>  Btw, is the quality of the openwifi.su data high enough so that the geolocator package will return even marginally useful data? I just tested this with a node in Leipzig, and I got a location in Braunschweig (the scan returned 46 BSSIDs on 2.4GHz, so there should be sufficient data to avoid getting the location that wrong?)
>>
>> You need to add the mesh BSSIDs to the blacklist as i wrote above. The openwifi.su backend needs atleast one known WiFi network in the database.
>
> Okay, after blacklisting the FFHL IBSS cell ID I get 0 results now (why is the FFHL IBSS located in Braunschweig though...?).

Because some ware has use the ffhl IBSS cell ID there or just setup a FFHL router and its Cell ID got catch by the war driving app.

> I think I will just change it to ignore all non-AP WLANs, there should be no need to blacklist anything by default then.

Ah yes this is a better idea. The new war driving app, which is planed as a GSoC18 project, should also ignore non-AP WLANs and even mobile hotspots.

>> I explicitly change it from a C-Lib into the geolocator corn to avoid airtime wasting by multiple access on scanning. For example before the geolocator, the hoodselector and an autolocator (witch place the location in the uci config) where using the C-Lib that means scanns twice for the same result. The geolocator includes the idea to update location position in an intervall where other programs can access. So maybe we should keep the lib out to force using the uci interface.
>
> Did you even look at my commit? I built a small Lua library from the code of this PR, it has nothing to do with a C library.

I think you miss understood me. I look at your commit and you created a lua lib witch can be called from other lua programs. Above I explained that each program witch use your lib will also call a wifi scan witch will waist airtime, in case if more than one program will request its position in an interval. This problem was happens before the geolocator exists by using a C-lib witch did a similar work like your lua lib on our FFNW routers. That is why I developed the geolocator and why I use the uci interface for providing actual coordinates. I think you lib is not a good way because of the airtime issue or do you think this is not relevant?

>> ubus was also my plan but I drop this idea while I was discussing with people from the Mozilla Location Service project and Zeeshan Ali who already develop a software calls gps-share to provide positions via ubus by using GPS HW in the background.
> 
> I only use ubus to get the interface list from netifd, instead of getting them from /etc/config/wireless via UCI.

Ah pardon, it was late yesterday ;)

> I have a few more ideas regarding this PR. WIll probably replace my Lua library again with one implementing the standard GLS API provided by Google and Mozilla (I assume I will implement it in C, to be able to HTTP POST). To use openwifi.su, they would need to provide a GLS endpoint (or a simple proxy would need to be written); I don't really want to support their non-standard API after all.

The problem by the GLS services by google was form me it is not opensource and I think you will need a API key

The key request was also a problem for GLS service by Mozilla

Both above projects did only provide HTTPS request witch means we have to put SSL on gluon. (The main reason why I didn´t use MLS)

> Furthermore, I'm unsure about what to do with the determined locations. Your PR currently writes it to the flash, which is not acceptable. I think it would be a nice solution to use a UCI state file in /var/state for it, so the results are only kept in RAM (a command to explicitly commit this state to flash can still be provided).

Yes I was also not sure about that. Another Idea which I got was to write a position only for once into the flash and refresh this position only if a tolerance value is overreach. Holding a position only in the RAM is not really optimal on reboots for programs for example  the hoodselector. Because that can mean in the worse case the hoodselector will change its hood 2 times after reboot. But at this point I am also not sure..

NeoRaider:
> I think you miss understood me. I look at your commit and you created a lua lib witch can be called from other lua programs. Above I explained that each program witch use your lib will also call a wifi scan witch will waist airtime, in case if more than one program will request its position in an interval. This problem was happens before the geolocator exists by using a C-lib witch did a similar work like your lua lib on our FFNW routers. That is why I developed the geolocator and why I use the uci interface for providing actual coordinates. I think you lib is not a good way because of the airtime issue or do you think this is not relevant?

The library is just code refactoring. There will be only a single program using the library in Gluon.

Regarding the HTTPS issue: I'm not sure how to proceed here. The reasoning of Mozilla etc. not to support HTTP makes sense; for privacy reasons the scan data should not be sent over unencrypted channels.

2tata:
>> I think you miss understood me. I look at your commit and you created a lua lib witch can be called from other lua programs. Above I explained that each program witch use your lib will also call a wifi scan witch will waist airtime, in case if more than one program will request its position in an interval. This problem was happens before the geolocator exists by using a C-lib witch did a similar work like your lua lib on our FFNW routers. That is why I developed the geolocator and why I use the uci interface for providing actual coordinates. I think you lib is not a good way because of the airtime issue or do you think this is not relevant?
> 
> The library is just code refactoring. There will be only a single program using the library in Gluon.

Ah ok.

> Regarding the HTTPS issue: I'm not sure how to proceed here. The reasoning of Mozilla etc. not to support HTTP makes sense; for privacy reasons the scan data should not be sent over unencrypted channels.

I am also not sure if MLS provides v6 as you probably read in the IRC

regarding to the privacy reasons, but what about all the respondd datas from gluon there available over public v6 ?

like:
http://[fd74:fdaa:9dc4:6800:12fe:edff:fe7b:2836]/cgi-bin/interfaces
http://[2a03:2260:1001:6800:12fe:edff:fe7b:2836]/cgi-bin/nodeinfo
etc. ?

2tata:
@NeoRaider I have change the geolocator to only request with Master mode driven scanned devices.
And I applyed your ubus request. Still the question how to handle the flash writing issue..

rubo77:
You could Set-up a Proxy service somewhere that accepts HTTP requests and Redirects them to https requests to the Mozilla Service

2tata:
@rubo77 MLS has no v6

2tata:
https://github.com/openwifi-su/sgps-core
An another project of my is already an MLS proxy. But without security.

But I dont like MLS that much... they keep all data closed and even for the service you need to ask them for a API key and you have to explain them why do you need it. I didint got a Key when I have told them I have build an MLS SSL Proxy... So I have not so good experience with there service.

ngobibibnbe:
goodmornig i am a student in ENSP in yaounde; our teacher talked us  about the work of gluon and Gsoc for the evolution of freifunk which really intrest me and for now i want to integrate the geolocalisation packages in the sources codes of openwrt and i have many difficulties to find the road of the work that i should do to achieve this integration.i am needing advices please .tks

T-X:
@ngobibibnbe: Awesome! Feel free to hop in on the IRC channel:
https://gluon.readthedocs.io/en/v2018.2.x/dev/basics.html#irc
There are many people hanging out who are probably able to help you. You might need to wait some minutes to get a response though.

Also, there is a mailing list available as noted here: https://github.com/freifunk-gluon/gluon. (But it wasn't used much recently. Most discussions happen on the IRC channel)

2tata:
@rotanid I did a rebase.

package/gluon-geolocator/files/usr/lib/micron.d/geolocator
```diff
@@ -0,0 +1 @@
+*/5 * * * *     /lib/gluon/geolocator/geolocator
```
mweinelt: How often would you expect someone to move their router? Five minutes is unusuable for a moving target, so that's probably not the use case. I'd say several hours (6/12/24) is more appropriate here.

package/gluon-geolocator/luasrc/lib/gluon/geolocator/geolocator
```diff
@@ -0,0 +1,130 @@
...
+
+local pos = locate()
+if not next(pos) then
+	os.exit(1)
+end
+if not uci:get_bool(GLC, "settings", "static_location") then
+	uci:set(LOC, uci:get_first(LOC, 'location'), 'latitude', pos.lat)
+	uci:set(LOC, uci:get_first(LOC, 'location'), 'longitude', pos.lon)
+	uci:save(LOC)
+	uci:commit(LOC)
```
mweinelt: With the given crontab this would cause a write to flash every five minutes, which is a bit excessive for my taste. In fact why commit at all, when you can tell the location didn't change significantly?

package/gluon-geolocator/luasrc/lib/gluon/geolocator/geolocator
```diff
@@ -0,0 +1,130 @@
...
+				end
+			end
+		end
+
+		return true
+	end)
+
+	assert(#found_bssids > 0, 'insufficient BSSIDs found')
+
+	local data = receive_json('http://openwifi.su/api/v1/bssids/' .. table.concat(found_bssids, ','))
```
mweinelt: I'm having a hard time with the privacy aspect of announcing your neighbouring WiFi networks in plain over the internet. What's it gonna take to get HTTPS up and running here? Just under 400 kB.

```
root@64289-rocket-carrier:~# opkg info libustream-wolfssl
Package: libustream-wolfssl
Version: 2018-07-30-23a3f283-1
Depends: libc, libubox, libwolfssl
Status: unknown ok not-installed
Section: libs
Architecture: powerpc_8540
Size: 4444
Filename: libustream-wolfssl_2018-07-30-23a3f283-1_powerpc_8540.ipk
Description: ustream SSL Library (wolfssl)

root@64289-rocket-carrier:~# opkg info ca-bundle
Package: ca-bundle
Version: 20180409-2
Depends: libc
Status: unknown ok not-installed
Section: base
Architecture: all
Size: 120447
Filename: ca-bundle_20180409-2_all.ipk
Description: System CA certificates as a bundle

root@64289-rocket-carrier:~# opkg info libwolfssl
Package: libwolfssl
Version: 3.15.3-stable-1
Depends: libc
Provides: libcyassl
Status: install ok installed
Section: libs
Architecture: powerpc_8540
Size: 270414
Filename: libwolfssl_3.15.3-stable-1_powerpc_8540.ipk
Description: wolfSSL (formerly CyaSSL) is an SSL library optimized for small
 footprint, both on disk and for memory use.
Installed-Time: 1549301037
```

And then I noticed the site has a broken HTTPS setup since some time in 2017:

```
openwifi.su uses an invalid security certificate. 
The certificate expired on 23. October 2017, 04:20:00 GMT+2. 
The current time is February 9, 2019, 4:01 AM. Error code: SEC_ERROR_EXPIRED_CERTIFICATE
```

package/gluon-geolocator/files/usr/lib/micron.d/geolocator
2tata: Could be a way, but then I would prefer a reboot script due to the interval time will be configurable over the configMode. But that also mean if some one changes the interval time via uci, he also need to rerun the reboot script manually...

package/gluon-geolocator/luasrc/lib/gluon/geolocator/geolocator
2tata: I am also not so happy with pushing though data plain over the internet. But the images will be to big with ssl on board. I have an another Idea to only send hashes of SSID+mac but this need time to implement and also time to be filled with datas. So actually there is only this solution.

T-X:
I still do not quite like the idea of adding and exposing new configuration options to the user for things that could easily be automized, too. I know, it might seem like just a small new knob. But a lot of small knobs make it messy, too.

What about focusing on the the main use-case first, which are stationary nodes (> 99.9% of current nodes)? And setting a fixed, non-configurable, longer interval first.

And then, later, after this PR was merged implement a small daemon written in C which:
* checks via a station dump on the mesh interface every 30 seconds or so if the neighborhood has changed significantly.
* if so, performs an early update.

(the advantage of a station dump would be that it's not as disruptive as a wifi scan)

package/gluon-geolocator/luasrc/lib/gluon/geolocator/geolocator
T-X: Maybe just always store it in /tmp/? (and update its readers accordingly)

package/gluon-geolocator/luasrc/lib/gluon/geolocator/geolocator
mweinelt: At some point we are going to cross the 4 MB either way. I'm not afraid of that, it just means lower end device will not be able to receive all possible features, which is reasonable, given the size constraint. On the other hand we would be able to build more interesting applications for properly sized routers.

mweinelt:
I agree that we can currently only properly target stationary nodes. A mobile application (moving target) would need to run at least at the lower bound of respondd, which is one minute, and I'm not even sure that this would be a proper interval if you wanted to see the movement on the map.

NeoRaider:
Also note that, at least with BATMAN IV, moving nodes while online is generally a bad idea, especially with the long originator interval we use. (I have no idea how BATMAN V or Babel perform in such usecases)

2tata:
@NeoRaider @T-X @mweinelt @rotanid I have rewrite parts of the geolocator. Now the corn file will be rewritten by an upgrade script and the interval has change to a hour value (default 12h). Also the geolocator will write only ones on the falsh if no position has been stored before. After that the node needs to move ~ +- 22m to write again. All other following requested position will be holded in RAM.

rubo77:
For info: This is blocked due to a missing encryption solution.

Maybe we can implement a simple stand alone encryption somehow, without Https?

package/gluon-geolocator/luasrc/lib/gluon/geolocator/geolocator
Adorfer: just for notice, openwifi.su has fixed the outdated ssl certificate aprox 2019-06-12

Adorfer:
> This is blocked due to a missing encryption solution.

Question: Would we be OK to go with a symetric (PSK) encryption? 
(I doubt it's suffienct for most people here. question is just: OK or NOK? In case of "good enough" we could look at stuff like https://github.com/kokke/tiny-AES-c )

mweinelt:
Pre-shared key (PSK) means at least every node knows the secret, which makes the key public knowledge, which makes it an insecure solution for secure encryption.

Adorfer:
> Pre-shared key (PSK) means at least every node knows the secret, which makes the key public knowledge, which makes it an insecure solution for secure encryption.

Thank you explaining why it is insecure. I hoped to have done this already with my post. 
that level of encryption "good enough" to unblock?

in other words: the blocker is not stating the attack vector to be mitigated, it's just "due to no encryption", not "due to missing asymetric encryption".

mweinelt:
TLS support in libustream will cost us roughly 300 kB:

```
# opkg info libustream-mbedtls
Package: libustream-mbedtls
Version: 2018-07-30-23a3f283-1
Depends: libc, libubox, libmbedtls
Status: install user installed
Section: libs
Architecture: powerpc_8540
Size: 5008
Filename: libustream-mbedtls_2018-07-30-23a3f283-1_powerpc_8540.ipk
Description: ustream SSL Library (mbedtls)
Installed-Time: 1557417673

root@64289-rocket-carrier:~# opkg info libmbedtls
Package: libmbedtls
Version: 2.16.1-1
Depends: libc
Status: install user installed
Section: libs
Architecture: powerpc_8540
Size: 177391
Filename: libmbedtls_2.16.1-1_powerpc_8540.ipk
Description: The aim of the mbedtls project is to provide a quality, open-source
 cryptographic library written in C and targeted at embedded systems.
 This package contains the mbedtls library.
Installed-Time: 1557417673

root@64289-rocket-carrier:~# opkg info ca-bundle
Package: ca-bundle
Version: 20190110-1
Depends: libc
Status: install user installed
Section: base
Architecture: all
Size: 116000
Filename: ca-bundle_20190110-1_all.ipk
Description: System CA certificates as a bundle
Installed-Time: 1557417673
```

CodeFetch:
@2tata Checking the node's location using a cron-job should not be the default behavior unless it is configured as "mobile". After a number of tests it should assume the data to be sufficient enough (and e.g. take an average value). I don't want a fixed node to jump around the map. Furthermore I don't know whether it is a good idea to just set the location like you would do manually. At least it needs a flag that this location was derived automatically. A location derived by trilateration might be enough to determine the correct domain, but e.g. a heat-mapping software will give erroneous results when it's being fed with approximated positions.

2tata:
On June 30, 2019 2:20:05 PM UTC, Vincent Wiemann <notifications@github.com> wrote:
>@2tata Checking the node's location using a cron-job should not be the
>default behavior unless it is configured as "mobile". After a number of
>tests it should assume the data to be sufficient enough (and e.g. take
>an average value). I don't want a fixed node to jump around the map.

The node will not jump around a location. I have never seen this behavior.

>Furthermore I don't know whether it is a good idea to just set the
>location like you would do manually. At least it needs a flag that this
>location was derived automatically. A location derived by trilateration
>might be enough to determine the correct domain, but e.g. a
>heat-mapping software will give erroneous results when it's being fed
>with approximated positions.

There is already an option to disable this service. The result is normally good enough for headmaps we already have proof this service in our network.

People are even able to improve the service dataset by using a war driving app

CodeFetch:
WLAN trilateration data has been mostly collected by mobile phones with GPS. As those mobile phones don't receive the GPS signals within the houses it's mostly data on street-levels. And mobile phones have undirected antennas. That means that the coordinates of a router using this service will always be inaccurate to a specific level. 

> The node will not jump around a location. I have never seen this behavior.

Nonetheless it's not nice to have this job running on stationary nodes after they have retrieved coordinates. The user knows well whether this node will ever be mobile and if he or she is lazy enough to not enter proper coordinates he or she should at least define whether this node is mobile or not. 

And I still doubt your statement, because I've written a software for WLAN trilateration a while ago, a reason why a friend of mine was offered a job at Google, and I know for sure that considering the RX power of beacons with a free line of sight, a great dataset, and without any reflections the best accuracy you can get is 10 meter. With routers behind walls on such a long distance, using such a bad dataset, it would be astonishing to have a accuracy of less than 30 meter.

> The result is normally good enough for headmaps we already have proof this service in our network.

If you have a proper heat-mapping software that respects the walls, buildings and heights above ground, it's not. I've written a software for this. Thus I know what I'm talking about. You need to mark these values as automatically retrieved somehow. They are useless for proper radio coverage interpolation.

mweinelt:
Hey @2tata,

we just talked this through and decided that with its privacy impact we are not going to merge this package.

hexa-

PR #1207, opened 2017-08-17
Update dns-cache.rst: next_node.name suggestion

A-Kasper:
This commit adds a suggestion for next_node.name, changes the example to nextnode.yourdomain.net and adds a hint, that browsers may open the searchpage if the hostname does not contain a dot.


rotanid:
please, let's continue with the existing pull request, so the relation between the discussion and the merge stays intact. i wrote additional information over there.

PR #1209, opened 2017-08-17
Doc: Adding GLUON_RELEASE for make manifest

SmithChart:
Adding the fact that GLUON_RELEASE is needed with 'make manifest'
if it is used with 'make'.

This fact was mentioned in the v2017.1 release notes but is not (clearly)
mentioned inside other parts of the documentation.

Signed-off-by: Chrissi^ <chris@tinyhost.de>


NeoRaider:
Thanks, I've applied your patch with some rewording.

PR #1210, opened 2017-08-19
docs: fastd: recommend better default mtu

mweinelt:
It is known that a fastd MTU of 1280 bytes will cause excessive fragmentation for full packets on networks that advertise to transport payloads of 1280 bytes, because it neglects the headers that are being added by the batman-adv encapsulation.

The following graphic is widely recognized as factually correct and it depicts this issue quite well.
MTU diagram for batman-adv networks

I therefore think the default documentation should reflect this by at least recommending a MTU value that allows for unfragmented IPv6 communication to pass the network, decreasing the burden on all of our embedded routers.

This will in turn allow new communities, who certainly heavily rely on gluons documentation, to make better choices when setting up their networks.

Thanks to @ohrensessel and others who didn't let the MTU discussion die, so we can build better networks.

### DIFF
--- a/docs/site-example/site.conf
+++ b/docs/site-example/site.conf
@@ -100,7 +100,7 @@
 
   mesh_vpn = {
     -- enabled = true,
-    mtu = 1280,
+    mtu = 1312,
 
     fastd = {
       -- Refer to http://fastd.readthedocs.org/en/latest/ to better understand
--- a/docs/user/faq.rst
+++ b/docs/user/faq.rst
@@ -15,3 +15,84 @@ interface. This DNS server must be announced in router advertisements (using
 on *batman-adv*. If your mesh does not have global IPv6 connectivity, you can setup
 your *radvd* not to announce a default route by setting the *default lifetime* to 0;
 in this case, the *radvd* is only used to announce the DNS server.
+
+
+What is a good MTU on the mesh-vpn
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Setting the MTU on the transport interface requires careful consideration, as
+setting it too low will cause excessive fragmentation and setting it too high
+may leave peers with a broken tunnel due to packet loss.
+
+Consider these key values:
+
+- Payload: Allow for the transport of IPv6 packets, by adhering to the minimum MTU
+  of 1280 Byte specified in RFC 2460
+  - and configure `MSS clamping`_ accordingly,
+  - and announce your link MTU via Router Advertisments and DHCP
+
+  .. MSS clamping: http://www.tldp.org/HOWTO/Adv-Routing-HOWTO/lartc.cookbook.mtu-mss.html
+
+- Encapsulation: Account for the overhead created by the configured mesh protocol
+  encapsulating the payload, which is
+  - up to 32 Byte (14 Byte Ethernet + 18 Byte batadv) for batman-adv compat v15 (v2014.0 and later)
+  - up to 28 Byte (14 Byte Ethernet + 14 Byte batadv) for batman-adv compat v14 (v2011.3.0 until and including v2013.4.0)
+
+- PMTU: What MTU does the path between your gateway and each of its peers support?
+
+For reference, the complete MTU stack looks like this:
+
+.. image:: https://cloud.hamburg.freifunk.net/thumbnail/de738ee3ac/1024/MTU%20within%20a%20batman-adv%20network%20v5.png
+
+Minimum MTU
+-----------
+
+Calculcate the minimum transport MTU by adding the encapsulation overhead to the
+minimum payload MTU required. This is the lowest recommended value, since going
+lower would cause unnecessary fragmentation for clients which respect the announced
+link MTU.
+
+Example: Our network currently uses batman-adv v15, it therefore requires up
+to 32 Bytes of encapsulation overhead on top of the minimal link MTU required for
+transporting IPv6.::
+
+  \        1312              1294          1280                                 0
+   \---------+-----------------+-------------+----------------------------------+
+    \TAP     |    batadv v15   |   Ethernet  |            Payload               |
+     \-------+-----------------+-------------+----------------------------------+
+      \      ^
+             |
+
+          MTU_LOW = 1280 Byte + 14 Byte + 18 Byte = 1312 Byte
+
+Maximum MTU
+-----------
+
+Calculating the maximum transport MTU is interesting, because it increases the
+throughput, by allowing larger payloads to be transported, but also more difficult
+as you have to take into account the tunneling overhead and each peers PMTU, which
+varies between providers.
+The underlying reasons are mostly PPPoE, Tunneling and IPv6 transition technologies
+like DS-Lite.
+
+Example: The peer with the smallest MTU on your network is behind DS-Lite and can
+transport IPv4 packets up to 1436 Bytes in size. Your tunnel uses IPv4 (20 Byte),
+UDP (8 Byte), Fastd (24 byte) and you require TAP (14 Byte) for Layer 2 (Ethernet)
+Tunneling.::
+
+  1436                1416     1408                    1384          1370    \
+    +-------------------+--------+-----------------------+-------------+------\
+    |        IP         |  UDP   |         Fastd         |     TAP     |    bat\
+    +-------------------+--------+-----------------------+-------------+--------\
+                                                                       ^         \
+                                                                       |
+
+       MTU_HIGH = 1436 Byte - 20 Byte - 8 Byte - 24 Byte - 14 Byte = 1370 Byte
+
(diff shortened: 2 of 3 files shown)


jplitza:
What is the point of aiming for unfragmented 1280 byte packets? Yes, that's what IPv6 mandates as minimum MTU. But no client, no device actually uses 1280 byte MTU, because most ignore what they are told in DHCP replies and RAs, what the MTU should be.

mweinelt:
That may be true, but some still do. Even then carefully calculated MSS Clamping reduces payload sizes to match the target MTU.

jplitza:
I'd really like to know which clients adhere to the advertised MTU. And MSS Clamping works for the current value, too, doesn't it?

T-X:
@jplitza: What about ICMPv6 Packet Too Big messages? At least for IPv6 I would expect it to work reliably as IPv6 dumped the intermediate fragmentation support, making Packet Too Big messages essential for any IPv6 communication to work reliable.

mweinelt:
@jplitza I can see this behaviour on my N5X running Stock Android 7.1.2 for example.

Yes, and MSS clamping works solely for TCP.

jplitza:
@T-X So you're suggesting that the gateway sends a Packet Too Big reply for each new connection as soon as the client sends a packet thats above 1280 byte? That sounds like it could work. 

@mweinelt Hm, interesting. Last time I checked, neither Android nor Ubuntu nor Windows seemed to care about it, but I don't remember which versions that were.

And because it seems relevant to the whole MTU and fragmentation discussion: [IP Fragmentation is broken (by Cloudflare)](https://blog.cloudflare.com/ip-fragmentation-is-broken/)

ohrensessel:
Windows 7 and 10 respect the MTU given by RAs afaik, but only for v6. Can
check again if required.

To force a MTU of 1280 you also set this mtu for the uplink interface of
the gateway, e.g the tunnel. Then even non-tcp packets would not exceed
1280 byte.

Am 20.08.2017 14:14 schrieb "Jan-Philipp Litza" <notifications@github.com>:

mweinelt:
Apple iOS 10.3.3 *does not* adhere to the advertised MTU. Happily sends SYN packets with a 1460 MSS, unfortunately there does not seem to be an option to view the MTU on these devices.

mweinelt:
> @jplitza I can see this behaviour on my N5X running Stock Android 7.1.2 for example.

Still valid for Android 8.0 (Oreo), which was released yesterday.

T-X:
From @ohrensessel:
> To force a MTU of 1280 you also set this mtu for the uplink interface of the gateway, e.g the tunnel. Then even non-tcp packets would not exceed 1280 byte.

Just in case anyone is wondering why this works: This is exactly the ICMPv6 Packet Too Big message case :-).

(The kernel gets a "large" packet, sees it does not fit through the 1280 bytes MTU tunnel, drops it and sends an "ICMPv6 Packet Too Big" back to the sender which retransmits with a smaller packet size, generally the one stated in the previous ICMPv6 message. By default, this ICMPv6 PTB behaviour happens automatically on a Linux router.)

Adorfer:
I do not get the point of restarting the complete MTU-fragemtation discussion the n-th time again. 
with all the arguments already well evaluated the last time. 
fragmented packets should be avoided if possible. 
(and if not possible on cer the tain cases, it does not help to divert to OS OS warsit does not make
what about the initial question

mweinelt:
When we last visited that topic there were very few devices, if any, that adhered to the MTU announced through DHCP and Router Advertisments. These device are imposing an additional load on our routers by sending full packets that require fragmentation.

With the OS/device situation changing we can now properly inform those devices so they adjust their outgoing packets accordingly. And Android makes up a big part of our network, probably yours as well.

This is why "OS wars", as you call them, are a necessary argument in this discussion to get the documentation updated.

Adorfer:
@mweinelt 
1) there a sweet spot for the fastd-mtu somewhere between 1312 and 1364 (ymmd), preferrably at the low end, as your PR states (which i fully support)
2) there is a number of devices which do not operate as we consider it as optimal, even if we apply the exhausitive lists of tweaks (dhcp-options, radvd-options, mss/clamping on the router etc.pp), which all may have one or the other downturns. 

**But whatever happens in section 2, i do not see why this would influence section 1.** 

selection of 1280 as fastd-mtu makes things even worse on for "non cooperative devices", but it may make things better if at least 1280 can be send unfragmented. .

(the only argument against selection of a good MTU could be _"punish bad designed clients"_. But this is nothing i would seriously consider.

christf:
@mweinelt while I already like the commit in this PR I do also see some benefit in explaining 1312 a little.

rotanid:
so, 1312 works great (better than 1280) here. first in experimental and since 2 weeks on the stable branch.
we measured around 30% performance increase for every device type, even WR841N

@mweinelt maybe add a short explanation for the suggested mtu value like @christf asked ?

A-Kasper:
Our Tunnel have an MTU of 1374. (Fastd)
radvd and dhcpd advertise an MTU of 1342

no problems. This Setup works for over 6 Month without Problems.

I took all known providers in Bochum and asked People to measure their
maxmtu for ipv6 and ipv4. after that i calculated the overheads and came to
these values.

attention: this maybe different with other isps...



2017-09-12 18:13 GMT+02:00 Andreas Ziegler <notifications@github.com>:

mweinelt:
Rough idea of how to document that chapter as follows. Some terms are still a bit wonky and I want you to give me feedback on things that look wrong. Actually I believe that collaborative editing would yield a better result faster, but let's start with this:

---

Setting the MTU requires careful consideration, as setting it too low will cause excessive fragmentation and setting it too high may leave peers with a broken tunnel due to packet loss.

Consider these key values:
- Payload: Allow for the transport of IPv6 packets, by adhering to the minimum MTU of 1280 Byte [RFC 2460]
  - and configure MSS clamping accordingly,
  - and announce your link MTU via Router Advertisments and DHCP
- Encapsulation: Account for the overhead created by the configured mesh protocol encapsulating the payload, which is
  - up to 32 Byte for batman-adv v15 (v2014.0 and later)
  - up to 28 Byte for batman-adv v14 (v2011.3.0 until and including v2013.4.0)
- Path MTU: What MTU does the path between your gateway and each of its peers allow?

Calculcate the minimum transport MTU by adding the encapsulation overhead to the minimum payload MTU required.

Calculating the maximum transport MTU is more difficult as you have to take the Path MTU into account, which varies alot by geographical region and provider. The underlying reasons are mostly PPPoE, Tunneling and IPv6 transition technologies like DS-Lite.

lemoer:
@mweinelt Maybe something like this :)?

```
\   1312              1294          1280                                 0
 \----+-----------------+-------------+----------------------------------+
  \   |    batadv v15   |   Ethernet  |            Payload               |
   \--+-----------------+-------------+----------------------------------+
    \

MTU_LOW = 1280 Byte + 14 Byte + 18 Byte = 1312 Byte
```

```
1436                 1416     1408                    1384          1370 \
  +-------------------+--------+-----------------------+-------------+----\
  |        IP         |  UDP   |         Fastd         |     TAP     |     \
  +-------------------+--------+-----------------------+-------------+------\
                                                                             \
MTU_HIGH = 1436 Byte - 20 Byte - 8 Byte - 24 Byte - 14 Byte = 1370 Byte
```

mweinelt:
```
1436                 1416     1408                    1384          1370 <-----> 1312              1294          1280                                 0
  +-------------------+--------+-----------------------+-------------+---  ...  ---+-----------------+-------------+----------------------------------+
  |        IP         |  UDP   |         Fastd         |     TAP     |     ...     |    batadv v15   |   Ethernet  |            Payload               |
  +-------------------+--------+-----------------------+-------------+---  ...  ---+-----------------+-------------+----------------------------------+
```

lemoer:
Hm. This looks more as if there are some bytes between TAP and batman layer.

lemoer:
```
\        1312              1294          1280                                 0
 \---------+-----------------+-------------+----------------------------------+
  \TAP     |    batadv v15   |   Ethernet  |            Payload               |
   \-------+-----------------+-------------+----------------------------------+
    \      ^
           |

        MTU_LOW = 1280 Byte + 14 Byte + 18 Byte = 1312 Byte
```

```
1436                1416     1408                    1384          1370    \
  +-------------------+--------+-----------------------+-------------+------\
  |        IP         |  UDP   |         Fastd         |     TAP     |    bat\
  +-------------------+--------+-----------------------+-------------+--------\
                                                                     ^         \
                                                                     |

     MTU_HIGH = 1436 Byte - 20 Byte - 8 Byte - 24 Byte - 14 Byte = 1370 Byte
```

mweinelt:
Yeah, this makes it much clearer.

mweinelt:
PR updated.

PR #1215, opened 2017-08-24
Docs: mesh_vpn default enabled

rubo77:


NeoRaider:
This is simply wrong.

rubo77:
So the default is false?

     enabled = false

Adorfer:
i understand it as _"the default for VPN should be 'disabled'. this is a political design decision. The documentation should reflect that, but not explain."_

PR #1216, opened 2017-08-27
Add domain specific configs to gluon

lemoer:
### intro
As freifunk networks become larger the ground noise in the batman layer 2 network increases. At about ~1000 nodes, we reach a point, where we want to split the network in different **domains** (layer 2 segments). So need to serve different configs. With current implementation in gluon, we need to build one firmware per domain, as there is a possibility for one `site.conf` per image. The idea is to introduce multiple domain configs per image, which could be changed without reflashing the router.

### about this pr
The gluon framework ensures, that the `site.conf` is valid before the image is build (at least in some limits). This is done per module (`package/gluon-pkg-name/check_site.lua`) with help of the `check_site_lib.lua`. This PR tries to extend the functionalities of the *check_site_lib* to verify domain specific configs also. This PR does not implement a functionality to read the domain values inside the firmware yet.

### ideas
- the domain config should be in `site/domains/domainname.conf`
- the structure is the same as in the `site/site.conf`
- the two configs (site.conf and selected domain.conf) should be merged.
- not all config values make sense in a domain specific way, so we *forbid* some values in the domain specific configs. the other way is also a subject

### state
- [x] domain specific configs are baked into the image (`/lib/gluon/domains/domainname.json`)
- [x] the functionality to forbid values has been implemented (e.g. `forbid_in_domain('config_mode.geo_location.show_altitude')`)
    - values, which do not make sense in a domain specific way are forbidden
    - some other values, which I think are not in first priority, are also forbidden for now
- [x] if there is an incorrect value while checking the site and domain configs, we print in which config the issue is
- [x] the `need_table(...)` function should be handled correctly
- [x] decide what should happen with packages, that do not know of domain specific values
    - ~~fail at compiletime, since they could use values, which should be forbidden in domain specific config?~~
    - don't care?
- [x] what happens if `site/domains/` directory is not existing?
- [x] switch from `forbid_in_*` to `need_*`, `need_domain_*`, `need_site_*`
- [x] mesh, ap, ... should be only allowed in domain config
- [x] site.conf does not need to be valid at all. -> having at least one `domain.conf` becomes necessary.
    - remove `if domain ~= nil` checks
    - stop verifying `check_domain(site, nil)`
- [x] test what happens, if no domain.conf exists
- [x] test one need_site_* value
- [x] test one need_domain_* value
- [x] test one need_* value
- [x] decide, how packages should ask for values of domain.conf?
    - generic approach, which is implemented in [site.lua](https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-core/luasrc/usr/lib/lua/gluon/site.lua), which is transparent for the packages
    - ~~implement the query in each package `if domain.opkg or if site.opkg ...`~~
- [x] upgrade behavior?
- [x] domain in uci value gluon.\@system[0].domain
- [x] implement mergeing in lua lib `site`
- [x] test merging in lua lib
- [x] merge domain and site in libgluonutil
- [x] add domain_seed
- [x] use domain_seed for mesh_lan
- [x] ensure, that the default domain exists
- [x] announce domain via respondd
- [x] [BUG] if a domain (only) value is not given, the error says it's missing in *site.conf*
- [x] [IMPROVEMENT] if a domain and site value is not given, the error says it's missing in *site.conf*

### DIFF
--- a/package/gluon-core/files/etc/config/gluon
+++ b/package/gluon-core/files/etc/config/gluon
@@ -1,0 +1,1 @@
+config system system
--- a/package/gluon-core/files/lib/gluon/domain_changed.sh
+++ b/package/gluon-core/files/lib/gluon/domain_changed.sh
@@ -1,0 +1,10 @@
+#!/bin/sh
+
+domain_code=$(uci get gluon.system.domain_code)
+
+[ -f /lib/gluon/domains/${domain_code}.json ] || (echo "file not found: /lib/gluon/domains/${domain_code}.json" >&2; exit 1) || exit 1
+
+for s in /lib/gluon/upgrade/*; do
+	echo -n ${s}:
+	(${s} && echo " ok") || echo " error"
+done
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/site.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/site.lua
@@ -29,7 +29,11 @@ local domain = (function(site)
 		dc = site['default_domain_code']
 	end
 
-	return read_json('/lib/gluon/domains/'..dc..'.json')
+	local domain = read_json('/lib/gluon/domains/'..dc..'.json')
+	if domain['domain_aliases'] and domain['domain_aliases'][dc] then
+		domain['domain_name'] = domain['domain_aliases'][dc]
+	end
+	return domain
 end)(site)
 
 
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/site.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/site.lua
@@ -1,11 +1,10 @@
-local site = (function()
-	local config = '/lib/gluon/site.json'
+local function read_json(path)
 
 	local json = require 'luci.jsonc'
 	local decoder = json.new()
 	local sink = decoder:sink()
 
-	local file = assert(io.open(config))
+	local file = assert(io.open(path))
 
 	while true do
 		local chunk = file:read(2048)
@@ -16,11 +15,36 @@ local site = (function()
 	file:close()
 
 	return assert(decoder:get())
-end)()
+end
+
+local site = read_json('/lib/gluon/site.json')
+local domain = (function(site)
+	local uci = require('simple-uci').cursor()
+	local fs = require "nixio.fs"
+	local sname = uci:get_first('gluon', 'system')
+
+	local dc = uci:get('gluon', sname, 'domain_code') or ''
+
+	if fs.stat('/lib/gluon/domains/'..dc..'.json', 'type')~='reg' then
+		dc = site['default_domain_code']
+	end
+
+	return read_json('/lib/gluon/domains/'..dc..'.json')
+end)(site)
 
 
 local wrap
 
+local function merge(t1, t2)
+	for k, v in pairs(t2) do
+		if (type(v) == "table") and (type(t1[k] or false) == "table") then
+			merge(t1[k], t2[k])
+		else
+			t1[k] = v
+		end
+	end
+	return t1
+end
 
 local function index(t, k)
 	local v = getmetatable(t).value
@@ -58,4 +82,4 @@ end
 
 module 'gluon.site'
 
-return wrap(site, _M)
+return wrap(merge(site, domain), _M)
--- a/package/gluon-site/Makefile
+++ b/package/gluon-site/Makefile
@@ -47,6 +47,8 @@ endef
 define Build/Compile
 	GLUON_SITEDIR='$(call qstrip,$(CONFIG_GLUON_SITEDIR))' lua -e 'print(require("cjson").encode(assert(dofile("../../scripts/site_config.lua"))))' > $(PKG_BUILD_DIR)/site.json
 	ls $(call qstrip,$(CONFIG_GLUON_SITEDIR))/domains/*.conf > /dev/null # at least one domain cfg has to exist
+	GLUON_SITEDIR='$(call qstrip,$(CONFIG_GLUON_SITEDIR))' lua -e 'print(assert(dofile("../../scripts/site_config.lua")).default_domain_code)' > $(PKG_BUILD_DIR)/default_domain_code
+	ls '$(call qstrip,$(CONFIG_GLUON_SITEDIR))'/domains/$$$$(cat $(PKG_BUILD_DIR)/default_domain_code).conf # ensure default_domain_code exists
 	for domain_cfg in `find $(call qstrip,$(CONFIG_GLUON_SITEDIR))/domains/ -iname \*.conf -printf "%f\n"`; do dc=$$$${domain_cfg%.conf}; GLUON_SITEDIR='$(call qstrip,$(CONFIG_GLUON_SITEDIR))' lua -e 'print(require("cjson").encode(assert(dofile("../../scripts/domain_config.lua")("'$$$${dc}'"))))' > $(PKG_BUILD_DIR)/domains/$$$${dc}.json; done
 	$(call GluonBuildI18N,gluon-site,$(GLUON_SITEDIR)/i18n)
 endef
--- a/package/gluon-site/Makefile
+++ b/package/gluon-site/Makefile
@@ -49,7 +49,15 @@ define Build/Compile
 	ls $(call qstrip,$(CONFIG_GLUON_SITEDIR))/domains/*.conf > /dev/null # at least one domain cfg has to exist
 	GLUON_SITEDIR='$(call qstrip,$(CONFIG_GLUON_SITEDIR))' lua -e 'print(assert(dofile("../../scripts/site_config.lua")).default_domain_code)' > $(PKG_BUILD_DIR)/default_domain_code
 	ls '$(call qstrip,$(CONFIG_GLUON_SITEDIR))'/domains/$$$$(cat $(PKG_BUILD_DIR)/default_domain_code).conf # ensure default_domain_code exists
-	for domain_cfg in `find $(call qstrip,$(CONFIG_GLUON_SITEDIR))/domains/ -iname \*.conf -printf "%f\n"`; do dc=$$$${domain_cfg%.conf}; GLUON_SITEDIR='$(call qstrip,$(CONFIG_GLUON_SITEDIR))' lua -e 'print(require("cjson").encode(assert(dofile("../../scripts/domain_config.lua")("'$$$${dc}'"))))' > $(PKG_BUILD_DIR)/domains/$$$${dc}.json; done
+	rm -f $(PKG_BUILD_DIR)/domains/*.json
+	for domain_cfg in `find $(call qstrip,$(CONFIG_GLUON_SITEDIR))/domains/ -iname \*.conf -printf "%f\n"`; do \
+		dc=$$$${domain_cfg%.conf}; \
+		GLUON_SITEDIR='$(call qstrip,$(CONFIG_GLUON_SITEDIR))' lua -e 'print(require("cjson").encode(assert(dofile("../../scripts/domain_config.lua")("'$$$${dc}'"))))' > $(PKG_BUILD_DIR)/domains/$$$${dc}.json; \
+		aliases=$$$$(GLUON_SITEDIR='$(call qstrip,$(CONFIG_GLUON_SITEDIR))' lua -e 'for alias_name, _  in pairs(dofile("../../scripts/domain_config.lua")("'$$$${dc}'")["domain_aliases"] or {}) do print(alias_name.." ") end'); \
+		for alias in $$$${aliases}; do \
+			ln -s $$$${dc}.json $(PKG_BUILD_DIR)/domains/$$$${alias}.json; \
+		done; \
+	done
 	$(call GluonBuildI18N,gluon-site,$(GLUON_SITEDIR)/i18n)
 endef
 
--- a/package/gluon-site/Makefile
+++ b/package/gluon-site/Makefile
@@ -8,7 +8,7 @@ PKG_VERSION:=$(if $(DUMP),x,$(GLUON_SITE_VERSION))
 
 
 PKG_CONFIG_DEPENDS := CONFIG_GLUON_RELEASE CONFIG_GLUON_SITEDIR
-PKG_FILE_DEPENDS := $(GLUON_SITEDIR)/site.conf $(GLUON_SITEDIR)/i18n/
+PKG_FILE_DEPENDS := $(GLUON_SITEDIR)/site.conf $(GLUON_SITEDIR)/domains/ $(GLUON_SITEDIR)/i18n/
 PKG_BUILD_DEPENDS := lua-cjson/host
 
 PKG_BUILD_DIR := $(BUILD_DIR)/$(PKG_NAME)
@@ -38,6 +38,7 @@ endef
 
 define Build/Prepare
 	mkdir -p $(PKG_BUILD_DIR)
+	mkdir -p $(PKG_BUILD_DIR)/domains
 endef
 
 define Build/Configure
@@ -45,12 +46,16 @@ endef
 
 define Build/Compile
 	GLUON_SITEDIR='$(call qstrip,$(CONFIG_GLUON_SITEDIR))' lua -e 'print(require("cjson").encode(assert(dofile("../../scripts/site_config.lua"))))' > $(PKG_BUILD_DIR)/site.json
+	ls $(call qstrip,$(CONFIG_GLUON_SITEDIR))/domains/*.conf > /dev/null # at least one domain cfg has to exist
+	for domain_cfg in `find $(call qstrip,$(CONFIG_GLUON_SITEDIR))/domains/ -iname \*.conf -printf "%f\n"`; do dc=$$$${domain_cfg%.conf}; GLUON_SITEDIR='$(call qstrip,$(CONFIG_GLUON_SITEDIR))' lua -e 'print(require("cjson").encode(assert(dofile("../../scripts/domain_config.lua")("'$$$${dc}'"))))' > $(PKG_BUILD_DIR)/domains/$$$${dc}.json; done
 	$(call GluonBuildI18N,gluon-site,$(GLUON_SITEDIR)/i18n)
 endef
 
 define Package/gluon-site/install
 	$(INSTALL_DIR) $(1)/lib/gluon
 	$(INSTALL_DATA) $(PKG_BUILD_DIR)/site.json $(1)/lib/gluon/
+	$(INSTALL_DIR) $(1)/lib/gluon/domains
+	$(CP) $(PKG_BUILD_DIR)/domains/*.json $(1)/lib/gluon/domains/
 	echo '$(GLUON_SITE_VERSION)' > $(1)/lib/gluon/site-version
 	echo '$(call qstrip,$(CONFIG_GLUON_RELEASE))' > $(1)/lib/gluon/release
 
--- a/package/gluon.mk
+++ b/package/gluon.mk
@@ -11,7 +11,24 @@ local site_json = f:read('*a')
 f:close()
 
 site = require('cjson').decode(site_json)
-$(shell cat '$(TOPDIR)/../scripts/check_site_lib.lua' '$(1)' | sed -ne '1h; 1!H; $$ {g; s/@/+@/g; s/\n/-@/g; p}')
+
+function check_domain(domain_code, domain)
+	$(shell cat '$(TOPDIR)/../scripts/check_site_lib.lua' '$(1)' | sed -ne '1h; 1!H; $$ {g; s/@/+@/g; s/\n/-@/g; p}')
+end
+
+local dir = os.getenv('IPKG_INSTROOT') .. '/lib/gluon/domains/'
+local pfile = io.popen('find '..dir..' -iname \\*.json')
+for domain_path in pfile:lines() do
+	local domain_code = string.gmatch(domain_path, '([^/]+).json$$')()
+	local f = assert(io.open(domain_path))
+	local domain_json = f:read('*a')
+	f:close()
+
+	domain = require('cjson').decode(domain_json)
+
+	check_domain(domain_code, domain)
+end
+pfile:close()
 END__GLUON__CHECK__SITE
 endef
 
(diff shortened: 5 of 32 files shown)


lemoer:
I think, i need to change the api of `need_table(varname, subcheck, required)` a little. The `subcheck` parameter is a callback function with signature `subcheck(key, value)` for now.

The following example is not easy to solve with this api in my oppinion:
``` lua
site = {
  arr_value = {
    somekey = {
      special_number = 42
    }
  }
}

domain = {
  arr_value = {
    somekey = {
      another_setting = 'magic'
    }
  }
}
```

We could now do the following to call the `subcheck` inside `need_table('arr_value', subcheck)`:
``` lua
subcheck('somekey', { special_number=42, another_setting='magic'})
```
In other words, we merge the keys. But then inside `subcheck(...)` it would be impossible to know where the value was from (domain or site). We could not drop a nice error message this way.

My idea would be to drop the pass of `value` to the `subcheck(...)` function.
All current modules in gluon do not use the `value` anyway. They instead reach the value by using the "json path" with the `key`. E.g:
``` lua
local prefix = string.format('autoupdater.branches[%q].', key)
need_string(prefix .. 'name')
```
This should work for all purposes and already covers proper error messages.

The only thing would be:
- change api from `subcheck(key, value)` to `subcheck(key)`
- merge keys from domain and site before calling `subcheck(key)`

lemoer:
Notes:
[ ] switch from `forbid_in_*` to `need_*`, `need_domain_*`, `need_site_*`
[ ] mesh, ap, ... should be only allowed in domain config

lemoer:
## allowed in both site and domain config
- autoupdater.branches
- autoupdater.branches.*.mirrors
- wifi*
- wifi*.channel
- wifi*.basic_rate
- wifi*.ap
- wifi*.ap.ssid
- wifi*.ap.disabled
- wifi*.ibss
- wifi*.ibss.mcast_rate
- wifi*.ibss.vlan
- wifi*.ibss.disabled
- wifi*.mesh
- wifi*.mcast_rate
- wifi*.disabled
- opkg
- opkg.lede
- opkg.extra
- opkg.extra.*
- ntp_servers
- dns
- dns.cacheentries
- dns.servers
- nextnode
- mesh
- mesh.batman_adv
- mesh.batman_adv.gw_sel
- mesh.batman_adv.routing_algo
- mesh_vpn.enabled
- mesh_vpn.mtu
- mesh_vpn.bandwidth_limit
- mesh_vpn.bandwidth_limit.enabled
- mesh_vpn.bandwidth_limit.ingress
- mesh_vpn.bandwidth_limit.egress
- mesh_vpn.fastd.methods
- mesh_vpn.fastd.groups
- mesh_vpn.fastd.groups.*.limit
- mesh_vpn.fastd.groups.*.groups
- mesh_vpn.fastd.groups.*.peers
- mesh_vpn.fastd.groups.*.peers.key
- mesh_vpn.fastd.groups.*.peers.remotes
- mesh_vpn.tunneldigger.brokers



## allowed in site only
- authorized_keys
- autoupdater.branch
- autoupdater.branches.*.name
- autoupdater.branches.*.good_signatures
- autoupdater.branches.*.pubkeys
- config_mode
- config_mode.owner
- config_mode.owner.obligatory
- config_mode.remote_login
- config_mode.geo_location
- config_mode.geo_location.show_altitude
- config_mode.remote_login.show_password_form
- config_mode.remote_login.min_password_length
- site_code
- site_name
- site_seed
- hostname_prefix
- timezone
- regdom
- wifi*.ibss.supported_basic_rates
- poe_passthrough
- wifi*.mesh.supported_basic_rates
- mesh_on_wan
- mesh_on_lan
- single_as_lan
- mesh_vpn.fastd.configurable
- mesh_vpn.fastd.syslog_level
- roles.default
- roles.list
- setup_mode.skip



## allowed in domain only
- next_node.mac
- next_node.ip4
- next_node.ip6
- prefix4
- prefix6
- wifi*.ibss.ssid
- wifi*.ibss.bssid
- wifi*.mesh.id
- extra_prefixes6

## helpful commands:
``` shell
grep -r need_ package/*/check_site.lua | cut -d: -f 2 | sed 's/need_/\nneed_/g' | grep need_ | grep need_site
grep -r need_ package/*/check_site.lua | cut -d: -f 2 | sed 's/need_/\nneed_/g' | grep need_ | grep need_domain
grep -r need_ package/*/check_site.lua | cut -d: -f 2 | sed 's/need_/\nneed_/g' | grep need_ | grep -v need_site | grep -v need_domain
```

lemoer:
```
/home/gluon/gluon/lede/staging_dir/hostpkg/bin/lua: stdin:50: domains/default.conf error: `site_code` is not allowed in domain specific config.
stack traceback:
        [C]: in function 'assert'
        stdin:50: in function 'with'
        stdin:186: in function 'need_site_string'
        stdin:209: in function 'check_domain'
        stdin:296: in main chunk
        [C]: in function 'dofile'
        (command line):1: in main chunk
        [C]: ?
postinst script ./usr/lib/opkg/info/gluon-core.postinst has failed with exit code 1
make[3]: *** [package/Makefile:67: package/install] Error 1
make[3]: Leaving directory '/home/gluon/gluon/lede'
make[2]: *** [package/Makefile:102: /home/gluon/gluon/lede/staging_dir/target-mips_24kc_musl-1.1.16/stamp/.package_install] Error 2
make[2]: Leaving directory '/home/gluon/gluon/lede'
make[1]: *** [/home/gluon/gluon/lede/include/toplevel.mk:200: world] Error 2
make[1]: Leaving directory '/home/gluon/gluon/lede'
make: *** [Makefile:133: all] Error 2
```
Yeay :)

lemoer:
@NeoRaider I think, I implemented your improvements. What do you say for now?

Next steps would be:
- [ ] decide, how packages should ask for values of domain.conf?
    - generic approach, which is implemented in [site.lua](https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-core/luasrc/usr/lib/lua/gluon/site.lua), which is transparent for the packages
    - implement the query in each package `if domain.opkg or if site.opkg ...`
- [ ] upgrade behavior?
- [ ] domain as uci value?
- [ ] integrate domain dependent config values in /lib/gluon/upgrade/*?

skorpy2009:
Maybe the "hostname_prefix" should be in both, site and domain config, because we will start with the zip code and separate by the first two digits. 60xxx and 61xxx will be different meshes.

lemoer:
@skorpy2009 Hm. I think this will introduce another level of complexity. I think the hostname_prefix is only used during config mode and maybe the domain will be set in config mode (or maybe first in the post upgrade process, aka the scripts in /lib/gluon/upgrade). So the changed hostname_prefix only read, during the second time in config mode. Normally there is already a hostname set yet, so the hostname_prefix won't be used neither.

Besides: I don't get the advantages, why one wants to store such information in the node name? Wouldn't be storing the domain in another uci value be enough? Maybe add a postal uci value, which is mandatory?

rotanid:
i agree with @lemoer , bad idea to try to store additional data in the hostname

lemoer:
I can first go on with my work, if I know, when the domain config should be changeable. We need a decision here!

1. Changing should be possible when the router is online (already in the first version of domain configs).
2. Changing should be only possible during post-upgrade procedure (forever)
3. First we implement the second choice and merge this behavior into gluon. If somebody (maybe the hoodselector) wants to extend the behavior, so the domains can be changed online, he is free to add another PR.

@NeoRaider I think this is your decision. I would be happy if you could state your position, so I could go on.

NeoRaider:
I think 2. is fine for now. Allowing online config changes doesn't really have anthing to do with separating the config into site and domain sections, so I'd like to postpone that to a later PR.

lemoer:
- [ ] path is broken /lib/gluon/default/default

lemoer:
[ ] libgluonutil?

lemoer:
- [ ] mesh lan out of domain_seed?

lemoer:
- [ ] look for uci_alloc_context(); in code and ensure free()

lemoer:
- [ ] rename site.json to base.json?

lemoer:
- [ ] libuci dependency in gluonutil

lemoer:
The binary for *gluon-print-site* is about 4K when the debug infos are stripped.

lemoer:
@NeoRaider I think, this is ready to review.

lemoer:
Some more work to allow aliases:
- [x] symlink
- [x] add check_site for domain_name and domain_aliases
- [x] domain_name should be adapted?
- [ ] exits in makefile
- [ ] rename aliases to domain_aliases

package/gluon.mk
```diff
@@ -11,7 +11,24 @@ local site_json = f:read('*a')
...
 
 site = require('cjson').decode(site_json)
-$(shell cat '$(TOPDIR)/../scripts/check_site_lib.lua' '$(1)' | sed -ne '1h; 1!H; $$ {g; s/@/+@/g; s/\n/-@/g; p}')
+
+function check_domain(domain_name, domain)
+	$(shell cat '$(TOPDIR)/../scripts/check_site_lib.lua' '$(1)' | sed -ne '1h; 1!H; $$ {g; s/@/+@/g; s/\n/-@/g; p}')
+end
+
+local dir = os.getenv('IPKG_INSTROOT') .. '/lib/gluon/domains/'
+local pfile = io.popen('find '..dir..' -iname \\*.json') -- TODO: not sure about the doublebackslash? why do we need it?
```
rotanid: leaving a comment as a marker for the TODO

rotanid:
please check your code indention again. the current lua code seems to use 3 spaces while yours sometimes has 2 or even 5 spaces

mweinelt:
Honestly swapping between site and domain is confusing at times.

Can we call the shared configuration `base`, and the invidual configurations either `site` or `domain`? Would probably increase readability and prevent confusion alot.

package/libgluonutil/src/libgluonutil.c
```diff
@@ -31,7 +31,41 @@
...
+ * Merges two JSON objects
+ *
+ * On conflicts, object a will be preferred.
+ *
+ * Internally, this functions merges all entries from object a into object b,
+ * so merging a small object a with a big object b is faster than vice-versa.
+ */
+static struct json_object * merge_json(struct json_object *a, struct json_object *b) {
+	if (!json_object_is_type(a, json_type_object) || !json_object_is_type(b, json_type_object)) {
+		json_object_put(b);
```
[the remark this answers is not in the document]
lemoer: I'm not sure, how this should be possible outside the function? `merge_json()` is called recursively.

package/gluon-core/files/lib/gluon/domain_changed.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+domain=$(uci get gluon.@system[0].domain)
```
mweinelt: The variable is not populated by default, also `gluon` would be an entirely new section, while `gluon-core` already exists.

mweinelt:
tests:
-  [x] illegal domain values are successfully migrated to the default domain

package/gluon-core/files/lib/gluon/domain_changed.sh
lemoer: Yes, neoraider said, that he want to merge all the `gluon-*` sections in one `gluon` section in near future. And I already use this new section. But I'm currently not sure how to introduce sections. I'll add a todo for this. 

But you pointed me to an other error. It should be `uci get gluon.system.domain`. Thanks.

mweinelt:
```
stdin:84: site.conf error: `next_node.mac` is not allowed in site config.
```

Do we actually require unique next_node mac addresses per domain? If not I'd want to preset it in `site.conf`.

NeoRaider:
`next_node.mac` should be disallowed in domain config, there is no reason to ever change this address.

NeoRaider:
Disregard my previous comment, I understood the reationale just a moment ago:

Disallowing it in site makes sense. New domains should just not set an address explicitly at all, so the Gluon default is used. It only needs to be set for old domains that use a different address.

package/gluon-core/src/respondd.c
```diff
@@ -0,0 +1,57 @@
...
+
+	struct json_object *domain_name = NULL;
+	json_object_object_get_ex(site, "domain_name", &domain_name);
+	if (domain_name)
+		json_object_get(domain_name);
+
+	json_object_put(site);
+
+	struct json_object *ret = json_object_new_object();
+	json_object_object_add(ret, "domain_name", domain_name);
```
genofire: Is there a reason, why you start use the `gluon-core` for a `respondd.c`.
Should not that be in the package `gluon-respondd` ??

mweinelt: FYI: As discussed earlier on IRC `domain_name` will be dropped and `system.domain_code` will be added.

package/gluon-core/files/lib/gluon/domain_changed.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-domain=$(uci get gluon.@system[0].domain)
+domain=$(uci get gluon.system.domain)
```
mweinelt: up until this change `/etc/config/gluon` looked like:

```
config system
    option 'domain' 'default'
```

What should it look like now?

mweinelt: The answer is

```
config system system
    option 'domain' 'default'
```

lemoer: yes. thanks

lemoer:
- [ ] maybe move respondd stuff to gluon-respondd package?
- [ ] what should happen with next_node.mac?

mweinelt:
- [x] Domain appears on status page
- [x] `system.domain_code` is exposed through nodeinfo

package/gluon-core/src/respondd.c
lemoer: >Is there a reason, why you start use the gluon-core for a respondd.c.
> Should not that be in the package gluon-respondd ??
Done

package/gluon-core/files/etc/config/gluon
```diff
@@ -0,0 +1 @@
+config system system
```
kb-light: why not using `/etc/config/gluon-core` which is already supplied by gluon-core?

package/gluon-core/luasrc/usr/lib/lua/gluon/site.lua
```diff
@@ -16,11 +15,40 @@ local site = (function()
...
+	local sname = uci:get_first('gluon', 'system')
+
+	local dn = uci:get('gluon', sname, 'domain') or ''
+
+	if fs.stat('/lib/gluon/domains/'..dn..'.json', 'type')~='reg' then
+		dn = site['default_domain']
+	end
+
+	local domain = read_json('/lib/gluon/domains/'..dn..'.json')
+	if domain['domain_aliases'] and domain['domain_aliases'][dn] then
```
kb-light: This currently makes not much sense to me. The file is already `..dn..'.json'`, why can't you just set the `domain_name` there?

kb-light:
`domain` (`default_domain`) vs. `domain_name` vs. `domain_code` is very confusing at the moment. According to `site_code` and `site_name`, I would expect spaces in a domain name, so I would prefer using `domain_code .. '.conf'` and therefore `default_domain_code` and `gluon.system.domain_code`.

It makes sense, that the SSID and the mesh-id are forced to domain configuration, but it than does not make much sense to allow `mesh_vpn.fastd.groups.*.peers.remotes` in site.conf. I would suggest to force the remotes to domain specific configuration, too, because it also belongs to "which nodes are allowed to mesh with each other and which are not".

package/gluon-core/luasrc/usr/lib/lua/gluon/site.lua
lemoer: Aliases for domains are created with symlinks. So in the symliked case `domain_name` is incorrect in the json file and has to be replaced with `domain['domain_aliases'][dn]`.

kb-light:
I flashed the current version without problems to one of my nodes after setting up `gluon.system.domain_code`.

mweinelt:
We have been testing the latest version and it looks good and works together with #1250.

lemoer:
As @neoraider told me yesterday he'd like to see a small change in the check_site_lib api (only cosmetic changes):

``` lua
need_string('foo.bar') -> need_string('foo.bar') -- (no change in this case)
need_domain_string('foo.bar') -> need_string(in_domain('foo.bar'))
need_site_string('foo.bar') -> need_string(in_site('foo.bar'))
```

package/gluon-core/files/etc/config/gluon
NeoRaider: I think calling the section 'core' would make sense, as this config is handled by *gluon-core* (and generally store config for package `gluon-X` in section `gluon.X`).

The commit title is not accurate, the top level of the UCI structure is called "package", so this would add a package "gluon" and a section "gluon.system" (or "gluon.core") to UCI.

lemoer:
@NeoRaider PR is updated and rebased to master.

lemoer:
Found a small bug:
```
ln: failed to create symbolic link '/home/gluon/gluon/lede/build_dir/target-i386_pentium4_musl-1.1.16/gluon-site/domains/suedstadt.json': File exists
ln: failed to create symbolic link '/home/gluon/gluon/lede/build_dir/target-i386_pentium4_musl-1.1.16/gluon-site/domains/nordstadt.json': File exists
```

package/gluon-site/Makefile
```diff
@@ -38,19 +38,33 @@ endef
...
+	ls $(call qstrip,$(CONFIG_GLUON_SITEDIR))/domains/*.conf > /dev/null # at least one domain cfg has to exist
+	GLUON_SITEDIR='$(call qstrip,$(CONFIG_GLUON_SITEDIR))' lua -e 'print(assert(dofile("../../scripts/site_config.lua")).default_domain_code)' > $(PKG_BUILD_DIR)/default_domain_code
+	ls '$(call qstrip,$(CONFIG_GLUON_SITEDIR))'/domains/$$$$(cat $(PKG_BUILD_DIR)/default_domain_code).conf # ensure default_domain_code exists
+	for domain_cfg in `find $(call qstrip,$(CONFIG_GLUON_SITEDIR))/domains/ -iname \*.conf -printf "%f\n"`; do \
+		dc=$$$${domain_cfg%.conf}; \
+		GLUON_SITEDIR='$(call qstrip,$(CONFIG_GLUON_SITEDIR))' lua -e 'print(require("cjson").encode(assert(dofile("../../scripts/domain_config.lua")("'$$$${dc}'"))))' > $(PKG_BUILD_DIR)/domains/$$$${dc}.json; \
+		aliases=$$$$(GLUON_SITEDIR='$(call qstrip,$(CONFIG_GLUON_SITEDIR))' lua -e 'for alias_name, _  in pairs(dofile("../../scripts/domain_config.lua")("'$$$${dc}'")["domain_aliases"] or {}) do print(alias_name.." ") end'); \
+		for alias in $$$${aliases}; do \
+			ln -s $$$${dc}.json $(PKG_BUILD_DIR)/domains/$$$${alias}.json; \
```
mweinelt: replace `ln -s` with `ln -sf` here.

lemoer: I decided to use `rm -r` to clean the whole directory `domains/` before entering the for loop, so the whole process should be a little more idempotent. Deleting a domain in site repo hat no effect before this additionional rm.

lemoer:
Pushed a fix.

rotanid:
@lemoer a small issue came up in #1310 , maybe you can fix this

lemoer:
Why should we continue executing upgrade scripts if one fails? I think this increases the chance to overlook the errors. E.g. the openWRT build scripts also fail if there is an error instead of trying to continue.

NeoRaider:
* Third-party packages should not be able to easily break the whole system
* Upgrade scripts can better be compared with /etc/uci-defaults scripts, which are also executed indepenently of each other
* If there is an issue breaking the scripts, configuration must still continue so the autoupdater is likely to work, so the issue can be solved in the next release

lemoer:
As domain_changed.sh is only to be called directly by the user, when they manually changed the `gluon.system.domain` uci value, I think there is no need for continuing. Do you still stay in your position?

NeoRaider:
I think it makes sense to unify the logic handling initial configuration and changing domains (eventually), so this logic should not behave differently.

lemoer:
Do you only mean it should behave equally or do you also mean it should be the same piece of code executing the scripts?

NeoRaider:
In the long term, it should be the same piece of code, but this is not a requirement for getting this PR merged.

package/gluon-core/files/etc/config/gluon
NeoRaider: I plan to pick up this commit first (with my suggested changes), so the rest of Gluon can start using it. This PR will need to be adjusted to the changed section name.

lemoer:
Updated.

NeoRaider:
I've just pushed 0b80f1b5ce816696b3f9b2bdaf9f5d8fe9ac0469, which allows us to avoid some code duplication.

NeoRaider:
I have applied parts of this PR that also work without actual multi-domain support. In 50812b162ca0bb6296db05838335e0e0cf2d0f79, I have made the following changes:

* gluon-client-bridge: allow AP SSID in domain config only (clients could roam between domains otherwise)
* gluon-mesh-vpn-fastd: allow peer keys in domain config only (reusing the same keys for multiple fastd instances breaks certain security guarantees (an attacker with control over the packet flow could make nodes connect to the wrong instance)
* gluon-mesh-vpn-core: allow bandwidth limit defaults in site config only (these default values are applied only in the initial configuration, so domain defaults would never have an effect)

NeoRaider:
I've pulled the rest of the changes into a local branch and will take it from here.

The most important piece missing now is an updated site documentation.

rotanid:
@NeoRaider isnt #1261 the documentation update? if that is not complete, please comment over there

lemoer:
@NeoRaider Thanks for reworking. I like the idea to merge stub no-ops as a prepare step.

(I think we can leave this PR/issue open, until the feature is merged into the master branch)

NeoRaider:
Multidomain support is merged now. It is disabled by default and can be enabled by setting GLUON_MULTIDOMAIN=1 in site.mk.

PR #1217, opened 2017-08-27
ramips-mt7620: add new subtarget with some Gl.Inet devices

ambassador86:
the cheap devices are shipped with openwrt. wpad-mini is replaced with hostapd-mini. the wifi does not support ibss + ap but 802.11s + ap works fine.
test installation: http://meshviewer.chemnitz.freifunk.net/#!v:m;n:e4956e41902c

### DIFF
--- a/targets/ramips-mt7620
+++ b/targets/ramips-mt7620
@@ -1,0 +1,11 @@
+packages -wpad-mini hostapd-mini
+
+# GL Innovations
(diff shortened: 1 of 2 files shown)


NeoRaider:
Merged with some adjustments to make it match other targets (*hostapd-mini* is added by Gluon by default and doesn't need to be listed explicitly).

PR #1224, opened 2017-09-15
treewide: check for NULL after uci_alloc_context()

lemoer:


NeoRaider:
The patch contains many instances of
```diff
-	uci_free_context(ctx);
+	if (ctx)
+		uci_free_context(ctx);
```
I'd prefer to avoid this by using a simple `return` or `return NULL` instead of `goto` (in `get_fastd()`, a `goto` is still needed, but the check can be removed by adding a new goto label).

lemoer:
@NeoRaider updated.

PR #1226, opened 2017-09-18
[RFC] Add pkg gluon-hoodselector from google summer of code 2016

2tata:
**This MR is now just for tracking!**

This patch contains the hoodselector and a domain.conf as an example.

The hoodselector is a software that creates decentralized, semi automated ISO OSI layer 2 network segmentation for batman-adv layer 2 routing networks. This program reads the geobased sub-networks called hoods(gluon domains) from the above mentioned domain configurations. The decision of choosing the right hood is made on following points: first, the hoodselector checks, if the router has a VPN connection. If it has, then its checks, if a geoposition was set on the router. Knowing the position of the router the hoodselector can find the right hood, because each hood is defined with geocoordinates. If the Router doesn't have a VPN connection e.g. as a mesh only router, the hoodselector triggers a WIFI scan and searches for neighboured mesh routers in other hoods. If there is an other router with a different BSSID/mesh ID but with the same mesh SSID, the router chooses it’s hood based on the neighboured BSSID. Ah deeper explanaton can be found here:
[Monitoring and quality assurance of open wifi networks out of client view](https://blog.freifunk.net/2016/05/23/monitoring-and-quality-assurance-open-wifi-networks-out-client-view/)

[Monitoring and quality assurance of open wifi networks out of client view (midterm evaluation)](https://blog.freifunk.net/2016/06/19/monitoring-and-quality-assurance-open-wifi-networks-out-client-view-midterm-evaluation/)

[Monitoring and quality assurance of open wifi networks out of client view (final evaluation)](https://blog.freifunk.net/2016/08/22/monitoring-and-quality-assurance-open-wifi-networks-out-client-view-final-evaluation/)

In the Nordwest Freifunk Community we already had deployed the hood networks since august 2016
Our current active hoods can be seen on the [Lutscher Lupe](https://srv11.ffnw.de/#/%7B%22overlays%22:%5B%22Nodes%22,%22Links%22,%22Clients%22,%22Hoods%22%5D,%22pos_lat%22:52.94201777829491,%22pos_lng%22:8.827514648437502,%22zoom%22:8%7D)

**This MR reference to #789 and replace #997** 

desired state diagram:
hoodselector

established state diagram:
gluon-hoodselector

- [x] '''check_site.lua''' validate default domain to do not have shapes.

- [x] '''check_site.lua''' validate non default domain to have rectangular shapes.

- [x] '''check_site.lua''' validate non default domain to have polygon shapes.

- [x] migrate VPN MODE to use site/domain configurations.

- [ ] migrate GATEWAY MODE to use site/domain configurations.

- [x] migrate SCAN MODE for known hoods to use site/domain configurations.

- [ ] SCAN MODE: discuss how to handle unknown hoods?
      - try getting hood informations via respondd over unknown mehs ID?

- [ ] migrate RADIO LESS MODE  to use site/domain configurations.

- [ ] migrate DEFAULT HOOD MODE to use site/domain configurations.

X  Use iwinfo lua lib instead of external call [ISSUE](https://git.ffnw.de/ffnw-firmware/packages/issues/114) and [MR](https://git.ffnw.de/ffnw-firmware/packages/merge_requests/63)

### DIFF
--- a/package/gluon-hoodselector/luasrc/usr/lib/lua/hoodselector/util.lua
+++ b/package/gluon-hoodselector/luasrc/usr/lib/lua/hoodselector/util.lua
@@ -1,0 +1,191 @@
+local fs = require 'nixio.fs'
+local json = require 'jsonc'
+local uci = require('simple-uci').cursor()
+local site = require 'gluon.site'
+
+local M = {}
+
+function M.split(s, delimiter)
+  local result = {};
+  for match in (s..delimiter):gmatch("(.-)"..delimiter) do
+    table.insert(result, match);
+  end
+  return result;
+end
+
+local PID = M.split(io.open("/proc/self/stat", 'r'):read('*a'), " ")[1]
+
+function M.log(msg)
+  if msg then
+    io.stdout:write(msg.."\n")
+    os.execute("logger hoodselector["..PID.."]: "..msg)
+  end
+end
+
+-- bool if direct VPN. The detection is realaise by searching the fastd network interface inside the originator table
+function M.directVPN(vpnIfaceList)
+  for _,vpnIface in ipairs(vpnIfaceList) do
+    local file = io.open("/sys/kernel/debug/batman_adv/bat0/originators", 'r')
+    if file ~= nil then
+      for outgoingIF in file:lines() do
+        -- escape special chars "[]-"
+        if outgoingIF:match(string.gsub("%[  " .. vpnIface .. "%]","%-", "%%-")) then
+          return true
+        end
+      end
+    end
+  end
+  return false
+end
+
+function M.fastd_installed()
+  if io.open("/usr/bin/fastd", 'r') ~= nil then
+    return true
+  end
+  return false
+end
+
+function M.tunneldigger_installed()
+  if io.open("/usr/bin/tunneldigger", 'r') ~= nil then
+    return true
+  end
+  return false
+end
+
+-- Get Geoposition. Return nil for no position
+function M.getGeolocation()
+  return {["lat"] = tonumber(uci:get('gluon-node-info', uci:get_first('gluon-node-info', 'location'), 'latitude')),
+  ["lon"] =  tonumber(uci:get('gluon-node-info', uci:get_first('gluon-node-info', 'location'), 'longitude')) }
+end
+
+function M.get_domains()
+  local list = {}
+  for domain_path in fs.glob('/lib/gluon/domains/*.json') do
+    table.insert(list, {
+      domain_code = domain_path:match('([^/]+)%.json$'),
+      domain = assert(json.load(domain_path)),
+    })
+  end
+  return list
+end
+
+-- Source with pseudocode: https://de.wikipedia.org/wiki/Punkt-in-Polygon-Test_nach_Jordan
+-- see also https://en.wikipedia.org/wiki/Point_in_polygon
+-- parameters: points A = (x_A,y_A), B = (x_B,y_B), C = (x_C,y_C)
+-- return value: −1 if the ray from A to the right bisects the edge [BC] (the lower vortex of [BC]
+-- is not seen as part of [BC]);
+--                0 if A is on [BC];
+--               +1 else
+function M.crossProdTest(x_A,y_A,x_B,y_B,x_C,y_C)
+  if y_A == y_B and y_B == y_C then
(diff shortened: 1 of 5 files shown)


lemoer:
@2tata Nice, that the work is going on here. Did you have a look at #1216 yet? I think, its nearly ready to merge.

What it does? It merges (selected) *domain* and *site* config into one config accessible via api. Same schema for settings is used in domain and in site configs. But since not all config settings make sense in a domain specific way, some are only allowed in site config. Some other values are only allowed in domain config. And the rest is allowed in both. This is implemented by extending the *check_site_lib.lua*. The existing ways to access the site (via site.lua or libgluonutil) now return a merged config from domain and site. Selecting a domain is handled via uci value *gluon.system.domain* for now.

What would be to do, to integrate hoodselector here:
- For now, we evaluate most parts of the config in /lib/gluon/upgrade/* scripts, which modify (persistent) uci variables. I think some work is done in netifd files, which directly read the site (without uci/flash writes). But this is the minor part.
- It would be very nice here, if we could change the district without persisting values to flash. Maybe you have an idea here?
- If the last step is done, the hoodselector could be a package, which changes just the actual domain via api.

What do you think?

2tata:
On 09/19/17 01:44, lemoer wrote:
> @2tata Nice, that the work is going on here. Did you have a look at #1216 yet? I think, its nearly ready to merge.

Yes I did.


> What would be to do, to integrate hoodselector here:
> - For now, we evaluate most parts of the config in /lib/gluon/upgrade/* scripts, which modify (persistent) uci variables. I think some work is done in netifd files, which directly read the site (without uci/flash writes). But this is the minor part.
> - It would be very nice here, if we could change the district without persisting values to flash. Maybe you have an idea here?

We had a discussion about that: should the hoodselector write conf persistent on flash. Our decision was: Yes,
because we would avoid scenarios where thousands of routers trying to connect to the default hood,  after rebooting.
The hoodselector write only on the flash if the hood/domain has changes.

> - If the last step is done, the hoodselector could be a package, which changes just the actual domain via api.
> 
> What do you think?

The hoodselector do nearly the same what do have explained above. In case if #1216 get merges it should be quit
easy to change the hoodselector to use the domain specific configs instead of the hood.json.

currently I am working on finalising the state machine whitch means "hoodselector will never leave neigbour hood if a mesh router connect to them."
and also on "Non rectangle polygones using Point in polygon test after Jordan".

cheers
Tarek

package/gluon-hoodselector/luasrc/usr/lib/lua/hoodselector/util.lua
```diff
@@ -0,0 +1,191 @@
...
+function M.getDefaultHood(jhood)
+  for _, h in pairs(jhood) do
+    if h.domain_code == site.default_domain() then
+      return h
+    end
+  end
+  return nil
+end
+
+function M.restart_services()
```
2tata: @NeoRaider : If that like you expect? I am not sure if I have all processes listed, which needs to be restarted.

CodeFetch:
@2tata As Gluon has multi-domain support now, why are we still talking about hoods? Is there a difference?

2tata:
hoods are equal to domains just a name

CodeFetch:
@2tata It would be nice if the domain-selector would work independent from the method to determine the correct domain. That means split it in a package gluon-domain-selector-core and e.g. gluon-domain-selector-geo-polygons. This way kpanic23 could provide a package gluon-domain-selector-geo-webapi with similar functionality. 
I don't like the idea of only having fixed borders for the domains, because it is frustrating having setup a node which has two connections to other nodes in different domains one with good signal strength and one with bad and unfortunately the coordinates say the node is in the domain of the node to which it has a bad connection. There needs to be a function in the gluon-domain-selector-core package to prioritize the methods. E.g. first use gluon-domain-selector-choose-neighbours which is a package that checks if there are wifi neighbours around and if that is the case it uses a simple formula to derive the domain which gives the best connectivity in average or even simpler just choose the domain of the node to which it has the best wifi connection. If that fails, because there are no neighbours around or the connection threshold is too low or whatever, use gluon-domain-selector-choose-geo which tries to derive the domain from one of the gluon-domain-selector-geo-* packages depending on which one is installed and provides the gluon-domain-selector-geo meta-package. If that fails e.g. because it does not know any coordinates, use gluon-domain-selector-choose-default which is a package that simply chooses the default domain.
I just think the complexity of the gluon-hoodselector  is too high to not separate it into different packages.

PR #1227, opened 2017-09-20
libgluonutil: add function that returns the mesh-prefix from site.conf

christf:
This function will allow to calculate the node-IP. currently it is only used in the babel PR but it migh be interesting for wider use as well.

### DIFF
--- a/package/libgluonutil/src/libgluonutil.c
+++ b/package/libgluonutil/src/libgluonutil.c
@@ -27,7 +27,7 @@
 #include "libgluonutil.h"
 
 #include <json-c/json.h>
-
+#include <arpa/inet.h>
 #include <stdio.h>
 #include <string.h>
 #include <stdlib.h>
@@ -108,6 +108,38 @@ struct json_object * gluonutil_wrap_and_free_string(char *str) {
 	free(str);
 	return ret;
 }
+
+
+bool gluonutil_get_mesh_prefix(struct in6_addr *result) {
+	struct json_object *site = gluonutil_load_site_config();
+	if (!site)
+		return false;
+
+	struct json_object *node_prefix = NULL;
+	if (!json_object_object_get_ex(site, "node_prefix6", &node_prefix)) {
+		json_object_put(site);
+		return false;
+	}
+
+	const char *str_prefix = json_object_get_string(node_prefix);
+	if (!str_prefix) {
+		json_object_put(site);
+		return false;
+	}
+
+	char *prefix_addr = strndup(str_prefix, strchrnul(str_prefix, '/')-str_prefix);
+
+	int rc = inet_pton(AF_INET6, prefix_addr, result);
+
+	free(prefix_addr);
+	json_object_put(site);
+
+	if (rc != 1 )
+		return false;
+
+	return true;
+}
+
 
 struct json_object * gluonutil_load_site_config(void) {
 	return json_object_from_file("/lib/gluon/site.json");
--- a/package/libgluonutil/src/libgluonutil.h
+++ b/package/libgluonutil/src/libgluonutil.h
@@ -27,10 +27,15 @@
 #ifndef _LIBGLUON_LIBGLUON_H_
 #define _LIBGLUON_LIBGLUON_H_
 
+#include <netinet/in.h>
+#include <stdbool.h>
+
+
 char * gluonutil_read_line(const char *filename);
 char * gluonutil_get_sysconfig(const char *key);
 char * gluonutil_get_node_id(void);
 char * gluonutil_get_interface_address(const char *ifname);
+bool gluonutil_get_mesh_prefix(struct in6_addr *retval);
 
 struct json_object * gluonutil_wrap_string(const char *str);
 struct json_object * gluonutil_wrap_and_free_string(char *str);


NeoRaider:
As mentioned in IRC, I'd like to support using separate prefixes for nodes and clients for batman-adv-based Gluon firmwares as well eventually. Therefore, I'd prefer a more generic name for this field - something like `node_prefix` or `node_prefix6` - or maybe you have a better idea for the name?

christf:
how about that one? This places node_prefix6 on the same level as prefix6 and removes information referring to babel..

package/libgluonutil/src/libgluonutil.c
```diff
@@ -109,6 +109,33 @@ struct json_object * gluonutil_wrap_and_free_string(char *str) {
...
+	struct json_object *node_prefix = NULL;
+	if (!json_object_object_get_ex(site, "node_prefix6", &node_prefix))
+		return NULL;
+
+	char *str_prefix = (char*)json_object_get_string(node_prefix);
+	if (!str_prefix)
+		return NULL;
+
+	// we can do the strchr here because in check_site.lua we have validated the format to be some address::/64
+	char *prefix_addr = strndup(str_prefix, strchr(str_prefix, '/')-str_prefix);
```
NeoRaider: Just use *strchrnul* instead of *strchr*, no need to rely on site checks (I'd like to make the C code as robust as possible, as we don't have Lua's error()/assert() here).

package/libgluonutil/src/libgluonutil.c
```diff
@@ -109,6 +109,33 @@ struct json_object * gluonutil_wrap_and_free_string(char *str) {
 	return ret;
 }
 
+
+struct in6_addr* gluonutil_get_mesh_prefix() {
```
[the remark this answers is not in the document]
christf: There are several options here.
I  like the allocate-space-and-pass-it-pattern.
What is you preference?

package/libgluonutil/src/libgluonutil.c
christf: done.

package/libgluonutil/src/libgluonutil.h
```diff
@@ -26,11 +26,14 @@
 
 #ifndef _LIBGLUON_LIBGLUON_H_
 #define _LIBGLUON_LIBGLUON_H_
+#include <arpa/inet.h>
```
NeoRaider: This should include `netinet/in.h` for `struct in6_addr`, `arpa/inet.h` should be moved to the main file.

package/libgluonutil/src/libgluonutil.h
```diff
@@ -26,11 +26,14 @@
...
+#include <stdbool.h>
```
NeoRaider: Add a few blank lines here (one blank line between include guards and include lines, two between the include lines and the other declarations).

package/libgluonutil/src/libgluonutil.c
```diff
@@ -109,6 +109,31 @@ struct json_object * gluonutil_wrap_and_free_string(char *str) {
...
+	if (!json_object_object_get_ex(site, "node_prefix6", &node_prefix))
+		return false;
+
+	char *str_prefix = (char*)json_object_get_string(node_prefix);
+	if (!str_prefix)
+		return false;
+
+	char *prefix_addr = strndup(str_prefix, strchrnul(str_prefix, '/')-str_prefix);
+
+	inet_pton(AF_INET6, prefix_addr, &(result->s6_addr));
```
NeoRaider: *inet_pton* expects an in6_addr pointer, just write `inet_pton(AF_INET6, prefix_addr, result);`

NeoRaider: You should also check the return value of *inet_pton* and only return true on success.

package/libgluonutil/src/libgluonutil.c
```diff
@@ -109,6 +109,31 @@ struct json_object * gluonutil_wrap_and_free_string(char *str) {
...
+bool gluonutil_get_mesh_prefix(struct in6_addr *result) {
+	struct json_object *site = gluonutil_load_site_config();
+	if (!site)
+		return false;
+
+	struct json_object *node_prefix = NULL;
+	if (!json_object_object_get_ex(site, "node_prefix6", &node_prefix))
+		return false;
+
+	char *str_prefix = (char*)json_object_get_string(node_prefix);
```
NeoRaider: This `(char*)` cast doesn't make sense, just make the variable a `const char *`.

christf: done.

package/libgluonutil/src/libgluonutil.c
christf: done

package/libgluonutil/src/libgluonutil.h
christf: done.

package/libgluonutil/src/libgluonutil.h
christf: done. What is the reasoning behind moving arpa/inet.h to the main file?

NeoRaider: One should always try to use the "smallest" include, especially in public header files, to avoid forcing others to compile more code than necessary when they use your headers. The header file only needs `struct in6_addr`, which is defined in `netinet/in.h` according to the docs.

christf:
my build with the fixes was successful, so removinb WIP.

package/libgluonutil/src/libgluonutil.h
```diff
@@ -27,10 +27,16 @@
 #ifndef _LIBGLUON_LIBGLUON_H_
 #define _LIBGLUON_LIBGLUON_H_
 
+#include <netinet/in.h>
+#include <arpa/inet.h>
```
NeoRaider: The `arpa/inet.h` in include can be removed here now.

package/libgluonutil/src/libgluonutil.c
```diff
@@ -109,6 +109,34 @@ struct json_object * gluonutil_wrap_and_free_string(char *str) {
...
+	if (!site)
+		return false;
+
+	struct json_object *node_prefix = NULL;
+	if (!json_object_object_get_ex(site, "node_prefix6", &node_prefix))
+		return false;
+
+	const char *str_prefix = json_object_get_string(node_prefix);
+	if (!str_prefix)
+		return false;
```
NeoRaider: This error path and the one above don't `json_object_put(site)`.

package/libgluonutil/src/libgluonutil.c
```diff
@@ -109,6 +109,34 @@ struct json_object * gluonutil_wrap_and_free_string(char *str) {
...
+
+	int rc = inet_pton(AF_INET6, prefix_addr, result);
+
+	free(prefix_addr);
+	json_object_put(site);
+
+	if (rc == 1 )
+		return true;
+
+	return false;
```
NeoRaider: Just a minor stylistic commend: I think it would be nice to have all the `return false` in if-branches, so the unconditional return at the end of the function is the `return true`.

package/libgluonutil/src/libgluonutil.c
christf: indeed, however that wasn't overlooked but actually new to the fixup. done.

package/libgluonutil/src/libgluonutil.c
christf: it does make sense to keep this the same style. done.

package/libgluonutil/src/libgluonutil.h
christf: done.

christf:
... finished the re-work.

NeoRaider:
Merged (with renamed function name).

PR #1229, opened 2017-09-29
config allow more width in style for input fields and selectboxes

rubo77:
With the limited `width` selectboxes did not resize to the maximum width if there were options that are longer than 20em

### DIFF
--- a/package/gluon-web-theme/sass/cascade.scss
+++ b/package/gluon-web-theme/sass/cascade.scss
@@ -351,7 +351,7 @@ input[type=image] {
 select,
 input[type=text],
 input[type=password] {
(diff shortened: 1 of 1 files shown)


rotanid:
typo in your commit message - seriously?

rotanid:
well ok, i corrected the commit message for you and merged it.

NeoRaider:
@rubo77 Wouldn't out layout look weird if we allow wider elements? Please provide screenshots!

rotanid:
as far as i understand it, the box will still be at 20em as long as the content fits in this space.
if it doesn't, it would - without this change - either look weird or hide part of the content

PR #1231, opened 2017-10-01
base-files: banner: Replace OpenWRT banner with a Freifunk banner

PolynomialDivision:
base-files: banner: replace openwrt banner with freifunk

Based on [Berlin Luci Banner](https://github.com/freifunk-berlin/firmware/blob/master/patches/700-banner-info.patch).

### DIFF
--- a/patches/openwrt/0900-replace-banner-with-freifunk.patch
+++ b/patches/openwrt/0900-replace-banner-with-freifunk.patch
@@ -6,7 +6,7 @@ Replace openwrt banner with freifunk
 Signed-off-by: Nick Hainke <vincent@systemli.org>
 
 diff --git a/package/base-files/files/etc/banner b/package/base-files/files/etc/banner
-index 799f7e5..628447c 100644
+index 799f7e5..a87c7c9 100644
 --- a/package/base-files/files/etc/banner
 +++ b/package/base-files/files/etc/banner
 @@ -1,14 +1,17 @@
@@ -15,12 +15,12 @@ index 799f7e5..628447c 100644
 - |   -   ||  _  |  -__|     ||  |  |  ||   _||   _|
 - |_______||   __|_____|__|__||________||__|  |____|
 -          |__| W I R E L E S S   F R E E D O M
-+  _____        _  __             _
-+ |  ___|      (_)/ _|           | |
-+ | |_ _ __ ___ _| |_ _   _ _ __ | | __
-+ |  _| '__/ _ \ |  _| | | | '_ \| |/ /
-+ | | | | |  __/ | | | |_| | | | |   <
-+ \_| |_|  \___|_|_|  \__,_|_| |_|_|\_\
++   _____ _     _    _  ____  _   _ 
++  / ____| |   | |  | |/ __ \| \ | |
++ | |  __| |   | |  | | |  | |  \| |
++ | | |_ | |   | |  | | |  | | . ` |
++ | |__| | |___| |__| | |__| | |\  |
++  \_____|______\____/ \____/|_| \_|                                                              
   -----------------------------------------------------
 - CHAOS CALMER (%C, %R)
 - -----------------------------------------------------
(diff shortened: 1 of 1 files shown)


patches/openwrt/0900-replace-banner-with-freifunk.patch
```diff
@@ -0,0 +1,41 @@
...
+- |       |.-----.-----.-----.|  |  |  |.----.|  |_
+- |   -   ||  _  |  -__|     ||  |  |  ||   _||   _|
+- |_______||   __|_____|__|__||________||__|  |____|
+-          |__| W I R E L E S S   F R E E D O M
++  _____        _  __             _
++ |  ___|      (_)/ _|           | |
++ | |_ _ __ ___ _| |_ _   _ _ __ | | __
++ |  _| '__/ _ \ |  _| | | | '_ \| |/ /
++ | | | | |  __/ | | | |_| | | | |   <
++ \_| |_|  \___|_|_|  \__,_|_| |_|_|\_\
```
mweinelt: > Gluon is a modular framework for creating OpenWrt-based firmwares for wireless mesh nodes. Several Freifunk communities in Germany use Gluon as the foundation of their Freifunk firmwares.

While several Freifunk communities use Gluon to build their firmwares, it is not a dedicated Freifunk firmware framework.

patches/openwrt/0900-replace-banner-with-freifunk.patch
```diff
@@ -15,12 +15,12 @@ index 799f7e5..628447c 100644
...
-+ | |_ _ __ ___ _| |_ _   _ _ __ | | __
-+ |  _| '__/ _ \ |  _| | | | '_ \| |/ /
-+ | | | | |  __/ | | | |_| | | | |   <
-+ \_| |_|  \___|_|_|  \__,_|_| |_|_|\_\
++   _____ _     _    _  ____  _   _ 
++  / ____| |   | |  | |/ __ \| \ | |
++ | |  __| |   | |  | | |  | |  \| |
++ | | |_ | |   | |  | | |  | | . ` |
++ | |__| | |___| |__| | |__| | |\  |
++  \_____|______\____/ \____/|_| \_|
```
PolynomialDivision: @mweinelt 
Maybe like this? But there is no need to merge this pull request. ^^

patches/openwrt/0900-replace-banner-with-freifunk.patch
```diff
@@ -29,8 +29,8 @@ index 799f7e5..628447c 100644
 -  * 3/4 oz Lime Juice       unstrained into a goblet.
 -  * 1 1/2 oz Orange Juice
 -  * 1 tsp. Grenadine Syrup
-+ Gluon Firmware (%N %V rev %C)
-+ %P - %S
++ Gluon Firmware (Chaos Calmer 15.05.1 rev Chaos Calmer)
++ Generic - x86/64
 + https://github.com/freifunk-gluon/gluon
 + https://freifunk.net/
```
PolynomialDivision: This should be deleted too...

NeoRaider:
We generally don't accept non-upstreamable patches in Gluon. We have a few old ones that will need considerable rework before being upstreamed (in total, Gluon contains about 10 patches that aren't upstream at the moment), but we'd rather like do decrease that number than add new ones.

rotanid:
there are custom packages out there that you can use to replace the default banner

PolynomialDivision:
@rotanid : Can you give me the name of this package? :)

mweinelt:
`gluon-banner` 
https://github.com/freifunk-ffm/packages/tree/master/ffffm-banner

PolynomialDivision:
Yeah but I don't like this solution... You can't add the placeholders and so on...
So no information about version and stuff like this...

I'm doing something like this now:
https://github.com/PolynomialDivision/site/commit/bf8434eaeee9e33f41eb6163b585cfec2f3a5bf5

PR #1233, opened 2017-10-03
docs: v2017.1.3 release notes

NeoRaider:


jplitza:
> Fortunately, the most severe (remote code execution) vulnerabilities are in the DHCP component of dnsmasq

I think this is only partly true. CVE-2017-14491 is an RCE in the DNS component and thus very well poses a threat to Gluon nodes with DNS caching enabled, doesn't it?

NeoRaider:
@jplitza Thanks for pointing that out, you're completely right. It seems I misread the advisory.

I'll fix that up and make the v2017.1.3 release tonight.

NeoRaider:
Release notes corrected and extended with some background information.

PR #1239, opened 2017-10-14
docs: mtu: fix reference to faq entry, add mtu diagram locally

mweinelt:
Fixes the broken link from site to the faq entry and adds the mtu diagramm to the repository.

Is there an easy way to preview/prerender the docs to see if changes check out?


rubo77:
I think you simply go into the docs directory and use

    make

belzebub40k:
> Is there an easy way to preview/prerender the docs to see if changes check out?

You can compile the docs locally with [sphinx](http://www.sphinx-doc.org/en/stable/) if you type `make html` in the docs folder.

mweinelt:
Thanks, the links look fixed now.

PR #1244, opened 2017-10-22
changed groupname gluon-fastd to gluon-mesh-vpn

lrnzo:
somewhere between gluon-v2016.2.7 and gluon-v2017.1.3 the groupname had changed, hence it should also change in the documentation


NeoRaider:
Thanks, I've applied your patch to both the v2017.1.x and the master branch (with improved commit message).

For future contributions, please always add a subsystem prefix to the commit message (package name, or simply "docs: " in case of documentation). Also, all fixes must go into master first before they are backported to the stable branches.

PR #1247, opened 2017-10-28
[RFC] add support for Ubiquiti EdgeRouter X-SFP

kb-light:
tested with v2017.1.x
requires #1173 to enable PoE within config mode

~EDIT: build seems to fail on master~
EDIT2: turns out, it was just an issue with my site.[conf|mk]


NeoRaider:
I removed the patch "ramips: simplify ubnt-erx-sfp device definition" (not all of its dependencies have been backported), the rest has been merged.

PR #1250, opened 2017-11-05
[WIP] new package gluon-config-mode-domain-select

kb-light:
First draft for gluon-config-mode-domain-select. (depends on #1216)

ToDo:

- [x] get config generation working after changing the domain (done, but I don't like the approach)
- [x] sort list by `domain_name`
- [x]  get it working with aliases (need to be tested)
- [ ] maybe integrate `get_domain_list()` to `gluon.site`
- [ ] evtl. eintragen in `https://github.com/freifunk-gluon/gluon/blob/master/package/feature`

### DIFF
--- a/docs/package/gluon-config-mode-domain-select.rst
+++ b/docs/package/gluon-config-mode-domain-select.rst
@@ -1,0 +1,19 @@
+gluon-config-mode-domain-select
+===============================
+This package provides a drop-down list for the config mode to select the domain
+the node will be placed in. If the selection has changed the upgrade scripts in
+``/lib/gluon/upgrade/`` are triggered to update the nodes configuration.
+
+Hiding domains could be useful for default or testing domains, which should not
+be accidentally selected by a node operater.
+
+domains/*.conf
+---------------------
+
+hide_domain \: optional (defaults to false)
+    - ``false`` shows this domain in drop-down list
+    - ``true`` hides this domain
+
+Example::
+
+  hide_domain = true
--- a/docs/site-example/i18n/de.po
+++ b/docs/site-example/i18n/de.po
@@ -16,6 +16,15 @@ msgstr ""
 "Freifunk-Knoten. Fülle das folgende Formular deinen Vorstellungen "
 "entsprechend aus und sende es ab."
 
+msgid "gluon-config-mode:domain"
+msgstr "Domäne"
+
+msgid "gluon-config-mode:domain-select"
+msgstr ""
+"Hier hast du die Möglichkeit, die Mesh-Domäne, in der sich dein Knoten "
+"befindet, auszuwählen. Bitte denke daran, dass sich dein Knoten nur mit den "
+"Knoten der ausgewählten Domäne verbinden kann."
+
 msgid "gluon-config-mode:pubkey"
 msgstr ""
 "<p>Dies ist der öffentliche Schlüssel deines Freifunk-Knotens. Erst nachdem "
--- a/docs/site-example/i18n/en.po
+++ b/docs/site-example/i18n/en.po
@@ -15,6 +15,15 @@ msgstr ""
 "Welcome to the setup wizard of your new Freifunk Alpha Centauri node. "
 "Please fill out the following form and submit it."
 
+msgid "gluon-config-mode:domain"
+msgstr "Domain"
+
+msgid "gluon-config-mode:domain-select"
+msgstr ""
+"Here you have the possibility of selecting the mesh domain in which your node "
+"is placed. Please keep in mind that your router only connects with the nodes "
+"of the selected domain"
+
 msgid "gluon-config-mode:pubkey"
 msgstr ""
 "<p>This is your Freifunk node's public key. The node won't be able to "
--- a/package/gluon-config-mode-domain-select/Makefile
+++ b/package/gluon-config-mode-domain-select/Makefile
@@ -1,0 +1,42 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-config-mode-domain-select
+PKG_VERSION:=1
+
+PKG_BUILD_DIR := $(BUILD_DIR)/$(PKG_NAME)
+
+include ../gluon.mk
+
+PKG_CONFIG_DEPENDS += $(GLUON_I18N_CONFIG)
+
+
+define Package/gluon-config-mode-domain-select
+  SECTION:=gluon
+  CATEGORY:=Gluon
+  TITLE:=UI for changing the node-config
+  DEPENDS:=+gluon-config-mode-core @GLUON_MULTIDOMAIN
+endef
+
+define Build/Prepare
+	mkdir -p $(PKG_BUILD_DIR)
+endef
+
+define Build/Configure
+endef
+
+define Build/Compile
+	$(call GluonBuildI18N,gluon-config-mode-domain-select,i18n)
+	$(call GluonSrcDiet,./luasrc,$(PKG_BUILD_DIR)/luadest/)
+endef
+
+define Package/gluon-config-mode-domain-select/install
+	$(CP) $(PKG_BUILD_DIR)/luadest/* $(1)/
+	$(call GluonInstallI18N,gluon-config-mode-domain-select,$(1))
+endef
+
+define Package/gluon-config-domain-select/postinst
+#!/bin/sh
+$(call GluonCheckSite,check_site.lua)
+endef
+
+$(eval $(call BuildPackage,gluon-config-mode-domain-select))
--- a/package/gluon-config-mode-domain-select/luasrc/lib/gluon/config-mode/wizard/0200-domain-select.lua
+++ b/package/gluon-config-mode-domain-select/luasrc/lib/gluon/config-mode/wizard/0200-domain-select.lua
@@ -1,6 +1,7 @@
 return function(form, uci)
 	local fs = require 'nixio.fs'
 	local json = require 'jsonc'
+	local site = require 'gluon.site'
 
 	local function get_domain_list()
 		local list = {}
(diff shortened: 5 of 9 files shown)


mweinelt:
I'm wondering what behaviour setting a domain (*manual*) should have versus a conflicting automatic decision making mechanism (*automat*). I think this is the right time & place to talk about this, because it would need to be reflected in the UI. Also I think we could use a common way to display not only domains, but also installed automatic mechanisms.

a. Can any automat be disabled from UCI or UI?
b. Is any automat overruled when a domain is set manually?
c. Should automat use the same UI element as the domains?
  1. Option `Automatic`
  2. Or a checkbox toggling automatic on/off
  3. Or a radio box switching between manual / automatic mode, 
      i. perhaps with checkboxes representing the individual registered automats
      ii. or with a place for each automation to hook in their individual settings, if any

kb-light:
a. that should be part of the automat. But we may provide a flag like `gluon.system.auto_domain_select`, which can be set via advanced settings. Or just as part of the drop-down, where it is possible to select "automatic domain selection" (which requires coordinates or something similar).

b. I would expect, that manual selection overwrites an automatic selection. Even if it is wrong.

c. not knowing which automates are available in future makes it hard to decide how they could be selected. But I think that the selection should be part of "advanced settings".

d. Currently I am just thinking about how we could update the configuration (automatically) after `gluon.system.domain` has changed.

mweinelt:
b. That ("Even if it is wrong") does not sound like a manual domain select should be on the default page, but an expert setting - and automatic should be the default.

c. They should probably all adhere to a `gluon.system.auto_domain_select.enabled` boolean

kb-light:
b. the problem is, that with manual selection you can force a wrong domain. but for automatic selcetion you even need manual entered data and I would expect, that it is easier to select the right domain, instead of setting the right coordinates. Additionally each community has different needs, so we could provide the same settings for the "advanced settings", too (`gluon-web-domain-select`).

docs/site-example/i18n/de.po
```diff
@@ -16,6 +16,15 @@ msgstr ""
...
 "entsprechend aus und sende es ab."
 
+msgid "gluon-config-mode:domain"
+msgstr "Domäne"
+
+msgid "gluon-config-mode:domain-select"
+msgstr ""
+"Hier hast du die Möglichkeit die Domäne, in der sich dein Knoten befindet, "
+"auszuwählen. Bitte denke daran, dass sich dein Knoten nur mit dem Netz der "
+"ausgewählten Domäne verbinden kann."
```
mweinelt: > nur mit Knoten der auswählten Domäne verbinden kann
Wir wollen ja auf den Mesh-Aspekten hinweisen, oder?

package/gluon-config-mode-domain-select/luasrc/lib/gluon/config-mode/wizard/0200-domain-select.lua
```diff
@@ -0,0 +1,59 @@
...
+			local domain_code = domain_path:match(path .. '/(.*)\.json$')
+			table.insert(list, domain_code)
+		end
+		return list
+	end
+
+	local s = form:section(Section, nil, translate('gluon-config-mode:domain-select'))
+
+	local o = s:option(ListValue, 'domain', translate('gluon-config-mode:domain'))
+	o.optional = false
```
mweinelt: Please set `o.default` so an already selected domain is preselected in the select box.

kb-light:
Sometimes the code seems to fail and not all domains were displayed

rotanid:
@kb-light now that #1216 has been merged, this can be updated to be also ready for final review

NeoRaider:
As mentioned on IRC, I have a different solution for the `ListValue.optional` handling in mind.

package/gluon-config-mode-domain-select/luasrc/lib/gluon/config-mode/wizard/0200-domain-select.lua
```diff
@@ -0,0 +1,48 @@
+return function(form, uci)
+	local site = require 'gluon.site'
+	local path = '/lib/gluon/domains'
+	local fs = require 'nixio.fs'
+	local json = require 'jsonc'
+
+	local function get_domain_list()
+		local list = {}
+		for domain_path in io.popen('find ' .. path .. '/ -iname \*.json -print'):lines() do
```
NeoRaider: - `\*` is no valid escape in Lua. Either use `\\*` to get a backslash in Shell, or replace all `'` with `"`, so you can just use `'*.json'` without any backslashes.
- use `-name` instead of `-iname`
- `-print` can be removed

... but even better would be to remove the whole *io.popen* and use *glob* from *nixio.fs* instead.

package/gluon-config-mode-domain-select/luasrc/lib/gluon/config-mode/wizard/0200-domain-select.lua
```diff
@@ -0,0 +1,48 @@
...
+			local domain_code = domain_path:match(path .. '/(.*)\.json$')
```
NeoRaider: `\.` is no valid escape in Lua, I assume you mean `%.`.

package/gluon-config-mode-domain-select/Makefile
```diff
@@ -0,0 +1,42 @@
...
+include ../gluon.mk
+
+PKG_CONFIG_DEPENDS += $(GLUON_I18N_CONFIG)
+
+
+define Package/gluon-config-mode-domain-select
+  SECTION:=gluon
+  CATEGORY:=Gluon
+  TITLE:=UI for changing the node-config
+  DEPENDS:=+gluon-config-mode-core
```
NeoRaider: Add `@GLUON_MULTIDOMAIN` to the DEPENDS.

docs/package/gluon-config-mode-domain-select.rst
```diff
@@ -0,0 +1,19 @@
+gluon-config-mode-domain-select
+===============================
+This Package provides a drop-down list for the config mode to select the domain
```
NeoRaider: "Package" -> "package"

docs/package/gluon-config-mode-domain-select.rst
```diff
@@ -0,0 +1,19 @@
...
+``/lib/gluon/upgrade/`` are triggered to update the nodes configuration.
+
+Hiding domains could be useful for default or testing domains, which should not
+be accidentally selected by a node operater.
+
+domains/*.conf
+---------------------
+
+hide_domain \: optional (defaults to false)
+    - ``false`` shows this domain within drop-down list
```
NeoRaider: "within" -> "in"?

docs/site-example/i18n/en.po
```diff
@@ -15,6 +15,15 @@ msgstr ""
 "Welcome to the setup wizard of your new Freifunk Alpha Centauri node. "
 "Please fill out the following form and submit it."
 
+msgid "gluon-config-mode:domain"
+msgstr "Domain"
+
+msgid "gluon-config-mode:domain-select"
+msgstr ""
+"Here you have the possibility of selecting the domain in which your node is "
```
NeoRaider: I propose to call this "mesh domain" / "Mesh-Domäne" in the examples.

package/gluon-config-mode-domain-select/luasrc/lib/gluon/config-mode/wizard/0200-domain-select.lua
```diff
@@ -0,0 +1,48 @@
...
+		if not domain.hide_domain then
+			o:value(domain.domain_code, domain.domain_name)
+		end
+	end
+
+	function o:write(data)
+		-- local core = uci:get_first('gluon', 'core')
+		-- uci:set('gluon', core, 'domain', data )
+		-- os.execute('gluon-reconfigure')
+		fs.writefile('/tmp/new_domain_code', data)
```
NeoRaider: What is the problem with the commented code, why write a file in `/tmp`?

Also, we use named sections in Gluon, so you can just use `uci:set('gluon', 'core', ...)` without *get_first*.

kb-light: I had some issues with the commented code before. I will test it again.

NeoRaider: The main issue I see with the bit above is that you set the value, but don't save it afterwards (`uci:save('gluon')`), so gluon-reconfigure does not see the new value yet. Usually, the save is implied by the returned array `return {'gluon'}`, but in this case the save happens too late.

Moving the gluon-reconfigure call to the reboot step is not a bad idea, but I don't like using `/tmp` for it.

NeoRaider:
Please rebase onto master. ListValue now properly supports optional and unset values. As `o.optional` defaults to false, you can just leave `o.default` unset to achieve the desired behaviour (don't explicitly add an empty entry to the list of choices, this is done internally).

kb-light:
rebased

package/gluon-config-mode-domain-select/luasrc/lib/gluon/config-mode/wizard/0200-domain-select.lua
```diff
@@ -0,0 +1,42 @@
...
+	for _, domain in pairs(get_domain_list()) do
+		if not domain.hide_domain then
+			o:value(domain.domain_code, domain.domain_name)
+		end
+	end
+
+	function o:write(data)
+		uci:set('gluon', 'core', 'domain', data)
+		uci:save('gluon')
+		os.execute('gluon-reconfigure')
```
NeoRaider: Hmm. We could either make the *gluon-reconfigure* conditional, so it is executed only when the domain has actually changed. Or we could defer the *gluon-reconfigure* to a later step, when UCI options have actually been saved, to get rid of the explicit *uci:save*. Both would be desirable, but achieving both at the same time is a bit tricky... I'll think of something.

mweinelt:
When the selected domain is hidden it cannot be preselected in the config mode. 
Instead the first domain gets preselected. 
In that case a better behaviour would be to preselect the default domain.

kb-light:
I think it should be better to preselect the hidden (configured) domain so that the domain won't change. I will look into it.

NeoRaider:
Merged with a bit of cleanup.

PR #1256, opened 2017-11-10
Add site.conf option for fastd peer upgrade

### LINKED ISSUE #557: Fastd keeps old peers when updating to firmware with new site.conf
Since we had two fastd connetions per node I added peergroups in the past to force the nodes to connect to gateways in different data centers.

Since we changed the peer limit to one and started to move the gateways around I like to remove the peergroups.

Sadly the old entries are not removed from the /etc/config/fastd during autoupdate.

CodeFetch:
> The site.conf option `delete_old_peers` removes old peer entries with uci option "net" mesh_vpn
> and their corresponding groups recursively.

Shouldn't this resolve #557?

### DIFF
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -213,6 +213,9 @@ mesh_vpn
 
     You can set syslog_level from verbose (default) to warn to reduce syslog output.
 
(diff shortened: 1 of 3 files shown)


CodeFetch:
Can't this be resolved properly?

PR #1257, opened 2017-11-10
[RFC] Add site.conf option for fastd peer upgrade

### LINKED ISSUE #557: Fastd keeps old peers when updating to firmware with new site.conf

CodeFetch:
> The site.conf option `delete_old_peers` removes old peer entries with uci option "net" mesh_vpn
> and their corresponding groups recursively.

Shouldn't this resolve #557?

BTW: Sorry for my inability to use Github. I don't like pull requests.

### DIFF
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -292,6 +292,13 @@ mesh_vpn
 
     You can set syslog_level from verbose (default) to warn to reduce syslog output.
 
+    While performing a sysupgrade fastd retains existing peer group and peer configurations to allow
+    incremental addition of peers and peer groups by default. This behaviour might be unwanted.
+    To make sure that fastd's peers and peer groups from previous firmware versions are removed 
+    (except those defined by the node2node VPN feature) set `sysupgrade_remove_old_peers` to `true`.
+    To remove peers and groups defined by the node2node VPN feature (which have ``n2n_vpn`` in their 
+    names) set `sysupgrade_remove_n2n_peers` to `true`.
+
     The `tunneldigger` section is used to define the *tunneldigger* broker list.
 
     **Note:** It doesn't make sense to include both `fastd` and `tunneldigger`
@@ -313,6 +320,8 @@ mesh_vpn
           methods = {'salsa2012+umac'},
           -- configurable = true,
           -- syslog_level = 'warn',
+          -- sysupgrade_remove_old_peers = false,
+          -- sysupgrade_remove_n2n_peers = false,
           groups = {
             backbone = {
               -- Limit number of connected peers from this group
--- a/package/gluon-mesh-vpn-fastd/check_site.lua
+++ b/package/gluon-mesh-vpn-fastd/check_site.lua
@@ -1,6 +1,8 @@
 local fastd_methods = {'salsa2012+gmac', 'salsa2012+umac', 'null+salsa2012+gmac', 'null+salsa2012+umac', 'null'}
 need_array_of({'mesh_vpn', 'fastd', 'methods'}, fastd_methods)
 need_boolean(in_site({'mesh_vpn', 'fastd', 'configurable'}), false)
+need_boolean('mesh_vpn.fastd.sysupgrade_remove_old_peers', false)
+need_boolean('mesh_vpn.fastd.sysupgrade_remove_n2n_peers', false)
 
 need_one_of(in_site({'mesh_vpn', 'fastd', 'syslog_level'}), {'error', 'warn', 'info', 'verbose', 'debug', 'debug2'}, false)
 
--- a/package/gluon-mesh-vpn-fastd/luasrc/lib/gluon/upgrade/400-mesh-vpn-fastd
+++ b/package/gluon-mesh-vpn-fastd/luasrc/lib/gluon/upgrade/400-mesh-vpn-fastd
@@ -92,6 +92,63 @@ function add_groups(prefix, groups, parent)
 	end
 end
 
+-- Checks if a peer or peer group entry belongs to the node2node VPN feature
+local function is_n2n(gp)
+	if gp and gp['.name']:find('n2n_vpn') then return true end
+	return false
+end
+
+--- Determine peers and peer groups from the old configuration that can be removed
+	-- @param group		Peer group as returned by uci:foreach or 'nil' for ALL
+	-- @param deln2n	Set 'true' for adding node2node VPN groups else 'false'
+	-- @return      	Array with first index being boolean indicating whether
+	--			any n2n peer groups have been restrained and second
+	--			being an array with the names of the items to delete.
+local function get_delgroups(group, deln2n)
+	if not deln2n and is_n2n(group) then
+		return false
+	end
+
+	local group_empty = true
+	local dellist = {}
+
+	if group then
+		uci:foreach('fastd', 'peer', function(peer)
+			if peer.group == group['.name'] then
+				if not deln2n and is_n2n(peer) then
+					group_empty = false
+				else
+					table.insert(dellist, peer['.name'])
+				end
+			end
+
+		end)
+	end
+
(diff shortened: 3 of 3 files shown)


H4ndl3:
Next time please just update your pull request instead of opening a new one.

christf:
please adjust the commit message to include the module being changed.
gluon-mesh-vpn-fastd: <commitmessage>

docs/user/site.rst
```diff
@@ -230,6 +235,8 @@ mesh_vpn
           methods = {'salsa2012+umac'},
           -- configurable = true,
           -- syslog_level = 'warn',
+          -- delete_old_peers = false,
+          -- delete_custom_peers = false,
```
christf: can you elaborate on the use case of delete_custom_peers?

docs/user/site.rst
```diff
@@ -213,6 +213,11 @@ mesh_vpn
 
     You can set syslog_level from verbose (default) to warn to reduce syslog output.
 
+    To make sure that old fastd peer groups/peers (excluding those which have ``custom`` in their name)
+    are not retained after sysupgrade (e.g. updating from firmware from another community)
```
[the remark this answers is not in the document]
CodeFetch: Yes, it does.

docs/user/site.rst
CodeFetch: This option is more about being able to retain custom peers. As the default for site values is always false I had to name it this way.
If you update from another community and your community does not support N2N or you want to abandon N2N or never planned on introducing it, you might want to get custom peers deleted also.

rotanid: what is N2N? i don't know about custom peers and N2N either, so this should be documented with this Pull Request so everyone knows what it is

christf: node to node VPN. I think it is great supporting that however I did not get this from the custom peers and rather would call them n2n_peers but that is just taking nit-picking to an extreme.

CodeFetch:
I changed the title to make clear that I didn't want to get this merged as it is, but to drive forth the fallen asleep topic about handling peers on update.
The peer update mechanism is the only one I know of that has severe problems with cross-community flashing and old config retainment. Thus I thought that something like this could easily solve our problems until we have a configuration management system that fits Gluon better. There might be better ways to achieve this and that's why I ask you „Shouldn't this resolve the problem? Do you know a better way?“.

CodeFetch:
@rotanid Sorry, I'm very busy at the moment. I'm going to clean this up after initial xlat464 support.

rotanid:
@CodeFetch it's been another year now, whats the progress/plan here?

i'm not sure why people open PRs and don't do anything afterwards - this way opening the PR is wasted time for the opener...

CodeFetch:
@rotanid Done.

rotanid:
@CodeFetch thanks!
i just noticed you're taking care of a feature here (n2n) which isn't supported by Gluon at the moment. (see #378 )
maybe you have it included in your local communities fork, but for upstream Gluon please remove the part about n2n for the time being.
after(!) doing it, it would be great if you could test your work.

CodeFetch:
@rotanid That's why I didn't want to call it n2n_vpn, but custom. The user may want to keep some "custom" entries whether they are being set by the n2n feature or by hand.

mweinelt:
IMO: If site.conf ships a peer group that should supersede the locally configured peer group of the same name, so just replace that, no knob please. 
At the same time just ignore other peer groups to allow for local reconfiguration as was inteded by this feature.

CodeFetch:
@mweinelt This was discussed earlier and NeoRaider did not want to break the ability to incrementally add peers to peer groups. 
@NeoRaider What was the reason to allow incremental addition of peers?

@rotanid I'll have a look at the node2node feature the next days. It shouldn't be that hard to get it merged, should it? christf said it is not considered before the babel release. I don't see a real reason for this. I just want to get these annoying 3 year old PRs merged. Having an established fastd node2node solution means I can be certain how the uci structure looks like and may simplify this PR. Thus you can set it blocked.

CodeFetch:
@rotanid Okay I've realized that it was not clear what I want to achieve here and I was confused about it at some point, too.

At the moment there is only a single fastd mesh package called "gluon-mesh-vpn-fastd". This should be more modular. The package "gluon-mesh-vpn-fastd-core" should provide the configuration and upgrade scripts that are common to each flavor.
fastd has a "forward" option which enables forwarding of packets between peers. In a regular setup where a node is connected to one or more supernodes, only supernodes should forward packets between their peers. Thus the fastd configuration of a regular node always disables forwarding and it is not possible to add any custom peers there.
I propose to call this package "gluon-mesh-vpn-fastd-n2s".

The forwarding option of the node2node feature will always disable forwarding, too. The node2node feature should be able to define "custom" peers and ones defined using site.conf (If the node is a regular one it uses the node2node feature to use custom peerings defined by the user. However if it is a supernode it will use the node2node feature to connect with other supernodes and e.g. nameservers which will either be defined in the site.conf or manually). In that case the term "custom" and "n2n" peers cannot be used interchangeably anymore.
I propose to call this package "gluon-mesh-vpn-fastd-n2n".

In the future another package may come up to allow supernode functionality using gluon which will allow forwarding between peers using even another fastd instance providing connectivity to regular nodes. This package will have a configuration option to permit nodes to connect without authorization and will be able to handle site.conf and custom peer entries, too (thus the firmware does not always have to be flashed when a new node's public is being submitted).
I propose to call this package "gluon-mesh-vpn-fastd-s2n". 

Thus "custom" peers should not only be a feature of n2n, but be provided by the "gluon-mesh-vpn-fastd-core" package. I'm going to change this pull-request to not rely on the naming of the options to decide whether a peer/peer group is custom, but introduce a boolean option called "custom" to peers/peer groups and fastd instances. If custom is set to true on one level of the configuration hierarchy all children will be considered custom. I'm also going to split the "gluon-mesh-vpn-fastd" package into "gluon-mesh-vpn-fastd-core" and "gluon-mesh-vpn-fastd-n2s".

christf:
Today in review we decided the following:

- There will be gluon-specific init-scripts that act on their default from site.conf without requiring a uci config.
- in case a configuration exists for the package in question in /etc/config/gluon, this config will be allowed to override the content from site
- peer groups must be marked with a flag "preserve" that defines whether the peer group is to be preserved on upgrade. The gluon-mesh-vpn-fastd-core package must be adjusted to reflect that.
- there will be no data migration when switching to this scheme as we cannot determine whether a peer group is a leftover or if it was placed intentionally by the user.

NeoRaider:
@christf's comment sort of mixes up the short-term and long-term solutions I proposed: Short-term we will simply add a flag `preserve`, long-term we'd like to move additional configuration to `/etc/config/gluon`, making such flags unnecessary.

CodeFetch:
So the only change is that the "custom" flag should be called "preserve" instead?

NeoRaider:
- There should be no site.conf option - old peers without `preserve` flag are removed unconditionally
- Remove everything that mentions "n2n" (this is not an official Gluon feature)

Basically, the whole implementation can be simplified to:
- First, remove all groups and peers with net "mesh_vpn" and without `preserve`
- Then, run the usual setup for new groups and peers
I assume this can be done in 5 lines or fewer...

CodeFetch:
@NeoRaider Maybe I misunderstand you...

>     * First, remove all groups and peers with net "mesh_vpn" and without `preserve`

I think one needs to check whether a group's subelements contain the `preserve` flag recursively first. That's what my code does.

CodeFetch:
Meanwhile I've seen that there is a more elegant way to do this. Thus I'll look into it next week.

NeoRaider:
Superseded by #2021.

PR #1259, opened 2017-11-12
[RFC] introduce autoupdater delay

kb-light:
This PR adds the possibility, to delay the regular autoupdate for about one hour. This may be useful if you are using PoE passthrough to power an other node and you want to prevent an update of both nodes at the same time.

This PR is not thought for covering issues like changing the meshid or doing domain splits/network segmentation/...


package/gluon-web-autoupdater/i18n/de.po
```diff
@@ -18,3 +18,15 @@ msgstr "Branch"
 
 msgid "Enable"
 msgstr "Aktivieren"
+
+msgid "Delay"
+msgstr "Verzögerung"
+
+msgid ""
+"You can delay the regular autoupdate for about an hour. This may be usefull "
```
rotanid: typo: "useful"

package/gluon-web-autoupdater/i18n/de.po
```diff
@@ -18,3 +18,15 @@ msgstr "Branch"
...
+"if using the PoE passthrough for powering an other node."
```
rotanid: "an other" -> "another"

CodeFetch:
Wouldn't it make sense to always activate this if PoE is enabled? Thus in a PoE daisy chain scenario the nodes behind the PoE passthrough update first. I don't think one hour is a good value. It should be up to 12 hours or something and randomized. That way if nodes are daisy chained more than one level the risk for a brick decreases to nearly zero.

kb-light:
The first idea was, to just check, if PoE passthrough is enabled, but:
- there are maybe more use cases than just PoE passthrough
- I want to keep the code simple
- the node operator should have the possibility, to enable/disable the delay
 So I decided, to add a new uci flag and only check that one.

12 hours seams a bad idea, because than the update will happen on the afternoon.

Additional, if daisy chaining more than two nodes with PoE passthrough keep an eye on the limited outputpower. (You may overload the first node)

CodeFetch:
I don't do daisy chaining, but people told me they were using up to 4 devices with one cable where they didn't have enough cable connections. Don't understand me wrong. I like this feature and think it has even more potential.

christf:
If supporting something like that, the timeout should be configurable by the person setting up the routers. However for communities that use staggered upgrade - spreading the autoupdate over a week this feature makes no sense. 

I consider this a hack that it is trying to solve something using a delay that can only properly be solved by taking the current network topology into account. It feels like solving missing synchronisation in code by inserting calls to sleep().

kb-light:
Giving the node operater the choice to setup a time window, would be the more complex version. If that is a requested feature, that would be merged, we can have a look at it. Alternatively we may allow to overwrite the random minute value with a fixed value, but that would need a rewrite of the cron file.

I do not understand, what this feature has to do with staggered updates. Even if the priority is > 0 two nodes can update at the same time. This feature is not used for installations/updates, where you need a specific order to not break your local setup. This should only be used for setups, where two (or maybe more) nodes should not update at the same time. Honestly this does not provide 100% guarantee that the nodes will not update at the same time, because there is also the fallback update.

Maybe it is a hack but there won't be anything like the "perfect autoupdater".

rubo77:
We could add an option to the auto Updater: "only update during good between XX: XX and yy:yy".

That would serve more use cases than just cascading updates

kb-light:
Thanks for your feedback, I will close this in favour of #1266.

PR #1268, opened 2017-11-24
explain that not setting dns.servers makes no sense

### LINKED ISSUE #1265: Support DNS cache using "discovered" DNS servers
Ideally, we'd like the node's DNS server (when used as a cache by clients) to use exactly those resolvers that it discovered via IPv6 RAs (our RAs contain DNS server information). This is to make sure that if one of our gateways is down, it doesn't even try to use that for DNS resolution.

Right now, what we do is set the `cacheentries` in the `dns` section, but we do *not* set any `server`. The consequence of that seems to be that it indeed provides a DNS cache, and just uses the servers it discovered. However, the [documentation](https://gluon.readthedocs.io/en/v2017.1.x/features/dns-cache.html) does not explain what happens when `cacheentries` is present but `server` is not, so I am worried that in some future update, the behavior may change and suddenly, the WAN resolver is used. Would it be possible to support and document configurations with `cacheentries`, but without `server`?

RalfJung:
Fixes #1265

### DIFF
--- a/docs/features/dns-cache.rst
+++ b/docs/features/dns-cache.rst
@@ -16,8 +16,11 @@ There are the following settings:
 
 If both options are set the node will cache as much DNS records as set with
 'cacheentries' in RAM. The 'servers' list will be used to resolve the received
-DNS queries if the request cannot be answered from cache.
-If these settings do not exist, the cache is not intialized and RAM usage will not increase.
+DNS queries if the request cannot be answered from cache.  If these settings do
+not exist, the cache is not intialized and RAM usage will not increase.  Note
+that not setting 'servers' here defeates the purpose; once the routers all
+announce the "next node" address for DNS, there is no way for nodes to
+automatically determine DNS servers.  They have to be baked into the firmware.
 
 When next_node.name is set, an A record and an AAAA record for the
 next-node IP address are placed in the dnsmasq configuration. This means that the content
--- a/docs/features/dns-cache.rst
+++ b/docs/features/dns-cache.rst
@@ -17,8 +17,8 @@ There are the following settings:
 If both options are set the node will cache as much DNS records as set with
 'cacheentries' in RAM. The 'servers' list will be used to resolve the received
 DNS queries if the request cannot be answered from cache.  If these settings do
-not exist, the cache is not intialized and RAM usage will not increase.  Note
-that not setting 'servers' here defeates the purpose; once the routers all
+not exist, the cache is not initialized and RAM usage will not increase.  Note
+that not setting 'servers' here defeats the purpose; once the routers all
 announce the "next node" address for DNS, there is no way for nodes to
 automatically determine DNS servers.  They have to be baked into the firmware.
 
--- a/docs/features/dns-cache.rst
+++ b/docs/features/dns-cache.rst
@@ -14,13 +14,17 @@ There are the following settings:
     servers
     cacheentries
 
-If both options are set the node will cache as much DNS records as set with
-'cacheentries' in RAM. The 'servers' list will be used to resolve the received
-DNS queries if the request cannot be answered from cache.  If these settings do
-not exist, the cache is not initialized and RAM usage will not increase.  Note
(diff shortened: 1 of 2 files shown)


docs/features/dns-cache.rst
```diff
@@ -16,8 +16,11 @@ There are the following settings:
 
 If both options are set the node will cache as much DNS records as set with
 'cacheentries' in RAM. The 'servers' list will be used to resolve the received
-DNS queries if the request cannot be answered from cache.
-If these settings do not exist, the cache is not intialized and RAM usage will not increase.
+DNS queries if the request cannot be answered from cache.  If these settings do
+not exist, the cache is not intialized and RAM usage will not increase.  Note
+that not setting 'servers' here defeates the purpose; once the routers all
```
rotanid: *defeats

RalfJung: Thanks, fixed.

christf:
I find "defeats the purpose" to be correct but unspecific.

How about something like:
Not setting dns.servers will lead to clients not being able to resolve hosts using dns.

It certainly is not perfect either but shows a clear consequence of not making the setting.

kb-light:
To keep the style consistent, please only one space after a dot.

RalfJung:
Thanks for your feedback, I will incorporate it soon.

> Maybe we should also force dns.servers to be present, if dns.cacheentries is set (via corresponding check_site.lua).

I just had the same thought :)

RalfJung:
Done, including the `check_site`. However, I have essentially 0 lua experience, so please carefully review that. Also, I am not sure how to test it: I changed my site.conf to not contain `dns.servers` any more, but the build still starts as usual. It is now rebuilding the kernel. Shouldn't `check_site` be ran before doing the time-consuming actual built of everything?

christf:
Having had another look at the corresponding issue, I think the PR should include am mention of RDNS having to be switched off on central gateways on batman networks for the dns cache to work.

rotanid:
@christf would you mind writing a bit more about this? why should you disable reverse dns / PTR ?!?

RalfJung:
I don't think @christf means reverse DNS. He's talking about putting DNS server information into the RA.

@christf it should be switched off? My understanding was that it should be changed to announce the next-hop IPv6.

rotanid:
ah, RDNSS so there was an "S" missing. all clear now, but i second your question to christf.

christf:
sorry for being unclear before.
* the central RA must announce the next-node address or none at all. @RalfJung  so yes your understanding is correct. 
* My remark to the central DNS information not being announced was regard to the babel network. There it is just overhead and never seen by clients.

RalfJung:
So do you think anything should be changed in the text of this PR?

christf:
@RalfJung  the original issue was wrt to configuration of central servers. I feel, adding a remark here would make sense. something like "RDNSS configuration on the gateways must point to the next-node-address in Batman-based networks"

RalfJung:
> "RDNSS configuration on the gateways must point to the next-node-address in Batman-based networks"

Considering that the nodes also send RAs, I am wondering if this is really necessary?  Anyway, it doesn't hurt, so I'm going to adapt the text accordingly.

rotanid:
sorry to interrupt again, but if we say that RDNSS has to announce next-node, shouldn't DHCP promote exactly the same?
either both or none or am i wrong?

RalfJung:
Done. My question above still stands.

rotanid:
ah, i guess the problem is, that the RDNSS sent by the nodes or by the supernodes shouldn't differ.
So maybe a better explanation could be: if you want to use this feature, the nodes/supernodes shouldn't announce something *different* than the next-node adresse: either nothing or next-node ?

RalfJung:
Like this?

christf:
@rotanid yes, they must not differ. that is the point.

christf:
@RalfJung I like it.

rotanid:
ok for me too, and as it is "only" document and can't break anything, i'll merge it ;)

PR #1274, opened 2017-12-07
ramips-rt305x: add D-Link DIR-615 rev D1-D4

ambassador86:
We have an ISP around Chemnitz who gave out these routers years ago. We got a request to support these for freifunk. Here it is: http://map.chemnitz.freifunk.net/#!v:m;n:002401c1a16c


rotanid:
thanks!

PR #1276, opened 2017-12-12
set alias to raspberry-pi-model-b-rev-2

2tata:

``` bash
Downloading 'http://autoupdate-lede.ffnw/stable/stable.manifest'
Connecting to 2a01:4f8:150:50a2::107:80
Writing to stdout

-                    100% |*******************************| 61344   0:00:00 ETA
Download completed (61344 bytes)
No matching firmware found (model raspberry-pi-model-b-rev-2)
No usable mirror found.
```

### DIFF
--- a/targets/brcm2708-bcm2708
+++ b/targets/brcm2708-bcm2708
@@ -1,3 +1,4 @@
 device raspberry-pi rpi
+manifest_alias raspberry-pi-model-b-rev-2
 factory -ext4-sdcard .img.gz
 sysupgrade -ext4-sdcard .img.gz
--- a/targets/brcm2708-bcm2709
+++ b/targets/brcm2708-bcm2709
@@ -1,3 +1,4 @@
 device raspberry-pi-2 rpi-2
+manifest_alias raspberry-pi-2-model-b-rev-1.1
 factory -ext4-sdcard .img.gz
 sysupgrade -ext4-sdcard .img.gz


NeoRaider:
I think it's better to use *manifest_alias* here instead of *alias*, no need to actually create symlinks.

Looking at `/lib/brcm2708.sh`, there are a lot more possible names... this might even warrant switching to *board_name* instead of *model* as basis for the brcm2708 image names eventually. (See also: related thread on the lede-dev ML regarding image naming. Maybe we will soon have a more robust solution for choosing the correct image, when that discussion has come to a conclusion.)

2tata:
On 12/13/17 00:20, Matthias Schiffer wrote:
> I think it's better to use *manifest_alias* here instead of *alias*, no need to actually create symlinks.
> 
> Looking at `/lib/brcm2708.sh`, there are a lot more possible names... this might even warrant switching to *board_name* instead of *model* as basis for the brcm2708 image names eventually. (See also: related thread on the lede-dev ML regarding image naming. Maybe we will soon have a more robust solution for choosing the correct image, when that discussion has come to a conclusion.)
> 

Ay, thanks for that information. I have change it to *manifest_alias*

cheers
Tarek

2tata:
This MR is more a small hot-fix. It should be replace if a proper solution in LEDE is implemented.

NeoRaider:
Merged with proper commit messages.

PR #1278, opened 2017-12-14
Update README.md

Rotzbua:
### Changes
updated links from http to https

### Reason
websites seems to support https now

### Backwards incompatibility?
no


rubo77:
@rotzbua
Site the online version also allows setting up a commit message. Just change the line "update readme" next time.

This time we already made a new commit with a correct messages message:

    docs: README.md: switch to https URLs

PR #1279, opened 2017-12-14
ar71xx: add support for TP-Link TL-WR1043N v5

### LINKED ISSUE #1260: support for TP-Link TL-WR1043N v5
there's a new revision of the WR1043 router.

the device TL-WR1043N v5 is very similar to the TL-WR1043ND v4.
the differences are:
- no detachable antennas
- no USB port

there's already a LEDE Pull Request by @Brother-Lal which has undergone a first review, should get accepted soon.
https://github.com/lede-project/source/pull/1506

a backport shouldn't be too hard.

as this device is very common in Freifunk networks (the main use case for Gluon) and there's a lack of affordable devices that are supported by Gluon and available to buy at the moment, i added this issue to the 2017.2 milestone.

blocktrron:
This adds support for TL-WR1043N v5.

closes #1260

### DIFF
--- a/patches/lede/0045-ar71xx-add-support-for-TP-Link-TL-WR1043N-v5.patch
+++ b/patches/lede/0045-ar71xx-add-support-for-TP-Link-TL-WR1043N-v5.patch
@@ -1,0 +1,367 @@
+From: Tim Thorpe <tim@tfthorpe.net>
+Date: Mon, 25 Sep 2017 04:38:49 -0500
+Subject: ar71xx: add support for TP-Link TL-WR1043N v5
+
+TP-Link TL-WR1043N v5 appears to be identical to the TL-WR1043ND v4,
+except that the USB port has been removed and there is no longer a
+removable antenna option.
+
+The software is more in line with the Archer series in that it uses a
+nested bootloader scheme.
+
+Specifications:
+
+ - QCA9563 at 775 MHz
+ - 64 MB RAM
+ - 16 MB flash
+ - 3 (non-detachable) Antennas / 450 Mbit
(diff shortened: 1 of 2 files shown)


rotanid:
two small things:
1) your commit message is missing the N
2) maybe we should name it tp-link-tl-wr1043n-v5 and not tp-link-tl-wr1043n-nd-v5 ? but i'm not really sure about that

blocktrron:
As far as i know the TL-WR1043 was never available with non-detachable Antennas before v5. So in my opinion we should continue in the current scheme. Altering it would additionally possibly break firmware-download wizards, etc.

rotanid:
@blocktrron i merged your PR. maybe you could try to cherry-pick this commit to the v2017.1.x branch? if this works as well, we can backport it

H4ndl3:
@blocktrron @rotanid I'd suggest using `tp-link-tl-wr1043n-v5` as target name in order to avoid confusion. That's the name of the router and people are used to the naming scheme with N = non detachable antennas and ND = detachable antennas. If that somehow breaks any download wizards then they have to be patched. Using a technically incorrect target name just for backwards compatibility is not something I'd suggest.

belzebub40k:
@H4ndl3 Images for versions 1-4 where also called `n-nd` but never hat models with none detachable antennas.

```
device tp-link-tl-wr1043n-nd-v4 tl-wr1043nd-v4
device tp-link-tl-wr1043n-nd-v5 tl-wr1043n-v5
```

H4ndl3:
@belzebub40k thanks for the hint, I guess it should stay that way then.

PR #1280, opened 2017-12-18
libbatadv: Add common batman-adv helper functions library

ecsv:
Interacting with batman-adv's genl interface requires some code and definitions which could be shared between different packages. libbatadv is trying to do this without providing any guarantee for ABI or API stability. It is only useful in very controlled environments like gluon.

It may be useful for

* gluon-mesh-batman-adv
* gluon-status-page-api
* gluon-radv-filterd (#838)


ecsv:
You might want to drop the last two patches when gluon-radv-filterd will not use the two functions (for example when they use my conversion to netlink).

christf:
I like the approach - adding a lib to wrap some complexity and reduce duplicate code.
As it stands, this PR adds more code than it removes, but I do not see this as a problem - there is stil some code that could benefit from it in the status page that will be migrated into gluon-mesh-batman-adv-core from #1053

package/libbatadv/src/batadv-genl.h
```diff
@@ -0,0 +1,110 @@
...
+#include <netlink/netlink.h>
+#include <netlink/genl/genl.h>
+#include <netlink/genl/ctrl.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdbool.h>
+
+#include "batman_adv.h"
+
+struct ether_addr;
```
NeoRaider: It seems more straightforward to me to just include `net/ethernet.h` instead of using a forward declaration.

package/libbatadv/src/batadv-genl.h
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * ARRAY_SIZE() - Get number of items in static array
+ * @x: array with known length
+ *
+ * Return:  number of items in array
+ */
+#ifndef ARRAY_SIZE
+#define ARRAY_SIZE(x) (sizeof(x) / sizeof(*(x)))
+#endif
```
NeoRaider: I'm not happy with defining a macro with such a generic in a library header file. We should either give it a proper namespace prefix, or keep the duplication in the files using it.

package/libbatadv/src/batadv-genl.h
```diff
@@ -0,0 +1,110 @@
...
+ * @type: type of the structure containing ptr
+ * @member: name of the member variable in struct @type
+ *
+ * Return: @type pointer of object containing ptr
+ */
+#ifndef container_of
+#define container_of(ptr, type, member) __extension__ ({ \
+	const __typeof__(((type *)0)->member) *__pmember = (ptr); \
+	(type *)((char *)__pmember - offsetof(type, member)); })
+#endif
```
NeoRaider: Dito.

package/libbatadv/src/batadv-genl.c
```diff
@@ -0,0 +1,444 @@
...
+#include <netlink/netlink.h>
+#include <netlink/genl/genl.h>
+#include <netlink/genl/ctrl.h>
+#include <net/ethernet.h>
+
+#include "batman_adv.h"
+
+#ifndef __maybe_unused
+#define __maybe_unused __attribute__((unused))
+#endif
```
NeoRaider: Defining macros in the compiler/libc implementation-internal namespace (`__` prefix) may be acceptable in the kernel, which is basically its own libc, but I would advise against it in normal userspace code.

ecsv: Ok, will be dropped

package/libbatadv/src/batadv-genl.h
ecsv: Will drop that with the translate-mac + get-tq patches

package/libbatadv/src/batadv-genl.h
ecsv: Ok, both will get a batadv_ prefix

ecsv:
@NeoRaider changes were now pushed old version can still be found in https://github.com/FreifunkVogtland/gluon/tree/libbatadv-orig

@christf The number of code lines  decreased but the number of comment lines increased. The helpers for radv-filterd were dropped and more efficient versions were proposed for radv-filterd (which use the internal structures of radv-filterd). The biggest increase is the new uapi header from the batman-adv kernel module - but this is mostly due to the way the kernel-doc is stored for each attribute/cmd

PR #1285, opened 2017-12-26
gluon-core: Firewall: allow Multicast Listener Query/Report/Done on mesh and client zone

### LINKED ISSUE #1282: gluon-respondd: no updates after first contact
At FFDA, we are experiencing our nodes on current Gluon master not sending respondd messages to our map backend. Nodes will send one message and won't answer on any other request. 

They appear to be offline on map while the status-page is still reachable and mesh links are also drawn.

Last known working commit currently is f7f659c2541011fb406243abbaaa16f0baedba74.

christf:
this will allow more icmp6 trafic in the mesh an local-client zone.
This is meant to fix #1282

### DIFF
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/140-firewall-rules
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/140-firewall-rules
@@ -15,7 +15,6 @@ end
 uci:foreach('firewall', 'zone', reject_input_on_wan)
 
 -- the client zone is set up by gluon-client-bridge
---
 uci:section('firewall', 'zone', 'mesh', {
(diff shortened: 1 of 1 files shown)


mweinelt:
@christf I only tested it with `130/0` and you seem to be missing the `/0` part on every numeric list item. I basically copied that syntax from the `wan` zone.

@T-X Is there a need for MLDv2 reports as well?

rotanid:
update from real-life & IRC: @neoraider is already working on this, too.

NeoRaider:
I've pushed a similar fix in 39284f00d0f309c8b0131f72dd2e4e11e4e2ab4a.

PR #1286, opened 2017-12-27
gluon-core: save firewall when setting dns config

christf:
When changing the firewall, it is sensible to save them in the same script.


NeoRaider:
There is already such a line just below the location you add it.

PR #1287, opened 2017-12-27
add current unix time to respondd statistics

christf:
This will add the current time to the respondd-provided data.

This will make sure we can detect from outside if a node has issues synchronizing its time.

### DIFF
--- a/package/gluon-respondd/src/respondd.c
+++ b/package/gluon-respondd/src/respondd.c
@@ -34,6 +34,7 @@
 #include <stdio.h>
 #include <string.h>
 #include <unistd.h>
+#include <time.h>
 
 #include <sys/vfs.h>
 
@@ -211,10 +212,23 @@ static struct json_object * get_rootfs_usage(void) {
 	return jso;
 }
 
+static struct json_object * get_time(void) {
+	struct timespec now;
+
+	if (clock_gettime(CLOCK_REALTIME, &now) != 0)
+		return NULL;
+
+	return json_object_new_int64(now.tv_sec);
+}
+
 static struct json_object * respondd_provider_statistics(void) {
 	struct json_object *ret = json_object_new_object();
 
 	json_object_object_add(ret, "node_id", gluonutil_wrap_and_free_string(gluonutil_get_node_id()));
+
+	json_object *time = get_time();
+	if (time != NULL)
+		json_object_object_add(ret, "time", time);
 
 	json_object_object_add(ret, "rootfs_usage", get_rootfs_usage());
 	json_object_object_add(ret, "memory", get_memory());


package/gluon-respondd/src/respondd.c
```diff
@@ -211,11 +212,21 @@ static struct json_object * get_rootfs_usage(void) {
 	return jso;
 }
 
+static struct json_object * get_time(void) {
+	struct timespec now;
+	clock_gettime(CLOCK_REALTIME, &now);
```
lemoer: Maybe EINVAL could happen here. I think EFAULT and EPERM are definitely no problem.

christf: yes, EINVAL could happen, but in that case nothing really breaks. The time will be invalid of course but that is fine, I guess....

christf: actually... when EINVAL happens, the field should not exist in the json. This I could do. 

Is that what you meant, @lemoer?

NeoRaider: We should just not add a time when *clock_gettime* fails (adding NULL would be fine, it shouldn't actually happen anyways).

lemoer: I think this is better. Another way would be to initialize, so it is not leaking any random stuff.

package/gluon-respondd/src/respondd.c
```diff
@@ -211,11 +212,21 @@ static struct json_object * get_rootfs_usage(void) {
...
+
+	struct json_object *jso = json_object_new_int(now.tv_sec);
```
NeoRaider: *json_object_new_int64*, for Y2038 support.

package/gluon-respondd/src/respondd.c
```diff
@@ -211,11 +212,21 @@ static struct json_object * get_rootfs_usage(void) {
...
+
+	return jso;
```
NeoRaider: No need to use a variable instead of returning it directly.

christf:
Feedback incorporated.

lemoer:
ACK. Vote for merge ;)

rotanid:
merged with slightly adjusted commit message

package/gluon-respondd/src/respondd.c
```diff
@@ -211,11 +212,24 @@ static struct json_object * get_rootfs_usage(void) {
...
+}
+
 static struct json_object * respondd_provider_statistics(void) {
 	struct json_object *ret = json_object_new_object();
 
 	json_object_object_add(ret, "node_id", gluonutil_wrap_and_free_string(gluonutil_get_node_id()));
 
+	json_object *time = get_time();
+	if (time != NULL)
+		json_object_object_add(ret, "time", time);
```
NeoRaider: We could do without the NULL check, as *json_object_object_add* would just add a JSON *null* value. All data consumers need to be able to deal with broken data anyways, and it's not an error that should actually ever happen.

PR #1288, opened 2017-12-28
Allow multiple names for nextnode - breaks current site.conf

christf:
This will allow us to specify more than one name to resolve to the nextnode-IPs

### DIFF
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -163,6 +163,7 @@ next_node \: package
     ::
 
       next_node = {
+        name = { 'nextnode.location.community.example.org', 'nextnode', 'nn' },
         ip4 = '10.23.42.1',
         ip6 = 'fdca:ffee:babe:1::1',
         mac = '16:41:95:40:f7:dc'
@@ -172,7 +173,8 @@ next_node \: package
     omitted, there will be no IPv4 or IPv6 anycast address. The MAC address
     defaults to ``16:41:95:40:f7:dc``; this value usually doesn't need to be
     changed, but it can be adjusted to match existing deployments that use a
-    different value.
+    different value. Each entry in the ``name``-field will be resolved as the
+    ipv4 and ipv6-address.
 
 mesh \: optional
     Options specific to routing protocols.
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -173,8 +173,13 @@ next_node \: package
     omitted, there will be no IPv4 or IPv6 anycast address. The MAC address
     defaults to ``16:41:95:40:f7:dc``; this value usually doesn't need to be
     changed, but it can be adjusted to match existing deployments that use a
-    different value. Each entry in the ``name``-field will be resolved as the
+    different value. Each entry in the ``name``-field will be resolved to the
     ipv4 and ipv6-address.
+    For this to work, clients must use the next-node as their resolver. In
+    batman-based networks this requires setting the central dhcp server to
+    deliver this address as DNS-server via DHCP-option. When running a radvd
+    inside the network, this should be set to deliver the next-node
+    ipv6-address via rdnss.
 
 mesh \: optional
     Options specific to routing protocols.
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -174,12 +174,12 @@ next_node \: package
     defaults to ``16:41:95:40:f7:dc``; this value usually doesn't need to be
     changed, but it can be adjusted to match existing deployments that use a
     different value. Each entry in the ``name``-field will be resolved to the
-    ipv4 and ipv6-address.
+    IPv4 and IPv6-address.
     For this to work, clients must use the next-node as their resolver. In
-    batman-based networks this requires setting the central dhcp server to
-    deliver this address as DNS-server via DHCP-option. When running a radvd
+    batman-based networks this requires setting the central DHCP server to
+    deliver this address as DNS server via DHCP option. When running a radvd
     inside the network, this should be set to deliver the next-node
-    ipv6-address via rdnss.
+    IPv6-address via rdnss.
 
 mesh \: optional
     Options specific to routing protocols.
--- a/package/gluon-core/check_site.lua
+++ b/package/gluon-core/check_site.lua
@@ -70,6 +70,12 @@ if need_table({'dns'}, nil, false) then
 	need_number({'dns', 'cacheentries'}, false)
 end
 
+if need_table('next_node', nil, false) then
+	need_string_match('next_node.ip6', '^[%x:]+$', false)
+	need_string_match('next_node.ip4', '^%d+.%d+.%d+.%d+$', false)
+	need_string_array('next_node.name', false)
+end
+
 need_string_match(in_domain({'next_node', 'ip6'}), '^[%x:]+$', false)
 need_string_match(in_domain({'next_node', 'ip4'}), '^%d+.%d+.%d+.%d+$', false)
 
--- a/package/gluon-core/check_site.lua
+++ b/package/gluon-core/check_site.lua
@@ -70,12 +70,7 @@ if need_table({'dns'}, nil, false) then
 	need_number({'dns', 'cacheentries'}, false)
 end
 
-if need_table('next_node', nil, false) then
-	need_string_match('next_node.ip6', '^[%x:]+$', false)
-	need_string_match('next_node.ip4', '^%d+.%d+.%d+.%d+$', false)
-	need_string_array('next_node.name', false)
-end
-
+need_string_array(in_domain({'next_node', 'name'}), false)
 need_string_match(in_domain({'next_node', 'ip6'}), '^[%x:]+$', false)
 need_string_match(in_domain({'next_node', 'ip4'}), '^%d+.%d+.%d+.%d+$', false)
 
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
@@ -33,10 +33,8 @@ local function set_dns_record(name, ip, sectionname)
 	})
 end
 
-uci:foreach("dhcp", "domain", function(s)
-	if (s['.name'] and string.match(s['.name'], "^nextnode[46]_")) then
-		uci:delete("dhcp", s['.name'])
-	end
+uci:delete_all('dhcp', 'domain', function(s)
+	return (s['.name'] and string.match(s['.name'], "^nextnode[46]"))
 end)
 
 for i, name in ipairs(next_node.name or {}) do
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
@@ -34,10 +34,8 @@ local function set_dns_record(name, ip, sectionname)
 end
 
 uci:foreach("dhcp", "domain", function(s)
-	for key, value in pairs(s) do
-		if (key == ".name" and string.gmatch(value, "^nextnode[46]")) then
-			uci:delete("dhcp", value)
-		end
+	if (s['.name'] and string.match(s['.name'], "^nextnode[46]_")) then
+		uci:delete("dhcp", s['.name'])
 	end
 end)
 
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
@@ -25,22 +25,25 @@ if dns.servers then
 	})
 end
 
-if next_node.name and next_node.ip4 then
-	uci:section('dhcp', 'domain', 'nextnode4', {
-		name = next_node.name,
-		ip = next_node.ip4,
+local function set_dns_record(name, ip, sectionname)
+	if not ip then return end
+	uci:section('dhcp', 'domain', sectionname, {
+		name = name,
+		ip = ip,
 	})
-else
-	uci:delete('dhcp', 'domain', 'nextnode4')
 end
 
-if next_node.name and next_node.ip6 then
-	uci:section('dhcp', 'domain', 'nextnode6', {
-		name = next_node.name,
-		ip = next_node.ip6,
-	})
-else
-	uci:delete('dhcp', 'domain', 'nextnode6')
+uci:foreach("dhcp", "domain", function(s)
+	for key, value in pairs(s) do
+		if (key == ".name" and string.gmatch(value, "^nextnode[46]")) then
+			uci:delete("dhcp", value)
+		end
+	end
+end)
+
+for i, name in ipairs(next_node.name or {}) do
+	set_dns_record(name, next_node.ip4, 'nextnode4_' .. i)
+	set_dns_record(name, next_node.ip6, 'nextnode6_' .. i)
 end
 
 uci:save('dhcp')
--- a/docs/site-example/site.conf
+++ b/docs/site-example/site.conf
@@ -83,6 +83,7 @@
   -- connected to using a known IP address.
   next_node = {
     -- anycast IPs of all nodes
+    -- name = { 'nextnode.location.community.example.org', 'nextnode', 'nn' },
     ip4 = '10.xxx.0.xxx',
     ip6 = 'fdxx:xxxx:xxxx::xxxx',
   },


package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
```diff
@@ -25,22 +25,37 @@ if dns.servers then
 	})
 end
 
+
+local function set_dns_record(name, host)
+		uci:section('dhcp', 'domain', 'nextnode4_' .. i, {
```
christf: the name should be a parameter to this function as well.

package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
```diff
@@ -25,22 +25,37 @@ if dns.servers then
...
+			name = name,
+			ip = host,
+		})
+end
+
+-- remove nextnode4_* and nextnode6_*
+uci:foreach("dhcp", "domain", function(s)
+	for key, value in pairs(s) do
+		if (key == ".name" and string.gmatch(value, "nextnode[46].*"))
```
christf: this seems to delete more than intended. investigate.

rotanid:
maybe not use "community.net" but something like "community.example.org" (RFC2606) ?

christf:
sure.

docs/user/site.rst
```diff
@@ -163,6 +163,7 @@ next_node \: package
     ::
 
       next_node = {
+        name = { 'nextnode.location.community.example.org', 'nextnode', 'nn' }
```
rotanid: missing comma at the end of the line

package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
```diff
@@ -25,22 +25,33 @@ if dns.servers then
...
+
 if next_node.name and next_node.ip4 then
-	uci:section('dhcp', 'domain', 'nextnode4', {
-		name = next_node.name,
-		ip = next_node.ip4,
-	})
-else
-	uci:delete('dhcp', 'domain', 'nextnode4')
+	for i,n in ipairs(next_node.name) do
+		set_dns_record(n, next_node.ip4, 'nextnode4_' .. i);
```
rotanid: semicolon at the end of the line not necessary

package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
```diff
@@ -25,22 +25,33 @@ if dns.servers then
...
 
 if next_node.name and next_node.ip6 then
-	uci:section('dhcp', 'domain', 'nextnode6', {
-		name = next_node.name,
-		ip = next_node.ip6,
-	})
-else
-	uci:delete('dhcp', 'domain', 'nextnode6')
+	for i,n in ipairs(next_node.name) do
+		set_dns_record(n, next_node.ip6, 'nextnode6_' .. i);
```
rotanid: semicolon at the end of the line not necessary

christf: done

package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
christf: done

package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
```diff
@@ -25,22 +25,33 @@ if dns.servers then
...
+		})
+end
+
+-- remove nextnode4_* and nextnode6_*
+uci:foreach("dhcp", "domain", function(s)
+	for key, value in pairs(s) do
+		if (key == ".name" and string.gmatch(value, "nextnode[46].*")) then
+			uci:delete("dhcp", value)
+		end
+	end
```
NeoRaider:
```lua
if string.match(s['.name'], '^nextnode[46]') then
```
- No need to loop over the pairs of *s*
- `^` anchor was missing (and `.*` was redundant)
- Use *match* instead of *gmatch*, was we're only interested in the first match

The comment does not really match the code: the code doesn't care about the underscore, the comment should be adjusted.

package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
```diff
@@ -25,22 +25,33 @@ if dns.servers then
...
+end)
+
 if next_node.name and next_node.ip4 then
-	uci:section('dhcp', 'domain', 'nextnode4', {
-		name = next_node.name,
-		ip = next_node.ip4,
-	})
-else
-	uci:delete('dhcp', 'domain', 'nextnode4')
+	for i,n in ipairs(next_node.name) do
```
NeoRaider: Missing space after comma.

package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
```diff
@@ -25,22 +25,33 @@ if dns.servers then
...
 end
 
 if next_node.name and next_node.ip6 then
-	uci:section('dhcp', 'domain', 'nextnode6', {
-		name = next_node.name,
-		ip = next_node.ip6,
-	})
-else
-	uci:delete('dhcp', 'domain', 'nextnode6')
+	for i,n in ipairs(next_node.name) do
```
NeoRaider: Dito.

package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
```diff
@@ -25,22 +25,33 @@ if dns.servers then
 	})
 end
 
+
+local function set_dns_record(name, host, sectionname)
+		uci:section('dhcp', 'domain', sectionname, {
+			name = name,
+			ip = host,
+		})
```
NeoRaider: Too deep indentation.

package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
```diff
@@ -25,22 +25,33 @@ if dns.servers then
...
 if next_node.name and next_node.ip6 then
-	uci:section('dhcp', 'domain', 'nextnode6', {
-		name = next_node.name,
-		ip = next_node.ip6,
-	})
-else
-	uci:delete('dhcp', 'domain', 'nextnode6')
+	for i,n in ipairs(next_node.name) do
+		set_dns_record(n, next_node.ip6, 'nextnode6_' .. i)
+	end
```
NeoRaider: How about using a single loop for IPv4 and IPv6, just guarded by
```lua
if next_node.name then
```
? The ip4/ip6 checks could be moved into *set_dns_record* as `if not host`

package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
christf: done.

package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
christf: name was not checked at all before because it is not mandatory I guess. We could introduce a check that, if it is there checks if it is a table.

package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
christf: done

package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
christf: done

package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
christf: done

package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
christf: host would be set in the ipv6 and in the ipv4 case, how would that help to distinguis? Also that would mean setting the section would have to move to yet another function or be duplicated due to the section name containing nextnoce4 or 6 respectively. Frankly I like this two-loop-approach more because I find it clearer. It may be a matter of taste though and if you really like the one-loop approach, I will adapt.

NeoRaider: Site checks are required even for optional fields, they are supposed to ensure that our update scripts never fail.

`host` is not set in both cases, the host field holds the IP address; this is confusing, lets rename it. I propose this code (untested):
```lua
local function set_dns_record(name, ip, sectionname)
    if not ip then return end
    uci:section('dhcp', 'domain', sectionname, {
        name = name,
        ip = ip,
    })
end

for i, name in ipairs(next_node.name or {}) do
    set_dns_record(name, next_node.ip4, 'nextnode4_' .. i)
    set_dns_record(name, next_node.ip6, 'nextnode6_' .. i)
end
```

package/gluon-core/check_site.lua
```diff
@@ -48,6 +48,7 @@ end
 if need_table('next_node', nil, false) then
 	need_string_match('next_node.ip6', '^[%x:]+$', false)
 	need_string_match('next_node.ip4', '^%d+.%d+.%d+.%d+$', false)
+	need_table('next_node.name', nil, false)
```
NeoRaider: This should be *need_string_array*.

rotanid:
@christf in case you didn't see, there's still an open review comment:
https://github.com/freifunk-gluon/gluon/pull/1288#discussion_r160487999

christf:
testing....

NeoRaider:
@christf What's the state of this?

rotanid:
@NeoRaider he updated this PR yesterday but only mentioned that on IRC

NeoRaider:
But the PR is not updated...

rotanid:
huh? the set_dns_record stuff was added yesterday as far as i can see... maybe you should specify more clearly what you want and @christf should add a comment after updating the PR regarding what he changed

NeoRaider:
Is there a problem with the code snippet I suggested? It would lead to much shorter code than what we have now.

christf:
Sorry for the delay. There is no problem with the suggested code. I just wanted to see this built and tested.

package/gluon-core/check_site.lua
```diff
@@ -70,6 +70,12 @@ if need_table({'dns'}, nil, false) then
 	need_number({'dns', 'cacheentries'}, false)
 end
 
+if need_table('next_node', nil, false) then
+	need_string_match('next_node.ip6', '^[%x:]+$', false)
+	need_string_match('next_node.ip4', '^%d+.%d+.%d+.%d+$', false)
+	need_string_array('next_node.name', false)
+end
```
NeoRaider: You have duplicated two lines from below. The `need_string_array` is not adjusted to the new check_site functions (and maybe we should validate the domain names at least a bit). The `need_table` is redundant now.

package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
```diff
@@ -25,22 +25,25 @@ if dns.servers then
...
-		ip = next_node.ip6,
-	})
-else
-	uci:delete('dhcp', 'domain', 'nextnode6')
+uci:foreach("dhcp", "domain", function(s)
+	for key, value in pairs(s) do
+		if (key == ".name" and string.gmatch(value, "^nextnode[46]")) then
+			uci:delete("dhcp", value)
+		end
+	end
```
NeoRaider: I think I commented before that
* the loop can be removed, just access `s['.name']` directly
* it's better to use *match* instead of *gmatch*

docs/user/site.rst
```diff
@@ -172,7 +173,8 @@ next_node \: package
     omitted, there will be no IPv4 or IPv6 anycast address. The MAC address
     defaults to ``16:41:95:40:f7:dc``; this value usually doesn't need to be
     changed, but it can be adjusted to match existing deployments that use a
-    different value.
+    different value. Each entry in the ``name``-field will be resolved as the
+    ipv4 and ipv6-address.
```
NeoRaider: * "will be resolved to the IPv4 and the IPv6 address"
* This needs more explanation (that this only does something when the clients use the node as resolver)

christf: done.

package/gluon-core/check_site.lua
christf: yes it isn't this was done before the merge of the multi-domain stuff. It is adjusted now.

christf: validation is not yet done... not sure what would be a good way. There is this https://stackoverflow.com/questions/106179/regular-expression-to-match-dns-hostname-or-ip-address which certainly seems like overkill.

NeoRaider: Lua patterns don't support operators on groups anyways, so something like `^[a-z0-9%-%.]+$` is probably the best we can easily do.

package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
christf: this is done.

christf:
The changes are pushed as fixup commit and meant to be squashed into one. That way you can see my adjustments separately.

docs/user/site.rst
```diff
@@ -172,7 +173,13 @@ next_node \: package
...
     defaults to ``16:41:95:40:f7:dc``; this value usually doesn't need to be
     changed, but it can be adjusted to match existing deployments that use a
-    different value.
+    different value. Each entry in the ``name``-field will be resolved to the
+    ipv4 and ipv6-address.
+    For this to work, clients must use the next-node as their resolver. In
+    batman-based networks this requires setting the central dhcp server to
+    deliver this address as DNS-server via DHCP-option. When running a radvd
+    inside the network, this should be set to deliver the next-node
+    ipv6-address via rdnss.
```
NeoRaider: There are still a few issues in the above text:
* IPv4, IPv6, DHCP should be capitalized consistently
* Many incorrect colons, should be "IPv6 address", "DNS server", "DHCP option"

package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
```diff
@@ -25,22 +25,23 @@ if dns.servers then
...
 
-if next_node.name and next_node.ip6 then
-	uci:section('dhcp', 'domain', 'nextnode6', {
-		name = next_node.name,
-		ip = next_node.ip6,
-	})
-else
-	uci:delete('dhcp', 'domain', 'nextnode6')
+uci:foreach("dhcp", "domain", function(s)
+	if (s['.name'] and string.match(s['.name'], "^nextnode[46]_")) then
```
NeoRaider: Remove the underscore from the match, so this will also remove old "nextnode4"/"nextnode6" sections.

package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
```diff
@@ -25,22 +25,23 @@ if dns.servers then
...
+		uci:delete("dhcp", s['.name'])
+	end
```
NeoRaider: Hmm, somehow one of my comments disappeared when submitting the review...

Deleting sections in a `uci:foreach` loop is not allowed and may lead to incorrect behaviour. Use `uci:delete_all` in its three-argument form (see `package/gluon-mesh-vpn-fastd/luasrc/lib/gluon/upgrade/400-mesh-vpn-fastd` for an example).

NeoRaider: I just looked at the libuci code, and deleting sections in a foreach loop should be safe after all. I vaguely remember a similar problem from a few years back, but it might have been a different issue. Switching to `uci:delete_all` would still make the code a bit shorter.

NeoRaider: It seems my first comment was right after all: I still get reproducible segfaults when I try to delete sections in a `uci:foreach` call.

docs/user/site.rst
christf: adjusted the capitalisation. for colons I need some help... I always put them in english text like salt on french fries....

package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
christf: done.

package/gluon-core/luasrc/lib/gluon/upgrade/820-dns-config
christf: Interestingly this code worked for me. Still I adjusted it.

NeoRaider:
Merged.

PR #1294, opened 2018-01-03
ebtables: Enable locking for concurrent ebtables processes

ecsv:
These changes are to enable locking to avoid problems when multiple ebtables processes try to modify the kernel data structures at the same time. Not using a lock can result in not applied modifications (add/delete) in the kernel.

During one of my tests, such a concurrency problem resulted in a node which was suppressing all RADVs. The reason for that was an empty RADV_FILTER. This could be fixed by restarting gluon-radv-filterd (when no other ebtables process was running at the same time).

These are required for the PRs #838 and #1113.


NeoRaider:
Rebased and applied.

PR #1296, opened 2018-01-06
mt7621: Add support for ZBT-WG3526-32M

CodeFetch:
Depends on #1300.

As I don't have one of those devices, please consider this as untested.
Unfortunately I grasped a similar patch exists in #962 ~which might even be identical~ too late. 

The general flash issue (32MB) was discussed [here](https://gist.github.com/dmke/d389bc364b3f73f525076eaee0019dc1).
The resulting patch was following:
http://lists.infradead.org/pipermail/linux-mtd/2016-December/070889.html

[Another patch](https://github.com/lede-project/source/commit/22d982ea003360d86ff1fef5114fb41dfa09ce9b#diff-3835f027e16e6416090c1833bc732af3) by nbd168 added the required mtd read operations for supporting ZBT-WG3526-32M and Digineo AC1200 Pro.

~This patch hopefully also resolves the [mt76](https://github.com/openwrt/mt76/issues/62) issue by bumping it to a new version.~ (Patch split - see #1300)

I'm going to clean up the patch files better when it's tested and working as expected.

PR #1298, opened 2018-01-07
build: introduce GLUON_WLAN_MESH build variable

### LINKED ISSUE #1242: Devices that don't support IBSS shouldn't be marked as BROKEN
Currently four targets (ralink/mediatek) are marked as broken with the remark:
> BROKEN: No AP+IBSS support

This is not necessarily anything that will ever be resolved.

Since 802.11s is working quite well and seems to be the successor of IBSS those devices are hardly broken, but only useful if the site already uses 802.11s mesh networking. In turn these could be marked conditional with a flag like `MESH_11s` instead of `BROKEN`.

2tata:
List additional targets if GLUON_WLAN_MESH=11s (default is 11s)

Closes #1242

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,11 +35,12 @@ $(GLUON_SITEDIR)/site.mk:
 
 include $(GLUON_SITEDIR)/site.mk
 
+GLUON_WLAN_MESH ?=11s
 
 GLUON_RELEASE ?= $(error GLUON_RELEASE not set. GLUON_RELEASE can be set in site.mk or on the command line)
 
 
-export GLUON_RELEASE GLUON_ATH10K_MESH GLUON_REGION GLUON_DEBUG
+export GLUON_RELEASE GLUON_WLAN_MESH GLUON_REGION GLUON_DEBUG
 
 show-release:
 	@echo '$(GLUON_RELEASE)'
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ $(GLUON_SITEDIR)/site.mk:
 
 include $(GLUON_SITEDIR)/site.mk
 
-GLUON_WLAN_MESH ?=11s
+GLUON_WLAN_MESH ?= 11s
 
 GLUON_RELEASE ?= $(error GLUON_RELEASE not set. GLUON_RELEASE can be set in site.mk or on the command line)
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -209,7 +209,7 @@ ar71xx-generic
   - My Net N750
 
 .. [#ath10k]
-  Device uses the ath10k WLAN driver; image is built for 11s by default unless GLUON_WLAN_MESH
+  Device uses the ath10k WLAN driver; images are built for 11s by default unless GLUON_WLAN_MESH
   is set as described in :ref:`getting-started-make-variables`
 
 ar71xx-nand
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -209,7 +209,7 @@ ar71xx-generic
   - My Net N750
 
 .. [#ath10k]
-  Device uses the ath10k WLAN driver; no image is built unless GLUON_ATH10K_MESH
+  Device uses the ath10k WLAN driver; image is built for 11s by default unless GLUON_WLAN_MESH
   is set as described in :ref:`getting-started-make-variables`
 
 ar71xx-nand
--- a/docs/user/getting_started.rst
+++ b/docs/user/getting_started.rst
@@ -160,9 +160,9 @@ GLUON_WLAN_MESH
   meshing in the same image due to WLAN firmware restrictions.
 
   Setting GLUON_MESH_MESH to ``11s`` or ``ibss`` will enable generation of images for ath10k devices
-  and install the firmware for the corresponding WLAN mode. Especially setting to ``11s`` will also enable
-  the generation of images for targets they're able to run only AP with 11s, like ralink or mediatek.
-  Defaults to ``11s``.
+  and install the firmware for the corresponding WLAN mode. Specifically setting to ``11s`` will also
+  enable the generation of images for targets that are able to run only AP+11s but not AP+IBSS, like
+  ralink or mediatek. Defaults to ``11s``.
 
 GLUON_BRANCH
   Sets the default branch of the autoupdater. If unset, the autoupdater is disabled
--- a/docs/user/getting_started.rst
+++ b/docs/user/getting_started.rst
@@ -156,14 +156,13 @@ Common variables
 ................
 
 GLUON_WLAN_MESH
-  While Gluon does support some hardware with ath10k-based 5GHz WLAN, these WLAN adapters don't work
-  well for meshing at the moment, so building images for these models is disabled by default. In addition,
-  ath10k can't support IBSS and 11s meshing in the same image due to WLAN firmware restrictions.
+  While Gluon does support some hardware with ath10k-based 5GHz WLAN. ath10k can't support IBSS and 11s
+  meshing in the same image due to WLAN firmware restrictions.
 
   Setting GLUON_MESH_MESH to ``11s`` or ``ibss`` will enable generation of images for ath10k devices
-  and install the firmware for the corresponding WLAN mode.
-
-  Setting GLUON_MESH_MESH to ``11s`` will aso list targets their able to run only AP + 11s.
+  and install the firmware for the corresponding WLAN mode. Especially setting to ``11s`` will also enable
+  the generation of images for targets they're able to run only AP with 11s, like ralink or mediatek.
+  Defaults to ``11s``.
 
 GLUON_BRANCH
   Sets the default branch of the autoupdater. If unset, the autoupdater is disabled
--- a/docs/user/getting_started.rst
+++ b/docs/user/getting_started.rst
@@ -155,13 +155,15 @@ usually be set on the command line or in ``site.mk``.
 Common variables
 ................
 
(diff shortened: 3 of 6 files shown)


targets/targets.mk
```diff
@@ -9,13 +9,16 @@ $(eval $(call GluonTarget,x86,generic))
 $(eval $(call GluonTarget,x86,geode))
 $(eval $(call GluonTarget,x86,64))
 
+ifeq ($(GLUON_WLAN_MESH),11s)
```
kb-light: I would expect that these targets are also build in case `BROKEN` is set (concerning `GLUON_WLAN_MESH = ibss`).

docs/user/getting_started.rst
```diff
@@ -155,14 +155,16 @@ usually be set on the command line or in ``site.mk``.
...
+GLUON_WLAN_MESH
   While Gluon does support some hardware with ath10k-based 5GHz WLAN, these WLAN adapters don't work
   well for meshing at the moment, so building images for these models is disabled by default. In addition,
   ath10k can't support IBSS and 11s meshing in the same image due to WLAN firmware restrictions.
 
-  Setting GLUON_ATH10K_MESH to ``11s`` or ``ibss`` will enable generation of images for ath10k devices
+  Setting GLUON_MESH_MESH to ``11s`` or ``ibss`` will enable generation of images for ath10k devices
   and install the firmware for the corresponding WLAN mode.
 
+  Setting GLUON_MESH_MESH to ``11s`` will aso list targets their able to run only AP + 11s.
```
lemoer: I think this text should be a little more generic now. Maybe removing the references to ath10k is sufficient. Or as an alternative, you could add other examples ralink/mediatek...

Makefile
```diff
@@ -35,11 +35,12 @@ $(GLUON_SITEDIR)/site.mk:
 
 include $(GLUON_SITEDIR)/site.mk
 
+GLUON_WLAN_MESH ?=11s
```
rotanid: missing space after the equality sign

docs/user/getting_started.rst
rotanid: i agree, the text should reflect the other targets as well, removing the text about ath10k is not necessary, as that content still applies. (this and lemoers comment affect this whole paragraph, not only the last line)
also, this last line has some english spelling/grammar issues. additionally, i'm not sure what "list" should mean in this context.

targets/targets.mk
rotanid: i second that, there should be an OR (pseudocode: if GLUON_WLAN_MESH = 11s OR BROKEN = 1 )

targets/targets.mk
```diff
@@ -10,12 +10,20 @@ $(eval $(call GluonTarget,x86,geode))
 $(eval $(call GluonTarget,x86,64))
 
 ifneq ($(BROKEN),)
+  AP_11S=1
```
rotanid: - the indention of this variable is different from the other lines
- the naming might introduce conflicts. maybe: GLUON_TARGET_11S_ONLY

rotanid:
i just noticed another problem:
what about devices like Ubnt ER-X ? It should build with IBSS and without BROKEN, because it has no wifi

docs/user/getting_started.rst
2tata: I change the text above also a bit, because the part about the default state was not correct.

rotanid:
thanks for your changes, i noticed some additional smaller issue with one of the changes.
(for those who saw my previous comment: it was crap, sorry )

targets/targets.mk
2tata: ACK

docs/user/getting_started.rst
```diff
@@ -155,13 +155,14 @@ usually be set on the command line or in ``site.mk``.
...
-
-  Setting GLUON_ATH10K_MESH to ``11s`` or ``ibss`` will enable generation of images for ath10k devices
-  and install the firmware for the corresponding WLAN mode.
+GLUON_WLAN_MESH
+  While Gluon does support some hardware with ath10k-based 5GHz WLAN. ath10k can't support IBSS and 11s
+  meshing in the same image due to WLAN firmware restrictions.
+
+  Setting GLUON_MESH_MESH to ``11s`` or ``ibss`` will enable generation of images for ath10k devices
+  and install the firmware for the corresponding WLAN mode. Especially setting to ``11s`` will also enable
+  the generation of images for targets they're able to run only AP with 11s, like ralink or mediatek.
```
rotanid: "Especially" -> "Specifically"
"they're" -> "that are"
"only AP with 11s" -> "only AP+11s but not AP+IBSS"

docs/index.rst
```diff
@@ -209,7 +209,7 @@ ar71xx-generic
   - My Net N750
 
 .. [#ath10k]
-  Device uses the ath10k WLAN driver; no image is built unless GLUON_ATH10K_MESH
+  Device uses the ath10k WLAN driver; image is built for 11s by default unless GLUON_WLAN_MESH
```
NeoRaider: "images are"

targets/targets.mk
```diff
@@ -10,12 +10,20 @@ $(eval $(call GluonTarget,x86,geode))
...
+ifeq ($(GLUON_WLAN_MESH),11s)
+GLUON_TARGET_11S_ONLY=1
+endif
+
+ifneq ($(GLUON_TARGET_11S_ONLY),)
 $(eval $(call GluonTarget,ramips,mt7620)) # BROKEN: No AP+IBSS support
 $(eval $(call GluonTarget,ramips,mt7628)) # BROKEN: No AP+IBSS support
 $(eval $(call GluonTarget,ramips,rt305x)) # BROKEN: No AP+IBSS support
-$(eval $(call GluonTarget,sunxi)) # BROKEN: Untested
 endif
```
NeoRaider: I'm not a fan of these code changes. I would do something like the following (code untested):
```
GLUON_WLAN_MESH_11s := $(filter 11s,$(GLUON_WLAN_MESH))
```
on the top, and
```
ifneq ($(GLUON_WLAN_MESH_11s)$(BROKEN),)
```
at the actual use site.

* The "BROKEN: No AP+IBSS support" comments must be removed.
* targets.mk should be reorderd so that the BROKEN devices are on the bottom again.
* We also need to make similar changes in the ramips-mt7621 target definition, where we currently only build the ER-X image without BROKEN
* Many devices of the ramips targets need a "BROKEN: untested", which was omitted so far, as the whole target was BROKEN. Maybe we should mark all devices of these targes as BROKEN for now and then remove the BROKEN check after we get test feedback.

rotanid: @NeoRaider 
ramips-mt7621: why would we need a change here? i though so at first, too - but ramips-mt7621 isn't broken anymore, please read the comments in there again - or am i wrong?

BROKEN: untested: i already use one or two of those, i can say for ramips-mt7628 vocore2 that it does work nicely.

NeoRaider: d-link-dir-860l-b1 and zbt-wg3526 have separate BROKEN guards. These should be moved under a 11s||BROKEN guard eventually as well if they work okay.

rotanid: their BROKEN guard is because of the flakey 2,4GHz wifi. this won't be fixed by this PR, but perhaps by updating mt76, see #1296

2tata:
On 01/08/18 09:49, Matthias Schiffer wrote:
> NeoRaider requested changes on this pull request.
> 
> 
> 
>> @@ -209,7 +209,7 @@ ar71xx-generic
>    - My Net N750
>  
>  .. [#ath10k]
> -  Device uses the ath10k WLAN driver; no image is built unless GLUON_ATH10K_MESH
> +  Device uses the ath10k WLAN driver; image is built for 11s by default unless GLUON_WLAN_MESH
> 
> "images are"
> 
>>  $(eval $(call GluonTarget,ramips,mt7620)) # BROKEN: No AP+IBSS support
>  $(eval $(call GluonTarget,ramips,mt7628)) # BROKEN: No AP+IBSS support
>  $(eval $(call GluonTarget,ramips,rt305x)) # BROKEN: No AP+IBSS support
> -$(eval $(call GluonTarget,sunxi)) # BROKEN: Untested
>  endif
> 
> I'm not a fan of these code changes. I would do something like the following (code untested):
> ```
> GLUON_WLAN_MESH_11s := $(filter 11s,$(GLUON_WLAN_MESH))
> ```
> on the top, and
> ```
> ifneq ($(GLUON_WLAN_MESH_11s)$(BROKEN),)
> ```
> at the actual use site.

Applied

> * The "BROKEN: No AP+IBSS support" comments must be removed.

Why shall I remove the comments? I understand it like the AP+11s targets
should also listed under BROKEN because of the comments reason?

> * targets.mk should be reorderd so that the BROKEN devices are on the bottom again.

Done

> * We also need to make similar changes in the ramips-mt7621 target definition, where we currently only build the ER-X image without BROKEN

You mean for following routers?
dir-860l-b1
zbt-wg3526

I didnt did that because of the comment for high package lost in using 11s.

cheers
Tarek

NeoRaider:
@2tata Please don't reply to Github reviews via mail, it makes it impossible to add further comments to the existing threads in the web interface.

A "BROKEN" comment means a device does not work correctly. We are currently redefining what this means, so calling these devices BROKEN is not accurate anymore; the comments must be removed for all devices that can be built without setting BROKEN.

Regarding the mt7621 devices: Okay, let's keep them BROKEN for now.

rotanid:
@2tata i have merged your code by squashing and using an adjusted commit message. i hope that's fine.

PR #1300, opened 2018-01-08
mt76: Bump version to 2017-12-17

CodeFetch:
Bumps mt76 to a new version while keeping the old mac80211 API.
This commit should fix WLAN LEDs and HT20 instabilities on e.g. ZBT-WG3526 and D-Link DIR-860L.


rotanid:
@CodeFetch needs rebase, change of LEDE patch numbering, because of dce388fdc7ede816a6eaa1753184939c66d45ec4

mweinelt:
My DIR860L unfortunately didn't even survive more than a day.

https://meshviewer.darmstadt.freifunk.net/#/de/map/908d7823af1c

rotanid:
@mweinelt i just noticed your device has a different chip, MT7062E whereas the ZBT-WG3526 has MT7603E - the changes mentioned the latter?
also, the TL-WR841N v13 also using mt76 has another chip, MT7628, which got different driver changes

heini66:
my zbt-wg3626 looks ( until now ) good for me ... 
http://map.freifunk-nordheide.de/#!v:m;n:78a3511367f0

heini66:
might be the different hardware at 2,4 be the problem why my 2,4ghz sucks on plain 2017.1.4 sucks and the keeping of the old mac80211 api isn't required?
sorry, unable to code it to test it. but if someone creates a patch, i'll give it a try on my zbt ...

CodeFetch:
@heini66 The API needs to be the same as from the kernel. Only the names of parameters were changed. There were no new parameters introduced or something...

rotanid:
@heini66 we are also testing this patch on ZBT-WG3526-16M and it looks good.
i achieved up to 90 Mbit/s via 2.4GHz with iperf3, average 58 Mbit/s
i guess i will build again without any workaround/quickfix to test long term stability.

rotanid:
the stability wasn't really good.
there were many improvements in the meantime but i don't have the time to update the backport.

PR #1000, opened 2016-12-30
add gluon-dns-config: a package that enables the next node to be used as dns cache

christf:
as dns server any server in the network can be used. for example an anycast-ip-address where dns is served ie on all gateways.

PR #1312, opened 2018-01-17
gluon-client-bridge: move IPv4 local subnet route to br-client

T-X:
This patch moves the prefix4 subnet route from the local-node veth
device to br-client (while keeping the next node ipv4 address on the
local node device).

This is in preparation to allow routing over the br-client interface
later.

### DIFF
--- a/package/gluon-client-bridge/luasrc/lib/gluon/upgrade/310-gluon-client-bridge-local-node
+++ b/package/gluon-client-bridge/luasrc/lib/gluon/upgrade/310-gluon-client-bridge-local-node
@@ -23,8 +23,7 @@ uci:section('network', 'device', 'local_node_dev', {
 local ip4, ip6
 
 if next_node.ip4 then
-	local plen = site.prefix4():match('/%d+$')
-	ip4 = next_node.ip4 .. plen
+	ip4 = next_node.ip4 .. '/32'
 end
 
 if next_node.ip6 then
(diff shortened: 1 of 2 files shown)


NeoRaider:
Please explain the usecase for this better.

T-X:
Moved this out of #1094 as I think this is a nice unification, too. That is, configure the next-node address in the same style for both IPv4 and IPv6.

Also, would be nice having this patch merged already to ease the usage and testing of #1094 via an external package. (this diff is the only one from #1094 affecting existing Gluon packages, everything else in #1094 are only additions via and within the new gluon-alt-esc-provider and gluon-alt-esc-client packages)

T-X:
Although, wait a minute. I haven't checked yet whether with the latest firewall refactoring the next node address still works with this diff.

CodeFetch:
@T-X I am not sure if the configuration for babel will allow it this way. The current plan is quite complex and not yet completely worked out. The firewall will most likely need another minor refactoring...

T-X:
Tested: next node address still works with Gluon master and this diff.

T-X:
@CodeFetch: Hm, this diff was proposed 8 months ago in #1094. And during that time that PR has had several rounds of updates and testing now.

Would it make sense to discuss how #1094 and 464xlat (which is what you are working on?) could be integrated together after there is a plan and prototype to look at for 464xlat? It is a little difficult to discuss like this.

T-X:
But I'm confident that #1094 and 464xlat are not mutually exclusive, if that is what you are worried about, @CodeFetch. In fact, I would love to be able to use 464xlat in a batman-adv network, too - while at the same time enabling a #1094 alt-esc-provider on the same node.

CodeFetch:
I haven't thought about that before. That's a great idea...

ecsv:
This PR broke the IPv4 anycast address. The DNS forwarder from #1000 is not working anymore correctly and the access to the status page via IPv4 anycast address is also broken. More details can be found in #1488

PR #1314, opened 2018-01-20
gluon-mesh-vpn-fastd: add public key to nodeinfo provider

mweinelt:
We are using the fastd exporter (https://github.com/freifunk-darmstadt/fastd-exporter) for Prometheus statistics and since the only stable key to link the graphs with other respondd data is the public key I propose adding it to the nodeinfo response.

Gotcha: The code is largely duplicated from the `get_fastd_version` function above it.

### DIFF
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -198,6 +198,12 @@ mesh_vpn
     defines the MTU of the VPN interface, determining a proper MTU value is described
     in the :ref:`FAQ <faq-mtu>`.
 
+    By default information that could be used to associate client traffic with a nodes
+    IP address is not advertised to protect the nodes privacy. This usually requires
+    the attacker to be able to observe the link over which the tunnel flows.
+    If this is of no concern in your threat-model this behaviour can be disabled by
+    setting *pubkey_privacy* to `false`.
+
     The `fastd` section configures settings specific to the *fastd* VPN
     implementation.
 
--- a/package/gluon-mesh-vpn-core/check_site.lua
+++ b/package/gluon-mesh-vpn-core/check_site.lua
@@ -1,5 +1,6 @@
 need_boolean(in_site({'mesh_vpn', 'enabled'}), false)
 need_number({'mesh_vpn', 'mtu'})
+need_boolean(in_site({'mesh_vpn', 'pubkey_privacy'}), false)
 
 need_boolean(in_site({'mesh_vpn', 'bandwidth_limit', 'enabled'}), false)
 need_number(in_site({'mesh_vpn', 'bandwidth_limit', 'ingress'}), false)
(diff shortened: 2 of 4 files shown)


mweinelt:
By default if fastd was not enabled the key is set to:
```
fastd.mesh_vpn.secret='generate'
```

Calling `/etc/init.d/fastd show_key mesh_vpn` will generate a secret.

Is this an issue?

NeoRaider:
@mweinelt The code should work similar to the info page in the "Advanced settings".

Hmm, I thought there was already an issue for this addition, but I can't find it right now. I don't really like the idea of adding the public key, as it makes it easier to correlate nodes and VPN sessions "from the outside", which might be undesirable because of privacy concerns.

rotanid:
> which might be undesirable because of privacy concerns
also thought this to be the reason it is hard to get this correlation at the moment. 
the thing is though, it should be easy at least for the operator of the vpn gateway. maybe there is a trick so at least these operators can correlate easier?

mweinelt:
Pubkey is now being advertised on the condition that the mesh-vpn fastd instance is enabled.

Via https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-web-admin/files/lib/gluon/web/view/admin/info.html#L16-L17

rotanid:
i would really like to know how this would affect the privacy, a real world example with explicitly naming the potential "attacker" and what would be gained and how

mweinelt:
@rotanid From my (probably limited) understanding of the issue, here's my take on things:

- The node is uniquely identifiable by it's mac address.
- The public key is always part of the fastd handshake.

Putting both together client traffic and tunnel traffic can be correlated and the traffic flow can then, probably by a nation state attacker, be attributed to the actual IP address of the node.

NeoRaider:
I don't think we should add Gluon-specific fields to non-Gluon UCI sections. It would be better to just get the value directly from the site in the respondd provider.

mweinelt:
Updated to not use UCI and use the value from the site instead, defaulting to `true`.

T-X:
I'm probably missing some basic cryptography knowledge, but concerning the node correlation:

Would it be an option to encrypt the handshake messages a fastd peer A sends to B with the public key of B? That way, only who has the private key of B is able to proof that the message was encrypted to B?

(what also came to my mind, probably for another ticket: should we enable the IPv6 privacy extension on br-wan? Otherwise we probably leak a node's MAC address via the IPv6 header of a fastd packet anyways)

NeoRaider:
@T-X 
Might be possible, but likely fatal for handshake performance, and the attack scenario is unclear. Also, I don't intend to change the handshake protocol of fastd any time soon.

Privacy extension support should have a way to opt-out, and is AFAIK not supported by netifd/odhcp6c at all at the moment.

NeoRaider:
Merged (I changed the doc part a bit).

rotanid:
here's the link to the merged commit for future reference: 42763d21dc6854e720cd2d8a58ae5fbf5b444f73

PR #1318, opened 2018-01-26
target files cleanup: variable checks and comments

rotanid:
this removes a check that was obsoleted and updates/cleans up the BROKEN comments

### DIFF
--- a/targets/ar71xx-generic
+++ b/targets/ar71xx-generic
@@ -1,9 +1,5 @@
-ATH10K_PACKAGES=
+ATH10K_PACKAGES='-kmod-ath10k kmod-ath10k-ct'
 ATH10K_PACKAGES_QCA9887=
-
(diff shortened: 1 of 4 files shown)


targets/ar71xx-generic
```diff
@@ -1,8 +1,6 @@
-ATH10K_PACKAGES=
-ATH10K_PACKAGES_QCA9887=
```
NeoRaider: I would like these variables always to be set, even when GLUON_WLAN_MESH has an invalid value. We could move the definitions from the "if 11s" to the top and only keep the "if ibss".

targets/ipq806x
```diff
@@ -1,5 +1,3 @@
-ATH10K_PACKAGES=
```
NeoRaider: Please keep the initialization, all variables should explicitly be set to an empty string.

rotanid:
@NeoRaider i updated the PR according to your suggestions

PR #1323, opened 2018-02-05
gluon-core: fix issue with legacy mesh on mesh_lan

### LINKED ISSUE #1322: gluon-reconfigure leads to deactivation of VXLAN
### Situation
A Node upgrading from gluon v2017.1.5 to current Gluon master leaves legacy-mesh mode on LAN/WAN enabled. 

After the domain is switched (from config mode or CLI + gluon-reconfigure), legacy mesh is used and the device is using VXLAN for wired mesh. 

Changing advanced settings from the Wizard Tab does not seem to be a desired impact.

kb-light:
When running gluon-reconfigure (and not upgrading from 2017.1.x) the removed line leads to legacy=0.
fixes #1322

### DIFF
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/220-interface-lan
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/220-interface-lan
@@ -17,7 +17,6 @@ uci:section('network', 'interface', 'mesh_lan', {
 	igmp_snooping = false,
 	proto         = 'gluon_wired',
 	index         = 4,
-	legacy        = old_proto == 'gluon_mesh',
 })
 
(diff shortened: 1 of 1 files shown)


NeoRaider:
Your fix is not correct, legacy must still be set to `old_proto == 'gluon_mesh'` if it doesn't have a value at all, otherwise legacy will be unset (thus disabled) on updates from v2017.1.x.

kb-light:
@NeoRaider isn't that done by line 53ff? (see 210-interface-wan for reference)

NeoRaider:
Indeed, you are right.

Issue #680, opened 2016-03-09
batman-adv crash when removing interface in certain configurations


belzebub40k:
When I configure more then 2 VLANs via UCI for mesh-connections and add them to bat0 the network stack isn't working correctly any longer. Any invocation of ip, ifconfig, brctl, batctl etc will result in a stuck system.
If I configure just two mesh-VLANs the system is working fine. It is also possible to add more VLANs to bat0 via "vconfig eth0.x" and "batctl if add eth0.x".

The strange part is before the crash happens the kernel always prints "tried to remove device eth0 from br-client" but eth0 itself shouldn't be part of any bridge or whatever. I also don't see this message when I configure just two mesh-VLANs.

See attached files for backtrace and /etc/config/network.
If you need further information please let me know.

EDIT:
- Tested with v2016.1.1 and master.
- Problem occurs on two pieces of hardware.

[dmesg.txt](https://github.com/freifunk-gluon/gluon/files/165051/dmesg.txt)
[network_config.txt](https://github.com/freifunk-gluon/gluon/files/165050/network_config.txt)

NeoRaider:
Interesting, I'll try to reproduce this with other hardware.

Which batman-adv version are you using, compat 14 or 15?

belzebub40k:
root@0216088335f8:/# batctl -v
batctl 2016.0 [batman-adv: 2016.0]

As far as I understand this refers to compat 15.

belzebub40k:
It looks like this problem is some how related to br-client. If I remove br-client completely from /etc/config/network I'm able to configure 4 mesh-VLANs (didn't try more).

However this isn't really an option as this device is intended to be a backbone node with connects several Ubiquiti devices running stock firmware. We have a management VLAN (3) on each of them which is then bridged with bat0 via br-client. Without this we wouldn't be able to access them from remote.

NeoRaider:
I've forwarded the issue to the batman-adv developers, they'll have a look.

belzebub40k:
Thanks. Is there a ticket I can track and contribute to?

I just found some more details that may be helpful:
I can leave br-client enabled but need to remove bat0 from it. And again it wont harm the system if I add it manually once everything is up.

NeoRaider:
I think there is no ticket yet, I'll open one.

I wasn't able to reproduce this issue on x86-vmware with the current Gluon master.

[...]

PR #1325, opened 2018-02-06
sunxi: LeMaker Banana Pi: remove BROKEN

belzebub40k:
Since the crashes described in https://github.com/freifunk-gluon/gluon/issues/680 are fixed I tested all three Banana Pi Models and everything looks fine. As the WiFi chip of the Pro model is not supported and IBSS/11s is not working with the R1 Model they are still marked as BROKEN.

### DIFF
--- a/targets/sunxi
+++ b/targets/sunxi
@@ -3,6 +3,13 @@ sysupgrade -ext4-sdcard .img.gz
 
 
 device lemaker-banana-pi sun7i-a20-bananapi
+
+if [ "$BROKEN" ]; then
+
+# WiFi chip not supported
 device lemaker-banana-pro sun7i-a20-bananapro
 
+# AP+IBSS or 11s not working
 device lamobo-r1 sun7i-a20-lamobo-r1
+
+fi
--- a/targets/targets.mk
+++ b/targets/targets.mk
@@ -7,6 +7,7 @@ $(eval $(call GluonTarget,brcm2708,bcm2708))
 $(eval $(call GluonTarget,brcm2708,bcm2709))
 $(eval $(call GluonTarget,mpc85xx,generic))
 $(eval $(call GluonTarget,ramips,mt7621))
+$(eval $(call GluonTarget,sunxi))
 $(eval $(call GluonTarget,x86,generic))
 $(eval $(call GluonTarget,x86,geode))
 $(eval $(call GluonTarget,x86,64))
@@ -22,5 +23,4 @@ $(eval $(call GluonTarget,ar71xx,mikrotik)) # BROKEN: no sysupgrade support
 $(eval $(call GluonTarget,brcm2708,bcm2710)) # BROKEN: Untested
 $(eval $(call GluonTarget,ipq806x)) # BROKEN: Untested
 $(eval $(call GluonTarget,mvebu)) # BROKEN: No AP+IBSS or 11s support
-$(eval $(call GluonTarget,sunxi)) # BROKEN: Untested
 endif


ctr49:
@belzebub40k Is there a particular reason why the Pro and R1 have been labeled BROKEN just because the built-in Wifi is not supported? Afaik the RPi3 also uses brcmfmac with a binary blob and is just marked broken for being untested. Previous RPis don't have WiFi at all and are not marked as broken.

I think many people use such Pis with an USB attached external WiFi adapter, so imho the presence or status of the built-in WiFi module shouldn't have an impact to the build status...

belzebub40k:
@ctr49 Apart from the not working WiFi interfaces I did not notice any problems. But AFAIK all devices that have WiFi need to support at least AP+802.11s. And as USB dongles aren't supported by Gluon anyway I don't see a reason to remove the BROKEN flag.

mweinelt:
USB dongles could be supported if they would be any good. Usually they are single radio (sometimes dualband) devices that don't support mesh at all. Weird chipsets from vendors like realtek whose drivers aren't mainline anyway.

ctr49:
@belzebub40k where does that definition come from? Why does Gluon have Raspberry Pi and Banana Pi builds at all then? Does it make a difference if a device doesn't have wifi or the built-in Wifi is just not supported (yet)?
Can you please elaborate on "USB dongles aren't supported by Gluon anyway"? Installing the necessary modules via opkg worked fine for me so far... I wasn't suggesting to add all possible usb-wifi kmods into the build, (although usb-core wouldn't hurt and technically there might be enough space as indicated by mweinelt).

@mweinelt yes there are good and bad devices. Specifically ALFA (and similar) devices are quite popular though. I've specifically got a TP-Link TL-WN7200ND adapter because the used rt2800-usb (Ralink) driver shouldn't be too bad: It is in mainline already and supports IBSS (probably no 802.1s though).

belzebub40k:
AP+802.11s is required according to the [Device Integration checklist](https://github.com/freifunk-gluon/gluon/wiki/Device-Integration-checklist). When I said _"USB dongles aren't supported by Gluon"_ I meant they are **not officially** support by the Gluon project. I can't give you any text that stats this but it was the answer I received from @NeoRaider when I asked a similar question.

ctr49:
According to the checklist I have to assume that WiFi is "not applicable" for devices not having WiFi. The same statement could be used for devices where WiFi is not (yet) supported though.

ctr49:
That means in contrast to RPi1/2 the RPi3 will also stay BROKEN unless brcmfmac is supported? I wouldn't hold my breath...

belzebub40k:
From a user perspective I would expect that a device which has a wifi chip and is supported by gluon has working wifi.

blocktrron:
@ctr49 Correctly. If you are an experienced user you can simply build yourself an image.

BROKEN is exactly for such cases.

ctr49:
We are turning in circles here. I understand your point. But no one has answered the question why there are "supported" devices that have no WiFi but devices that have WiFi which is not supposed to be used cannot be supported.

Scenario 1)
No one (except me ) is really planning to use such a device for Gluon. Then the change wouldn't hurt. This is very unlikely though, see scenario 2.

Scenario 2)
People want to use the Banana Pro / RPi 3 like me. If you look at the forums there are several threads about running FF on top of a RPi. RPi3 will have the same problem as the Banana Pro (it has Wifi, but apparently the bcrmfmac is not supported in Gluon)

Scenario 3)
People expect that Banana Pro and Raspi would be working *with built-in* WiFi and ask questions why it doesn't work. This can simply be solved by documenting the limitation. If there is enough demand someone may even work on getting it working.

I cannot see a scenario where removing broken would cause harm other than unavailability of something that is even more unavailable while it is marked as broken.

I know that I can build my own. But this is about using a couple of devices in a Freifunk project. Yes I could also ask my community to build with BROKEN or have a modified target file in the fork. But I cannot believe, that I'm the only one who need Gluon in the main build for a device that has WiFi but that shouldnt be used.
Today's gadget have many hardware features. Is it really the expectation that we only use devices where *every component* of that device (even if I don't want to use it) has to be supported already? Where is the incentive to improve something then?

How about moving it from BROKEN into a $GLUON_WLAN_MESH" = 'ibss' section?
(assuming the wifi generally works but doesn't support mesh) I see a few devices that are using this already...

NeoRaider:
It is planned to move from a single BROKEN flag to a more fine-grained classification of devices (#1432).

Generally, removing a BROKEN flag for a partially supported device is always a question of what users expect from a device. For example, Broadcom WLAN is completely unusable for mesh networking, so we will never support that WLAN chip for anything else than simple AP mode. Will a user seeing that the RPi3 is "supported" by Gluon expect the RPi3 to work as a full-blown mesh node? I don't know...

In any case, we'd like Gluon support status to be a good indication for users to decide on a device to get, which is why we're reluctant to remove the BROKEN flag for devices what have built-in WiFi that is not fully supported. (What kind of WiFi chip does the LeMaker Banana Pi have? If it does not support meshing, we need to revert this PR for now.)

If we had more fine-grained tagging, I don't think there would be a problem with providing images for partially-supported devices by default, as long as the limitations are made clear on the fimware download pages.

ctr49:
The Banana **Pi** doesn't have any WiFi, this is why **this** commit is fine. My request was about the Banana **Pro** which uses the same driver as the RPi3 (bcrmfmac). I'd expect it to work in client and AP mode, possibly IBSS but not as mesh (both for RPi3 and Banana Pro)

PR #1327, opened 2018-02-06
ar71xx: remove bs-partition ro-flag for UniFi AC

blocktrron:
This removes the read-only flag from the bs (bootselect) partition
on UniFi AC devices. This allows to correct the indicator from which
partition the device is booting its kernel from.

See also:
 - freifunk-gluon/gluon#1301
 - https://github.com/openwrt/openwrt/pull/727
 - https://bugs.lede-project.org/index.php?do=details&task_id=662


rotanid:
@NeoRaider this has been merged upstream today, so we can merge it too, i guess?
maybe even backport it to v2017.1.x ?

NeoRaider:
I've backported it to lede-17.01 and updated `modules`, so it will in v2017.1.x as soon as this update is applied there.

Issue #94, opened 2014-04-12
ignored txpower offset on Ubiquiti NanoStation M2


T-X:
It seems that the 12 dBi antenna of the Ubiquiti NanoStation M2 is properly set in the firmware, at least a 12dB txpower offset is noted in iwinfo. However, the txpower is not reduced accordingly:

<pre>
$ iwinfo
wlan0     ESSID: "luebeck.freifunk.net"
          Access Point: 24:A4:3C:AC:F4:AC
          Mode: Master  Channel: 1 (2.412 GHz)
          Tx-Power: 30 dBm  Link Quality: 40/70
          Signal: -70 dBm  Noise: -89 dBm
          Bit Rate: 43.6 MBit/s
          Encryption: none
          Type: nl80211  HW Mode(s): 802.11bgn
          Hardware: 168C:002A 0777:E012 [Ubiquiti NanoStation M2]
          TX power offset: 12 dB
          Frequency offset: none
          Supports VAPs: yes

wlan0-1   ESSID: "02:d1:11:37:fc:38"
          Access Point: 02:D1:11:37:FC:38
          Mode: Ad-Hoc  Channel: 1 (2.412 GHz)
          Tx-Power: 30 dBm  Link Quality: 69/70
          Signal: -41 dBm  Noise: -89 dBm
          Bit Rate: 102.3 MBit/s
          Encryption: unknown
          Type: nl80211  HW Mode(s): 802.11bgn
          Hardware: 168C:002A 0777:E012 [Ubiquiti NanoStation M2]
          TX power offset: 12 dB
          Frequency offset: none
          Supports VAPs: yes
</pre>

The "Tx-Power" field seems to include the tx power offset (https://dev.openwrt.org/browser/branches/attitude_adjustment/package/iwinfo/src/iwinfo_cli.c?rev=36121#L433). However it should say "Tx-Power: 20 dBm".

The reg domain is set correctly, too:

<pre>
$ iw reg get
country DE:
        (2400 - 2483 @ 40), (N/A, 20)
        (5150 - 5250 @ 80), (N/A, 20), NO-OUTDOOR
        (5250 - 5350 @ 80), (N/A, 20), NO-OUTDOOR, DFS
        (5470 - 5725 @ 80), (N/A, 27), DFS
        (57240 - 65880 @ 2160), (N/A, 40), NO-OUTDOOR
</pre>

<pre>
$ iwinfo wlan0-1 txpowerlist
  12 dBm (  15 mW)
[...]

PR #1329, opened 2018-02-12
gluon-core: add outdoor channel support for 5 ghz radios

mweinelt:
```
    Add the `wifi5.outdoor_chanlist` site configuration that
    allows specifying an outdoor channel range that can be
    switched to for regulatory compliance.
    
    Upon enabling the outdoor option the device will
     - configure the `outdoor_chanlist` on all 5 GHz radios
     - which may enable DFS/TPC, based on the regulatory domain
     - disable ibss/mesh on the 5 GHz radio, as DFS *will*
       break mesh connections
     - allow for htmode reconfiguration on 5 GHz radios
    
    The outdoor option can be toggled from
     - Advanced Settings
       - W-LAN
         - Outdoor Installation
    
    The `preserve_channel` flag overrules the outdoor channel
    selection.
```

## Configuring from Config Mode

config mode

## Configuring from CLI

Outdoor disabled
```shell
root@64xxx-f81a67a5f222:~# uci set gluon.wireless.outdoor='0'
root@64xxx-f81a67a5f222:~# /lib/gluon/upgrade/200-wireless 
root@64xxx-f81a67a5f222:~# uci show wireless.radio0.channel; uci show wireless.radio1.channel; uci show wireless.mesh_radio0.disabled
wireless.radio0.channel='44'
wireless.radio1.channel='1'
wireless.mesh_radio0.disabled='0'
root@64xxx-f81a67a5f222:~# wifi
root@64xxx-f81a67a5f222:~# iw dev
phy#1
	Interface client1
		ifindex 18
		wdev 0x100000003
		addr f8:1a:67:a5:f2:21
		ssid darmstadt.freifunk.net
		type AP
		channel 1 (2412 MHz), width: 20 MHz, center1: 2412 MHz
		txpower 20.00 dBm
	Interface mesh1
		ifindex 17
		wdev 0x100000002
		addr ae:02:43:9d:a4:05
		type mesh point
		channel 1 (2412 MHz), width: 20 MHz, center1: 2412 MHz
		txpower 20.00 dBm
phy#0
	Interface client0
		ifindex 16
		wdev 0x3
		addr ae:02:43:9d:a4:00
		ssid darmstadt.freifunk.net
		type AP
		channel 44 (5220 MHz), width: 20 MHz, center1: 5220 MHz
		txpower 20.00 dBm
	Interface mesh0
		ifindex 15
		wdev 0x2
		addr ae:02:43:9d:a4:01
		type mesh point
		channel 44 (5220 MHz), width: 20 MHz, center1: 5220 MHz
		txpower 20.00 dBm
```

Outdoor enabled
```shell
root@64xxx-f81a67a5f222:~# uci set gluon.wireless.outdoor='1'
root@64xxx-f81a67a5f222:~# /lib/gluon/upgrade/200-wireless 
root@64xxx-f81a67a5f222:~# uci show wireless.radio0.channel; uci show wireless.radio1.channel; uci show wireless.mesh_radio0.disabled
wireless.radio0.channel='100'
wireless.radio1.channel='1'
wireless.mesh_radio0.disabled='1'
root@64xxx-f81a67a5f222:~# wifi
root@64xxx-f81a67a5f222:~# iw dev
phy#1
	Interface client1
		ifindex 21
		wdev 0x100000005
		addr f8:1a:67:a5:f2:21
		ssid darmstadt.freifunk.net
		type AP
		channel 1 (2412 MHz), width: 20 MHz, center1: 2412 MHz
		txpower 20.00 dBm
	Interface mesh1
		ifindex 20
		wdev 0x100000004
		addr ae:02:43:9d:a4:05
		type mesh point
		channel 1 (2412 MHz), width: 20 MHz, center1: 2412 MHz
		txpower 20.00 dBm
phy#0
	Interface client0
		ifindex 19
		wdev 0x4
		addr ae:02:43:9d:a4:00
		type AP
		channel 100 (5500 MHz), width: 20 MHz, center1: 5500 MHz
		txpower 22.00 dBm
root@64xxx-f81a67a5f222:~# logread | grep DFS
Mon Feb 12 01:22:49 2018 daemon.notice hostapd: client0: interface state COUNTRY_UPDATE->DFS
Mon Feb 12 01:22:49 2018 daemon.notice hostapd: client0: DFS-CAC-START freq=5500 chan=100 sec_chan=0, width=0, seg0=0, seg1=0, cac_time=60s
Mon Feb 12 01:23:49 2018 daemon.notice hostapd: client0: DFS-CAC-COMPLETED success=1 freq=5500 ht_enabled=1 chan_offset=0 chan_width=1 cf1=5500 cf2=0
Mon Feb 12 01:23:49 2018 daemon.notice hostapd: client0: interface state DFS->ENABLED

```

## TODO

- [x] site validation
  - [x] configurable list of acceptable outdoor channels (ranges and individual channels)
  - [x] channel/chanlist validation
- [x] upgrade scripts
  - [x] auto-enable for obvious 5ghz outdoor devices
- [x] config mode
  - [x] show section in wizard, when enabled
  - [x] enable outdoor mode and configure htmode in advanced settings
  - [x] language strings
- [x] docs


blocktrron:
In addition to this, it would be great to have some devices pre-configured to use outdoor channels. Devices such as Nanostation M5, CPE510, UniFi AP AC Mesh (probably impossible atm. due to wrong board detection).

mweinelt:
Simon Wunderlich's slides on "Dynamic Frequency Selection in 5 GHz mesh networks"
https://www.battlemesh.org/BattleMeshV7/Agenda?action=AttachFile&do=get&target=2014-05-17_wbmv7_DFS.pdf

Intelligent handling of mesh connections is out of scope for this issue.

MPW1412:
As comments were requested in the gluon developer mumble meeting tonight: I would like to see this merged, as we have a lot of Ubiquity AC Mesh Pro devices here in Münsterland, which are running with original firmware so far and could be flashed with Gluon then. For now we install the rocket m2 for mesh and use a dual band access point at the same place. This could be reduced to one dual band device and by this simplifies setups.

Vote fore merge. :)

mweinelt:
- config mode lets you configure htmode per radio now.
- config only shows up when you actually have a 11a/11na/11ac radio

config mode

looking for people willing to test now!

mweinelt:
I tested this for the `DFS-ETSI` regdom, the 5 GHz radio automatically uses an txpower setting according to {regdom, device capability}.

The global regulatory domain seems to be what's used, the per phy value apparently is not.

```
# iw reg get
global
country DE: DFS-ETSI
	(2400 - 2483 @ 40), (N/A, 20), (N/A)
	(5150 - 5250 @ 80), (N/A, 20), (N/A), NO-OUTDOOR, AUTO-BW
	(5250 - 5350 @ 80), (N/A, 20), (0 ms), NO-OUTDOOR, DFS, AUTO-BW
	(5470 - 5725 @ 160), (N/A, 27), (0 ms), DFS
	(5725 - 5875 @ 80), (N/A, 14), (N/A)
	(57000 - 66000 @ 2160), (N/A, 40), (N/A)

phy#1
country US: DFS-FCC
	(2402 - 2472 @ 40), (N/A, 30), (N/A)
	(5170 - 5250 @ 80), (N/A, 23), (N/A), AUTO-BW
	(5250 - 5330 @ 80), (N/A, 23), (0 ms), DFS, AUTO-BW
	(5490 - 5730 @ 160), (N/A, 23), (0 ms), DFS
	(5735 - 5835 @ 80), (N/A, 30), (N/A)
	(57240 - 63720 @ 2160), (N/A, 40), (N/A)

phy#0
country US: DFS-FCC
	(2402 - 2472 @ 40), (N/A, 30), (N/A)
	(5170 - 5250 @ 80), (N/A, 23), (N/A), AUTO-BW
	(5250 - 5330 @ 80), (N/A, 23), (0 ms), DFS, AUTO-BW
	(5490 - 5730 @ 160), (N/A, 23), (0 ms), DFS
	(5735 - 5835 @ 80), (N/A, 30), (N/A)
	(57240 - 63720 @ 2160), (N/A, 40), (N/A)
```

UAP-AC-Mesh
```
# iw dev
[...]
phy#0
	Interface client0
		ifindex 20
		wdev 0x6
		addr ba:f5:60:dd:39:70
		ssid darmstadt.freifunk.net
		type AP
		channel 140 (5700 MHz), width: 20 MHz, center1: 5700 MHz
		txpower 27.00 dBm
```

WDR4900v1
```
# iw dev
[...]
phy#0
	Interface client0
		ifindex 28
		wdev 0x5
		addr ae:02:43:9d:a4:00
		type AP
		channel 120 (5600 MHz), width: 20 MHz, center1: 5600 MHz
		txpower 22.00 dBm
```

mweinelt:
The last commit adds an improved channel and chanlist validation:

wifi24.channel and wifi5.channel now check against an array of valid channels:
```
*** site.conf error: expected wifi24/channel to be one of the given array [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14], but it is 0
```

```
*** site.conf error: expected wifi5/channel to be one of the given array [34, 36, 38, 40, 42, 44, 46, 48, 50, 52, 54, 56, 58, 60, 62, 64, 100, 102, 104, 106, 108, 110, 112, 114, 116, 118, 120, 122, 124, 126, 128, 132, 134, 136, 138, 140, 142, 144, 149, 151, 153, 155, 157, 159, 161, 165, 169, 173], but it is 45
```

Also chanlist validation implements what hostapd describes to be a valid chanlist:

```
# Channel list restriction. This option allows hostapd to select one of the
# provided channels when a channel should be automatically selected.
# Channel list can be provided as range using hyphen ('-') or individual
# channels can be specified by space (' ') separated values
# Default: all channels allowed in selected hw_mode
#chanlist=100 104 108 112 116
#chanlist=1 6 11-13
```

mweinelt:
I currently see a `500 Bad Gateway` after clicking `Save & Restart`. I don't get an error message and haven't found the error yet.

mweinelt:
Model detection for UAP-AC-M is stuck in the openwrt pull request queue (https://github.com/openwrt/openwrt/pull/767) and OpenWrt 18.5 is supposed to be branched in 3 days.

IMO this is the most interesting device for this PR, but the review/merge isn't happening.

MPW1412:
Can we do anything to support you on this? I'd love to see the UAP-AC-M{-Pro,} properly supported, as we're having more and more of those devices. So far they run with stock firmware. With gluon we'd have at least 2,4 GHz mesh.

CodeFetch:
There was a talk on the last battlemesh where they said that DFS in ath9k is broken as it has far too many false-positives (maybe it was a lightning talk as it is not on the schedule).
They said they have implemented DFS in hostapd which works reliable and it was tested in a certified environment. Maybe it was „Opennet Initiative“. I don't know... https://wiki.opennet-initiative.de/wiki/OpenWrt_DFS/Experiment

I oppose merging this unless we are either sure that DFS works correctly (which I doubt) or whoever did that talk releases their work and we confirmed it's working.
It is not okay to merge something into Gluon that makes communities think they can switch to outdoor channels now while there is evidence that it is broken.

Adorfer:
"false positives" are better than "no 5G outdoor at all". 
If somebody can improve the DFS later: why not.
But at least it's start. 
Or do you assume that "not merging" will improve the reliability of dectection significantly/instantly?

CodeFetch:
@Adorfer You are right. I misunderstood what the PR does. As it falls back to indoor channels when no radar-free channel was detected, it is safe to merge this and I didn't see that one can define devices for which outdoor channels can be used before. 
Only the DFS detection in ath9k is broken. Ath10k should be fine.
As far as I remember it can happen that a node continuously detects radar without any radar at all.
Markus said „WDR4900v1 (didn't find a radar-free channel … weird)“. It seems to be the case for this device as mine does not find a radar free channel, too.
Nonetheless I've written an e-mail to the Opennet people hoping that we can fix DFS completely soon.

@mweinelt We have a very strong CW radar in our city. It uses channels 116-132. Can you please change the outdoor channel siteconf definition to be a list instead  so that one can define e.g. outdoor_chanlist = {'100-112', '136', '140'}, add a siteconf check to make sure that the default channel is always an indoor channel and mention this in the docs?

The only problem I still see is when radar was detected once and the node switches to the indoor channel it is unlikely that it ever switches to an outdoor channel again or am I missing something?

mweinelt:
@CodeFetch:

```lua
outdoor_chanlist = '100-112,136,140',
```

CodeFetch:
@mweinelt Oh sorry Martin... Did I get it right the first time that when I activate outdoor channels it does not use the default (indoor) channel anymore? 
Are you sure that it can't happen that a node detects radar on all outdoor channels hence not being reachable for some time at all?
In the talk they said some ath9k devices detect radar very often and some even always. I don't know if it only differs per location, device or chipset, but I'd like to be sure beforehand...

mweinelt:
When you activate outdoor mode it sets:
```
channel=auto
```
That removes the indoor channel from the hostapd configuration

As I described in https://github.com/freifunk-gluon/gluon/pull/1329#issuecomment-369060790 I indeed had the situation that a WDR4900v1 didn't find a radar-free channel and therefore didn't set up the radio. I didn't research this any further.

I've tried this on ath9k/ath10k/mt76, all seem to be working (to some degree?).

2tata:
The Outdoor setting semms not applyed via configmode. Mesh is still on:


root@ffnw-98ded0727ce8:# uci show | grep wireless
gluon.wireless=wireless
gluon.wireless.outdoor='1'
gluon.wireless.outdoor_radio0_htmode='HT40'
gluon-core.@wireless[0]=wireless
wireless.radio0=wifi-device
wireless.radio0.type='mac80211'
wireless.radio0.hwmode='11a'
wireless.radio0.path='platform/ar934x_wmac'
wireless.radio0.htmode='HT20'
wireless.radio0.channel='44'
wireless.radio0.country='DE'
wireless.mesh_radio0=wifi-iface
wireless.mesh_radio0.ifname='mesh0'
wireless.mesh_radio0.network='mesh_radio0'
wireless.mesh_radio0.device='radio0'
wireless.mesh_radio0.mesh_fwding='0'
wireless.mesh_radio0.mesh_id='t-ffnw-mesh_02:00:0A:12:E0:00'
wireless.mesh_radio0.mcast_rate='12000'
wireless.mesh_radio0.mode='mesh'
wireless.mesh_radio0.disabled='0'
wireless.mesh_radio0.macaddr='b6:8e:3f:0d:88:09'
wireless.client_radio0=wifi-iface
wireless.client_radio0.ifname='client0'
wireless.client_radio0.network='client'
wireless.client_radio0.disabled='0'
wireless.client_radio0.device='radio0'
wireless.client_radio0.mode='ap'
wireless.client_radio0.macaddr='b6:8e:3f:0d:88:08'
wireless.client_radio0.ssid='testing.nordwest.freifunk.net'
root@ffnw-98ded0727ce8:~# 

The wifi is still on ch44.
Do I miss something in the configmode?

2tata:
I think we should avoid redundancy config entrys in the configmode like `Device will be installed outdoors` and `Node will be installed outdoors` especially with inconsistent descriptions.

2tata:
After config mode, the CDE510 does not apply the new wireless settings:

```
root@ffnw-98ded0727ce8:~# uci show | grep wireless
gluon.wireless=wireless
gluon.wireless.outdoor_radio0_htmode='HT20'
gluon.wireless.outdoor='1'
gluon-core.@wireless[0]=wireless
wireless.radio0=wifi-device
wireless.radio0.type='mac80211'
wireless.radio0.hwmode='11a'
wireless.radio0.path='platform/ar934x_wmac'
wireless.radio0.htmode='HT20'
wireless.radio0.channel='44'
wireless.radio0.country='DE'
wireless.mesh_radio0=wifi-iface
wireless.mesh_radio0.ifname='mesh0'
wireless.mesh_radio0.network='mesh_radio0'
wireless.mesh_radio0.device='radio0'
wireless.mesh_radio0.mesh_fwding='0'
wireless.mesh_radio0.mesh_id='t-ffnw-mesh_02:00:0A:12:E0:00'
wireless.mesh_radio0.mcast_rate='12000'
wireless.mesh_radio0.mode='mesh'
wireless.mesh_radio0.disabled='0'
wireless.mesh_radio0.macaddr='b6:8e:3f:0d:88:09'
wireless.client_radio0=wifi-iface
wireless.client_radio0.ifname='client0'
wireless.client_radio0.network='client'
wireless.client_radio0.disabled='0'
wireless.client_radio0.device='radio0'
wireless.client_radio0.mode='ap'
wireless.client_radio0.macaddr='b6:8e:3f:0d:88:08'
wireless.client_radio0.ssid='testing.nordwest.freifunk.net'
```
After manually calling `/lib/gluon/upgrade/200-wireless` the wifi settings seems applied.

```
root@ffnw-98ded0727ce8:~# uci show | grep wireless
gluon.wireless=wireless
gluon.wireless.outdoor_radio0_htmode='HT20'
gluon.wireless.outdoor='1'
gluon-core.@wireless[0]=wireless
wireless.radio0=wifi-device
wireless.radio0.type='mac80211'
wireless.radio0.hwmode='11a'
wireless.radio0.path='platform/ar934x_wmac'
wireless.radio0.htmode='HT20'
wireless.radio0.country='DE'
wireless.radio0.channel='auto'
wireless.radio0.channels='100-140'
wireless.client_radio0=wifi-iface
wireless.client_radio0.ifname='client0'
wireless.client_radio0.network='client'
wireless.client_radio0.disabled='0'
wireless.client_radio0.device='radio0'
wireless.client_radio0.mode='ap'
wireless.client_radio0.macaddr='b6:8e:3f:0d:88:08'
wireless.client_radio0.ssid='testing.nordwest.freifunk.net'
wireless.mesh_radio0=wifi-iface
wireless.mesh_radio0.ifname='mesh0'
wireless.mesh_radio0.network='mesh_radio0'
wireless.mesh_radio0.device='radio0'
wireless.mesh_radio0.mesh_fwding='0'
wireless.mesh_radio0.mesh_id='t-ffnw-mesh_02:00:0A:12:E0:00'
wireless.mesh_radio0.mcast_rate='12000'
wireless.mesh_radio0.mode='mesh'
wireless.mesh_radio0.disabled='1'
wireless.mesh_radio0.macaddr='b6:8e:3f:0d:88:09'
```

iwinfo
```
root@ffnw-98ded0727ce8:~# iwinfo 
client0   ESSID: "testing.nordwest.freifunk.net"
          Access Point: B6:8E:3F:0D:88:08
          Mode: Master  Channel: 104 (5.520 GHz)
          Tx-Power: 17 dBm  Link Quality: 60/70
          Signal: -50 dBm  Noise: -95 dBm
          Bit Rate: 115.6 MBit/s
          Encryption: none
          Type: nl80211  HW Mode(s): 802.11an
          Hardware: unknown [Generic MAC80211]
          TX power offset: unknown
          Frequency offset: unknown
          Supports VAPs: yes  PHY name: phy0

```

Adorfer:
just as a wild guess: in initial config mode the /etc/config/wireless is not fully populated, so many keys might not not exists like the radio0-interface, so some of the radio commands may run into errors when creating 
wireless.radio0.channel='auto'
wireless.radio0.channels='100-140'

2tata:
On 07/27/18 17:33, Adorfer wrote:
> just as a wild guess: in initial config mode...

It`s also not work on reenterd config mode.

2tata:
@mweinelt ^

package/gluon-web-wifi-config/luasrc/lib/gluon/config-mode/model/admin/wifi-config.lua
```diff
@@ -1,5 +1,6 @@
 local fs = require 'nixio.fs'
```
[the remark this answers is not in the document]
mweinelt: not mine, rebasing now.

package/gluon-web-wifi-config/luasrc/lib/gluon/config-mode/model/admin/wifi-config.lua
```diff
@@ -25,6 +26,19 @@ local function txpower_list(phy)
 	return new
 end
 
+local function has_5ghz_radio()
```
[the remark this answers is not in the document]
mweinelt: Done.

package/gluon-config-mode-outdoor/i18n/de.po
```diff
@@ -0,0 +1,9 @@
+msgid ""
+"Please enable this option in case the node is to be installed outdoors "
+"to comply with local frequency regulations."
+msgstr ""
+"Wenn der Knoten im Freien aufgestellt werden soll, dann aktiviere bitte "
+"diese Option um den örtlichen Frequenzbestimmungen zu entsprechen."
+
+msgid "Node will be installed outdoors"
+msgstr "Knoten wird im Außenbereich betrieben"
```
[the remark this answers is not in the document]
mweinelt: done

package/gluon-config-mode-outdoor/i18n/gluon-config-mode-outdoor.pot
```diff
@@ -0,0 +1,7 @@
+msgid ""
+"Please enable this option in case the node is to be installed outdoors "
+"to comply with local frequency regulations."
+msgstr ""
+
+msgid "Node will be installed outdoors"
+msgstr ""
```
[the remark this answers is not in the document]
mweinelt: done

package/gluon-web-wifi-config/i18n/de.po
```diff
@@ -49,3 +49,25 @@ msgstr ""
...
+"to stay on the same channel permanently."
+msgstr ""
+"Ist das Gerät für den Einsatz im Freien konfiguriert, wird ein WLAN-Kanal auf "
+"dem 5-GHz-Band sowie eine Sendeleistung entsprechend den gesetzlichen "
+"Frequenzregulatorien gewählt. Gleichzeitig wird die dynamische Frequenzwahl "
+"(DFS; Radarerkennung) aktiviert und die Mesh-Funktionalität deaktiviert, da "
+"sich Nachbarknoten dauerhaft auf demselben Kanal befinden müssen."
+
+msgid "HT Mode"
+msgstr "HT-Modus"
```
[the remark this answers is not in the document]
mweinelt: done

package/gluon-web-wifi-config/i18n/gluon-web-wifi-config.pot
```diff
@@ -33,3 +33,20 @@ msgid ""
...
+msgid ""
+"Configuring the device for outdoor use tunes the 5 GHz radio to a frequency "
+"and transmission power that conforms with the local regulatory requirements. "
+"It also enables dynamic frequency selection (DFS; radar detection). At the "
+"same time, mesh functionality is disabled as it requires neighbouring nodes "
+"to stay on the same channel permanently."
+msgstr ""
+
+msgid "HT Mode"
+msgstr ""
```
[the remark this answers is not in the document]
mweinelt: done

package/gluon-setup-mode/luasrc/lib/gluon/upgrade/350-outdoor-preset
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/lua
```
mweinelt: This is the script that detects outdoor models and sets the preset. It will run unless the outdoor variable was already discovered. 

If autodiscovery fails debug here.

package/gluon-web-wifi-config/i18n/fr.po
```diff
@@ -46,3 +46,21 @@ msgstr ""
...
+msgid "Node will be installed outdoors"
+msgstr "FIXME"
+
+msgid ""
+"Configuring the node for outdoor use tunes the 5 GHz radio to a frequency "
+"and transmission power that conforms with the local regulatory requirements. "
+"It also enables dynamic frequency selection (DFS; radar detection). At the "
+"same time, mesh functionality is disabled as it requires neighbouring nodes "
+"to stay on the same channel permanently."
+msgstr "FIXME"
```
mweinelt: Needs french translation

package/gluon-web-wifi-config/i18n/fr.po
```diff
@@ -46,3 +46,21 @@ msgstr ""
 "<br /><br />Ici vous pouvez aussi configurer la puissance d'émmission se votre Wi-Fi. "
 "Prenez note que les valeurs fournies pour la puissance de transmission prennent "
 "en compte les gains fournis par l'antenne, et que ces valeurs ne sont pas toujours disponibles ou exactes."
+
+msgid "Outdoor installation"
+msgstr "Installation extérieure"
+
+msgid "Node will be installed outdoors"
+msgstr "FIXME"
```
mweinelt: Needs french translation

package/gluon-setup-mode/luasrc/lib/gluon/upgrade/350-outdoor-preset
mweinelt: @Tata Discovery will only happen on first boot of device afair.

package/gluon-web-wifi-config/i18n/fr.po
NeoRaider: Please just leave missing translations empty, so the original string is used. Missing translations are not blockers for PRs.

package/gluon-config-mode-outdoor/i18n/de.po
```diff
@@ -0,0 +1,10 @@
...
+
```
2tata: Sorry I don't mean a newline like that. In the version before there was the line ending broken. Now the line ending seems ok, but the "newline" should be remove.

package/gluon-config-mode-outdoor/i18n/gluon-config-mode-outdoor.pot
```diff
@@ -0,0 +1,8 @@
...
+
```
2tata: same here

package/gluon-config-mode-outdoor/luasrc/lib/gluon/config-mode/wizard/0250-outdoor.lua
```diff
@@ -0,0 +1,23 @@
+
```
2tata: This Option is still redundant. There should be only one checkbox in the config mode. Maybe you can remove this file or replace it with a banner like the autoupdater banner to inform the user wether the outdoor mode is active or not.

mweinelt:
5 GHz radio detection broken due to an early return statement I pushed yesterday(?), issue found by @fooker and @Noodlesalat. Thanks alot!

mweinelt:
Thanks alot, amended and pushed!

package/gluon-config-mode-outdoor/luasrc/lib/gluon/config-mode/wizard/0250-outdoor.lua
2tata: Ok, so than maybe you can remove the config inside the "advanced mode" or at leas the redundant checkboxes.

mweinelt: The rationale is:

- Enable outdoor mode automatically when the model is primarily used outdoors
- Inform users in wizard when outdoor mode is enabled and allow them to disable it
- Allow advanced users to enable outdoor mode, even when it wasn't autodetected
- Allow advanced users to configure the HT mode

ecsv:
Just some small remarks (just to have it somewhere):

freifunk-gluon is also used/tested in non-EU countries (as far as I heard). Some of these countries/regulatory domains (FCC, IC) have special limits for outdoor installations and thus require (on QCA HW) special CTLs. These are part of the board data files (BDFs) which are (for ath10k) are stored in the board.bin/board-2.bin. As far as I can see, these are not handled by this PR.


And there were a couple of problems when we tested ath10k (ipq40xx) in the last months on DFS channels. For example, it continued to transmit non-CSA related frames on the channel when a radar signal was detected and the CSA was already started. Another problem was that a device send probe requests on radar channels even when it was not yet clear that there is no radar on the channel - so the channel was in no mean marked as available. The two patches for that can be found at

* https://patchwork.kernel.org/patch/10545939/
* https://patchwork.kernel.org/patch/10517903/

Another problem is that the ETSI weather radar channels (120-128) require a more ridiculous^W strict CAC, longer CAC, better detection rate and a longer blocking time in case a radar is detected. This is *not* (as far as I know) implemented in the linux kernel. And it vendors also tend to not certify for these channels and instead just block them.

And there is the problem that hostapd didn't stop when it ran out of channels in the chanlist - it just continue to transmit on the *unavailable* channel. You can play around that using "dfs_simulate_radar" in /sys/kernel/debug (actual position depends on the device) when you want to try whether this is also the case with your firmware:

     echo 1 > /sys/kernel/debug/ieee80211/phy0/ath9k/dfs_simulate_radar

And there were some problems with the detection rate on ath9k and earlier ath10k models (QCA9882). The problem on ath9k seemed to be HW related and the ath10k could have been related to something else (we never tried to certify it again).

mweinelt:
One remaining issue seems to be that on some firmware builds outdoor mode doesn't activate after initial configuration from config mode, while `gluon.wireless.outdoor` is still set to `1`. 

A `gluon-reconfigure` reconfigures the device properly and `wifi` applies the correct configuration.

Ideas welcome.

christf:
As discussed on IRC, we need a mechanism that can trigger gluon-reconfigure which is requred when toggling the outdoor flag and when changing the domain. 
When both is done, gluon-reconfigure should execute exactly once.

Noodlesalat:
There were two bugs that are now fixed:
- If the outdoor mode was used in a non-multidomain environment, the outdoor mode settings were not applied to the devices, because the 200-wireless upgrade script was not executed. This is fixed by a "gluon-reconfigure" [here](https://github.com/freifunk-gluon/gluon/pull/1329/commits/631923558c8c5ee2eab2de365a9522ccb10128fe#diff-888cf5bb0760b5b2c235dd8da3900ad3R162)
- When upgrading a node with outdoor mode settings applied, the mesh network got activated even it was disabled in the nodes config. This was fixed by changing `local outdoor_disabled = false` to `local outdoor_disabled = nil` [here](https://github.com/freifunk-gluon/gluon/pull/1329/commits/631923558c8c5ee2eab2de365a9522ccb10128fe#diff-88c2a03cf9e1c43cb50605324d07ba31R188)

I still have concerns about the usage of `gluon-reconfigure`. I don't know if [this](https://github.com/freifunk-gluon/gluon/pull/1329/commits/631923558c8c5ee2eab2de365a9522ccb10128fe#diff-888cf5bb0760b5b2c235dd8da3900ad3R162) line is executed if the outdoor mode is enabled by default ("preset the outdoor mode on suitable devices") and the user just clicks "Save & restart" in the config mode. Imho it is better to execute `gluon-reconfigure` one time after the user clicked "Save & restart".

NeoRaider:
We can't really rely on the "save & restart" button, as the "advanced settings" pages save separately. The user might just do a power cycle or reboot using SSH instead of using the button.

Having a persistent flag (possibly in UCI) triggering the reconfigure early during reboot might be a passable solution.

mweinelt:
Rebased and added preset for Unifi AP AC Mesh Pro.

mweinelt:
I reordered the outdoor preset upgrade script to 180, so it runs before `200-wireless`. That should fix the wireless reconfiguration from preset after the initial setup.

mweinelt:
Both `gluon-config-mode-outdoor` and `gluon-web-wifi-config` now execute `200-wireless` (instead of `gluon-reconfigure` in `gluon-web-wifi-config`'s case). This is a little bit less invasive.

Noodlesalat:
Tested in all imaginable variations. The outdoor-mode works!

package/gluon-config-mode-outdoor/luasrc/lib/gluon/config-mode/wizard/0250-outdoor.lua
```diff
@@ -0,0 +1,28 @@
...
```
NeoRaider: Remove this blank line.

package/gluon-config-mode-outdoor/luasrc/lib/gluon/config-mode/wizard/0250-outdoor.lua
```diff
@@ -0,0 +1,28 @@
...
+	outdoor.default = outdoor_mode
+
+	function outdoor:write(data)
+		if data ~= outdoor_mode then
+			uci:set('gluon', 'wireless', 'outdoor', data)
+			uci:commit('gluon.wireless')
+			os.execute('/lib/gluon/upgrade/200-wireless')
+			uci:commit('wireless')
+		end
+	end
```
NeoRaider: - Instead of commiting the data in outdoor:write, return `{'gluon', 'wireless'}` from the outer function.
- Section-specific commits like `gluon.wireless` are not useful. The file will be written anyways.

package/gluon-web-wifi-config/luasrc/lib/gluon/config-mode/model/admin/wifi-config.lua
```diff
@@ -8,7 +9,6 @@ local function txpower_list(phy)
 	local off  = tonumber(iwinfo.nl80211.txpower_offset(phy)) or 0
 	local new  = { }
 	local prev = -1
-	local _, val
```
NeoRaider: Loop variables are implicitly local.

package/gluon-config-mode-outdoor/luasrc/lib/gluon/config-mode/wizard/0250-outdoor.lua
```diff
@@ -0,0 +1,28 @@
+
+return function(form, uci)
+	local outdoor_mode = uci:get_bool('gluon', 'wireless', 'outdoor')
+
+	if not outdoor_mode then
+		-- only shown when enabled. enabled through preset or web-wifi-config.
+		return
+	end
```
NeoRaider: A checkbox you can only disable, but not enable seems weird. Either always show the checkbox (if the wifi supports 5GHz), or only show a note like the autoupdater.

package/gluon-web-wifi-config/luasrc/lib/gluon/config-mode/model/admin/wifi-config.lua
```diff
@@ -97,7 +108,57 @@ uci:foreach('wireless', 'wifi-device', function(config)
...
+				data = nil
+			end
+			uci:set('gluon', 'wireless', 'outdoor_' .. radio .. '_htmode', data)
+		end
+	end)
+end
+
+
 function f:write()
+	uci:commit('gluon.wireless')
```
NeoRaider: Again, a section-specific commit is not useful

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -49,22 +49,37 @@ if not sysconfig.gluon_version then
...
 
+local function is_outdoor()
+	return uci:get_bool('gluon', 'wireless', 'outdoor')
+end
+
 local function get_channel(radio, config)
-	local channel
 	if uci:get_first('gluon-core', 'wireless', 'preserve_channels') then
-		channel = radio.channel
+		return radio.channel
```
NeoRaider: You must only return here if `radio.channel` is set.

package/gluon-config-mode-outdoor/luasrc/lib/gluon/config-mode/wizard/0250-outdoor.lua
Noodlesalat: The idea is that the checkbox only shows up if the outdoor-mode is enabled. It is meant as a little reminder that the outdoor-mode is enabled (for example if the autodetection system enabled the outdoor-mode) at the moment and you can disable it directly in the wizard.
You can also enable the outdoor-mode on non-auto-enabled devices via the wifi tab in the advanced settings.

imho this keeps the wizard clean from outdoor-mode related messages that you do not need. (e.g. devices that are not made for outdoor usage)

package/gluon-core/luasrc/lib/gluon/upgrade/180-outdoor-preset
```diff
@@ -0,0 +1,31 @@
...
+    'tp-link-cpe510-v1.0',
+    'tp-link-cpe510-v1.1',
+    'ubiquiti-nanostation-m5',
+    'ubiquiti-nanostation-m5-xw',
+    'ubiquiti-nanostation-loco-m5',
+    'ubiquiti-nanostation-loco-m5-xw',
+    'ubiquiti-unifi-ac-mesh',
+    'ubiquiti-unifi-ac-mesh-pro',
+}
+local model = platform_info.get_image_name()
```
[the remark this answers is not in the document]
mweinelt: That is not going to work for the `unifiac-lite` platform and the Unifi AC-Mesh, respectively the `unifiac-pro` platform and the Unifi AC Mesh Pro

mmalte:
>      - disable ibss/mesh on the 5 GHz radio, as DFS *will*
>        break mesh connections

Is this really necessary? I'm running quite some outdoor installations on stock firmware, some of them haven't seen an DFS event in years. Vendors like Ubiquiti and Cisco are using 5GHz for their proprietary mesh protocols, of cause this breaks for some time whenever they need to change frequency, but it is a valuable addition most of the time.

package/gluon-core/luasrc/lib/gluon/upgrade/180-outdoor-preset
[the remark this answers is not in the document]
NeoRaider: Hmm, `get_model()` would still be better than `get_image_name()` then (and I would still guard it using `platform.match` and only rely on `get_model()` where necessary).

package/gluon-config-mode-outdoor/luasrc/lib/gluon/config-mode/wizard/0250-outdoor.lua
NeoRaider: It is very weird to allow disabling an option through the wizard, but being unable to revert this change without the advanced settings (and it would even be possible to create a firmware that has the wizard checkbox, but not the advanced settings page, leaving only SSH or a factory reset to revert it).

mweinelt:
Sure, mesh on outdoor channels would be awesome. Someone from Bowers and Wilkins is currently contributing support for that to hostapd, including support for encryption, channel switch announcements, reassociation after channel switches.

I'm not quite sure yet how reliable (think race-conditions) CSA will be in dense meshes though.

https://www.spinics.net/lists/hostap/msg04760.html

CodeFetch:
That's great news! I'm still not very happy that I can't find anything about the wpa_supplicant radar pattern detection. It would be nice if there wouldn't be that many false-positives in ath9k. The wpa_supplicant implementation was tested in a lab and seems reliable. I wrote Simon Wunderlich if he knew who did the lightning talk on the BattleMesh, but unfortunately he doesn't. Does somebody here know a person from the Opennet Initiative? The didn't reply to my mail. I would feel much better if we have a reliable DFS detection that does not get triggered all the time in the worst case.

mweinelt:
How did you come across the impression that wpa_supplicant would do pattern matching for DFS? Are you talking about stations submitting that information to APs? Either way the pattern matching is usually part of the radios driver, hostapd simply handles the events.

CodeFetch:
There was a lightning talk on the last BattleMesh where they said that the ath9k radar pattern matching was broken due to a vast amount of false-positives (and maybe they also said that it couldn't detect all kinds of patterns) and that they exposed it to userspace and implemented radar pattern detection in wpa_supplicant. They certified their implementation in a lab. Unfortunately the talk is not on the schedule as it was likely a lightning talk and I am frustrated I can't find out who they were. The Opennet initiative worked on something similar (but their website states implemented in hostapd), but I couldn't get in touch with them, yet. Having a certified DFS implementation for Gluon would be awesome.

package/gluon-config-mode-outdoor/luasrc/lib/gluon/config-mode/wizard/0250-outdoor.lua
Noodlesalat: Maybe we can show the checkbox on auto-enabled devices everytime and do not show it (when outdoor-mode is disabled) on all other 5GHz devices.
Then you could enable and disable the outdoor-mode on all auto-enabled devices directly in the wizard and for all other 5GHz devices the advanced settings are the way to go. This would keep the wizard clean for the majority of all devices and still offers the outdoor-mode-settings.
@NeoRaider @mweinelt Does this sound like a good solution for you?

Noodlesalat:
I have already tested the version based on https://github.com/freifunk-gluon/gluon/pull/1329/commits/eaadb32c4dca4e6b1ca81eb9600cd679482f642e:
My test devices was a AVM FB 4040. I adjusted the preset in order to auto-enable the outdoor-mode on this device. Here are my results:

Singledomain (site repo from Freifunk Fulda)
- [ ] outdoor-mode enabled by default?
- [ ] "Save&Restart" → outdoor-mode enabled, HT20
- [ ] disable outdoor-mode in the wizard → outdoor-mode enabled, HT20
- [ ] enable outdoor-mode in the advanced settings → checkbox in the wizards shows up, outdoor-mode enabled, HT20
- [ ] set htmode to HT40 in the advanced settings → checkbox in the wizards shows up, outdoor-mode enabled, HT40
- [ ] try different combinations in the advanced settings and watched `uci show`

Multidomain (site repo from Freifunk Darmstadt)
- [ ] outdoor-mode enabled by default?
- [ ] "Save&Restart" → outdoor-mode enabled, HT20
- [ ] disable outdoor-mode in the wizard → outdoor-mode enabled, HT20
- [ ] enable outdoor-mode in the advanced settings → checkbox in the wizards shows up, outdoor-mode enabled, HT20
- [ ] set htmode to HT40 in the advanced settings → checkbox in the wizards shows up, outdoor-mode enabled, HT40
- [ ] try different combinations in the advanced settings and watched `uci show`

CodeFetch:
Before merging this we should have a solution for #1662.
While we should be working on modularizing everything, we actually increase the complexity of existing packages. This package could and should be completely independent.

docs/user/site.rst
```diff
@@ -181,6 +181,20 @@ wifi24 \: optional
 wifi5 \: optional
     Same as `wifi24` but for the 5Ghz radio.
 
+    Additionally a range of channels that are safe to use outsides on the 5 GHz band can
```
CodeFetch: As I think this should be a separate package, you should write something like "with the package xxx".

package/gluon-config-mode-outdoor/luasrc/lib/gluon/config-mode/wizard/0250-outdoor.lua
```diff
@@ -0,0 +1,28 @@
+return function(form, uci)
```
CodeFetch: I'd prefer having this as a subsection or similar of the WiFi config section with #1662.

package/gluon-core/check_site.lua
```diff
@@ -27,12 +27,22 @@ need_string_array({'ntp_servers'}, false)
...
+				34, 36, 38, 40, 42, 44, 46, 48, 50, 52, 54, 56, 58, 60, 62,
+				64, 96, 100, 102, 104, 106, 108, 110, 112, 114, 116, 118, 
+				120, 122, 124, 126, 128, 132, 134, 136, 138, 140, 142, 144, 
+				149, 151, 153, 155, 157, 159, 161, 165, 169, 173 }
+			need_one_of({config, 'channel'}, channels)
+			need_chanlist({config, 'outdoor_chanlist'}, channels, false)
+		end
 
 		local rates = {1000, 2000, 5500, 6000, 9000, 11000, 12000, 18000, 24000, 36000, 48000, 54000}
 		local supported_rates = need_array_of(in_site({config, 'supported_rates'}), rates, false)
```
CodeFetch: This check should be done in a package-specific check_site.lua

package/gluon-core/luasrc/lib/gluon/upgrade/180-outdoor-preset
```diff
@@ -0,0 +1,31 @@
...
+
+uci:section('gluon', 'wireless', 'wireless', {
+    outdoor = util.contains(outdoor_models, model) ~= false
+})
+
+uci:save('gluon.wireless')
```
CodeFetch: There should be a single upgrade file after 200 wireless.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -49,22 +49,37 @@ if not sysconfig.gluon_version then
...
+
 local function get_channel(radio, config)
 	local channel
 	if uci:get_first('gluon-core', 'wireless', 'preserve_channels') then
+		-- preserved channel always wins
 		channel = radio.channel
+	elseif  (radio.hwmode == '11a' or radio.hwmode == '11na') and is_outdoor() then
+		-- actual channel will be picked and probed from chanlist
+		channel = 'auto'
 	end
```
CodeFetch: There should be a single independent upgrade file after 200 wireless.

package/gluon-web-wifi-config/i18n/de.po
```diff
@@ -49,3 +49,25 @@ msgstr ""
...
+"to stay on the same channel permanently."
+msgstr ""
+"Ist der Knoten für den Einsatz im Freien konfiguriert, wird ein WLAN-Kanal auf "
+"dem 5-GHz-Band sowie eine Sendeleistung entsprechend den gesetzlichen "
+"Frequenzregulatorien gewählt. Gleichzeitig wird die dynamische Frequenzwahl "
+"(DFS; Radarerkennung) aktiviert und die Mesh-Funktionalität deaktiviert, da "
+"sich Nachbarknoten dauerhaft auf demselben Kanal befinden müssen."
+
+msgid "HT Mode"
+msgstr "HT-Modus"
```
CodeFetch: This translation section should be in a file in an independent package-specific location.

package/gluon-web-wifi-config/i18n/fr.po
```diff
@@ -46,3 +46,9 @@ msgstr ""
 "<br /><br />Ici vous pouvez aussi configurer la puissance d'émmission se votre Wi-Fi. "
 "Prenez note que les valeurs fournies pour la puissance de transmission prennent "
 "en compte les gains fournis par l'antenne, et que ces valeurs ne sont pas toujours disponibles ou exactes."
+
+msgid "Outdoor installation"
+msgstr "Installation extérieure"
+
+msgid "HT Mode"
+msgstr "Mode HT"
```
CodeFetch: This translation section should be in a file in an independent package-specific location.

package/gluon-web-wifi-config/i18n/gluon-web-wifi-config.pot
```diff
@@ -33,3 +33,20 @@ msgid ""
...
+msgid ""
+"Configuring the node for outdoor use tunes the 5 GHz radio to a frequency "
+"and transmission power that conforms with the local regulatory requirements. "
+"It also enables dynamic frequency selection (DFS; radar detection). At the "
+"same time, mesh functionality is disabled as it requires neighbouring nodes "
+"to stay on the same channel permanently."
+msgstr ""
+
+msgid "HT Mode"
+msgstr ""
```
CodeFetch: This translation section should be in a file in an independent package-specific location.

package/gluon-web-wifi-config/luasrc/lib/gluon/config-mode/model/admin/wifi-config.lua
```diff
@@ -1,4 +1,5 @@
 local iwinfo = require 'iwinfo'
```
CodeFetch: This model should be an "inframap" or similar using #1662

CodeFetch:
@mweinelt Sorry for being so pedantic about modularization, but there are two important projects that don't want to merge with Gluon because it is not modular enough. On the one hand LibreMesh and on the other AREDN-Mesh. AREDN-Mesh has totally different requirements on an outdoor mode. Both projects have very skilled programmers and joining efforts is a worthwhile goal.

blocktrron:
@CodeFetch This is the first time i've heard Gluon has the goal to merge with those projects.

CodeFetch:
@blocktrron Merging with LibreMesh is on the plan for more than 3 years now. We've talked about that with the LibreMesh developers on the last BattleMesh and last consent was that after we've successfully modularized the packages to be independent from Batman and Babel integration is fine, we should have a look at how to support BMX7. That AREDN-Mesh wants to have a look at switching to Gluon is new. I have the names of devices in #94 corresponding to vendor device IDs from their firmware and extracted the other power offsets from the firmware. The want to get rid of their Perl-Scripts and switch to a cleaner solution.

NeoRaider:
"Merging with LibreMesh" is unrealistic, and I don't even consider it desirable, as it will further reduce the diversity in mesh firmware frameworks. We can work together and try to create modules that are usable by both projects, but I don't think either project will disappear.

package/gluon-config-mode-outdoor/luasrc/lib/gluon/config-mode/wizard/0250-outdoor.lua
NeoRaider: Still a bit weird, but it sounds acceptable to me.

NeoRaider:
@CodeFetch Modularization can be achieved in a later refactoring.

CodeFetch:
@NeoRaider Yes, it can be done later. I wasn't clear about that. "Merging" was the wrong word. They'd use Gluon as their base system when it's modular enough and as you say merging efforts of common packet development and system stability, device support patching etc.

package/gluon-config-mode-outdoor/luasrc/lib/gluon/config-mode/wizard/0250-outdoor.lua
mweinelt: I need to commit gluon before running `200-wireless` or else it will not apply outdoor settings.

package/gluon-web-wifi-config/luasrc/lib/gluon/config-mode/model/admin/wifi-config.lua
mweinelt: Which is why the line can be removed from what I understand. It's not directly related to the pr though.

package/gluon-core/luasrc/lib/gluon/upgrade/180-outdoor-preset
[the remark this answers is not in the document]
mweinelt: With https://github.com/freifunk-gluon/gluon/pull/1329#discussion_r266271237 I would need to outsource the logic into a library. Is `gluon.platform` the right place for an `is_outdoor_device` function?

package/gluon-config-mode-outdoor/luasrc/lib/gluon/config-mode/wizard/0250-outdoor.lua
NeoRaider: `uci:save('gluon')` should be sufficient before running the upgrade script, the actual commit can then be done via the return.

docs/user/site.rst
NeoRaider: As this feature modifies the logic of wifi interface configuration, moving this into a separate package will make the implementation significantly more complex. If we are actually planning to modularize the interface config (which I don't see at this point), this PR is not the place to discuss it.

package/gluon-core/check_site.lua
NeoRaider: As long as the channel selection logic is not moved to a separate package, this is the correct place to check these values.

package/gluon-core/luasrc/lib/gluon/upgrade/180-outdoor-preset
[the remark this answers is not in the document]
NeoRaider: That would be fine.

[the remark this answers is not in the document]
mweinelt: Addressed in 2ca9951ae9a6c34c8e931cfe3edfcbb8934ba2b0

package/gluon-config-mode-outdoor/luasrc/lib/gluon/config-mode/wizard/0250-outdoor.lua
mweinelt: Addressed in df0ce8f4ea9357dd223d97a393911bedcaee8852

package/gluon-config-mode-outdoor/luasrc/lib/gluon/config-mode/wizard/0250-outdoor.lua
mweinelt: Addressed in df0ce8f4ea9357dd223d97a393911bedcaee8852

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -207,22 +222,33 @@ util.foreach_radio(uci, function(radio, index, config)
...
 	uci:set_list('wireless', radio_name, 'supported_rates', config.supported_rates())
 	uci:set_list('wireless', radio_name, 'basic_rate', config.basic_rate())
 
-
 	local ibss_disabled = is_disabled('ibss_' .. radio_name)
 	local mesh_disabled = is_disabled('mesh_' .. radio_name)
 
 	configure_ibss(config.ibss(), radio, index, suffix,
 		first_non_nil(
+			outdoor_disabled,
```
NeoRaider: Unless I'm overlooking something, this logic will never re-enable the ibss / mesh VIFs, even after outdoor mode is switched off again. I think the best solution would be to delete the ibss / mesh VIF sections completely when outdoor mode is enabled.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -207,22 +222,33 @@ util.foreach_radio(uci, function(radio, index, config)
 	uci:set('wireless', radio_name, 'htmode', htmode)
 	uci:set('wireless', radio_name, 'country', site.regdom())
 
+	local outdoor_disabled = nil
+	if (radio.hwmode == '11a' or radio.hwmode == '11na') then
+		if is_outdoor() then
+			uci:set('wireless', radio_name, 'channels', config.outdoor_chanlist())
```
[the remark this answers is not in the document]
mweinelt: According to hostapd.conf if chanlist is unset it will default to all channels allowed in the selected hw_mode:
```
# Channel list restriction. This option allows hostapd to select one of the
# provided channels when a channel should be automatically selected.
# Channel list can be provided as range using hyphen ('-') or individual
# channels can be specified by space (' ') separated values
# Default: all channels allowed in selected hw_mode
#chanlist=100 104 108 112 116
#chanlist=1 6 11-13
```

But I assume that this will break in mac80211.sh here, is that correct?

https://github.com/openwrt/openwrt/blob/master/package/kernel/mac80211/files/lib/netifd/wireless/mac80211.sh#L97

package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua
```diff
@@ -27,3 +27,20 @@ function match(target, subtarget, boards)
 
    return true
 end
+
+function is_outdoor_device()
+   if match('ar71xx', 'generic', {
+      'cpe510-520-v1',
+      'ubnt-nano-m',
+      'ubnt-nano-m-xw'}) then
```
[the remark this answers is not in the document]
mweinelt: Fixed in 84b914d258527a822a89f848d39bc93492d15d02

package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua
```diff
@@ -27,3 +27,20 @@ function match(target, subtarget, boards)
...
+      return true
+
+   elseif platform.match('ar71xx', 'generic', {'unifiac-lite'}) and platform_info.get_model() == 'Ubiquiti UniFi-AC-MESH' then
```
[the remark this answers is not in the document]
mweinelt: Fixed in 84b914d258527a822a89f848d39bc93492d15d02

scripts/check_site.lua
```diff
@@ -305,6 +332,11 @@ function need_array_of(path, array, required)
 	return need_array(path, function(e) need_one_of(e, array) end, required)
 end
 
+function need_chanlist(path, channels, required)
+	local valid_chanlist = check_chanlist(channels)
+	return need(path, valid_chanlist, required, 'be a valid chanlist')
```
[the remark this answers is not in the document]
mweinelt: Fixed in d857b3367bcce996247dde2dcdf1eac7ad1396c9

package/gluon-core/luasrc/lib/gluon/upgrade/180-outdoor-preset
```diff
@@ -0,0 +1,18 @@
...
+	os.exit(0)
+end
+
+local platform_info = require 'platform_info'
+
+uci:section('gluon', 'wireless', 'wireless', {
+    outdoor = platform_info.is_outdoor_device()
+})
+
+uci:save('gluon.wireless')
```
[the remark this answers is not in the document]
mweinelt: Fixed in 492c22e092cd39e5fadf7efd49412d11412f54c0

package/gluon-core/luasrc/lib/gluon/upgrade/180-outdoor-preset
```diff
@@ -0,0 +1,18 @@
...
+
+if uci:get('gluon', 'wireless', 'outdoor') ~= nil then
+	os.exit(0)
+end
+
+local platform_info = require 'platform_info'
+
+uci:section('gluon', 'wireless', 'wireless', {
+    outdoor = platform_info.is_outdoor_device()
+})
```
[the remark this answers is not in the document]
mweinelt: Fixed in 492c22e092cd39e5fadf7efd49412d11412f54c0

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
mweinelt: Fixed in 64778022828edd97fe6da4d50690d802f84d9f7c

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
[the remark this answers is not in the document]
NeoRaider: https://github.com/openwrt/openwrt/blob/82625374eea0613027d8dbe2bcfdce7beadaa883/package/kernel/mac80211/files/lib/netifd/wireless/mac80211.sh#L311 will not add anything if channel_list is empty, so the resulting config will be valid.

package/gluon-web-wifi-config/luasrc/lib/gluon/config-mode/model/admin/wifi-config.lua
```diff
@@ -97,7 +108,57 @@ uci:foreach('wireless', 'wifi-device', function(config)
 	end
 end)
 
+
+if site.wifi5.outdoor_chanlist() and has_5ghz_radio() then
```
NeoRaider: The other two places where the outdoor mode is set - the preset and the config wizard - don't care if outdoor_chanlist is set, so this code shouldn't either. Otherwise you could upgrade from a firmware with outdoor channels to a firmware without, leaving you without an option to disable outdoor mode.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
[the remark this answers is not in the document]
mweinelt: Okay, but that leaves us with the issue that people with outdoor mode enabled but without a proper chanlist might use an indoor channel outdoors.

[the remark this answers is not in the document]
NeoRaider: The only option I see that can't lead to unexpected behaviour when no outdoor channels are defined after a firmware upgrade would to make the outdoor channel list mandatory... or fetch that information from the regdb?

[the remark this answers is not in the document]
mweinelt: I'll take a peek into the latter.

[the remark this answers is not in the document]
mweinelt: Tried iwinfo.nl80211.freqlist, unfortunately it doesn't expose regulatory information. Only `(channel, mhz, restricted)` per channel. No channel was restricted, so I'm assuming this is the live radar state, whether or not that channel has radar activity.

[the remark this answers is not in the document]
mweinelt: The nl80211 interface for regdb information is not exposed through iwinfo and I don't see myself writing that, so making the outdoor channel list mandatory it is.

package/gluon-web-wifi-config/luasrc/lib/gluon/config-mode/model/admin/wifi-config.lua
mweinelt: Wrt https://github.com/freifunk-gluon/gluon/pull/1329#discussion_r278135451 the current design decision is to make the chanlist mandatory everywhere.

This was addressed in b74fc36b99b8e7a196308a432dda597b5e55e2ca (preset) and 9aed7b22fed9dfd4c93d2ff616b0d282c8ce13c9 (wizard).

mweinelt: The chanlist was now made optional as we have a proper fallback channel selection through a1b317f31eb35f11c7fd32288468cdbecb9290ab.

I dropped https://github.com/freifunk-gluon/gluon/commit/b74fc36b99b8e7a196308a432dda597b5e55e2ca and https://github.com/freifunk-gluon/gluon/commit/9aed7b22fed9dfd4c93d2ff616b0d282c8ce13c9, and removed the channel list condition from web-wifi-config in e522014f44bb8d54b5f2a44e6cd409c1d880fa9f.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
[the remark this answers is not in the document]
mweinelt: Solved by setting `country3=0x4f` in the hostapd configuration.

```

# The third octet of the Country String (dot11CountryString)
# This parameter is used to set the third octet of the country string.
#
# All environments of the current frequency band and country (default)
#country3=0x20
# Outdoor environment only
#country3=0x4f
# Indoor environment only
#country3=0x49
# Noncountry entity (country_code=XX)
#country3=0x58
# IEEE 802.11 standard Annex E table indication: 0x01 .. 0x1f
# Annex E, Table E-4 (Global operating classes)
#country3=0x04
```

package/gluon-core/luasrc/lib/gluon/upgrade/180-outdoor-preset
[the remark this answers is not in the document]
NeoRaider: 1. Rather than doing nothing when sysconfig.gluon_version is set, outdoor should be set to false in this case
2. I think it would make sense to have a site.conf option that can force the default to on or off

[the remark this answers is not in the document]
mweinelt: 1. Wow, I didn't properly read what you wrote earlier, sorry. Fixed in f6b9febb436ee39e4db64c4ddac09155f6226ad2.
2. TODO.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -49,22 +49,37 @@ if not sysconfig.gluon_version then
...
 	local channel
 	if uci:get_first('gluon-core', 'wireless', 'preserve_channels') then
+		-- preserved channel always wins
 		channel = radio.channel
+	elseif  (radio.hwmode == '11a' or radio.hwmode == '11na') and is_outdoor() then
+		-- actual channel will be picked and probed from chanlist
+		channel = 'auto'
 	end
-
+
```
TobleMiner: Please remove trailing whitespace

rubo77: She sa line 231, but aren't those cut out by Gluon SRC diet anyway?

mweinelt: Fixed, thanks!

package/gluon-core/luasrc/lib/gluon/upgrade/180-outdoor-preset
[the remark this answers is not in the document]
mweinelt: 2. https://github.com/freifunk-gluon/gluon/pull/1329/commits/662b84ec82db95a0afe7806801fbaa8ff94d6a68

NeoRaider:
#1716 caused a few merge conflicts...

docs/site-example/site.conf
```diff
@@ -69,6 +69,8 @@
   -- for channel.
   wifi5 = {
     channel = 44,
+    outdoor_chanlist = '100-140',
+    outdoor_preset = true;
```
NeoRaider: `;` should be a `,`.

How about making this a tristate (true = default to outdoor mode enabled, false = default to disabled, nil = default is device-dependent)?

mweinelt: So rather than `outdoor_preset` this would become `outdoor` with states 
- `true` all devices with 5 GHz use outdoor channels (eliminates 5 GHz mesh completely)
- `false` no device uses outdoor channels (same as not defining `outdoor_chanlist`)
- `nil` preset devices use outdoor channels

NeoRaider: One alternative option would be to replace the `nil` with a string like `'preset'` (where 'preset' is the default, usually making nil equivalent to 'preset'). This would allow domain configs to override a true/false site config with an explicit 'preset' option.

mweinelt: Tristate done.

belzebub40k:
I just looked into the feature after it got merged but it looks like there is a mismatch between what is checked in `check_site.lua` ...

https://github.com/freifunk-gluon/gluon/blob/ddb11ddd69528f2578e8a3b1f3a515abe5189937/package/gluon-core/check_site.lua#L47

... and what is read in `180-outdoors`.

https://github.com/freifunk-gluon/gluon/blob/ddb11ddd69528f2578e8a3b1f3a515abe5189937/package/gluon-core/luasrc/lib/gluon/upgrade/180-outdoors#L17

Also the documentation says:

> The `outdoors` option in turn allows to configure when outdoor mode will be enabled.

mweinelt:
Thanks, fixed in ad1a1906ab2aa620417406cc8d112d7549b56640.

NeoRaider:
Hmm, should the site.conf setting really be placed under wifi5? While it doesn't matter for 2.4GHz, other frequency bands that may be added in the future will likely have similar restrictions, so I'd see the outdoor mode more as a global (toplevel) flag.

NeoRaider:
> Hmm, should the site.conf setting really be placed under wifi5? While it doesn't matter for 2.4GHz, other frequency bands that may be added in the future will likely have similar restrictions, so I'd see the outdoor mode more as a global (toplevel) flag.

@mweinelt Opinions on this?

mweinelt:
Yes, long-term you are correct.

Short-term (still several years) we're not going to see another frequency band for outdoor usage, are we? 

We can certainly change this now, I'm not sure we need to.

A-Kasper:
@mweinelt I see a massive IoT Trend. Devices in 433 MHz are getting more popular. Also first Freifunk Communities are starting LoRaWAN stations. I think more and more devices will have support for theses applications. I'm not sure about regulatory restrictions in this Band.

mweinelt:
@A-Kasper I'm not sure how that relates to the topic.

PR #1333, opened 2018-02-19
Update URL for OpenWrt repository

H4ndl3:
Old OpenWrt versions have been moved to https://github.com/openwrt/archive, therefore the URL to the OpenWrt repository needs to be updated.


NeoRaider:
At the moment, https://github.com/openwrt/chaos_calmer is the primary Chaos Calmer repo with the most recent fixes.

H4ndl3:
Huh, that is a bit confusing, I thought the repo has moved entirely upon seeing the chaos_calmer branch inside of it.

PR #1335, opened 2018-02-20
Explaining how to find first parameter for a new device's entry in th…

MPW1412:
…e target file.


docs/dev/hardware.rst
```diff
@@ -30,6 +30,9 @@ the output of the following command (on the target device), so the autoupdater
 can work::
 
     lua -e 'print(require("platform_info").get_image_name())'
+    
+Best practise is to generate an image first by using a parameter at will and then
```
mweinelt: I'd mention the objective first, as is done with the second parameter below.
> Best practice to get the first parameter is …

MPW1412: This should be clear in the context, I guess. Because that is explained in the paragraph before.

mweinelt: Ah yes, I missed that.

mweinelt: > by using an arbitrary value
instead of
> by using a parameter at will

- "at will" seems more idiomatic to me
- also we should be talking about the value of the parameter I guess.

MPW1412:
I added another phrase to accommodate to @mweinelt's wish to make the procedure more clear. I don't think „value“ is a good word here, as a name is not a countable thing. But I'm not a native speaker. 

Motion to merge or close. It's just a tiny improvement.

rotanid:
what about: "an arbitrary name" ?

NeoRaider:
I see no problem with the word "value", and "arbitrary" is more common in technical context than "at will".

rotanid:
rebased and merged.

PR #1337, opened 2018-02-24
[RFC] gluon-debug: add new package to debug process runtime

lemoer:
This package could be helpful, when we want to debug bogus gluon nodes. It provides cpu runtime options per process. It adds a respondd request type `proc`.

Query with:
``` shell
gluon-neighbour-info -r proc -d ::1 -p 1001
```
Output format:
``` json
{
  "<PID>": {
     "n": "<PROCESSNAME>",
     "u": UTIME, "s": STIME, "cu": CUTIME, "cs": CSTIME
  },
  "<PID>": {
     "n": "<PROCESSNAME>",
     "u": UTIME, "s": STIME, "cu": CUTIME, "cs": CSTIME
  }
}
```

It not written very well, but it's only a proof of concept yet. What do you think, is it helpful to have such a debugging option in gluon?


mweinelt:
There shouldn't be any harm in spelling out the key names to something meaningful, if repeated often enough it should vanish in the compression anyway. But doing so will make it more readable.

rotanid:
this package's Makefile needs to be adjusted to the treewide changes which were made in March&April

lemoer:
Closed in favour of #1549.

PR #1340, opened 2018-03-07
batman-adv: multicast TT fixes and cleanups

T-X:
The first one adds a fix that might potentially result in multicast packet
loss once we would enable multicast_mode again.

The second one avoids some small but unnecessary overhead. More
importantly though, it is supposed to ease further multicast improvements
later (e.g. no need for a multicast sending node to determine overlap
between WANT_ALL_IPV4/6 flags and TT entries while on fast-path).


NeoRaider:
Master was updated to batman-adv 2018.0, please rebase.

T-X:
Done.

PR #1341, opened 2018-03-07
gluon-web-model: add hostname validator

rubo77:


rubo77:
i like to see this reimplemented like in luci, so I can repair our package https://github.com/freifunk-kiel/ffki-packages/tree/lede/gluon-config-mode-hostname-no-pretty

NeoRaider:
We have no use for a hostname validator like this, as we allow arbitrary strings as human-readable node names; I don't think there is a good reason to limit users in their choice of node names.

Note that even for DNS purposes, most unicode characters are valid by using IDN encoding (Punycode) (only a few ASCII special characters are forbidden). While not directly related, I would generally advise against a DNS zone with all nodes, as problems like naming conflicts will also lead to confusion.

Furthermore:

* You extended the .js source, but not the generated .js
* You did not extend the server-side Lua validator

This leads me to the assumption that you did not test your change at all. As mentioned before, please don't waste our time and properly test all your changes before opening a PR.

rubo77:
Just for info: the uglyfy command is in the comment in the top of the page: 

    uglifyjs javascript/gluon-web-model.js -o files/lib/gluon/web/www/static/gluon-web-model.js -c -m --support-ie8

and the server-side Lua validator is found at 

    package/gluon-web-model/luasrc/usr/lib/lua/gluon/web/model/datatypes.lua

PR #1342, opened 2018-03-07
French translation: WLAN -> Wi-Fi

rubo77:


NeoRaider:
A quick lookup in a dictionary gives me the spellings "wifi" and "WIFI", but not "Wi-Fi": https://dict.leo.org/franz%C3%B6sisch-deutsch/WLAN

edeso:
i guess https://en.wikipedia.org/wiki/Wi-Fi is the proper spelling according to the alliance behind the standard.

..ede

rubo77:
The french wikipedia page for "wifi" redirects to "wi-fi" too:

https://fr.wikipedia.org/wiki/wifi

rubo77:
also French telecom providers use this term:

https://www.orange.com/ -> search for "wifi" there

NeoRaider:
@edeso As this is about spelling of the French word and not about the standard, that's not a very strong argument.

@rubo77 Well, that Wikipedia page starts with "A Wi-Fi, also spelled wifi, [...]"... Nevertheless, I've merged it like this, as both spellings seem to be in use.

PR #1346, opened 2018-03-17
Check mesh-id, AP SSID and IBSS SSID length

2tata:


NeoRaider:
Thanks, merged. I edited the commit messages to fix a typo and make the wording of both commits match, please double-check the messages yourself next time.

NeoRaider:
Hmm, I might have merged this a bit too quickly. I made another change to accept 0-character ESSIDs/mesh IDs, as there is no technical reason not to allow them.

Note that these matches might still not be fully accurate: The Lua match functions match on *characters*, while an ESSID is limited to 32 *bytes*. The OpenWrt/LEDE build system should use C locale, but I'm not completely sure how this interacts with Unicode strings in Lua files.

PR #1347, opened 2018-03-18
gluon-config-mode-domain-select: allow hiding of individual domain codes

mweinelt:
Allow hiding invidiual domain codes and names from the config mode drop-down.

### DIFF
--- a/package/gluon-config-mode-domain-select/check_site.lua
+++ b/package/gluon-config-mode-domain-select/check_site.lua
@@ -1,1 +1,2 @@
 need_boolean(in_domain({'hide_domain'}), false)
+need_array_of(in_domain({'hide_domain_codes'}), table_keys(need_table(in_domain({'domain_names'}), nil, true)), false)
--- a/package/gluon-config-mode-domain-select/check_site.lua
+++ b/package/gluon-config-mode-domain-select/check_site.lua
@@ -1,2 +1,3 @@
 need_boolean(in_domain({'hide_domain'}), false)
-need_array_of(in_domain({'hide_domain_codes'}), table_keys(need_table(in_domain({'domain_names'}), nil, true)), false)
+valid_domain_codes = table_keys(need_table(in_domain({'domain_names'})))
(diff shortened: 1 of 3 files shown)


NeoRaider:
Hmm, this makes the *hide_domain* feature redundant. Maybe we can merge both, i.e. setting *hide_domain_codes* to *true* instead of a list of names hides all of them? I guess we would need to introduce some new check_site facility for such alternatives though...

mweinelt:
Only redundant in that if you completely mirror all domain_codes the behaviour is the same. Instead I need something to only hide single domain_codes, not the whole domain.

What we currently do is:
 - label domain with numeric identifier, e.g. `dom6`
 - create aliases that reflect municipalities/quarters and add them as aliases to one of those numeric domains
 - the numeric identifier is an organizational name, nothing a user should be encouraged to choose in the config-mode

NeoRaider:
I have an implementation in mind for check_site alternatives ("any one of the following conditions must be met"), so I would like to proceed with merging *hide_domain* and *hide_domain_codes* into a single option. I think I like the name *hide_domain_codes* better.

package/gluon-config-mode-domain-select/check_site.lua
```diff
@@ -1 +1,2 @@
 need_boolean(in_domain({'hide_domain'}), false)
+need_array_of(in_domain({'hide_domain_codes'}), table_keys(need_table(in_domain({'domain_names'}), nil, true)), false)
```
NeoRaider: This would be more readable if you assigned the return value of `table_keys(need_table(in_domain({'domain_names'})))` to a local variable. A value is required by default, so the `, nil, true` can be omitted.

mweinelt:
The issue with that is that currently need will result in a val_error, instead of returning, on error.

NeoRaider:
Merged with some changes. I went for the name "hide_domain" eventually, as we the term "domain code" only internally so far. Consequently existing configs that set *hide_domain* to *true* don't need to be adjusted.

PR #1354, opened 2018-03-28
Domain fix for tunneldigger and mesh eth interfaces

2tata:
* Allow tunneldigger configurations only in domains.
* Allow mesh_on_wan, mesh_on_lan and single_as_lan in both site and domain configurations.


NeoRaider:
> Allow mesh_on_wan, mesh_on_lan and single_as_lan in both site and domain configurations.

This is not possible, all of these values are evaluated before a domain is chosen.

NeoRaider:
I've merged the other patch. Please also make sure your commits always have a proper package name prefix, I had to fix this up again.

PR #1357, opened 2018-04-03
RFC: batman-adv: directed multicast support

T-X:
The following changes and patches add multicast-to-multi-unicast support between nodes. It supports IPv6 link-local multicast with up to 16 target nodes and if exceeded, either falls back to flooding (e.g. for important ones like ICMPv6 Neighbor Solicitations) or drops them if the gluon-ebtables-filter-multicast is installed.

This should fix our current issue of larger mesh networks having a significant amount of and overhead through ICMPv6 Neighbor Solicitations.

Recap of what we already have:

- Multicast listener awareness: A node knows which client device is interested in which multicast transmissions (bridge IGMP/MLD snooping, IGMP/MLD querier per node, multicast-capable batman-adv TT). -> Since current Gluon master
- Multicast-to-unicast from node to client: A node delivers multicast packets via individual unicast transmissions (only) to clients interested in it instead of broadcasting. -> Since Gluon v2016.2

This adds the missing piece to have directed, brodcast-less multicast for the whole path from one client to another.

For batman-adv:
- It adds some patches to batman-adv to allow not just multicast-to-single-unicast but also multicast-to-multi-unicast forwarding for one thing.
- For another, adds a patch to batman-adv to allow it to forward a frame via multicast-to-unicast but not via flooding. This is indicated to batman-adv via a "noflood" packet mark.

For gluon-ebtables-filter-multicast:
- Relaxes the filtering rules to allow a packet which would previously be dropped to pass if it were transsmitted via up to 16 unicast transmissions (so to up to 16 nodes in the mesh).

For gluon-mesh-batman-adv:
- Reenables the multicast_mode
- Sets multicast_fanout limit to 16.
- Assigns the marking number for the batman-adv noflood mark

This needs a good amount of (more) testing from other people, therefore sending it as RFC. 

Once your community has fully migrated to a Gluon v2018.1 you can test this patchset: v2018.1 has the whole listener side ready. This patch only touches the multicast sender side, so you can add this patchset to dedicated nodes which have the multicast sending devices while receiving on any Gluon >=v2018.1 node.

### DIFF
--- a/package/gluon-ebtables-filter-multicast/files/lib/gluon/ebtables/355-mcast-drop
+++ b/package/gluon-ebtables-filter-multicast/files/lib/gluon/ebtables/355-mcast-drop
@@ -1,1 +1,6 @@
+if os.execute('batctl -v | grep -q "batman-adv: 2013.4"') ~= 0 then
+	rule ('MULTICAST_OUT -p IPv6 --ip6-dst ff02::1/128 -j DROP')
+	rule ('MULTICAST_OUT -p IPv6 --ip6-dst ff02::/ff0f:: -j mark --set-mark 0x4 --mark-target RETURN')
+end
+
 rule ('MULTICAST_OUT -j DROP')
--- a/package/gluon-mesh-batman-adv/luasrc/lib/gluon/upgrade/310-gluon-mesh-batman-adv-mesh
+++ b/package/gluon-mesh-batman-adv/luasrc/lib/gluon/upgrade/310-gluon-mesh-batman-adv-mesh
@@ -21,6 +21,7 @@ uci:section('batman-adv', 'mesh', 'bat0', {
 	routing_algo = routing_algo,
 	multicast_mode = true,
 	multicast_fanout = 16,
+	noflood_mark = '0x4/0x4',
 })
 uci:save('batman-adv')
 
--- a/package/gluon-mesh-batman-adv/luasrc/lib/gluon/upgrade/310-gluon-mesh-batman-adv-mesh
+++ b/package/gluon-mesh-batman-adv/luasrc/lib/gluon/upgrade/310-gluon-mesh-batman-adv-mesh
@@ -19,7 +19,8 @@ uci:section('batman-adv', 'mesh', 'bat0', {
 	gw_sel_class = gw_sel_class,
 	hop_penalty = 15,
 	routing_algo = routing_algo,
-	multicast_mode = false,
+	multicast_mode = true,
+	multicast_fanout = 16,
 })
 uci:save('batman-adv')
 
--- a/patches/packages/routing/0002-batman-adv-mesh-wide-multi-target-multicast-to-unicast-support.patch
+++ b/patches/packages/routing/0002-batman-adv-mesh-wide-multi-target-multicast-to-unicast-support.patch
@@ -1,0 +1,971 @@
+From: Linus Lüssing <linus.luessing@c0d3.blue>
+Date: Thu, 29 Mar 2018 20:44:01 +0200
+Subject: batman-adv: mesh-wide, multi-target multicast-to-unicast support
+
+We already have multicast-to-unicast support between client and node.
+And batman-adv has multicast-to-unicast conversion capabilities if there
+is only one receiver. Now this patchset adds node-to-node
+multicast-to-unicast support to multiple targets as well.
+
+This should save airtime for ICMPv6 Router Solicitations, for instance.
+Here we have multiple targets for the same IPv6 multicast destination
+address. However, the number of destination nodes is still rather small
+in this case, i.e. all gateways, making multiple unicast transmissions
+less costly than flooding a packet through the whole mesh.
+
+Also, this allows playing with multicast streaming to a limited set of
+destination nodes (default setting: 16).
+
+Also, nearly all nodes should be running a batman-adv version which
+was compiled with CONFIG_BATMAN_ADV_MCAST=y (e.g. Gluon v2018.1 or
+later) for this feature to take effect.
+
+Signed-off-by: Linus Lüssing <linus.luessing@c0d3.blue>
+
+diff --git a/batman-adv/files/etc/config/batman-adv b/batman-adv/files/etc/config/batman-adv
+index 6e6017a31468c8e595ec13b09e27529f28f3eaec..8a4e881537ca48dc8e611bab9e22918a8dd02fe4 100644
+--- a/batman-adv/files/etc/config/batman-adv
++++ b/batman-adv/files/etc/config/batman-adv
+@@ -12,9 +12,11 @@ config 'mesh' 'bat0'
+ 	option 'bridge_loop_avoidance'
+ 	option 'distributed_arp_table'
+ 	option 'multicast_mode'
++	option 'multicast_fanout'
+ 	option 'network_coding'
+ 	option 'hop_penalty'
+ 	option 'isolation_mark'
++	option 'noflood_mark'
+ 
+ # yet another batX instance
+ # config 'mesh' 'bat5'
+diff --git a/batman-adv/files/lib/batman-adv/config.sh b/batman-adv/files/lib/batman-adv/config.sh
+index 3a746a3c0e280b9a01ae67ea27bde65619b77ebc..d45c4009d24d0d2c416d23134aed755bc36b9413 100644
+--- a/batman-adv/files/lib/batman-adv/config.sh
++++ b/batman-adv/files/lib/batman-adv/config.sh
+@@ -12,8 +12,8 @@ bat_config()
+ {
+ 	local mesh="$1"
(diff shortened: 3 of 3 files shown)


T-X:
So far I tested this patchset in the following two scenarios:

### Virtual environment 

**Scenario:**
- 6 nodes (6 KVM instances)
- line topology

**Tested:**
- verified via tcpdump that multicast packets are transmitted correctly via individual batman-adv unicast packets
- verified that multicast_fanout works, that is that batman-adv correctly falls back to flooding if exceeded
- verified that the "noflood" mark works, that is that if a packet were to exceed multicast_fanout that such packets are correctly dropped instead of flooded.

### Testbed (non-virtual)

**Scenario:**

mcast-to-ucast-setup

- 3 nodes (top: TP-Link WDR4300, middle+bottom: WDR3600; meshed on 2.4GHz only as thick walls give no chance for 5GHz)
- 6 laptops (1x sender + 1x receiver in the cellar, 4x receivers at the top)
- Sender in the cellar transmitts three 640x320px videos simultaneously in a loop, using a dedicated IPv6 link-local multicast address for each video
- Receivers sign up for all three multicast addresses


**Screenshot from a client device, running 3x VLC and H.O.R.S.T.**

Commands:

- Sender:
<pre><code>$ vlc --loop ./video1.mp4 --sout="#rtp{dst=[ff12::23%enp0s25],port=5023,mux=ts,sap,name=foobar}" --sout-keep
$ vlc --loop ./video2.mp4 --sout="#rtp{dst=[ff12::42%enp0s25],port=5042,mux=ts,sap,name=foobar}" --sout-keep
$ vlc --loop ./video3.mp4 --sout="#rtp{dst=[ff12::123%enp0s25],port=5123,mux=ts,sap,name=foobar}" --sout-keep
</code></pre>

- Receiver:
<pre><code>$ vlc "rtp://@[ff12::23%wlan0]:5023"
$ vlc "rtp://@[ff12::42%wlan0]:5042"
$ vlc "rtp://@[ff12::123%wlan0]:5123"
</code></pre>

**Observations**

will edit and insert here

T-X:
PS: In the video streaming test I also did a brief mobility/handover test. With one of the laptops at the top I moved down to the cellar and back up, briefly stopping at each node.

When moving around the lift before reaching the middle node the streams paused for a few seconds. As expected that metal beast left no chance to wpa_supplicant to perform any smooth handover. Other than that, video playback continued smoothly with only a few video frames lost. At each node I verified that the laptop and its wpa_supplicant had switched to the new AP.

patches/packages/routing/0005-batman-adv-mesh-wide-multi-target-multicast-to-unicast-support.patch
```diff
@@ -0,0 +1,1091 @@
...
+++	unsigned int limit = 2 * atomic_read(&bat_priv->multicast_fanout);
+++	int ret;
+++
+++	ret = batadv_mcast_forw_tt_send(bat_priv, skb, vid, &limit);
+++	if (ret != NET_XMIT_SUCCESS)
+++		return ret;
+++
+++	ret = batadv_mcast_forw_want_all_send(bat_priv, skb, vid, &limit);
+++	if (ret != NET_XMIT_SUCCESS)
+++		return ret;
```
blocktrron: In conjunction with Line 423, this seems to have the effect of not delivering the packet to nodes with the want-all flag set in case no other node has the multicast address in the transtable.

T-X: Urgh, you are absolutely right, line 423 should return NET_XMIT_SUCCESS instead of NET_XMIT_DROP... will fix this PR later today. Many thanks for spotting and identifying the issue, @blocktrron! And thanks to you and @mweinelt for testing this PR.

T-X:
Changelog v2:
- Fixed return value in "batadv_mcast_forw_tt_send()" for an empty TT entry: the want-all lists still need to be tried. This fixes potential packetloss. - Discovered by @mweinelt and @blocktrron

mweinelt:
I can confirm the issues we encountered are fixed in v2. Thanks!

T-X:
Changelog v3:
* Updated to Gluon master branch (which is basically Gluon v2018.1 right now)

As soon as some community returns some positive feedback for this patchset in their otherwise Gluon v2018.1 based production network network then I will remove the "RFC".

As I said before, no need to apply this patchset on all nodes yet, just some test node(s) is fine as long as (most of) the rest of your network runs v2018.1.

T-X:
Some useful commands to check whether your network is and which multicast addresses are or are not suitable for tests:

* Number of unsuportive nodes: ``$ batctl mf | grep " -"``
* Number of receiver nodes for a specific multicast address: ``$ batctl tg -m -H | sed "s/...\([^ ]*\) .*/\1/" | sort | uniq -c | sort -n -r | less``

Make sure that for the multicast address you are playing with has a total, summed unsupportive node count plus receiver node count matching your multicast address < 16.

Also, with the VLC example above your multicast receivers should show up in this "number of receivers" table. Also on your multicast receiving host running Linux have a look at ``$ ip maddr show`` to find the IPv6 multicast address to MAC multicast address mapping.

christf:
we need to understand if #1468 is induced by this patch. Happily removing blocked when we know this is not the case.

mweinelt:
It is unclear if the multicast patches are the cause of the crash in #1468, so this is not blocked but instead requires a decision on how to proceed.

We in Darmstadt have been testing this patch since its inception and had no issues with it in months.

christf:
The patch is meant to appeal to a wider audience in order to gain momentum for an upstream merge. I find the approach to be valid and I am interested in the feature.

As discussed on IRC I see two risks:
1. If merged everyone automatically has the patch. Maybe it would be beneficial to roll this out only when consciously selected
2. We have had the discussion previously whether to accept patches because they must be maintained and we have usually decided for upstreaming patches if possible. This is the goal all along so I support merging this patch given (1) is addressed.

rotanid:
i'm against merging a non-upstreamed patch like this.
if upstream doesn't want it even though they have been told about the FFDA test results, they will hopefully name reasonst. but afaik no one submitted it yet.

T-X:
Just some background information why I haven't submitted the patches to upstream yet:

Linux network subsystem maintainers are increasingly demanding a "new" interface for configurations (sysfs -> netlink). And have recently expressed this towards batman-adv which does not have a netlink interface for configuration options yet. Therefore any patch introducing new configuration options, including this one, will likely be rejected for now.

@ecsv has provided some initial [patches to introduce netlink support](https://patchwork.open-mesh.org/project/b.a.t.m.a.n./list/?series=188&state=%2A&archive=both) for configuration options in batman-adv. However, this will probably need a bit more time. Until then I can't submit this patch upstream.

I'm also absolutely in favor of an "upstream-first" policy. I guess since many communities have switched to a multi-domain setup now, there is no hurry?

Otherwise, if multicast overhead is an urgent issue for some larger communities I guess we could also interpret this pull-request as a fix. And apply it regardless of upstreaming status (and I'm definitely going to upstream). But I currently have no overview of the urgency.

T-X:
Changelog v4:
* rebased to master (new Patch numbers)
* fixed a double free bug in the batman-adv multicast-to-multi-unicast patch -> led to crashes, especially on netifd restart/shutdown (#1468)
* changed the default multicast_fanout from 1 to 16 in the batman-adv multicast-to-multi-unicast patch (no functional change, as it was set to 16 by the netifd scripts, too; just to sync with the proposal that was sent to the batman-adv mailing list)

T-X:
Changelog v5:
* fixed skb memory leaks in error cases

T-X:
Changelog v6:

* rebased to master (no changes/conflicts)
* fixed the batadv_mcast_forw_tt_send() again, which was fixed in v2 but reintroduced by v4

ecsv:
The fanout change will be part of the batman-adv 2019.2 release

rotanid:
@T-X one of your commit messages should read "gluon-mesh-batman-adv:" instead of "batman-adv:"

@ecsv is "the fanout change" the whole upstream part of this PR?

ecsv:
@rotanid No, there are also a noflood_mark which is not upstream. I am unsure about its usefulness in upstream batman-adv since it is rather tied to the gluon specific 355-mcast-drop. But we also have this ap_isolation which is rather similar.

package/gluon-ebtables-filter-multicast/files/lib/gluon/ebtables/355-mcast-drop
```diff
@@ -1 +1,6 @@
+if os.execute('batctl -v | grep -q "batman-adv: 2013.4"') ~= 0 then
```
mweinelt: batman-adv-legacy was dropped after v2019.1 was branched.

PR #1363, opened 2018-04-11
Switch to ebtables-tiny, build in most kernel modules

NeoRaider:
ebtables-tiny is a cleane up version of ebtables, with all table/match extensions we need linked in statically and far less features. The --concurrent switch is removed, as ebtables-tiny always uses a lockfile.

The other patches get rid of a few unnecessary kernel module and attempt to build as many modules into the kernel as possible to reduce the per-module overhead of loadable modules. A number of patches is backported from the OpenWrt master to fix issues caused by the bigger kernel on some devices.

Overall, the sysupgrade images shrink by about 2 erase blocks (2*64KB): The kernel grows by 2 erase blocks, while the rootfs shrinks by 4. RAM usage is reduced by 2~3 MB due to the removed modules.

Test feedback is welcome, especially from communities with OOM/load issues.


mweinelt:
TP-Link 1043v2

- 65% (before) vs 63% (after) memory usage after 30 minutes uptime

1043v2-uptime
1043v2-memory

NeoRaider:
@mweinelt

The "free" memory should show a more significant difference than "available" memory (free + cache + buffers), which is probably what your graphs show. This means that cache use has dropped (probably due to the smaller rootfs). I hope that the ~1MB more "available" memory together with the reduced cache usage will avoid the cache-related load issues we observed.

mweinelt:
The formula used is `total-free/total` with data taken from respondd.

mweinelt:
Memory usage looks a lot better indeed, roughly 5% (~3.2MB) have freed up. Tests with the usual suspects will probably happen this weekend.

memory

PR #1366, opened 2018-04-14
gluon-status-page, gluon-mesh-batman-adv: make statuspage mesh-protoc…

christf:
…ol-agnostic

this PR will make the staus page backend independent from batctl.

this replaces the backend-part from #1053


package/gluon-mesh-batman-adv/src/Makefile
```diff
@@ -33,3 +33,4 @@ LDLIBS += $(LIBBATADV_LDLIBS)
 
 respondd.so: respondd.c
 	$(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) -shared -fPIC -D_GNU_SOURCE -o $@ $^ $(LDLIBS) -lgluonutil -liwinfo -luci
+
```
NeoRaider: It seems this line was added accidentally?

package/gluon-status-page-mesh-batman-adv/Makefile
```diff
@@ -0,0 +1,23 @@
...
+
+PKG_NAME:=gluon-status-page-mesh-batman-adv
+PKG_VERSION:=1
+
+include ../gluon.mk
+
+define Package/gluon-status-page-mesh-batman-adv
+  TITLE:=Batman-data provider for gluon-status-page
+  SECTION:=gluon
+  CATEGORY:=Gluon
```
NeoRaider: The above two lines can be dropped since 994c94918a73a673f25a65fd6795fd89e19decec

package/gluon-status-page/files/lib/gluon/status-page/www/cgi-bin/dyn/neighbours-nodeinfo
```diff
@@ -6,6 +6,6 @@ badrequest() {
         exit 1
 }
 
-( batctl if | cut -d: -f1 | grep -qxF "$QUERY_STRING" ) 2>/dev/null || badrequest
+( ubus call network.interface dump | jsonfilter -e "@.interface[@.proto='gluon_mesh' && @.up=true].device"| grep -qxF "$QUERY_STRING" ) 2>/dev/null || badrequest
```
NeoRaider: Hmm, I think the complexity of this command is high enough that we should encapsulate the whole thing up to the `||` in a small script provided by gluon-core. Something like `gluon-is-meshif`. I guess the `grep` could also be replaced by an extension of the jsonfilter expression.

package/gluon-status-page-mesh-batman-adv/src/Makefile
```diff
@@ -0,0 +1,40 @@
...
+  endif
+  LIBBATADV_CFLAGS += $(shell $(PKG_CONFIG) --cflags $(LIBBATADV_NAME))
+  LIBBATADV_LDLIBS += $(shell $(PKG_CONFIG) --libs $(LIBBATADV_NAME))
+endif
+CFLAGS += $(LIBBATADV_CFLAGS)
+LDLIBS += $(LIBBATADV_LDLIBS)
+
+neighbours-batadv: neighbours-batadv.c
+	$(CC) $(CFLAGS) $(CFLAGS_JSONC) $(LDFLAGS) $(LDFLAGS_JSONC) -o $@ $^ $(LDLIBS)
+
```
NeoRaider: Ugh, I hate the amount of boilerplate code in this file... I'll clean this up very soon, in all Makefiles that use libbatadv.

package/gluon-status-page/src/stations.c
```diff
@@ -27,6 +27,8 @@ static struct json_object *get_stations(const struct iwinfo_ops *iw, const char
     json_object_object_add(station, "signal", json_object_new_int(entry->signal));
     json_object_object_add(station, "noise", json_object_new_int(entry->noise));
     json_object_object_add(station, "inactive", json_object_new_int(entry->inactive));
+    json_object_object_add(station, "rx_rate", json_object_new_int(entry->rx_rate.rate));
+    json_object_object_add(station, "tx_rate", json_object_new_int(entry->tx_rate.rate));
```
NeoRaider: This change looks unrelated, and we should not add these values unless the frontend also uses them.

package/gluon-status-page-mesh-batman-adv/Makefile
```diff
@@ -0,0 +1,23 @@
...
+  PROVIDES:=gluon-status-page-data-provider
+  DEPENDS:=+gluon-status-page +gluon-mesh-batman-adv +libnl-tiny +libbatadv +libjson-c
```
NeoRaider: *gluon-mesh-batman-adv* is a virtual package. Remove the + sign prefixing the dependency - never use selecting dependencies with virtual packages!

package/gluon-status-page-mesh-batman-adv/Makefile
```diff
@@ -0,0 +1,23 @@
...
+PKG_NAME:=gluon-status-page-mesh-batman-adv
+PKG_VERSION:=1
+
+include ../gluon.mk
+
+define Package/gluon-status-page-mesh-batman-adv
+  TITLE:=Batman-data provider for gluon-status-page
+  SECTION:=gluon
+  CATEGORY:=Gluon
+  PROVIDES:=gluon-status-page-data-provider
```
NeoRaider: I think this PROVIDES is somewhat useless; as long as *gluon-status-page-mesh-batman-adv* depends on *gluon-status-page*, the result will be a circular dependency, which should always be avoided.

As the status page does not have a public API anymore, the status-page frontend for a specific mesh protocol will always be provided by the same package the backend comes in. As such, there is no common functionality in different backend packages that would justify a PROVIDES.

package/gluon-status-page/files/lib/gluon/status-page/www/cgi-bin/dyn/neighbours-nodeinfo
christf: agreed - how about we create gluon-get-mesh-ifs  that contains everything up to the Igrep. This would be usable in a more universal way than when incorporating the grep for a specific interface....

package/gluon-status-page-mesh-batman-adv/Makefile
christf: done

package/gluon-status-page/files/lib/gluon/status-page/www/cgi-bin/dyn/neighbours-nodeinfo
christf: done

package/gluon-status-page/src/stations.c
christf: in general I agree but since we are splitting backend and frontend here, I left it in the commit. I will remove it now, but I really do like those stats.

package/gluon-status-page-mesh-batman-adv/Makefile
christf: done.

package/gluon-core/files/usr/bin/get-mesh-ifs
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+ubus call network.interface dump | jsonfilter -e "@.interface[@.proto='gluon_mesh' && @.up=true].device"
```
NeoRaider: Maybe name it "gluon-mesh-interfaces"? In any case, it should start with "gluon-".

package/gluon-l3roamd/files/etc/init.d/gluon-l3roamd
```diff
@@ -10,7 +10,7 @@ start_service () {
 	procd_set_param stdout 1
 	procd_set_param stderr 1
 	procd_set_param respawn ${respawn_threshold:-3660} ${respawn_timeout:-5} ${respawn_retry:-0}
-	interfaces=$(for dev in $( ubus call network.interface dump | jsonfilter -e "@.interface[@.proto='gluon_mesh' && @.up=true].device");do echo " -m $dev"; done;
+	interfaces=$(for dev in $(/usr/bin/get-mesh-ifs);do echo " -m $dev"; done;
```
NeoRaider: Remove the full path here, just use the name of the script.

NeoRaider:
I've prepared the frontend part in 35ade80f1c827fd5d61528152a8a4e58ceccc7fa.

package/gluon-core/files/usr/bin/get-mesh-ifs
christf: sure. I like verbs, so I would go with gluon-get-mesh-interfaces ok?

NeoRaider: Ok

package/gluon-l3roamd/files/etc/init.d/gluon-l3roamd
christf: removed the path here but kept it in the scripts that are called by the statuspage.

NeoRaider: Why, do those script not have a proper PATH set?

christf:
rename and rebase is done.

package/gluon-status-page/files/lib/gluon/status-page/www/cgi-bin/dyn/neighbours-nodeinfo
```diff
@@ -6,6 +6,6 @@ badrequest() {
         exit 1
 }
 
-( ubus call network.interface dump | jsonfilter -e "@.interface[@.proto='gluon_mesh' && @.up=true].device"| grep -qxF "$QUERY_STRING" ) 2>/dev/null || badrequest
+( /usr/bin/gluon-get-mesh-interfaces | grep -qxF "$QUERY_STRING" ) 2>/dev/null || badrequest
```
NeoRaider: The other absolute paths in these files don't make much sense either, let's not introduce another one.

NeoRaider: Sorry about changing my mind again, but I think I would prefer the verb "list" to "get" in the script name.

christf: done. - for the paths I had to touch the files again anyways. I like list better than get too.

NeoRaider:
Thanks, merged.

PR #1368, opened 2018-04-16
gluon-status-page: when prettifying undefined numbers, show 0 instead of NaN

christf:
this will show 0 instead of NaN on the status-page when we are prettifying undefined values in case of missing data.


NeoRaider:
I don't think this is a good idea. As status page frontend and backend are always provided by the same node now, there should never be missing data (and 0 is a very bad replacement, it should show something like "n/a" or "-" (we show "-" in the neighbour list for unavailable fields).

christf:
I agree "0" probably is not the best choice.

The idea stems from the traffic statistics. There we have RX/TX/Forward.
Forward is somewhat batman specific and shows up as a rather ugly NaN value at the moment in the babel firmware. I am not even sure a corresponding value can be identified with sane amount of work in the babel network.
At the same time, I am not entirely sure of the value of the "forward"-field in the status page so one could argue that the best approach would be dropping it instead of adjusting the frontend to data that wasn't provided.

NeoRaider:
Let's make it "-" for consistency with the neigh list. Additional comments following in a moment.

package/gluon-status-page/javascript/status-page.js
```diff
@@ -39,6 +39,10 @@
 	function prettyPrefix(prefixes, step, d) {
 		var prefix = 0;
 
+		if (d == undefined) {
```
NeoRaider: Always use `===` instead of `==`.

package/gluon-status-page/javascript/status-page.js
```diff
@@ -39,6 +39,10 @@
...
+			return "-"
+		}
```
NeoRaider: Remove {} for single statement.

package/gluon-status-page/javascript/status-page.js
```diff
@@ -39,6 +39,10 @@
 	function prettyPrefix(prefixes, step, d) {
 		var prefix = 0;
 
+		if (d == undefined) {
+			return "-"
```
NeoRaider: Add semicolon for consistent style.

NeoRaider: The returned value should be `"- "`, so there is a space between the `-` and the unit.

christf: I thought about that. returning "-" here should lead to no unit at all,  am I reading this wrong?

NeoRaider:
You also forgot to update the minified status-page.js.

package/gluon-status-page/javascript/status-page.js
NeoRaider: prettyBits and prettyBytes will still add "bps"/"B".

NeoRaider:
Thanks, merged. I have regenerated the minimized version again, as using my version of uglify-js (3.3.21) resulted in a smaller file.

PR #1369, opened 2018-04-20
x86: add packages: leds, sensors, gpio, usb, wifi and tools

rotanid:
as we have enough space on x86 devices (but no individual profiles) i'd like to add some packages to the x86 target which are useful especially for the pcengines APU2 device.
those are build and run tested on this device.

i'm not really interested in discussions about single packages, so please just name the ones i have to remove from the pull request, if you have doubts about some.

### DIFF
--- a/targets/x86-64
+++ b/targets/x86-64
@@ -1,7 +1,18 @@
 config 'CONFIG_VDI_IMAGES=y'
 config 'CONFIG_VMDK_IMAGES=y'
+
+ATH10K_PACKAGES='kmod-ath10k-ct ath10k-firmware-qca9887 ath10k-firmware-qca988x'
+if [ "$GLUON_WLAN_MESH" = 'ibss' ]; then
+	ATH10K_PACKAGES='kmod-ath10k-ct ath10k-firmware-qca9887-ct ath10k-firmware-qca988x-ct'
+fi
+
 packages 'kmod-3c59x' 'kmod-8139cp' 'kmod-8139too' 'kmod-e100' 'kmod-e1000' 'kmod-e1000e' 'kmod-forcedeth' 'kmod-igb' 'kmod-natsemi' 'kmod-ne2k-pci'
 packages 'kmod-pcnet32' 'kmod-r8169' 'kmod-sis900' 'kmod-sky2' 'kmod-tg3' 'kmod-tulip' 'kmod-via-rhine' 'kmod-via-velocity'
+packages 'kmod-ath9k'
+packages $ATH10K_PACKAGES
+packages 'kmod-gpio-button-hotplug' 'kmod-gpio-nct5104d' 'kmod-hwmon-core' 'kmod-hwmon-k10temp' 'kmod-leds-gpio' 'kmod-leds-apu2' 'kmod-sp5100_tco'
+packages 'kmod-usb-core' 'kmod-usb-ohci' 'kmod-usb2' 'kmod-usb3' 'kmod-usb-serial'
+packages 'bash' 'wget-nossl'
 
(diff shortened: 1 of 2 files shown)


targets/x86-64
```diff
@@ -2,6 +2,9 @@ config 'CONFIG_VDI_IMAGES=y'
 config 'CONFIG_VMDK_IMAGES=y'
 packages 'kmod-3c59x' 'kmod-8139cp' 'kmod-8139too' 'kmod-e100' 'kmod-e1000' 'kmod-e1000e' 'kmod-forcedeth' 'kmod-igb' 'kmod-natsemi' 'kmod-ne2k-pci'
 packages 'kmod-pcnet32' 'kmod-r8169' 'kmod-sis900' 'kmod-sky2' 'kmod-tg3' 'kmod-tulip' 'kmod-via-rhine' 'kmod-via-velocity'
+packages 'kmod-ath9k' 'kmod-ath10k'
```
NeoRaider: We should distinguish IBSS and 11s like in other targets and install kmod-ath10k-ct for IBSS. This is also missing the necessary ath10k firmware packages.

mweinelt:
I'd opt for a more comfortable shell, like a bash.

targets/x86-64
rotanid: i'll look into this.

rotanid:
@Adorfer installing another shell shouldn't change the default, but we simply try it out.

rotanid:
hm, just noticed that some packages are in upstreams default package set for x86-64: https://github.com/openwrt/openwrt/blob/master/target/linux/x86/64/target.mk#L3
@NeoRaider can we rely on them staying there and therefore leave them out here?

targets/x86-64
mweinelt:
```
# opkg list | grep "ath10k-firmware" | awk '{ print $1 }'
ath10k-firmware-qca4019
ath10k-firmware-qca4019-ct
ath10k-firmware-qca4019-ct-htt
ath10k-firmware-qca6174
ath10k-firmware-qca9887
ath10k-firmware-qca9887-ct
ath10k-firmware-qca9887-ct-htt
ath10k-firmware-qca9888
ath10k-firmware-qca9888-ct
ath10k-firmware-qca9888-ct-htt
ath10k-firmware-qca988x
ath10k-firmware-qca988x-ct
ath10k-firmware-qca988x-ct-htt
ath10k-firmware-qca9984
ath10k-firmware-qca9984-ct
ath10k-firmware-qca9984-ct-htt
ath10k-firmware-qca99x0
ath10k-firmware-qca99x0-ct
ath10k-firmware-qca99x0-ct-htt
```

mweinelt:
Considering how often that list was touched in the last years (https://github.com/openwrt/openwrt/commits/master/target/linux/x86/64/target.mk) I guess that is a pretty safe bet.

However I think it wouldn't hurt to duplicate 4 NIC drivers to have a complete list of available drivers here.

targets/x86-64
rotanid: question is, does it make sense to include all of those? at least i know of chips that work with  ath10k-firmware-qca9888 and ath10k-firmware-qca988x but better with the latter

mweinelt: https://de.varia-store.com/produkt/31905-compex-wle600vx-802-11ac-a-b-g-n-minipci-express-karte.html

I own the WLE600VX and it identifies as `Network controller: Qualcomm Atheros QCA986x/988x 802.11ac Wireless Network Adapter`, seems to be a `QCA9882`. Is often sold with PC Engines APU2 I guess.

PR #1385, opened 2018-05-06
ramips-mt7621: add support for Ubiquiti EdgePoint R6

rotanid:
this adds support for the Ubiquiti EdgePoint R6 which uses exactly the same hardware as the Ubiquiti EdgeRouter X-SFP
source: https://openwrt.org/toh/ubiquiti/ubiquiti_edgerouter_x_er-x_ka

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -302,6 +302,7 @@ ramips-mt7621
 
   - EdgeRouter X
   - EdgeRouter X-SFP
+  - EdgePoint R6
 
 ramips-mt7628
 ^^^^^^^^^^^^^
--- a/targets/ramips-mt7621
+++ b/targets/ramips-mt7621
@@ -34,6 +34,7 @@ factory
 sysupgrade '.tar'
 
 device ubnt-erx-sfp ubnt-erx-sfp
+alias ubnt-ep-r6
 packages '-hostapd-mini'
 factory
 sysupgrade '.tar'


oszilloskop:
I have learned that the EdgePoint R6 does not accept a UBNT **X-SFP** LEDE/GLUON initramfs image via the well-known console solutions. However, the EdgePoint R6 router can still be programmed using a UBNT X-SFP LEDE/GLUON initramfs **kernel** binary. This requires a serial TTL interface and a TFTP server.

You will get a UBNT EdgePoint R6 running with a UBNT X-SFP Gluon firmware by doing following:
1) Transfer a UBNT-X-SFP initramfs kernel binary into the SDRAM via TFTP.
2) Perform a sysupgrade of a UBNT-X SFP file via a web browser (http://192.168.1.1).

You will find a useable LEDE/GLUON UBNT-X-SFP initramfs kernel file here: https://github.com/oszilloskop/UBNT_ERX_Gluon_Factory-Image/blob/master/back-to-stock-ramips-mt7621-ubnt-erx-sfp-initramfs-kernel.bin

Note I:
Also please have look at https://github.com/oszilloskop/UBNT_ERX_Gluon_Factory-Image#per-konsole-zur%C3%BCck-zur-ubnt-stockfirmware. The procedure described there is to be applied with the UBNT-X-SFP initramfs kernel binary (instead of the stock firmwarel image described there).

Note II:
Es ist ein wildes Gefrickel :o)

rotanid:
@oszilloskop so you found no way to flash this device without a serial interface, right?

oszilloskop:
@rotanid : You are right.

rotanid:
ok, so let's close this PR until there's a way without a serial interface.

PR #1389, opened 2018-05-07
ar71xx: add support for TP-Link TL-WR940N v6

achterin:
This PR backports support for TP-Link TL-WR940N v6 to master.

As of now, this PR is untested, as my device is still on its way.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -259,7 +259,7 @@ ar71xx-tiny
   - TL-WR743N/ND (v1, v2)
   - TL-WR841N/ND (v3, v5, v7, v8, v9, v10, v11, v12)
   - TL-WR843N/ND (v1)
-  - TL-WR940N (v1, v2, v3, v4, v5)
+  - TL-WR940N (v1, v2, v3, v4, v5, v6)
   - TL-WR941ND (v2, v3, v4, v5, v6)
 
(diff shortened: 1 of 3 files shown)


achterin:
Testnode online: https://meshviewer.chemnitz.freifunk.net/#!v:m;n:b04e269026a4

rotanid:
did you only start it or did you explicitly test every part?
(i know, we need a list of what to test... i wrote one a few months ago on the freifunk forum... )
the hardware isn't really new, so testing the meshing stability/speed for situation without VPN may be enough

this device still has only 32megs ram and 4megs flash, so it won't be great now and it won't last 5 years

achterin:
I did some testing of course.
So far i tested client connection via lan & wlan with direct uplink or via mesh and did some speedtests. vpn was always enabled and the throughput was pretty stable. The only thing i did not test as of now is lan-mesh.

So far everything looks good to me, no load issues, no ram issues: https://stats.chemnitz.freifunk.net/d/000000003/node?var-node=b04e269026a4&from=now-24h&to=now&refresh=1m&orgId=1

If you insist on testing the lan-mesh, i'll do that tonight and report back.

rotanid:
thank you for your contribution

PR #1391, opened 2018-05-10
v2017.1.x: docs: add missing devices

achterin:
This PR adds 2 devices to the docs that were recently added.
Devices are:

- GL Innovations GL-AR300M
- Ubiquiti UniFi AC Mesh
- TP-Link TL-WR940N v5
- TP-Link TL-WR1043N/ND

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -134,6 +134,7 @@ ar71xx-generic
 * GL Innovations
 
   - GL-AR150
+  - GL-AR300M
   - GL-iNet 6408A (v1)
   - GL-iNet 6416A (v1)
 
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -204,6 +204,7 @@ ar71xx-generic
   - UniFi AP Pro
   - UniFi AP Outdoor
   - UniFi AP Outdoor+
+  - UniFi AC Mesh [#ath10k]_
```
NeoRaider: The name and the ordering are inconsitent: It should be called UniFi AP AC Mesh and be ordered between the Lite and the Pro.

achterin: But that device has no "AP" in its Name. Its just called "UniFi AC Mesh". I just checked their productpage.
You are right though about the ordering. Will fix that when i get back home.

rotanid: well, you're partly correct, it is also named UAP-AC-M by the vendor...
but upstream OpenWrt doesn't mention the "AP" so we should change this in master branch and your naming should be ok

achterin:
I've fixed the ordering and added 2 more devices, which are already supported.

rotanid:
LGTM, merged

Issue #1360, opened 2018-04-07
RFC: Adding Einverständnis zum Datenschutz to configmode


A-Kasper:
Because many Networks are using maps with nodes.json etc. it is necessary to ask for an Einverständniserklärung zum Datenschutz and inform what is stored where.
For me there are two ways to do this: 
1. need to register the node in the community and doing it that way,
or
2. add an optional Datenschutzerklärungspage to configmode with accepted hook and refuse connecting to mesh and meshvpn as long as its not accepted. 

I would like to have the second possibility.

T-X:
3. Add an explanation regarding what a Freifunk node is (especially: a node contributing to a public(!) network, therefore being publically visible) to your community website. Something people will (and will like to!) read before downloading a firmware.

I think a general explanation would be more helpful than a checkbox with a legal disclaimer which no one will read anyway.

A-Kasper:
It is not important that someone reads it, or better: it is not our problem if the Box is clicked without reading it. For legal reasons we need to ask to accept the conditions. Else we may not fullfill dsgvo requierements on map-servers. 

If there is a legal page in config mode we would be able to realize to inform our users as requiered. There is nothing to hide, so it's easy to inform what is saved where and why and ask to accept. doing this in gluon would be the easiest way.

a general explanation would not be enough. you need to need to accept it with clicking somewhere.

A-Kasper:
...and you need to care that you don't store data without Permission. So if you ask on a webpage, you'll need to track and register each node and owner and work with whitelists..I see many things why I don't want manual router registration.

A-Kasper:
yes. something like this. maybe also with respondd module sending the
accepted conditions so you could use data on maps etc if allowed and
elsewise skip them etc

2018-04-08 17:20 GMT+02:00 Ruben Barkow <notifications@github.com>:

NeoRaider:
Does is have to be a checkbox, or is a simple text like "By completing the installation of this node, you accept the conditions" with a link sufficient, as it is often done with terms and conditions in online shops?

In any case, @T-X's solution sounds much better: Add the terms and conditions for the GDPR to your website before allowing to download the images. Cluttering the firmware with legal texts is not acceptable in my opinion.

NeoRaider:
Also, which information would be covered by the GDPR? There is some basic topology data that must always be available, and that is not personal in nature. AFAICT, the "contact information" and possibly "location" fields are the only data that might be considered personal information.

[...]

PR #1394, opened 2018-05-13
gluon-config-mode-contact-info: provide enhancements for DSGVO

christf:
for german, english and french translation.
This patchset was created after discussing DSGVO with Bea on WCW.

Big thank you to @shiromarieke for the french translation.

this addresses #1360


christf:
... When requiring the contact field via site.conf these texts do not match any more.
There are two ways around that:
a) have the community that requires this also overwrite the texts. To me this seems reasonable. If there is a process set up by a community that demands a contact field, it should explained.
b) remove the option to require the contact field as mandatory. From a DSGVO perspective this is valid and it would make sure there is no inconsistency. However I see this as a limitation to gluon. There may be a valid use case.

NeoRaider:
I vote for removing the "mandatory" option again. It was added by @rubo77 referring to the Pico-Peering Agreement (PPA), which requires providing contact information to peers.

I don't see any compelling reason to follow the PPA to the letter: Freifunk and community networks in general are *our* networks - we can change the rules as we deem sensible.

package/gluon-config-mode-contact-info/i18n/de.po
```diff
@@ -16,12 +16,16 @@ msgstr "Kontakt"
...
+"not required to operate a node. If you chose to enter data, it will be "
+"stored on this node and can be deleted by yourself at any time."
 msgstr ""
 "Bitte hinterlege hier einen Hinweis, um anderen zu ermöglichen, "
 "Kontakt mit dir aufzunehmen. Beachte, dass dieser Hinweis auch "
 "<em>öffentlich</em> im Internet, zusammen mit den Koordinaten "
-"deines Knotens, einsehbar sein wird."
+"deines Knotens, einsehbar sein wird. Für den Betrieb sind diese "
+"Informationen nicht erforderlich. Eine Speicherung erfolgt auf diesem Router. "
+"Diese Daten können Durch Dich in diesem Menü eigenständig gelöscht werden."
```
NeoRaider: * "Durch" klein
* "Dich" klein (im Rest der Texte schreiben wir "Du" auch klein)
* "in diesem Menü" -> "auf dieser Seite" oder so

package/gluon-config-mode-contact-info/i18n/fr.po
```diff
@@ -16,11 +16,16 @@ msgstr "Informations de Contact"
 msgid ""
 "Please provide your contact information here to allow others to contact "
 "you. Note that this information will be visible <em>publicly</em> on "
-"the internet together with your node's coordinates."
+"the internet together with your node's coordinates. This information is "
+"not required to operate a node. If you chose to enter data, it will be "
+"stored on this node and can be deleted by yourself at any time."
```
NeoRaider: It is relevant that this data is also collected in a central location (map server), and potentially cached for a very long time?

The FFHL meshviewer will show all information of offline nodes for several months after a node has disappeared. Not caching this information at all would solve the problem, but it would also make the information less useful, as you can't contact the operator of an unexpectedly offline node.

NeoRaider:
Another thing I just noticed: you only change the translation files, but not the original message in the Lua code.

rubo77:
I would like to keep the mandatory option.

It would just have to be a second text block that is shown if `mandatory==true`

That text block should contain the right wording, that is needed, if you collect user-data and keep them for a certain time. As far as I understand, it is still allowed to collect data, you just have to claim how you store them and how users may delete them and so on.

christf:
@rubo77 I disagree. We should not make it easy to collect personal data. An option like this does make it very easy without protecting the community from the resulting pitfalls. I am with neoraider on that one.

package/gluon-config-mode-contact-info/i18n/fr.po
christf: DSGVO beschreibt "dauerhafte Speicherung". A cache is no permanently saved data (the closest translation I can come up with). OTOH several months defenitely is an issue wrt DSGVO because the intent of shutting down a node is not clear. It could have been done to not do Freifunk any more or to just stop providing their data. Ithink that however is a question of sane defaults for meshviewer  not for gluon.
At the same time it probably does not hurt to mention cache retention time (which should be in the order of days at most.)

A-Kasper:
This does not fullfill dsgvo..as already mentioned we need an agreement to use contactinfo and ips in mapservers. if this data is stored in nodes.json without permission it violates dsgvo. we need a permissiontext in gluon. If not mapservers violate dsgvo and thats a big problem.

rotanid:
i disagree. you as a firmware provider don't know what the public will do with the data. the node provides the data to the public, anyone can do anything with the data.
it's like putting it on your public website, you can't control what people/machines will do with it.
therefore, you shouldn't write anything specific, like about some map server someone might run.

christf:
@A-Kasper read the text. the permission is given. This hinges on the field not being mandatory. My rework will remove the ability to set contact info as mandatory.
Problem solved.

package/gluon-config-mode-contact-info/i18n/fr.po
christf: in #gluon, rotanid mentioned that this would not help because the data may be processed by anyone. Really, it is public, which the text already states.

In my rework I will attempt to make the consequence of this data being public more obvious.

A-Kasper:
@rotanid:
are you allowed to crawl mailaddresses from websites and use them in your databases? the answer is no. you need a permission.

@christf
for me it's mandatory. The sense of the ability to set this to mandatory didn't change with dsgvo. I just want nodes in my network where i'm able to contact the owner. also maps show ip adresses. these are also personenbezogene daten.

It is possible to show a legal text with permission hooks and then add this hook status in respondd info.

doing this the functionality would not be destroyed and communities would be able to provide dsgvo compatible services.

I don't understand why this easiest way to handle the legal issues (informing and requesting permission as it is requiered) is denied.

A-Kasper:
christf: what you are able to do is not the same as what you are allowed to do. it's important to respect that you are not allowed to store my data without permission. it is not important if you are able to do so. 

dsgvo says datenverarbeitung. if you do that you need one of the named permissions. I don't see which permission like "permission by law" matches here. 

Nobody likes this stuff, but why don't just ask for permission and inform as requested in dsgvo and everything is fine? Viewing a text in config mode and setting/sending a hook is doable

horstfffl:
for me just adding more "Kleingedrucktes" may help on the juristic side but imo its not enough.
we should be a good example and enable nodeowners to simply change what specific info is send via alfred/respondd.
this could be achieved with a datenschutz config-mode page where you have checkboxes for each of the data-sets meant to be send.

ps
checkboxes: opt-in or opt-out is another good question.

A-Kasper:
horstfffl: I agree, but time is running. For me the first thing is to do what is needed. This is the Kür.
If there is possibility with gluon I'll use the Introductiontext and ask for adding an "Einverstanden. " in the contact field in front of the address, but this is really dirty. My skills are not good enough to code it myself

christf:
@A-Kasper I am not asking for permission because it is simply not possible to handle. The data is public. Anyone can access it. 
If you require this data to operate a node you must ensure that DSGVO is kept and there is no actual decision for the user except to not run a Freifunk node. If iti s a voluntary setting the consent to make the data given is an actual option.

If you want to make this field mandatory in your community, feel free to do so. But how do you want to protect the data entered from scraping? This is not just a map-problem. There is no central contact@ for stuff that is being scraped de-centrally. Thus it is impossible to obtain permission to do that.


We should have an IP-address-discussion somewhere else. There will be a wiki page for this stuff shortly.

A-Kasper:
I don't understand where you see the problem.

As a Nodeowner I'm just using my own data. the dsgvo doesn't match for own data. --> no problem in statuspage and respondd answers

As a mapserver admin i have to follow the dsgvo. i'm allowed to use data if i have the permission. 
freifunk is an open mesh network. you can't ask for the permission on a specific server (or better it makes not much sense). the best place to ask is the node itself. 

I already presented a solution that fullfills dsgvo legal issues:
show a legal page/section in configmode, where communities can place their ppa and datenschutz texts. add two hooks to accept them. 
send hook status via respondd.

on mapservers filter out respondd answers without placed hook.

this solves the problem and i think it shouldn't be very complicate to implement. 

In my mind gluon should make it possible to stay on safe side for legal issues. Communities could still don't use such a package if they don't want it.

workarounds without such a package will be:
1. requesting to add a string like einverstanden in contact field and changing infotext
2. changing text for map hook with datenschutz and tell by swtting this hook the node is shown blabla...permission.

but this is dirty. i would like to have this simple "show legaltext" package and filter out nodes without hook on mapserverside

package/gluon-config-mode-contact-info/i18n/de.po
christf: done.

package/gluon-config-mode-contact-info/i18n/fr.po
christf: and done.

mdrenger:
while the discussion #1360 does not seem to have achieved consensus, this PR should not be merged.

- transparency
being clear and transparent is a nice thing we should do. The firmware should continue have an option to make contact information mandatory. Freifunk groups may ask for consent or operate on legitimate interest (1.b).

- Maps
Asking for consent on behalf of map server admins should not be part of the firmware. Anyone can operate a map server and firmware providers are not identical with map admins. Network operators can argue legitimate interest to harvest that data.

Please refrain from asking for consent unnecessarily.

mdrenger:
also, if you want to go down that road of additional non-network related information on respondd:
please make sure to ask node operators to provide mandatory site information ( name, street, location and electronic contact ) as well as a data privacy statement for services provided by the node.

package/gluon-config-mode-contact-info/i18n/de.po
christf: sorry for this, somehow the correction didn't make it into the commit. This is fixed now for good.

christf:
We do not need consensus to merge a PR. I will not merge it myself because I have raised it.

This PR achieves that gluon is DSGVO-proof according to a lawyer specializing in Datenschutz. If that is a goal of the project then it should either be merged or an alternative approach should be pursued by someone else who has also at the very least consulted an authority in the field.

A-Kasper:
gluon is dsgvo proof. yes.
offering a mapserver collecting and storing data is not. many communities are using mapservices.

rotanid:
i added this to the milestone 2018.1 as it is rather important for many people on the legal side.

@A-Kasper i want to separate the discussions between this PR, making Gluon "DSGVO-proof", and your suggested changes, allowing for some consent regarding a mapserver or similar services like a nodelist. please use #1360 for your ideas or open an own PR for the suggested changes.

@christf is the latest version build&run-tested?
also, did you ask the lawyer you talked to about the nodes which already have that information on them? the people running those nodes added the contact info without reading the text we now want to add.
i know it's an unpleasant topic and we might decide that we ignore it, but i won't decide this alone.

A-Kasper:
because i don't know a community without mapserver I need to Nack this PR. The common gluon ökosystem is not dsgvo proof with this. it's a dealbreaker because we would still need to shutdown these really common services Or do some stupid hacks with other gluon texts and settings. 

We need a solution for more than the firmware.
Also this PR doesn't care about nodes in service. the owner never sees the text and does something to accept it. Also DSGVO wants extra checkboxes. or something you active click to accept datenschutz

christf:
@rotanid The fact that many users will have previously entered information was not considered in our discussion. 
This PR was build- and run-tested with german, english and frencht language settings in the browser.

rubo77:
In the [example for the community-specific translation](https://github.com/freifunk-gluon/gluon/blob/master/docs/site-example/i18n/de.po#L36)s in config mode we could add an option to agree to the terms in the email message body, that is generated by the link in the reboot page.

This already can be done by all communities i.e. like this: https://github.com/freifunk-kiel/site-ffki/commit/11dba734b5f0c2e09dc0416217961cde2ab38f7c

rotanid:
as neoraider also gave his consent via IRC, i'll merge this now.
this does explicitly NOT mean, that other/further changes are impossible, they will always be possible if there's consensus among the people with commit rights, just open a new pull request with a decent elaborated description

Issue #1378, opened 2018-05-03
HW support for GL.iNet AR-750


Commifreak:
-> https://www.gl-inet.com/ar750/

OpenWRT supports it: https://openwrt.org/toh/hwdata/gl.inet/gl.inet_gl-ar750 since commit: https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=2e5252d346e2ec832a203af778b5c1d949f0ae5f

Is there a chance that gluon/FF support it?

mweinelt:
OpenWrt support was added after LEDE 17.01 and has not been backported. Therefore support is only possible until after Gluons master branch was upodated to the next OpenWrt release or if the device support was backported to 17.01, but it is quite late for that imo.

mweinelt:
If the device was in OpenWrt master you will need to do a backport to lede-17.01 first.

achterin:
Why would I have to do that? Right now the backport for "TP-Link Archer C7 v4" is almost accepted and the code for this comes straight from OpenWrt master as well.
Sorry if this is a dumb question as I am fairly new to the gluon development.

rotanid:
@mweinelt as upstream does reject almost any hardware related backport, we don't apply this restriction as far as i know.

Commifreak:
@achterin I've tried ti build gluon with the ffharz site:

```
make GLUON_TARGET=ar71xx-generic GLUON_REGION=eu
```

which exited normally. I've got several images now, but none that has AR750 in it. Where do I find an image for the AR750 to flash it with gluon? Am I missing something?

...time passes...

Got it! ->
```
./output/images/sysupgrade/gluon-ffharz-0.9.1~20180504-gl-ar750-sysupgrade.bin
```

But thats for sysupgrade only, where's the factory image?

pbriesch:
So far as I know there is OpenWrt/LEDE preinstalled. Therefore no factory image is needed, only sysupgrade.

Commifreak:
Yep, just read that on the openwrt device description. I'll give it a try. Stay tuned ;)

achterin:
I think I've messed up the port config. There was an openwrt change on master, that was not applied onto lede-2017.1, that changed the port naming.
I've updated my branch and it would be nice if you could rebuild and test again.
Sorry for the trouble.

Commifreak:
No problem, I build it again asap and let you know if it works!

Commifreak:
[...]

PR #1395, opened 2018-05-14
ar71xx-generic: Add support for GL.iNet GL-AR750

achterin:
This PR backports hw support for GL.iNet GL-AR750.
The backport against 2017.1.x got already tested by @Commifreak and he will test this patchset asap. As of now, this is should be seen as untested until he reports back.

Fixes #1378

PR #1396, opened 2018-05-14
2017.1.x: ar71xx-generic: Add support for GL.iNet GL-AR750

achterin:
This PR backports hw support for GL.iNet GL-AR750 to v2017.1.x.

This PR got already tested by @Commifreak as can be seen in ticket #1378. The master backport still awaits his testing.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -135,6 +135,7 @@ ar71xx-generic
 
   - GL-AR150
   - GL-AR300M
+  - GL-AR750 [#ath10k]_
   - GL-iNet 6408A (v1)
   - GL-iNet 6416A (v1)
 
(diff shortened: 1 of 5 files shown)


NeoRaider:
We should really limit hardware backports to a minimum. Either a device is really well-suited for community networks (low-priced powerful hardware, easily available, ...), and there are no other comparable devices, or the maintenance effort is just not worth it.

rotanid:
@NeoRaider what is the core of your comment, a NACK? i think we should accept backports for our newest stable branch (which still is v2017.1.x at the moment) as long as the support is working good and it doesn't affect too much (like updating mt76 would)

NeoRaider:
My opinion on the matter is:
1. If it can be backported to master, a v2017.1.x backport is usually okay as well.
2. Don't backport at all, wait for the next OpenWrt release, unless a device is "important".

Commifreak:
The master BP seem to do its job:

bildschirmfoto 2018-05-16 um 16 18 49

rotanid:
the comment i just left at #1395 does affect this PR even more, as in stable release branches we absolutely don't want to break anything...

> i just saw that one of the patches you like to merge also touches code that is relevant to already supported devices.
> i'm considering to require a test of these after the changes, too... but maybe you have good reasons why this isn't necessary?

PR #1404, opened 2018-05-23
ar71xx: add support for Alfa AP121F

CodeFetch:
Adds support for Alfa AP121F USB powered travel router (looks like an USB ethernet adapter).
It has 16 MB flash and 64 MB memory and low power consumption. There is only a sysupgrade file for it. The description of the flashing procedure is in the commit message.
Everything seems to work and the primary MAC is correct, but maximum TX power is 17 dBm and wireless has 1x1 chains (max. 150 MBit/s).
The case be opened easily to access the UART interface.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -112,6 +112,7 @@ ar71xx-generic
 * ALFA Network
 
   - AP121
+  - AP121F
   - AP121U
   - Hornet-UB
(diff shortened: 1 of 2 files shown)


rotanid:
thanks!
off-topic, do you know where to buy this one?

PR #1405, opened 2018-05-24
Backport: GDPR changes backported to v2017.1.x

MPW1412:
Hello,

I just backported @christf's [GDPR-Commit](https://github.com/freifunk-gluon/gluon/commit/ab7f1acd0c83a77936c452eab1c6b3e9dcc612b3) from Gluon-Master to v2017.1.x so it can be used by the communities to build new firmware.

I compiled and tested firmware based on this backport: https://firmware.freifunk-muensterland.de/domaene16/versions/v3.3.3/

Regards,
Matthias


rotanid:
@MPW1412 you used @christf as author of one of your three commits, but changed the commit message rather drastically, i don't think that is OK.
please either do only one commit with yourself as the author and honor the work by others in the extended commit message, e.g. by adding
`Based-on-patch-by: Vorname Nachname <E-MAIL>`
or: leave the original commit almost untouched and add bigger changes as a second commit.
please also use commit messages that look like other Gluon commits, starting with a matching prefix, see the original commit on this GDPR issue as an example.

rotanid:
#1416

PR #1407, opened 2018-05-26
ar71xx: add support for TP-Link RE355

achterin:
This PR backports support for TP-Link RE355 v1.

This device is almost identical to the RE450. Both devices share the same problem: They get shipped with 64MB or 128MB RAM.
I only have a 64MB device to test with and as ath10k will not function properly the device will most likely reboot due to OOM errors.
So far everything (Mesh, Client connectivity, Uplink via eth) is working and a device with 128MB RAM will run just fine.

Testnode can be found here: https://meshviewer.chemnitz.freifunk.net/#!v:m;n:18a6f7450cce

### DIFF
--- a/patches/lede/0091-ar71xx-add-support-for-TP-Link-RE355.patch
+++ b/patches/lede/0091-ar71xx-add-support-for-TP-Link-RE355.patch
@@ -1,0 +1,338 @@
+From: Henryk Heisig <hyniu@o2.pl>
+Date: Tue, 6 Feb 2018 22:08:28 +0100
+Subject: ar71xx: add support for TP-Link RE355
+
+This device is identical as TP-Link RE450
+RE355 is a dual-band AC1200 router, based on Qualcomm/Atheros
+QCA9558+QCA9880.
+
+Specification:
+
+720/600/200 MHz (CPU/DDR/AHB)
+64/128 MB of RAM (DDR2)
+8 MB of FLASH (SPI NOR)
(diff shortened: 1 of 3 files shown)


rotanid:
we should only merge this as BROKEN for the reason you already wrote.
as discussed on IRC, the RE450 was merged without BROKEN before we knew about the existence of examples with only 64MB. we can't add broken now for this one, as it would leave those devices without future updates.

patches/lede/0091-ar71xx-add-support-for-TP-Link-RE355.patch
```diff
@@ -0,0 +1,338 @@
...
+Subject: ar71xx: add support for TP-Link RE355
+
+This device is identical as TP-Link RE450
+RE355 is a dual-band AC1200 router, based on Qualcomm/Atheros
+QCA9558+QCA9880.
+
+Specification:
+
+720/600/200 MHz (CPU/DDR/AHB)
+64/128 MB of RAM (DDR2)
```
mweinelt: 64/128 = 0.5 MB of RAM? :)

achterin:
I added the BROKEN-flag and a footnote to the docs.

NeoRaider:
Please stop backporting more devices. We already have too many hardware support patches in Gluon.

I will accept being outvoted by the other maintainers, but I'm against any hardware support that requires adding patches against OpenWrt/LEDE in Gluon.

PR #1409, opened 2018-05-31
Move contact field help to site localization for easier customization

RalfJung:
I guess this also needs to be mentioned in the released notes so that people don't forget to update their site localization when upgrading gluon?


rotanid:
i suggest do also do this  for the part `e.g. E-mail or phone number` , not only the longer text.
actually, i would prefer this for even more parts aside from the contact info, but that's another story.

RalfJung:
> i suggest to also do this for the part e.g. E-mail or phone number , not only the longer text.

Done.

package/gluon-config-mode-contact-info/luasrc/lib/gluon/config-mode/wizard/0500-contact-info.lua
```diff
@@ -5,16 +5,9 @@ return function(form, uci)
...
-		.. 'you. Note that this information will be visible <em>publicly</em> on '
-		.. 'the internet together with your node\'s coordinates. This means it can '
-		.. 'be downloaded and processed by anyone. This information is '
-		.. 'not required to operate a node. If you chose to enter data, it will be '
-		.. 'stored on this node and can be deleted by yourself at any time.'
-	))
+	local s = form:section(Section, nil, translate("gluon-config-mode:contact-help"))
 
-	local o = s:option(Value, "contact", pkg_i18n.translate("Contact info"), pkg_i18n.translate("e.g. E-mail or phone number"))
+	local o = s:option(Value, "contact", pkg_i18n.translate("Contact info"), translate("gluon-config-mode:contact-note"))
```
NeoRaider: No unqualified *translate* function exists anymore in the Gluon master, so this can't work. A quick grep for "site_i18n" should show you how it is done now.

I would prefer not to remove the old text, but only replace it if it is set in the site i18n, defaulting to the old text if unset. You can use the *_translate* function of the i18n module that returns *nil* if a translation is unset; this is preferable for all site i18n texts anyways, so the *gluon-config-mode:contact-note* is never actually shown.

rotanid:
oh, it is? that sounds nice, documentation on this would be great.

NeoRaider:
@rubo77 This is nonsense. It was "working" with v2017.1.x because of the arbitrary ordering the i18n files were loaded. In the master this is fixed, and independent i18n files can't override each other anymore.

package/gluon-config-mode-contact-info/luasrc/lib/gluon/config-mode/wizard/0500-contact-info.lua
RalfJung: Oh sorry, I guess I shouldn't just have ported this from the 2017.1.x branch.

I have found calls to `site_i18n._translate('gluon-config-mode:reboot')`, so I guess that's what I am supposed to use now. But where does the default text (in case nothing is set) go then?

NeoRaider: The default text can stay as is, so the call would become something like
```lua
site_i18n._translate('gluon-config-mode:reboot') or pkg_i18n.translate('... the current text ...')
```

Don't forget to initialize site_i18n, as it is not used in *0500-contact-info.lua* so far.

RalfJung:
Okay, I fixed the `site_i18n` and added a fallback for `contact-note` and tested that this actually works.

I can squash later if you want; for now I wanted to avoid force-pushing. Also, should there be a fallback for `contact-help`?

NeoRaider:
I'd like to have fallback for everything - basically, don't change the package pot/po files at all, so nobody is required to adjust their site i18n.

NeoRaider:
Merged in 1649635be46f37e62802f92a8ee53f831ed0e31b. I made a minor modification to your last version, so pkg_i18n is only queried for the help string when there is no override defined in site_i18n.

RalfJung:
Awesome, thanks :)

rotanid:
@RalfJung thanks for the idea and work! it allows me to drop a patch and if my upcoming PRs (based on your work) are accepted, even two more patches :)

Issue #1258, opened 2017-11-12
Kernel error: unregister_netdevice: waiting for mesh-vpn to become free. Usage count = 1


RalfJung:
I am seeing the following kernel error on some of my nodes:
```
unregister_netdevice: waiting for mesh-vpn to become free. Usage count = 1
```

In the past, I used to see an error very much like this on gateways; usually in the context of doing a `reboot` and then finding the machine stuck (and recoverable only via the server console). Batman had some bugs where it did not properly detach some interfaces. However, I haven't seen this error on gateways ever since we updated them to Batman 2017.x.

I am running two nodes, both with Gluon 2017.1.3. One, for testing, is an x86 node running in a VM via libvirt. There, I can pretty much reproduce the above error as follows:
* Start the VM, make sure it is connected.
* Put the *host* (i.e., my laptop) into stand-by.
* Wait a little while (I usually get into the bus and drive home, or something like that).
* Resume the host. Check the VM output, and I see the above error.

I then have to do a `reboot` of the VM (i.e., the node) to get it into a working state again.

Just now, I saw this error the first time on real hardware (TP-Link TL-WR841N/ND v10). I was trying to do `/etc/init.d/tunneldigger restart`, and it got stuck. `logread` shows that tunneldigger got the termination signal, but then failed to terminate. I sent `kill -KILL` to the result of `pidof tunneldigger`, to no avail. `dmesg` shows the above error, and it kept repeating. Ultimately I had to `reboot` the device.

Now I am a little worried that other devices might run into a similar issue when stopping services for an auto-update, and will just not work until someone notices and manually power cycles them.

T-X:
[...]

PR #1411, opened 2018-06-02
batman-adv: Fix best gw refcnt after netlink dump

ecsv:
This fixes a problem similar to the one described in #1258.

This patch will not apply on 2017.1.x. An adjusted version can be found in https://github.com/FreifunkVogtland/gluon/tree/batadv/gw_refcnt_fix-2017.1.x

----

A reference for the best gateway is taken when the list of gateways in the mesh is sent via netlink. This is necessary to check whether the currently dumped entry is the currently selected gateway or not. This information is then transferred as flag BATADV_ATTR_FLAG_BEST.

After the comparison of the current entry is done, batadv_*_gw_dump_entry() has to decrease the reference counter again. Otherwise the reference will be held and thus prevents a proper shutdown of
the batman-adv interfaces (and some of the interfaces enslaved in it).

### DIFF
--- a/patches/packages/routing/0002-batman-adv-Fix-best-gw-refcnt-after-netlink-dump.patch
+++ b/patches/packages/routing/0002-batman-adv-Fix-best-gw-refcnt-after-netlink-dump.patch
@@ -1,0 +1,115 @@
+From: Sven Eckelmann <sven@narfation.org>
+Date: Sat, 2 Jun 2018 16:25:16 +0200
+Subject: batman-adv: Fix best gw refcnt after netlink dump
+
+A reference to the best gateway is taken when the list of gateways in the
+mesh is sent via netlink. This is necessary to check whether the currently
+dumped entry is the currently selected gateway or not. This information is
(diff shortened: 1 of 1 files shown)


rotanid:
i tested this by using your branches:
https://github.com/FreifunkVogtland/gluon/tree/batadv/gw_refcnt_fix
https://github.com/FreifunkVogtland/gluon/tree/batadv/gw_refcnt_fix-2017.1.x
i couldn't reproduce my issue described in #1258 anymore with these branches, while another device with unpatched v2017.1.x running the test at the same time still shows the issues
thanks for your fix, ACK from my side

NeoRaider:
Thanks for finding and fixing this!

rotanid:
@ecsv feel free to do a PR for v2017.1.x Gluon branch, as i already tested your fix for that branch.

rubo77:
Es this bug also present in 2016?

rotanid:
@rubo77 no.

Issue #1243, opened 2017-10-22
High load on some devices after v2017.1.x update


mweinelt:
The devices were fine before, but with 2017.1.x high loads appeared. They seem to originate somewhere in the kernelspace.

Only some models are affected, but then all devices of that model experience this issue:

- TP-Link WR842ND v2
- TP-Link WR1043ND v1
- Ubiquiti Nanostation Loco M2

Probably more, but since our Grafana is currently down it's cumbersome to find more.

NeoRaider:
I've seen the issue on a WR841ND (I think it was v9 or v10). Possibly, all models with 32MB are affected?

mweinelt:
Some 841 V9 are affected, not all though.
The 841 V10 looks unaffected.

rotanid:
we also noticed that devices with more mesh neighbours are more likely to be affected - what a surprise!

Tarnatos:
I can confirm that dir Freifunk Nord

mweinelt:
Devices on the FFDA network with load > 2.0

```json
{
  "TP-Link TL-MR3420 v1": [
    8.63
  ],
  "TP-Link TL-WR841N/ND v11": [
    7.85,
    12.47,
    3.44,
    3.9
  ],
  "TP-Link TL-WR940N v4": [
    4.48
  ],
  "Ubiquiti AirRouter": [
    3.09
  ],
  "TP-Link TL-WR710N v1": [
    6.2
  ],
  "TP-Link TL-WR842N/ND v2": [
    2.28,
    12.43,
    8.98,
    15.45,
    8.98,
    3.47,
    7.22,
    10.07,
    12.14,
    9.2,
    7.25,
    8.48,
    2.91,
    6.22,
    8.91,
    3.14,
    2.13,
    9.9,
    10.92
  ],
  "Ubiquiti PicoStation M2": [
    9.97
  ],
  "Linksys WRT160NL": [
    4.69
  ],
  "TP-Link TL-WR710N v2.1": [
    4.19,
    11.26
  ],
  "TP-Link TL-WR1043N/ND v1": [
    6.98,
    3.57,
    6.25,
    2.23,
    5.62
  ],
  "TP-Link TL-WR841N/ND v9": [
    4.98,
    4.55,
    4.08,
    2.67,
    4.89,
    2.57,
    9.66,
    7.81
  ],
  "TP-Link TL-WA850RE v1": [
    7.71
  ],
[...]

PR #1418, opened 2018-06-05
RFC: gluon-mesh-batman-adv: unconditionally rate limit ARP

T-X:
With a reasoning similar to "gluon-ebtables: unconditionally segment
IGMP/MLD" also make the ARP rate limiting mandatory.

It turned out to be very common that there is a client device with
an application scanning the IP subnet, causing congestion and high
load for any community which did not add the gluon-ebtables-limit-arp
package yet.

Therefore just always add gluon-ebtables-limit-arp via a dependency.

Signed-off-by: Linus Lüssing <linus.luessing@c0d3.blue>

### DIFF
--- a/package/gluon-ebtables-limit-arp/Makefile
+++ b/package/gluon-ebtables-limit-arp/Makefile
@@ -8,7 +8,7 @@ include ../gluon.mk
 
 define Package/gluon-ebtables-limit-arp
   TITLE:=Ebtables limiter for ARP packets
-  DEPENDS:=+gluon-core +gluon-ebtables gluon-mesh-batman-adv \
+  DEPENDS:=+gluon-core +gluon-ebtables \
 	+@GLUON_SPECIALIZE_KERNEL:KERNEL_BRIDGE_EBT_LIMIT \
 	+@GLUON_SPECIALIZE_KERNEL:KERNEL_BRIDGE_EBT_MARK \
 	+@GLUON_SPECIALIZE_KERNEL:KERNEL_BRIDGE_EBT_MARK_T
--- a/package/gluon-mesh-batman-adv/Makefile
+++ b/package/gluon-mesh-batman-adv/Makefile
@@ -7,7 +7,7 @@ include ../gluon.mk
 
 define Package/gluon-mesh-batman-adv/common
   PROVIDES:=gluon-mesh-batman-adv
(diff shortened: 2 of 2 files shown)


NeoRaider:
* You're creating a circular dependency - this will likely not even build
* I consider our ebtables support too fragile to make this default, in particular regarding the limit reset issue, and the general unfitness of the ebtables kernel interface for frequent changes...

T-X:
Changes in v2:
* fixed circular dependency

T-X:
@NeoRaider: I fully agree with you, that this is far from being a beautiful approach. However I'm worried that too many communities are not selecting this package (probably because they are not aware of it) and are therefore having serious issues in their network.

I also fear that this makes evaluating feedback from communities regarding #1243 quite a bit more complicated.

To me, pragmatically, the network health and stability improvements of always having this package selected seems to outweigh the potential (so far theoretical?) stability concerns regarding the design of this approach. To me it seems worth it, at least until someone comes up with a better approach.

rotanid:
i would agree with @T-X  - as long as no one proves that he has very good technical reasons to not use this package, we could revert the change then

rotanid:
during todays meeting it was decided to postpone this and not include it in the upcoming 2018.1 release. 
the idea will be tested though during the development cycle of the 2018.2 release, but with a slightly different implementation where you are able to remove the package optionally

2tata:
Maybe its better to leave the dependency on `gluon-mesh-batman-adv` for `gluon-ebtables-limit-arp`. Also remove `gluon-ebtables-limit-arp` as a dependency from `gluon-mesh-batman-adv` again. (all in all reset the changes) and add `gluon-ebtables-limit-arp` to gluon features. So the package will be automatically selected if `gluon-mesh-batman-adv` has selected and individual communities are still able to deselect the package by adding:

```
GLUON_SITE_PACKAGES := \
        -gluon-ebtables-limit-arp
```

rotanid:
@2tata something like this was meant with "but with a slightly different implementation where you are able to remove the package optionally"

2tata:
@rotanid Yes, the optionally package remove will be possible like I wrote above. (the minus at the beginning)

rotanid:
@2tata feel free to open a PR for this

rotanid:
closing as #1516 was merged

PR #1424, opened 2018-06-08
ar71xx: add support for OpenMesh A40/A60

ecsv:


rotanid:
info about the A60:
https://openwrt.org/toh/hwdata/open-mesh/open-mesh_a60

PR #1425, opened 2018-06-08
Add FF3L site config to the list of communities

J0WI:


docs/user/site.rst
```diff
@@ -617,7 +618,7 @@ This is a non-exhaustive list of site-repos from various communities:
 * `site-ffho <https://git.ffho.net/freifunkhochstift/ffho-site>`_ (Hochstift)
 * `site-ffhgw <https://github.com/lorenzo-greifswald/site-ffhgw>`_ (Greifswald)
 * `site-ffka <https://github.com/ffka/site-ffka>`_ (Karlsruhe)
-* `site-ffki <http://git.freifunk.in-kiel.de/ffki-site/>`_ (Kiel)
+* `site-ffki <https://git.freifunk.in-kiel.de/ffki-site/>`_ (Kiel)
```
ecsv: Why did you change the URL for Kiel? This is not mentioned in the commit message and doesn't seem to belong here.

rotanid:
what @ecsv writes.
and also, please use a commit message that looks similar to the other commits changing the list of communities.
suggestion:
docs: add link to site config of Freifunk Dreiländereck

docs/user/site.rst
J0WI: I though it's not worth to create a separate commit. Spited into #1427.

J0WI:
Rebased.

PR #1440, opened 2018-06-18
2018.1.x: mac80211/ath10k-ct: reduce memory consumption of ath10k

ecsv:
ath10k uses a rather high number of buffers to communicate with the QCA firmware. Especially the HTC (host-target-communication) and HTT (host-target-transport) can take up a lot of memory when data is
transferred over a ath10k wifi link.

Even a 256MB device with three radios can go OOM (while sending to threedevices using UDP unicast/multicast packets) with the default buffer limits. Unfortunately, this also reduced the maximum 5GHz throughput on an IPQ40xx device without RPS/XPS optimizations from 494/432 Mbit/s for TCP transfers (download/upload) to 438/343 Mbit/s. These throughput values should still be good enough for a Freifunk installation.

----

This is a partial backport from OpenWrt  https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=cc189c0b7fa015978b04bb663a75b1da726376b5

### DIFF
--- a/patches/lede/0098-mac80211-limit-ath10k-memory-hunger.patch
+++ b/patches/lede/0098-mac80211-limit-ath10k-memory-hunger.patch
@@ -1,0 +1,73 @@
+From: Christian Lamparter <chunkeey@gmail.com>
+Date: Mon, 14 Nov 2016 04:57:38 +0100
+Subject: mac80211: limit ath10k memory hunger
+
+This is a problem with 128MiB RAM
+
+Signed-off-by: Christian Lamparter <chunkeey@gmail.com>
+[sven@narfation.org: remove QCA4019 part of change, adjust commit message]
+Signed-off-by: Sven Eckelmann <sven@narfation.org>
+
+Origin: backport, https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=cc189c0b7fa015978b04bb663a75b1da726376b5
+
+diff --git a/package/kernel/mac80211/patches/960-0010-ath10k-limit-htt-rx-ring-size.patch b/package/kernel/mac80211/patches/960-0010-ath10k-limit-htt-rx-ring-size.patch
+new file mode 100644
+index 0000000000000000000000000000000000000000..d49be4df703e8039509ecea23de8b5bd8e590765
+--- /dev/null
++++ b/package/kernel/mac80211/patches/960-0010-ath10k-limit-htt-rx-ring-size.patch
+@@ -0,0 +1,11 @@
++--- a/drivers/net/wireless/ath/ath10k/htt.h
+++++ b/drivers/net/wireless/ath/ath10k/htt.h
++@@ -199,7 +199,7 @@ enum htt_rx_ring_flags {
++ };
++ 
++ #define HTT_RX_RING_SIZE_MIN 128
++-#define HTT_RX_RING_SIZE_MAX 2048
+++#define HTT_RX_RING_SIZE_MAX 512
++ 
++ struct htt_rx_ring_setup_ring {
++ 	__le32 fw_idx_shadow_reg_paddr;
+diff --git a/package/kernel/mac80211/patches/960-0011-ath10k-limit-pci-buffer-size.patch b/package/kernel/mac80211/patches/960-0011-ath10k-limit-pci-buffer-size.patch
+new file mode 100644
+index 0000000000000000000000000000000000000000..54cc2ce7847d9256a78d3c60101ba6cd9baf496a
+--- /dev/null
++++ b/package/kernel/mac80211/patches/960-0011-ath10k-limit-pci-buffer-size.patch
+@@ -0,0 +1,38 @@
++--- a/drivers/net/wireless/ath/ath10k/pci.c
+++++ b/drivers/net/wireless/ath/ath10k/pci.c
++@@ -127,7 +127,7 @@ static struct ce_attr host_ce_config_wla
++ 		.flags = CE_ATTR_FLAGS,
++ 		.src_nentries = 0,
++ 		.src_sz_max = 2048,
++-		.dest_nentries = 512,
+++		.dest_nentries = 128,
++ 		.recv_cb = ath10k_pci_htt_htc_rx_cb,
++ 	},
++ 
++@@ -136,7 +136,7 @@ static struct ce_attr host_ce_config_wla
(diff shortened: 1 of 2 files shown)


ecsv:
A different thing which could be tested to work around OOMs while doing transfers over wifi is to disable the new AQM implementation. This should be doable by setting wake_tx_queue in ieee80211_ops to NULL. This would be in net/wireless/ath/ath10k/mac.c ath10k_ops for ath10k and drivers/net/wireless/ath/ath9k/main.c ath9k_ops for ath9k. But this is more an theoretical idea and not tested yet.

ecsv:
The change for ath10k-ct was also prepared for OpenWrt in PR https://github.com/openwrt/openwrt/pull/1077 (btw. different paths have to be patches in gluon 2017.1.x and OpenWrt 18.06)

greearb:
I would not commit something that makes performance worse for everyone.  Instead, use something like my fwcfg feature in the ath10k-ct driver to allow users to configure those sorts of things to the value they prefer.

rotanid:
@greearb in the case of Gluon and Freifunk, this speed reduction doesn't bother anyone.
but of course we follow upstream OpenWrt, so the decision has to be made there.

christf:
When given the choice of speed vs stability the right choice is stability.

rotanid:
@christf the thing is that upstream likely rejects this patch - and don't want to use patches (most of the time) which are not upstreamable...

rotanid:
as ibss (-> ath10k-ct) is deprecated anyways and ath10k devices with only 64M memory are BROKEN, i don't think it's worth another non-upstream patch.
but ultimately it's someone else's decision to make.

ecsv:
> as ibss (-> ath10k-ct) is deprecated anyways

Doesn't look this way. There are a couple of targets which select it unconditionally:

* targets/ar71xx-generic
* targets/ipq806x
* targets/x86.inc

> and ath10k devices with only 64M memory are BROKEN

The problem also affects ath10k-only dual or tri-radio devices with only 128/256 MB. You can find such devices for example in OpenWrt's ipq40xx target

In general: I will not invest any time in implementing a special ath10k-ct-only feature. Especially not when it depends on a ath10k-ct feature which the ath10k-ct author is not upstreaming to ath10k. And also not because HTT_RX_RING_SIZE_MAX is a preprocessor variable which is then (indirectly) used to fill a *const* struct. The latter is also true for the PCI buffer size change.

greearb:
On 07/06/2018 11:04 PM, Sven Eckelmann wrote:
>     as ibss (-> ath10k-ct) is deprecated anyways
>
> Doesn't look this way. There are a couple of targets which select it unconditionally:
>
>   * targets/ar71xx-generic
>   * targets/ipq806x
>   * targets/x86.inc
>
>     and ath10k devices with only 64M memory are BROKEN
>
> The problem also affects ath10k-only dual or tri-radio devices with only 128/256 MB. You can find such devices for example in OpenWrt's ipq40xx target
>
> In general: I will not invest any time in implementing a special ath10k-ct-only feature. Especially not when it depends on a ath10k-ct feature which the ath10k-ct author is not upstreaming to ath10k. And also not because HTT_RX_RING_SIZE_MAX is a preprocessor variable which is then (indirectly) used to fill a /const/ struct. The latter is also true for the PCI buffer size change.

I have tried to get the ath10k-ct features supported upstream, but I cannot force
Kalle to accept patches.  The last set of 3 patches I posted to implement some minor
support for ath10k-ct firmware are just rotting in patchwork in 'New' state.

Anyway, if someone comes up with some other way to configure things at run-time
that may be acceptable upstream, then post that patch instead.

Thanks,
Ben

-- 
Ben Greear <greearb@candelatech.com>
Candela Technologies Inc  http://www.candelatech.com

rotanid:
@ecsv maybe related? #1495

ecsv:
> @ecsv maybe related? #1495

When https://github.com/freifunk-gluon/gluon/issues/1495#issuecomment-409071951 is master then it could be related. At least I am not aware that the buffer size changes in ath10k were backported to LEDE (this is why I've created .#1440 with the changes for ath10k and ath10k-ct)

@blocktrron @mweinelt

mweinelt:
Yes, they are gluon master /openwrt 18.06.

Adorfer:
is possible to Backport to 2018.1.x or would this be considered as "new function"?

greearb:
If you want to discuss 802.11ax or anything else that is not related to memory usage, then please start a new topic.

Adorfer:
no need for that, since that is not my intention to discuss.

rotanid:
we discussed this patch on IRC today with @blocktrron and @christf 
we tend to reject this patch, as should be there with master branch and therefore v2018.2 release anyway.
also, we don't see a big problem in v2018.1.x branch with supported ath10k devices which would justify adding another patch.
what do you think about this @ecsv ?

ecsv:
Fine with me. Communities can just apply the patches manually when they have ath10k devices which have this kind of problem with low memory ath10k devices. The biggest problem I've seen were with ath10k dual/triband devices (ipq40xx) and these are not yet supported by 2018.1.x.

Just as reminder: It is not there in the master branch for ath10k-ct (GLUON_WLAN_MESH = 'ibss'; or ipq806x and x86)

blocktrron:
Good point on IBSS. If we consider IBSS as legacy i would view this as a non issue, but if this is not the case we should apply this patch to master.

ecsv:
> Good point on IBSS. If we consider IBSS as legacy i would view this as a non issue, but if this is not the case we should apply this patch to master.

Small remark: The master branch (openwrt 18.06) would need a slightly different patch (the ath10k-ct package is a little bit special and you need to patch the correct subdirectory in owrt 18.06): https://github.com/openwrt/openwrt/pull/1077

rotanid:
ok, so we'll close this PR and if someone wants to adjust the patch for 18.06 with ath10k-ct, we can consider a  new PR

ecsv:
> > Ok, will do that. Will take a while until the build is finished.
> 
> ok, thought you don't want to, after reading your comment in [openwrt/openwrt#1077](https://github.com/openwrt/openwrt/pull/1077)

I didn't want to implement any ath10k-ct only feature to enable/disable it at runtime. It was not about adjusting the minimal ath10k patch for the ath10k-ct in OpenWrt 18.06 (which I basically did already).

PR #1441, opened 2018-06-19
Show sha256sum to verify upgrade checksum

J0WI:
This changes the checksum shown in the upgrade page from MD5 to SHA-256. Upstream OpenWRT/LEDE is already doing this.

### DIFF
--- a/package/gluon-web-admin/files/lib/gluon/config-mode/view/admin/upgrade_confirm.html
+++ b/package/gluon-web-admin/files/lib/gluon/config-mode/view/admin/upgrade_confirm.html
@@ -12,7 +12,7 @@ You may obtain a copy of the License at
 <h2><%:Upgrade firmware%></h2>
 
 <p>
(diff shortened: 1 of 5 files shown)


rubo77:
shouldn't there be a check on both, md5 and sha256 for a certain period of time, until all nodes are on a new update system and all firmware is served with sha256 checksums?

mweinelt:
This does not relate to the autoupdater, so the change is probably just fine.

package/gluon-web-admin/files/lib/gluon/config-mode/view/admin/upgrade_confirm.html
```diff
@@ -21,7 +21,7 @@ <h2><%:Upgrade firmware%></h2>
 
 <p>
   <ul>
-    <li>md5sum: <code><%=checksum%></code></li>
+    <li>sha256sum: <code><%=checksum%></code></li>
```
NeoRaider: Hmm, simply labeling it "Checksum" would look nicer... in that case we should also make this translatable. Well, that can be done in another PR.

PR #1442, opened 2018-06-19
[RFC] Add support for bootloader images

blocktrron:
In light of new devices, which require special images for flashing the device thru the vendor bootloader. Factory images are currently described as "installable with the vendor GUI", which brings the need for an additional category.


NeoRaider:
Being installable through a GUI is not a hard requirement for a factory image. I'd simply make the image that makes it the easiest to switch from the stock firmware to OpenWrt/Gluon the "factory" image - if the easiest method is the bootloader, that's okay.

NeoRaider:
I see your point. Still, "bootloader" seems a bit specific for this special case; maybe call it "other" and allow some control over the image name in the target definition (so the string "tftp" will appear in the filename)?

rotanid:
as discussed during the last meeting which you attended, we don't take PRs against the next branch. so why shouldn't we immediately close this?

blocktrron:
I like this idea, it would also add the possibility of "ramboot" images, like the ones needed for the ocedo devices.

mweinelt:
RFCs are there for discussing a direction of development before wasting time in an approach that would never be merged. That's what this PR is for.

NeoRaider:
This PR was closed automatically, as the next branch does not exist anymore.

PR #1448, opened 2018-06-26
gluon-config-mode-geo-location: allow overriding geo location description text

rotanid:
allow overriding geo location description text

build and run tested on an TL-WR1043ND v2

inspired by #1409 and @RalfJung

### DIFF
--- a/docs/site-example/i18n/de.po
+++ b/docs/site-example/i18n/de.po
@@ -80,3 +80,8 @@ msgstr ""
 
 msgid "gluon-config-mode:contact-note"
 msgstr "z.B. E-Mail oder Telefonnummer"
+
+msgid "gluon-config-mode:geo-location-help"
+msgstr ""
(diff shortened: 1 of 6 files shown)


rotanid:
@rubo77 i don't fully understand.
are you talking about those 4 lines? https://github.com/tecff/gluon/blob/4361f5b762cd3a9e15506f10ca0df1b85e095b24/package/gluon-config-mode-geo-location/i18n/de.po#L20-L30
i don't see a reason to change those.

rubo77:
No

Change 

    "If you want the location of your node to be displayed on the map, you can "
"enter its coordinates here."

To the new short string

rotanid:
@rubo77 now i understand.
no, this can't be changed because we want the site conf setting to be optional, not mandatory.

NeoRaider:
Merged, with the same changes as in #1449.

PR #1449, opened 2018-06-26
gluon-config-mode-hostname: allow overriding hostname description text

rotanid:
allow overriding hostname description text

build and run tested on an TL-WR1043ND v2

inspired by #1409 and @RalfJung

### DIFF
--- a/docs/site-example/i18n/de.po
+++ b/docs/site-example/i18n/de.po
@@ -80,3 +80,7 @@ msgstr ""
 
 msgid "gluon-config-mode:contact-note"
 msgstr "z.B. E-Mail oder Telefonnummer"
+
+# please find the default translation for this text in package/gluon-config-mode-hostname/i18n/
+msgid "gluon-config-mode:hostname-help"
+msgstr ""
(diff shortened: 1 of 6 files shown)


rotanid:
@rubo77 you review contains no information at all, what's the matter?
changing https://github.com/tecff/gluon/blob/4361f5b762cd3a9e15506f10ca0df1b85e095b24/package/gluon-config-mode-hostname/i18n/de.po#L13-L14 ?
i don't see a reason for this

rubo77:
So we have to repeat the same long English string one less

rotanid:
@rubo77 now i understand.
no, this can't be changed because we want the site conf setting to be optional, not mandatory.

NeoRaider:
Hmm, I'm not happy with duplicating all those texts - new communities will start with a copy of *site-example/i18n*, so they would now get their own copies of the texts by default.

We could add those entries with empty translations to site-example/i18n/*.po (empty translations will make msgmerge happy, but will be treated like a nonexistent translation by po2lmo), and add comments there refering to the po files with the default texts in the package directory. This should be done for the already duplicated contact field texts as well.

rotanid:
@NeoRaider i have updated the commit with a possible implementation of your idea.
if this draft does not make you happy, please tell exactly what would make you happy or push to this branch yourself

PR #1454, opened 2018-07-02
docs: private-wlan: add uci command to set MAC

CodeFetch:
@rotanid noticed that a command  is missing in the docs to set the MAC address of the private WLAN when it is being configured manually. 

This commit adds the missing command to set the special Gluon MAC address of the private WLAN interface to the documentation.

### DIFF
--- a/docs/features/private-wlan.rst
+++ b/docs/features/private-wlan.rst
@@ -7,14 +7,19 @@ Please note that you should not enable ``mesh_on_wan`` simultaneously.
 The private WLAN can be enabled through the config mode if the package ``gluon-web-private-wifi`` is installed.
 You may also enable a private WLAN using the command line::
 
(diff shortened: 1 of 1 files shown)


rotanid:
thanks, but according to https://gluon.readthedocs.io/en/latest/dev/mac_addresses.html we should use "3" or "7" for the private wifi, not "4" - or am i missing something?

CodeFetch:
@rotanid I've updated the commit and checked it's working correctly now.

rotanid:
thanks for this better solution, but i strongly disagree with "0 is usually the ID of the 2.4GHz radio", we have loads of devices (like Archer C7, WDR4900, ...) where radio0 is the 5GHz radio...

CodeFetch:
@rotanid I've completely removed the comment about the radio ID as I think we can expect someone using SSH to understand what it is.

PR #1455, opened 2018-07-04
ar71xx-generic: Reduce SquashFS blocksize to 64K

ecsv:
Some 32 MB devices like the Nanostation M2 suffer from sudden high loads combined with a squashfs related OOM reboot:

    logd invoked oom-killer: gfp_mask=0x2420848, order=0, oom_score_adj=0
    CPU: 0 PID: 774 Comm: logd Not tainted 4.4.135 #0
    Stack : 804214dc 00000000 00000001 80480000 8182fa3c 80474803 804028d0 00000306
            804e378c 00001ade 00000040 00000000 00000000 800a7f10 00000006 00000000
            00000000 00000000 804063e0 80c69994 804e6542 800a5e8c 02420848 00000000
            00000001 801fd600 00000000 00000000 00000000 00000000 00000000 00000000
            00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
            ...
    Call Trace:
    [<800721cc>] show_stack+0x54/0x88
    [<800d5468>] dump_header.isra.4+0x48/0x130
    [<800d5c38>] check_panic_on_oom+0x48/0x84
    [<800d5d64>] out_of_memory+0xf0/0x324
    [<800d9888>] __alloc_pages_nodemask+0x6b8/0x724
    [<800d2960>] pagecache_get_page+0x154/0x270
    [<80134cb0>] __getblk_slow+0x15c/0x374
    [<80160418>] squashfs_read_data+0x1c8/0x6e8
    [<80164628>] squashfs_readpage_block+0x32c/0x4d8
    [<801622a4>] squashfs_readpage+0x5bc/0x6d0
    [<800dd030>] __do_page_cache_readahead+0x1f8/0x264
    [<800d479c>] filemap_fault+0x1a8/0x458
    [<800efc1c>] __do_fault+0x64/0xd0
    [<800f2824>] handle_mm_fault+0x4a4/0xb40
    [<80076e98>] __do_page_fault+0x134/0x470
    [<80060820>] ret_from_exception+0x0/0x10

Reduction of the SquashFS blocksize should reduce the problem slightly on these nodes. The image size will increase slightly but this should be no problem for ar71xx-generic (in contrast to ar71xx-tiny).

----

This is currently  (since 2018-07-04 14:00) on https://stats.freifunk-vogtland.net/d/ffv_node/nodeinfo?var-node=002722c0bf5b&orgId=1&from=1530688589748&to=now-1m


CodeFetch:
This is a sane way to save RAM and lower the time for decompression for devices which have more flash, but it is hard to fit such an image to 4MB (ar71xx-tiny). We should change it for ar71xx-generic only.

rotanid:
all targets supported by gluon except ar71xx-tiny don't have devices with <8M flash.
therefore i would agree with @ecsv to enable it for every target, except ar71xx-tiny.

ecsv:
Since I don't know what is now expected, I've prepared three branches:

* https://github.com/FreifunkVogtland/gluon/tree/ar71xx-generic/squashfs_64k (this PR; only ar71xx-generic)
* https://github.com/FreifunkVogtland/gluon/tree/generic/squashfs_64k (all devices which use squashfs)
* https://github.com/FreifunkVogtland/gluon/tree/generic-notiny/squashfs_64k (all devices which use squashfs - except ar71xx-tiny)

CodeFetch:
It's a relatively small difference in memory foot print if you use 64K or 256K and have no memory pressure, but 64K blocks can be decompressed much faster. You don't need to read a 256K block for reading a 20 bytes file. So I think it is the combination of both differences that let it perform better. The disadvantage is a worse compression of course and I think if we still want to support 4MB devices in the long term we need to use 256K as the compression ratio is much better. For 4MB devices a better workaround might be to not compress incompressible files like executables. They can then be read without decompression and thus with an arbitrary block size as we need the fixed block size only for decompression. It is slightly more efficient regarding memory usage and increases reading speed on page faults.

As long as we don't have the ability to use uncompressed files, using 64K for devices with more than 4MB flash seems reasonable to me. We should merge this PR as it is and have a look at how to increase free memory on devices with the load bug while cleaning and upstreaming the squashfs support for uncompressed files if possible.

rotanid:
we decided on IRC to merge this now and see how it works out. if problems arise, we can revert it before the release.
if there are no problems, we can consider it for other targets, too.

PR #1458, opened 2018-07-08
ar71xx: ag71xx: reduce NAPI weight and default size of TX/RX ring buffers

CodeFetch:
These values were even lower previously and Gluon worked fine. They were increased to get more throughput on gigabit capable devices. Most ar71xx devices have a gigabit network card, but no gigabit ports and it is unlikely that we will have Gluon networks with such throughputs anytime soon.

As to https://lwn.net/Articles/139884/ a too large NAPI weight can
cause increased system load as a driver may turn off polling and
return to interrupt-driven mode if the quota was not filled in
the last polling cycle creating the need of more interrupts to 
be handled. This commit decreases the NAPI weight from 64 to 32.

As fastd is a UDP tunnel with a constant stream of Batman management packets running over this interface which does not fill the quota often, we see many interrupts.

Furthermore this commit decreases the default TX/RX ring buffer
sizes from 128/256 to 64/128 as the buffers will unlikely be 
filled with a throughput <= 100 Mbit/s.

### DIFF
--- a/patches/lede/0102-ar71xx-ag71xx-reduce-NAPI-weight-and-default-size-of-TX-RX-ring-buffers.patch
+++ b/patches/lede/0102-ar71xx-ag71xx-reduce-NAPI-weight-and-default-size-of-TX-RX-ring-buffers.patch
@@ -1,0 +1,38 @@
+From: Vincent Wiemann <me@codefetch.de>
+Date: Sun, 8 Jul 2018 00:12:08 +0200
+Subject: ar71xx: ag71xx: reduce NAPI weight and default size of TX/RX ring buffers
+
+As to https://lwn.net/Articles/139884/ a too large NAPI weight can
+cause increased system load as a driver may turn off polling and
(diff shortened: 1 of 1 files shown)


rotanid:
@CodeFetch what about cases where the private wifi functionality is used, does this still allow to use more than 100 MBit/s ?
but if i'm not mistaken, this is no hard limit to 100 MBit/s with these changes?

CodeFetch:
@rotanid As far as I understand you can still get ~300 Mbit/s with weight 32. 
Decreasing the buffers will make it more likely that packets are being dropped under high load when the throughput is also high. As I read somewhere the buffer size should automatically increase up to the max value if that happens - apart from that I haven't found it in the code and don't know where too look, yet I think we won't see any issues.

Btw I've just realized that there is a patch in OpenWrt already that reduces the NAPI weight to 32 and there have been other reasonable patches that might increase ag71xx performance. I'm going to add them to this pull request...

PR #1459, opened 2018-07-09
ar71xx/ag71xx: backport patches

CodeFetch:
This commit backports patches from OpenWrt's master branch which
increase cache performance, reduce NAPI weight and fix bugs.


NeoRaider:
Merging this to master doesn't make sense, as I intend to switch the master to OpenWrt 18.06 as soon as possible.

AFAICT, this does not fix any bugs that are relevant for us, so this is just optimization. I don't really like adding so many patches to a stable branch if it doesn't fix serious problems... Do you have any numbers on the overall performance increase?

CodeFetch:
@NeoRaider Okay, I close it and wait for 18.06. Initially I only wanted to decrease the NAPI weight and the size of the ring buffers, as it is too high for our use case. Thought it would be a good idea to just get it up to date as there were many improvements, but I don't believe that it is a big difference. You know ag71xx much better... Do you think the MDIO memory leak can affect us? Does the false sharing only affect SMPs?

Issue #1392, opened 2018-05-12
Add Support for TP-Link TL-WR841N (v13)


christf:
841 v13 have 8MB flash and 64 MB RAM and are sporting an mt76x8 chip. nbd just mentioned that support is in recent openwrt and mesh + AP should be working however testing is required. Let's add it as broken until we know what works and what doesn't.

NeoRaider:
We should do so after the v2018.1 release.

We are already backporting too many devices when it would be much easier to wait for the switch to the OpenWrt master. While a backport may be justified for the most promising devices, adding a known BROKEN device (AFAIK, there is no sane way to flash the 841v13 yet) only adds to the maintenance burden.

rotanid:
in Gluon master it would be rather broken, as we gave up on backporting the mt76 updates.

regarding flashing, it has to be done via TFTP which is not worse than the SSH-way for the already supported devices like UAP AC Lite, Pro, Mesh, ... - i don't think this would need to be marked as broken.

christf:
@rotanid I agree that just for the flashing method the broken flag is not justified. The suggestion for BROKEN stems from the fact that we don't know yet how these devices work in typical freifunk situations yet. In any case, this is a minor detail.

Adorfer:
if flashing would require RS232 console access (and making contact on the main PCB) prior to TFTP download the image: In this case i would agree as well to status "broken".
a procedure requiring the removal of screws/breaking seals/placing needles/soldering: not good. 

But as long as it can be done from the outside and proven working (even with some known glitches): Not BROKEN for me.

kpanic23:
Well, technically it is possible to flash the device via the stock web interface. The only problem is, that the stock firmware overwrites part of the image. This could theoretically be mitigated by adding 128k of padding to the image before the kernel. I have no idea if that would be possible...

>Tests showed that the GUI upgrade routine copies value of "Additional
Hardware Version" from existing firmware into offset "0x2023c" in
provided file, _before_ storing it in flash. In case of vendor firmware
upgrade files (which all include U-Boot image and two headers), this
offset points to the matching field in kernel+rootfs firmware part
header. Unfortunately, in case of LEDE factory image file which contains
only one header, it points to the offset "0x2023c" in kernel image. This
leads to a corrupted kernel and ends up with a "soft-bricked" device. 

[...]

PR #1470, opened 2018-07-10
ramips: add TP-Link TL-WR841N v13

blocktrron:
This PR adds support for TP-Link TL-WR841N v13.

Note that this PR is not ready for merging. Following things need to be done to fully support Gluon features:

TP-Link TL-WR841N v13 #1392 (calling everyone!)
- [x] Verify correct MAC
- [x] Verify correct autoupdater-filename

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -343,6 +343,7 @@ ramips-mt7628
 * TP-Link
 
   - Archer C50 v3 [#80211s]_
+  - TL-WR841N v13 [#80211s]_
 
 * VoCore
 
--- a/targets/ramips-mt76x8
+++ b/targets/ramips-mt76x8
@@ -4,6 +4,10 @@ device tp-link-archer-c50-v3 tplink_c50-v3
 factory
 extra_image -squashfs-tftp-recovery -bootloader .bin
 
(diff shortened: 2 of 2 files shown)


skorpy2009:
@blocktrron do you have any information regarding the performance?

blocktrron:
@skorpy2009 Performance is around 8 Mbit/s in an empty L2 domain.

2tata:
test for 841 v13:
Flashable from vendor firmware via tftp.
Sysupgrade works.
Can forget configuration (if applicable).
Seems having working autoupdate
gluon profile name must match image name:

        root@tp-wr841n-v13-test:~# lua -e 'print(require("platform_info").get_image_name())'
        tp-link-tl-wr841n-v13

Supports all wired network ports on the device.
Have correct port assignment (WAN/LAN).
Association with wifi AP is possible.
Association with 802.11s mesh works.
ap/mesh mode works parallel
power/sys led lit while the device is on and displaied config mode blink sequence.
Radio led shows activity.
Switchport leds is map to their respective port.

**Doesn't shows link state and activity!**

reset/wps button returns device into config mode.
primary mac match address on device label.

PR #1472, opened 2018-07-11
gluon-ebtables: build kernel with ebt_dnat and ebt_redirect

T-X:
The redirect and dnat target are needed for gluon-alt-esc-client to forward frames to the selected, alternative gateways.

### DIFF
--- a/package/gluon-core/Config.in
+++ b/package/gluon-core/Config.in
@@ -104,6 +104,10 @@ config KERNEL_BRIDGE_EBT_ARP
 	bool
 	select KERNEL_BRIDGE_NF_EBTABLES
 
+config KERNEL_BRIDGE_EBT_DNAT
+	bool
+	select KERNEL_BRIDGE_NF_EBTABLES
+
 config KERNEL_BRIDGE_EBT_IP
 	bool
 	select KERNEL_BRIDGE_NF_EBTABLES
@@ -124,6 +128,9 @@ config KERNEL_BRIDGE_EBT_MARK_T
 	bool
 	select KERNEL_BRIDGE_NF_EBTABLES
 
+config KERNEL_BRIDGE_EBT_REDIRECT
+	bool
+	select KERNEL_BRIDGE_NF_EBTABLES
 
 # Not all of the following modules are really required for Gluon, but fw3 pulls
 # them in, so we add them to the kernel config to reduce the number of loaded


CodeFetch:
Can you briefly explain why they are needed, please?

T-X:
gluon-alt-esc-client adds rules like the following into ebtables:

> $ ebtables-tiny -t nat -A ALT_ESC_DNAT -p IPv4 -j redirect --redirect-target ACCEPT

or:

> ebtables-tiny -t nat -A ALT_ESC_DNAT -p IPv4 -j dnat --to-dst "$exit4" --dnat-target ACCEPT

For that to work both ebtables-tiny needs to recognize dnat and redirect options. And the kernel needs to have the according netfilter code to be able to install them.

Regarding why those rules are used, have a look at the descriptions provided in #1094.

NeoRaider:
I'm hesitant to accept these patches, as I'd really like to avoid to have another major release with ebtables and iptables. Switching to nftables should be our priority.

rubo77:
Maybe you can set this pr to 2018.1.x branch and only accept it there.

For master we Will have to refactor the alt ESC package then to nftables

NeoRaider:
@rubo77 v2018.1.x is a stable branch, no new features are acceptable.

T-X:
From an egoistic perspective I would argue that it is a regression ;-). The gluon-alt-esc packages worked before and with v2018.1 they don't. And there is no possible workaround from the package itself.

Granted, that disregards that gluon-alt-esc is non-upstream. So that argument of mine is not a good one as non-upstream packages should not take support for granted.

On the other hand, if I remember correctly from the meeting some months ago the PR for gluon-alt-esc was neither merged nor closed because more community feedback was requested.

However, if it is not easy to play with and not easily testable by other communities, then I think it might be a little difficult to gather such feedback. In that sense, as long as the gluon-alt-esc PR is undecided, I would argue to interpret this change as a fix of a regression.

Does that make sense?

---

There is one more change that comes to my mind, an offer I could make, maybe as a compromise: I could half the size of this by dropping the ebtables redirect part, only adding ebtables dnat to ebtables-tiny. Instead I would need to backport this fix for the kernel then though:

[bridge: ebtables: fix reception of frames DNAT-ed to bridge device/port](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=cf3cb246e277d98987aa8d62ef2730dbee2f5fa7)

(and would need to update two lines in the gluon-alt-esc-client package, to use ebtables dnat in the place where I used ebtables redirect, too)

rotanid:
@NeoRaider i kind of agree with @T-X and also think, that we don't really lose anything accepting this PR...

NeoRaider:
Adding the two new symbols in `package/gluon-core/Config.in` should be fine, as the switch to nftables will likely take some time. As for the dependencies on these new symbols, having them in the alt-esc packages that actually use them seems more appropriate to me.

rotanid:
@T-X do you intend to adjust this PR as requested in the last comment?

christf:
at the moment gluon-alt-esc cannot be merged due to this rework. I would like to see gluon-alt-esc merged and I think this is the last piece missing.

T-X:
Changelog v2:
* moved dependencies from gluon-ebtables to gluon-alt-esc-client (other PR, #1094) and tested that change (works, thanks @neoraider!)

rotanid:
@T-X is there any reason to keep this PR here separate from #1094? you could add it as a separate commit as part of #1094 as it's only needed for that package and merging it without the package doesn't help anything

T-X:
Hi @rotanid. If #1094 were merged soon, then yes, then I could move this patch to the other pull-request indeed.

Otherwise, if #1094 would need a little longer and more discussion before merging, then it would be great to have this PR, #1472, and #1568 merged sooner to be able to use the gluon-alt-esc package via external modules in a vanilla Gluon release.

rotanid:
hm, seems fair. and @christf also acked on IRC, so let's merge this.

rubo77:
can this be ported back to 2018.1.2 please? so We can include the alt-esc package there again in our next firmware update.

rotanid:
we don't usually backport such things.

bobcanthelpyou:
@rubo77 it is part of branch v2018.2.x

PR #1474, opened 2018-07-11
build: add support for other images

blocktrron:
Successor for #1442 because of master branch switching to OpenWRT 18.06.

----

In light of future devices in need of special images for initial
flashing which are, in contrast to factory images, only flashable
from the bootloader, we add a new category for images which don't fit
into factory nor sysupgrade.

Devices in need for those images contain the AVM Fritz!Box 4040 and
various TP-Link Mediatek models.

Example for the Fritz!Box 4040:

 > other -squashfs-eva.bin bootloader.bin

results in the eva image landing in the "other" folder suffixed with
"bootlaoder.bin".

### DIFF
--- a/scripts/common.inc.sh
+++ b/scripts/common.inc.sh
@@ -38,6 +38,10 @@ sysupgrade() {
 	:
 }
 
+other() {
+	:
+}
+
 no_opkg() {
 	:
 }
--- a/scripts/copy_output.sh
+++ b/scripts/copy_output.sh
@@ -7,6 +7,8 @@ set -e
 
 default_factory_ext='.bin'
 default_factory_suffix='-squashfs-factory'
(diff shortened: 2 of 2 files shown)


NeoRaider:
Thanks, I reworked your code a bit to reduce code duplication and support multiple "other" images per device. I also renamed the target command to *extra_image* and added an extension argument.

Merged in 7460431faa5b53e9d1b5405bde1aa9a99c2667b9.

heini66:
what about the needed initramfs-kernel images for first flashing ubnt erx devices ?

NeoRaider:
@heini66 The ERX initramfs images can't be built with Gluon's usual .config, so this would need to be implemented as a completely separate target... I don't think it is worth the effort the implementation would take.

Issue #1434, opened 2018-06-13
Checklist for device integration testing


mweinelt:
What needs to be verified to do integration testing for a new device. Let's create a checklist.

@Brother-Lal wanted to take care of that.

Make use of our notes from the meeting here: https://md.darmstadt.ccc.de/gluon-device-support-policy#

rotanid:
additional ideas for this from my forum post: https://forum.freifunk.net/t/archer-c7-v-4-0/16384/7

Brother-Lal:
Checklist for now:

    - [ ] Flash from stock possible
    - [ ] Flash back to stock possible
    - [ ] sysupgrade is working (check with flashing + erasing settings)
    - [ ] autoupdater working (implied by sysupgrade working?)
    - [ ] Working WAN/LAN Ports + associated LED(s)
    - [ ] Wifi LED working and blinking
    - [ ] Wifi: 11s and AP Mode working in tandem / at least one of them
    - [ ] Is there a possibility for user triggered reset (button,POE-Reset, etc)
    - [ ] MAC: Is the official MAC(usually wlan0) the same as printed on the device/packaging?
    - [ ] ... More?

Any other additions I might have missed?

Adorfer:
MAC: clarifiy this point that this should be checked like "idem to meshid"  (=default routername-suffix, i do not know how this will be on the map for babel)

Adorfer:
- [  ] Config-mode is listeninging on a wired interface
- [  ] WAN/LAN ports are not inverted by accident. (perhaps change to "working WAN/LAN" to "working LAN/WAN and mapped correctly")

rubo77:
```
- [ ] tested with x clients
- [ ] tested with x Mbit throughput
```

csamsel:
```
- [ ] tested fastd: x Mbit VPN throughput
- [ ] tested tunneldigger: x Mbit VPN throughput
```

T-X:
Throughput tests sounds like a nice idea! However I think we would need to specify and document how to do them then as we would otherwise get varying results.

Brother-Lal:
Throughput tests will vary with several factor: MTU,bandwidth available, encryption, load on gateways, etc...
I wouldn't add that, as it's really depending on to many factors. A general area (10-13 MBits) would be ok, but not really dependable anyway....
The same with the clients, neighbours and different VPNs - without a real world scenario over a good timeframe (2/3 weeks) no conclusions can be drawn. 

[...]

PR #1475, opened 2018-07-12
ar71xx: add support for OCEDO Koala

blocktrron:
This adds support for the OCEDO Koala, a Wireless-AC Access Point with PoE support, spec wise comparable to an Archer C7 v2.

While this device most likely can't be bought (even on eBay), some Communities have gotten hold of a fairly large quantity of those devices. Therefore it would be great to add this device.

Checklist from here (https://github.com/freifunk-gluon/gluon/issues/1434#issuecomment-399168117) is completely tested, however flashing is possible only from a tftpboot image, but can be done without attaching UART.


blocktrron:
No.1 MTBM - thanks a lot!

PR #1478, opened 2018-07-14
ar71xx: add support for TP-Link RE355

achterin:
Support is marked as broken because this device is sold in two variants,
one with 64M RAM and another with 128M. As of now 64M is not enough
for ath10k with 5GHz enabled.
As there is no indication known which variant one will get better mark
it as broken.

### DIFF
--- a/targets/ar71xx-generic
+++ b/targets/ar71xx-generic
@@ -225,6 +225,10 @@ if [ "$BROKEN" ]; then
 device tp-link-archer-c60-v1 archer-c60-v1 # BROKEN: OOM with 5GHz enabled in most environments
 fi
 
+if [ "$BROKEN" ]; then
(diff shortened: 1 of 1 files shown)


pbriesch:
Does this apply for RE455 as well?

blocktrron:
To be honest, i think adding those devices doesn't solve the problem. Personally i would like to see the broken flag removed and instead the 5GHz radio being disabled by default (ideally with a warning message in advanced config mode).

achterin:
@pbriesch yes, this also applies to the RE450v1. We talked about this on irc and the consensus was, that we dont add the BROKEN flag to already supported devices.

@blocktrron i like your idea. if the gluon-devs are ok with this approach i'll start writing patches.

rotanid:
@achterin have you done testing as specified in our device support policy draft?
quote: "integration tested"

rotanid:
merged.
please open an RFC issue or something like this for the suggested other change

PR #1482, opened 2018-07-15
 gluon-web-admin: add site-commit to info page 

rubo77:
This will enable the communities to add links to their build structure and repositories to the info page.

(Tested on a build on tag 2018.1.x)

## screenshot

auswahl_004

### DIFF
--- a/package/gluon-web-admin/files/lib/gluon/config-mode/view/admin/info.html
+++ b/package/gluon-web-admin/files/lib/gluon/config-mode/view/admin/info.html
@@ -23,7 +23,8 @@
 		{ _('Hostname'), pretty_hostname.get(uci) },
 		{ _('MAC address'), sysconfig.primary_mac },
 		{ _('Hardware model'), platform.get_model() },
-		{ _('Gluon version'), util.trim(util.readfile('/lib/gluon/gluon-version')) },
+		{ _('Gluon version') .. " / " .. _('Site version'), util.trim(util.readfile('/lib/gluon/gluon-version')) 
+			.. " / " .. util.trim(util.readfile('/lib/gluon/site-version')) },
 		{ _('Firmware release'), util.trim(util.readfile('/lib/gluon/release')) },
 		{ _('Site'), site.site_name() },
 		{ _('Public VPN key'), pubkey },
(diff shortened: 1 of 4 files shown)


NeoRaider:
1. It does not make sense to allow changing texts that just describe general facts. In this case, I don't see a point in adding a text at all, please remove the additional section. The text will not help end users in any way (they don't know that these numbers mean), while it is obvious for people who know how to build firmwares.
2. Please order the site version below Gluon version.
3. I'm wondering if we should make this new field configurable in site.conf (allowing to show or hide it), possibly hiding it by default. This depends on whether people find this additonal number primarily useful or primarily annoying; we already have a various numbers on this page...

rubo77:
I added an extra configurable text, so each community can put their link to the firmware-Repository they used there.

rubo77:
we couls also put the site-commit directly after the gluon commit:

> Gluon Version: v2018.1-2-g1234568 / Site: v2016.1.5-322-g34123456+

rubo77:
I updated the screenshot

H4ndl3:
I don't see the point of this PR. The information you intend to display is targeted towards end users, not advanced users but end users will most likely never see the configuration page. 

1. Information regarding to how the firmware was built would be better suited on the page where you download the firmware. 
2. To flash a firmware you have to download it first so a link to the firmware repository would only be helpful for subsequent manual firmware upgrades. Most nodes will be flashed automatically by the autoupdater and those who want to upgrade their node manually already know where to download the firmware.

rubo77:
OK, so if there is no vote for this extra  optional text, I will remove it from this PR.

rubo77:
please vote :+1:  here, If you like to have an optional text (that will only be added, if you add some in your i18n file)

rubo77:
I removed the extra text. now this PR should only add the site-version to the info page

rubo77:
@mweinelt, that would implement that the tracking remote of the site repo and of the used gluon repo would have to be stored on the node also. at the moment we only have the site-repo and gluon-repo commit-IDs (and dirty-flags). See: #937 

I am fine with having to explain how things work on our firmware-download pages. There we can also explain what a dirty flag is and how they can  reconstruct from the IDs in config mode how we built our firmware

Adorfer:
very often when users report problems "from somewhere" and i am trying to help, the maximum i finde (easily without googeling) is the directory of the images... So if by default in the Firmware there would be a link to the repository (decentralized git...), this would help a lot. 

In plus we could open the aspect that a software "distributed" under GPL has to state where the source can be obtained. putting this info inside the images itself would help a lot.

Issue #1488, opened 2018-07-21
APs send out ARP broadcast with anycast IP to mesh (and clients) with local MAC address


ecsv:
It seems like ARP packets are sent around in the mesh which contain different values for the anycast MAC address. These packets are received by the clients - which are then not able to communicate with their connected AP anymore over IPv4. This is especially problematic when dnsmasq is used as DNS forwarder on the node.

----

I wanted to ask about this on IRC but it seems hackint thought that I am a spammer. So here the full text

    [2018-07-21 19:34:41] <ecsv> I've deployed here the dnsmasq forwarder on all nodes (gluon 2018.1) - but it made quite some problems and ruined my friday. but at least I've noticed an interesting property: problem only happens over ipv4
    [2018-07-21 19:34:41] <ecsv> ok, the clients request the DNS records via ipv4 anycast address or via ipv6 anycast address. the problem only happens via ipv4
    [2018-07-21 19:34:41] <ecsv> a similar problem can be seen via icmp echo requests. I have roughly a 12% packet loss when pinging the ipv4 anycast address (300x 1s ping) but 0% loss when using the link local anycast address
    [2018-07-21 19:34:41] <ecsv> arp-limiter was disabled btw.
    [2018-07-21 19:34:41] <ecsv> does anybody else notices the same problem?
    [2018-07-22 00:06:54] <ecsv> ok, problem seems to be that from time to time, the client thinks that the anycast ipv4 address is on a completely different device (on the other end of the mesh)
    [2018-07-22 00:06:54] <ecsv> and why does he seem to think so? because APs on the other end of of the mesh sends out an ARP broadcast (with the anycast ip address) which asks about some client mac which is not known to it
    [2018-07-22 00:06:54] <ecsv> now my client device thinks "hey, cool, why the anycast ipv4 address just changed its mac address. now I have to sent all my traffic to the other one. probing is the cool kids way to handle it"
    [2018-07-22 00:06:54] <ecsv> important question now would be: why do APs send out ARP requests with the anycast IP address but not anycast mac address (16:41:95:40:f7:dc) to get the mac address for some clients (maybe because they wanted to sent some DNS replies)
    [2018-07-22 00:06:54] <ecsv> and why is this traffic allowed to enter the mesh. the last question can be answered using ebtables. (and the first one is a little harder and I will just ignore it for now)
[...]

PR #1489, opened 2018-07-22
2018.1.x: Filter IPv4 anycast specific ARP packets from/to bat0 and always use anycast MAC address for them

ecsv:
The commit b3762fc61cca ("gluon-client-bridge: move IPv4 local subnet route to br-client (#1312)") moves the IPv4 prefix from the local-port interface to br-client. A client requesting an IPv4 connection to the IPv4 anycast address of the node (the device running gluon) will create following packets:

1. ARP packet from client to get the MAC of the mac address of the anycast IPv4 address
2. ARP reply from node to client with the anycast MAC address for the IPv4 anycast address
3. IPv4 packet from client which requires reply (for example ICMP echo request)
4. ARP request for the client MAC address for its IPv4 address in prefix4 (done with the mac address of br-client and transmitted over br-client)
5. IPv4 packet from node (transmitted over br-client with br-client MAC address) as reply for the client IPv4 packet (for example ICMP echo reply)

The step 4 is extremely problematic here. ARP replies with the anycast IPv4 address must not be submitted or received via bat0 - expecially not when it contains an node specific MAC address as source. When it is still done then the wrong MAC address is stored in the batadv DAT cache and ARP packet is maybe even forwarded to clients. This latter is especially true for ARP requests which are broadcast and will be flooded to the complete mesh.

Clients will see these ARP packets and change their own neighbor IP (translation) table. They will then try to submit the packets for IPv4 anycast addresses to the complete wrong device in the mesh. This will for example break the access to the status page to the connected device or the anycast DNS forwarder implementation. Especially the latter causes extreme latency when clients try to connect to server using a domain name or even breaks the connection setup process completely. Both are caused by the unanswered DNS requests which at first glance look like packet loss.

An node must therefore take care of:

* not transmitting ARP packets related to the anycast IPv4 address over bat0
* drop ARP packets related to the anycast IPv4 when they are received on bat0 from a still broken node
* don't accept ARP packets related to the anycast IPv4 replies on local node when it comes from bat0

The step 4 and 5 are problematic here because packets use the node specific MAC addresses from br-client instead of the anycast MAC address. The client will receive the ARP packet with the node specific MAC address and change their own neighbor IP (translation) table. This will for example break the access to the status page to the connected device or the anycast DNS forwarder implementation when the client roams to a different node.

----

This is for the v2018.1x branch. This branch doesn't seem to have support for babel and a revert of b3762fc61cca ("gluon-client-bridge: move IPv4 local subnet route to br-client (#1312)") should not be rejected with "but this will not work with babeld".

See #1488 for the related ticket.

### DIFF
--- a/package/gluon-client-bridge/luasrc/lib/gluon/upgrade/310-gluon-client-bridge-local-node
+++ b/package/gluon-client-bridge/luasrc/lib/gluon/upgrade/310-gluon-client-bridge-local-node
@@ -23,7 +23,8 @@ uci:section('network', 'device', 'local_node_dev', {
 local ip4, ip6
 
 if next_node.ip4 then
-	ip4 = next_node.ip4 .. '/32'
+	local plen = site.prefix4():match('/%d+$')
+	ip4 = next_node.ip4 .. plen
 end
 
 if next_node.ip6 then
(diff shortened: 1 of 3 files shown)


T-X:
I understood @ecsv that he offered this PR, #1489, for v2018.1. And #1490 for master.

Now that @neoraider seems to have decided to merge #1490 both into master and v2018.1.x, why was there still a need to merge this PR (#1489)?

Am I missing something?

NeoRaider:
Huh? I merged #1489 into v2018.1 and #1490 into master. Note that both PRs are completely identical anyways.

PR #1492, opened 2018-07-24
gluon-config-mode-theme: background grey of clicked buttons

rubo77:
This is a small easy enhancement of the config mode: If you click a submit button it will change its colour to grey, so you have a feedback that you really clicked the button.


rotanid:
i think it's a good idea.

rubo77:
I just tested a build with this and it works as expected

PR #1503, opened 2018-08-06
docs: getting_started: dependency for time package

rubo77:


rotanid:
1) pls, your commit message ;) capital letter (and i'm not sure about those marks around the time
2) just FYI, the dependency will be gone again with OpenWrt 19.01 ^^

rubo77:
changed the commit message.

so then it should be there until we update to 19.01, shouldn't it?

rotanid:
yes, that information was really just that: information.

PR #1508, opened 2018-08-08
ramips-mt7620: add support for TP-Link Archer C20 v1

ambassador86:
I used the TFTP method as advised by openwrt wiki.

- [x]  must support upgrade mechanism
  - [x]  must have working sysupgrade
  - [ ]  must have working autoupdate
maybe there is an upstream problem as the image (gluon and openwrt) fail the image check because of an additional dash in the name
`Device tplink,c20-v1 not supported by this image
Supported devices: c20v1
`

- [x] wired network
- [ ] wifi (if applicable)
only on 2,4 GHZ as 5GHz is not supported at the moment
only on 2,4 GHZ as 5GHz is not supported at the moment
  - [ ] ap/mesh mode must work in parallel on all usable radios

- [x] led mapping
  - [x] power/sys led

- [ ] radio leds
they are both turned off
  - [ ] should show activity
 - [x] switchport leds

issues resolved

### DIFF
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac
@@ -47,6 +47,8 @@ elseif platform.match('ar71xx', 'generic', {'archer-c5', 'archer-c58-v1',
 elseif platform.match('ipq40xx', nil, {'avm,fritzbox-4040',
                                        'openmesh,a42', 'openmesh,a62'}) then
   table.insert(try_files, 1, '/sys/class/net/eth0/address')
+elseif platform.match('ramips', 'mt7620', {'tplink,c20-v1'}) then
+  table.insert(try_files, 1, '/sys/class/net/eth0/address')
 end
 
 
--- a/targets/ramips-mt7620
+++ b/targets/ramips-mt7620
@@ -7,3 +7,9 @@ factory
 
 device gl-mt750 gl-mt750
 factory
+
+# TP-Link
+if [ "$BROKEN" ]; then
+device tp-link-archer-c20-v1 tplink_c20-v1 # BROKEN: only 2.4GHz wifi working
+fi
+
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -340,6 +340,10 @@ ramips-mt7620
   - GL-MT300N [#80211s]_
   - GL-MT750 [#80211s]_
 
+* TP-Link
+
+  - Archer C20 (v1) [#80211s]_
+
 ramips-mt7621
 ^^^^^^^^^^^^^
 


targets/ramips-mt7620
```diff
@@ -7,3 +7,10 @@ factory
 
 device gl-mt750 gl-mt750
 factory
+
+# TP-Link
+# BROKEN: only 2.4GHz wifi working
```
rotanid: please add this mark at the end of the "device" line as it's done for the other devices

blocktrron:
From the dts, the correct MAC address should reside on eth0. Add the device here to set the correct Primary-MAC: https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac

Radio LEDs probably will never work as the drive does not implement needed triggers.

ambassador86:
Hey @blocktrron the MAC adress for eth0 is the same as printed on the case.

blocktrron:
@ambassador86 So you will need to add something along the lines of 
```
elseif platform.match('ramips', 'mt7620', {'tplink,c20-v1'}) then
  table.insert(try_files, 1, '/sys/class/net/eth/address')
```

in 010-primary-mac

ambassador86:
With vanilla Openwrt I can choose which LED to use for wireless. In my opinion the naming of both is not correct as the 5GHz symbol lights up in the standard config. I whould open a bug report for this in Openwrt.

package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac
```diff
@@ -47,9 +47,10 @@ elseif platform.match('ar71xx', 'generic', {'archer-c5', 'archer-c58-v1',
 elseif platform.match('ipq40xx', nil, {'avm,fritzbox-4040',
                                        'openmesh,a42', 'openmesh,a62'}) then
   table.insert(try_files, 1, '/sys/class/net/eth0/address')
+elseif platform.match('ramips', 'mt7620', {'tplink,c20-v1'}) then
+  table.insert(try_files, 1, '/sys/class/net/eth0/address')
 end
 
-
```
blocktrron: Deletion of this line seems to be done by mistake?

targets/ramips-mt7620
```diff
@@ -7,3 +7,9 @@ factory
 
 device gl-mt750 gl-mt750
 factory
+
+# TP-Link
+if [ "$BROKEN" ]; then
+device tp-link-archer-c20-v1 tplink_c20-v1 # BROKEN: only 2.4GHz wifi working
```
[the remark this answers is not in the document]
rotanid: i don't think it makes sense to even merge it as broken with these downsides, especially broken upgrade, therefore the comment isn't necessary, but a fix is

ambassador86:
I wrote a PR for OpenWRT to add the support for sysupgrade in https://github.com/openwrt/openwrt/pull/1406

rotanid:
your fix made it into the 18.06 branch: https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=e4678c6c01fac468f677633de1bc5116fafefc38

rotanid:
>  With vanilla Openwrt I can choose which LED to use for wireless. In my opinion the naming of both is not correct as the 5GHz symbol lights up in the standard config. I whould open a bug report for this in Openwrt.

have you made progress on this issue with the LEDs?

ambassador86:
Hello @rotanid, the sysupgrade is now working. I had no luck with the LEDs so far.

ambassador86:
With a current OpenWRT build under gluon the radio is not supporting mesh and ap mode in parallel anymore.

J0WI:
Sysupgrade may work now: https://github.com/openwrt/openwrt/pull/1406

rotanid:
@J0WI thanks, but i already linked this here few days ago ;)
there's still the issue with the LEDs.

PR #1514, opened 2018-08-19
Update ar71xx-generic

heini66:
Add other image to first flash avm's 4020 device

### DIFF
--- a/targets/ar71xx-generic
+++ b/targets/ar71xx-generic
@@ -36,6 +36,7 @@ factory
 # AVM
 device avm-fritz-box-4020 fritz4020 fritz4020
 factory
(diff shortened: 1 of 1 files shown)


blocktrron:
The 4020 doesn't need a separate Image for bootstrapping via EVA. Such an image is also not generated while building OpenWRT for the device. So this patch would break copying the files.
A separate Image is needed for the 4040 as it contains a second bootloader for the device. The 4020 doesn't need such a separate image.

While we could make a second copy of the sysupgrade image to store it in the "other" subdirectory, i think this is only wasting space as instructions are pointing to the sysupgrade image anyway.

PR #1515, opened 2018-08-19
gluon-config-mode-geo-location-osm: OpenStreetMap-based location picker

NeoRaider:
This PR contains the revised version of the OpenStreetMap-based location picker for the config mode (#1484).

Changes:
- Better integration with gluon-web
- OpenLayers 5 (instead of 2)
- Lots of changed strings (both package names and user-visible texts)
- Extend gluon-config-mode-geo-location rather than replacing it (avoiding to build multiple variants of the same code)

The one big remaining issue is that OpenLayers 5.1.3 does not work with XHTML (or HTML5 delivered with application/xhtml+xml Content-Type). This has been fixed in https://github.com/openlayers/openlayers/pull/8524, so I hope the next OpenLayers release will just work.


docs/user/site.rst
```diff
@@ -407,6 +407,13 @@ config_mode \: optional
     package. Set *geo_location.show_altitude* to *true* if you want the altitude
     field to be visible.
 
+    The *geo_location.osm* section is only relevant when the *gluoin-config-mode-geo-location-osm*
```
2tata: gluoin-config-mode-geo-location-osm -> gluon-config-mode-geo-location-osm

package/gluon-config-mode-geo-location-osm/check_site.lua
```diff
@@ -0,0 +1,4 @@
+need_number(in_site({'config_mode', 'geo_location', 'osm', 'center', 'lon'}))
+need_number(in_site({'config_mode', 'geo_location', 'osm', 'center', 'lat'}))
```
2tata: Maybe we can have default values like 0.0/0.0 to reduce site.conf parameters.

NeoRaider: Certainly not. This is a value that *must* be provided by each site, 0.0/0.0 simply doesn't make sense.

package/gluon-config-mode-theme/sass/gluon.scss
```diff
@@ -284,6 +284,12 @@ code {
   margin-bottom: 2em;
 }
 
+.gluon-osm-map {
+  width: 100%;
+  height: 40em;
```
2tata: 40em seems to much maybe 25 is better the map looks nicer.

NeoRaider: The width of the main content scales between 40em and 60em (depending on the window width), meaning that with a height of 40em the map is square in narrow windows, and has a 3:2 aspect ratio in wide windows. This is already close to the usual widescreen display aspect ratio of 16:10.

Given that the relevant areas for sites should be circular on average, a square seems like the best approximation to me. Further decreasing the height would make the map less square, so we should either increase the height, or keep the current value...

NeoRaider:
Screenshot:

image

rotanid:
i think it's a nice solution.
btw., the fix to openlayers was merged.

mweinelt:
This would be an ideal basis for a picking a domain, e.g. through boundary box overlays, like at https://darmstadt.freifunk.net/mitmachen/domains/.

2tata:
@NeoRaider There seems a bug in thins code. If I have "Advertise node position" checked and had already select a position on the map. After Pressing "Save & restart" the configmode will be just reload but not completed. Only after I have unchecked "Advertise node position" I can apply the formula.

NeoRaider:
Hmm, I will look into this.

NeoRaider:
I fixed two issues in the code:
* missing `validate()` override
* map was not using the previously submitted lon/lat values when form was rejected because of unrelated validation failures

2tata:
@NeoRaider I can confirm the bugfix. Now everything seems working as expected.

docs/user/site.rst
```diff
@@ -407,6 +407,13 @@ config_mode \: optional
     package. Set *geo_location.show_altitude* to *true* if you want the altitude
     field to be visible.
 
+    The *geo_location.osm* section is only relevant when the *gluon-config-mode-geo-location-osm*
+    package is used. The *center.lon* and *center.lat* values are mandatory in this case and
+    define the default center of the map when no position has been picked yet. The *zoom* level
+    defaults to 12 in this case. *openlayers_url* allows to override the base URL of the
+    *build/ol.js* and *css/ol.css* files (the default is
+    ``https://cdn.rawgit.com/openlayers/openlayers.github.io/master/en/v5.1.3``).
```
2tata: url can be update to https://cdn.rawgit.com/openlayers/openlayers.github.io/master/en/v5.2.0/build/ol.js
that contains your upstream fixes.

rubo77:
What is a valis URL? still the same as in the original PR?

        openlayers_url = 'http://dev.openlayers.org/OpenLayers.js',

2tata:
No its not. But you don't need to set a URL because a default value is already given. Furthermore there exists some documentation in docs/user/site.rst.

PR #1516, opened 2018-08-22
package/features: add gluon-ebtables-limit-arp on batman packages as Opt out

2tata:
This MR is an alternative idea for #1418

The gluon-ebtables-limit-arp is selected as default when the batman package is selected.
For communities who don't want this package, the can deselect it as described below:
```
GLUON_SITE_PACKAGES := \
        -gluon-ebtables-limit-arp
```

### DIFF
--- a/package/features
+++ b/package/features
@@ -23,4 +23,5 @@ packages 'web-advanced & autoupdater' \
 	'gluon-web-autoupdater'
 
 packages 'status-page & (mesh-batman-adv-14 | mesh-batman-adv-15)' \
-        'gluon-status-page-mesh-batman-adv'
+        'gluon-status-page-mesh-batman-adv' \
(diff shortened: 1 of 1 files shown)


rotanid:
i added a space before the backslash and updated the commit message.

package/features
```diff
@@ -23,4 +23,5 @@ packages 'web-advanced & autoupdater' \
 	'gluon-web-autoupdater'
 
 packages 'status-page & (mesh-batman-adv-14 | mesh-batman-adv-15)' \
-        'gluon-status-page-mesh-batman-adv'
+        'gluon-status-page-mesh-batman-adv' \
+        'gluon-ebtables-limit-arp'
```
NeoRaider: Why would this depend on the status-page feature?

PR #1518, opened 2018-08-24
gluon-mesh-vpn-core: add abstraction layer for vpn interface return

2tata:
This PR should intergrade a missing abstraction that give developers the option to get the mesh VPN interface name, with hasn't to be aware of every vpn-tunneling-solution.


NeoRaider:
You can simplify this whole code to `return 'mesh-vpn'`. The name is hardcoded in *gluon-mesh-vpn-core*.

Also, please rename the Lua file to `mesh-vpn.lua`, so it matches the package name (with `-core` removed).

PR #1528, opened 2018-09-06
docs: enhance  example values for gluon-config-mode-geo-location-osm

rubo77:
Latitude and longitude was swapped and the example Layer URL is not reacheable so it should be set in comments.

### DIFF
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -428,11 +428,11 @@ config_mode \: optional
             show_altitude = true,
             osm = {
               center = {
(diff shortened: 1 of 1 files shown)


rubo77:
ffac is the official TLD of freifunk alpha centauri ;)

rotanid:
@rubo77 maybe ffac.example.org or at least some domain containing a dot?
(and your commit message, as was already said)

rubo77:
changed that.

PR #1529, opened 2018-09-09
gluon-config-mode-hostname: add option to use default hostname

blocktrron:
Before 7827f89, mandatory hostname field in config mode was pre-filled with the default hostname.

This commit adds the config_mode.hostname.prefill option for controlling the default value.

### DIFF
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -403,6 +403,9 @@ config_mode \: optional
     and the node's primary MAC address is assigned. Manually setting a hostname
     can be enforced by setting *hostname.optional* to *false*.
 
+    To not prefill the hostname-field in config-mode with the default hostname,
+    set *hostname.prefill* to *false*.
+
     By default, no altitude field is shown by the *gluon-config-mode-geo-location*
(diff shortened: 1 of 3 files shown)


rotanid:
what happens, if someone leaves the prefilled default in place?
in my opinion, if a custom hostname is mandatory, it shouldn't accept the default value - otherwise the "optional = false" is pretty useless together with prefill = true

blocktrron:
@rotanid If the user leaves the prefilled default, the node has a unique hostname (for us '64xxx-<NODEID>').

The motivation behind this is due to an increased number of keys which do not follow our name-scheme. This was not the case when the hostname was prefilled and not only used as a placeholder.

'optional=false' and 'prefill=true' was exactly the behavior we had before the mentioned commit. I'll admit that it might not be sensible (although it was the behavior for years) for every community but at least for us it is the best solution.

rotanid:
how can "64xxx-" be unique?
but my point wasn't about uniqueness, but about having to change the default (this allows to use the default as part of the changed name, like "examplestreet-001122334455")
but i think i got your intention and it looks valid, just doesn't cover my idea

PR #1533, opened 2018-09-17
mt76x8: flag target as broken

blocktrron:
The MT76x8 is currently not stable enough for worry-free use with Gluon.
It suffers from reboots in intervals as little as 12 hours or WiFi
freezes for several hours.

It may be unflagged as soon the situation with mt76 got better.


rubo77:
What is this had upon?

We have a C50, that worked fine for days

rotanid:
@rubo77 based upon experience from my and @blocktrron s test nodes.
also, the mt76 driver maintainer knows about issues when Apple/iOS devices are connecting - which might be the reason for the issues we are observing.
it will be investigated and probably be fixed, but until then: BROKEN.

PR #1535, opened 2018-09-23
ath10k-ct: reduce memory consumption

ecsv:
ath10k-ct uses a rather high number of buffers to communicate with the QCA/Candelatech firmware. Especially the HTC (host-target-communication) and HTT (host-target-transport) can take up a lot of memory when data is transferred over a ath10k wifi link.

Even a 256MB device with three radios can go OOM (while sending to three devices using UDP unicast/multicast packets) with the default buffer limits. Unfortunately, this also reduced the maximum 5GHz throughput on an IPQ40xx device from 494/432 Mbit/s for TCP transfers (download/upload) to
438/343 Mbit/s. These throughput values should still be good enough for a Freifunk installation. And these changes are already used by ath10k in OpenWrt 18.06.

----

There was already a pull request for LEDE 17.01 (gluon 2018.1.x)  in #1440. This one was not accepted before the release of 2018.1 and later rejected for 2018.1.x. See also the failed PR for OpenWrt 18.06 https://github.com/openwrt/openwrt/pull/1077 . The original ath10k (no -ct) driver was patched in https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=cc189c0b7fa015978b04bb663a75b1da726376b5

This change will affect QCA 802.11 11ac based hardware:

* every target  when using GLUON_WLAN_MESH = 'ibss'
* target ipq806x
* target x86

### DIFF
--- a/patches/openwrt/0008-ath10k-ct-reduce-memory-consumption.patch
+++ b/patches/openwrt/0008-ath10k-ct-reduce-memory-consumption.patch
@@ -1,0 +1,78 @@
+From: Sven Eckelmann <sven@narfation.org>
+Date: Mon, 18 Jun 2018 21:03:37 +0200
+Subject: ath10k-ct: reduce memory consumption
+
+ath10k uses a rather high number of buffers to communicate with the QCA
+firmware. Especially the HTC (host-target-communication) and HTT
+(host-target-transport) can take up a lot of memory when data is
+transferred over a ath10k wifi link.
+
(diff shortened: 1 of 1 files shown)


mweinelt:
Assigning to @NeoRaider because he wanted to discuss the ath10k memory usage and upstream defaulting to ath10k-ct with other openwrt members.

NeoRaider:
I've rebased and merged this for now, until we've worked out a way to deal with this upstream.

PR #1537, opened 2018-09-24
Bugfix: package/gluon-config-mode-geo-location: delet nodes Position after disable location option by reenterd config-mode. Also set location independend from sharing

2tata:
Fix an issue after reentering the config-mode. In case a user has enterd the config-mode again and is disabling the location option the previously stored coordinates will be not deleted. Furthermore, this MR makes able to set a position independent from the option to share the nodes location.

### DIFF
--- a/package/gluon-config-mode-geo-location/luasrc/lib/gluon/config-mode/wizard/0400-geo-location.lua
+++ b/package/gluon-config-mode-geo-location/luasrc/lib/gluon/config-mode/wizard/0400-geo-location.lua
@@ -32,26 +32,41 @@ return function(form, uci)
 
 	local o
 
-	local share_location = s:option(Flag, "location", pkg_i18n.translate("Advertise node position"))
-	share_location.default = uci:get_bool("gluon-node-info", location, "share_location")
-	function share_location:write(data)
-		uci:set("gluon-node-info", location, "share_location", data)
+	local own_latitude = uci:get("gluon-node-info", location, "latitude")
+	local own_longitude = uci:get("gluon-node-info", location, "longitude")
 
-		-- The config mode does not have a nicer place to put this at the moment...
-		if not show_altitude then
+	local set_location = s:option(Flag, "location", pkg_i18n.translate("Set node position"))
+	set_location.default = own_latitude or own_longitude
+
+	-- Delete already saved coordinates
+	function set_location:write(data)
+		if not data then
+			uci:delete("gluon-node-info", location, "latitude")
+			uci:delete("gluon-node-info", location, "longitude")
+			uci:set("gluon-node-info", location, "share_location", false)
+		end
+
+		if not show_altitude or not data then
 			uci:delete("gluon-node-info", location, "altitude")
 		end
+	end
+
+	local share_location = s:option(Flag, "share_location", pkg_i18n.translate("Advertise node position"))
+	share_location.default = uci:get_bool("gluon-node-info", location, "share_location")
+	share_location:depends(set_location, true)
+	function share_location:write(data)
+		uci:set("gluon-node-info", location, "share_location", data)
 	end
 
 	local map
 	if osm then
 		map = s:option(osm.MapValue, "map", osm.options())
-		map:depends(share_location, true)
+		map:depends(set_location, true)
 	end
 
 	o = s:option(Value, "latitude", pkg_i18n.translate("Latitude"), pkg_i18n.translatef("e.g. %s", "53.873621"))
-	o.default = uci:get("gluon-node-info", location, "latitude")
-	o:depends(share_location, true)
+	o.default = own_latitude
+	o:depends(set_location, true)
 	o.datatype = "float"
 	function o:write(data)
 		uci:set("gluon-node-info", location, "latitude", data)
@@ -61,8 +76,8 @@ return function(form, uci)
 	end
 
 	o = s:option(Value, "longitude", pkg_i18n.translate("Longitude"), pkg_i18n.translatef("e.g. %s", "10.689901"))
-	o.default = uci:get("gluon-node-info", location, "longitude")
-	o:depends(share_location, true)
+	o.default = own_longitude
+	o:depends(set_location, true)
 	o.datatype = "float"
 	function o:write(data)
 		uci:set("gluon-node-info", location, "longitude", data)
@@ -77,7 +92,7 @@ return function(form, uci)
 			pkg_i18n.translatef("e.g. %s", "11.51")
 		)
 		o.default = uci:get("gluon-node-info", location, "altitude")
-		o:depends(share_location, true)
+		o:depends(set_location, true)
 		o.datatype = "float"
 		o.optional = true
 		function o:write(data)
--- a/package/gluon-config-mode-geo-location/i18n/de.po
+++ b/package/gluon-config-mode-geo-location/i18n/de.po
@@ -26,6 +26,9 @@ msgstr "Breitengrad"
 msgid "Longitude"
 msgstr "Längengrad"
 
+msgid "Set node position"
+msgstr "Knotenposition setzen"
+
 msgid "Advertise node position"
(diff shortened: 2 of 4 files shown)


rubo77:
Why do you need that?

I assume it is needed for a new package, of so, then this should only changed if that new package is really installed also

kpanic23:
This would really be a nice to have. With packages existing to automagically selecting the right domain for given coordinates, it would be great to be able to set (and unset) the node's coordinates without having to publish them via respondd. This way nodes can auto-select the domain without appearing on the map.
At the moment, you have to publish the coordinates to be able to set them in config mode.

2tata:
I need to change something more, because at the moment its the coordinates fields a are shown permanently and the fils are mark as required. Means without coordinates it will be not possible to complete the config mode.

2tata:
This change is related to what @kpanic23 told. It will give node owners the option to provide coordinates e.g. for the hoodselector or any community related domain selection, without showing the node on the map.

rubo77:
Then the option should be made independent only if the hoodselector or any community related domain selection is present.

It would be unnecessary confusing for the use in a config mode of a domain without such packages.
Some inexperienced Users would disable this without understanding that then it is of no use for their community. And that's probably not what they want.

2tata:
> Then the option should be made independent only if the hoodselector or any community related domain selection is present.

This is already given in some communities.

> It would be unnecessary confusing for the use in a config mode of a domain without such packages.
> Some inexperienced Users would disable this without understanding that then it is of no use for their community. And that's probably not what they want.

This option makes the gluon framework a bit more flexible wether or not if 
packages like the hoodselector or any other position related services present in 
gluon.

cheers
Tarek

christf:
@rotanid @NeoRaider @mweinelt how do you feel about this one? this is a 3-line patch. We should be able to for a decision on whether to merge or decline within a month.

rotanid:
merged it with adjusted commit message: 104d7455fa721fb0198a38a5a05fd4e8d9b72a72

NeoRaider:
I'm not happy with this at all. The commit doesn't only show the position fields even when the position is not shared - it forces the user always to provide a position!

mweinelt:
Reverted and reopened so changes can be made.

CodeFetch:
@mweinelt Please close this again as it does not make any sense. What I see here is a PR without content and a misleading name. If 2tata wants to geolocate the node dynamically this should not change anything on the existing config-mode-geo-location package.

rotanid:
@2tata do you plan to rebase and fix this?

CodeFetch:
@rotanid If the user decides to not share a location, why should you share it against his or her will? That is what the PR does and thus NeoRaider closed it. I don't understand why mweinelt reopened it.

rubo77:
aus einem chat:

> tata 07:20
> ja, will mich die Tage drum kümmern

rotanid:
@CodeFetch this was an error, not the intention of the PR. therefore, the change was reverted until a proper version without this problem exists.

rotanid:
@2tata do you plan to implement the necessary changes?

2tata:
> @2tata do you plan to implement the necessary changes?

I think I will be not available for that in the next 2 month ... after that I plan to finish it. Shall I close it?

rotanid:
@2tata no need to close it if you still plan to do it, it just wont make it in 2019.1 then.

2tata:
> @2tata no need to close it if you still plan to do it, it just wont make it in 2019.1 then.

@rotanid  could you also add the patch welcome tag please? If some other would like to finish it in the next 2 month, I don't mind.

rotanid:
@2tata maybe you have time for this again, as 3 months went by?

2tata:
@rotanid I will finish it tomorrow noon.

2tata:
@rotanid rebased and tested on a tl841n

package/gluon-config-mode-geo-location/luasrc/lib/gluon/config-mode/wizard/0400-geo-location.lua
```diff
@@ -32,26 +32,46 @@ return function(form, uci)
...
+	local own_longitude = uci:get("gluon-node-info", location, "longitude")
 
-		-- The config mode does not have a nicer place to put this at the moment...
-		if not show_altitude then
+	local set_location = s:option(Flag, "location", pkg_i18n.translate("Set node position"))
+	if own_latitude or own_longitude then
+		set_location.default = true
+	else
+		set_location.default = false
+	end
```
mweinelt:
```suggestion
	set_location.default = own_latitude or own_longitude
```

package/gluon-config-mode-geo-location/luasrc/lib/gluon/config-mode/wizard/0400-geo-location.lua
```diff
@@ -32,26 +32,46 @@ return function(form, uci)
...
-	share_location.default = uci:get_bool("gluon-node-info", location, "share_location")
-	function share_location:write(data)
-		uci:set("gluon-node-info", location, "share_location", data)
+	local own_latitude = uci:get("gluon-node-info", location, "latitude")
+	local own_longitude = uci:get("gluon-node-info", location, "longitude")
 
-		-- The config mode does not have a nicer place to put this at the moment...
-		if not show_altitude then
+	local set_location = s:option(Flag, "location", pkg_i18n.translate("Set node position"))
+	if own_latitude or own_longitude then
```
mweinelt:
```suggestion
	set_location.default = own_latitude or own_longitude
```

package/gluon-config-mode-geo-location/luasrc/lib/gluon/config-mode/wizard/0400-geo-location.lua
mweinelt:
```suggestion:-5
	set_location.default = own_latitude or own_longitude
```

package/gluon-config-mode-geo-location/luasrc/lib/gluon/config-mode/wizard/0400-geo-location.lua
```diff
@@ -32,26 +32,46 @@ return function(form, uci)
...
+		set_location.default = true
+	else
+		set_location.default = false
+	end
+
+	-- Delete already saved coordinates
+	function set_location:write(data)
+
```
mweinelt: remove empty line

2tata: applied

package/gluon-config-mode-geo-location/luasrc/lib/gluon/config-mode/wizard/0400-geo-location.lua
2tata: applied

rotanid:
i think the code is OK, but we need structured testing because of the problem we had last time:
1) don't share or even set anything
2) don't share but set location
2.1) switch to 1)
3) share and set location
3.1) switch to 1)
3.2) switch to 2)

2tata:
@rotanid I have tested all of your six mentenorid cases three days before

rotanid:
4: what happens during upgrades from previous Gluon versions?
4.1) no location was saved and sharing disabled
4.2) location data is there, but sharing was disabled
4.3) location data is there, sharing enabled

i expect:
4.1) no change
4.2) location data stays, location set is set to true, sharing stays disabled
4.3) location data stays, location set is set to true, sharing stays enabled

rotanid:
i also tested the scenarios now.
last thing is: the commit message.
but i'm not quite sure about a commit message, maybe something like this?

```
gluon-config-mode-geo-location: allow to set location independently from     sharing it

also properly delete location data when
 "set location" is not set
```

2tata:
@rotanid how is it now ?

rotanid:
@2tata adjusted it slightly and merged.

PR #1539, opened 2018-09-27
v2018.1.x: Add gluon support for GL.iNet GL-MiFi

cmarxmeier:
bug in lede patch for gl-mifi
patches/lede/0086-ar71xx-set-status-led-for-the-gl-boards.patch
gl-mifi:greeen:wan -> gl-mifi:green:wan

### DIFF
--- a/patches/lede/0086-ar71xx-set-status-led-for-the-gl-boards.patch
+++ b/patches/lede/0086-ar71xx-set-status-led-for-the-gl-boards.patch
@@ -318,7 +318,7 @@ index 42f4415d7fe0252aadf39e2ca50f96566c023728..412c562fa042e7abb0ccb35208bb5582
  static struct gpio_led gl_mifi_leds_gpio[] __initdata = {
  	{
 -		.name = "gl-mifi:wan",
-+		.name = "gl-mifi:greeen:wan",
++		.name = "gl-mifi:green:wan",
  		.gpio = GL_MIFI_GPIO_LED_WAN,
  		.active_low = 0,
  	},
--- a/targets/ar71xx-generic
+++ b/targets/ar71xx-generic
@@ -7,7 +7,7 @@ if [ "$GLUON_WLAN_MESH" = 'ibss' ]; then
 	ATH10K_PACKAGES_QCA9887='-kmod-ath10k kmod-ath10k-ct -ath10k-firmware-qca9887 ath10k-firmware-qca9887-ct'
 fi
 
-MIFI_PACKAGES='rt2800-usb-firmware kmod-mppe kmod-usb-net kmod-usb-net-rndis kmod-usb-net-qmi-wwan kmod-usb-acm kmod-usb-core kmod-ath kmod-ath9k kmod-ath9k-common kmod-cfg80211 kmod-crypto-aead kmod-crypto-crc32c kmod-crypto-ecb kmod-crypto-hash kmod-crypto-manager kmod-crypto-null kmod-crypto-pcompress kmod-crypto-sha1 kmod-fs-exfat kmod-fs-ext4 kmod-fs-ntfs kmod-fs-vfat kmod-fuse kmod-gpio-button-hotplug kmod-lib-crc-ccitt kmod-lib-crc-itu-t kmod-lib-crc16 kmod-lib-textsearch kmod-mac80211 kmod-mii kmod-ppp kmod-pppoe kmod-pppox kmod-rt2800-lib kmod-rt2800-usb kmod-rt2x00-lib kmod-rt2x00-usb kmod-scsi-core kmod-slhc kmod-tun kmod-udptunnel4 kmod-udptunnel6 kmod-usb-net-huawei-cdc-ncm kmod-usb-net-ipheth kmod-usb-net-sierrawireless kmod-usb-ohci kmod-usb-serial kmod-usb-serial-cp210x kmod-usb-serial-option kmod-usb-serial-sierrawireless kmod-usb-serial-wwan kmod-usb-serial-wwan kmod-usb-storage kmod-usb-uhci kmod-usb-wdm kmod-usb2'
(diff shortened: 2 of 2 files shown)


patches/lede/0086-ar71xx-set-status-led-for-the-gl-boards.patch
```diff
@@ -318,7 +318,7 @@ index 42f4415d7fe0252aadf39e2ca50f96566c023728..412c562fa042e7abb0ccb35208bb5582
  static struct gpio_led gl_mifi_leds_gpio[] __initdata = {
  	{
 -		.name = "gl-mifi:wan",
-+		.name = "gl-mifi:greeen:wan",
++		.name = "gl-mifi:green:wan",
```
mweinelt: Are you modifying a patch that was pulled from OpenWrt? If so, why didn't you submit it upstream instead?

cmarxmeier:
its a gluon lede patch with mistyping in original

patches/lede/0086-ar71xx-set-status-led-for-the-gl-boards.patch
blocktrron: It's in fact updated in OpenWRT (also in the 18.06 release).

https://github.com/openwrt/openwrt/commit/58fe5279318b0bcc398ea5bde68db4ce204daa8b

So we should backport this patch instead of fiddling in existing patches.

cmarxmeier: the pull request for the mistype is just a correction of an active lede patch in order to get wlan led working. To keep it stupid simple is would suggest to update 2018.1.x branch lede patch with correction like i did in my source that's currently compiling for 8 Domains in Rhein-Sieg

cmarxmeier: the pull request regarding target/ar71-generic is valid for 2018.1 and master. lede and openwrt both support gl-mifi .
gluon-2018.1.x based firmware is in productive test for more than 24 hours now:
https://map.freifunk-siegburg.de/#!v:m;n:e4956e44b048

NeoRaider:
* All new hardware support must land in the Gluon master first before it is backported to stable branches
* Backported patches should be added using `git cherry-pick` where possible, rather than editing existing patches (you didn't even refresh the patch files, the file hashes in the patch headers should change when the content changes)

mweinelt:
Check https://gluon.readthedocs.io/en/v2018.1.x/dev/basics.html for a guide on how to update patches.

targets/ar71xx-generic
```diff
@@ -7,6 +7,8 @@ if [ "$GLUON_WLAN_MESH" = 'ibss' ]; then
 	ATH10K_PACKAGES_QCA9887='-kmod-ath10k kmod-ath10k-ct -ath10k-firmware-qca9887 ath10k-firmware-qca9887-ct'
 fi
 
+MIFI_PACKAGES='rt2800-usb-firmware kmod-mppe kmod-usb-net kmod-usb-net-rndis kmod-usb-net-qmi-wwan kmod-usb-acm kmod-usb-core kmod-ath kmod-ath9k kmod-ath9k-common kmod-cfg80211 kmod-crypto-aead kmod-crypto-crc32c kmod-crypto-ecb kmod-crypto-hash kmod-crypto-manager kmod-crypto-null kmod-crypto-pcompress kmod-crypto-sha1 kmod-fs-exfat kmod-fs-ext4 kmod-fs-ntfs kmod-fs-vfat kmod-fuse kmod-gpio-button-hotplug kmod-lib-crc-ccitt kmod-lib-crc-itu-t kmod-lib-crc16 kmod-lib-textsearch kmod-mac80211 kmod-mii kmod-ppp kmod-pppoe kmod-pppox kmod-rt2800-lib kmod-rt2800-usb kmod-rt2x00-lib kmod-rt2x00-usb kmod-scsi-core kmod-slhc kmod-tun kmod-udptunnel4 kmod-udptunnel6 kmod-usb-net-huawei-cdc-ncm kmod-usb-net-ipheth kmod-usb-net-sierrawireless kmod-usb-ohci kmod-usb-serial kmod-usb-serial-cp210x kmod-usb-serial-option kmod-usb-serial-sierrawireless kmod-usb-serial-wwan kmod-usb-serial-wwan kmod-usb-storage kmod-usb-uhci kmod-usb-wdm kmod-usb2'
```
blocktrron: There seem to be a lot of packages which are not necessary here (e.g. kmod-ath, kmod-ath9k, kmod-fs-exfat kmod-fs-ext4, kmod-fs-ntfs, kmod-fs-vfat, kmod-fuse, kmod-gpio-button-hotplug, ...)

cmarxmeier: jepp - list was taken from manufacturer's  distro openwrt 18.06 image.

cmarxmeier: 20180926_ff_unterwechs_gl-mifi
but filling less than 5% of nvram storage with everything included.

mweinelt: I'd be fine with firmware and userspace utilities to get the LTE radio going, since that seems to be its primary purpose, but most filesystems, USB device support and several kmods are not required for the device to work.

mweinelt:
Please update the PR title to say something like

"ar71xx-generic: Add support for GL.iNet GL-MiFi"

NeoRaider:
There is still no PR for the Gluon master. This device *must* be added to to the master, before we can even consider adding a backport to v2018.1.x.

blocktrron:
@cmarxmeier Judging from the datasheet, the device supports qmi protocol, so instead of currently 56 packages, the device should work with only `kmod-usb-net-qmi-wwan` and `uqmi` package. It does also support serial communication, but i would not want to use that for a high-speed connection such as 4G.

GL.iNet is also referencing some HiSilicon (Huawei) based card, but it does not seem to be purchaseable from their shop.

cmarxmeier:
modem is embedded by mini pci-e module 3,3v so with matching openwrt kernel driver in image (or later opkg install of module) should be usable.
serial drivers were tested with Holdpeak package to read Multimeter values from shell.
img_20180930_154729_800x

blocktrron:
This multimeter-thing is not embedded into the device? Then we should not include packages for supporting it, users can install them via opkg.

Regarding package welection for the wwan module: You should try to establish a connection using qmi with the two packages mentioned in my former commit. Currently your package selection includes firmware for Ralink USB-wireless adapters, redundant definition of ath9k and serial modem stuff for an entirely other vendor. This stuff might fit into the generated image but does definitely not belong there, so we should ship a minimal package selection to get the features of the device (w/o external accessories) working.

blocktrron:
@cmarxmeier This device is defined in the ar71xx target, thus this image will only work with atheros chipsets (It won't even boot with a MediaTek SoC). So we do not have to explicitly add definitions for such stuff (we also don't do this for other devices). Also GL.iNet explicitly mentions the devices SoC on their webpage.

We cannot cover all use cases with packages we include in upstream gluon. Communities can decide to add/remove specific packages themselves. 

I can't 'do with the package selection what i like ' as i neither own such a device nor I'm a Gluon maintainer. I'm just trying to assist in pushing this PR into a direction it has the possibility to be merged.

rotanid:
@cmarxmeier he's trying to help you get this merged and has experience there, so please try to implement changes he suggests, if you want this device to be supported in Gluon

rotanid:
we won't backport this device to v2018.1.x, closing.

PR #1550, opened 2018-10-04
don't cache index pages which contain meta redirect tags

### LINKED ISSUE #1530: Browser cache, redirections and related issues
When flashing multiple nodes without setting a custom name, they all end up with the same hostname, generated using the first node's host ID. 

I would propose adding "no-cache"-headers to the config mode pages as well as to the status page:

The redirection of different gluon version's status pages differ, if these get cached one can't reach the status page without clearing the browser's cache.

Example: You open the status page using the nextnode address. On Gluon 2018.1 it gets redirected to /cgi-bin/status.
Now, if you try to open a different node's status page running Gluon 2017.1 is not possible, since you get redirected automatically and that path does not exist in older Gluon versions.
The same happens if you downgrade to an older Gluon version.

That redirection should be done using HTTP 303 instead of 301, so it doesn't get cached.

rotanid:
fixes #1530

### DIFF
--- a/package/gluon-status-page/files/lib/gluon/status-page/www/index.html
+++ b/package/gluon-status-page/files/lib/gluon/status-page/www/index.html
@@ -1,6 +1,9 @@
 <!DOCTYPE html>
 <html xmlns="http://www.w3.org/1999/xhtml">
 	<head>
+		<meta http-equiv="Cache-Control" content="no-cache, no-store, must-revalidate" />
+		<meta http-equiv="Pragma" content="no-cache" />
+		<meta http-equiv="Expires" content="0" />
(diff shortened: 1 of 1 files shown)


MyIgel:
Is there a possibility to set the equivalent headers?

```
Cache-Control: no-cache, no-store, must-revalidate
Pragma: no-cache
Expires: 0
```

mweinelt:
Much more difficult because we simply want the redirects not to be cached and the webservers just delivers static files.

NeoRaider:
We should be able to add the options mentioned by @MyIgel as meta http-equiv as well.

rotanid:
@NeoRaider i implemented this now.

rotanid:
backported this to v2018.1.x branch

PR #1551, opened 2018-10-07
Add gluon support for GL.iNet GL-MiFi

cmarxmeier:


NeoRaider:
Why are all these packages necessary? We already include all packages that OpenWrt includes by default; ideally, there should be no model-specific package lists in Gluon (the only exception being ath10k-ct at the moment, which we will hopefully get rid of when we drop IBSS support).

cmarxmeier:
all kernel modules from manufacturer openwrt image were included cause most  communities do not offer kernel packages for download. 
Makes no sense to purchase a 150 Euro mobile router if you cannot access its internal sd-card and all hardware features it offers without compiling gluon yourself.

mweinelt:
We are looking for a shared set of packages that make the device usable but not overloaded with things people might eventually use.

If a kmod or firmware package is required to enable hardware support it should ideally land directly in OpenWrt. Think WiFi, 3G/4G, USB-Controller, hardware that is directly attached to the device when shipped.

If the kmod or firmware, like often the case with WiFi, doesn't suit the purpose of Gluon, namely Mesh and AP at the same time, a local overwrite in Gluon is appropriate.

If additional packages are needed it is up to the community to decide which to include via their `site.mk`¹, and lots of communities do. After the build process all built modules are in `output/modules/` and can be served via a simple HTTP server serving static files and configured in `site.conf` to be pulled by opkg. So even individuals can include some packages later on, if the community allows for it.

We will not be tracking such a vast package list downstream.

[1] https://git.darmstadt.ccc.de/ffda/site/blob/master/site.mk#L51
[2] https://gluon.readthedocs.io/en/latest/user/site.html (search for "opkg")

cmarxmeier:
device is shipped with sd-card slot onboard,  mini pci-e slot for lte modem and nano sim slot on mainboard
see http://download.gl-inet.com/download/mifi/ for more documentation

cmarxmeier:
http://images.freifunk-rhein-sieg.net holds 2018.1.x sysupgrade images for gl-mifi for RSK-Domains and 3 Troisdorf domains. Kernel related packages are offered accordingly in package directory beside image folder.

mweinelt:
> device is shipped with sd-card slot onboard, mini pci-e slot for lte modem and nano sim slot on mainboard
Great, so thats a change that should be done in OpenWrt.

> http://images.freifunk-rhein-sieg.net holds 2018.1.x sysupgrade images for gl-mifi 
The issue is not that we don't believe it works, just the requirement to do things properly, so the maintenance burden is not unnecessarily repeated downstream and everyone can profit from upstream changes.

targets/ar71xx-generic
```diff
@@ -9,6 +9,7 @@ if [ "$GLUON_WLAN_MESH" = 'ibss' ]; then
 fi
 
 
+
```
mweinelt: drop this newline

cmarxmeier:
device already handed out back to its owner roman@freifunk-troisdorf.de
I'm currently busy with https://map.freifunk-troisdorf.de/#!/de/map/0c806307e7ac for 100*wlan project in Troisdorf - so please no unnessassary notifiers for already done gl-mifi.

rotanid:
this can be reopened as soon as someone is willing to do the testing according to the checklist. closing.

PR #1552, opened 2018-10-09
docs: move site configuration repositories list to the wiki

mweinelt:
allows for easier editing and hopefully more updates.


christf:
@mweinelt I added the linked page to the sidebar. I think it deserves some visibility.

Issue #1377, opened 2018-04-27
Scheduled domain migration


NeoRaider:
As proposed in #1364, there should be a way to automatically migrate between mesh domains, ideally scheduled to avoid breaking meshing for longer than necessary. Such a scheduled migration could be used for a number of migration scenarios, like switching between VXLAN and legacy wired, IBSS and 11s, or to another wifi channel. Switching routing protocols, e.g. batadv 14 -> 15 or batadv -> babel are not possible though, as we can't install them in the same firmware. We could also do away with IBSS support when we drop VXLAN this way.

I think a scheduled migration is superior to parallel operation of mesh technologies (like IBSS+11s, or VXLAN+legacy), as weak nodes don't deal well with the additional traffic (especially in dense meshes where there are already multiple neighbours without seeing each twice).

One open question would be how to handle the scheduled switch when the time can't be determined because there is no mesh connection. Possible approaches include:
* Signalling between nodes (broadcasts on wired mesh, beacon stuffing for wireless)
* Alternating between both domains periodically (maybe switching every hour?) until an uplink connection is possible again

T-X:
> I think a scheduled migration is superior to parallel operation of mesh technologies

In my opinion this depends. For Freifunk networks which often have many nodes and nodes with a limited amount of RAM I agree.

However I wouldn't say "superior". A parallel operation has its own advantages, too:

* Safer upgrade process: The new protocols can be verified before disabling the old one.
* With a scheduled update we have one shot and no more updates spread over weeks. Though some fallback/alternating between domains might mitigate this, this would still add complexity and might potentially cause unintended network splits?
* Parallel operation would be nice to have for comparisons and to be able to test new protocols in a large mesh network without affecting users before fully switching to it.
* Backup: Regarding protocols we so far have a very homogeneous setup. This makes us vulnerable to yet unknown but critical bugs in a mesh protocol. Running a secondary, auxiliary mesh protocol with slower update intervals could be interesting as a backup to increase the overall robustness.

But like I said, this only works if devices are beefy enough and/or if the network isn't too large.

mweinelt:
Fixed in https://github.com/freifunk-gluon/gluon/pull/1555

PR #1555, opened 2018-10-18
gluon-scheduled-domain-switch: add package

blocktrron:
This package allows to automatically switch to another domain, either at a given point in time or after the node was offline long enough.

It's primary goal is to allow for communities which still use IBSS to migrate over to 802.11s without having to run both protocols at the same time, which might lead to overloaded routers.

Depending how #1377 is scoped, this might close it.

### DIFF
--- a/docs/package/gluon-scheduled-domain-switch.rst
+++ b/docs/package/gluon-scheduled-domain-switch.rst
@@ -1,0 +1,38 @@
+gluon-scheduled-domain-switch
+=============================
+
+This package allows to switch a routers domain at a given point
+in time. This is needed for switching between incompatible transport
+protocols (e.g. 802.11s and IBSS or VXLAN).
+
+Nodes will switch when the defined *switch-time* has passed. In case the node was
+powered off while this was supposed to happen, it might not be able to aquire the
+correct time. In this case, the node will switch after it has not seen any gateway
+for a given period of time.
+
+site.conf
+---------
+All those settings have to be defined exclusively in the domain, not the site.
+
+domain_switch : optional (needed for domains to switch)
+    target_domain :
+        - target domain to switch to
+    switch_after_offline_mins :
+        - amount of time without reachable gateway to switch unconditionally
+    switch_time :
+        - UNIX epoch after which domain will be switched
+    connection_check_targets :
+        - array of IPv6 addresses which are probed to determine if the node is
+	  connected to the mesh
+
+Example::
+
+  domain_switch = {
+    target_domain = 'new_domain',
+    switch_after_offline_mins = 120,
+    switch_time = 1546344000, -- 01.01.2019 - 12:00 UTC
+    connection_check_targets = {
+      '2001:4860:4860::8888',
+      '2001:4860:4860::8844',
+    },
+  },
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
@@ -260,3 +260,12 @@ function foreach_radio(uci, f)
 		end
 	end
 end
+
+function get_uptime()
+	local uptime_file = readfile("/proc/uptime")
+	if uptime_file == nil then
+		-- Something went wrong reading "/proc/uptime"
+		return nil
+	end
+	return tonumber(uptime_file:match('^[^ ]+'))
+end
--- a/package/gluon-scheduled-domain-switch/Makefile
+++ b/package/gluon-scheduled-domain-switch/Makefile
@@ -1,0 +1,13 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-scheduled-domain-switch
+PKG_VERSION:=1
+
+include ../gluon.mk
+
+define Package/gluon-scheduled-domain-switch
+  TITLE:=Allows scheduled migrations between domains
+  DEPENDS:=+gluon-core @GLUON_MULTIDOMAIN
+endef
+
+$(eval $(call BuildPackageGluon,gluon-scheduled-domain-switch))
--- a/package/gluon-scheduled-domain-switch/check_site.lua
+++ b/package/gluon-scheduled-domain-switch/check_site.lua
@@ -1,0 +1,6 @@
+if need_table(in_domain({'domain_switch'}), check_domain_switch, false) then
+	need_domain_name(in_domain({'domain_switch', 'target_domain'}))
+	need_number(in_domain({'domain_switch', 'switch_after_offline_mins'}))
+	need_number(in_domain({'domain_switch', 'switch_time'}))
+	need_string_array_match(in_domain({'domain_switch', 'connection_check_targets'}), '^[%x:]+$')
+end
--- a/package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-check-connection
+++ b/package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-check-connection
@@ -1,0 +1,36 @@
+#!/usr/bin/lua
+
+local unistd = require 'posix.unistd'
+local util = require 'gluon.util'
+local site = require 'gluon.site'
+
+local offline_flag_file = "/tmp/gluon_offline"
+local is_offline = true
+
+-- Check if domain-switch is scheduled
+if site.domain_switch() == nil then
+	-- Switch not applicable for current domain
+	os.exit(0)
+end
+
+-- Check reachability of pre-defined targets
+for _, ip in ipairs(site.domain_switch.connection_check_targets()) do
+	local exit_code = os.execute("ping -c 1 -w 10 " .. ip)
+	if exit_code == 0 then
+		is_offline = false
+		break
+	end
+end
+
+if is_offline then
+	-- Check if we were previously offline
+	if unistd.access(offline_flag_file) then
+		os.exit(0)
+	end
+	-- Create offline flag
+	local f = io.open(offline_flag_file, "w")
+	f:write(tostring(util.get_uptime()))
+	f:close()
+else
+	os.remove(offline_flag_file)
+end
--- a/package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
+++ b/package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
@@ -1,0 +1,67 @@
+#!/usr/bin/lua
+
+local uci = require('simple-uci').cursor()
+local unistd = require 'posix.unistd'
+local util = require 'gluon.util'
+local site = require 'gluon.site'
+
+-- Returns true if node was offline long enough to perform domain switch
+function switch_after_min_reached()
+	if not unistd.access("/tmp/gluon_offline") then
+		return false
+	end
+
+	local switch_after_sec = site.domain_switch.switch_after_offline_mins() * 60
+
+	local current_uptime = util.get_uptime()
+	if current_uptime == nil then
+		return false
+	end
+
+	local f = util.readfile("/tmp/gluon_offline")
+	if f == nil then
+		return false
+	end
+	local offline_since = tonumber(f)
+
+	local offline_time_sec = current_uptime - offline_since
+
+	if offline_time_sec > switch_after_sec then
+		return true
+	end
+	return false
+end
+
+-- Returns true in case switch time has passed
+function switch_time_passed()
+	local current_time = os.time()
+	local switch_time = site.domain_switch.switch_time()
+
+	return switch_time < current_time
+end
+
+if site.domain_switch() == nil then
+	-- Switch not applicable for current domain
+	print("No domain switch defined for the current domain.")
+	os.exit(0)
+end
+
+local current_domain = uci:get("gluon", "core", "domain")
+local target_domain = site.domain_switch.target_domain()
+
+if target_domain == current_domain then
+	-- Current and target domain are equal
+	print("Domain '" .. target_domain .. "' equals current domain.")
+	os.exit(1)
+end
+
+if not switch_after_min_reached() and not switch_time_passed() then
+	-- Neither switch-time passed nor switch_after_min reached
+	os.exit(0)
+end
+
+uci:set("gluon", "core", "domain", target_domain)
+uci:commit("gluon")
+
+os.execute("gluon-reconfigure")
+os.execute("reboot")
--- a/package/gluon-core/check_site.lua
+++ b/package/gluon-core/check_site.lua
@@ -3,15 +3,6 @@ need_string(in_site({'site_name'}))
 
 -- this_domain() returns nil when multidomain support is disabled
 if this_domain() then
-	function need_domain_name(path)
-		need_string(path)
-		need(path, function(default_domain)
-			local f = io.open(os.getenv('IPKG_INSTROOT') .. '/lib/gluon/domains/' .. default_domain .. '.json')
-			if not f then return false end
-			f:close()
-			return true
-		end, nil, 'be a valid domain name')
-	end
 	need_domain_name(in_site({'default_domain'}))
 
 	need_table(in_domain({'domain_names'}), function(domain)
--- a/package/gluon-scheduled-domain-switch/luasrc/lib/gluon/upgrade/950-gluon-scheduled-domain-switch
+++ b/package/gluon-scheduled-domain-switch/luasrc/lib/gluon/upgrade/950-gluon-scheduled-domain-switch
@@ -1,0 +1,20 @@
+#!/usr/bin/lua
+
+local json = require 'jsonc'
+local site = require 'gluon.site'
+local unistd = require 'posix.unistd'
+
+local cronfile = "/usr/lib/micron.d/gluon-scheduled-domain-switch"
+
+-- Check if domain switch is scheduled
+if site.domain_switch() == nil then
+	-- In case no domain switch is scheduled, remove cronfile
+	os.remove(cronfile)
+	os.exit(0)
+end
+
+-- Only in case domain switch is scheduled
+local f = io.open(cronfile, "w")
+f:write("* * * * *  /usr/bin/gluon-check-connection\n")
+f:write("*/5 * * * *  /usr/bin/gluon-switch-domain\n")
+f:close()
--- a/scripts/check_site.lua
+++ b/scripts/check_site.lua
@@ -305,6 +305,15 @@ function need_array_of(path, array, required)
 	return need_array(path, function(e) need_one_of(e, array) end, required)
 end
 
+function need_domain_name(path)
+	need_string(path)
+	need(path, function(domain_name)
+		local f = io.open(os.getenv('IPKG_INSTROOT') .. '/lib/gluon/domains/' .. domain_name .. '.json')
+		if not f then return false end
+		f:close()
+		return true
+	end, nil, 'be a valid domain name')
+end
 
 local check = assert(loadfile())
 


blocktrron:
I've just noticed there are some things to do in terms of code-style (Tabs vs. Spaces), will fix that soon.

docs/package/gluon-scheduled-domain-switch.rst
```diff
@@ -0,0 +1,30 @@
+gluon-scheduled-domain-switch
+=============================
+
+This package allows to switch a routers domain at a given point
+in time. This is needed e.g. for communities switching mesh-protocols
+(IBSS / 802.11s) without using both simultaneously.
```
mweinelt: switching between incompatible mesh or wireless transport protocols

docs/package/gluon-scheduled-domain-switch.rst
```diff
@@ -0,0 +1,30 @@
...
+
+Nodes will switch if this defined *switch-time* has passed. In case the node was
```
mweinelt: if the defined

docs/package/gluon-scheduled-domain-switch.rst
```diff
@@ -0,0 +1,30 @@
...
+powered off when this was supposed to happen, it might not be able to set it's clock
```
mweinelt: while this was supposed to happen, it might not be able to get aquire the correct time.

docs/package/gluon-scheduled-domain-switch.rst
```diff
@@ -0,0 +1,30 @@
...
+        - amount of time without reachable gateway to switch unconditionally
+    switch_time :
+        - UNIX epoch after which domain will be switched
+
+Example::
+
+  domain_switch = {
+    target_domain = 'new_domain',
+    switch_after_min = 120,
+    switch_time = 1546344000, -- 01.01.2019 - 12:00 GMT
```
mweinelt: UTC

docs/package/gluon-scheduled-domain-switch.rst
```diff
@@ -0,0 +1,30 @@
...
+    switch_after_min :
+        - amount of time without reachable gateway to switch unconditionally
+    switch_time :
+        - UNIX epoch after which domain will be switched
+
+Example::
+
+  domain_switch = {
+    target_domain = 'new_domain',
+    switch_after_min = 120,
```
mweinelt: switch_after_offline_mins

package/gluon-scheduled-domain-switch/Makefile
```diff
@@ -0,0 +1,13 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-scheduled-domain-switch
+PKG_VERSION:=1
+
+include ../gluon.mk
+
+define Package/gluon-scheduled-domain-switch
+  TITLE:=Allows to switch to another domain at a site-specified time
```
mweinelt: Allows scheduled migrations between domains

package/gluon-scheduled-domain-switch/files/usr/bin/gluon-check-gw-reachability
```diff
@@ -0,0 +1,16 @@
+#! /bin/sh
```
mweinelt: whitespace unnecessary

blocktrron:
Addressed all issues pointed out by @mweinelt

docs/package/gluon-scheduled-domain-switch.rst
```diff
@@ -2,12 +2,12 @@ gluon-scheduled-domain-switch
 =============================
 
 This package allows to switch a routers domain at a given point
-in time. This is needed e.g. for communities switching mesh-protocols
-(IBSS / 802.11s) without using both simultaneously.
+in time. This is needed e.g. for communities switching between
```
mweinelt: drop the `e.g.` and `communities`

docs/package/gluon-scheduled-domain-switch.rst
```diff
@@ -2,12 +2,12 @@ gluon-scheduled-domain-switch
...
+incompatible wireless transport protocols (e.g. 802.11s and IBSS).
 
-Nodes will switch if this defined *switch-time* has passed. In case the node was
-powered off when this was supposed to happen, it might not be able to set it's clock
-correctly. In this case, the node will switch after it has not seen any gateway
+Nodes will switch if the defined *switch-time* has passed. In case the node was
```
mweinelt: when the defined

docs/package/gluon-scheduled-domain-switch.rst
rotanid: issue in example still open

genofire:
Maybe a `map` of domains is good:
 e.g. `old_domain1` to `new_domain1` and `old_domain2` to `new_domain2`

blocktrron:
@genofire This is indirectly archieved by the abitlity to put the settings into the respective domain files. This way, it is also possible to e.g. configure `switch_after_offline_mins `on a per-domain base.

docs/package/gluon-scheduled-domain-switch.rst
blocktrron: fixed

docs/package/gluon-scheduled-domain-switch.rst
blocktrron: fixed

docs/package/gluon-scheduled-domain-switch.rst
blocktrron: fixed

package/gluon-scheduled-domain-switch/files/usr/bin/gluon-check-gw-reachability
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+GW_COUNT=$(batctl gwl -H 2> /dev/null | wc -l)
```
christf: this does work exclusively for batman. The IBSS => 802.11s switch *might* be relevant or other mesh protocols too.
OTOH it is unlikely that someone builds a new network with more recently supported mesh protocol.

mweinelt: As only one mesh protocol can be selected at this time a migration between them is impossible. Correct me if I'm wrong.

blocktrron: Fixed: https://github.com/freifunk-gluon/gluon/pull/1555/commits/82df55ea0049736cd0ade23ea5fa31c621ad13d4#diff-0219e17e415d8777b3329280a95a57a5R11

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
```diff
@@ -0,0 +1,78 @@
...
+	local switch_after_sec = site.domain_switch.switch_after_offline_mins() * 60
+
+	local f = io.open("/tmp/gluon_gw_unreach")
+	if not f then
+		return false
+	end
+	local offline_since = f:read('*a')
+	f:close()
+
+	local offline_time_sec = current_epoch - offline_since
```
T-X: Not a Lua expert. But could this over- or underflow in 2038 or could the automatic string to float conversion result in wrong values? (date --date="@`echo $((2**31))`" => "Tue 19 Jan 04:14:08 CET 2038".

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
```diff
@@ -0,0 +1,78 @@
...
+
+	if offline_time_sec > switch_after_sec then
```
T-X: Could this result in a race condition right after boot, when no time was fetched via NTP yet?

For instance like:
1) Device booted, timestamp is equivalent to something like 1970
2) No gateways were discovered yet
3) /usr/sbin/gluon-check-gw-reachability writes a very low value to /tmp/gluon_gw_unreach
4) NTP updates the time a recent one.
5) /usr/sbin/gluon-check-gw-reachability is not called yet (or it is called but with no gateways found yet, or batctl erroring)
6) Now /usr/bin/gluon-switch-domain is called and switches immediately as it thinks there was unreachability for >40 years.

Maybe add some sanity check for any retrieved time? For instance, if any retrieved time is < 2018 then do not switch?

T-X:
If I remember correctly then @neoraider was thinking about some rollback mechanism, too. So, if after switching to the new domain settings and if gateways are unreachable then, then revert to the previous domain. Could probably be implemented and added later, too, but @blocktrron, have you spend some thoughts on something like this?

Would it be like switching back and forth between old and new domain settings after some timeout? (Could that run into some undesired side-effects?)

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
blocktrron: The initial time after booting correlates to the time the image was built (and maybe the latest date in /etc, but i'm not sure on this one).

So a sanity check based on a minimal value will not help as all images will have been built in 2018.

A possible approach would be to compare the offline-time to the systems uptime, what do you think about that? This should also mitigate the possible race condition reading and writing this file.

blocktrron:
A little bit of background here: This is pretty much a stripped down, tidied up version of the [package ](https://github.com/freifunk-darmstadt/ffda-packages/tree/master/ffda-domain-director) we used to migrate the network of Freifunk Darmstadt.

We thought about this also but dropped this idea due to the fact we might run into issues in larger meshes, switching back an forth between domains. We never wanted nodes to go back. If you run into the fallback switch, you will see a node at the point all other nodes have switched, even if this means you will have no mesh connection for ~5 days (IMHO. a reasonable timespan between rolling out the scheduled domain switch and the actual switch-date is 1 Week or maybe 10 days).

So a back-and-forth switching is not necessary, all nodes will be back online after a week.

As long as undesired side-effects go: I've never thought about potential problems, except for the time to stabilize a larger mesh with different fallback-intervals and domain states. This will sort out itself after time, but i think the one-time switch is superior.

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
T-X: Using uptime instead of the Unix epoch for this issue is a great idea! I like that one.

I do not quite understand your last remark, how it fixes the read/write race condition though. The following could still happen, couldn't it?

1) Device booted, gets connected and sees some gateways
2) Uptime increases to n days. Still many days from the configured switching date away, though.
3) Gateway connectivity is lost somehow, for a short amount of time. But enough to make /usr/sbin/gluon-check-gw-reachability start writing its file. 
4) Now  /usr/bin/gluon-switch-domain is called while gluon-check-gw-reachability has only written the first (few) digit(s) of the current uptime so far. /usr/bin/gluon-switch-domain then compares this cut value to its current uptime and thinks ~n days of unreachability have passed.

T-X:
> So a back-and-forth switching is not necessary, all nodes will be back online after a week.

Unless you've made some mistake in the new domain settings. Or have other bugs that only show up with the new domain settings.

While the domain switching might seem safer than the autoupdater because it does not need to flash the whole system, that might make it less safer without a revert option: With the autoupdater you can validate your images/changes via beta branches and an updating timespan. With the domain-switch it's an "all-in / all-jump-now" approach.

With the presumption that humans will make mistakes and will create broken domain settings (or will run into new protocol or driver bugs that will only surface in the combination of the new domain settings and this particular network), do we have enough safety measures for the scheduled-domain-switch approach to accommodate for that?

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
blocktrron: You are absolutely right, in this case the switch would still happen.

mweinelt:
> Unless you've made some mistake in the new domain settings. Or have other bugs that only show up with the new domain settings.

Certainly, but I'd argue that as usual any firmware doing such a major migration should be properly tested ahead of time. Additionally signing a firmware manifest should imply that it has received testing, more so with a domain migration, anything else feels unreasonable. 

Apart from misconfiguration I don't believe the general concept of this PR is particularly error-prone. It makes sure that nodes definitely migrate at some point, it's reliability basically comes down to how reliable you can roll out new firmware before the switch time is up. I don't necessarily believe wiggling back and forth between old and new configuration is strictly necessary, although it would be a nice addition.

The primary culprit for me is that you have to configure the switch time at build time, unable to further control and monitor the switch at run time, like in a remote-controlled approach.
- Set it too long, you're fine, it just takes forever.
- Set it too short, you risk splitting your network.

Fortunately with a non-parallel approach we don't run into load issues along the way. I'd say most communities can reliably roll out a firmware to most devices in about a week, with most nodes receiving the update very quickly, and only a few select routers lagging behind, because most famously
- their uplink is bad, 
- they were offline between firmware rollout and migration

So I'd say a recommendation of at least 10-14 days between firmware rollout and the actual switch would be suitable. Obviously more days are needed if you have configured a slower rollout.

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
NeoRaider: Lua switches to doubles when integers are insufficient (originally, Lua 5.1 did not support integers at all, and it always used floating-point math; OpenWrt backported integer support from newer Lua versions). I'm not sure of what kind of conversion issues you are thinking.

I think adding a *tonumber* around the `f:read()` would make the code a bit clearer.

I assume the current os.time() implementation will break in 2038, but that's a different story... glibc and kernel developers are working on that issue.

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
NeoRaider: Any sane date implementation will use stdio for its output, so short writes (up to 1KB in musl, I think 4KB is common in other libcs) happen atomically (with some caveats, but let's not go into detail too much here).

If you really want to be sure, you could write to a temporary file and then move it into place, but I don't consider this an actual issue.

T-X: @blocktrron: Why not switching to store the uptime in /tmp/gluon_gw_unreach, too? Seems to me that this would simplify the checks a lot. Just compare the uptime in that file and the current uptime. No need to worry about NTP at all (at least for the switch_after_min_reached() check).

blocktrron: @T-X great idea! I've implemented your suggestion, so we do not rely on the date and time anymore.

blocktrron:
Just tested my changes and they behave as expected. Also switched to using system uptime instead of date and time to determine the offline-duration of the node.

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
```diff
@@ -0,0 +1,84 @@
...
+	if not unistd.access("/tmp/gluon_gw_unreach") then
+		return false
+	end
+
+	local current_epoch = os.time()
+	local switch_after_sec = site.domain_switch.switch_after_offline_mins() * 60
+
+	local current_uptime_handle = io.popen('cat /proc/uptime | cut -d "." -f1')
+	local current_uptime = tonumber(current_uptime_handle:read('*all'))
+	current_uptime_handle:close()
```
NeoRaider: Please don't use popen here, read the file directly and use Lua to split the string.

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
```diff
@@ -0,0 +1,84 @@
...
+
+	local f = io.open("/tmp/gluon_gw_unreach")
+	if not f then
+		return false
+	end
+	local offline_since = tonumber(f:read('*a'))
```
NeoRaider: You can simplify this a bit by using gluon.util.readfile().

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
```diff
@@ -0,0 +1,84 @@
...
+	-- Neither switch-time passed nor switch_after_min reached
+	return 0
+end
+
+uci:set("gluon", "core", "domain", target_domain)
+uci:commit("gluon")
+
+local gluon_reconfigure_handle = io.popen("gluon-reconfigure")
+gluon_reconfigure_handle:read('*all')
+gluon_reconfigure_handle:close()
```
NeoRaider: Use os.execute for gluon-reconfigure as well.

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
```diff
@@ -0,0 +1,84 @@
...
+if not domain_exists(target_domain) then
+	-- Domain does not exist
+	print("Domain '" .. target_domain .. "' does not exist.")
+	return 1
+end
+
+if target_domain == current_domain then
+	-- Current and target domain are equal
+	print("Domain '" .. target_domain .. "' equals current domain.")
+	return 0
```
NeoRaider: I think this should return 1 as well, as it's not a proper configuration.

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
```diff
@@ -0,0 +1,84 @@
...
+		return true
+	end
+
+	return false
+end
+
+if site.domain_switch() == nil or site.domain_switch.target_domain() == nil or site.domain_switch.switch_after_offline_mins() == nil or site.domain_switch.switch_time() == nil then
+	-- Switch not applicable for current domain
+	print("No domain switch defined for the current domain.")
+	return 1
```
NeoRaider: I guess this should return 0, as not all domains need to have a domain switch configured?

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
```diff
@@ -0,0 +1,84 @@
...
+	local switch_time = site.domain_switch.switch_time()
+
+	if switch_time < current_epoch then
+		return true
+	end
+
+	return false
+end
+
+if site.domain_switch() == nil or site.domain_switch.target_domain() == nil or site.domain_switch.switch_after_offline_mins() == nil or site.domain_switch.switch_time() == nil then
```
NeoRaider: All fields *must* be checked in a site checker script. The checks for the individual fields could be omitted here then.

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
```diff
@@ -0,0 +1,84 @@
...
+local current_domain = uci:get("gluon", "core", "domain")
+local target_domain = site.domain_switch.target_domain()
+
+if not domain_exists(target_domain) then
+	-- Domain does not exist
+	print("Domain '" .. target_domain .. "' does not exist.")
+	return 1
+end
+
+if target_domain == current_domain then
```
NeoRaider: These checks (`domain_exists(target_domain)` and `target_domain == current_domain`) should be added to a site checker script as well.

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
blocktrron: This does not work as os.execute does not wait for termination of gluob-reconfigure.

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
blocktrron: IMHO this is not the case as one might want to migrate only a specific domain alias while remaining in the same "technical" domain.

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
NeoRaider: It does - os.execute returns the exit code of the passed command.

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
NeoRaider: Ah, this makes sense. The first condition should still be checked at build time.

blocktrron:
Updated this PR. Tested on one node without issues.

I hope i did catch everything. :smile:

blocktrron:
Fixed two minor inconsistencys (see fixup commit).

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -253,3 +253,12 @@ function foreach_radio(uci, f)
...
 	end
 end
+
+function get_uptime()
+	local uptime_file = readfile("/proc/uptime")
+	if uptime_file == nil then
+		-- Something went wrong reading "/proc/uptime"
+		return nil
+	end
+	return tonumber(string.sub(uptime_file, 0, string.find(uptime_file, "%.") - 1))
```
NeoRaider: I think it would make sense to split at the space, not the dot. How about this?
```lua
return tonumber(uptime_file:match('^[^ ]+'))
```

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -253,3 +253,12 @@ function foreach_radio(uci, f)
...
+end
```
NeoRaider: Missing newline

package/gluon-scheduled-domain-switch/check_site.lua
```diff
@@ -0,0 +1,6 @@
+if need_table(in_domain({'domain_switch'}), check_domain_switch, false) then
+	need_domain_name(in_domain({'domain_switch', 'target_domain'}), true)
+	need_number(in_domain({'domain_switch', 'switch_after_offline_mins'}), true)
+	need_number(in_domain({'domain_switch', 'switch_time'}), true)
+	need_string_array_match(in_domain({'domain_switch', 'connection_check_targets'}), '^[%x:]+$', true)
```
NeoRaider: It's common to leave out the *required* argument when it is the default (true).

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-check-connection
```diff
@@ -0,0 +1,36 @@
...
+-- Check reachability of pre-defined targets
+for _, ip in ipairs(site.domain_switch.connection_check_targets()) do
+	local exit_code = os.execute("ping -c 1 -w 10 " .. ip)
+	if exit_code == 0 then
+		is_offline = false
+		break
+	end
+end
+
+if is_offline == true then
```
NeoRaider: Drop the `== true`

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-check-connection
```diff
@@ -0,0 +1,36 @@
...
+	if unistd.access(offline_flag_file) then
+		os.exit(0)
+	end
+	-- Create offline flag
+	local f = io.open(offline_flag_file, "w")
+	f:write(tostring(util.get_uptime()))
+	f:close()
+else
+	os.remove(offline_flag_file)
+end
```
NeoRaider: Missing newline

docs/package/gluon-scheduled-domain-switch.rst
```diff
@@ -0,0 +1,38 @@
+gluon-scheduled-domain-switch
+=============================
+
+This package allows to switch a routers domain at a given point
+in time. This is needed for switching between incompatible wireless
+transport protocols (e.g. 802.11s and IBSS).
```
skorpy2009: not only wireless but also for vxlan?

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
```diff
@@ -0,0 +1,72 @@
...
+local util = require 'gluon.util'
+local site = require 'gluon.site'
+
+-- Returns true if node was offline long enough to perform domain switch
+function switch_after_min_reached()
+	if not unistd.access("/tmp/gluon_offline") then
+		return false
+	end
+
+	local current_epoch = os.time()
```
NeoRaider: Just call the variable *current_time*?

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
```diff
@@ -0,0 +1,72 @@
...
+
+	if offline_time_sec > switch_after_sec then
+		return true
+	end
+	return false
+end
+
+-- Returns true in case switch time has passed
+function switch_time_passed()
+	local current_epoch = os.time()
```
NeoRaider: Just call the variable *current_time*?

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
```diff
@@ -0,0 +1,72 @@
...
+	local switch_time = site.domain_switch.switch_time()
+
+	if switch_time < current_epoch then
+		return true
+	end
```
NeoRaider: Can be simplified to `return switch_time < current_epoch`

package/gluon-scheduled-domain-switch/Makefile
```diff
@@ -0,0 +1,13 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-scheduled-domain-switch
+PKG_VERSION:=1
+
+include ../gluon.mk
+
+define Package/gluon-scheduled-domain-switch
+  TITLE:=Allows scheduled migrations between domains
+  DEPENDS:=+gluon-core +gluon-mesh-batman-adv
```
NeoRaider: I think the dependency on batman-adv can be dropped now?

docs/package/gluon-scheduled-domain-switch.rst
blocktrron: yes

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
blocktrron: yup, works too

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
blocktrron: fixed

package/gluon-scheduled-domain-switch/check_site.lua
blocktrron: removed the explicit `true`

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-check-connection
blocktrron: fixed

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-check-connection
blocktrron: fixed

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
blocktrron: I#ve dropped it altogether as it isn't used in this function.

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
blocktrron: fixed

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
blocktrron: corrected

package/gluon-scheduled-domain-switch/Makefile
blocktrron: indeed it can now be dropped.

blocktrron:
Updated PR.

 - Addressed everything @NeoRaider and @skorpy2009 pointed out.
 - Rebased onto current master

package/gluon-scheduled-domain-switch/Makefile
```diff
@@ -0,0 +1,13 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-scheduled-domain-switch
+PKG_VERSION:=1
+
+include ../gluon.mk
+
+define Package/gluon-scheduled-domain-switch
+  TITLE:=Allows scheduled migrations between domains
+  DEPENDS:=+gluon-core
```
NeoRaider: I found another thing... Please add `@GLUON_MULTIDOMAIN` to DEPENDS, so the package can only be selected in multi-domain configurations.

blocktrron: Fixed

PR #1559, opened 2018-10-24
docs: features/autoupdater: fix typo and add GLUON_TARGET for make clean

bobcanthelpyou:


docs/features/autoupdater.rst
```diff
@@ -8,10 +8,10 @@ Building Images
 
 By default, the autoupdater is disabled (as it is usually not helpful to have unexpected updates
 during development), but it can be enabled by setting the variable GLUON_BRANCH when building
-to override the default branch set in the set in the site configuration.
+to override the default branch set in the site configuration.
```
rubo77: maybe we can formulate this differently to make clear, that the value set in the site.conf is NEVER used at all. either you override it with GLUON_BRANCH or the autoupdater is disabled, thus doesn't use the value set  in the site.conf either.

docs/features/autoupdater.rst
```diff
@@ -35,9 +35,9 @@ A fully automated nightly build could use the following commands:
 ::
 
     git pull
-    (cd site && git pull)
+    (git -C site pull)
     make update
-    make clean
+    make clean GLUON_TARGET=ar71xx-generic
```
rubo77: why do you have to add a target here?

bobcanthelpyou: `make clean` doesn't work for me, with GLUON_TARGET only 

It is mention in the Getting Started guide, too
https://gluon.readthedocs.io/en/v2018.1.x/user/getting_started.html#cleaning-the-build-tree

docs/features/autoupdater.rst
bobcanthelpyou: Which default branch is set building an image without `GLUON_BRANCH`?

rubo77: the branch set in site.conf is set, but the autoupdater will be disabled too, so the branch that is set is not used unless you re-enable the autoupdater by hand. I think this is not a very good solution right now, because there is no disable setting in site.conf. You use the fact that GLUON_BRANCH is missing during build as "switch" to disable the autoupdater.

Anyway as long as it is like it is, we should explain this here, so someone new will understand this behaviour.

Issue #1510, opened 2018-08-08
Config Mode Hostname


mweinelt:
In config mode provide a hostname that the node resolves to itself, so we can use a name to reference to its IP addresses. Just like we already do with the nextnode address.

This requires us to set the nameserver DHCP option, if we don't do that already, and set up RDNSS.

#### site.conf
```lua
config_mode = {
    name = { 'setup', 'setup.ffda.io' },
}
```

NeoRaider:
Unfortunately, this conflicts with the idea that the user is in a private WLAN for internet access while connected to the node in config mode (e.g. for loading a map to display). The node should not advertise DNS in config mode to avoid breaking internet connectivity...

kaechele:
Maybe [umdns](https://openwrt.org/docs/guide-developer/mdns#umdns) could be useful here?

PR #1563, opened 2018-11-04
[RFC] gluon-setup-mode: add wifi interfaces

belzebub40k:
Since more and more manufactures drop LAN interfaces on their notebooks and don't include a wired NIC in form of an adapter we should provide a way to configure a node via WiFi. This PR is just a rough idea to start the discussion on how to implement it. We were already discussing the existing code on the Freifunk Hessen Meetup and came to the conclusion that using the primary_mac as PSK doesn't work for all devices. Other ideas are to reduce the txpower to the minimum and set maximum connections to 1.

### DIFF
--- a/package/gluon-setup-mode/files/lib/gluon/setup-mode/rc.d/S20network
+++ b/package/gluon-setup-mode/files/lib/gluon/setup-mode/rc.d/S20network
@@ -13,6 +13,21 @@ delete_interface() {
         [ "$1" = 'loopback' ] || uci_remove network "$1"
 }
 
+delete_wifi_iface() {
+	uci_remove wireless "$1"
+}
+
+add_wifi_iface() {
+	uci_add wireless wifi-iface "setup_$1"
+	uci_set wireless "setup_$1" ifname "setup${1#radio}"
+	uci_set wireless "setup_$1" network "setup"
+	uci_set wireless "setup_$1" device "$1"
+	uci_set wireless "setup_$1" mode 'ap'
+	uci_set wireless "setup_$1" ssid 'Gluon Setup'
+	uci_set wireless "setup_$1" encryption 'psk2'
+	uci_set wireless "setup_$1" key "$(lua -e 'print(require("gluon.sysconfig").primary_mac:gsub(":",""):lower())')"
+}
+
 prepare_config() {
 (
 	export UCI_CONFIG_DIR=/var/gluon/setup-mode/config
@@ -33,6 +48,14 @@ prepare_config() {
 	uci_set network setup netmask "$SETUP_MODE_NETMASK"
 
 	uci_commit network
+
+	cp /etc/config/wireless "$UCI_CONFIG_DIR"
+
+	config_load wireless
+	config_foreach delete_wifi_iface wifi-iface
+	config_foreach add_wifi_iface wifi-device
+
+	uci_commit wireless
 )
 }
 


mweinelt:
Both accessible at the same time is easily possible.

A few points
- maxassoc only makes sense if we use the first 2.4 GHz radio we find
- add the primary mac into the ESSID so we can distinguish between different devices in config mode
- finding a safe psk might be impossible, but low txpower *might* be sufficient

edeso:
technically config mode can be provided via LAN and WIFI at the same time surely. 
question is if it is desirable by default. an open WLAN with an unprotected config interface is a proper attack vector.

a solution that explicitly expects a WLAN activation eg. 20s reset as opposed to the 10s for config mode now. of course the device must react on button up in this and not as now do the reboot after 10s button down.

rotanid:
in addition to remarks already made: i don't want this enabled by default. this would not only be a security issue but also "break" features like the new OSM map location picker.

rubo77:
it could be an extra package that has to be built in a second firmware branch, so you can choose, which firmware to download (in case you don't have a USB-LAN adapter with you.)

rotanid:
or a button press in config mode, as you already suggested?

mweinelt:
Neither of these ideas makes the approach secure. Not every router has a button for that.

rubo77:
rotanids id sounds fine: when you are in config mode already, long-press the reset button again for 10s, then you get in wifi config mode

edeso:
@mweinelt i like @rotanid 's suggestion. for devices without reset button there is obviously currently a way already to enter config mode. doing the same from within config mode already would be identical to the two times reset press.

rotanid:
@mweinelt i also agree to "lower txpower" and "unique ESSID".
regarding PSK, i vote for a well-known PSK which the community has to set. this way, the config-mode isn't accessible by a random stranger without any effort.
together with low txpower and not-enabled-by-default, i'd be fine with this

belzebub40k:
Ok, to sum this up we want something like this:
- explicitly enable this feature via _site.conf_
- WiFi should only be enabled on user interaction _(reset/wifi/wps button, what ever is available)_
- authentication via site-wide PSK
- unique SSID _(fixed string + hostname/nodeid)_
- `txpower` reduced to minimum
- maybe limit to (first) 2.4 GHz radio and set `maxassoc` to 1

christf:
How about we do this the other way around, like have the node connect to an access point and download a file containing a secret that is used as wpa passphrase?
It still means you have to trust all the nodes on your network if you do this. It also means the node does not  open up to *any* communication source within range.

belzebub40k:
@christf How should the node be able to connect to an AP and how should the WPA key reach the user? Connecting to an AP would only be possible if there is an other node in reach (or an other open one without captive portal) which is very unlikely.

NeoRaider:
"Well-known PSK" is no viable solution IMO. Commercial hardware vendors are forced to give their devices unique keys that are not derived from MAC address etc (at least California has passed such a law by now, other states/countries are hopefully going to follow).

I propose to use one of the following solutions for accessing the config mode from devices without Ethernet:
1. Use a LED + phone app to transmit a PSK (can support Gluon nodes without any Ethernet)
2. Add support for a DHCP client mode (user would connect WAN port to home network, node announces itself via MDNS or something simpler/smaller, user could access config mode over their home WLAN; requires a WAN port to exist)

blocktrron:
I really like the idea of using LEDs to provide an interface to the user. This may also be extended for troubleshooting purposes.

We might have the problem not all devices will have a suitable status-LED (or a LED at all), but i think this should be no blocker.

christf:
it could be downloaded from a "well-known" location using the wan-port.

I do not see a way to connecting to WPA secured network.

yanosz:
I'd prefere WPS PBC (pushbutton). WPA is set up without a PSK, while just one device can be connected per push.

belzebub40k:
WPS PBC sounds like a good idea but it is not supported by hostapd-mini. It would only be possible if we use an other package which is not possible on 4MB devices. If the gluon maintainers are fine with using different hostapd/wpad packages on devices/targets with more than 4MB flash I can try to implement it.

mweinelt:
DHCP Client on WAN combined with mDNS sounds easiest and would also resolve #1510. It also has the benefit that it's in most cases the correct way of connecting your node to your WAN.

Regarding the inclusion of a bigger hostapd package - this would likely make this package non-default. I'm not sure about our stance on per target features though.

blocktrron:
Some ideas:

 - Regarding Key derivation from the MAC-Adress: What about using a randomized MAC-Address for the WiFi interface an utilizing the Primary MAC as a PSK? You might be able to correlate a device in setup mode using the Map and get the primary MAC this way, but this might be a bit of a stretch.

 - Using very low TX Power for setup-mode. This way we can limit the range in which the Node can be configured (but might run into problems with shitty WiFi adapters?)

 - Using the WPS and WiFi buttons to activate/deactivate the configmode WiFi. This way let's say a push on each of those Button activates the configmode WiFi for 5 minutes. This is not the case if the node has never been configured. If it's a first time flashed node, the WiFi is enabled continously.

I know the first point might spark some controversy but IMHO. we will be desperately in need of configmode-WiFi in a year or so.

rubo77:
For first time flashing it is not important to have a secure password. They will know pre-shared-key will be enough.
We should ask for a change of this password on first time use though

blocktrron:
I think this does not have the desired effect. The Config-Mode should be accessible to everyone who has physical access to the device itself, just like it is via Ethernet.

CodeFetch:
@NeoRaider That's not the company you are working for, is it? :grin: https://www.youtube.com/watch?v=aqHauk3bNFA
Actually I like the idea of using LEDs, but I think we should not depend on a phone app and the pass phrase does not need to be that secure as it is temporary. A four digit number should be sufficient. 1234 - blink 1 time, pause, blink 2 times, pause, blink 3 times, pause, blink 4 times, long pause, repeat. If the timing is chosen well, it should be no problem to use it. The "WiFi config mode" should be turned off if nobody associates within 5 Minutes or after a 10 Minute timeout of the client.

blocktrron:
I've tinkered a bit with your branch @belzebub40k . I've modified the branch to use a randomly generated BSSID with the node-id / primary mac as a password.

However, this approach will fail on ramips based devices as they only support vAP addresses which are derived from the radios MAC. My approach fails there.

So we are either back to using a phone app which reads an 8 digit PSK via LED or we are using a different approach and try to read the factory-psk, hoping it is stored in flash somewhere. Exclusively using the Node-ID would fail either way, as some AVM models at least do not have a MAC address printed on the casing.

What do you think?

CodeFetch:
@blocktrron 
> However, this approach will fail on ramips based devices as they only support vAP addresses which are derived from the radios MAC. My approach fails there.

We should prefer WPS.

> So we are either back to using a phone app which reads an 8 digit PSK via LED or we are using a different approach and try to read the factory-psk, hoping it is stored in flash somewhere. Exclusively using the Node-ID would fail either way, as some AVM models at least do not have a MAC address printed on the casing.

8 digit PSK can be computed within < 30s using a GPU if you intercept the handshaking. WPA2 is a dinosaur with fixed PBKDF2 using ridiculous 4096 iterations.
At least on Atheros devices only a random WPS pin is stored in the partition where the MAC is stored.
We should not use that, but use WPS like in it's security recommendations. That means generating a random WPS pin every time and only activating WPS when a button has been pressed. The PSK and the SSID should be randomly computed on every boot.

> I dislike the idea of the user setting a pin for the config-mode WiFi as most users will tend to forget it and thus the config-mode WiFi will not have the desired effect of being usable with just the physical device present.

Yes.
I tried to find good values for a blinking pattern and it is quite useful. I think you don't need a phone app for that. Only a pause between the digits long enough to type it. I'm going to make a pull-request if you give me some time (super busy atm). The problem I see is that we have to patch the hostapd-mini-package configuration to support WPS as it will not be accepted upstream (with WPS it's not minimal).

blocktrron:
@CodeFetch 

I strongly dislike the idea of relying on WPS as the only manner of connecting to the network, even though it might seem to solve our problem of being the same on every platform, as WPS implementations on the client side are not always existent. As an additional method however PB-WPS might be an option.

> 8 digit PSK can be computed within < 30s using a GPU if you intercept the handshaking.

Honestly, i don't see the problem. I dislike the idea of having a completely-open VAP for even a minute as this would open up a completely unsecured bridge to the WAN. A 8 digit PSK would be good enough in my opinion as it would mitigate the "activated by accident case" with no handshakes taking place enough.

CodeFetch:
@blocktrron With an 8 digit PSK you can just leave the network open to not give the user a false feeling of security, because it is not secure. Every auto-wifi-crack-tool can crack it if it intercepts the traffic within 3 minutes.

Please show me a modern device that does not support WPS.

blocktrron:
@CodeFetch 

> With an 8 digit PSK you can just leave the network open to not give the user a false feeling of security, because it is not secure.

My intent is to mitigate the "button pressed by accident". I think someone waiting for a handshake to a Freifunk Node in Setup-Mode via WiFi which is only active e.g. for 5 minutes is a stretch.

> Please show me a modern device that does not support WPS.

PBC is not supported in many NetworkManager GUIs on Linux, afaik even Ubuntu. Someone inexperienced with the command line might not be able to connect  to the Network.

Android P also removed WPS support.

CodeFetch:
@blocktrron Seems you're right. I didn't know that WPS support is so bad on linux network managers. I already knew that Google sucks. That's a shame. When you say PBC is not supported did you mean, that PIN mode is also not supported? My suggestion was not to use PBC, but just to start WPS when you push the button on the router and signal the PIN code using the LEDs. 
Android P removed WPS support, because Google is incompetent. When WPS is properly implemented, it is much more secure than weak passwords.
The only security issues with WPS are:

- Using PBC
- Using a weak random number generator for the salt on the router
- Using WPS PINs that can be derived from the MAC address
- Not using a new WPS PIN on every connection attempt
- Client devices accepting a connection to a WPS router after they have been tricked into revealing the first 4 digits of the PIN (and I assume Android devices were so stupid doing that and Google didn't assigned priority fixing this, but wanted to release their shitty brand-new Android), because to an attacker they may just have revealed the last 4 digits...

If we want to signal the PSK using the LEDs, we really need a light communication app and do not choose a weak or fixed password. We could then always signal the non-human-readable PSK in config mode and after you push the WPS or in case it is missing the Reset button it starts WPS and shows the WPS PIN instead.

blocktrron:
I did some further research. It seems the only real platform that ever supported WPS is Windows and Android up until Version 9. Neither iOS nor Mac OS X ever supported WPS.

So we definitely need to use a PSK.

CodeFetch:
@blocktrron With 16 digits between 1-9 it would take at most 40 days to crack the password using a SP35 Yukon Power. That's should be secure enough for a temporary password for the next 10 years. With 8 digits it's less than 0.1 seconds. I'd really love to see this feature, but I don't have a clue how to accomplish this securely without taking the MAC address into account or using an app with light communication.

christf:
I like the phone-app-approach that reads a PSK. If we cannot build that in a reliable way, the next best idea imho is using a well-known PSK and append a digit that is flashed via LED and interpreted and entered by the user.

CodeFetch:
@christf The problem with that approach is that the PBKDF2 parameters of WPA2 are so weak (fixed 4096 rounds) that likely services will come up soon (if there aren't any yet) to crack them if you intercept the salt (and these services will be happy about known plain-text in PSKs). There are people that do war-driving as a hobby. If it becomes that easy to do the cracking automatically, I don't want to be the one who suggested this approach. WPS is thought to be used for just entering a few digit PINs, but PSK is not thought to use a weak key. It is negligent that routers allow you to enter an arbitrary PSK and don't hash it with a random value and tell you "use that password! I won't help you making your network insecure", but vendors decided the customer is king and if they want to be hacked, let them or our competitor will do it and their no-brain customers will say to our customer "What is wrong with you password, dude? My WiFi password is just 'banana123'."  "My router forced me to use that password"  "Seems like your router sucks. You should better buy such a cool one like I have".

Another approach could be to use a simple PIN, but hash it computation intensive to generate a good PSK. We could even offer that as a website where you can enter the PIN your router shows and hash it using JavaScript, but I dislike that kind of centralization.

mweinelt:
If WPA2 is the issue that means we need to rely on WPA3 and SAE where attacks can only be mounted online, against the running node. WPA3 would also give us a good protection even when weak passwords are used.

But that obviously relies on the availability of capable client devices.

rotanid:
i think you are over-engineering this...

CodeFetch:
@rotanid You're right. That was considering worst-case scenarios. 

I think it's okay with a **random** PSK-prefix set in the site.conf, low transmit power and at least 4 digits from 1 to 9 signaled with the LEDs. That needs a skilled hacker (who knows gluon and the communities' prefix), a very well-designed cracking service (that recognizes it as a gluon router, knows the location and its communities' prefix), or a strong AI (that will likely know other ways to crack it and has better things to do) to attack this successfully. Until the last two are reality, there are no devices with WPA2 anymore and if a hacker wastes his or her time for something like that, the target probably knows who it was :see_no_evil:.
Still I think this should be activated using a button and if the device does not have such a button this feature should not be used at all.

@belzebub40k Here is a shell script for the blinking pattern that I wanted to use for the WPS PIN. It might help you and has been tested. Note that urandom gives enough entropy for such a setup, but use random if you like...
The reason why it's a shell script and not Lua is that the LEDs are managed using shell scripts.
https://paste.ee/p/LKlXv

mweinelt:
> I would deactivate the WLAN after 5 minutes no matter whether a client is still connected. If someone needs longer, he or she needs to press the button again and get a new password.

Actually if we limit the AP to a single station we could leave the AP up as long as a station is connected.

CodeFetch:
The deauth sent from the AP to the STA is unencrypted (WLAN deauthentication attack). Is it different the other way round?

CodeFetch:
By the way there is [a service for WPA2 cracking](https://gpuhash.me/) already and I confirmed it works quite well.

Ann0n:
Perhaps it should only be required to prove physical access while in config mode.
So the process is something like:
1) enter config mode
2) lan is active
3) if no lan connection is detected wifi is activated and always has a known predetermined configuration
4) you make the changes you want and choose save
5) you get show a new page with a configuration summary, 
on this page you are asked to press and hold a specific button on the router
6) you get a countdown shown, and when it gets to 0, you are to release the button, doing so within a second or two then allows writing the data

CodeFetch:
@Ann0n Where is the security here? We've found out that security is only possible by using a trusted third-party provider.

The router creates a open WLAN network where only a single web page is available and a WPA-encrypted network where one has access to the config mode and where only one false login attempt is allowed.
On the web page of the open network there is written a seemingly random string which only changes on reboots. The router's status LED is signalling a specific number with a blinking pattern. The user enters the seemingly random string and the number into e.g. a form on a trusted website or e.g. a Python script where the seemingly random string is hashed with the number -> This is the WPA key the user can use to login to the config-mode WLAN.

Ann0n:
I really agree with 

> i think you are over-engineering this...

If you can enter the config mode at any time by pressing the buttons, physical access gives you the keys to the kingdom.
So basically you can make the whole wifi thing do whatever you want, but it won't really matter, since at any later time someone can change anything they want, if they have physical access.

I am not trying to say your idea is bad, but I think it is overly complicated considering how easy it is to get into config mode.

I left out a step in my process above. In between choosing save and the summary page, the configuration would need to be locked. This way you would always see and be able to review what is setup before it being committed. That would be the security, that you say what is being committed, when you commit it. At any point after that, anyone with physical access could change anything anyway.

CodeFetch:
> > i think you are over-engineering this...

No, security is not "over-engineering". Today's security requirements are and need to be much higher, because you always need to keep in mind, that we are near a singularity of mind where such security flaws can be exploited automatically. I know it sounds like science-fiction, but think of a superhuman AI that wants to hack everything and control everything. It would hack your router. Yes, there is no such AI at the moment (or maybe you just don't know, because it controls everything already ;-)), but it's only a matter of time. Thus the first thing you learn studying IT security is: Security by obscurity gives no security at all and giving the user a false feeling of security is the worst thing one can do.

> This way you would always see and be able to review what is setup before it being committed.

Output can be faked once the system is corrupted or using a MITM attack.

> At any point after that, anyone with physical access could change anything anyway.

When a system is corrupted, you can't trust anything you see anymore. The only thing you can do to recover the system into a trustful state again is to rewrite the flash chip with an SPI flasher.

CodeFetch:
I just remember the end of our discussion in our community. Only using a hash-like approach for third-party-trust is not enough.

The router needs to have a pre-defined public key of a authentication service run by the community.

That means the router encrypts the WPA keyphrase together with the number the LED shows using the public key of the community's tool. The user needs to enter the resulting chiffre in the web-interface of the community's tool together with the number the LED shows. The tool decrypts the chiffre using its private key and checks whether the number is correct and if it matches, it shows the WPA keyphrase to the user who can then login to the encrypted config-mode WLAN network. If it doesn't match, it adds the chiffre to a blacklist and does not accept further checks with that. 

So the chances for an attacker to guess the correct number are 1/x where x is the range of the possible numbers shown by the LED. This is the most secure approach one can get, but of course introduces a single point of failure.

Ann0n:
> Thus the first thing you learn studying IT security is: Security by obscurity gives no security at all and giving the user a false feeling of security is the worst thing one can do.

Exactly. That is why putting some kind of encryption on freifunk WIFI doesn't make sense, as an example.

To me is seems like what you want to do is exactly that, "Security by obscurity".

I say as long as someone can connect a cable and use the button for accessing config mode, nothing will be more secure than that button. Or am I missing some magical access control made by some kind of fantasy AI from another planet? I think 99% of your https://github.com/freifunk-gluon/gluon/pull/1563#issuecomment-522708386 is not required. If you don't share my view of the router being insecure with the current config mode setup, then just say so. You don't need to share your skynet fantasies on github. ;)

I posted in this issue, because I looked at the software and noticed the button press config mode is the most insecure point. It took me about a whole 3 hours of looking at the software to do so.

If someone really wants to gain control over a router you have, I doubt they will sit around waiting for you to put it in config mode, and then manipulate it. What role should Gluon fill? I assumed it should be an easy way for people to join into the freifunk or another mesh community. Reading a blinking LED seems to me to be something which could easily be problematic and annoy people. Getting a phone app, which might not work for whatever reason to do, sounds even worse.

I read through your last post, and your plans would make the whole thing dependent on a web service. This is something many people complain about companies doing.

Either way, it seems to me that you and your community pretty much already made up its mind about what needs to be done. To bad that information was not available here to begin with.

CodeFetch:
> If you don't share my view of the router being insecure with the current config mode setup, then just say so. You don't need to share your skynet fantasies on github. ;)

Requiring physical access is the best security one can get. If you have an encrypted hard-drive and someone nasty wants access to it, he or she will likely torture or blackmail you, but it's still the best security you can have. These are no skynet fantasies. It will be reality in our lifetime and people who don't realize that are doomed.

> I read through your last post, and your plans would make the whole thing dependent on a web service. This is something many people complain about companies doing.

You already have to trust the infrastructure of the community you use when you don't run your own infrastructure. There is no problem here, because you could build your own firmware with your own keys and e.g. use an offline-tool.

CodeFetch:
> I say as long as someone can connect a cable and use the button for accessing config mode, nothing will be more secure than that button.

Yes, if you enter the configuration as morse code.

MPW1412:
> So we are either back to using a phone app which reads an 8 digit PSK

I read that a couple of times. Almost all current notebooks have a camera build in. If we morse an alphanumeric ten digit code, someone who can read morse, can get it without recognition engine. And if you don't want to learn morse, you could use a JavaScript based recognition web-app, which could be run on a smartphone or notebook.

And just in case the recognition fails, you could provide the morse alphabet in the app to decode it yourself.

CodeFetch:
As there is a similar discussion in an [OpenWrt pull-request](https://github.com/openwrt/openwrt/pull/2408) and people there tend to use a simple/not that secure approach, we should follow this as a complex setup with a third-party provider is not in sight.
Thus a method needs to be implemented in OpenWrt to retrieve the vendor's PSK or otherwise the label MAC address needs to be used as PSK and the SSID/BSSID needs to be set randomly at initial device configuration.

christf:
We have not discussed neoraiders second suggestion yet. For this, the speed at which DHCP is replied will define whether communication is possible, isn't it? 
I like the idea of being able to configure those nodes without re-plugging them afterwards and without the need to fuss with LAN cables. @neoraider could you detail how the address distribution would work with DHCP?

For IPv6 and uradvd I think this could be easier but that would be a slight variation of your idea.

mgiganto:
> Thus a method needs to be implemented in OpenWrt to retrieve the vendor's PSK or otherwise the label MAC address needs to be used as PSK and the SSID/BSSID needs to be set randomly at initial device configuration.

This method is implemented in my [PR 2445](https://github.com/openwrt/openwrt/pull/2445), this PR is to configure/recover some oem values and includes the PSK, SSID, and country for devolo magic 2 wifi device.

CodeFetch:
@MPW1412 Hm... Can you test this using my script? It's quite accurate. The problem I see is when e.g. the node is on a pole. The app will have a hard time reading it. On the other hand cameras and morse apps are getting better every day... Brilliant!

CodeFetch:
I've written a bash script (works nice with bash, but might need some changes for busybox ash) which generates a random PSK and signals it using the status LED in morse code. It would be nice if someone could test it and bring it in a state where it runs with ash, because I don't have time for it atm. I've tested it successfully with the android app `Morse Code Agent`. 

```
#!/bin/sh

# Duration in minutes to keep the WLAN active
run_max_minutes=5

# Upper bound of PSK length
psk_len_upper=30

# Lower bound of PSK length
psk_len_lower=20

# Duration of a morse dot (in seconds)
# Also used as pause between symbols
# 100 milliseconds works good with +30 fps
# morse decoders (e.g. a mobile phone app)
dot_duration="0.1"

# End of configuration 
#######################

. /lib/functions/leds.sh
. /etc/diag.sh

# busybox sleep does not support floating point - workaround uses lua
function func_sleep {
	eval "lua -e 'local c=os.clock function s(n) local t0=c() while c()-t0 <= n do end end s($1)'"
}

time_stop=$(( $run_max_minutes * 60 ))
time_start=0
time_cmd='echo $(($(cut -d " " -f 1 < /proc/uptime | cut -d "." -f1) - $time_start))'

# Morse dash's duration is defined as 3 times morse dot's
# Also used as pause between characters
dash_duration=$(expr 3*$dot_duration | bc)

# Difference between dot and dash (3 times)
dotdash_duration=$(expr 2*$dot_duration | bc)

# Difference between dash (3) and words (7)
dashword_duration=$(expr 4*$dot_duration | bc)

# Implemented morse alphabet
morse_alphabet="A .- B -... C -.-. D -.. E . F ..-. G --. H .... I .. J .--- K -.- L .-.. M -- N -. O --- P .--. Q --.- R .-. S ... T - U ..- V ...- W .-- X -..- Y -.-- Z --.. 1 .---- 2 ..--- 3 ...-- 4 ....- 5 ..... 6 -.... 7 --... 8 ---.. 9 ----. 0 -----"

# Range of the PSK length
psk_len_range=$(($psk_len_upper - $psk_len_lower))

# Get a random number between 0 and 99
psk_len_rnd=$(</dev/urandom tr -dc 0-9 | head -c 2 | sed 's/^0//')

# Calculate random PSK length using the random number
psk_len=$(($(($(($psk_len_range * $psk_len_rnd)) / 99)) + psk_len_lower))

# Get a random PSK consisting of alphanumeric characters
psk=$(</dev/urandom tr -dc 0-9A-Z | head -c $psk_len)
echo "WPA PSK: $psk"

##TODO Setup WLAN

# Convert PSK to morse code
psk_chars=$(echo "$psk" | sed -e 's/\(.\)/\1 /g')
morse_psk_chars=""
for psk_char in $psk_chars; do
	last_morse_char=""
	for char in $morse_alphabet; do
		if [ "$last_char" = "$psk_char" ]; then
			morse_psk_chars="$morse_psk_chars $char"
			break;
		fi
		last_char=$char
	done
done
echo "Morse code: $morse_psk_chars"

# Display the morse code with timeout
time_start=$(eval $time_cmd)
led_off $status_led
func_sleep $dash_duration
func_sleep $dashword_duration
while [ $(eval $time_cmd) -lt $time_stop ]; do
	for morse_char in $morse_psk_chars; do
		morse_char_symbols=$(echo "$morse_char" | sed -e 's/\(.\)/\1 /g')
		for symbol in $morse_char_symbols; do
			if [ "$symbol" = "." ]; then
				led_on $status_led
				func_sleep $dot_duration
				led_off $status_led
			fi
			if [ "$symbol" = "-" ]; then
				led_on $status_led
				func_sleep $dash_duration
				led_off $status_led
			fi
			func_sleep $dot_duration
		done
		func_sleep $dotdash_duration
	done
	func_sleep $dashword_duration
done

##TODO Tear down WLAN
status_led_set_timer 1000 300

```

MPW1412:
@CodeFetch, did you find a recognition engine yet? I couldn't either find one for Android nor for Linux.

CodeFetch:
@MPW1412 
> I've tested it successfully with the android app `Morse Code Agent`.

mweinelt:
Hey @belzebub40k ,

we decided that this changeset is a good candidate for the community repository that is about to be created. As this has an obvious security impact we cannot support adding this to the Gluon core project at this time.

CodeFetch:
I'm pretty sure we'll see a solution in https://github.com/openwrt/openwrt/pull/2408 and don't need a community package for this.

CodeFetch:
> We have not discussed neoraiders second suggestion yet. For this, the speed at which DHCP is replied will define whether communication is possible, isn't it?
> I like the idea of being able to configure those nodes without re-plugging them afterwards and without the need to fuss with LAN cables. @NeoRaider could you detail how the address distribution would work with DHCP?

I agree this would be great.

mweinelt:
> Hey @belzebub40k ,
> 
> we decided that this changeset is a good candidate for the community repository that is about to be created. As this has an obvious security impact we cannot support adding this to the Gluon core project at this time.

I still think an easy setup possibility on mobile devices is an important step for Gluon, sadly with its security drawbacks we won't accept it in the core gluon repository.

Please retarget this pull request over in the [community-packages](https://github.com/freifunk-gluon/community-packages) repository, where this change would be very welcome.

PR #1568, opened 2018-11-15
Local node routing table separation

T-X:
This pull request includes one minor fix for the source MAC for IPv6 packets from the local-node ULA. And two changes similar to the previous attempt from commit b3762fc61cca ("gluon-client-bridge: move IPv4 local subnet route to br-client (#1312)"). But without the drawbacks @ecsv had discovered with it (3ef28a4684).

Conceptually, the new approach is to use a separate routing table for the local-node interface. Which we will only jump into if a node tries to use its IPv4 or IPv6 local-node address. This should ensure, that we never use the local-node interface's MAC address for any purpose other than direct client-to-node communication (for instance, routed packets will always use the unique MAC address from br-client).

This is also in preparation for the gluon-alt-esc packages (#1094).

### DIFF
--- a/package/gluon-client-bridge/luasrc/lib/gluon/upgrade/310-gluon-client-bridge-local-node
+++ b/package/gluon-client-bridge/luasrc/lib/gluon/upgrade/310-gluon-client-bridge-local-node
@@ -23,8 +23,7 @@ uci:section('network', 'device', 'local_node_dev', {
 local ip4, ip6
 
 if next_node.ip4 then
-	local plen = site.prefix4():match('/%d+$')
-	ip4 = next_node.ip4 .. plen
+	ip4 = next_node.ip4 .. '/32'
 end
 
 if next_node.ip6 then
--- a/package/gluon-mesh-batman-adv/luasrc/lib/gluon/upgrade/320-gluon-mesh-batman-adv-client-bridge
+++ b/package/gluon-mesh-batman-adv/luasrc/lib/gluon/upgrade/320-gluon-mesh-batman-adv-client-bridge
@@ -25,7 +25,18 @@ uci:section('network', 'interface', 'client', {
 
 uci:delete('network', 'client_lan')
 
+uci:section('network', 'rule', 'local_node_rule', {
+	src = next_node.ip4 .. '/32',
+	lookup = 2,
+})
+uci:set('network', 'local_node_rule', 'in', 'loopback')
+
 uci:delete('network', 'local_node_route')
+uci:section('network', 'route', 'local_node_route', {
+	interface = 'local_node',
+	target = site.prefix4(),
+	table = 2,
+})
 
 uci:section('network', 'rule6', 'local_node_rule6', {
 	src = next_node.ip6 .. '/128',
--- a/package/gluon-mesh-batman-adv/luasrc/lib/gluon/upgrade/320-gluon-mesh-batman-adv-client-bridge
+++ b/package/gluon-mesh-batman-adv/luasrc/lib/gluon/upgrade/320-gluon-mesh-batman-adv-client-bridge
@@ -8,6 +8,7 @@ local site = require 'gluon.site'
 local sysconfig = require 'gluon.sysconfig'
 local util = require 'gluon.util'
 local uci = require('simple-uci').cursor()
+local next_node = site.next_node({})
 
 
 uci:section('network', 'interface', 'client', {
@@ -26,11 +27,17 @@ uci:delete('network', 'client_lan')
 
 uci:delete('network', 'local_node_route')
 
+uci:section('network', 'rule6', 'local_node_rule6', {
+	src = next_node.ip6 .. '/128',
+	lookup = 2,
+})
+uci:set('network', 'local_node_rule6', 'in', 'loopback')
+
 uci:delete('network', 'local_node_route6')
 uci:section('network', 'route6', 'local_node_route6', {
-	interface = 'client',
+	interface = 'local_node',
 	target = site.prefix6(),
-	gateway = '::',
+	table = 2,
 })
 
 uci:save('network')
--- a/package/gluon-mesh-batman-adv/luasrc/lib/gluon/upgrade/320-gluon-mesh-batman-adv-client-bridge
+++ b/package/gluon-mesh-batman-adv/luasrc/lib/gluon/upgrade/320-gluon-mesh-batman-adv-client-bridge
@@ -25,6 +25,11 @@ uci:section('network', 'interface', 'client', {
 
 uci:delete('network', 'client_lan')
 
+uci:section('network', 'route', 'br_client_route', {
+	interface = 'client',
+	target = site.prefix4(),
+})
+
 uci:section('network', 'rule', 'local_node_rule', {
 	src = next_node.ip4 .. '/32',
 	lookup = 2,


christf:
I am uncertain whether we should we diverge routing for babeld and batman.

T-X:
Hi @christf. When making this PR I was wondering and uncertain why those routes were in the batman-adv package upgrade scripts, too.

Do you mean you see some conflict with the babel integration? Then please elaborate, then I'd love to hear and learn what you are doing and need differently. Or are you just suggesting to do what I had been wondering about initially, to move these routes and tables from gluon-mesh-batman-adv (package/gluon-mesh-batman-adv/luasrc/lib/gluon/upgrade/320-gluon-mesh-batman-adv-client-bridge) to gluon-client-bridge (package/gluon-client-bridge/luasrc/lib/gluon/upgrade/310-gluon-client-bridge-local-node)?

T-X:
Ok, great.

"local_node_route6" was already set up via 320-gluon-mesh-batman-adv-client-bridge. @christf, can I just move my changes and additions for "local_node_route", "local_node_route6", "local_node_rule", "local_node_rule6" and "br_client_route" from 320-gluon-mesh-batman-adv-client-bridge to 310-gluon-client-bridge-local-node? Or would that break something for Babel?

christf:
I do not think it should break anything but it would be interesting to see a test.

rotanid:
how's the testing of this coming along? it would be nicer to get good work not only done, but also tested and afterwards maybe merged because of good results :)

CodeFetch:
@christf This should not interfere with the Babel setup as the local node route table lookup is /128 /32. If it does it should be easy to fix.
@rotanid Please merge it so that https://github.com/freifunk-gluon/gluon/pull/1094 can be merged.

rotanid:
@CodeFetch i won't merge if there's not even basic testing with a babel-build.

bobcanthelpyou:
I try to do a test build of this PR with the help of the FFFFM babel knowledge, next week(-end)
https://wiki.ffm.freifunk.net/firmware:babel
https://github.com/freifunk-ffm/site-ffffm/tree/christf_next

@rotanid what comes to your mind when basic testing has to be done?

And what should be tested when doing tests with this PR especially?

CodeFetch:
On layer 3 this patch should work this way:
1. (local_node_rule) IPv4 packet from our local node IP address (indirectly means client is local)? -> put in routing table 2
1. (local_node_rule6) IPv6 packet from our local node IP address (indirectly means client is local)? -> put in routing table 2
2. (babel) packet destined to a client on another node (shorter cidr)? -> output on interface where other node is / forward to other node
3. (br_client_route) IPv4 packet destined to a client in the site-prefix? -> output on client bridge
(this is needed as T_X does not use the site-ipv4-prefix for the local-node interface anymore, but the local-node IPv4 address. IPv6 routes use the link local address for their routes which is based on the MAC address of the clients together with the interface where the MAC address can be reached. Thus this is not needed for IPv6 here)

Routing table 2:
(local_node_route) IPv4 packet destined to site-IPv4-prefix? -> output on local-node interface
(local_node_route) IPv6 packet destined to site-IPv4-prefix? -> output on local-node interface

local-node interface is a veth and the packet outputs on local-port which is in the client bridge together with the wifi client interfaces and the bat0 client interface. This patch only has the purpose that packets from the local node IP address use the MAC address of the local-port interface.

@bobcanthelpyou That means you have to check everything ordinary like does the clients still have gateway reachability, do they receive router announcements - say check if the clients have internet connection. Then as a client check if you can reach the status page at the local node's IPv4/IPv6 address and as a client on another node check if you can reach the status page of this node with its global IPv6 address.

rotanid:
@bobcanthelpyou @T-X @christf has someone now tested this with a babel build?

bobcanthelpyou:
It is still on my todo-list but i forgot about it, and can't start at the moment (cause i am afk for the next weeks)

rotanid:
> @bobcanthelpyou @T-X @christf has someone now tested this with a babel build?

status?

package/gluon-client-bridge/luasrc/lib/gluon/upgrade/310-gluon-client-bridge-local-node
```diff
@@ -23,8 +23,7 @@ uci:section('network', 'device', 'local_node_dev', {
 local ip4, ip6
 
 if next_node.ip4 then
-	local plen = site.prefix4():match('/%d+$')
-	ip4 = next_node.ip4 .. plen
+	ip4 = next_node.ip4 .. '/32'
```
mweinelt: @christf Please test this change with the babeld setup.

package/gluon-mesh-batman-adv/luasrc/lib/gluon/upgrade/320-gluon-mesh-batman-adv-client-bridge
```diff
@@ -24,13 +25,35 @@ uci:section('network', 'interface', 'client', {
...
+uci:section('network', 'route', 'br_client_route', {
+	interface = 'client',
+	target = site.prefix4(),
+})
+
+uci:section('network', 'rule', 'local_node_rule', {
+	src = next_node.ip4 .. '/32',
+	lookup = 2,
+})
+uci:set('network', 'local_node_rule', 'in', 'loopback')
```
NeoRaider: Please merge this *set* into the *section* above (same for `local_node_rule6` below).

I assume this was not done because `in` is a keyword in Lua. The correct syntax for this (and arbitrary Lua values as table keys) is
```lua
{
  ...
  ['in'] = 'loopback',
},
```

NeoRaider: ... and is the `in` setting even necessary at all?

package/gluon-mesh-batman-adv/luasrc/lib/gluon/upgrade/320-gluon-mesh-batman-adv-client-bridge
```diff
@@ -24,13 +25,35 @@ uci:section('network', 'interface', 'client', {
 
 uci:delete('network', 'client_lan')
 
+uci:section('network', 'route', 'br_client_route', {
+	interface = 'client',
+	target = site.prefix4(),
+})
```
NeoRaider: Hmm, I wonder if such a change will lead to incorrect source address selection in some cases. I'm not sure if source address selection for hosts with multiple addresses is as well-specified for IPv4 as it is for IPv6 (at least, I don't know the relevant RFCs), but I'm thinking about the following:

When the Gluon node wants to contact a local client, it will now select this route at first (before source address selection), with interface client - which doesn't have any IPv4 address. Now, how is the source address selected now? What tells Linux to use the address of local-node and not, for example, the br-wan address?

The issue might be solved by adding a "source" attribute with the next-node address to this route definition, but I haven't fully thought this through yet.

For IPv6, a similar issue cannot occur, as br-client should always have an address, which would be selected (at least when gluon-radvd is included, which is the default now).

NeoRaider: @T-X

Hmm, I wonder if this route could lead to packets with the local-node source IPv4 address leaving through br-client, with the br-client MAC address, or if packets would go through another round of routing after choosing the source address, and then end up in routing table 2. If the former is the case, I don't think we should merge the IPv4 part of these changes...

Would adding a source address attribute with the local-node IPv4 address to this route be a problem for alt-esc?

rotanid:
@T-X are you still interested in this? you have not commented on the feedback which was given 9 months ago...

T-X:
Closing this PR as I'm currently not working on it. And instead have an alternative approach in mind (https://www.open-mesh.org/projects/open-mesh/wiki/OpenHarbors).

PR #1573, opened 2018-11-17
Do not count clients inactive for more than 60s 

### LINKED ISSUE #1135: inaccuracy in reported wifi client count
I was irritated that the client count for my DIR860L was 0 on hopglass, but 8 on meshviewer, so I started looking into the respondd data that the node delivers.

And sure enough `wifi` doesn't show any clients at all.

```
"clients": {
  "total": 0,
  "wifi": 0,
  "wifi24": 8,
  "wifi5": 0
},
```

This is on 88f2e3e4e2de3666705cc4bc4198c728e7fa4ebe

NeoRaider:
Only compile-tested so far, I won't be able to run-test it before tomorrow.

Fixes #1135.

### DIFF
--- a/package/gluon-mesh-babel/src/respondd.c
+++ b/package/gluon-mesh-babel/src/respondd.c
@@ -63,6 +63,8 @@
 #define STRINGIFY(s) _STRINGIFY(s)
 #include <stdlib.h>
 
+#define MAX_INACTIVITY 60000
+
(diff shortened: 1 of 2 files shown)


NeoRaider:
Tested and working as expected. I've fixed the commit message of the second commit to match my changes: Both clients and neighbours are filtered by their inactivity time now.

PR #1577, opened 2018-11-21
Wireguard prep

christf:
This contains patches that prepare gluon for a wireguard build, separating hopefully uncontroversial changes from wireguard which induces quite a bit of complexity.


package/gluon-iptables-clamp-mss-to-pmtu/files/lib/gluon/mesh-vpn/iptables-mss.rules
```diff
@@ -1,3 +1,3 @@
 *mangle
--A FORWARD -o mesh-vpn -p tcp -m tcp --tcp-flags SYN,RST SYN -d 64:ff9b::/96 -j TCPMSS --clamp-mss-to-pmtu
```
rotanid: is there an explanation on the removal of this destination part? maybe i'm just missing it.

rotanid:
@christf please change the "bump modules" commit so it looks like the others, including the shortlog etc.

package/gluon-iptables-clamp-mss-to-pmtu/files/lib/gluon/mesh-vpn/iptables-mss.rules
christf: clamping should be done on all traffic passing through vpn interfaces, not just on traffic destined for nat64. When introducing the package I wanted to keep the scope of the rule as small as possible. During usage I had to learn & accept that the initial scope was too small, thus removing the destination.

christf:
@rotanid thank you for your review. Done.

package/gluon-mesh-babel/files/etc/sysctl.d/40-gluon-babel.conf
```diff
@@ -0,0 +1,2 @@
+# this is used for ecn in ipv6 and ipv4 networks
+net.ipv4.tcp_ecn=1
```
NeoRaider: This does not seem like it needs to be specific to Babel, maybe move this to gluon-core?

PR #1582, opened 2018-11-25
docs: cherry-pick from master of all relevant changes for 2018.1.x

rubo77:
Since https://gluon.readthedocs.io is always built on this branch, I think we should backport all fitting docs changes here too


NeoRaider:
The time command is not needed by Gluon v2018.1.x.

rubo77:
The time commit is removed and the new batman_V commit added

PR #1588, opened 2018-11-26
ar71xx-generic: Add support for CPE210 V3

kpanic23:
Since upstream support seems to be "problematic", why don't we just add support via a patch until the device is finally supported upstream? 

~I've tested everything but upgrading.~

  - [X] webinterface
  - [X] tftp
  - [ ] other: <specify>
- wired network
- wifi (if applicable)
- led mapping
  - power/sys led
  - radio leds
    - [X] should show activity
  - switchport leds
    - [X] should map to their respective port (or switch, if only one led present) 
- [X] primary mac should match address on device label (or packaging) (https://gluon.readthedocs.io/en/latest/dev/hardware.html#notes)

### DIFF
--- a/patches/openwrt/0012-firmware-utils-tplink-safeloader-Add-CPE210-v3.patch
+++ b/patches/openwrt/0012-firmware-utils-tplink-safeloader-Add-CPE210-v3.patch
@@ -1,0 +1,34 @@
+From: Robert Marko <robimarko@gmail.com>
+Date: Sat, 11 Aug 2018 17:12:01 +0200
+Subject: firmware-utils: tplink-safeloader: Add CPE210 v3
+
+Add TP-Link CPE210 v3 to the support list.
+Its identical to the v2.
+
+Signed-off-by: Robert Marko <robimarko@gmail.com>
+diff --git a/tools/firmware-utils/src/tplink-safeloader.c b/tools/firmware-utils/src/tplink-safeloader.c
+index b2cc96bc3ce765df62bee837bdf0e3cf2d53be1b..8164c1ce314c5c658f2fe071c648059e7b4e1caf 100644
+--- a/tools/firmware-utils/src/tplink-safeloader.c
++++ b/tools/firmware-utils/src/tplink-safeloader.c
+@@ -155,7 +155,7 @@ static struct device_info boards[] = {
+ 		.last_sysupgrade_partition = "support-list",
+ 	},
+ 
+-	/** Firmware layout for the CPE210 V2 */
++	/** Firmware layout for the CPE210 V2 and V3 */
+ 	{
+ 		.id     = "CPE210V2",
+ 		.vendor = "CPE210(TP-LINK|UN|N300-2|00000000):2.0\r\n",
+@@ -170,7 +170,11 @@ static struct device_info boards[] = {
+ 			"CPE210(TP-LINK|US|N300-2|55530000):2.0\r\n"
+ 			"CPE210(TP-LINK|UN|N300-2):2.0\r\n"
+ 			"CPE210(TP-LINK|EU|N300-2):2.0\r\n"
+-			"CPE210(TP-LINK|US|N300-2):2.0\r\n",
++			"CPE210(TP-LINK|US|N300-2):2.0\r\n"
++			"CPE210(TP-LINK|EU|N300-2|45550000):3.0\r\n"
++			"CPE210(TP-LINK|UN|N300-2|00000000):3.0\r\n"
++			"CPE210(TP-LINK|UN|N300-2):3.0\r\n"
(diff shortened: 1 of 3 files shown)


kpanic23:
```
root@CPE210v30:~# autoupdater -f
Retrieving manifest from http://upd4.ff3l/experimental/sysupgrade/experimental.manifest ...
Stopping cron...
Stopping haveged...
Stopping micrond...
Stopping sysntpd...
Stopping gluon-radvd...
Stopping uhttpd...
Stopping sse-multiplexd...
Stopping gluon-respondd...
vm.drop_caches = 3
Downloading image:  7620 / 7620 KiB
packet_write_wait: Connection to fdc7:3c9d:ff31:1:e80:63ff:fe07:e71e port 22: Broken pipe
```
Looks good :)

mweinelt:
A major issue always lies in the question of how long we would need to keep these patches downstream. With the upstream merge request being blocked because of different opinions - that doesn't sound very promising. And that is what prevents me from merging that without further discussion here and upstream.

rotanid:
@kpanic23 upstream doesnt look too bad i think, it doesnt depend on the openwrt guys atm but on the contributor

kpanic23:
The intention of this PR was that, based on my current experience with TP-Link, we should include support for this device as soon as possible, or rather, as long as it is available ;)
If we wait until it is merged upstream and our first Gluon version based on the next OpenWrt release is released, the device will be probably not in production anymore...

kpanic23:
Which is not meant to criticise the Gluon and/or OpenWRT-team, but rather the ever-shorteing product lifecycles of TP-Link devices ;)

mweinelt:
I agree that it's probably important to get it in before v2018.2, just so it still can be bought.

blocktrron:
This might be overruled, but i would feel MUCH better if we could get ath79 support going here before we merge the device to ar71xx downstream.

rotanid:
as there is no progress on the upstream Pull Request, we'll close this here.
feel free to re-open or open a new PR once it is merged upstream, maybe you can work on this instead of the original author.

adrianschmutzler:
I've taken over the effort to support this device on ar71xx.
I hope both ath79 (https://github.com/openwrt/openwrt/pull/1843) and ar71xx (https://github.com/adrianschmutzler/openwrt/commit/ef471a5057f8d3d68d0e31f39598fc5f95d1795a) support will be merged into master soon.

Support for master (and openwrt-19.xx) will include dynamic partitioning, which is not available in openwrt-18.06. Thus, backporting requires changes to the partitions defined for the device.
I'm currently testing backport patches to openwrt-18.06 for Freifunk Franken firmware based on openwrt-18.06.2. (Also fixing the cross-upgrade-posibility problem raised in the initial OpenWrt PR)

If you are still interested in patches for openwrt-18.06.2, I would provide those after successful testing.

rotanid:
@adrianschmutzler thanks for your work. as soon as you have tested openwrt-18.06 patches of the now-merged upstream ones, someone else can use them to test with gluon.
it must be possible to upgrade from earlier 18.06 based builds and later to upgrade to 19.x.
also, upgrading from early 18.06 based builds to 19.x shouldn't get broken if possible.

adrianschmutzler:
My rebase can be found here:
https://github.com/adrianschmutzler/openwrt/commits/cpev3gluon

I will not send a gluon PR, since I'm not familiar with gluon. However, everything should be similar to the CPE210v2, so this should be an easy one.

Since the support patch for the OpenWrt master is just a "copy" of the v2, I essentially did the same for openwrt-18.06. In target/linux/ar71xx/image/generic-tp-link.mk, where a lot happened in between, this might actually look different if you compare the two. Especially note that no dynamic partitions are available in openwrt-18.06 branch.

I tested sysupgrading from an 18.06.2-based build to a build with the most recent master, which went well went (both with Freifunk Franken-based builds, so no pure OpenWrt). Since partitions in 18.06.x are fixed, there should not be problems with earlier versions.
After 19.xx, when kernel 4.19 is coming, well have to see ... (maybe one has to switch to the OKLI loader then)

rotanid:
@kpanic23 maybe you can build based on his branch and test this. if it works, feel free to update your pull request!

kpanic23:
Sorry, I don't have a V3 to test anymore. But in the meantime, we have 8 V3s running flawlessly running a firmware based on my PR.

adrianschmutzler:
@kpanic23 The current PR will work equally well as one based on the new patches. The only difference is that the old PR is supposed to allow crossupgrading between CPE210v2 and CPE210v3.

kpanic23:
superseded by #1695

adrianschmutzler:
superseded by https://github.com/freifunk-gluon/gluon/pull/1696

PR #1589, opened 2018-11-27
gluon-mesh-batman-adv: Use own namespace for batman-adv-legacy

T-X:
This patchset allows loading and running the mainline batman-adv (compat 15) and the batman-adv-legacy (compat 14) fork simultaneously.

Which module to use is determined via the routing_algo parameter in the site configuration (which is copied to /etc/config/batman-adv on update): If set to BATMAN_IV_LEGACY then the batman-adv-legacy (compat 14) module is used. Otherwise (e.g. BATMAN_IV or BATMAN_V) the mainline batman-adv module is chosen.

Together with the gluon-scheduled-domain-switch package (freifunk-gluon/gluon#1555) this should ease the migration from compat 14 to compat 15 and speed-up the deprecation of batman-adv-legacy.

Note: This pull-request should only be applied after applying freifunk-gluon/batman-adv-legacy#13 and after applying freifunk-gluon/packages#206.


T-X:
Changelog v2:
* updated packages hash in modules file to the current merge result of freifunk-gluon/packages#206
* fixed gluon-ebtables-limit-arp calls to the correct batctl binary by setting and using PATH to use the batctl wrapper
* fixed alfred startup with a patch to let it automatically determine the netlink name to use ("batadv" vs. "batadv_legacy")

T-X:
Closing this in favor of #1622.

PR #1591, opened 2018-11-29
ar71xx: add support for UniFi-AC-Mesh-Pro

achterin:
This PR adds proper support/detection for the UniFi AC Mesh Pro devices. These devices are hardware clones of the UniFi AC Pro but with an outdoor casing.
This PR is compile-tested only because I don't have a device on hand anymore. I've used the original backport successfully on 5 devices back in july last year though.

I will update the checklist from time to time once testusers report back:
  - [ ] webinterface
  - [ ] tftp
  - [x] other: ssh into device, mtd on both partitions, dd nullbyte for bootselect
  - [x] must have working autoupdate [1]
- wired network
- wifi (if applicable)
- led mapping
  - power/sys led
  - radio leds
    - no radio ledss present
  - switchport leds
    - no switch leds

[1]
`root@C-Henrietten24TestMeshPro:~# lua -e 'print(require("platform_info").get_ima
ge_name())'
ubiquiti-unifi-ac-mesh-pro`

Signed-off-by: Christoph Krapp <achterin@googlemail.com>

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -223,6 +223,7 @@ ar71xx-generic
   - Rocket M2/M5 Ti
   - Rocket M2/M5 XW
   - UniFi AC Mesh [#ath10k]_
+  - UniFi AC Mesh Pro [#ath10k]_
   - UniFi AP
   - UniFi AP AC Lite [#ath10k]_
   - UniFi AP AC LR [#ath10k]_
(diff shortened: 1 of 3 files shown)


mweinelt:
The patch is included in OpenWrt master and could be backported to 18.06. When that happens we can drop the patch here, but until then I think this looks ready to merge.

rotanid:
@NeoRaider maybe you can cherry-pick upstream 987b961537b7002eda21df97dd8bfebe8882bc6d into 18.06 branch? this would save us a patch in Gluon :-)

Issue #864, opened 2016-08-28
Make ath10k firmware selectable in site.mk


NeoRaider:
Provide choice between AP+IBSS and AP+11s capable firmwares.

PR #1592, opened 2018-11-30
docs: fix some old urls or 'upgrade' to https

bobcanthelpyou:
* `wiki.openwrt.org` seems to redirecting to `https://oldwiki.archive.openwrt.org/` which is not usable, lately

* `https://git-scm.com/book/en/Git-Basics-Tagging` redirects to `https://git-scm.com/book/en/v2/Git-Basics-Tagging`
Q: not sure if we should keep the link or use version 2?

* Some links where upgraded to https
Q; Is that useful at all?

* Also noted some of the issues are linked as pull requests in the release notes or the other way round, e.g.
`https://github.com/freifunk-gluon/gluon/issues/1097` -> `https://github.com/freifunk-gluon/gluon/pull/1097` or
`https://github.com/freifunk-gluon/gluon/pull/864` -> `https://github.com/freifunk-gluon/gluon/issues/864`
Q: Should that be fixed?

### DIFF
--- a/docs/features/monitoring.rst
+++ b/docs/features/monitoring.rst
@@ -33,7 +33,7 @@ There are two packages responsible for distribution of the information. For
 one, information is distributed across the mesh using alfred_. Information
 between neighbouring nodes is exchanged using `gluon-respondd`.
 
-.. _alfred: http://www.open-mesh.org/projects/alfred
+.. _alfred: https://www.open-mesh.org/projects/alfred
 
 alfred (mesh bound)
 ~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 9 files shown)


NeoRaider:
- Using v2 of the book seems like a good idea, so the links don't break when chapters get renamed in a potential future version
- Upgrading to HTTPS where supported is always a good idea, but nothing we should invest much time into
- If you already have found a few example for incorrectly links issues/pull requests, we can fix them, but again, it's not important enough to invest much time into

bobcanthelpyou:
Updated the Git tags link to v2 and added a commit with a switch of issues and pull requests links

PR #1594, opened 2018-12-02
gluon-core: fix pattern %v in opkg URLs

belzebub40k:
$VERSION_ID is a lowercase value


rubo77:
What were the consequences of this bug before it was fixed?

We used %v in our site.

belzebub40k:
%v would return 18.06-snapshot instead of 18.06.or what ever the OpenWrt version is.

PR #1599, opened 2018-12-14
RFC: update sphinx config to v1.4.9

bobcanthelpyou:
The sphinx config was (or seems to be) created by sphinx-quickstart v1.2.3 (Debian 8)

The pull request contain a sphinx config created by sphinx-quickstart v1.4.9 (Debian 9), the option set should be the same as used for v1.2.3

Some difference between sphinx config of Gluon (-) and a fresh sphinx config (+) created with the same option set and the same toctree
```
$ diff -u docs/conf.py debian9/conf.py 
--- docs/conf.py	2018-12-14 21:13:25.716435203 +0100
+++ debian9/conf.py	2018-12-08 00:24:12.122909851 +0100
@@ -62,7 +62,7 @@
 # The short X.Y version.
 version = '2018.1+'
 # The full version, including alpha/beta/rc tags.
-release = version
+release = '2018.1+'
 
 # The language for content autogenerated by Sphinx. Refer to documentation
 # for a list of supported languages.
@@ -156,7 +156,7 @@
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-# html_static_path = ['_static']
+html_static_path = ['_static']
 
 # Add any extra paths that contain custom files (such as robots.txt or
 # .htaccess) here, relative to this directory. These files are copied
$ diff -u --color=always docs/index.rst debian9/index.rst 
--- docs/index.rst	2018-12-14 21:13:25.716435203 +0100
+++ debian9/index.rst	2018-12-14 21:22:08.481951853 +0100
@@ -434,4 +437,7 @@
 Indices and tables
 ==================
 
+* :ref:`genindex`
+* :ref:`modindex`
 * :ref:`search`
+
```

* `html_static_path` was commented out by v1.2.3, too
* `modindex` seems to be only for python modules, still removed
* `genindex` creates a link as "Index" to an empty page https://gluon.readthedocs.io/en/v2018.1.x/genindex.html removing `genindex` removes the Index-link but the file is still created (with v1.4.9 maybe it is fixed by later versions), maybe readthedocs uses a newer sphinx version where the file is removed, too.
* the Makefile used by Gluon is equal with the default Makefile by v1.2.3 and the Makefile in the pull request is equal with the default Makefile by v1.4.9


bobcanthelpyou:
I am not sure what effect it will have to the readthedocs output, but i expect it would be the same, just the link to `genindex` would be removed. I locally generated the html output on Debian 8 and 9 and compared those files. Maybe someone else should check that, too. (I didn't compare the output with the html download of the readthedocs version, yet)

The only config settings which are new / different to the old config (but the default settings) are 
```
author = 'Project Gluon'
language = None
exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store']
todo_include_todos = False
html_theme = 'alabaster'
```
I guess the `html_theme` will be overwritten by readthedocs?

The make file added some new output formats but those are not used by Gluon i guess. Maybe the `dummy` output could be useful for a CI check or something.

I just updated it, cause i had a look in the differences of the default sphinx config and the Gluon sphinx config (i guess i added the notable changes to the request text).

All these changes might not be useful at all and the config in Debian 10 with a newer sphinx version is totally different, again.

NeoRaider:
Thanks, this is rebased and merged now! I went a step further and updated the config all the way to the 1.8.4 defaults.

PR #1604, opened 2018-12-22
ar71xx: add support for FRITZ!Repeater 300E

mweinelt:
Signed-off-by: Martin Weinelt <martin@darmstadt.freifunk.net>

Slight issue: The radio is dualband and is currently autoconfigured for 5 GHz. We are lacking a sane default and/or a toggle.

```
root@64283-fritz300e:~# uci show wireless
wireless.radio0=wifi-device
wireless.radio0.type='mac80211'
wireless.radio0.hwmode='11a'
wireless.radio0.path='pci0000:00/0000:00:00.0'
wireless.radio0.htmode='HT20'
wireless.radio0.country='DE'
wireless.radio0.channel='48'
wireless.mesh_radio0=wifi-iface
wireless.mesh_radio0.ifname='mesh0'
wireless.mesh_radio0.network='mesh_radio0'
wireless.mesh_radio0.device='radio0'
wireless.mesh_radio0.mesh_fwding='0'
wireless.mesh_radio0.mesh_id='ffda-mesh-dom1'
wireless.mesh_radio0.mcast_rate='12000'
wireless.mesh_radio0.mode='mesh'
wireless.mesh_radio0.disabled='0'
wireless.mesh_radio0.macaddr='3e:34:64:7b:0c:09'
wireless.client_radio0=wifi-iface
wireless.client_radio0.ifname='client0'
wireless.client_radio0.network='client'
wireless.client_radio0.disabled='0'
wireless.client_radio0.device='radio0'
wireless.client_radio0.mode='ap'
wireless.client_radio0.macaddr='3e:34:64:7b:0c:08'
wireless.client_radio0.ssid='darmstadt.freifunk.net'
```

```
root@64283-fritz300e:~# iw phy
Wiphy phy0
	max # scan SSIDs: 4
	max scan IEs length: 2257 bytes
	max # sched scan SSIDs: 0
	max # match sets: 0
	max # scan plans: 1
	max scan plan interval: -1
	max scan plan iterations: 0
	Retry short limit: 7
	Retry long limit: 4
	Coverage class: 0 (up to 0m)
	Device supports AP-side u-APSD.
	Device supports T-DLS.
	Available Antennas: TX 0x3 RX 0x3
	Configured Antennas: TX 0x3 RX 0x3
	Supported interface modes:
		 * IBSS
		 * managed
		 * AP
		 * AP/VLAN
		 * monitor
		 * mesh point
		 * P2P-client
		 * P2P-GO
		 * outside context of a BSS
	Band 1:
[...]

PR #1606, opened 2018-12-27
gluon-core: distribute dualband radios

blocktrron:
This commit distributes dualband radios evenly on 2.4 GHz and 5GHz with 2.4 GHz being prioritised higher than 5 GHz.

This means in case a device has only a single radio and this radio supports operation in both bands, it will be set to 2.4 GHz.

#1604 depends on this PR

 - [x] Tested on a single-radio dualband device (AVM FRITZ!WLAN Repeater 300E)
 - [x] Tested on a dual-radio dualband device (OCEDO Panda)
 - [x] Tested on a single-radio singleband device (vocore2)
 - [x] Tested on a dual-radio singleband device (OCEDO Koala)

### DIFF
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
@@ -10,6 +10,43 @@ local uci = require('simple-uci').cursor()
 -- Initial
 if not sysconfig.gluon_version then
 	uci:delete_all('wireless', 'wifi-iface')
+
+	-- First count all radios with a fixed frequency band.
+	-- This is needed to distribute devices which have radios
+	-- capable of operating in the 2.4 GHz and 5 GHz band need
+	-- to be distributed evenly.
+	local radio_band_count = {band24=0, band5=0}
+	util.foreach_radio(uci, function(radio, index, config)
+		local hwmodes = iwinfo.nl80211.hwmodelist(util.find_phy(radio))
+		if (hwmodes.a or hwmodes.ac) and hwmodes.g then
+			-- Dualband - do nothing in this step
+		elseif hwmodes.g then
+			-- 2.4 GHz
+			radio_band_count["band24"] = radio_band_count["band24"] + 1
+		elseif hwmodes.a or hwmodes.ac then
+			-- 5 GHz
+			radio_band_count["band5"] = radio_band_count["band5"] + 1
+		end
+	end)
+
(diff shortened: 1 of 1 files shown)


package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -148,6 +148,37 @@ local function fixup_wan(radio, index)
...
+		-- Dualband - do nothing in this step
+	elseif hwmodes.a then
+		-- 5 GHz
+		radio_band_count["band5"] = radio_band_count["band5"] + 1
+	elseif hwmodes.g then
+		radio_band_count["band24"] = radio_band_count["band24"] + 1
+	end
+end)
+
+-- Now we have the number of all fixed 2.4G GHz and 5 GHz radios.
```
rotanid: "2.4G GHz" ?
and i wouldnt write "now we have.." but rather something like "Using the number of all ..."

rotanid:
i just tested a single-band 2.4GHz device (vocore2) and your changes didn't have any effect (so like it should be)
i did a clean flash and the wifi worked like before.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -10,6 +10,43 @@ local uci = require('simple-uci').cursor()
...
+	-- capable of operating in the 2.4 GHz and 5 GHz band need
+	-- to be distributed evenly.
+	local radio_band_count = {band24=0, band5=0}
+	util.foreach_radio(uci, function(radio, index, config)
+		local hwmodes = iwinfo.nl80211.hwmodelist(util.find_phy(radio))
+		if (hwmodes.a or hwmodes.ac) and hwmodes.g then
+			-- Dualband - do nothing in this step
+		elseif hwmodes.g then
+			-- 2.4 GHz
+			radio_band_count["band24"] = radio_band_count["band24"] + 1
```
NeoRaider: Hmm, we could use this code and clean up legacy hwmode values '11ng' and '11na' (that we have to handle in multiple places otherwise). How about using the strings '11n' and '11a' as map keys in *radio_band_count* - the increment + uci:set could be refactored into a local function that way, where only radio_name and one of these two strings need to be passed.

NeoRaider: Hmm, or not. At least when guarded by `if not sysconfig.gluon_version then`, this will obviously not help with upgrading legacy configurations...

PR #1608, opened 2018-12-29
ramips-mt76x8: remove broken flag

blocktrron:
This commit removes the broken flag from all devices in the mt76x8
subtarget.

The stability of the mt76 driver for the mt7628 and mt7612 has greatly
improved in the last half-year. It might be still behind ath9k and
ath10k but it is suitable for daily use.

This affects the following devices:

 - GL.iNet MT300N v2
 - TP-Link Archer C50 v3
 - TP-Link TL-WR841 v13

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -372,6 +372,15 @@ ramips-mt7621
 ramips-mt76x8
 ^^^^^^^^^^^^^
 
+* GL.iNet
+
+  - GL-MT300N v2 [#80211s]_
+
+* TP-Link
+
+  - TL-WR841N v13 [#80211s]_
+  - Archer C50 v3 [#80211s]_
+
 * VoCore
 
   - VoCore2 [#80211s]_
--- a/targets/ramips-mt76x8
+++ b/targets/ramips-mt76x8
@@ -1,9 +1,8 @@
-if [ "$BROKEN" ]; then
-
 # GL.iNet
 
 device gl-mt300n-v2 gl-mt300n-v2
 factory
+
 
 # TP-Link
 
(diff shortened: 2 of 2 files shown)


rubo77:
I thought it only runs fine if you do MoL?

https://forum.freifunk.net/t/installation-und-erfahrungen-mit-tp-link-archer-c50-v3/19520/23?u=rubo77

rotanid:
please read also comments after the one you linked.
blocktrron and myself were observing mt76x8 devices over a long period and came to the conclusion already written in this PR here, see above.

blocktrron:
@rubo77 
It might be a bit confusing but just to clarify: If i would have the choice i would pick a recent ath9k based device over every mt76x8 based one.

The thing is - yes the driver is not that stable as ath9k or ath10k. But with ath9k options slowly disappearing and being replaced with chips we don't know yet if they can be supported (QCN5502), having the option of less stable MediaTek devices is something we can only be happy about.

mt76 has improved greatly stability wise even for mesh applications (which are not in the focus of the development afaik) so i think we are heading in the right direction there. As we might come to the point of people HAVING to replace their TL-WR841N, low cost options will be very appreciated.

kueken:
Do you actually have no desire to at least take over version 14 (4/32) in principle?
In principle it is still completely sufficient for Freifunk and would be helpful to the community as a 15 € router.
Probably many have bought the device 'accidentally' and it is now lying around unused.

The cheapest alternative is currently the 801n (d) v5 from TPlink.
Also the same chipset.
Then there would be Netgear 6120.
Just a question.
Openwrt is built and runs for the 841n V14.
I also get Sysupgrade built for Freifunk, but the router cannot be flashed via sysupgrade?

rotanid:
@kueken this place (and old PR only vaguely related to your request) is not here to discuss new device support - please open a new pull request (PR) which provides the code for the device support you want and/or let's discuss it via IRC or mailing list.

PR #1609, opened 2019-01-02
add renamed target information to changelog

RalfJung:
I think target name changes should be mentioned in the changelog.  We had two new firmware CI builds fail because the build script still used the old target names.


RalfJung:
This is touching the 2018.2 release notes so I submitted the PR against the 2018.2 branch, not sure if that was the right call.

NeoRaider:
Thanks, merged with some changes.

PR #1613, opened 2019-01-07
Removed BROKEN because sysupgrade works now

mortzu:
OpenWRT on Mikrotik supports sysupgrade now. Tested on mANTBox 15s (RB921GS-5HPacD-15S). I suggest removing BROKEN.

### DIFF
--- a/targets/targets.mk
+++ b/targets/targets.mk
@@ -3,6 +3,7 @@ GLUON_WLAN_MESH_11s := $(filter 11s,$(GLUON_WLAN_MESH))
 $(eval $(call GluonTarget,ar71xx,generic))
 $(eval $(call GluonTarget,ar71xx,tiny))
 $(eval $(call GluonTarget,ar71xx,nand))
+$(eval $(call GluonTarget,ar71xx,mikrotik))
 $(eval $(call GluonTarget,brcm2708,bcm2708))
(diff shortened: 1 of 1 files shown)


rubo77:
Please edit your commit message so it starts with "target mikrotic:"

mweinelt:
`ar71xx` or `ar71xx-mikrotik` would be more appropriate as the prefix.

rotanid:
currently we have two device specifications in ar71xx-mikrotik.
marking this target as "unbroken" needs more than just some new information

* the sysupgrade process has to be tested with devices of both types we have in there
* the device integration has to be completely done: https://github.com/freifunk-gluon/gluon/wiki/Device-Integration-checklist

txt-file:
IMHO the best way to go is to unbreak the target and mark the devices inside the target as broken/untested.

txt-file:
I wrote a different commit for that change. https://gitlab.com/txt.file/gluon/commit/97a01a1ddf9f2fcd9227276743a7e158ef5f8879

My iea is to use aliases as we are doing with the ubiquity devices. Mark tested devices as stable and untested devices as broken. Then we can check the device integration device by device.

I am currently compile-test my change as I am lacking the time to test it on my wAP ac.

rotanid:
i agree with this, let's go on with very specific testing, e.g. in #1614

Issue #1584, opened 2018-11-25
IBSS mesh and client mode doesn't work in parallel with ath10k-ct


oliver:
After the general 5GHz wifi problems with AC7 and 2018.1.1+bremen1 firmware were solved (in #1561), we found out that _meshing_ on 5GHz still does not work.

This affects all FFHB 2018.1 firmwares so far, and has (only) been tested on Archer C7 v2 so far.

Symptoms are that a 5GHz-only device (eg. CPE510) does not create a working wifi mesh connection to the AC7.

Example device: http://og-ac7-testing-2.nodes.ffhb.de (https://map.ffhb.de/#!/en/map/f4f26d70b277).
The AC7 is in the same room as a CPE510 (https://map.ffhb.de/#!/en/map/c4e984b0a84a) and a WDR3600 (https://map.ffhb.de/#!/en/map/c4e984d5138e), both with 2017.1.8+bremen1 (ie. stable) firmware.
The Stable-FW devices are meshing nicely. The AC7 is meshing only via 2.4GHz. As result, the WDR3600 meshes with both other devices while the CPE510 meshes only with WDR3600.

Interestingly the devices appear to see each other via IBSS, but still don't get a mesh connection working:
```
root@cpe510-og-1:~# iw dev ibss0 station dump
Station aa:85:d3:37:f0:5e (on ibss0)
	inactive time:	0 ms
	rx bytes:	167231521
	rx packets:	1267310
	tx bytes:	15690420
	tx packets:	65650
	tx retries:	2531
	tx failed:	0
	rx drop misc:	0
	signal:  	-58 [-60, -62] dBm
	signal avg:	-59 [-61, -63] dBm
	tx bitrate:	144.4 MBit/s MCS 15 short GI
	rx bitrate:	144.4 MBit/s MCS 15 short GI
	expected throughput:	46.875Mbps
	authorized:	yes
	authenticated:	yes
	associated:	yes
	preamble:	long
	WMM/WME:	yes
	MFP:		no
	TDLS peer:	no
	DTIM period:	0
	beacon interval:100
	short slot time:yes
	connected time:	5629 seconds
Station 0e:9e:42:44:91:aa (on ibss0)
	inactive time:	30 ms
	rx bytes:	73632280
	rx packets:	567327
	tx bytes:	0
	tx packets:	0
	tx retries:	0
	tx failed:	0
	rx drop misc:	0
	signal:  	-61 [-63, -65] dBm
[...]

PR #1614, opened 2019-01-07
mikrotik-mantbox-15s-r2: add support for mANTBox 15s

mortzu:
  - [ ] webinterface
  - [x] tftp
  - [ ] other: <specify>
- wired network
- wifi (if applicable)
- led mapping
  - power/sys led
    - [ ] lit while the device is on
  - radio leds
    - [ ] should show activity
  - switchport leds
- [ ] reset/wps button must return device into config mode
- [ ] primary mac should match address on device label (or packaging)

### DIFF
--- a/targets/ar71xx-mikrotik
+++ b/targets/ar71xx-mikrotik
@@ -1,8 +1,15 @@
 config 'CONFIG_GLUON_SPECIALIZE_KERNEL=y'
 
+ATH10K_PACKAGES='kmod-ath10k-ct ath10k-firmware-qca9887 ath10k-firmware-qca988x'
+if [ "$GLUON_WLAN_MESH" = 'ibss' ]; then
+	ATH10K_PACKAGES='kmod-ath10k-ct ath10k-firmware-qca9887-ct ath10k-firmware-qca988x-ct'
+fi
+
 # Enable ath5k in addition to ath9k
 # ath5k cards are commonly used with Mikrotik hardware
 packages 'kmod-ath5k'
+
+packages $ATH10K_PACKAGES
 
 device mikrotik-nand-64m nand-64m
 factory
--- a/targets/ar71xx-mikrotik
+++ b/targets/ar71xx-mikrotik
@@ -1,18 +1,19 @@
 config 'CONFIG_GLUON_SPECIALIZE_KERNEL=y'
 
-ATH10K_PACKAGES='kmod-ath10k-ct ath10k-firmware-qca9887 ath10k-firmware-qca988x'
-if [ "$GLUON_WLAN_MESH" = 'ibss' ]; then
-	ATH10K_PACKAGES='kmod-ath10k-ct ath10k-firmware-qca9887-ct ath10k-firmware-qca988x-ct'
-fi
-
 # Enable ath5k in addition to ath9k
 # ath5k cards are commonly used with Mikrotik hardware
 packages 'kmod-ath5k'
 
-packages $ATH10K_PACKAGES
+if [ "$GLUON_WLAN_MESH" = '11s' ]; then
+device mikrotik-mantbox-15s-r2 nand-large
+packages kmod-ath10k-ct ath10k-firmware-qca988x
+factory
+fi
 
+if [ "$BROKEN" ]; then
 device mikrotik-nand-64m nand-64m
 factory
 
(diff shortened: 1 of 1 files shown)


targets/ar71xx-mikrotik
```diff
@@ -1,9 +1,16 @@
 config 'CONFIG_GLUON_SPECIALIZE_KERNEL=y'
 
+ATH10K_PACKAGES='kmod-ath10k-ct ath10k-firmware-qca9887 ath10k-firmware-qca988x'
```
mweinelt: I'm not sure about the firmware, that you're including here.
> Wireless 5 GHz chip model: QCA9892

targets/ar71xx-mikrotik
```diff
@@ -1,9 +1,16 @@
...
+if [ "$GLUON_WLAN_MESH" = 'ibss' ]; then
+	ATH10K_PACKAGES='kmod-ath10k-ct ath10k-firmware-qca9887-ct ath10k-firmware-qca988x-ct'
+fi
```
mweinelt: Adding packages without testing them is not helpful. If you're testing for Freifunk Bremen you've probably only tested IBSS. Please remember that support for 802.11s is a requirement and there have been issues with ath10k-ct and IBSS mesh (see #1584).

targets/ar71xx-mikrotik
```diff
@@ -1,9 +1,16 @@
...
+
 # Enable ath5k in addition to ath9k
 # ath5k cards are commonly used with Mikrotik hardware
 packages 'kmod-ath5k'
 
+packages $ATH10K_PACKAGES
```
mweinelt: Please only add the package assignment below a device that actually needs it, not globally for the target, as right now the packages will also be part of the devices that don't make use of it.

targets/ar71xx-mikrotik
```diff
@@ -1,9 +1,16 @@
...
+
 device mikrotik-nand-64m nand-64m
 factory
```
mweinelt: Add an alias for your specific model below the matching device statement, so we can have proper human-readable and parsable filenames. That way we can also track the tested device support.

targets/ar71xx-mikrotik
rotanid: i think this could be difficult as you can exchange the wifi modules in some mikrotik devices (like in other devices like x86 APU2 or mvebu turris omnia)

mweinelt: Okay.

targets/ar71xx-mikrotik
mortzu: What should be the right firmware? This one works

targets/ar71xx-mikrotik
mortzu: Yeah. Right. IBSS does not work with ath10k concurrent with client network. I've changed the support only for 11s.

targets/ar71xx-mikrotik
mweinelt: Interesting, I don't see a firmware matching the 9892 number, but if that one works, so be it.

blocktrron: @mweinelt The 3rd digit indicates "SoHo / Enterprise". They re basically identical in terms of required driver.

mweinelt: Okay, so drop  `ath10k-firmware-qca9887`.

targets/ar71xx-mikrotik
```diff
@@ -4,8 +4,16 @@ config 'CONFIG_GLUON_SPECIALIZE_KERNEL=y'
 # ath5k cards are commonly used with Mikrotik hardware
 packages 'kmod-ath5k'
 
+if [ "$GLUON_WLAN_MESH" = '11s' ]; then
+device mikrotik-mantbox-15s-r2 nand-large
```
mweinelt: The first parameter to device is the name used by the autoupdater. To discover the name that should go there run the following command on your preliminary image:

```
```

mweinelt:
This PR needs to be updated to work with the new lua based target system.

mweinelt:
Hey, this pull has been idle for quite some time and we agreed to close it.

Feel free to ask us to reopen when you're going to continue working on it.

PR #1619, opened 2019-01-13
Docs: elaborate automatic updates regarding user consent

### LINKED ISSUE #1618: Auto-Updater & Documentation: Regarding user consent
The current documentation on automatic updates and its implementation appears incomplete regarding user consent. Two shortcomings appear to exists:

- [ ] The config-mode wizard does not ask users regarding consent on automatic-updates. It's an expert feature not presented within the wizard
- [ ] The documentation does not outline, that enabling automatic updates in the site-configuration  may result in remote access to nodes, network and data without explicit user consent. An active configuration appears to be favorable for public installations managed by a group of people - But for community networks (e.g. Freifunk) this is rather problematic. It allows backdoorish access  to private devices, networks and data. It appears reasonable to document a proper warning explaining the situation and suggesting, that user consent should be gained out-of-band.

yanosz:
Addressing #1618

### DIFF
--- a/docs/features/autoupdater.rst
+++ b/docs/features/autoupdater.rst
@@ -1,7 +1,15 @@
 Autoupdater
 ===========
 
-Gluon contains an automatic update system which can be configured in the site configuration.
+Gluon contains an automatic update system, which can be configured in the site configuration.
+
+Please mind that asking users for consent is not implemented within the automatic update system.
+If automatic updates are enabled by default, users are neither asked nor notified when updates are available or installed.
+
+It is reasonable to ask users for consent before administrators gain access to private devices
+and networks via automatic updates. This could be done before downloading, e.g. by providing different download options.
+Some community networks have guidelines regarding automatic update policies, that require agreeing explicitly.
+(e.g. `Freifunk Memorandum of Understanding <https://github.com/freifunk/MoU/blob/master/FreifunkMemorandumofUnderstanding_en.md>`).
 
 Building Images
 ---------------


docs/features/autoupdater.rst
```diff
@@ -3,6 +3,13 @@ Autoupdater
 
 Gluon contains an automatic update system which can be configured in the site configuration.
 
+Please mind, that asking users for consent is not implemented within the automatic update system.
```
rotanid: i think the comma has to be removed here

rotanid:
please:
- lower case beginning of the commit message
- mention the fixed issue not in the first line of the commit message, as it gets too long, you can add a blank line and additional information after that

i thought you may want to add similar information to the config mode and place the autoupdater in a more visible way?

docs/features/autoupdater.rst
mweinelt: Please remember that...

docs/features/autoupdater.rst
```diff
@@ -3,6 +3,13 @@ Autoupdater
...
+Also, users are neither asked nor notified when updates are available or installed.
```
mweinelt: Connect the sentences with a comma, they are directly related. Drop "Also ,".

docs/features/autoupdater.rst
```diff
@@ -3,6 +3,13 @@ Autoupdater
...
+
+It appears reasonable to ask users for consent before administrators gain access to private devices and networks via automatic updates. This could be done before downloading, e.g. by providing different download options.
```
mweinelt: That sounds very vague ("it appears that...") and because it is a serious matter it should not.

docs/features/autoupdater.rst
```diff
@@ -3,6 +3,13 @@ Autoupdater
...
+Some community networks have guidelines regarding automatic update policies (e.g. Freifunk Memorandum of Understanding).
```
mweinelt: The MoU only talks about consent, yet if you want to reference it please as quote & attribution.

docs/features/autoupdater.rst
mweinelt: The effort to provide a dedicated set of images with a disabled autoupdater feels disproportionate. I'd much prefer a in-firmware solution.

docs/features/autoupdater.rst
kaechele: If we include this at all I guess we should clarify the difference between what we mean by consent vs. the already existing ability of turning the Autoupdater on or off in config mode.
Especially how consent must be informed, i.e. ensuring the user is aware of the implications of their consent.

yanosz:
Thanks for your feedback. I changed some parts accordingly.

Regarding the achievement of this text: It's supposed to inform builders problematic configurations for freifunk networks.. It is suppose to document gluon as is, without modification. 

I agree, that a more visible part is better, but its ongoing: From my impression, automatic updates are a matter of dispute in the Freifunk community. It seems hard to propose a design balancing the different interest. 
Thus I think its reasonable to document the issue as it is.

Please correct typos you encounter while merging. It doesn't look fruitful to have a github discussion on that.

rotanid:
> Please correct typos you encounter while merging. It doesn't look fruitful to have a github discussion on that.

this is not about the discussion, it is about "who does the work". your suggestion moves the work load to other people, which prefer the author getting his/her contribution merge-button-click-ready.

yanosz:
@rotanid : I don't think, that we can go one, here. I respect your decision not to merge this PR as it is.

rotanid:
@yanosz maybe we can update this docs part together with the changes to the config mode you propose?
this change here would be partly obsoleted by those changes and would need to be changed again.

PR #1620, opened 2019-01-15
[WIP] target ar71xx-mikrotik: add missing OpenWrt image types

txt-file:
Add the missing OpenWrt image types according to https://openwrt.org/toh/mikrotik/common#downloading_openwrt_images

sorted alphanumeric, compile tested

### DIFF
--- a/targets/ar71xx-mikrotik
+++ b/targets/ar71xx-mikrotik
@@ -1,9 +1,5 @@
 config 'CONFIG_GLUON_SPECIALIZE_KERNEL=y'
 
-# Enable ath5k in addition to ath9k
-# ath5k cards are commonly used with Mikrotik hardware
-packages 'kmod-ath5k'
-
 device mikrotik-rb-nor-flash-16M rb-nor-flash-16M
 factory
 
--- a/targets/ar71xx-mikrotik
+++ b/targets/ar71xx-mikrotik
@@ -4,8 +4,17 @@ config 'CONFIG_GLUON_SPECIALIZE_KERNEL=y'
 # ath5k cards are commonly used with Mikrotik hardware
 packages 'kmod-ath5k'
 
+device mikrotik-rb-nor-flash-16M rb-nor-flash-16M
+factory
+
+device mikrotik-rb-nor-flash-16M-ac rb-nor-flash-16M-ac
+factory
+
 device mikrotik-nand-64m nand-64m
 factory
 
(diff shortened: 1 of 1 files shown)


rotanid:
@txt-file you added the x86 squashfs change to this PR, which is very wrong. please leave this PR only about mikrotik.

txt-file:
true. I commited to the wrong branch. Removed the commit.

targets/ar71xx-mikrotik
```diff
@@ -4,8 +4,17 @@ config 'CONFIG_GLUON_SPECIALIZE_KERNEL=y'
 # ath5k cards are commonly used with Mikrotik hardware
 packages 'kmod-ath5k'
```
mweinelt: This is certainly not the case for 802.11ac compatible devices anymore. Can you speak to that?

txt-file: This i probably true for older RouterBoards which had miniPCI connectors. As we probably won't support these (old) devices I can remove the line. According to [wireless.wiki.kernel.org](https://wireless.wiki.kernel.org/en/users/drivers/atheros) ath5k only supports a/b/g. I'll remove the line in my next version of the patch.

txt-file:
I'lll prepare a new patch with added devices and broken/untested flags.

mweinelt:
I just realised what you meant by "added devices". I don't think it's wise to track all possible devices as aliases below without having a test protocol. Nobody is going to maintain that alias list anyway.

Please, just stick to the different image types and reasonable kmod/firmware packages.

txt-file:
I just removed kmod-ath5k. Any further changes?

Should I add devices / aliases or not?

mweinelt:
The device aliases would imply someone took a deeper look at the individual device, as that is not the case I'd rather have them left out.

targets/ar71xx-mikrotik
txt-file: https://github.com/freifunk-gluon/gluon/pull/1620/commits/78386e79c99e5a7b6525a54080afe316ec9061de removes 'kmod-ath5k'

txt-file:
alright

mweinelt:
Replace "target ar71xx-mikrotik" with just "ar71xx-mikrotik" and then I think we're ready to merge.

mweinelt:
Applied with the commit message changed. Thanks!

PR #1621, opened 2019-01-17
mt7620: add support for Nexx WT3020 (8M)

Dark4MD:
  
 * Works easy with the default Firmware

* wired network
  
* wifi (if applicable)
  
* led mapping
 
  * power/sys led
    
  * radio leds
    
    * [ ]  should show activity

      * device has only one LED for power status

  * switchport leds
    
      * device has only one LED for power status

* it just have an reset button and it works fine

* [ ]  primary mac should match address on device label (or packaging) (https://gluon.readthedocs.io/en/latest/dev/hardware.html#notes)

* it doesn't have a mac adress on the case or sticker printed and the packaging is long gone, so i can't confirm, neither deny this.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -352,6 +352,10 @@ ramips-mt7620
   - GL-MT300N [#80211s]_
   - GL-MT750 [#80211s]_
 
+* Nexx
+
+  - WT3020AD/F/H (8M)
+
 ramips-mt7621
 ^^^^^^^^^^^^^
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -354,7 +354,7 @@ ramips-mt7620
 
 * Nexx
 
-  - WT3020AD/F/H (8M)
+  - WT3020AD/F/H
 
 ramips-mt7621
 ^^^^^^^^^^^^^
--- a/targets/ramips-mt7620
+++ b/targets/ramips-mt7620
@@ -1,4 +1,5 @@
 # GL Innovations
+
 device gl-mt300a gl-mt300a
 factory
 
@@ -7,5 +8,7 @@ factory
 
 device gl-mt750 gl-mt750
 factory
+
+# Nexx
 
 device nexx-wt3020-8m wt3020-8M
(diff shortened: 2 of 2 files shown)


blocktrron:
Regarding the primary mac: Have you tried to identify a matching MAC-Address when running the stock firmware? This should be enough as the primary mac has to be device-independent in a mesh.

Dark4MD:
I just checked the OpenWRT Wiki for a way to get back to stock and found this hint that says, that the mac address of the device is stored in /dev/mtd2 with offset 4 and just checked it.
I have the same MAC Address in Gluon like the one that is stored in /dev/mtd2.

targets/ramips-mt7620
```diff
@@ -7,3 +7,5 @@ factory
 
 device gl-mt750 gl-mt750
 factory
+
```
mweinelt: Add the vendor in a comment, as above for GL.iNet.

targets/ramips-mt7620
```diff
@@ -7,3 +7,5 @@ factory
...
+device nexx-wt3020-8m wt3020-8M
```
mweinelt: I wonder if it's worth it to create aliases for the different 8M models (WT3020AD, WT3020F, WT3020H).

mweinelt:
Start your commit message with the target your changes are against, for example:

`ramips-mt7620: add support for Nexx WT3020 (8M)`

docs/index.rst
```diff
@@ -352,6 +352,10 @@ ramips-mt7620
   - GL-MT300N [#80211s]_
   - GL-MT750 [#80211s]_
 
+* Nexx
+
+  - WT3020AD/F/H (8M)
```
mweinelt: As only the WT3020A has 4 MB Flash the "(8M)" suffix is unnecessary.

mweinelt:
Squashed and merged in https://github.com/freifunk-gluon/gluon/commit/04cca280a1ff27f7dfa51031210d4e08d86607e9.

Thanks!

Issue #211, opened 2014-10-13
Poor wireless performance compared to stock firmware


sargon:
Our test setup consists of one tp-wr841n and two laptops one connected via wire and one via wireless.
In the step we tested throughput with the stock firmware which results in:

```
[root@pan ~]# iperf -s
WARNING: interval too small, increasing from 0.00 to 0.5 seconds.
------------------------------------------------------------
Server listening on TCP port 5001
TCP window size: 85.3 KByte (default)
------------------------------------------------------------
[  4] local 192.168.0.10 port 5001 connected with 192.168.0.100 port 58049
[ ID] Interval       Transfer     Bandwidth
[  4]  0.0- 0.5 sec  4.12 MBytes  69.2 Mbits/sec
[  4]  0.5- 1.0 sec  5.45 MBytes  91.4 Mbits/sec
[  4]  1.0- 1.5 sec  5.19 MBytes  87.0 Mbits/sec
[  4]  1.5- 2.0 sec  4.17 MBytes  70.0 Mbits/sec
[  4]  2.0- 2.5 sec  5.18 MBytes  86.9 Mbits/sec
[  4]  2.5- 3.0 sec  5.10 MBytes  85.6 Mbits/sec
[  4]  3.0- 3.5 sec  5.49 MBytes  92.1 Mbits/sec
[  4]  3.5- 4.0 sec  5.40 MBytes  90.6 Mbits/sec
[  4]  4.0- 4.5 sec  5.48 MBytes  91.9 Mbits/sec
[  4]  4.5- 5.0 sec  5.49 MBytes  92.1 Mbits/sec
[  4]  5.0- 5.5 sec  5.52 MBytes  92.6 Mbits/sec
[  4]  5.5- 6.0 sec  5.25 MBytes  88.1 Mbits/sec
[  4]  6.0- 6.5 sec  5.52 MBytes  92.7 Mbits/sec
[  4]  6.5- 7.0 sec  5.53 MBytes  92.7 Mbits/sec
[  4]  7.0- 7.5 sec  5.24 MBytes  87.9 Mbits/sec
[  4]  7.5- 8.0 sec  5.13 MBytes  86.1 Mbits/sec
[  4]  8.0- 8.5 sec  5.20 MBytes  87.2 Mbits/sec
[  4]  8.5- 9.0 sec  4.95 MBytes  83.1 Mbits/sec
[  4]  9.0- 9.5 sec  5.41 MBytes  90.8 Mbits/sec
[...]

PR #1622, opened 2019-01-18
batman-adv-legacy: batman-adv co-existence

T-X:
This patchset allows installing the mainline batman-adv (compat 15) and the batman-adv-legacy (compat 14) fork simultaneously.

Which module to use is determined via the routing_algo parameter in the site configuration: If set to BATMAN_IV_LEGACY then the batman-adv-legacy (compat 14) module is used. Otherwise (e.g. BATMAN_IV or BATMAN_V) the mainline batman-adv module is chosen.

Together with the gluon-scheduled-domain-switch package (freifunk-gluon/gluon#1555) this should ease the migration from compat 14 to compat 15 and speed-up the deprecation of this fork.

Note: This pull-request should only be applied after applying freifunk-gluon/packages#211.


T-X:
There was no new compatibility version number for five five years now.

With compat-15, TVLVs (type-version-length-values) were introduced with the intent of introducing new features and changes without creating incompatibilities. So it might(tm) be the last compat version bump. Or at least a compat version increment is unlikely to happen in the next years.

T-X:
Changelog v2:
* rebased to current master
* "modules: update Gluon packages" commit hashes updated to potential package branch after https://github.com/freifunk-gluon/packages/pull/211 merge

T-X:
Will be updated and simplified after the batctl v2013.4 removal (#1654 and https://github.com/freifunk-gluon/packages/pull/213).

rotanid:
i removed to "blocked"-status as the mentioned PRs have been merged

T-X:
Changelog v4:
* rebased to current master
* "modules: update Gluon packages" commit hashes updated to potential package branch after freifunk-gluon/packages#211 merge
* Removed UCI config template for batman-adv and readded it in gluon-mesh-batman-adv-common; fixes an unconfigured gluon-mesh-batman-adv-legacy if only this package was installed

T-X:
Changelog v5:
* rebased to current master:
  * 0004-batman-adv-remove-module-autoload.patch => 0005-batman-adv-remove-module-autoload.patch
  * 0005-batman-adv-remove-UCI-config-template.patch => 0006-batman-adv-remove-UCI-config-template.patch

package/gluon-mesh-batman-adv/files/etc/config/batman-adv
```diff
@@ -0,0 +1,20 @@
...
+	option 'orig_interval'
+	option 'bridge_loop_avoidance'
+	option 'distributed_arp_table'
+	option 'multicast_mode'
+	option 'network_coding'
+	option 'hop_penalty'
+	option 'isolation_mark'
+
+# yet another batX instance
+# config 'mesh' 'bat5'
```
NeoRaider: Moving this file to gluon-mesh-batman-adv is quite ugly. If we really want to merge coexistence before updating to batadv v2019.2 (and thus before Gluon v2019.1), splitting this into `/etc/config/batman-dav` and `/etc/config/batman-adv-legacy` would seem preferable to me - but we could also defer this until after the release.

T-X:
Changelog v6:
* rebased to current master
* "modules: update Gluon packages" commit hashes updated to potential package branch after freifunk-gluon/packages#211 merge
* kept /etc/config/batman-adv in the batman-adv routing feed package, as #211 now moved the config file to its own /etc/config/batman-adv-legacy
* Patch "gluon-mesh-batman-adv: update batman-adv-legacy config" was added to set our preferences for this new /etc/config/batman-adv-legacy file, too

T-X:
The advantages of adding this to Gluon v2019.1 instead of postponing: With a migration path in place this should allow us to remove batman-adv-legacy in Gluon v2019.2 (and removing batman-adv-legacy might make the batman-adv (config) update for Gluon v2019.2 easier?).

package/features
```diff
@@ -23,8 +23,8 @@ packages 'web-advanced' \
 packages 'web-advanced & autoupdater' \
 	'gluon-web-autoupdater'
 
-packages 'status-page & (mesh-batman-adv-14 | mesh-batman-adv-15)' \
+packages 'status-page & (mesh-batman-adv-legacy | mesh-batman-adv)' \
```
NeoRaider: I would prefer to keep the package names unchanged until we actually get rid of compat 14.

package/gluon-alfred/Makefile
```diff
@@ -7,7 +7,7 @@ PKG_RELEASE:=1
 include ../gluon.mk
 
 define Package/gluon-alfred
-  DEPENDS:=+gluon-core +gluon-respondd +gluon-neighbour-info gluon-mesh-batman-adv +micrond +alfred
+  DEPENDS:=+gluon-core +gluon-respondd +gluon-neighbour-info gluon-mesh-batman-adv-common +micrond +alfred
```
NeoRaider: I think this dependency (and all similar) should be left unchanged - they actually depend on one of the batadv implementations, not just the common parts.

package/gluon-mesh-batman-adv/Makefile
```diff
@@ -18,37 +18,37 @@ define Package/gluon-mesh-batman-adv/common
...
+	+@GLUON_SPECIALIZE_KERNEL:KERNEL_CRC16 \
+	+@GLUON_SPECIALIZE_KERNEL:KERNEL_LIBCRC32C
 endef
 
-define Package/gluon-mesh-batman-adv-14/install
+define Package/gluon-mesh-batman-adv-common/install
 	$(Gluon/Build/Install)
-
-	$(INSTALL_DIR) $(1)/lib/gluon/mesh-batman-adv
-	echo 14 > $(1)/lib/gluon/mesh-batman-adv/compat
```
NeoRaider: If you remove the compat file, you also need to adjust the respondd provider that reads it.

package/gluon-mesh-batman-adv/check_site.lua
```diff
@@ -2,4 +2,4 @@
 need_boolean(in_domain({'mesh', 'vxlan'}), not this_domain())
 
 need_number({'mesh', 'batman_adv', 'gw_sel_class'}, false)
-need_one_of({'mesh', 'batman_adv', 'routing_algo'}, {'BATMAN_IV', 'BATMAN_V'}, false)
+need_one_of({'mesh', 'batman_adv', 'routing_algo'}, {'BATMAN_IV', 'BATMAN_IV_LEGACY', 'BATMAN_V'}, false)
```
NeoRaider: We need to make sure that only options for batman-adv versions that are installed may be selected.

package/gluon-mesh-batman-adv/files/lib/netifd/proto/gluon_bat0.sh
```diff
@@ -42,10 +42,17 @@ proto_gluon_bat0_setup() {
 	echo 1 > /proc/sys/net/ipv6/conf/primary0/disable_ipv6
 	ip link set primary0 address "$primary0_mac" mtu 1532 up
 
-	local routing_algo="$(uci -q get batman-adv.bat0.routing_algo || echo 'BATMAN_IV')"
-	(echo "$routing_algo" >/sys/module/batman_adv/parameters/routing_algo) 2>/dev/null
-
-	echo bat0 > /sys/class/net/primary0/batman_adv/mesh_iface
+	local routing_algo="$(uci -q get batman-adv.bat0.routing_algo || uci -q get batman-adv-legacy.bat0.routing_algo || echo 'BATMAN_IV')"
```
NeoRaider: You need to read from the site config whether batman-adv or batman-adv-legacy is to be used (and I don't think it makes sense to ever store the routing algo in /etc/config/batman-adv-legacy at all).

package/gluon-mesh-batman-adv/luasrc/lib/gluon/upgrade/310-gluon-mesh-batman-adv-mesh
```diff
@@ -23,6 +24,17 @@ uci:section('batman-adv', 'mesh', 'bat0', {
...
+uci:section('batman-adv-legacy', 'mesh', 'bat0', {
+	orig_interval = 5000,
+	gw_mode = gw_mode_legacy,
+	gw_sel_class = gw_sel_class,
+	hop_penalty = 15,
+	routing_algo = routing_algo,
+	multicast_mode = true,
+})
+uci:save('batman-adv-legacy')
+
```
NeoRaider: - These sections are applied using a procd hotplug trigger. bat0 should only be configured in batman-adv or batman-adv-legacy depending on the site config, never both
- multicast_mode doesn't make sense for -legacy

rotanid:
@T-X do you want to address the change requests?

package/gluon-mesh-batman-adv/files/lib/netifd/proto/gluon_bat0.sh
```diff
@@ -42,10 +42,17 @@ proto_gluon_bat0_setup() {
...
+
+	if [ "$routing_algo" = "BATMAN_IV_LEGACY" ]; then
+		modprobe batman-adv-legacy
```
NeoRaider: Hmm, another thought: With the hoodselector, we now support runtime domain switches. While I would still consider this support "experimental", should we handle this here by unloading the other batadv kmod here? If it is working well, we could also avoid adding the patch to disable the autoloading.

T-X:
Oh, hm, runtime domain switches... I'm not quite sure if that's a good idea with batman-adv-legacy. While for the current batman-adv module we didn't have any shutdown bug reports for a while now, I'm aware of at least one shutdown bug that is still present in batman-adv-legacy. The fix is quite invasive:

* batman-adv: fix rare race conditions on interface removal ([8a436b94d731873b](https://git.open-mesh.org/batman-adv.git/commit/8a436b94d731873b))

T-X:
Changelog v7:
* rebased to current master
* "modules: update Gluon packages" commit hashes updated to potential package branch after freifunk-gluon/packages#211 merge
* removed "mesh-batman-adv-14 -> mesh-batman-adv-legacy" and "mesh-batman-adv-15 -> mesh-batman-adv" renames
* left gluon-mesh-batman-adv dependencies unchanged
* updated respondd provider to not use /lib/gluon/mesh-batman-adv/compat anymore
* extended gluon-mesh-batman-adv/check_site.lua to also check the necessary package for a specified routing_algo was added
* gluon_bat0.sh: always determine routing_algo + module from site instead of UCI
* only add "bat0" section in either /etc/config/batman-adv or /etc/config/batman-adv-legacy, depending on which version was chosen
* avoid adding routing_algo and multicast_mode to /etc/config/batman-adv-legacy

Unchanged:
* no runtime swapping support, keep requiring full reboot (batman-adv-legacy is currently not unload-safe)

NeoRaider:
A small note that is not directly related to this PR: please *never* update the file `modules` in PRs with commit IDs that are not in the referenced repos yet. There is no way to know in advance how commits end up in the packages repo, so these updates only create additional work when merging...

NeoRaider:
Thanks, merged with some changes!

PR #1624, opened 2019-01-18
2018.2: ar71xx: add support for AVM FRITZ!WLAN Repeater 300E

blocktrron:
Backport the AVM FRITZ!WLAN Repeater 300E to the latest stable branch.

Although it might be discontinued, it is commonly available on eBay for around 15-20 Euros. As low-cost alternatives with 64M RAM and 16MB Flash are sparse, this should give a great value despite the rather weak (400MHz) CPU.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -135,6 +135,7 @@ ar71xx-generic
 * AVM
 
   - Fritz!Box 4020 [#avmflash]_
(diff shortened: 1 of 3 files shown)


rotanid:
although i'm not against this backport, i'd prefer are more broad backport strategy, let's discuss this on IRC #gluon , i just wrote something there.

mweinelt:
Cherry picked with all the other commits from master.

Issue #1623, opened 2019-01-18
gl.inet-gl-ar750 autoupdate fails on model


mmalte:
The Autoupdate of gl.inet-gl-ar750 fails with:

```
root@ffac-Bergdriesch-GL-AR750:/tmp# autoupdater 
Retrieving manifest from http://updates.freifunk-aachen.de/experimental/sysupgrade/experimental.manifest ...
autoupdater: warning: no matching firmware found (model gl.inet-gl-ar750)
Retrieving manifest from http://updates.ffac.rocks/experimental/sysupgrade/experimental.manifest ...
autoupdater: warning: no matching firmware found (model gl.inet-gl-ar750)
Retrieving manifest from http://updates.aachen.freifunk.net/experimental/sysupgrade/experimental.manifest ...
autoupdater: warning: no matching firmware found (model gl.inet-gl-ar750)
autoupdater: error: no usable mirror found
```
there is an inconsistency between cpuinfo and board_name:

```
root@ffac-Bergdriesch-GL-AR750:~# cat /tmp/sysinfo/board_name
gl-ar750
```

```
root@ffac-Bergdriesch-GL-AR750:~# cat /proc/cpuinfo
system type		: Qualcomm Atheros QCA9533 ver 2 rev 0
machine			: GL.iNet GL-AR750
processor		: 0
cpu model		: MIPS 24Kc V7.4
BogoMIPS		: 432.53
wait instruction	: yes
microsecond timers	: yes
tlb_entries		: 16
extra interrupt vector	: yes
hardware watchpoint	: yes, count: 4, address/irw mask: [0x0ffc, 0x0ffc, 0x0ffb, 0x0ffb]
isa			: mips1 mips2 mips32r1 mips32r2
ASEs implemented	: mips16
shadow register sets	: 1
kscratch registers	: 0
package			: 0
core			: 0
VCED exceptions		: not available
VCEI exceptions		: not available
```
The manifest offers updates for:

`gl-ar750 2018.2-1~exp20190117`

I'm happy to provide additional information if needed.

mweinelt:
> The first parameter defines the Gluon profile name, which is used to refer to the device and is part of the generated image name. The profile name must be same as the output of the following command (on the target device), so the autoupdater can work:

```
lua -e 'print(require("platform_info").get_image_name())'
```

Run the above and then - can you provide a PR to fix that?

mmalte:
```
[...]

PR #1625, opened 2019-01-18
 ar71xx: fix incorrect gl-ar750 image name

mmalte:
```
root@ffac-Bergdriesch-GL-AR750:~# lua -e 'print(require("platform_info").get_image_name())'
gl.inet-gl-ar750
```

Not sure if this is the way it should be fixed, but this way the manifest file is accepted by the autoupdater.

### DIFF
--- a/targets/ar71xx-generic
+++ b/targets/ar71xx-generic
@@ -79,7 +79,7 @@ factory
 device gl-ar300m gl-ar300m
 factory
 
-device gl-ar750 gl-ar750
+device gl.inet-gl-ar750 gl-ar750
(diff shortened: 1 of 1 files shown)


blocktrron:
The commit title should be prefixed with the target (in this case `ar71xx`)

I would also change the commit title to something like
`ar71xx: fix incorrect gl-ar750 image name`

mmalte:
fixes #1623

mweinelt:
Thanks!

rotanid:
@mmalte we wanted a changed commit message, not PR title ;) next time!

PR #1626, opened 2019-01-19
gluon-core: assert WiFi driver provides 4 MAC-addresses

blocktrron:
Disclaimer
----
We should not be affected by this bug yet as there are (on first sight) no affected MediaTek devices with the affected driver in Gluon. I ran into the Bug using an Archer C2 v2 on next.

Commit-message
----
Gluon has multiple ways to obtain unique MAC-addresses. They are either
provided by the WiFi driver or derived from the primary MAC-address.

Quoting the same file:

> It's necessary that the first 45 bits of the MAC address don't
> vary on a single hardware interface, since some chips are using
> a hardware MAC filter. (e.g 'rt305x')

This currently fails in case the rt35xx based chips mac address differs
from the primary MAC. In this case, the MAC address for the client0 radio
(vif 1) comes from the WiFi driver. As there is only a single
MAC-address provided by '/sys/class/ieee80211/phyX/addresses' but the
MAC-address for mesh 0 (vif 2) is derived from the Node-ID, resulting in
different first 45 bits. The WiFi won't come up altogether in this case.

This commit verifies at least 4 MAC-Addresses are provided by the WiFi
driver. If this is not the case, all MAC-addresses are derived from the
primary MAC. This way, affected radios are working correctly.

### DIFF
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
@@ -212,14 +212,21 @@ end
 local function get_wlan_mac_from_driver(uci, radio, vif)
 	local primary = sysconfig.primary_mac:lower()
 
-	local i = 1
-	for addr in get_addresses(uci, radio) do
-		if addr:lower() ~= primary then
-			if i == vif then
-				return addr
-			end
+	local addresses = {}
+	for address in get_addresses(uci, radio) do
+		if address:lower() ~= primary then
+			table.insert(addresses, address)
+		end
+	end
 
-			i = i + 1
+	-- Make sure we have at least 4 addresses
+	if #addresses < 4 then
+		return nil
+	end
+
+	for i, addr in ipairs(addresses) do
+		if i == vif then
+			return addr
 		end
 	end
 end


blocktrron:
I've just now realized we only need 4 unique non-primary mac-addresses as we are working on a per-interface basis here.

Updated my PR.

NeoRaider:
Hmm, I know I intentionally left out this check to use at least one address from ath9k/ath10k devices, and only fall back to the generated addresses for the other VIFs... but I guess this doesn't really matter.

We might also run into trouble in case the hardware/driver doesn't like to use any addresses except for the ones listed in `/sys/class/ieee80211/phyX/addresses` - I think I remember such issues on mt76? But I don't see any way to deal with this case when there aren't sufficient addresses provided by the driver, so we might as well merge this. Minor comments will follow in a moment.

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -212,15 +212,25 @@ end
...
+	local addresses = {}
+	for address in get_addresses(uci, radio) do
+		if address:lower() ~= primary then
+			table.insert(addresses, address)
+		end
+	end
 
-			i = i + 1
+	-- Make sure we have at least 4 addresses
+	if table.getn(addresses) < 4 then
```
NeoRaider: It's more idiomatic to write `#addresses` to the the length of the table.

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -212,15 +212,25 @@ end
...
+		return nil
+	end
+
+	local i = 1
+	for _, addr in ipairs(addresses) do
```
NeoRaider: We can just write `for i, addr in ...` and avoid counting ourselves.

blocktrron: done

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
blocktrron: done

blocktrron:
Changed accordingly. Currently compiling, will test soon.

blocktrron:
Just tested the current state, works without a problem on my device.

@NeoRaider mt76 is not affected by this PR as it provides 8 addresses per radio.

blocktrron:
@NeoRaider regarding the ath9k/ath10k thing: If you would like to preserve the behavior, we can exclude ath9k/ath10k from this patch. However, this should not make a difference in practice (except for all WiFi client BSSIDs become randomized)

NeoRaider:
I'll merge it like this for now, we can add an exception for ath9k/ath10k later if any problems arise.

PR #1627, opened 2019-01-21
ramips-mt7620: add support for Xiaomi MiWifi Mini

Dark4MD:
  
 * There are two ways to flash. Both are written in the OpenWRT Wiki. One via an Code Injection Bug and one via an Account on miwifi.com
* https://openwrt.org/toh/xiaomi/mini#quick_openwrt_installation

* wired network
  
* wifi (if applicable)
  
* mesh is working but it depends on PR #1626 to get it working on both radios, without it only mesh on radio0 (the extra wifi chip) is working.

* led mapping
 
  * power/sys led
    
  * radio leds
    
    * [ ]  should show activity

      * device has only one LED for power status

  * switchport leds
    
      * device has only one LED for power status

* it just have an reset button and it works fine

* [x]  primary mac should match address on device label (or packaging) (https://gluon.readthedocs.io/en/latest/dev/hardware.html#notes)

* fixed


blocktrron:
#1626 should fix the issues with the 2.4 radio, can you confirm that?

achterin:
@ambassador86 and I experimented with similar devices (TP-Link Archer C20v1 & ASUS RT-AC51U) and had the same issue. If you disable the 5GHz chip completely (edit dts-boardfile in OpenWrt) the 2.4GHz radio is capable of running AP & Mesh-Mode in parallel.
@blocktrron i will test this tonight on my asus device and report back.

Dark4MD:
@blocktrron i will try this and report back as soon as i got it running with the potential fix and yes, the image name is the one i get as an output with "lua -e 'print(require("platform_info").get_image_name())'"

Dark4MD:
After applying the Fix in #1626 mesh on both radios is working.

package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac
```diff
@@ -28,6 +28,8 @@ if platform.match('ar71xx', 'generic', {'tl-wdr3600', 'tl-wdr4300',
   table.insert(try_files, 1, '/sys/class/ieee80211/phy1/macaddress')
 elseif platform.match('ramips', 'mt7621', {'dir-860l-b1'}) then
   table.insert(try_files, 1, '/sys/class/ieee80211/phy1/macaddress')
+elseif platform.match('ramips', 'mt7620', {'miwifi-mini'}) then
+  table.insert(try_files, 1, '/sys/class/net/eth0/address')
```
rotanid: are you sure this is the only way, not like with mt7621 ?

rotanid:
maybe you could link the openwrt wiki article in your PR text above if you mention it.

i would suggest to remove the broken flag and instead mark this PR as broken so it isn't merged before #1626

package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac
mweinelt: And mt7620 before mt7621.

Dark4MD: i tried the other ways, phy1/macaddress is off by one in the last digit, only the address on eth0 is matching the one on the case.

mweinelt:
When you're done please squash your changes into a single commit.

Dark4MD:
I open a new, clean PR

PR #1637, opened 2019-01-30
meta: add template for bug reports

mweinelt:
As low-quality bug reports have recently been creeping up I felt it was time to come up with a bug reporting template that asks for the nitty-gritty, by reminding reporters of useful questions that can be answered ahead of time.

This will hopefully help with triaging issues more quickly.


achterin:
Maybe add the device device-version as well? For several devices it is crucial to know which version is used as they might differ drastically (other architecture etc.).

rubo77:
looks great!

maybe add a hint for unexperienced github- and markdown-users that they should use the "Preview"-Tab to see the outcome of the Bugreport before committing.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,71 @@
...
+---
+
+<!--
+
+Please carefully fill out the questionaire below to help improve the
+timely triaging of issues. Walk through the questions below and use
+them as an inspiration for what information you can provide.
+
+Make use of codeblocks (three backticks before and after) where
+appropriate (configuration exceperts, log output, etc.). Example:
```
rotanid: *excerpts

skorpy2009:
Maybe also ask if / how one can reproduce the behavior

mweinelt:
Is reproducing an issue not already covered by asking for "everything" that was used to build the firmware. For the parts the user actually did we ask "What were you doing when you first noticed the problem?".

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,71 @@
...
+```
+
+You can use the "Preview" tab to check how your issue is going to look
+before you actually send it in.
+
+Thank you for taking the time to report a bug with the Gluon project.
+
+-->
+
+### Bugreport
```
NeoRaider: Missing space

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,71 @@
...
+
+**What is the problem?**
+<!-- 
+- What is not working as expected?
+- How is it misbehaving?
+- When did the problem first start showing up?
+- What were you doing when you first noticed the problem?
+- On which devices (vendor, model and revision) is it misbehaving?
+- Does the issue appear on multiple devices or targets?
+-->
```
NeoRaider: I'm wondering whether having so many questions in a single comment is a good idea. People without a technical background might write their answers inside the comment...

mweinelt: I'd say people without any technical background rarely have accounts on GitHub to open an issue in the first place. Adding documentation on how to fill out the bug report inside of HTML comments is quite normal.

I can add another comment that tells the reporter to watch out and not put their text inside a comment, if that helps at all, seeing as we've amassed quite some text already.

mweinelt: @neoraider On second thought, if you're only wondering I'd say that we should just give it a try and makes changes as we go. This is only our first issue template after all.

rubo77:
This is great, but now we need another template for other issues than bugs, otherwise, you can only add bugs here: https://github.com/freifunk-gluon/gluon/issues/new/choose

PR #1640, opened 2019-02-02
ar71xx: fix incorrect GL.iNet image names

oszilloskop:
In current Gluon versions the content of the manifest file does not match with the device identifier of a GL.iNet GL-AR150. To solve this issue the manifest content and device identifier must be correspond.

Solution:
This commit adjusted the GL-AR150 Gluon profile name which is used to refer to the device and image name.

### DIFF
--- a/targets/ar71xx-generic
+++ b/targets/ar71xx-generic
@@ -72,14 +72,19 @@ factory
 # GL Innovations
 
 device gl-inet-6408a-v1 gl-inet-6408A-v1
+
 device gl-inet-6416a-v1 gl-inet-6416A-v1
 
-device gl-ar150 gl-ar150
+device gl.inet-gl-ar150 gl-ar150
+manifest_alias gl-ar150
 factory
-device gl-ar300m gl-ar300m
+
+device gl.inet-gl-ar300m gl-ar300m
+manifest_alias gl-ar300m
 factory
 
 device gl.inet-gl-ar750 gl-ar750
+manifest_alias gl-ar750
 packages $ATH10K_PACKAGES
 factory
 


bobcanthelpyou:
Ob noch weitere GL Router betroffen sind?
(gl-ar750: https://github.com/freifunk-gluon/gluon/pull/1625)

blocktrron:
See this upstream commit:

https://github.com/openwrt/openwrt/commit/bb1df2d75bcc27614505ad9979f3eb2ad4c1fd40

So yes, we should do a proper fix for all these affected devices. This includes changing the device name to the current format and adding aliases for backwards compatibility, as the requested string is depended on the current running firmware.

oszilloskop:
> So yes, we should do a proper fix for all these affected devices.
Yes, for example I see a problem with communities that are still stuck with Gluon v2016.2.x and want to update it soon (like us).

```
Gluon: 2016.2.*:
root@GL-AR150:~# lua -e 'print(require("platform_info").get_image_name())'
gl-ar150
```
```
Gluon: 2018.2:
root@GL-AR150:~# lua -e 'print(require("platform_info").get_image_name())'
gl.inet-gl-ar150
```

But first step by step.

oszilloskop:
What is the better course of action?
Rather close this PR and process all affected  devices in one go?
Or let the developers who can test with their own devices process the devices individually?

blocktrron:
I would go for fix covering all affected boards. Can you include these in your PR?

oszilloskop:
> Can you include these in your PR?

I'm afraid I can't. It is my first PR. I am not trained in PR changes. :o)
I will create a new one and then close this one.

rotanid:
@oszilloskop please learn it, there will always be change requests.
did you do the changes with the github.com website or on your PC? the latter is the only useful way, then you can to your additional changes in the branch "oszilloskop-GL-AR150", and do "git commit -a --amend" to add all those changes to the existing commit. if you have further question it would be ideal if you join us on IRC, irc.hackint.org, channel #gluon

oszilloskop:
@rotanid 
I do it on Github, with a macOS-GUI and quite rarely inside a shell (only push and pull 😄 )
Because this PR and also the commit message is very close to the GL-AR150 but the newer one will bei for GL-AR150 and GL-AR300M (and possible GL-MiFi #1551 ) I'll do it the way I intended.

About IRC:
Its me, its Jason

oszilloskop:
Okay, this PR is ready for merge from my side.
- It does fix the profile names for GL.iNet gl-ar150 and gl-ar300m.
- The profile name for gl-ar750 is already changed in #1625
- A hint in gl-MiFi PR #1551 was set

**This PR does not use aliases for backwards compatibility!**

NeoRaider:
To allow upgrading form the old version to the new one, add a manifest alias for the old model name to the device sections, for example:
```
device gl.inet-gl-ar150 gl-ar150
manifest_alias gl-ar150
factory
```
(and as the device sections are getting longer, a blank line between the different devices wouldn't hurt)

NeoRaider:
And while you're at it, please also add an alias for the GL-AR750, as this was overlooked in #1625.

oszilloskop:
Done

mweinelt:
LGTM, can you please squash these commits into one?

oszilloskop:
Done

PR #1641, opened 2019-02-03
ramips-mt76x8: add support for TP-Link Archer C50 v4

blocktrron:
This backports the TP-Link Archer C50 v4.

We are dropping the following upstream commits. They add support for the
TP-Link recovery-flag which enabled the web-recovery. As they are not
needed for the router to work, we drop them for now.

28cd2ca base-files: sysupgrade: support additional mtd options
1e06482 mtd: add logic for TP-Link ramips recovery magic

----

  - [x] webinterface
  - [x] tftp
  - [ ] other: <specify>
    ```
    root@64xxx-0c80637f6a04:~# lua -e 'print(require("platform_info").get_image_name())'
    tp-link-archer-c50-v4
    ```
- wired network
- wifi (if applicable)
- led mapping
  - power/sys led
  - radio leds
    - [x] should show activity
  - switchport leds

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -384,6 +384,7 @@ ramips-mt76x8
 
   - TL-WR841N v13 [#80211s]_
   - Archer C50 v3 [#80211s]_
+  - Archer C50 v4 [#80211s]_
 
 * VoCore
 
--- a/patches/openwrt/0012-tools-mktplinkfw2-add-split-uboot-layout.patch
+++ b/patches/openwrt/0012-tools-mktplinkfw2-add-split-uboot-layout.patch
@@ -1,0 +1,26 @@
+From: David Bauer <mail@david-bauer.net>
+Date: Wed, 28 Nov 2018 23:56:31 +0100
+Subject: tools: mktplinkfw2: add split-uboot layout
+
+This commit adds the split-uboot partition layout used by the
+Archer C50 v4 to mktplinkfw2.
+
+Signed-off-by: David Bauer <mail@david-bauer.net>
+
+diff --git a/tools/firmware-utils/src/mktplinkfw2.c b/tools/firmware-utils/src/mktplinkfw2.c
+index dead49e7af8839bac5d1dee7445cf8921208c970..35db848bbab9f8315526dfa24ed351ad58878dde 100644
+--- a/tools/firmware-utils/src/mktplinkfw2.c
++++ b/tools/firmware-utils/src/mktplinkfw2.c
+@@ -146,6 +146,12 @@ static struct flash_layout layouts[] = {
+ 		.kernel_la	= 0x80000000,
+ 		.kernel_ep	= 0x80000000,
+ 		.rootfs_ofs	= 0x140000,
++	}, {
++		.id		= "8MSUmtk", /* Split U-Boot OS */
++		.fw_max_len	= 0x770000,
++		.kernel_la	= 0x80000000,
++		.kernel_ep	= 0x80000000,
++		.rootfs_ofs	= 0x140000,
+ 	}, {
+ 		.id		= "8MLmtk",
+ 		.fw_max_len	= 0x7b0000,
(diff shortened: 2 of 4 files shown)


mweinelt:
If the resulting image is not directly flashable, is it worth it carrying the patches? For example how would you concatenate two files on Windows?

blocktrron:
I can not give you a definitive answer on this. While you will need to fiddle yourself an image, the resulting binary can be flashed thru the Webinterface, avoiding TFTP.

I know this is a tradeoff, but it's an attractive option in the 25-30 Euro range IMHO.

kpanic23:
> For example how would you concatenate two files on Windows? 

`copy /b file1+file2 file`

rotanid:
please add the device to the list in docs/index.rst

blocktrron:
@rotanid added the device to the docs.

rubo77:
Isn't this as unstable as  my v3?

My node is crashing every few hours, if I use them as mesh-only node with a lot of other WiFis in the Air.

blocktrron:
@rubo77 Hardware-wise it is identical to the v3.

I have 3 mt76x8 based devices running here for the last weeks without major issues. See https://meshviewer.darmstadt.freifunk.net/#/de/map/b039560bc5ab

Are your builds based on the latest Gluon master?

rubo77:
Its this one: https://mesh.freifunk.in-kiel.de/#!v:m;n:b04e26454480 running in a deep mesh with bad uplink connectivity. built on gluon-v2018.2-26-gf9634ca0+ which is f9634ca0 with just an extra patch to build an extra target

rubo77:
still seems to crash, already only an uptime of 41m now

blocktrron:
collect serial logs, see whats causing it.

rotanid:
the linked device should run much longer than it did so far to get useful information.
i don't see any problems, yet, it didn't "reboot" again after your mentioned "reboot"/"crash".
how do you know it is a crash? does the firmware contain any kind of "bugfix-package" that resets the wifi or the device? if so, remove this package from your test-builds, please, not only for this device.
can you be sure the "reboot" wasn't a power-, user-, update-related one?

T-X:
Check /sys/kernel/debug/crashlog. If it's not there, then the device did not crash.

kevin-olbrich:
> Isn't this as unstable as my v3?

I had to deploy one of those and it has 100% uptime with both 2018.2 and 2018.2.1.
The customer said, it is working perfectly fine. (Batman IV, Tunneldigger)

I've just got the first v4 (3 devices) to be installed at another location.

rubo77:
I just built the master branch and there was no image in the factory folder, only in sysupgrade, is this indended?

blocktrron:
@rubo77 see the installation instructions for the v4 in the OpenWrt wiki. You'll need to build an image yourself due to legal issues.

rubo77:
ah ok: https://openwrt.org/toh/tp-link/archer-c50#archer_c50_v4_installation_instructions

>  **WARNING: Do NOT flash the squashfs-sysupgrade.bin to the C50 v4 via the TPlink firmware's web GUI. It will brick the router! **

shouldn't we put a txt file in the `factory` folder for such dangerous devices?

in the txt file we could provide such warning and a link to the install instructions

rotanid:
the sysupgrade file was never used for vendor web GUI flashing, so whoever tries this hasn't read any instructions...

blocktrron:
@rubo77 This is something which should be dealt with using device metadata instead of placing textfiles into image-output directories.

rubo77:
ok, sounds good, and where do we put those device metadata? where can the users find it?

Adorfer:
since i received support requests from users failing with their c50v4 in the past: it has to be made more obvious/clear for communities who just offer "raw image directories" of their prebuilt binaries. 
Dropping some index-file (plain ascii) directly in the directory with descriptions would help a lot for those.

PR #1642, opened 2019-02-04
ramips-mt7620: add support for ASUS RT-AC51U as broken

ambassador86:
This device is based on the MediaTek MT7620A with 580MHz, 16MB Flash and 64MB RAM. The radios are MediaTek MT7620A and MediaTek MT7610E. 
For now this device should remain as broken. The LED config is not correct and the 5GHz is not working. Test device: http://map.chemnitz.freifunk.net/#!v:m;n:708bcdea0a38

  * [x]  tftp
* wired network  
* wifi (if applicable)
  * [x]  association with AP must be possible on  2.4GHZ only
  * [x]  association with 802.11s mesh must be working on  2.4GHZ only
  * [x]  ap/mesh mode must work in parallel on 2.4GHZ only
* led mapping
  * power/sys led
  * radio leds
    * [ ]  should show activity
  * switchport leds
* [x]  primary mac should match address on device label (or packaging)

### DIFF
--- a/targets/ramips-mt7620
+++ b/targets/ramips-mt7620
@@ -1,3 +1,9 @@
+# ASUS
+if [ "$BROKEN" ]; then
+device asus-rt-ac51u rt-ac51u # BROKEN: no 5GHz usable, LEDs not fully supported
+factory
+fi
+
 # GL Innovations
 
 device gl-mt300a gl-mt300a


rotanid:
thanks. i've got a few questions:

- is someone already working on the LEDs and the 5ghz wifi?
- will the 5ghz wifi work, is there a chance?
- what about the primary mac?
- the "factory" keyword means to NOT generate a factory image, is this intentional?

achterin:
@rotanid 5ghz wifi will work once gluon swiches to openwrt 19.x. in openwrt 18.06.x only 2.4ghz support for this device is included. a factory image is not needed as the device is flashed via uboot recovery with the sysupgrade image. just check the [commit message](https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=5f932988c3b65af14bd6e48a05cc2204b7aeb419) for reference.

blocktrron:
Afaik the rt2800 driver lacks the necessary triggers for the 2.4GHz LEDs to work. In OpenWRT, LEDs are controlled via a netdev trigger on wlan0 which fails for us.

ambassador86:
The MAC adress looks fine. In my opinion this PR is ready now. The 5 GHz wifi will be availiable as soon as gluon uses a newer OpenWRT version and the LEDs are not fixable at the moment. Hopfully this is enough to integrate it as broken.

rotanid:
did you look whether the fixes in OpenWrt master can be backported?

could you please add the reason for the BROKEN flag to the file? look at other broken devices for examples on how to specify it

achterin:
yes it can be done but will be a huge mess. we would need several patches to backport 5ghz support and also need #1626 merged to get both interfaces up.

blocktrron:
@rotanid In addition the MT7610 driver does not support MAC-address patching yet. See https://github.com/openwrt/mt76/issues/219 and https://github.com/openwrt/mt76/pull/209

rotanid:
the question came up on IRC, but i guess your comments are a big NO.
the reason has to be added nevertheless.

targets/ramips-mt7620
```diff
@@ -1,3 +1,9 @@
+# Asus
```
mweinelt: The vendors name is ASUS (in caps).

rotanid:
@ambassador86 thanks. leave a comment after making changes, please, as we don't get notified otherwise.
merging it.

PR #1644, opened 2019-02-08
ath10k/ath10k-ct: support multicast and management rate control

ecsv:
Drivers with software rate control can directly use the selected multicast rate for multicast/broadcast frames and the minimal basic rate for management frames. But drivers with offloaded rate control algorithms must be informed about such upper layer decisions to configure the hardware/firmware.

A new BSS_CHANGED_MCAST_RATE is introduced in mac80211 to automatically inform all drivers. ath10k can detect this event and forward it via WMI to the driver. The already existing BSS_CHANGED_BASIC_RATES can be used to select the management rate.

Without the WMI commands, a low rate (not necessarily one from the basic rates) is used for bcast/mcast/management frames. This means that the /etc/config/wireless settings basic_rate and mcast_rate would have no effect on the rates selected by this driver for the mentioned frames.

----

The ath10k-ct and mac80211 packages in OpenWrt master already support the broadcast/multicast interface (because they are using ath10k from 4.19 as their base). The management frame changes where part of 4.20 and are not yet part of OpenWrt master. But I have forwarded backports of these changes as OpenWrt PR:

* https://github.com/openwrt/openwrt/pull/1801
* https://github.com/openwrt/openwrt/pull/1802

### DIFF
--- a/patches/openwrt/0014-mac80211-ath10k-support-for-multicast-and-management-rate-control.patch
+++ b/patches/openwrt/0014-mac80211-ath10k-support-for-multicast-and-management-rate-control.patch
@@ -1,0 +1,420 @@
+From: Sven Eckelmann <sven@narfation.org>
+Date: Fri, 8 Feb 2019 21:20:20 +0100
+Subject: mac80211: ath10k: support for multicast and management rate control
+
+Drivers with software rate control can directly use the selected multicast
+rate for multicast/broadcast frames and the minimal basic rate for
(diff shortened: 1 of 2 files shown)


mweinelt:
Thank you.

blocktrron:
@ecsv This might come late - but I've just discovered, that management rate doesn't seem to be set for at least IPQ4019. It still sends out beacons at 1Mbit/s when configuring a 6Mbit/s basic rate.

Can you reproduce this or am i just holding it wrong?

PR #1646, opened 2019-02-11
docs: missing uci commit in gluon-reconfigure example

belzebub40k:
The example to switch the domain via commandline doesn't commit the uci changes and a node would fall back to its previous domain after a reboot.

### DIFF
--- a/docs/features/multidomain.rst
+++ b/docs/features/multidomain.rst
@@ -93,6 +93,7 @@ Switching the domain
 ::
 
     uci set gluon.core.domain="newdomaincode"
+    uci commit gluon.core.domain
     gluon-reconfigure
     reboot
 


NeoRaider:
This doesn't make sense, the last upgrade script 998-commit run by gluon-reconfigure commits everything.

mweinelt:
@NeoRaider Weird, I thought so too but missed it while quickly grepping `/lib/gluon/upgrade`. In that case let's revert the change, as 998-commit is the better way to persist the domain after the change actually happened. Agreed?

NeoRaider:
Agreed.

belzebub40k:
Sorry also missed that. I only checked `package/gluon-core/luasrc/...` for upgrade scripts but not `package/gluon-core/files/...`

Issue #116, opened 2014-06-13
Auto-change SSID when router has no Internet connection


rubo77:
To inform the user, that the node has no internet connnection, the SSID could automatically change to "freifunk (offline)" while the router has no net and it is changed back, when the internetconnection is back.

This would solve a lot of struggling for the clients.

RubenKelevra:
We use a concrete implementation for this, parsing and using batctl to check the (selected) gateway connectivity. We use this implementation more than a year, but want to refactor it, before giving it to everyone.

If somebody is interested in the crappy bash-code, take a look at https://github.com/FF-NRW/offline-ssid/

Yes it throws out all clients which are connected currently, but all clients which knows the new ssid, will connect again. Else this gives the ability to select two different wireless-profiles, for both ssids. I think no userfriendly system currently leaves an wifi-connection open, with link-local, if the internet-connection aren't working.

So this gives the user the choice to add an offline-freifunk also to it's known networks.

Else there is an batman-adv issue: When one gateway goes down, batman-adv stick on it, till a timeout of 300 seconds occour. In this time, no new client can use the node, although other gateways work. Now my script detects this, in seconds, and change the ssid. Taking this node from the online-ssid, and gives the client the ability to select another node in range, with online-ssid.

A similar issue occours at boot-time, it takes some seconds to establish the fastd-connection and/or get batman to select an gateway. In this time, no client can successfully connect, so our nodes boot with a 'boot-ssid' and change it to online or offline, after this.

rubo77:
@RubenKelevra: Great! Any negative side-effects?

RubenKelevra:
@rubo77 No.

ohrensessel:
Apart from clients beeing kicked out of the wlan. Or did you solve that?
Am 06.04.2015 3:45 nachm. schrieb "@RubenKelevra" <notifications@github.com

RubenKelevra:
They are not kicked, the connections are graceful ended. Since the SSID change this behavior is excpected. 

Else this is the only thing which makes sense here:
If the clients have the other ssid in there profiles, they can reconnect. Else the clients search for a router which is still "online".

rubo77:
Sounds great so far, what could be a reason not to implement this?

ohrensessel:
This behaviour is the exact opposite to how I understand the Freifunk idea.
Internet (and therefore gateway access) is only a service within the
network. Even in a local mesh without any mesh-vpn uplink communication
[...]

PR #1649, opened 2019-02-14
add package gluon-offline-ssid to automatically change the SSID on no Internet connection

rubo77:
The package is thoroughly tested as an external package in various domains already.

Details see [gluon-ssid-changer.rst](https://github.com/rubo77/gluon/blob/ssid-changer/docs/package/gluon-ssid-changer.rst)

source: https://github.com/Freifunk-Nord/gluon-ssid-changer

solves #116


ambassador86:
Hello @rubo77,
I really like this functionality, one thing that came in my mind when looking through the documentation: Does it work with BATMAN_V? It has no TQ. Instead it will use the speed. I assume turning of the TQ check will also work with BATMAN_V.

rubo77:
@ambassador86: you are right, I added a hint to the Readme

CodeFetch:
Thank you for this. It's really annoying being connected to a node that has no gateway connectivity.

Could you make the check of whether a gateway is reachable independent from the package itself?
Now that we support different mesh protocols soon it would be nice if you could trigger an external script e.g. provided by gluon-mesh-batman-adv to test the availability of a gateway connection. You could let your package depend on a meta-package e.g. "gluon-gateway-status" and let gluon-mesh-batman-adv provide the meta-package.
Do you need uci configurable parameters or could you directly read the site.conf instead?

T-X:
In https://github.com/freifunk-gluon/gluon/pull/1555#discussion_r244013492 it was decided to check for connectivity via IP addresses provided in the site.conf.

So either add a connection_check_targets to ssid_changer, too, or move the connection_check_targets out of domain_switch and reuse it?

CodeFetch:
@T-X I'd prefer your second suggestion, as we don't want to introduce redundancy even if it is something as simple as a connection check (which might become more complex over time). I'd suggest to avoid having switch-cases like "if batman else if babel" in packages like the status page, but instead providing own packages for such functionalities which "provide" a specific functionality. That way one does not need to make changes in X packages when e.g. a new meshing protocol comes up. I know that's a bad example, but what I mean is to keep things more modular and interchangeable. Some communities might want to define a connection check only to be successful if there is a specific bandwidth available or if the DNS is also working etc.

rotanid:
we also use an ssid-changer, but this alone doesnt mean that i think it should be available upstream.
you could for example add a reason why your implementation is better than other implementations.

and i already know which criticism others will have:

- no bash please, use lua or C ( see https://github.com/Freifunk-Nord/gluon-ssid-changer/issues/12 )
- too many options/settings maybe

rubo77:
We optimized a lot on the code in Kiel so it works stable and is not interfering with the AU for example. It must not be the best solution, but we should add _something_ to the gluon base at least.

Great ideas already, thanks!

We will accept PRs at https://github.com/Freifunk-Nord/gluon-ssid-changer to enhance the code and implement outsourcing of the gateway-check.

I hope we make the best out of it together ;)

mweinelt:
> We will accept PRs at https://github.com/Freifunk-Nord/gluon-ssid-changer to enhance the code and implement outsourcing of the gateway-check.

If you're not willing to work on the pull request this is essentially a dead end for merging it here and I'm not willing to do a review in that case.

rubo77:
I posted the wrong url. I meant the underlying branch for this PR here: 
https://github.com/rubo77/gluon/tree/ssid-changer

sure. I am willing to work on this. just thought, sometimes it is easy for someone to add a PR for his thoughts, than explaining which file should go where, and so on...

rotanid:
ok, that sounds better.
as i already wrote, we won't accept this with bash code, as we have this policy for a while now.
i know that not everyone can code in lua, me neither.

regarding the settings, maybe add a list of every configurable thing in your first comment of this PR together with the default value.
afterwards we can discuss if every setting is really necessary and if the suggested defaults are OK.
i think this package should work with defaults only, too, that's why the defaults should be sane.

T-X:
Out of curiosity, does anyone have a number regarding how many communities are using an essid changer?

Also, which other ssid changers are out there? Would be interesting to compare. Maybe some have nice ideas that might be worth adopting.

NeoRaider:
One big obstacle in the implementation of SSID changers is that they have to modify UCI config at runtime, which we've tried to avoid in Gluon so far. I'm not sure how realistic the addition of dynamic VIF management via ubus to netifd would be (or if I can bring to fruition my long-term plan to remove netifd and UCI from Gluon's network config altogether first).

rubo77:
This solution doesn't need UCI

docs/package/gluon-ssid-changer.md
```diff
@@ -0,0 +1,70 @@
...
+no connection to any gateway.
+This SSID can be generated from the nodes hostname with the first
+and last part of the nodename or the mac address, to allow observers to 
+recognise which node is down. The script is called once a minute by micron.d
+and it will change from online to offline-SSID maximum once every (definable)
+timeframe.
+
+You can enable/disable it in the config mode.
+
+Once every timeframe it checks if there's still a gateway reachable. There can be
```
CodeFetch: It checks if a gateway is reachable once in an interval. Different algorithms can be selected to determine whether a gateway is reachable.

docs/package/gluon-ssid-changer.md
```diff
@@ -0,0 +1,70 @@
...
+selected two algorithms to analyze if a gateway is reacheable:
+
+- `tq_limit_enabled=true`: (not working with BATMAN_V) define an upper and lower
+  limit to turn the offline_ssid on and off in-between these two values the SSID
```
CodeFetch: lower bound to toggle the offline_ssid.

docs/package/gluon-ssid-changer.md
```diff
@@ -0,0 +1,70 @@
...
+  will never be changed to prevent it from toggeling every minute.
+- `tq_limit_enabled=false`: there will be only checked, if the gateway is
```
CodeFetch: ?

docs/package/gluon-ssid-changer.md
```diff
@@ -0,0 +1,70 @@
...
+  reachable with:
+
+        batctl gwl -H
+
+Depending on the connectivity, it will be decided if a change of the SSID is 
+necessary: There is a variable `switch_timeframe` (for ex.  1440 = 24h) that
```
CodeFetch: Leave this out "Depending on the connectivity, it will be decided if a change of the SSID is 
necessary"

docs/package/gluon-ssid-changer.md
```diff
@@ -0,0 +1,70 @@
...
+  will never be changed to prevent it from toggeling every minute.
+- `tq_limit_enabled=false`: there will be only checked, if the gateway is
+  reachable with:
+
+        batctl gwl -H
+
+Depending on the connectivity, it will be decided if a change of the SSID is 
+necessary: There is a variable `switch_timeframe` (for ex.  1440 = 24h) that 
+defines a time interval after which a successful check that detects an offline
+state will result in a single change of the SSID to "FF_Offline_$node_hostname".
```
CodeFetch: The parameter `switch_timeframe` defines the period of time after which gateways are considered unreachable resulting in the change of the SSID to "FF_Offline_$node_hostname" (in minutes).

docs/package/gluon-ssid-changer.md
```diff
@@ -0,0 +1,70 @@
...
+Only the first few minutes (also definable in a variable `first`) the 
+OFFLINE_SSID may also be set. All other minutes a checks will just be counted
+and reported in the log and whenever an online state is detected the SSID will
+be set back immediately back to normal. when the timeframe is reached, there
```
CodeFetch: Too complicated

docs/package/gluon-ssid-changer.md
```diff
@@ -0,0 +1,70 @@
+gluon-ssid-changer
+==================
+
+This package adds a script to change the SSID to an Offline-SSID when there is
```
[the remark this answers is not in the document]
rubo77: not really, because it only changes from online to offline every timeframe, but could change back from offline to online every minute.

but i will rephraze this to be clearer.

docs/package/gluon-ssid-changer.md
rubo77: This is not how it works.

the timeframe is just for measurement. during that timeframe, every minute it is recorded, if the gateway is reacheable or not. If half of the checks result in not-reacheable, the SSID will be changed to the Offline SSID

CodeFetch: I should have looked at the code first. Maybe I still misunderstand how it works. If you have a 24 hour switch_timeframe and the first 12 hours every check is successful, doesn't it take another 12 hours of failure until the failure state is recognized and vice versa?

rubo77: yes, as you wish. we recomment a 60 minutes timeframe, so nodes don't go into offline mode too lightly

rubo77: I renamed to rest, and rephrazed the whole block.

rubo77:
> @T-X : In [#1555 (comment)](https://github.com/freifunk-gluon/gluon/pull/1555#discussion_r244013492) it was decided to check for connectivity via IP addresses provided in the site.conf.


maybe we could use this file then? https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-check-connection

rubo77:
> maybe add a list of every configurable thing in your first comment of this PR together with the default value.
> afterwards we can discuss if every setting is really necessary and if the suggested defaults are OK.
> i think this package should work with defaults only, too, that's why the defaults should be sane.

The config values and defaults are visible in [lib/gluon/upgrade/500-ssid-changer](https://github.com/rubo77/gluon/blob/ssid-changer/package/gluon-ssid-changer/luasrc/lib/gluon/upgrade/500-ssid-changer#L14)

package/gluon-ssid-changer/luasrc/lib/gluon/upgrade/500-ssid-changer
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/lua
+
+local site = require 'gluon.site'
+
+local uci = require('simple-uci').cursor()
+
+if site.ssid_changer == nil then
+	-- print('ssid_changer not defined in site.conf')
```
rotanid: obsolete code?

package/gluon-ssid-changer/luasrc/lib/gluon/config-mode/model/admin/ssid-changer.lua
```diff
@@ -0,0 +1,41 @@
+local uci = require('simple-uci').cursor()
+local util = require 'gluon.util'
+
+local pkg_i18n = i18n 'gluon-ssid-changer'
+
+local f = Form(pkg_i18n.translate('Offline-SSID'))
+
+local s = f:section(Section, nil, pkg_i18n.translate(
+	'Here you can enable to automatically change the SSID to the Offline-SSID '
+  .. 'when the node has no connection to the selected Gateway.'
```
rotanid: indentation wrong maybe?

package/gluon-ssid-changer/luasrc/lib/gluon/config-mode/model/admin/ssid-changer.lua
```diff
@@ -0,0 +1,41 @@
...
+))
+
+local enabled = s:option(Flag, 'enabled', pkg_i18n.translate('Enabled'))
+enabled.default = uci:get_bool('ssid-changer', 'settings', 'enabled')
+
+-- local prefix = s:option(Value, 'prefix', pkg_i18n.translate('First part of the Offline SSID'))
```
rotanid: many commented out lines in this file?

package/gluon-ssid-changer/luasrc/lib/gluon/config-mode/model/admin/ssid-changer.lua
```diff
@@ -0,0 +1,41 @@
...
+			-- prefix = prefix.data or 'FF_Offline_'
+			-- suffix = suffix.data or 'nodename'
+			-- tq_limit_min = tq_limit_min.data or '45'
+			-- tq_limit_enabled = tq_limit_enabled.data or '0'
+		})
+	else
+		uci:set('ssid-changer', 'settings', 'enabled', '0')
+	end
+
+	uci:commit('ssid-changer')
```
rotanid: i'm not sure whether commit is correct here, maybe there's a central commit already?

package/gluon-ssid-changer/Makefile
```diff
@@ -0,0 +1,36 @@
...
+include $(TOPDIR)/../package/gluon.mk
+
+define Package/$(PKG_NAME)
+  TITLE:=changes the SSID to an Offline-SSID so clients don't connect to an offline WiFi
+  DEPENDS:=+gluon-core +micrond
+endef
+
+define Package/$(PKG_NAME)/description
+  Script to change the SSID to an Offline-SSID when there is no connection to
+  any gateway. This SSID can be generated from the nodes hostname with the first
```
rotanid: node -> node's

package/gluon-ssid-changer/Makefile
```diff
@@ -0,0 +1,36 @@
...
+  and last part of the nodename or the mac address, to allow observers to 
+  recognise which node is down. The script is called once a minute by micron.d
```
rotanid: micrond

package/gluon-ssid-changer/luasrc/lib/gluon/config-mode/model/admin/ssid-changer.lua
rubo77: Comments are deleted by gluonSourceDiet.

I wasnt sure, if we need those lines here again maybe?

package/gluon-ssid-changer/Makefile
```diff
@@ -0,0 +1,36 @@
...
+  and it will change from online to offline-SSID maximum once every (definable)
```
rotanid: change -> switch
here offline is written in small letters, everywhere else with a big "O".
"maximum once" doesn't seem to be correct english, maybe... "once per definable timeframe at most."

package/gluon-ssid-changer/luasrc/lib/gluon/config-mode/model/admin/ssid-changer.lua
rotanid: having a comment with information is not the same as having code that isn't used.
i can't say if you need this again, but if you do you should uncomment it ;)

mweinelt:
The name `gluon-ssid-changer` would leave some room for interpretation, while a name like `gluon-offline-ssid` would convey the packages functionality much better.

Also if the package is specific to Gluon please save UCI data below `gluon.<pkgname>`.

rotanid:
i agree with your first two ideas, but not the third one.
to be able to debug it is necessary to be able to connect, and the SSID containing the nodename shows clearly which node in your current area has a problem, which is helpful for debugging.

package/gluon-ssid-changer/luasrc/lib/gluon/config-mode/model/admin/ssid-changer.lua
rubo77: I removed the code for now. If we want to make those values configurable later too, we could add the code from git history in the Freifunk Nord branch

package/gluon-ssid-changer/luasrc/lib/gluon/config-mode/model/admin/ssid-changer.lua
rubo77: I am not sure either. ... are there some experts here sure? ;)

rubo77:
> Also if the package is specific to Gluon please save UCI data below `gluon.<pkgname>`.

If we change the storage point in the config, should we also add some lines that clean up the stored variables by the original Freifunk-Nord ssid-changer?

rotanid:
> If we change the storage point in the config, should we also add some lines that clean up the stored variables by the original Freifunk-Nord ssid-changer?
Gluon can't take care of community-specific packages...

package/gluon-ssid-changer/luasrc/lib/gluon/config-mode/model/admin/ssid-changer.lua
rotanid: just compare with other config mode packages and afterwards try it out.

rotanid:
things i didn't review yet: shell code (has to be replaced anyway), README

package/gluon-ssid-changer/luasrc/lib/gluon/config-mode/model/admin/ssid-changer.lua
```diff
@@ -0,0 +1,28 @@
...
+	'Here you can enable to automatically change the SSID to the Offline-SSID '
+	.. 'when the node has no connection to the selected Gateway.'
+))
+
+local enabled = s:option(Flag, 'enabled', pkg_i18n.translate('Enabled'))
+enabled.default = uci:get_bool('ssid-changer', 'settings', 'enabled')
+
+function f:write()
+	if enabled.data then
+		uci:section('ssid-changer', 'settings', 'settings', {
```
blocktrron: No need to create a section here as it is created by `500-ssid-changer` in any case.

package/gluon-ssid-changer/luasrc/lib/gluon/upgrade/500-ssid-changer
```diff
@@ -0,0 +1,21 @@
...
+
+	uci:section('ssid-changer', 'settings', 'settings', {
+		enabled           = uci:get('ssid-changer', 'settings', 'enabled') or site_enabled,
+		switch_timeframe  = site.ssid_changer.switch_timeframe() or '30',
+		first             = site.ssid_changer.first() or '5',
+		prefix            = site.ssid_changer.prefix() or 'FF_Offline_',
+		suffix            = site.ssid_changer.suffix() or 'nodename',
+		tq_limit_enabled  = site.ssid_changer.tq_limit_enabled() or false,
+		tq_limit_max      = site.ssid_changer.tq_limit_max() or 45,
+		tq_limit_min      = site.ssid_changer.tq_limit_min() or 35,
```
blocktrron: I would go for required site.conf entries instead of magic default values.

package/gluon-ssid-changer/check_site.lua
```diff
@@ -0,0 +1,9 @@
+need_boolean({'ssid_changer', 'enabled'}, false)
+need_number({'ssid_changer', 'switch_timeframe'}, false)
+need_number({'ssid_changer', 'first'}, false)
+need_string({'ssid_changer', 'prefix'}, false)
+need_one_of({'ssid_changer', 'suffix'}, {'nodename', 'mac', 'none'}, false)
+if need_boolean({'ssid_changer','tq_limit_enabled'}, false) then
```
blocktrron: In case we switch to required site.conf variables, the `tq_limit_enabled` could be dropped and implicitly assumed by a dictionary containing `tq_limit_max` and `tq_limit_min` (see example below)

```
ssid_changer = {
  ..
  tq_limit = {
    max = xx,
    min = xx,
  },
  ..
},
```

package/gluon-ssid-changer/luasrc/lib/gluon/config-mode/model/admin/ssid-changer.lua
blocktrron: should be correct as each module in the "Advanced Setup" tab takes care of persisting it's state itself.

package/gluon-ssid-changer/luasrc/lib/gluon/upgrade/500-ssid-changer
rotanid: i wrote the complete opposite, that it should work with defaults only :D

mweinelt:
> Another thought is, that if the focus of your mesh network is Internet access, why not disable the wireless client network or make the network hidden? Those could be other useful actions this package could take care of.

To be clear, I thought of these actions as other actions the package could offer instead of the just the offline SSID. But obviously that depends on someone having interest in these actions.

rubo77:
I want to rewrite the shell script in lua now. I think the script should go into `luasrc/usr/bin/change-ssid`

Is there an easy way do develop this on my laptop? or do I have to develop on a node all the time? I have installed lua locally, but if I start any lua script from the gluon source, my mashine always lacks the required libraries: `local unistd = require 'posix.unistd'; 
local util = require 'gluon.util' ...`

rubo77:
I had some time today to transcode the shell script into lua: https://github.com/freifunk-gluon/gluon/pull/1649/commits/6c559ff4102c1f3a83b4357266a38c867636d9bb

It is not in a working state, but I need please some fedback, if I am on the right way, then I will continue to make this run in Lua

package/gluon-ssid-changer/check_site.lua
rubo77: i'd rather like to have `tq_limit_enabled` as required and `tq_limit_min` and max optional with sound default values

rubo77:
> The name `gluon-ssid-changer` would leave some room for interpretation, while a name like `gluon-offline-ssid` would convey the packages functionality much better.
> 
> Also if the package is specific to Gluon please save UCI data below `gluon.<pkgname>`.

done

package/gluon-ssid-changer/luasrc/lib/gluon/config-mode/model/admin/ssid-changer.lua
rubo77: OK, I will change it to just 

    		uci:set('offline-ssid', 'settings', 'enabled', '1')

CodeFetch:
@rubo You are right. gluon-node-info, gluon-setup-mode etc. have their own config files. I think this is the correct way as OpenWrt packages like fastd do it alike. Another point are sysupgrade files... They execute uci commit xxx and shouldn't interfere with each other.

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
```diff
@@ -0,0 +1,233 @@
...
+      return s .. '} '
+   else
+      return tostring(o)
+   end
+end
+dump(phy); os.exit()
+
+-- only once every timeframe minutes the SSID will change to the Offline-SSID
+-- (set to 1 minute to change immediately every time the router gets offline)
+local MINUTES = uci:get('offline-ssid', 'settings', 'switch_timeframe') or 30
```
CodeFetch: You don't need to use capital letters for variable names and the config file should be called gluon-offline-ssid.

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
```diff
@@ -0,0 +1,233 @@
...
+
+function shell_exec(command)
+  local file = assert(io.popen(command, 'r'))
+  local output = file:read('*all')
+  file:close()
+  return output
+end
+
+function file_exists(name)
+   local f=io.open(name,"r")
```
CodeFetch: Use spaces before and after equal sign and after a comma

CodeFetch: I'm not sure, but I think you can write return assert(io.popen(command, 'r')):read('*a') instead in one line.

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
```diff
@@ -0,0 +1,233 @@
...
+  end
+else
+  -- 'none'
+  local SUFFIX=''
+end
+local OFFLINE_SSID=PREFIX .. SUFFIX
+
+-- get all SSIDs (replace \' with TICX and back to keep a possible tic in an SSID)
+ONLINE_SSIDs = shell_exec('uci show | grep wireless.client_radio[0-9]\.'
+  .. '| grep ssid | awk -F \'=\'  \'{print $2}\'') .. " "  or "~FREIFUNK~"
```
CodeFetch: In Lua please and without hardcoded FREIFUNK
(use uci:foreach)

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
```diff
@@ -0,0 +1,233 @@
...
+local TQ_LIMIT_ENABLED = uci:get('offline-ssid', 'settings', 'tq_limit_enabled') or '0'
+
+local CHECK
+if ( TQ_LIMIT_ENABLED == 1 ) then
+  --  upper limit, above that the online SSID will be used
+  local TQ_LIMIT_MAX = uci:get('offline-ssid', 'settings', 'tq_limit_max') or '45'
+  --  lower limit, below that the offline SSID will be used
+  local TQ_LIMIT_MIN = uci:get('offline-ssid', 'settings', 'tq_limit_min') or '35'
+  -- grep the connection quality of the currently used gateway
+  local GATEWAY_TQ=shell_exec('batctl gwl | grep -e "^=>" -e "^\*" | awk -F \'[()]\' \'{print $2}\' | tr -d " "')
```
CodeFetch: Please also write this in Lua. Maybe I've already coded that in Lua. I can look for my script and give it to you tomorrow. Before merging this, we should put all the connection checks into a separate package and let this package depend on it.

rubo77:
I added a commit of the current state

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
```diff
@@ -0,0 +1,241 @@
...
+        else
+          shell_exec('logger -s -t "gluon-offline-ssid" -p 5 "could not set to offline state: did neither endnd SSID ' .. ONLINE_SSID .. ' nor ' .. OFFLINE_SSID .. '. Please reboot"')
+        end
+      end
+    end
+    -- else print("Minute ' .. M .. ', just count ' .. OFF_COUNT .. '")
+  end
+  
+  local f = io.open(TMP, 'w+')
+  f:write(OFF_COUNT + 1)
```
CodeFetch: io.popen(tmp, 'w'):write(off_count + 1)
The capital variable names look pretty in shell, but not in Lua.

Furthermore your config file is /etc/config/gluon-offline-ssid. Thus you have to use uci get gluon-offline-ssid instead of uci get offline-ssid.

rubo77:
I added the first running lua version! :)

still a lot of bash_exec() calls, but that can be minimized later

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
rubo77: 1. decapitalization done ;)

2. not changed: other packages don't call their config files gluon-* either

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
rubo77: rewritten

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
rubo77: only `w`? without `+`?

CodeFetch:
@rubo77 I don't write the comments for fun. Other packages name their config files gluon-* and btw your package does it, too. Your config file is /etc/config/gluon-offline-ssid and thus you need to use uci get gluon-offline-ssid, because your package does not create a /etc/config/offline-ssid file.
Look at the /etc/config directory of a Gluon node. You'll find e.g. gluon-node-info, gluon-setup-mode, gluon-wan-dnsmasq etc.

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
CodeFetch: You don't close the file, but assign a file pointer, you don't use assert, you open the file in read-write mode instead of write-only mode while never reading the file.

rubo77: yes, seems to work with only `w`

rubo77:
> Your config file is /etc/config/gluon-offline-ssid and thus you need to use uci get gluon-offline-ssid

you're right. fixed

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
rubo77: 2. also fixed

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
```diff
@@ -0,0 +1,241 @@
+#!/usr/bin/lua
+
+local uci = require("simple-uci").cursor()
+local util = require 'gluon.util'
+local site = require 'gluon.site'
+
+-- todo:
+-- pgrep -f autoupdater >/dev/null && safety_exit 'autoupdater running'
```
rotanid: look for something like this in other packages, be it upstream or not

CodeFetch:
@rubo77 This is the connection check of ffho-autoupdater-wifi-fallback
```
local function connectivity_check()
  local f = io.open('/sys/kernel/debug/batman_adv/bat0/gateways', 'r')
  if f then
    for line in f:lines() do
      local gateway_mac = line:match('^=?>? +([0-9a-f:]+)')
      if gateway_mac then
        if os.execute('batctl ping -t5 -c1 ' .. gateway_mac .. ' > /dev/null 2>&1') == 0 then
          return true
        end
      end
    end
    f:close()
end
```

Maybe you can do it similarly.

rotanid:
> Maybe you can do it similarly.

it should be done like other already included packages, as was already written by someone else.

CodeFetch:
@rotanid Do you mean the domain switcher?
Maybe you know what the problem is:
`pgrep -f autoupdater >/dev/null && safety_exit 'autoupdater running'`

should be equal to

 if os.execute('pgrep -f autoupdater > /dev/null 2>&1') == 0 then
      safety_exit("autoupdater running")
end

But actually os.execute always returns 0

rubo77:
I could use my `shell_exec` function with parameter true.

But I thought the AU already created a pidfile, I could just check. So I don't need a pgrep call then?

CodeFetch:
@rubo77 Why do you need to check if the autoupdater is running at all?

rubo77:
I don't remember why, but it is fixed now anyway :) 

And I fixed some other bugs, now it seems to work all correct.

I built a firmware with it so the Makefile is ok too

package/gluon-offline-ssid/luasrc/lib/gluon/config-mode/model/admin/offline-ssid.lua
```diff
@@ -0,0 +1,26 @@
+local uci = require('simple-uci').cursor()
```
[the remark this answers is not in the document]
rubo77: what do you mean?

package/gluon-offline-ssid/luasrc/lib/gluon/upgrade/500-gluon-offline-ssid
```diff
@@ -0,0 +1,21 @@
...
+local uci = require('simple-uci').cursor()
+
+if site.offline_ssid ~= nil then
+	local site_enabled = site.offline_ssid.enabled() or '1'
+
+	uci:section('offline-ssid', 'settings', 'settings', {
+		enabled           = uci:get('gluon-offline-ssid', 'settings', 'enabled') or site_enabled,
+		switch_timeframe  = site.offline_ssid.switch_timeframe() or '30',
+		first             = site.offline_ssid.first() or '5',
+		prefix            = site.offline_ssid.prefix() or 'FF_Offline_',
```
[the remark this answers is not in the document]
rubo77: Do you want to change the default prefix?

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
```diff
@@ -0,0 +1,247 @@
...
+local first = tonumber(uci:get('gluon-offline-ssid', 'settings', 'first') or '5')
+
+-- the offline-ssid will start with this prefix use something short to leave space for the nodename
+-- (no '~' allowed!)
+local prefix = uci:get('gluon-offline-ssid', 'settings', 'prefix') or 'FF_Offline_'
+
+local disabled = uci:get('gluon-offline-ssid', 'settings', 'enabled') == '0' or false
+if disabled then 
+  print("offline-ssid is disabled")
+end
```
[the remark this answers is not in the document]
rubo77: no, if you disable it on a running node, the script has to continue running to set it back to the online-ssid

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
```diff
@@ -0,0 +1,247 @@
...
+  if ( string.len(suffix) > 30 - string.len(prefix) ) then
+    -- calculate the length of the first part of the node identifier in the offline-ssid
+    local half = math.floor((28 - string.len(prefix) ) / 2)
+    -- jump to this charakter for the last part of the name
+    local skip = string.len(suffix) - half
+    -- use the first and last part of the nodename for nodes with long name
+    suffix = string.sub(suffix,0,half) .. '...' .. string.sub(suffix, skip)
+  end
+elseif settings_suffix == 'mac' then
+  suffix = uci:get('wireless', 'client_radio0', 'macaddr')
```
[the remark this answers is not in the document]
rubo77: is this really needed? In this case the mac is only needed for the ssid-name. I think short code is more important, than that it is 10000% really the identical MAC.

package/gluon-offline-ssid/luasrc/lib/gluon/config-mode/model/admin/offline-ssid.lua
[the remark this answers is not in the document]
CodeFetch: All other packages have consistent naming for these files without "gluon-".
Another thing is that you should create two different packages for that. One gluon-offline-ssid for the core functionality and another one gluon-web-offline-ssid that depends on gluon-offline-ssid for the config-mode.

package/gluon-offline-ssid/luasrc/lib/gluon/upgrade/500-gluon-offline-ssid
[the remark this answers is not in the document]
CodeFetch: Yes, because Gluon is not intended to be used by Freifunk only (in a distant future we will hopefully merge LibreMesh into it).

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
```diff
@@ -0,0 +1,247 @@
...
+-- (set to 1 minute if you want to change immediately every time the router gets offline)
+local minutes = tonumber(uci:get('gluon-offline-ssid', 'settings', 'switch_timeframe') or '30')
+
+-- the first few minutes directly after reboot within which an offline-ssid always may be activated
+-- (must be <= switch_timeframe)
+local first = tonumber(uci:get('gluon-offline-ssid', 'settings', 'first') or '5')
+
+-- the offline-ssid will start with this prefix use something short to leave space for the nodename
+-- (no '~' allowed!)
+local prefix = uci:get('gluon-offline-ssid', 'settings', 'prefix') or 'FF_Offline_'
```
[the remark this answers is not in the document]
rubo77: But what if someone fiddles with the UCI config so a value is not defined anymore? this will make sure, that a default value is still be used

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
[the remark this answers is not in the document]
CodeFetch: Then exit after you've checked, that the online-SSID is set.

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
```diff
@@ -0,0 +1,247 @@
...
+
+local disabled = uci:get('gluon-offline-ssid', 'settings', 'enabled') == '0' or false
```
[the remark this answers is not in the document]
rubo77: sure? what if the uci call fails? it will be `nil` then, is this `~='0'`?

package/gluon-offline-ssid/files/etc/config/gluon-offline-ssid
```diff
@@ -0,0 +1 @@
+config settings 'settings'
```
rubo77: Is this OK? I wonder if `main settings` would be better?

original issue here: https://github.com/Freifunk-Nord/gluon-ssid-changer/issues/17

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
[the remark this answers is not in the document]
rubo77: it kind of does, if you follow the code: but I can optimize to break before setting `current_ssid and 
offline_ssid_in_phy`

package/gluon-offline-ssid/files/etc/config/gluon-offline-ssid
CodeFetch: You could set it to config offline-ssid 'settings' as e.g. the autoupdater does I think

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
```diff
@@ -0,0 +1,247 @@
...
+    -- else print("minute ' .. m .. ', just count ' .. off_count .. '")
+  end
+  
+  assert(io.open(tmp, 'w')):write(off_count + 1)
+end
+
+if hup_needed == 1 then
+  -- send hup to all hostapd to load the new ssid
+  shell_exec('killall -hup hostapd')
+  hup_needed = 0
```
CodeFetch: I think you can remove hup_needed = 0 at this point

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
```diff
@@ -0,0 +1,247 @@
...
+          break
+        end
+        if online_ssid_in_phy then
+          -- set offline
+          shell_exec('logger -s -t "gluon-offline-ssid" -p 5 ' .. msg .. '"' .. off_count .. ' times offline, ssid is ' .. current_ssid .. ', change to ' .. offline_ssid .. '"')
+          shell_exec('sed -i "s~^ssid=' .. ssid.ssid .. '~ssid=' .. offline_ssid .. '~" ' .. hostapd)
+          hup_needed = 1
+        else
+          shell_exec('logger -s -t "gluon-offline-ssid" -p 5 "could not set to offline state: did neither find ssid ' .. ssid.ssid .. ' nor ' .. offline_ssid .. '. please reboot"')
+        end
```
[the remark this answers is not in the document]
rubo77: fixed, and I think so too: 

> Maybe we should provide this functionality in gluon-util as I've seen it in quite a number of packages now.

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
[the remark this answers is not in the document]
rubo77: I enhanced the flow (see last commit)

package/gluon-offline-ssid/luasrc/lib/gluon/upgrade/500-gluon-offline-ssid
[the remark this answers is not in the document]
rubo77: I'd suggest just "Offline_" then

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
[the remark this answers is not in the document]
CodeFetch: Yes, that's what I meant. Remove them from the sysupgrade file then. Or does this break something?

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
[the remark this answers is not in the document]
CodeFetch: You're right. Sorry... Use a variable called "enabled" instead and in the line after if not enabled.

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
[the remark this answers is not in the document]
CodeFetch: Sorry again. What I really meant was to use the primary MAC.
And it's not gluon-util, but sysconfig...
```
sysconfig.primary_mac
```
You can see the MAC address of the client interface using e.g. iwinfo, but the info you may need is the primary MAC on the sticker/map.

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
```diff
@@ -0,0 +1,247 @@
...
+
+-- return_value is optional to return instead of the output
+function shell_exec(command, return_value)
+  if return_value then
+    return os.execute(command .. ' &> /dev/null')
+  end 
+  local file = assert(io.popen(command, 'r'))
+  local output = file:read('*all')
+  file:close()
+  return output
```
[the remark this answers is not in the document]
rubo77: doesn't the file have to be closed before the function returns?

package/gluon-offline-ssid/luasrc/lib/gluon/config-mode/model/admin/offline-ssid.lua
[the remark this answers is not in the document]
rubo77: +1 for the "other thing" ;)

But what has to be changed here now in the first place at the position you wrote this comment?

Or did you want to write a global comment and just selected the first line for it?

rotanid:
info from IRC, kind of a review by @NeoRaider :
he doesn't want it to be configurable via UCI oder config-mode. the script should get it's settings directly from the site or domain config, so the behaviour is always consistent for each domain/site.

package/gluon-offline-ssid/luasrc/lib/gluon/config-mode/model/admin/offline-ssid.lua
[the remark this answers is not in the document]
rubo77: gluon-web-offline-ssid is extracted now

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
```diff
@@ -0,0 +1,247 @@
...
+end
+
+if shell_exec('pgrep -f autoupdater', true) == 0 then
+  safety_exit("autoupdater running")
```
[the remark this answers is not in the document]
rubo77: we can't remember any reason either... removed

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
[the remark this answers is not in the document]
rubo77: it is different: in the sysupgrade file it will create a reasonable default, that will be stored in the config files for UCI.
In this file it will create a temporary reasonable value if the config files are messed up or missing.

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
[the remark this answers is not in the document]
rotanid: dont we *restart* the wifi service?

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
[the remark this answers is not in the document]
rubo77: I want, that the script is enabled in all cases, the only case it is disabled should be if there is a '0' in uci -enabled field.

it should be enabled if the enabled entry in uci is missing or anything other than '0'.

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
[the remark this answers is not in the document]
CodeFetch: If you don't close a file descriptor in Lua, it will either be closed by the garbage collector if you don't have a reference anymore or it will be closed by a POSIX kernel on program termination.
I assumed the garbage collector would be more clever with Lua 5.1 and directly close the file as it is not associated with a variable, but I did some testing and found out that the garbage collector still only runs "from time to time". Thus you can leave the code as is, but it wouldn't really hurt having a few open file descriptors.

package/gluon-offline-ssid/luasrc/lib/gluon/config-mode/model/admin/offline-ssid.lua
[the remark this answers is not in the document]
CodeFetch: luasrc/lib/gluon/config-mode/controller/admin/gluon-offline-ssid.lua
should be named 
luasrc/lib/gluon/config-mode/controller/admin/offline-ssid.lua

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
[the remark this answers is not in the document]
CodeFetch: @rotanid The script changes the hostapd config in /var/run which resolves to /tmp/run. Then hostapd is just being killed and procd restarts it.

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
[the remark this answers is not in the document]
CodeFetch: Then you should rename the uci entry to ".disabled". Having not set an enabled field for a section that supports one means it is disabled.

package/gluon-offline-ssid/luasrc/lib/gluon/config-mode/model/admin/offline-ssid.lua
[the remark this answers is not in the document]
rubo77: which one? I think it is already, no?

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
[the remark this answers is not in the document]
rubo77: ok, then we leave it as it is

CodeFetch:
@rubo77 I just realized that gluon-util offers functions for shell execution (util.exec) and for string trimming (util.trim). Please use them where possible.

rubo77:
`util.exec` doesn't do `assert()` around `popen()` isn't that needed?

And I need a fuinction to return the exit-code of the command anyway.

`util.trim` is added

CodeFetch:
@NeoRaider @rubo77 Indeed I think it should as a failed popen returns NULL instead of a file descriptor.
My suggestion is the following (haven't tested, yet):
```
local function readall(f)
	if not f then
		return nil
	end

	return f:read('*a'), f:close()
end

function readfile(file)
	return readall(assertt(io.open(file)))
end

function exec(command)
	local out, ret = readall(assert(io.popen(command)))
	return out, ret
end
```
now you can use it this way 
```
local _, ret = util.exec(ssid_grep .. ssid.ssid .. '" ' .. hostapd)
if not ret then
	(...)
end
```
or without using an additional variable
```
if not select(2, util.exec(ssid_grep .. ssid.ssid .. '" ' .. hostapd)) then
	(...)
end
```

rubo77:
I got it here: https://stackoverflow.com/a/5243210/1069083

but the user there doesn't remember, why he used assert there either.

I will have to test it, but I am on holidays for 2 weeks now...

CodeFetch:
See my
> EDIT: @rubo77 NeoRaider does not use assert(), because he checks whether the file pointer is nil and returns nil in that case himself. Thus if you need assert, you can still do assert(util.exec( ... ))...

rubo77:
We should check for both: `system.@system[0].hostname` for the internal hostname
And Optionally: `system.@system[0].pretty`

NeoRaider:
> We should generate the offline SSID from both: `system.@system[0].hostname` for the internal hostname
> And Optionally: `system.@system[0].pretty_hostname`
Just use the pretty_hostname module like `package/gluon-config-mode-hostname/luasrc/lib/gluon/config-mode/wizard/0100-hostname.lua` (and some other scripts)

rubo77:
I already use the pretty_hostname function: https://github.com/freifunk-gluon/gluon/blob/34281512f8833be9b894da9ea8bc64af48072935/package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid#L102

Is it possible that the pretty_hostname is empty? Or will it default to the default_hostname? Then that part should be fine like it is.

rubo77:
@CodeFetch : how do I distinguish with your suggestion between getting the output of a command and getting the return-value?

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
rubo77: This is still open, see below: https://github.com/freifunk-gluon/gluon/pull/1649#issuecomment-475932614

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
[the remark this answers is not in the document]
rubo77: I don't think an extra logger function in util is needed, as `logger` is already such a function in openwrt. maybe the extra options can be ommitted, I didn't check

CodeFetch:
@rubo77 The first return value is the output, the second is the return value. Sorry, I don't understand your question.

package/gluon-offline-ssid/luasrc/usr/bin/gluon-offline-ssid
rubo77: > we should put all the connection checks into a separate package and let this package depend on it.

#1684

rubo77:
Thoughts about using a ping to check the connection:

- In a large mesh, if all nodes ping the same IP at the same time, this could result in thausands of pings in the same second and this could be **misinterpreted** by the target or the hoster of the target as a **DOS attack**
- if the target of the ping is an external server, that is only reacheable through the gateway and a VPN to the internet, a failure in that VPN to the internet would result in a mesh-wide alarm of all offline-checks although the mesh is still intact.

I can imagine two intentions, why a community wants an offline-SSID:

1. nodes should indicate, that they cannot reach any gateway
2. nodes should indicate, that the internet is not working

If you only want to indicate a not working gateway connection, a ping is not the best solution, the implemented gateway connection check will suffice

If you want to check the connection to the internet, the ping would be a better solution, but We should ensure, not all pings are set up at the same time, (maybe add a random `sleep` to the script?)

rubo77:
> @CodeFetch:
>      `return f:read('*a'), f:close()`
> ...
>  The first return value is the output, the second is the return value. 

If we globally change the  function util.readall() wouldn't that affect calls in other packages, where there are no zwo return vlaues expected?

NeoRaider:
> > @CodeFetch:
> > `return f:read('*a'), f:close()`
> > ...
> > The first return value is the output, the second is the return value.
> 
> If we globally change the function util.readall() wouldn't that affect calls in other packages, where there are no zwo return vlaues expected?

I don't think any changes to readall() are necessary; the function is fine as it is. And indeed a change might break some callers.

NeoRaider:
Oops, I accidentally deleted a comment that was not mine...

rubo77:
@CodeFetch:
can you re-post the comment, neoraider deleted?

NeoRaider:
I assume a lof of this PR will be replaced anyways when #1684 is done, so a review in its current state is not very useful.

CodeFetch:
I don't see the comment anymore. It was likely not important.
@NeoRaider Regarding the exec function, I don't like that it discards the status code.

rubo77:
I think it is fine, we have two existing functions util.exec, which returns just the output and os.execute to return just the exit code

CodeFetch:
But none that returns both, but you're right it's not relevant for the PR...

rubo77:
There seems something odd with devices that have 2.4 and 5GHz: only one of the ssids is set back to normal, the other stays in offline state.

log:
```
# /usr/bin/gluon-offline-ssid
* de:ad:be:ef:ff:04 (195) 8e:a2:f8:50:93:99 [  mesh-vpn]: 10.0/2.0 MBit
  de:ad:be:ef:ff:06 (255) 8e:a2:f8:50:93:99 [  mesh-vpn]: 10.0/2.0 MBit
  de:ad:be:ef:ff:01 (195) 8e:a2:f8:50:93:99 [  mesh-vpn]: 10.0/2.0 MBit
  de:ad:be:ef:ff:00 (195) 8e:a2:f8:50:93:99 [  mesh-vpn]: 10.0/2.0 MBit
uptime in minutes:13.58, every 1 minutes, countdown:0
node is considered offline
ssid=FF_Offline_R6120-Ha...merweg-53
ssid FF_Offline_R6120-Ha...merweg-53
http://kiel.freifunk.net/ is correct

```

mweinelt:
Hi @rubo77 

we decided that this changeset is a good candidate for the community repository that is about to be created. 

Please stay tuned.

package/gluon-offline-ssid/Makefile
```diff
@@ -0,0 +1,22 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-offline-ssid
+PKG_VERSION:=7
+
+include ../gluon.mk
+
+define Package/$(PKG_NAME)
+  TITLE:=changes the SSID to an Offline-SSID so clients don't connect to an offline WiFi
+  DEPENDS:=+gluon-core +micrond
```
rotanid: https://github.com/Freifunk-Nord/gluon-ssid-changer/commit/f6db033d6568ea27805c9694b20085b8dec5ba87#diff-2e2f87a246ba417fe54fa668193fa2db

package/gluon-web-offline-ssid/Makefile
```diff
@@ -0,0 +1,13 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-web-offline-ssid
+PKG_VERSION:=1
+
+include ../gluon.mk
+
+define Package/$(PKG_NAME)
+  TITLE:=adds an option to enable and disable the offline-ssid changer in advanced config mode
+  DEPENDS:=+gluon-core +micrond +gluon-offline-ssid
```
rotanid: https://github.com/Freifunk-Nord/gluon-ssid-changer/commit/f6db033d6568ea27805c9694b20085b8dec5ba87#diff-2e2f87a246ba417fe54fa668193fa2db

rubo77: added that fix and rebased to current master

rubo77:
The jenkins build sais:

    checking consistency... /home/jenkins/workspace/gluon-ci_PR-1649/docs/package/gluon-offline-ssid.rst: WARNING: document isn't included in any toctree


How can I fix that?

package/gluon-web-offline-ssid/Makefile
rotanid: i think you changed it in the wrong Makefile!

rubo77: oh yes, fixed an rebased

rubo77:
Rebased to master and added link in docs/index.rst

mweinelt:
Please retarget this pull request over in the [community-packages](https://github.com/freifunk-gluon/community-packages) repository, where this change would be very welcome.

Closing, since the decision in https://github.com/freifunk-gluon/gluon/pull/1649#issuecomment-549560209 is final.

blocktrron:
Another point to consider here is that reloading the hostapd config leads to radios operating on DFS frequencies to re-perform CAC, which will stop clients from connecting for 60 seconds. I just came across this yesterday when preparing patches for another hostapd issues but this might be a bug in hostapd.

rubo77:
@blocktrron, so you have a solution for this?
https://github.com/Freifunk-Nord/gluon-ssid-changer/issues/26

blocktrron:
@rubo77 you could patch hostapd to make the scanning process asynchronous.

Even then you might still encounter issues, as the kernel might return an error when rescheduling a running CAC. Solution would be to provide a kernel API for hostapd to query CAC state for a specific channel.

That being said, OpenWrt 20.xx changes how hostapd works (hostapd is now running as a service and only reconfigured by netifd). Current implementations relying on wireless restarts using `wifi` and config reloads using interrupts will most likely break and need to adapted to use ubus.

rubo77:
This is way above my programming skill

since many communities use the ssid-changer already, we need someone, that knows how to do it and add a patch to get a running ssid-changer on 2020.xx.

## help wanted :)

blocktrron:
I've had to develop a way to dynamically change the SSID braodcasted (no association intended) for a different project. This obviously has multiple shortcomings (e.g. the SSID is never updated in the mac80211 subsys), however this might be a starting point for someone pursuing this SSID changer.

```
From 4a83ffbd717c8618aa0b170945e6076a8f91dfe5 Mon Sep 17 00:00:00 2001
From: David Bauer <mail@david-bauer.net>
Date: Mon, 19 Oct 2020 23:25:57 +0200
Subject: [PATCH] hostapd: ubus: add interface to update SSID

Add a method which allows to dynamically alter the SSID broadcasted on
the air.
---
 .../services/hostapd/src/src/ap/ubus.c        | 30 +++++++++++++++++++
 1 file changed, 30 insertions(+)

diff --git a/package/network/services/hostapd/src/src/ap/ubus.c b/package/network/services/hostapd/src/src/ap/ubus.c
index acc735ca54..ec2099343a 100644
--- a/package/network/services/hostapd/src/src/ap/ubus.c
+++ b/package/network/services/hostapd/src/src/ap/ubus.c
@@ -544,6 +544,35 @@ hostapd_bss_wps_cancel(struct ubus_context *ctx, struct ubus_object *obj,
 }
 #endif /* CONFIG_WPS */
 
+enum {
+	SET_SSID_STR,
+	__SET_SSID_MAX,
+};
+
+static const struct blobmsg_policy set_ssid_policy[__SET_SSID_MAX] = {
+	[SET_SSID_STR] = { "ssid", BLOBMSG_TYPE_STRING },
+};
+
+static int
+hostapd_ubus_set_ssid(struct ubus_context *ctx, struct ubus_object *obj,
+		       struct ubus_request_data *ureq, const char *method,
+		       struct blob_attr *msg)
+{
+	struct hostapd_data *hapd = container_of(obj, struct hostapd_data, ubus.obj);
+	struct blob_attr *tb[__SET_SSID_MAX];
+	char *buf;
+
+	blobmsg_parse(set_ssid_policy, __SET_SSID_MAX, tb, blob_data(msg), blob_len(msg));
+
+	if (!tb[SET_SSID_STR] || blobmsg_data_len(tb[SET_SSID_STR]) > SSID_MAX_LEN)
+		return UBUS_STATUS_INVALID_ARGUMENT;
+
+	buf = blobmsg_data(tb[SET_SSID_STR]);
+	hapd->conf->ssid.ssid_len = os_strlen(buf);
+	os_memcpy(hapd->conf->ssid.ssid, buf, hapd->conf->ssid.ssid_len);
+	return 0;
+}
+
 static int
 hostapd_bss_update_beacon(struct ubus_context *ctx, struct ubus_object *obj,
 			struct ubus_request_data *req, const char *method,
@@ -1311,6 +1340,7 @@ static const struct ubus_method bss_methods[] = {
 	UBUS_METHOD_NOARG("wps_status", hostapd_bss_wps_status),
 	UBUS_METHOD_NOARG("wps_cancel", hostapd_bss_wps_cancel),
 #endif
+	UBUS_METHOD("set_ssid", hostapd_ubus_set_ssid, set_ssid_policy),
 	UBUS_METHOD_NOARG("update_beacon", hostapd_bss_update_beacon),
 	UBUS_METHOD_NOARG("get_features", hostapd_bss_get_features),
 #ifdef NEED_AP_MLME
-- 
2.28.0
```

rubo77:
So which gluon version will use the new Openwrt, where the ssid-changer would not work anymore?

rotanid:
depending on the discussion in #2107 it is next one or the one after that ;-)

rubo77:
So it seems like the ssid-changer only works up until 2020.2.x. Since gluon 2021.1 it is not working any more

rotanid:
i disagree, there was no change in v2021.1.x that would break it. at least in our fork we didn't do any change in 1.5 years and it still works with the newest release v2021.1.2
though, your statement might depend very hard on which fork is being used, as there are many.

rubo77:
@rotanid we use the Freifunk-Nord fork

rubo77:
We use this fork: https://github.com/Freifunk-Nord/gluon-ssid-changer

And I have to elaborate: the ssid-changer still works with the newest release v2021.1.2.

So it will be only on master-builds, where it will not work any more, I guess, and we will need a new solution for the next release then.

PR #1650, opened 2019-02-16
ramips-mt76x8: add support for Netgear R6120

blocktrron:
MT7628/MT7612 based dualband-ac WiFi router. Very similar to the Archer C50 v3/v4 but with double the Flash (16MiB instead of 8 MiB).

This is also the first MT7628 based dualband device which can be flashed directly from the Vendor Web-Interface.

Price:
 - [New: ~40€ ](https://geizhals.de/netgear-r6120-r6120-100pes-a1624501.html)
 - [Refurbished (Amazon WHD): 30€](https://www.amazon.de/gp/offer-listing/B071HWVV5J/ref=dp_olp_used?ie=UTF8&condition=used)

-----

  - [x] webinterface
  - [ ] tftp
  - [x] other: nmrp recovery tool
    ```
    root@64xxx-b039560bc5ab:~# lua -e 'print(require("platform_info").get_image_name())'
    netgear-r6120
    ```
- wired network
- wifi (if applicable)
- led mapping
  - power/sys led
  - radio leds
        - Device has only one WiFi LED. Shows status of 2.4 GHz radio
    - [x] should show activity
  - switchport leds
        - Shows Link state. Activity is not shown as there is no netdev trigger tied to it. This is fairly common (maybe update this requirement)

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -380,6 +380,10 @@ ramips-mt76x8
 
   - GL-MT300N v2 [#80211s]_
 
+* NETGEAR
+
+  - R6120 [#80211s]_
+
 * TP-Link
 
   - TL-WR841N v13 [#80211s]_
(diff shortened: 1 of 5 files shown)


rotanid:
it is supported in OpenWrt master and does fit our checklist regarding "worth a backport", merging.

PR #1651, opened 2019-02-16
Minor changes of web references

bobcanthelpyou:
just some minor changes
* ecdsa utils new home (?) https://github.com/freifunk-gluon/ecdsautils
* update some urls by following 301 'moved permanently' redirects
* switch http to https

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -447,7 +447,7 @@ Footnotes
   is explicitly set to something other than *11s*
 
 .. [#avmflash]
-  For instructions on how to flash AVM devices, visit https://www.fritzfla.sh
+  For instructions on how to flash AVM devices, visit https://fritzfla.sh
 
 License
 -------
--- a/contrib/sign.sh
+++ b/contrib/sign.sh
@@ -15,7 +15,7 @@ The script may be performed multiple times to the same document
 to indicate an approval by multiple developers.
 
 See also
- * ecdsautils on https://github.com/tcatm/ecdsautils
+ * ecdsautils on https://github.com/freifunk-gluon/ecdsautils
 
 EOHELP
 	exit 1
--- a/contrib/sigtest.sh
+++ b/contrib/sigtest.sh
@@ -8,8 +8,8 @@ sigtest.sh checks if a manifest is signed by the public key <public>. There is
 no output, success or failure is indicated via the return code.
 
 See also:
- * ecdsautils in https://github.com/tcatm/ecdsautils
- * http://gluon.readthedocs.org/en/latest/features/autoupdater.html
+ * ecdsautils in https://github.com/freifunk-gluon/ecdsautils
+ * https://gluon.readthedocs.io/en/latest/features/autoupdater.html
 
 EOHELP
     exit 1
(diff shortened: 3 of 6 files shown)


patches/openwrt/0008-ipq40xx-add-support-for-the-ZyXEL-NBG6617.patch
```diff
@@ -80,7 +80,7 @@ to get the challange value/seed.
 |012345678901
 
 This seed/value can be converted to the password with the help of this
-bash script (Thanks to http://www.adslayuda.com/Zyxel650-9.html authors):
+bash script (Thanks to https://www.adslayuda.com/zyxel650-actualizacion_del_firmware.html authors):
```
rotanid: you can't edit this one, this is a commit in openwrt that we simply cherry-picked.

docs/index.rst
```diff
@@ -447,7 +447,7 @@ Footnotes
   is explicitly set to something other than *11s*
 
 .. [#avmflash]
-  For instructions on how to flash AVM devices, visit https://www.fritzfla.sh
+  For instructions on how to flash AVM devices, visit https://fritz-tools.readthedocs.io/de/latest/
```
rotanid: i think it is intended to keep the easiert url, let's wait for @blocktrron

mweinelt: Yeah, I think it's a vanity url that is primarily easy to remember. Let's keep it that way, maybe drop the ẁww.` though :)

patches/openwrt/0008-ipq40xx-add-support-for-the-ZyXEL-NBG6617.patch
bobcanthelpyou: Just thought that the old link might become invalid in the future. I will remove the commit.

docs/index.rst
bobcanthelpyou: I am not sure about it, from my point of view i will not remember the short url, when using the readthedocs page, nothing is saying anything about something like the short link (beside of fritz), i will just remember "fritz tools" i guess.

Also short urls might become invalid in the long run, might be used for tracking people, .. i don't know.

Maybe replacing it with https://fritz-tools.readthedocs.io/en/latest/ would be better as the Gluon docs are using en, too?

But i could remove/update the change (as requested) of cause.

rotanid: let's wait for the tool's author so he can decide.

patches/openwrt/0008-ipq40xx-add-support-for-the-ZyXEL-NBG6617.patch
rotanid: a commit message written by someone else, contained in a public repository, is written into stone for eternity ;-) we are simply copying. also, this copy will become obsolete as soon as Gluon is based on OpenWrt 19.x

docs/index.rst
bobcanthelpyou: yes please

blocktrron: I don't consider readthedocs as the final solution, as the performance is rather bad. Also, in case Readthedocs ceases to exist or we have other problems with their service, we can easily switch to a self-hosted solution.

So i would leave the fritzfla.sh vanity URL in place as people are referring to the flash-tool as fritzflash already.

Regarding the domain/hosting situation: This is currently handled by me, but no access logs or anything else are created for it.

bobcanthelpyou: undone the change but remove `www` as suggested.

PR #1656, opened 2019-02-25
gluon-*batman-adv: Also list neighbors with non-best direct link

ecsv:
Links between two direct neighbors are not always the best route between
these devices. The flag BATADV_ATTR_FLAG_BEST would not be set for these
originator entries and the respondd module would just ignore this entry.

This causes missing links in meshviewer and similar tools. And when the
link quality is nearly equal and but fluctuates slightly, these links will
from time to time appear and disappear on the map.

Or ff these neighbors are not accepted and returned to the status page then
some of the neighbor entries will show a name, (acceptable) signal strength
and mac address but no TQ value.

### DIFF
--- a/package/gluon-mesh-batman-adv/src/respondd.c
+++ b/package/gluon-mesh-batman-adv/src/respondd.c
@@ -667,9 +667,6 @@ static int parse_orig_list_netlink_cb(struct nl_msg *msg, void *arg)
 				      BATADV_ARRAY_SIZE(parse_orig_list_mandatory)))
 		return NL_OK;
 
-	if (!attrs[BATADV_ATTR_FLAG_BEST])
-		return NL_OK;
-
 	orig = nla_data(attrs[BATADV_ATTR_ORIG_ADDRESS]);
 	dest = nla_data(attrs[BATADV_ATTR_NEIGH_ADDRESS]);
 	tq = nla_get_u8(attrs[BATADV_ATTR_TQ]);
--- a/package/gluon-mesh-batman-adv/src/respondd.c
+++ b/package/gluon-mesh-batman-adv/src/respondd.c
@@ -695,6 +695,7 @@ static int parse_orig_list_netlink_cb(struct nl_msg *msg, void *arg)
 
 	json_object_object_add(obj, "tq", json_object_new_int(tq));
 	json_object_object_add(obj, "lastseen", json_object_new_double(lastseen / 1000.));
+	json_object_object_add(obj, "best", json_object_new_boolean(attrs[BATADV_ATTR_FLAG_BEST]));
 	json_object_object_add(interface, mac1, obj);
 
 	return NL_OK;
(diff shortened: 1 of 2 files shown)


CodeFetch:
@ecsv The only problem I see is that the function is called "originators" and originators are per definition the best next hop. We need to add a flag to the respondd-data and visualize it on the status page to make clear "this is not a best next hop". Not showing the nodes on the status page and the map is of course wrong. Thank you for pointing this issue out.

ecsv:
Ehrm, since when are originators best next hops?

CodeFetch:
Sorry I just edited it "best next hop to something"

ecsv:
> Sorry I just edited it "best next hop to something"

I seem to have a different definition for that. https://www.open-mesh.org/projects/open-mesh/wiki/BATMANConcept

CodeFetch:
@ecsv Okay thus the definition of originators is still correct with your changes, but BATADV_ATTR_FLAG_BEST does mean it is the best entry to this originator. 
Shouldn't there be a best entry for every originator? Doesn't removing this flag mean you get duplicate originator entries?

ecsv:
Problem is here that we want to have only the information for the neighors. So it goes through the list of originators, compares the originator and next-hop mac address and only accepts entries which have equal originator and next-hop entry. Before that (and even when the debugfs file was parsed) only the best next hop entry was checked.

I've reimplemented the behavior of the debugfs parser when I wrote the netlink part and made a note that this behavior has to be changed (because it seems to be wrong).... and then I forgot about it. Luckily, I've observed this weird behavior again and thus prepared the changes in this PR.

PS: Keep in mind that the debugfs file had one line per originator and the mac address after the originator was the best next hop.

CodeFetch:
@ecsv Why do you give me a thumbs down emoji? Don't you know Microsoft uses that for rating.

CodeFetch:
@ecsv Can you please just add something like `json_object_object_add(obj, "best", json_object_new_boolean(attrs[BATADV_ATTR_FLAG_BEST]);` to respondd.c?

T-X:
Looks like a good idea to me. This will give valuable extra information regarding the robustness / redundancy / potentiall fallback links in the topology. Code looks fine, too.

T-X:
What confused me a bit, which is not directly related to this patch, is the `` if (memcmp(orig, dest, 6) != 0)``.  A check which seems to have been part of the debugfs version earlier already via a ``if (strcmp(mac1, mac2))``. Had to read the batman-adv code a few times concerning the neigh_node, wasn't sure whether that would play nicely with the secondary interface MAC. But it seems to work.

(I'm wondering whether it would be easier and more intuitive for users if we were adding a BATADV_ATTR_DIRECT_NEIGH attribute to batman-adv, which would be set to true if ``neigh_node->orig_node == orig_node in *_orig_dump_subentry(). So that userspace could check this flag instead of this memcmp())

ecsv:
Was this now approved by @T-X or not? There is no status change in his review (still set to "waiting for review").

I will definitely not add the BATADV_ATTR_DIRECT_NEIGH flag in this PR. This is a thing which first has to go to upstream first.

rotanid:
@ecsv let's treat the "code looks fine, too" as an ACK.
i'm merging.

PR #1657, opened 2019-02-25
ath10k/ath10k-ct: fix incorrect multicast/broadcast rate setting

ecsv:
If no mcast_rate is set for the wifi-iface then there is no rate_idx (0)
set for the bss. This breaks for example 5GHz meshpoint interfaces because
0 maps to a CCK rate (11Mbit/s).

It must also be avoided that the ath10k-ct internal state for the rates is
not synced with the mac80211 rates state. Otherwise, the user specified
rate (e.g. a wifi-iface mcast_rate for a meshpoint interface) will only be
set on startup. And a short while after that, ath10k-ct specific code in
ath10k_check_apply_special_rates is missing a valid rate in its own
structures and is then recalculating a new default rate. This default rate
is in most situations not the requested rate.

Fixes: a399b607357b ("ath10k/ath10k-ct: support multicast and management rate control")
Forwarded: https://github.com/openwrt/openwrt/pull/1861
Forwarded: https://github.com/openwrt/openwrt/pull/1862


mweinelt:
Thanks!

PR #1658, opened 2019-02-25
gluon-mesh-wireless-sae: Add support for SAE on 802.11s mesh

### LINKED ISSUE #1636: Add support for encrypted 802.11s connections (SAE)
Building on #1586, where we achieve more secure station connections, enduser traffic may still be routed via 802.11s connections, which at this point are entirely unprotected, and can be passively sniffed to recover previously encrypted traffic.

Therefore we should allow to enable SAE (Simultaenous Authentication of Equals) on 802.11s mesh connections. They would need to use a well-known PSK, and therefore offer little to no mitigation of active attacks, but since routers would generate a session key from the PSK and their MAC addresses we would protect the traffic after a secure connection was made from a fairly easy mountable passive attacks.

mweinelt:
Setting a pre-shared key on the wireless mesh will enable encryption
using the SAE key agreement scheme and require encryption for all mesh
links between any two wireless mesh nodes.

<s>Disables the ar71xx-tiny target and other devices with 4 MB of flash
when the GLUON_WLAN_MESH_SAE variable is set to 1 as they don't have
enough free space for the required packages.</s>

Fixes #1636

Signed-off-by: Martin Weinelt <martin@darmstadt.freifunk.net>


mweinelt:
I'm still talking to people about the security guarantees we can expect, especially for the case of public mesh networks where the PSK is well-known.

CodeFetch:
@mweinelt I've learned never to do crypto if it is not secure, because it misleads the users to make false assumptions, but that's actually what OWE does. Hardened security is of course only achievable in the case of SAE using a WOT. My approach would be quite radical - use a common PSK only for the first association with a specific node, negotiate a SK and save it in flash. If SAE does not support different PSKs for different MPs, we need to patch it or not use it at all. Of course an attacker could be doing MITM right from the beginning, but it's the same with OWE.

mweinelt:
> I've learned never to do crypto if it is not secure because it misleads the users to make false assumptions

You could argue that any crypto is better than none, but I'll take that criticism.

> but that's actually what OWE does.

You are saying OWE is insecure? How so?

> My approach would be quite radical - use a common PSK only for the first association with a specific node, negotiate a SK and save it in flash.

Sure, TOFU with the well-known PSK then reneg a different one, sounds good.

> If SAE does not support different PSKs for different MPs, we need to patch it or not use it at all. 

I haven't checked, but I don't think wpa_supplicant supports different PSKs per BSSID. But honestly, Gluon might also be applicable for private mesh networks, so SAE is still a good thing to have either way.

> Of course an attacker could be doing MITM right from the beginning, but it's the same with OWE.

You can certainly roll out something like WPA-EAP onto your mesh network to get some authentication going, if that helps. But OWE is certainly an improvement on open.

> The mesh is still not secure, because it is a mesh network...

This is specific to public mesh networks.

> Another thing is using this approach we could use an additional signature used by the nodes on their first handshake to verify their identity and let the nodes submit these entries to a supernode they trust. If there are deviations we at least know that some nodes are corrupted. Of course you could also submit false deviations, but to a human (who knows which nodes to trust) it would be more obvious which nodes are going rude.

A protocol modification is certainly out of scope for this pull request. And I'd honestly wouldn't want a "trust the supernodes" scheme.

CodeFetch:
> > but that's actually what OWE does.
> 
> You are saying OWE is insecure? Besides the lack of mutual authentication that all WiFi authentication schemes besides EAP have, how so?

Exactly that point. Not using a PSK and not doing authentication is insecure. It's better than no encryption as you need to actively intervene for hacking, but try to tell someone non-technical that encryption doesn't mean authentication and what an evil twin attack is.

> 
> > If SAE does not support different PSKs for different MPs, we need to patch it or not use it at all.
> 
> I haven't checked, but I don't think wpa_supplicant supports different PSKs per BSSID. But honestly, Gluon might also be applicable for private mesh networks, so SAE is still a good thing to have either way.

You're right. I have only thought about open infrastructure... The limitiation with multiple PSKs and SAE itself seems to be the drivers and their partially closed-source binary blobs. Ath9k would be capable of doing that.

> A protocol modification is certainly out of scope for this pull request. And I'd honestly wouldn't want a "trust the supernodes" scheme for a mesh network.

That was not my intention, but only to collect the data to be able to detect such problems.

mweinelt:
I'm not a crypto nerd, I couldn't get anyone with more knowledge than me to take a stab at this, so here are a few quotes from the SAE paper. I don't believe SAE is well-equipped for public mesh networks.

> Both parties to the exchange must have a common view of the password. It deserves mentioning that 
> the view of the password could, infact, be derived from a one-way function of a raw password.
> Employing such a technique would add an additional burdento an attacker who has stolen a mesh
> point provisioned with a common view of the password.

I don't have the slightest clue how, what I believe to be, obfuscation through a hash function could improve anything. The raw password will need to be saved somewhere, no? :hankey: 

> It should be noted, though, that compromise of the password can enable attack against all future runs of the protocol because a man-in-the middle attack cannot be detected by either side.

That's reasonable, as the password is used to authenticate each other.

My takeaway: One could argue we are worse off with plain mesh connections.

The paper is unfortunately paywalled, so for reference: 

Simultaneous Authentication of Equals: A Secure, Password-Based Key Exchange for Mesh Networks
by Dan Harkins
https://ieeexplore.ieee.org/document/4622764

yanosz:
.. and alternative might be an approach similar to AeroFlux[1]: Users are able to initiatie a WPA connection to a trusted device acting as a NfV (e.g. their own raspberry pi). But that's even more challanging.

[1] https://www.net.t-labs.tu-berlin.de/~stefan/ons14wireless.pdf

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -167,6 +167,11 @@ local function configure_mesh(config, radio, index, suffix, disabled)
 		macfilter = macfilter,
 		maclist = maclist,
 	})
+
+	if config.psk(false) then:
+		uci:set('wireless', 'wifi-iface', name, 'encryption', 'psk2+ccmp')
```
txt-file: psk2+ccmp means WPA2. Your pull request description is about WPA3. WPA3 would be 'sae' according to https://openwrt.org/docs/guide-user/network/wifi/basic#wpa_modes

mweinelt:
> Derive key automatically from domain_seed, don't let someone define a "secret", because it is not secret at all

As the `domain_seed` is multidomain specific I've chosen the `prefix6` variable instead.

I've moved the config below `wifi.mesh.sae{,_passphrase}`, I believe it is a documentation issue to convey to users that there is little gain from setting the passphrase in public mesh networks. On the other hand this keeps the option for private mesh networks to pop up and have a secure mesh network.

mweinelt:
Revamped and eager for a review.

mweinelt:
Seems to be working!

```
root@64xxx-nbg6617:~# logread -f
Wed Oct 23 23:10:21 2019 daemon.notice wpa_supplicant[3140]: mesh0: new peer notification for 22:a9:42:f4:f6:91
Wed Oct 23 23:10:22 2019 daemon.notice wpa_supplicant[3140]: mesh0: mesh plink with 22:a9:42:f4:f6:91 established
Wed Oct 23 23:10:22 2019 daemon.notice wpa_supplicant[3140]: mesh0: MESH-PEER-CONNECTED 22:a9:42:f4:f6:91
Wed Oct 23 23:10:24 2019 daemon.notice wpa_supplicant[2567]: mesh1: new peer notification for 22:a9:42:f4:f6:95
Wed Oct 23 23:10:24 2019 daemon.notice wpa_supplicant[2567]: mesh1: mesh plink with 22:a9:42:f4:f6:95 established
Wed Oct 23 23:10:24 2019 daemon.notice wpa_supplicant[2567]: mesh1: MESH-PEER-CONNECTED 22:a9:42:f4:f6:95
```

package/gluon-mesh-wireless-sae/check_site.lua
```diff
@@ -0,0 +1,2 @@
+need_boolean(in_domain({'wifi', 'mesh', 'sae'}), false)
+need_string(in_domain({'wifi', 'mesh', 'sae_passphrase'}), false)
```
CodeFetch: I'd prefer to have a section in the site.conf like:
sae {
    passphrase = 'xyz'
}
and automatically assume it's activated when this entry exists...

Everything else looks good. One thing that is missing is a note in the documentation.

What I wonder about is if setting SAE here allows backward-compatibility with WPA2 (Dragonblood attack). I'd prefer to have the WPA2 mode disabled if possible. I'm not up-to-date about whether it was fixed in OpenWrt...

CodeFetch: Hm... I think you added a separate flag, because the passphrase could be hashed from the site prefix. Shouldn't this be the site ID instead?

mweinelt: > I'd prefer to have a section in the site.conf like:
> sae = {
> passphrase = 'xyz'
> }
> and automatically assume it's activated when this entry exists...

The idea is to not require public mesh networks to explicitly set a passphrase, when no passphrase will be secure anyway. On the other hand I don't want to exclude someone who'd want it in a private mesh network to be able to actually use a proper shared secret. Therefore the requirement to split that up. The current config structure looks like  this:

```lua
wifi = {
  mesh = {
    sae = true,
    sae_passphrase = nil, -- only set this if you are running a *private* mesh network
  },
},
```

> Everything else looks good. One thing that is missing is a note in the documentation.

Documentation is still on the todo list.

> What I wonder about is if setting SAE here allows backward-compatibility with WPA2 (Dragonblood attack). I'd prefer to have the WPA2 mode disabled if possible. I'm not up-to-date about whether it was fixed in OpenWrt...

I've grepped hostap.git for CVE-2019-9494 and the latest related commit is 
hostap_2_7-721-gcff138b07.

We're currently using `hostapd-mini - 2019-08-08-ca8c2bd2-2`, which is
```
❯ git describe ca8c2bd2
hostap_2_9
```

Please decide for yourself if this is fixed. I currently don't have the time to research this.

```
commit cff138b0747fa39765cbc641b66cfa5d7f1735d1
Author: Jouni Malinen <jouni@codeaurora.org>
Date:   Sat Mar 2 16:05:56 2019 +0200

    SAE: Use constant time operations in sae_test_pwd_seed_ffc()
    
    Try to avoid showing externally visible timing or memory access
    differences regardless of whether the derived pwd-value is smaller than
    the group prime.
    
    This is related to CVE-2019-9494.
    
    Signed-off-by: Jouni Malinen <jouni@codeaurora.org>

commit f8f20717f87eff1f025f48ed585c7684debacf72
Author: Jouni Malinen <jouni@codeaurora.org>
Date:   Sat Mar 2 12:45:33 2019 +0200

    SAE: Use const_time selection for PWE in FFC
    
    This is an initial step towards making the FFC case use strictly
    constant time operations similarly to the ECC case.
    sae_test_pwd_seed_ffc() does not yet have constant time behavior,
    though.
    
    This is related to CVE-2019-9494.
    
    Signed-off-by: Jouni Malinen <jouni@codeaurora.org>

commit 90839597cc4016b33f00055b12d59174c62770a3
Author: Jouni Malinen <jouni@codeaurora.org>
Date:   Sat Mar 2 12:24:09 2019 +0200

    SAE: Mask timing of MODP groups 22, 23, 24
    
    These groups have significant probability of coming up with pwd-value
    that is equal or greater than the prime and as such, need for going
    through the PWE derivation loop multiple times. This can result in
    sufficient timing different to allow an external observer to determine
    how many rounds are needed and that can leak information about the used
    password.
    
    Force at least 40 loop rounds for these MODP groups similarly to the ECC
    group design to mask timing. This behavior is not described in IEEE Std
    802.11-2016 for SAE, but it does not result in different values (i.e.,
    only different timing), so such implementation specific countermeasures
    can be done without breaking interoperability with other implementation.
    
    Note: These MODP groups 22, 23, and 24 are not considered sufficiently
    strong to be used with SAE (or more or less anything else). As such,
    they should never be enabled in runtime configuration for any production
    use cases. These changes to introduce additional protection to mask
    timing is only for completeness of implementation and not an indication
    that these groups should be used.
    
    This is related to CVE-2019-9494.
    
    Signed-off-by: Jouni Malinen <jouni@codeaurora.org>

commit 362704dda04507e7ebb8035122e83d9f0ae7c320
Author: Jouni Malinen <jouni@codeaurora.org>
Date:   Tue Feb 26 19:34:38 2019 +0200

    SAE: Avoid branches in is_quadratic_residue_blind()
    
    Make the non-failure path in the function proceed without branches based
    on r_odd and in constant time to minimize risk of observable differences
    in timing or cache use. (CVE-2019-9494)
    
    Signed-off-by: Jouni Malinen <jouni@codeaurora.org>

commit 6513db3e96c43c2e36805cf5ead349765d18eaf7
Author: Jouni Malinen <jouni@codeaurora.org>
Date:   Tue Feb 26 13:05:09 2019 +0200

    SAE: Minimize timing differences in PWE derivation
    
    The QR test result can provide information about the password to an
    attacker, so try to minimize differences in how the
    sae_test_pwd_seed_ecc() result is used. (CVE-2019-9494)
    
    Use heap memory for the dummy password to allow the same password length
    to be used even with long passwords.
    
    Use constant time selection functions to track the real vs. dummy
    variables so that the exact same operations can be performed for both QR
    test results.
    
    Signed-off-by: Jouni Malinen <jouni@codeaurora.org>

commit c93461c1d98f52681717a088776ab32fd97872b0
Author: Jouni Malinen <jouni@codeaurora.org>
Date:   Fri Mar 8 00:24:12 2019 +0200

    OpenSSL: Use constant time selection for crypto_bignum_legendre()
    
    Get rid of the branches that depend on the result of the Legendre
    operation. This is needed to avoid leaking information about different
    temporary results in blinding mechanisms.
    
    This is related to CVE-2019-9494 and CVE-2019-9495.
    
    Signed-off-by: Jouni Malinen <jouni@codeaurora.org>

commit 6e34f618d37ddbb5854c42e2ad4fca83492fa7b7
Author: Jouni Malinen <jouni@codeaurora.org>
Date:   Wed Feb 27 18:38:30 2019 +0200

    Add helper functions for constant time operations
    
    These functions can be used to help implement constant time operations
    for various cryptographic operations that must minimize externally
    observable differences in processing (both in timing and also in
    internal cache use, etc.).
    
    This is related to CVE-2019-9494 and CVE-2019-9495.
    
    Signed-off-by: Jouni Malinen <jouni@codeaurora.org>

commit d42c477cc794163a3757956bbffca5cea000923c
Author: Jouni Malinen <jouni@codeaurora.org>
Date:   Tue Feb 26 11:43:03 2019 +0200

    OpenSSL: Use constant time operations for private bignums
    
    This helps in reducing measurable timing differences in operations
    involving private information. BoringSSL has removed BN_FLG_CONSTTIME
    and expects specific constant time functions to be called instead, so a
    bit different approach is needed depending on which library is used.
    
    The main operation that needs protection against side channel attacks is
    BN_mod_exp() that depends on private keys (the public key validation
    step in crypto_dh_derive_secret() is an exception that can use the
    faster version since it does not depend on private keys).
    
    crypto_bignum_div() is currently used only in SAE FFC case with not
    safe-prime groups and only with values that do not depend on private
    keys, so it is not critical to protect it.
    
    crypto_bignum_inverse() is currently used only in SAE FFC PWE
    derivation. The additional protection here is targeting only OpenSSL.
    BoringSSL may need conversion to using BN_mod_inverse_blinded().
    
    This is related to CVE-2019-9494 and CVE-2019-9495.
    
    Signed-off-by: Jouni Malinen <jouni@codeaurora.org>
```

> Hm... I think you added a separate flag, because the passphrase could be hashed from the site prefix. Shouldn't this be the site ID instead?

I wanted it to be the `domain_seed`, but since that is only available in multidomain setups I opted for `prefix6`, which is a mandatory setting that is also different per domain. Site specific information isn't.

CodeFetch:
It looks like there is no consensus about how to fix the vulnerabilities. Thus changes in the protocol might happen which break backward-compatibility (the proposed changes would indeed break backward-compatibility).  As far as I can see especially VU871675 has not been addressed, yet. Therefore SAE should be considered experimental.

mweinelt:
Do you have a reference you can quote with relation to the hostapd state? If so we could add that to the package documentation.

CodeFetch:
@mweinelt The problem is not hostapd not implementing countermeasures timely or something. The problem is that there is no consensus about how to fix some of the Dragonblood vulnerabilities at the moment, at all.
https://wpa3.mathyvanhoef.com/

The most [important patches](https://w1.fi/security/) for hostapd are in OpenWrt's version. What I wanted to know is if the transition mode can somehow be deactivated for our networks (as we don't rely on WPA2 compatibility) to counter VU871675. 

The problem I mentioned is that it is not clear how to counter some of the vulnerabilities, yet. This is still an open discussion. There have been protocol changes which break backwards-compatibility:
`Fortunately, as a result of our research, both the Wi-Fi standard and EAP-pwd are being updated with a more secure protocol. Although this update is not backwards-compatible with current deployments of WPA3, it does prevent most of our attacks. ` 
Other such protocol changes might follow. This is not an issue that can be fixed by doing this or that - There is no solution where everyone agrees, yet. See WPA3 as a draft...

blocktrron:
@CodeFetch 802.11s is not affected by the downgrade attack, as it only implements SAE due to the equality of node roles.

CodeFetch:
@blocktrron Thank you for that information. I couldn't find anything about that. Can you give a hint where I can find information about this? As far as I know the handshakes are done for each and every node individually. From the specifications it should also be possible to use different PSKs for each link although it is not implemented in hostapd, yet. That's why I'm not sure about this point and I think 802.11s networks don't get the same attention like AP/STA networks.

Do you understand if those proposed "breaking" changes to the WPA3 protocol came into effect already or whether they are still being planned? To me it looks like the WiFi alliance tries to limit the damage on cost of security. That's why I wouldn't trust the current state.

package/gluon-mesh-wireless-sae/luasrc/lib/gluon/upgrade/205-wireless-mesh-sae
```diff
@@ -0,0 +1,31 @@
...
+local site = require 'gluon.site'
+local hash = require 'hash'
+local uci = require('simple-uci').cursor()
+
+
+local function configure_sae(vif, enable)
+	if enable then
+		uci:set('wireless', vif, 'encryption', 'sae')
+		uci:set('wireless', vif, 'key', site.wifi.mesh.sae_passphrase() or hash.md5(site.prefix6()))
+		uci:set('wireless', vif, 'ieee80211w', 2)
```
blocktrron: The activation of MFP has to be done conditionally only when the device supports MFP.

This is a bit of a problem for devices, which do not support MFP, however this is most likely a small edge case, as only Atheros <9160 do not seem to support it.

mweinelt: That basically affects the wr2543n/nd v1. Anything else?

mweinelt:
Added package documentation.

package/gluon-mesh-wireless-sae/luasrc/lib/gluon/upgrade/205-wireless-mesh-sae
mweinelt: I guess I can make that contingent on `platform.supports_mfp` from #1876.

mweinelt: The big question is now: What should be the default (off, optional, required)?

Depends on:
- which radios cannot handle mfp?
  - enabling mfp would exclude these devices/targets from wireless mesh
- is there a difference between optional and required for radios that cannot handle mfp?
  - as in: can we default to optional, because that gives us the best security
  - or: will setting mfp to optional already incur breakage

NeoRaider: In the kernel headers I even see a feature flag NL80211_EXT_FEATURE_MFP_OPTIONAL: "Driver supports the %NL80211_MFP_OPTIONAL value in %NL80211_ATTR_USE_MFP."

This might mean that there are drivers that will work with "required", but break with "optional"? Not entirely sure how to interpret this.

blocktrron: I did not come across such a device (which doesn't mean there are such cases out there).

Honestly, i would not make much about it. MFP is required for WiFi 5 certification and (to my knowledge) all those drivers correctly support it. We are only cutting off AR9132 devices and ramips-rt305x as well as ramips-mt7620.

Looking at the ramips case, these devices cannot even so SAE. The AR9132 can so SAE without MFP, however this chip is flawed in many ways and the obscure TP-Link TL-WR2543 is the only device with 64M RAM we are supporting.

As SAE mesh only makes sense when using with 802.11w, i would propose to check 802.11w capability per-radio and enable SAE only in this case. When 802.11w is not supported, it should be disabled.

Thoughts on that?

docs/package/gluon-mesh-wireless-sae.rst
```diff
@@ -0,0 +1,49 @@
...
+of SAE relies upon the authentication through a shared secret.
+
+In the context of public mesh networks a shared secret is an
+obvious oxymoron. Still this functionality may provide an improvement
+over unencrypted mesh connections in that it protects against a
+passive attacker who did not observe the key agreement. In addition
+we enable Management Frame Protection (802.11w) on wireless mesh
+interfaces to prevent protocol-level deauthentication attacks.
+
+If `wifi.mesh.sae` is enabled a shared secret will automatically be
```
blocktrron: > If `wifi.mesh.sae` is enabled, a shared secret will automatically be

docs/package/gluon-mesh-wireless-sae.rst
```diff
@@ -0,0 +1,49 @@
...
+=======================
+
+This package adds support for SAE on 802.11s mesh connections.
+
+Enabling this package will require all 802.11s mesh connections
+to be encrypted using the SAE key agreement scheme. The security
+of SAE relies upon the authentication through a shared secret.
+
+In the context of public mesh networks a shared secret is an
+obvious oxymoron. Still this functionality may provide an improvement
```
blocktrron: Still, this functionality may provide an improvement

docs/package/gluon-mesh-wireless-sae.rst
```diff
@@ -0,0 +1,49 @@
...
+wifi.mesh.sae \: optional
+  - ``true`` enables SAE on 802.11s mesh connections
+  - ``false`` disables SAE on 802.11s mesh connections
+  - defaults to ``false``
+
+wifi.mesh.sae_passphrase \: optional
+  - sets a shared secret used to authenticate any two mesh nodes,
+    crucial for private mesh networks
+  - should not be set, if the shared secret is shared with untrusted
+    third parties, like in a publish mesh network
```
blocktrron: I think `must` is the better terminology here, as it practically doesn't matter whether or not it is set in this case.

Someone might have a reason to set it individually instead of relying on the generated value.

mweinelt: > Someone might have a reason to set it individually instead of relying on the generated value.

Hence the `should`.

package/gluon-mesh-wireless-sae/luasrc/lib/gluon/upgrade/205-wireless-mesh-sae
mweinelt: > i would propose to check 802.11w capability per-radio and enable SAE only in this case. 

Those devices would then be able to mesh among each other, but no with the remaining nodes of the network. I would rather want to signal, that if your node cannot do SAE it does not do wireless mesh at all.

mweinelt: > When 802.11w is not supported, it should be disabled.

Is that ("it") wireless mesh in general or SAE in particular?

txt-file:
I have a 802.11s running at home and am configuring it via luci.
luci does not show an option to configure 802.11w if the wifi is set to 802.11s+sae. Therefore I wonder if 802.11w even applies to 802.11s networks.

txt-file:
Since luci for OpenWrt 19.07 802.11w option is only shown for ap, ap-wds, sta or sta-wds mode. In luci for OpenWrt 18.06 it was only shown for ap or ap-wds.

luci master: https://github.com/openwrt/luci/blob/cf09705ee394ea9cad90581df09a81ec48b48906/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js#L1660
luci for 19.07: https://github.com/openwrt/luci/blob/openwrt-19.07/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js#L1578
luci for 18.06: https://github.com/openwrt/luci/blob/openwrt-18.06/modules/luci-mod-admin-full/luasrc/model/cbi/admin_network/wifi.lua#L1051

mweinelt:
> I have a 802.11s running at home and am configuring it via luci.
> luci does not show an option to configure 802.11w if the wifi is set to 802.11s+sae. Therefore I wonder if 802.11w even applies to 802.11s networks.

https://github.com/openwrt/openwrt/blob/master/package/network/services/hostapd/files/hostapd.sh#L1016-L1030

mweinelt:
So `ieee80211w` is currently not propagated into the wpa_supplicant config. Also wpa_supplicant does not complain when adding `ieee80211w` to the config (`mode=5`, `key_mgmt=SAE`).

```
# ieee80211w: whether management frame protection is enabled
# 0 = disabled (default unless changed with the global pmf parameter)
# 1 = optional
# 2 = required
# The most common configuration options for this based on the PMF (protected
# management frames) certification program are:
# PMF enabled: ieee80211w=1 and key_mgmt=WPA-EAP WPA-EAP-SHA256
# PMF required: ieee80211w=2 and key_mgmt=WPA-EAP-SHA256
# (and similarly for WPA-PSK and WPA-PSK-SHA256 if WPA2-Personal is used)
# WPA3-Personal-only mode: ieee80211w=2 and key_mgmt=SAE
```

But even when relying on wpa_supplicants default for `ieee80211w`, which is disabled, I'm seeing MFP/PMF on the MBSS Link below.

```
Station 72:f2:91:1d:c9:f9 (on mesh0)
	inactive time:	10 ms
	rx bytes:	167282
	rx packets:	1464
	tx bytes:	29536
	tx packets:	150
	tx retries:	0
	tx failed:	0
	rx drop misc:	6
	signal:  	-82 [-82, -92, -95, -95] dBm
	signal avg:	-80 [-79, -91, -95, -95] dBm
	Toffset:	3653634491 us
	tx bitrate:	57.8 MBit/s MCS 5 short GI
	rx bitrate:	43.3 MBit/s MCS 4 short GI
	rx duration:	96682 us
	last ack signal:-95 dBm
	mesh llid:	0
	mesh plid:	0
	mesh plink:	ESTAB
	mesh local PS mode:	ACTIVE
	mesh peer PS mode:	ACTIVE
	mesh non-peer PS mode:	ACTIVE
	authorized:	yes
	authenticated:	yes
	associated:	yes
	preamble:	long
	WMM/WME:	yes
	MFP:		yes
	TDLS peer:	no
	DTIM period:	2
	beacon interval:100
	connected time:	52 seconds
```

So I guess we need to do some code diving into wpa_supplicant, because from reading 802.11-2016 it appears MFP on MBSS is a thing as soon as a RSNA (WPA2 assocation) is in play.

> Management frame protection protocols in an MBSS apply to individually addressed frames after
establishment of the RSNA MTK, and to group addressed frames indicated as “Group Addressed Privacy” in Table 9-47. Robust management frame protection is implemented with CCMP and GCMP.

How that affects SAE I don't know yet.

mweinelt:
The history of `mesh_rsn.c` talks about implementing SAE as the "robust security network" for mesh links, so I think we can safely assume that 11w works on 11s links with SAE, as we've seen above.

I've dropped the explicit setting of `ieee80211w` in this package as it had no effect at all, but since it apparently gets auto-enabled I've update the docs to that effect.

PR #1662, opened 2019-03-07
[RFC] gluon-web-model: Make model-maps extensible

CodeFetch:
> It was not possible to modularize single sections of forms in the config mode.
> This commit adds support for ´Supramaps´. Using supramaps it is possible to create an additional directory where inframaps reside.
> E.g. a Lua model file with a same-named directory (excluding the file extension) defines a Form element and returns it. This is a supramap.
> If Lua files exist in that directoryn the Form element is available in their environment as "Supramap".
> Thus these files can add additional sections which are "inframaps" and must return the modified supramap.
> This allows e.g. modularizing the password field in the gluon-web-admin package as an own package without hampering usability.

I haven't tested a build with these modifications, yet. Is there a whitelist for LuaSrcDiet or does it ignore variable names starting with capital letters? I'd like to have this feature to increase modularity of several gluon-web-packages including the mesh-vpn-fastd package. Then I can merge N2N support properly. Do you know a better way to achieve this?

### DIFF
--- a/package/gluon-web-model/luasrc/usr/lib/lua/gluon/web/model.lua
+++ b/package/gluon-web-model/luasrc/usr/lib/lua/gluon/web/model.lua
@@ -4,6 +4,7 @@
 
 module('gluon.web.model', package.seeall)
 
+local glob = require 'posix.glob'
 local unistd = require 'posix.unistd'
 local classes = require 'gluon.web.model.classes'
 
@@ -11,12 +12,13 @@ local util = require 'gluon.web.util'
 local instanceof = util.instanceof
 
 -- Loads a model from given file, creating an environment and returns it
-local function load(filename, i18n)
+local function load(filename, i18n, supra)
+	supra = supra or {}
 	local func = assert(loadfile(filename))
 
 	setfenv(func, setmetatable({}, {__index =
 		function(tbl, key)
-			return classes[key] or i18n[key] or _G[key]
+			return supra[key] or classes[key] or i18n[key] or _G[key]
 		end
 	}))
 
@@ -36,7 +38,9 @@ return function(config, http, renderer, name, pkg)
 	local hidenav = false
 
 	local modeldir = config.base_path .. '/model/'
-	local filename = modeldir..name..'.lua'
+	local filename = modeldir .. name .. '.lua'
+	local infradir = modeldir .. name .. '/'
+	local infrafnames = glob.glob(infradir .. '*.lua') or {}
 
 	if not unistd.access(filename) then
 		error("Model '" .. name .. "' not found!")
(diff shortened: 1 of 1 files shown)


NeoRaider:
Can you give an example how LuaSrcDiet affects what you're trying? It should not minimize non-local names.

CodeFetch:
@NeoRaider Ah thank you. I haven't tested it, yet. I was just unsure about how it does the diet. I'm unsure about the naming, too as this can only be used with forms. Maybe I should add a check whether the returned "map" is an instance of a Form. Actually it's not really a map. 

At the moment all these points are not really important as I just need a working feature to separate the sections in different files to allow separation of packages. Improvements can be done afterwards, but it's your decision if and in which form it gets merged...

NeoRaider:
I don't think any of this is necessary. Components can just be provided by regular Lua modules, without any need to glob for files to load.

CodeFetch:
@NeoRaider The goal of this is to modularize the packages to allow to specify the packages that one wants in the site.mk file. Gluon is not modular enough at the moment to use it for different purposes than Freifunk. E.g. there should be a gluon-web-wifi-core package, a gluon-web-wifi-private-wlan, a gluon-web-wifi-outdoor-mode, a gluon-web-wifi-client, a gluon-web-wifi-mesh etc. package. Not one "provides all" package. The same for gluon-web-vpn-core. There should be a gluon-web-vpn-fastd-n2s, gluon-web-vpn-fastd-n2n, gluon-web-fastd-s2n etc. package. They should provide additional files that are e.g. being embedded as a section in the form of gluon-web-*-core. Is it clear what I want to achieve here? It's only possible with globbing.

NeoRaider:
Such globbing would only be useful in very specific parts of the web UI. I don't think such a facility should be a default feature of gluon-web, but something that is explicitly used by a model script.

Also, we don't really support multiple VPNs or multiple mesh protocols at the same time, so rather than globbing, such modular models could just load a module with a special name, rather than globbing for multiple files (see the the optional OSM feature in `gluon-config-mode-geo-location/luasrc/lib/gluon/config-mode/wizard/0400-geo-location.lua` for an example). Globbing can still be a viable solution, but only in a few well-considered places.

(Picking up the discussion from #1329, I'm not sure if now it a good time for other projects to consider using Gluon as a base for their platform - I still hope we can revamp our network configuration to become more flexible eventually (see also the [network-rewrite milestone](https://github.com/freifunk-gluon/gluon/milestone/10), but my current plan for that involves a complete replacement of netifd, fw3, most of UCI, and possibly other tools with something better suited for our requirements.)

CodeFetch:
I understand your concerns, but I don't see any other way. At the moment we are using globbing already for the models. The concept can of course be improved. That's why I opened this as a RFC. What I'm thinking of is that we could have "core"-models/supramaps similar to the categories of OpenWrt's LUCI, but not with nested navigation, but a single form with multiple sections instead. That is cleaner, more modern UI design. At the moment we have e.g. the problem, that the state is getting lost when we switch to the Expert mode. There should be no different site for that. Instead the General setup page should be a category with the same tabbed navigation like in the Expert mode. And the expert forms' categories should be toggled using a Expert mode checkbox. It's hard to explain with words only. I'm going to make a design concept so that I can show you, what I want to achieve. Another issue that would be solved with this approach is responsive design. On my devices with a smaller display I have to scroll horizontally to see all links. It's getting too much. There should only be e.g. "General", which is the only visible category when the expert mode checkbox is disabled, "WLAN", "Ports", "VPN", "Plugins" and "System". Where "Plugins" should be used by community-maintained packages that are too specific to be upstreamed.

CodeFetch:
I decided to rework Gluon's web interface based on the new LuCI2 concept. A helping hand would be awesome.

PR #1665, opened 2019-03-09
docs: faq: add note for supported hardware

CodeFetch:
A table with hardware supported by Gluon can be found on the OpenWrt Wiki. This commit adds a corresponding FAQ entry.

### DIFF
--- a/docs/user/faq.rst
+++ b/docs/user/faq.rst
@@ -1,10 +1,20 @@
 Frequently Asked Questions
 ==========================
 
+.. _faq-hardware:
+
+What hardware is supported?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+A table with hardware supported by Gluon can be found on the `OpenWrt Wiki`_.
+If you want to find out if your device can potentially be supported 
+have a look at :doc:`../dev/hardware` for detailed hardware requirements.
+
+.. _OpenWrt Wiki: https://openwrt.org/toh/views/toh_gluon_supported
+
 .. _faq-dns:
 
-DNS does not work on the nodes
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Why does DNS not work on the nodes?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Gluon nodes will ignore the DNS server on the WAN port for everything except
 the mesh VPN, which can lead to confusion.
@@ -18,8 +28,8 @@ in this case, the *radvd* is only used to announce the DNS server.
 
 .. _faq-mtu:
 
-What is a good MTU on the mesh-vpn
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+What is a good MTU on the mesh-vpn?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Setting the MTU on the transport interface requires careful consideration, as
 setting it too low will cause excessive fragmentation and setting it too high


mweinelt:
A machine-readable file in the Gluon repository is something I would prefer for this scenario. Especially since several third-party tools maintain machine-readable data independently from Gluon already.

CodeFetch:
@mweinelt What do you mean with "machine-readable"? Every HTML-table is easily machine-readable. XML was for years the standard until JSON came up. I'd just put a ["Giant Table"](https://sphinx-rtd-theme.readthedocs.io/en/stable/demo/lists_tables.html) in the docs and add a note to the device support pull-request template to extend the table in the docs by the hardware specifications. That way rubo does not need to care for the OpenWrt table anymore. Is there a common use-case for a need of a machine-readable table? I assume everyone who needs such data can write a script that extracts it within at most 5 Minutes. Are you thinking about  e.g. something like this? https://hannover.freifunk.net/download.html

mweinelt:
Yes, that is exactly one of the use cases I see.

CodeFetch:
@mweinelt What I'd like to see is a human-readable table in the docs or at least a link to a file in the docs in a human-readable format. JSON is not very human-readable. For you machine-readability seems to be a priority, to me it's human-readability. How can we achieve this without unnecessary redundancy?

mweinelt:
A human-readable format could automatically be created and updated from machine-readable data with a webhook or a CI job, just like we render the current documentation. The machine-readable data would in my opinion simply be much more beneficial.

CodeFetch:
Hm... Webhooks and CI jobs are a bit heavy. I'd prefer a simple shell script as a git pre-commit hook that updates the documentation...

mweinelt:
A script to do the rendering would be part of the solution either way. We can certainly revisit automation later on if we think it's beneficial.

My current thought on git hooks vs pipeline is:
The disadvantage of git hooks would be their need to be set them up locally for every maintainer, whereas a CI job would reduce the burden on every following contributor.

CodeFetch:
I have never used Github's CI, but only Gitlab's and only for unit testing and deployment - not for adding additional commits or something like that (with a bot?). Maybe I don't understand how you want to achieve this. Isn't Github's CI a premium feature?

We could set the hooks directory using `git config core.hooksPath scripts/hooks` on "make update" and introduce "make init" for those who don't want to load all dependencies. That looks pretty simple to me compared to CI.

mweinelt:
GitHub has integrations with for example Travis-CI ans Circle-CI which both have a free plan, which would be find for such jobs.

But as I stated above the automation part is something we could revisit later.

rubo77:
We could use the machine readable openwrt database.

Then we would omit redundancy again.

It is easy to mark a device as Gluon compatible in the openwrt db

CodeFetch:
@rubo77 I've thought about the machine-readable file. There are some reasons why I think it is a good idea. We can get rid of the target.mk-files respectively generating them automatically which can lead to a cleaner design in the long-term. We could add build flags for packages which they can use to extract data from the "targets.json". For example the antenna offsets of Ubiquity devices can be defined there and an additional package can take care of setting them correctly. Besides that we should not rely on an external resource for a supported device table, information like this does not seem to be that important for OpenWrt like it is for us. It is great that we have the OpenWrt supported hardware table. This makes it easy to create such a targets.json file, but in the future I think we should edit it as a local resource.

But my time is sparse and I only wanted to merge a FAQ entry in the docs and now I'm creating a JSON-file with targets-data and a parser in bash that generates a docs entry and the targets.mk-files automatically from that file.

rotanid:
i'm fine with this doc change, there may be more advanced / better solutions but they can still be done later.

Issue #1661, opened 2019-03-03
add support for triple radio configurations


blocktrron:
Currently, when using a device with three radios, Gluon does not create client and mesh VAPs on the third radio.

With the emerge of triband devices, Gluon should support the third radio and maybe bring it up in a sensible configuration. I propose inn case of two 5GHz radios: 

 - one 5GHz radio used for meshing on indoor channels
 - one used for serving clients on DFS / radar channels

With the outdoor mode deactivated, the mesh radio should be deactivated. This is due to most triband devices have a limited range of selectable channels for each radio.

I don't consider this issue as urgent. The only device currently affected is the OpenMesh A62 which shouldn't be very common.

Personally i ran into this very problem when testing an AVM Repeater 3000 with Gluon.

Djfe:
List of Hardware that would profit from this (not all of them have openwrt support but I checked whether their target has support in openwrt master):
* OpenMesh A62
* Plasma Cloud PA2200
* AVM Fritz!Repeater 3000 and 6000
* AVM Fritz!Box 4060
* LibreRouter v1
* Linksys MR8300/EA8300/(EA7500)
* Xiaomi AX3600 (und AX9000)
* GL.iNet GL-B2200
* (Linksys Velop (WHW01 v1 und WHW03 v2))

This wasn't done to raise any urgency but just to raise awareness which kind of devices this affects.
Even if it takes another year, I don't care too much since there are other devices than these available that we can recommend. But a mesh network being assigned to it's own radio sounds very appealing to me. I'm looking forward to testing whatever comes out of this :)

maurerle:
@blocktrron @rotanid this is fixed with #3563

PR #1666, opened 2019-03-09
[RFC] add support for dual 5GHz radio configurations

blocktrron:
This commit adds support for devices with the following properties (#1661):
 - Features two 5GHz radios
 - Each of the two radios only supports channel 36-64 or 100-144

This is the case for most (if not all) triple-radio devices based on the IPQ40xx platform. Examples are the _OpenMesh A62_, _ASUS Lyra_, _AVM FRITZ!Repeater 3000_.

On first boot, the device will be configured like following:

 - One 5GHz radio on DFS channels
   - Mesh deactivated
 - One 5GHz radio on indoor (default) channel
   - Client disabled

The channel range for the 5GHz is obtained from the outdoor chanlist introduced by the outdoor mode. This site.conf parameter could therefore be renamed (currently not part of this PR).

Speaking of outdoor mode: In case of activated outdoor mode, the mesh radio will not work due to DFS channels marked as unusable by the driver.

This PR could be extended to support devices with two 5GHz radios which support the entire frequency spectrum (similar to how the radio distribution https://github.com/freifunk-gluon/gluon/commit/888cddb662847c3c082f150fb2c9ad8d95f62324 works).

### DIFF
--- a/package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
+++ b/package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
@@ -3,6 +3,8 @@
 local util = require 'gluon.util'
 
 local uci = require('simple-uci').cursor()
+local sysconfig = require 'gluon.sysconfig'
+local iwinfo = require 'iwinfo'
 
 
 local function is_disabled(config, name)
@@ -13,6 +15,8 @@ local function is_disabled(config, name)
 	return config.disabled(false)
 end
 
+local has_client_radio = false
+
 util.foreach_radio(uci, function(radio, index, config)
 	local radio_name = radio['.name']
 
@@ -21,6 +25,10 @@ util.foreach_radio(uci, function(radio, index, config)
 
 	local ap = config.ap
 	local disabled = is_disabled(ap, name)
+
+	if not util.supports_channel(radio, config.channel()) then
+		has_client_radio = true
+	end
 
 	uci:delete('wireless', name)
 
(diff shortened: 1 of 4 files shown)


rotanid:
thanks. i added a blocked label as this seems to depend on #1329 which is waiting for review.

targets/ramips-mt7621
```diff
@@ -35,9 +35,7 @@ fi
 device ubnt-erx ubnt-erx
 packages '-hostapd-mini'
 factory
-sysupgrade '.tar'
```
bobcanthelpyou: Why is the sysupgrade of ramips-mt7621 touched in that context?

mweinelt: Because the patchset was rebased against next, not master.

CodeFetch:
@mweinelt Why is this still blocked? Why can't it be part of v2019.1?

mweinelt:
Not blocked but scheduled for v2019.2, we will not have any dual 5 GHz radio devices before that anyway.

mweinelt:
Hey @blocktrron, time to rebase this.

blocktrron:
@mweinelt I'm planning on proposing the abstraction of the wireless configuration, as we now somewhat have to split the user intent from the runtime configuration to not drown in the logic.

CodeFetch:
Great news! 
Maybe you remember I told you about qboost a while ago on IRC.
It's an undocumented TDMA feature implemented in the ath10k firmware which seems to be Ubiquiti's AirMax.
DD-WRT folks dug deeper and added [experimental support](https://svn.dd-wrt.com/changeset/41499) for it. This means we can create an OpenWrt patch and do some testing.
Three years ago I created a [hacky fork](https://github.com/freifunk-graviton/graviton) of Gluon for testing some TDMA features using ath9k hMAC, but it was disappointing. When I find the time I'm going to create a new version so that we can get rid of closed-source firmware in our mesh networks completely. Thus it would be nice if the second 5 GHz radio could connect with the backbone PtmP nodes using qboost as a slave (PtmP nodes will act as bridges using WDS like AirMax devices).
Just wanted to tell you to keep that in mind.

blocktrron:
This needs to be rewritten to take in mind config abstraction as well as the additional MAC space allocation. I'm closing this for now and will hopefully one day find time to get back to this.

PR #1667, opened 2019-03-09
docs: remove unavailable devices (Picostation M5 and TL-WR710N v1/v2.1)

bobcanthelpyou:
Can't find a device called Picostation M5. Only a Picostation M2 is available.

TL-WR710N v1, v2.1 are part of ar71xx-generic not ar71xx-tiny
```
$ grep wr710n targets/*
targets/ar71xx-generic:device tp-link-tl-wr710n-v1 tl-wr710n-v1
targets/ar71xx-generic:device tp-link-tl-wr710n-v2.1 tl-wr710n-v2.1
targets/ar71xx-tiny:device tp-link-tl-wr710n-v2 tl-wr710n-v2
```

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -223,7 +223,7 @@ ar71xx-generic
   - Loco M2/M5 XW
   - Nanostation M2/M5
   - Nanostation M2/M5 XW
-  - Picostation M2/M5
+  - Picostation M2
   - Rocket M2/M5
   - Rocket M2/M5 Ti
   - Rocket M2/M5 XW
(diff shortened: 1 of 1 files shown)


mweinelt:
Great find, please be a bit more thorough on the explanation in the commit messsage, use a summary in the first line, add explanations in separate lines below that.

mweinelt:
Merged with slight adapation to the picostation explanation. Thanks!

bobcanthelpyou:
Thx @mweinelt for the git commit reference. I am lazy that i put the explanation in the merge requests mostly and in the first line of the commit message. I am never sure what to write, and which are the best fitting words. Thanks for the precise words.

PR #1669, opened 2019-03-13
ipq806x: add support for NETGEAR R7800

blocktrron:
  - [x] webinterface
  - [x] tftp
  - [x] other: <nmrpflash utility
    ```
    root@64367-nighthawk-x4s:~# lua -e 'print(require("platform_info").get_image_name())'
    netgear-nighthawk-x4s-r7800
    ```
- wired network
- wifi (if applicable)
- led mapping
  - power/sys led
  - radio leds
    - [ ] should show activity
  - switchport leds

### DIFF
--- a/targets/ipq806x
+++ b/targets/ipq806x
@@ -1,7 +1,31 @@
-ATH10K_PACKAGES='-kmod-ath10k kmod-ath10k-ct -ath10k-firmware-qca99x0 ath10k-firmware-qca99x0-ct'
+# The QCA9980 was discontinued by Qualcomm. It didn't receive any firmware-update for over 4 years.
+# See https://github.com/kvalo/ath10k-firmware/tree/master/QCA99X0/hw2.0
+# 802.11s was never implemented for the chip's firmware. It will most likely be broken forever.
+# The QCA9984 on the other hand works fine for 11s meshes on both bands.
+
+QCA9980_PACKAGES='-kmod-ath10k kmod-ath10k-ct -ath10k-firmware-qca99x0 ath10k-firmware-qca99x0-ct'
+QCA9984_PACKAGES='kmod-ath10k -kmod-ath10k-ct ath10k-firmware-qca9984 -ath10k-firmware-qca9984-ct'
 
 
+#
+# QCA9980
+#
+
+if [ "$BROKEN" ]; then
 # TP-Link
-
 device tp-link-archer-c2600 tplink_c2600
-packages $ATH10K_PACKAGES
+packages $QCA9980_PACKAGES
+fi
+
+#
+# QCA9984
+#
+
+# -- ToDo --
+# The BROKEN flag for the target can be removed when switching to an OpenWrt base newer than 18.06,
+# as in 18.06 the LEDs are not yet functional. This is already fixed in the current OpenWrt master.
+
+# NETGEAR
+device netgear-nighthawk-x4s-r7800 netgear_r7800
+factory .img
+packages $QCA9984_PACKAGES
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac
@@ -49,6 +49,8 @@ elseif platform.match('ar71xx', 'generic', {'archer-c5', 'archer-c58-v1',
 elseif platform.match('ipq40xx', nil, {'avm,fritzbox-4040',
                                        'openmesh,a42', 'openmesh,a62'}) then
   table.insert(try_files, 1, '/sys/class/net/eth0/address')
+elseif platform.match('ipq806x', nil, {'netgear,r7800'}) then
+  table.insert(try_files, 1, '/sys/class/net/eth1/address')
 elseif platform.match('ramips', 'mt7621', {'dir-860l-b1'}) then
   table.insert(try_files, 1, '/sys/class/ieee80211/phy1/macaddress')
 end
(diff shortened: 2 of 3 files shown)


targets/ipq806x
```diff
@@ -1,7 +1,29 @@
-ATH10K_PACKAGES='-kmod-ath10k kmod-ath10k-ct -ath10k-firmware-qca99x0 ath10k-firmware-qca99x0-ct'
+# The QCA9980 is discontinued from Qualcomm. It didn't receive any firmware-update since >4 years.
```
mweinelt: "The QCA9980 **was** discontinued **by** Qualcomm. It didn't receive any firmware update **for** over 4 years."

Also add the link to the firmware repository for reference. 
https://github.com/kvalo/ath10k-firmware/tree/master/QCA99X0/hw2.0

targets/ipq806x
```diff
@@ -1,7 +1,29 @@
...
+# Any kind of meshing was never implemented for this chip. It is (most likely) broken forever.
```
mweinelt: What does "Any kind of meshing" represent? Only 11s or also IBSS? Be explicit here. 
Also no need to put "most likely" in brackets.

targets/ipq806x
```diff
@@ -1,7 +1,29 @@
...
+# The QCA9984 on the other hand works fine for 11s meshes on both bands.
 
+QCA9980_PACKAGES='-kmod-ath10k kmod-ath10k-ct -ath10k-firmware-qca99x0 ath10k-firmware-qca99x0-ct'
+QCA9984_PACKAGES='kmod-ath10k -kmod-ath10k-ct ath10k-firmware-qca9984 -ath10k-firmware-qca9984-ct'
 
-# TP-Link
 
+#
```
mweinelt: If it is broken indefinately we can either decide to drop and not accept QCA9980 radios completely, or use this file to remind us of tried-and-failed devices, in which case I would like an if broken block surrounding all QCA9980 devices.

Yes I know the target is currently marked as BROKEN, but that makes unbreaking it easier at a later point, when everything is already reasonably documented.

targets/ipq806x
```diff
@@ -1,7 +1,29 @@
...
+packages $QCA9980_PACKAGES
+
+
+#
+# QCA9984
+#
+
+# -- ToDo --
+# The BROKEN flag for QCA9984 can be removed when switching to an OpenWrt base newer than 18.06,
+# as in 18.06 the LEDs are not yet functional. This is already fixed in the current OpenWrt master.
```
mweinelt: If the LEDs are currently broken for the radio add a BROKEN flag around QCA9984 and unflag the target as BROKEN in general, because the reason in `targets.mk` only reflects on QCA9980.

blocktrron: As no device in this target is currently fully functional, i would leave the target as a whole broken. For easier unbreaking however, i would add a BROKEN clause explicitly for the QCA9980. Sounds good?

targets/ipq806x
blocktrron: I do net own such a device, so i can't test them. The Firmware definitely does not support 11s, however it might support IBSS but as a single VAP (no additional client VAP). I can try to explain the situation more, but being 100% explicit is not possible for me here.

targets/ipq806x
mweinelt: But the broken remark for the target does not apply for QCA9984, it is misleading at best now.

https://github.com/freifunk-gluon/gluon/blob/master/targets/targets.mk#L25

mweinelt: Ok, unbreaking the target seems like the wrong move here, sorry. But updating the remark seems to be in order.

targets/ipq806x
CodeFetch: Why don't you use the Candelatech firmware? It supports QCA99X0. Can you please explain why you select the outdated QCA firmware at all?

CodeFetch: Okay I read the thread about ipq806x. The ct-driver was updated some months ago with changes to QCA99X0. Can you please test the router with ct-firmware only? Don't select both firmwares. And if there are issues, please fill a bug report at CT. This driver is being actively developed.

blocktrron: @CodeFetch I do not own a QCA99x0 based device. They can be considered the AR1A of Wave 2 chips so unless someone funds me i won't invest any time there. We weren't able to bring up any mesh/AP combination up at all with any firmware/driver combination some months back.

For the QCA99x4: The -ct firmware crashed for me multiple times within 24 hours so i will stick with the one from QCA here. Same goes for the driver. I see no reason to switch to the -ct one when the one from QCA works flawlessly for our scenario.

CodeFetch: Thank you for the hint with the crashing. I assume you would get SWBA overrun messages, too. Then it is a known bug.

blocktrron:
PR updated.

targets/ipq806x
rotanid: @blocktrron you missed to change this with your recent update

targets/ipq806x
```diff
@@ -1,7 +1,31 @@
-ATH10K_PACKAGES='-kmod-ath10k kmod-ath10k-ct -ath10k-firmware-qca99x0 ath10k-firmware-qca99x0-ct'
+# The QCA9980 is discontinued from Qualcomm. It didn't receive any firmware-update since >4 years.
+# See https://github.com/kvalo/ath10k-firmware/tree/master/QCA99X0/hw2.0
+# 802.11s was never implemented for the chips firmware. It is most likely broken forever.
```
rotanid: chip's
i'm also not sure if "It will most likely be broken forever." is better?

blocktrron: fixed

targets/ipq806x
blocktrron: fixed

PR #1678, opened 2019-03-17
ramips-mt76x8: add support for TP-Link TL-WR902AC v3

ambassador86:
  * [x]  tftp - use gluon-...-tp-link-tl-wr902ac-v3-bootloader.bin file
    * [x]  must keep/forget configuration (if applicable) _think `sysupgrade [-n]` or `firstboot`_
  * [x]  must have working autoupdate   _usually means: gluon profile name must match image name_
`root@ffc-d80d170318c4:~# lua -e 'print(require("platform_info").get_image_name())'
tp-link-tl-wr902ac-v3
root@ffc-d80d170318c4:~# 
`

* wired network  
  * [X]  should support all network ports on the device
  * [X]  must have correct port assignment (WAN/LAN) -- WAN only
* wifi
  * [X]  association with AP must be possible  
  * [X]  association with 802.11s mesh must be working  
  * [X]  ap/mesh mode must work in parallel 
* led mapping
  * power/sys led
    * [X]  lit while the device is on
    * [X]  should display config mode blink sequence
  * radio leds
    * [x]  should show activity
  * switchport leds
    * [?]  should show link state and activity -- the LED at the RJ45 connector shows the link state but no activity
* [X]  wps button must return device into config mode
* [X]  primary mac should match address on device label (or packaging) (https://gluon.readthedocs.io/en/latest/dev/hardware.html#notes)

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -345,6 +345,7 @@ ramips-mt76x8
 
   - TL-MR3420 v5
   - TL-WR841N v13
+  - TL-WR902AC v3
   - Archer C50 v3
   - Archer C50 v4
 
--- a/targets/ramips-mt76x8
+++ b/targets/ramips-mt76x8
@@ -43,6 +43,12 @@ device('tp-link-tl-wr841n-v13', 'tl-wr841n-v13', {
 	},
 })
 
+device('tp-link-tl-wr902ac-v3', 'tplink_tl-wr902ac-v3', {
+	factory = false,
+	extra_images = {
+		{'-squashfs-tftp-recovery', '-bootloader', '.bin'},
+	},
+})
 
 -- VoCore 2
 


rotanid:
thanks for your contribution.
the problem is the not-yet-working LED. if someone flashes the device with this, the led won't work even after an update to a new version unless the device is completely reset, like with firstboot.
additionally, this device would be BROKEN even with a backport for the LED, which is against our device integration policy (having a backport for a BROKEN device).
conclusion:
this has to wait until Gluon switches to the newer OpenWrt branch.

ambassador86:
Hello @rotanid,
thanks for your comment. At the moment the device is not as usable as it could be. Waiting until gluon uses a more recent version of OpenWRT sound reasonable to me. Hopefully the device can be used with both radios then too.

mweinelt:
This PR needs to be updated to work with the new lua based target system.

CodeFetch:
@ambassador86 

```
device('tp-link-tl-wr902ac-v3', 'tplink_tl-wr902ac-v3', {
	broken = true, -- no 5GHz WLAN, eth-LED not working
	factory = false,
	extra_images = {
		{'-squashfs-tftp-recovery', '-bootloader', '.bin'},
	},
})
```

ambassador86:
Hello @CodeFetch ,
I updated the PR accordingly

mweinelt:
@rotanid Thanks for the update of the title, but that does not change the commit message.

@ambassador86 Please update the commit message to add the device revision.

rotanid:
you are right, but i think any work on this PR can wait until Gluon is based on OpenWrt 19.07 branch and then the broken flag may also be obsolete

ambassador86:
I will build a new image for this device and postpone the further work then.

mweinelt:
@ambassador86 Gluon master branch is now tracking OpenWrt 19.07, so you should probably be fine to continue here.

ambassador86:
Thanks @mweinelt for the hint. I will test it this evening.

ambassador86:
Hello @mweinelt,
the current state of the device is updated above. I have removed the broken flag as the 5GHz is now working and usable. The LED at the RJ45 Connector is showing the link state now. The WAN Led remains dark. Depends on your opinion if this is good enough to be merged without broken flag. It is also okay to have it in place.

rotanid:
> The WAN Led remains dark. Depends on your opinion if this is good enough to be merged without broken flag. It is also okay to have it in place.

i think this depends on things like "is it fixable at all" "is there a patch upstream" and "is it fixable via sysupgrade without losing config later".
@blocktrron what do you think?

blocktrron:
@rotanid I will have a look at the GPL code later and see whether or not this is an easy fix.

Otherwise - a WAN LED is only set on firstboot, thus it would still need the broken flag if it is non-working.

rotanid:
@blocktrron did you have a look?

@ambassador86 please adjust your commit message text, your reason for BROKEN is not up-to-date

rotanid:
@ambassador86 your commit message still contains a remark about 5hgz not being usable, but as far as i understood, it is now OK in OpenWrt 19.07.x or Gluon Master?

@blocktrron ping regarding LED

ambassador86:
Hello @rotanid thanks for keeping heads up. I corrected the commit message as OpenWrt 19.07 is enough for this router to work.  In #1866 the 5GHz is also stable.

rotanid:
we decided to reduce the number of devices being merged as blocked, if there's a good chance to fix the reason before merging.
therefore, i'm applying the blocked-label and we welcome an upstream patch for the LED which can be included in this PR after it has been merged upstream.

rotanid:
@ambassador86 the LED might work now with the latest openwrt bump done by @blocktrron - probably you are already aware of it

blocktrron:
@ambassador86 We might have another problem here - Can you verify which MAC address is used on the 5GHz part? The MT7610 (from my observation) seems to have a generic Ralink MAC address, which will lead to problems as soon as two devices with the same MAC participate in a mesh.

(Copy from https://github.com/freifunk-gluon/gluon/pull/1866#issuecomment-570402437)

ambassador86:
@blocktrron everthing is working as expected now. I added the device to the supported list and removed the broken flag.

PR #1460, opened 2018-07-09
Support ingress bandwidth shaping with tunneldigger

kaechele:
Fix for #1447 

Signed-off-by: Felix Kaechele <felix@kaechele.ca>

Issue #1447, opened 2018-06-26
simple-tc reduces bandwidth far below given limit


RalfJung:
I have set
```
uci set simple-tc.mesh_vpn.limit_ingress=10000
uci set simple-tc.mesh_vpn.limit_egress=1000
uci commit simple-tc && /etc/init.d/tunneldigger restart
```
With these settings, when I do a speedtest, I get <3 MBit/s download speed (and ~1 MBit/s upload as expected).

With an ingress limit of `30000`, download speed increases to ~5 MBit/s.

When I turn off simple-tc, I get 25 MBit/s for downloads (my DSL line has 30 MBit/s so that's as expected).

Seems like simple-tc reduces the download speed much further than it should? I had a look at the device load during these tests, but there was always >50% idle.

This is with a WR841N/ND v10 and a firmware based on Gluon v2018.1.8, in a tunneldigger/l2tp-based network.

NeoRaider:
TL;DR: ingress filters are bad, the behaviour is known.

We don't have any way to control how fast a VPN peer sends traffic; such a feature would need to be implemented in the VPN software, or possibly a layer on top of the VPN that transmits the limit and installs an egress filter on the other side.

The only thing our ingress filter does is to drop packets when we are over the limit, and hope that TCP connections behave sanely and reduce throughput. For non-TCP protocols, the situation is even worse, as they often do not have any kind of control (so they will simply experience packet loss when the ingress filter kicks in). And even for TCP, behaviour in the presence of packet loss can be very bad with some TCP implementations, especially when there are multiple parallel TCP streams.

RalfJung:
I see. So what one would want is for tunneldigger to tell the broker on the gateway side to please do egress filtering. I can see how that makes more sense. :)
I think I even saw TODOs for that somewhere in the tunneldigger sources or so.

I can look into adding support for that in tunneldigger. Does simple-tc have support for cooperating with the VPN client in that regard?

kaechele:
Tunneldigger already has that support (-L flag on the client makes the server set the requested downstream bandwidth limit). It's just that the support on the Gluon side is missing to actually pass that flag to tunneldigger when
1. a limit is set in config mode and
2. tunneldigger is the method used for VPN.

RalfJung:
Awesome, thanks a lot!

PR #1682, opened 2019-03-23
Bugfix: gluon-config-mode-mesh-vpn: 0300-mesh-vpn.lua limit_ingress 

2tata:
fix limit_ingress default value in case of default disabled simple-tc for tunneldigger.

In case if 'limit_bw_down' set for tunneldigger should simple-tc still be enabled?

If simple-tc enabled by default via site.conf the script '500-mesh-vpn' from 'gluon-mesh-vpn-core' will set 'limit_ingress' to 0 and also enabled simple-tc. Does 0 have a special meaning in this situation like "ignore it"?

Bug appears here: #1460 
Related to #1447 

Thanks to @lrnzo 

### DIFF
--- a/package/gluon-config-mode-mesh-vpn/luasrc/lib/gluon/config-mode/wizard/0300-mesh-vpn.lua
+++ b/package/gluon-config-mode-mesh-vpn/luasrc/lib/gluon/config-mode/wizard/0300-mesh-vpn.lua
@@ -50,7 +50,7 @@ return function(form, uci)
 	o:depends(limit, true)
 	o.default = uci:get("simple-tc", "mesh_vpn", "limit_ingress")
 	if has_tunneldigger then
-		-- Check if limit_bw_down exsist if not take value vom limit_ingress
+		-- Check if limit_bw_down exsists. If not, take value from limit_ingress
 		local limit_bw_down = uci:get("tunneldigger", "mesh_vpn", "limit_bw_down")
 		if limit_bw_down ~= nil then
 			o.default = limit_bw_down
--- a/package/gluon-config-mode-mesh-vpn/luasrc/lib/gluon/config-mode/wizard/0300-mesh-vpn.lua
+++ b/package/gluon-config-mode-mesh-vpn/luasrc/lib/gluon/config-mode/wizard/0300-mesh-vpn.lua
@@ -50,7 +50,7 @@ return function(form, uci)
 	o:depends(limit, true)
 	o.default = uci:get("simple-tc", "mesh_vpn", "limit_ingress")
 	if has_tunneldigger then
-		-- Check if limit_bw_down exsists. If not, take value from limit_ingress
+		-- Check if limit_bw_down exists. If not, take the value from limit_ingress
 		local limit_bw_down = uci:get("tunneldigger", "mesh_vpn", "limit_bw_down")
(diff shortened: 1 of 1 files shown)


lrnzo:
I still do not understand why the check for ```has_tunneldigger``` is applied on the ```limit_ingress``` only and not on the ```limit_egress```. Maybe I fundamentally do not understand something basic here?

2tata:
On 3/23/19 10:32 PM, lrnzo wrote:
> I still do not understand why the check for ```has_tunneldigger``` is applied on the ```limit_ingress``` only and not on the ```limit_egress```. Maybe I fundamentally do not understand something basic here?
> 

The best is you have a look at #1447 . There is some description.

vg
Tarek

package/gluon-config-mode-mesh-vpn/luasrc/lib/gluon/config-mode/wizard/0300-mesh-vpn.lua
```diff
@@ -48,10 +48,13 @@ return function(form, uci)
 
 	o = s:option(Value, "limit_ingress", pkg_i18n.translate("Downstream (kbit/s)"))
 	o:depends(limit, true)
+	o.default = uci:get("simple-tc", "mesh_vpn", "limit_ingress")
 	if has_tunneldigger then
-		o.default = uci:get("tunneldigger", "mesh_vpn", "limit_bw_down")
-	else
-		o.default = uci:get("simple-tc", "mesh_vpn", "limit_ingress")
+		-- Check if limit_bw_down exsist if not take value vom limit_ingress
```
rotanid: "... exists. If not, take the value from ..."

2tata: I guess it was late in the night ...

rotanid: just copy&paste my suggestion, because two issues are still there ;)

NeoRaider:
I'm not happy with the solution yet, as there is no "single source of truth" for the configured limit anymore. When the user sets a limit in a fastd-based firmware and later upgrades to a tunneldigger-based firmware, the limit is not migrated.

There should be a single UCI value that is read and written by gluon-config-mode-mesh-vpn for both fastd and tunneldigger; on writes, additional VPN-specific values like the tunneldigger limit might be updated as well.

rotanid:
good point... @2tata maybe you can improve this?

2tata:
I agree with @NeoRaider I will do an another MR this week to simplify it. Still the above question: 

> If simple-tc enabled by default via site.conf the script '500-mesh-vpn' from 'gluon-mesh-vpn-core' will set 'limit_ingress' to 0 and also enabled simple-tc. Does 0 have a special meaning in this situation like "ignore it"?

NeoRaider:
@2tata No, that logic is simply broken. Another issue in the same few lines of code: tunneldigger.mesh_vpn.limit_bw_down and simple-tc.mesh_vpn.limit_ingress are overwritten when site.mesh_vpn.bandwidth_limit.enabled is true, regardless of a previously set value.

PR #1684, opened 2019-03-24
gluon-check-connection: extracted from gluon-scheduled-domain-switch

rubo77:
Extract the connection-check for example, to reuse it in gluon-offline-ssid package #1649 .

The check so far only pings some definable IP6 addresses. In the old ssid-changer, we only used a batctl check if a gateway is reachable instead. It seems like the ping solution is the better, but are there any other suggestions?

We could also add a TQ-check to the gateway as in the offline-ssid package, so it is treated as offline, if the TQ sinks below a threshold.

### DIFF
--- a/docs/package/gluon-scheduled-domain-switch.rst
+++ b/docs/package/gluon-scheduled-domain-switch.rst
@@ -24,8 +24,8 @@ domain_switch : optional (needed for domains to switch)
 
 check_connection :
     targets :
-        - array of IPv6 addresses which are probed to determine if the node is
-	  connected to the mesh
+        - array of IPv6 addresses which are probed to determine if the node can
+          reach any time-servers to acquire the correct time
 
 Example::
 
@@ -41,3 +41,6 @@ Example::
       '2001:4860:4860::8844',
     },
   },
+
+Note: the package ``gluon-check-connection`` is installed as dependency but it can
+also be used stand-alone, e.g. to detect if a connection to the internet is working.
--- a/docs/package/gluon-scheduled-domain-switch.rst
+++ b/docs/package/gluon-scheduled-domain-switch.rst
@@ -22,11 +22,6 @@ domain_switch : optional (needed for domains to switch)
     switch_time :
         - UNIX epoch after which domain will be switched
 
-check_connection :
-    targets :
-        - array of IPv6 addresses which are probed to determine if the node can
-          reach any time-servers to acquire the correct time
-
 Example::
 
   domain_switch = {
@@ -42,5 +37,6 @@ Example::
     },
   },
 
-Note: the package ``gluon-check-connection`` is installed as dependency but it can
-also be used stand-alone, e.g. to detect if a connection to the internet is working.
+Note: the package ``gluon-check-connection`` is installed as dependency so you
+have to define ``check_connection`` too. A node is considered online if any of
+the specified ip addresses can be reached via ping.
--- a/docs/package/gluon-scheduled-domain-switch.rst
+++ b/docs/package/gluon-scheduled-domain-switch.rst
@@ -21,7 +21,9 @@ domain_switch : optional (needed for domains to switch)
         - amount of time without reachable gateway to switch unconditionally
     switch_time :
         - UNIX epoch after which domain will be switched
-    connection_check_targets :
+
+check_connection :
+    targets :
         - array of IPv6 addresses which are probed to determine if the node is
 	  connected to the mesh
 
@@ -31,7 +33,10 @@ Example::
     target_domain = 'new_domain',
     switch_after_offline_mins = 120,
     switch_time = 1546344000, -- 01.01.2019 - 12:00 UTC
-    connection_check_targets = {
+  },
+
+  check_connection = {
+    targets = {
       '2001:4860:4860::8888',
       '2001:4860:4860::8844',
     },
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -419,6 +419,22 @@ autoupdater \: package
     All configured mirrors must be reachable from the nodes via IPv6. If you don't want to set an IPv6 address
     explicitly, but use a hostname (which is recommended), see also the :ref:`FAQ <faq-dns>`.
 
+.. _user-site-check_connection:
+
+check_connection \: optional
+    ``targets`` defines an array of IPv6 addresses which are probed for
+    connectivity, for example determine if the node can reach any time-servers
+    to acquire the correct time. This is e.g. used for the
+    *gluon-scheduled-domain-switch* package
+    ::
+
+      check_connection = {
+        targets = {
+          '2001:4860:4860::8888',
+          '2001:4860:4860::8844',
+        },
+      },
+
 .. _user-site-config_mode:
 
 config_mode \: optional
--- a/package/gluon-check-connection/Makefile
+++ b/package/gluon-check-connection/Makefile
@@ -1,0 +1,13 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-check-connection
+PKG_VERSION:=1
+
+include ../gluon.mk
+
+define Package/gluon-check-connection
+  TITLE:=Checks if a node can ping definable targets
+  DEPENDS:=+gluon-core
+endef
+
+$(eval $(call BuildPackageGluon,gluon-check-connection))
--- a/package/gluon-check-connection/check_site.lua
+++ b/package/gluon-check-connection/check_site.lua
@@ -1,0 +1,1 @@
+need_string_array_match({'check_connection', 'targets'}, '^[%x:]+$')
--- a/package/gluon-check-connection/luasrc/usr/bin/gluon-check-connection
+++ b/package/gluon-check-connection/luasrc/usr/bin/gluon-check-connection
@@ -4,17 +4,11 @@ local unistd = require 'posix.unistd'
 local util = require 'gluon.util'
 local site = require 'gluon.site'
 
-local offline_flag_file = "/tmp/gluon_offline"
+local offline_flag_file = "/var/gluon/check-connection/offline"
 local is_offline = true
 
--- Check if domain-switch is scheduled
-if site.domain_switch() == nil then
-	-- Switch not applicable for current domain
-	os.exit(0)
-end
-
 -- Check reachability of pre-defined targets
-for _, ip in ipairs(site.domain_switch.connection_check_targets()) do
+for _, ip in ipairs(site.check_connection.targets()) do
 	local exit_code = os.execute("ping -c 1 -w 10 " .. ip)
 	if exit_code == 0 then
 		is_offline = false
--- a/package/gluon-check-connection/luasrc/usr/bin/gluon-check-connection
+++ b/package/gluon-check-connection/luasrc/usr/bin/gluon-check-connection
@@ -4,7 +4,8 @@ local unistd = require 'posix.unistd'
 local util = require 'gluon.util'
 local site = require 'gluon.site'
 
-local offline_flag_file = "/var/gluon/check-connection/offline"
+local offline_flag_file_path = "/var/gluon/check-connection"
+local offline_flag_file = offline_flag_file_path + "/offline"
 local is_offline = true
 
 -- Check reachability of pre-defined targets
@@ -20,6 +21,10 @@ if is_offline then
 	-- Check if we were previously offline
 	if unistd.access(offline_flag_file) then
 		os.exit(0)
+	end
+	-- check if path exists
+	if !access(offline_flag_file_path, F_OK) then
+		stat.mkdir(offline_flag_file_path)
 	end
 	-- Create offline flag
 	local f = io.open(offline_flag_file, "w")
--- a/package/gluon-scheduled-domain-switch/Makefile
+++ b/package/gluon-scheduled-domain-switch/Makefile
@@ -7,7 +7,7 @@ include ../gluon.mk
 
 define Package/gluon-scheduled-domain-switch
   TITLE:=Allows scheduled migrations between domains
-  DEPENDS:=+gluon-core @GLUON_MULTIDOMAIN
+  DEPENDS:=+gluon-core +gluon-check-connection @GLUON_MULTIDOMAIN
 endef
 
 $(eval $(call BuildPackageGluon,gluon-scheduled-domain-switch))
--- a/package/gluon-scheduled-domain-switch/check_site.lua
+++ b/package/gluon-scheduled-domain-switch/check_site.lua
@@ -2,5 +2,4 @@ if need_table(in_domain({'domain_switch'}), nil, false) then
 	need_domain_name(in_domain({'domain_switch', 'target_domain'}))
 	need_number(in_domain({'domain_switch', 'switch_after_offline_mins'}))
 	need_number(in_domain({'domain_switch', 'switch_time'}))
-	need_string_array_match(in_domain({'domain_switch', 'connection_check_targets'}), '^[%x:]+$')
 end
--- a/package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
+++ b/package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-switch-domain
@@ -5,9 +5,10 @@ local unistd = require 'posix.unistd'
 local util = require 'gluon.util'
 local site = require 'gluon.site'
 
--- Returns true if node was offline long enough to perform domain switch
+-- Returns true if node was offline long enough to perform domain switch (set by gluon-check-connection)
 local function switch_after_min_reached()
-	if not unistd.access("/tmp/gluon_offline") then
+	local offline_flag_file = "/var/gluon/check-connection/offline"
+	if not unistd.access(offline_flag_file) then
 		return false
 	end
 
@@ -18,7 +19,7 @@ local function switch_after_min_reached()
 		return false
 	end
 
-	local f = util.readfile("/tmp/gluon_offline")
+	local f = util.readfile(offline_flag_file)
 	if f == nil then
 		return false
 	end


CodeFetch:
Personally I prefer the ping check as when there is a node which just has a bad connection, I don't want it to be shown as offline. A bad connection is better than no connection.

rubo77:
What does the variable `check_domain_switch` do in this line?

https://github.com/freifunk-gluon/gluon/blob/a536b23cc11a5750eb74c0aa79db6362af5f69c6/package/gluon-scheduled-domain-switch/check_site.lua#L1

rubo77:
I am building a firmware with this package right now, at least the build seems to work OK already

rotanid:
i added the RFC label, please request to remove it as soon as you are finished from your side.
BTW: github has a new function to mark a pull request as a DRAFT (i dont know if you can switch to DRAFT after creating a PR, if not you can use it in the future)

CodeFetch:
@skorpy2009 If you have a problem with that, articulate it instead of giving thumbs down. If you don't want bad connections to work, define a threshold using the supported rates. At the moment the lowest recommended value by Gluon is 6 MBit/s. Thus if the nodes can't handle out that rate the ping check won't work either. Using the offline-ssid for defining the threshold is a wrong approach.

rotanid:
@CodeFetch 45Mbit/s over a TQ of 30? in theory, maybe your values are correct, but in practise definitely far from reality, especially for non-TCP protocols.

CodeFetch:
@rotanid I've build long distance links with Ubiquiti hardware some years and I have achieved such TQ/throughput ratios in reality. The only reason why you don't achieve that with Gluon is because you have your retransmissions being handled on the gateway which introduces a high jitter.

rotanid:
ok, maybe, but we really went off-topic.

regarding the PR's changes, any opinions there?
to me it looks good, has it been build and run tested @rubo77 ?

rubo77:
I built it in our ci successful here:
 
https://freifunk.in-kiel.de/firmware/multidomain/2018.2.2.0-643~multi/

Maybe you can test if everything works?

docs/package/gluon-scheduled-domain-switch.rst
```diff
@@ -31,8 +33,11 @@ Example::
     target_domain = 'new_domain',
     switch_after_offline_mins = 120,
     switch_time = 1546344000, -- 01.01.2019 - 12:00 UTC
-    connection_check_targets = {
+  },
+
+  connection_check = {
```
CodeFetch: As you change the name of the site.conf field here, can you please change it to check_connection instead?
I assume that the package was named this way to allow a unified naming for check packages later. E.g. check-throughput, check-wan-connection, check-batman-iv-tq or whatever. Thus it would be nice if the config options start with check, too.

rotanid: good catch, the same order of those words should be used in the whole package @rubo77

docs/package/gluon-scheduled-domain-switch.rst
```diff
@@ -31,8 +33,11 @@ Example::
...
     switch_time = 1546344000, -- 01.01.2019 - 12:00 UTC
-    connection_check_targets = {
+  },
+
+  check_connection = {
+    targets = {
       '2001:4860:4860::8888',
       '2001:4860:4860::8844',
-    },
+    }
```
NeoRaider: Keep the trailing comma

package/gluon-check-connection/Makefile
```diff
@@ -0,0 +1,13 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-check-connection
+PKG_VERSION:=1
+
+include ../gluon.mk
+
+define Package/gluon-check-connection
+  TITLE:=Checks if a node can ping definable targets
+  DEPENDS:=+gluon-core @GLUON_MULTIDOMAIN
```
NeoRaider: Why @GLUON_MULTIDOMAIN?

package/gluon-check-connection/check_site.lua
```diff
@@ -0,0 +1,3 @@
+if need_table(in_domain({'check_connection'}), nil, false) then
+	need_string_array_match(in_domain({'check_connection', 'targets'}), '^[%x:]+$')
```
NeoRaider: I don't think `in_domain` necessarily makes sense here. You could also define a connection check target in the internet, in which case a single setting could be used for a while site.

package/gluon-check-connection/check_site.lua
```diff
@@ -0,0 +1,3 @@
+if need_table(in_domain({'check_connection'}), nil, false) then
```
NeoRaider: Remove the table check, the inner check implies the outer.

docs/package/gluon-scheduled-domain-switch.rst
```diff
@@ -31,8 +33,11 @@ Example::
     target_domain = 'new_domain',
     switch_after_offline_mins = 120,
     switch_time = 1546344000, -- 01.01.2019 - 12:00 UTC
-    connection_check_targets = {
+  },
+
+  check_connection = {
```
NeoRaider: I'd prefer to call this `connection_check`, not `check_connection`. (Also in the package name)

package/gluon-check-connection/luasrc/usr/bin/gluon-check-connection
```diff
@@ -7,14 +7,8 @@ local site = require 'gluon.site'
 local offline_flag_file = "/tmp/gluon_offline"
```
NeoRaider: Let's clean up names and paths a bit:
- Let's rename the script to `gluon-connection-check` (if we're going with my suggestion to rename the package)
- A better place for the flag file would be `/var/gluon/connection-check/offline`
- The cron job and micrond dependency can be moved from the domain switch package as well

CodeFetch:
@NeoRaider gluon-check-connection would be better if we add other check-packages later.

docs/package/gluon-scheduled-domain-switch.rst
rubo77: the idea was to add more of those packages later: check_gw_connection, check_xyz,...

NeoRaider:
@CodeFetch Hmm, good point. Let's keep the current name.

docs/package/gluon-scheduled-domain-switch.rst
rubo77: ok, so we keep the current name check_connection

package/gluon-check-connection/luasrc/usr/bin/gluon-check-connection
rubo77: since we keep the name check-connectin now, I will move the flag to `/var/gluon/check-connection/offline`

I hope, that there will be no write-action done to the flash-rom there either. (Which was the reason, we chose /tmp to store the flag)

rubo77: I am not sure, how to move the cron from the scheduled domain switch to the new package, as the cron job should only run, in case domain switch is scheduled

rubo77:
I see a problem: If we use the package gluon-check-connection for the domain switch and also for the offline-ssid, there could be needed different IP targets to ping: one to check, if time-servers are reacheable for the domain switch and other IPs to ping to check if the internet connection to the outside is working

package/gluon-check-connection/luasrc/usr/bin/gluon-check-connection
NeoRaider: /tmp and /var are the same directory on OpenWrt, just symlinked.

Having two cronjobs, one in domain-switch and one in ssid-changer, isn't nice either. There would need to be some central facility in the check-connection package that other package can interface with to enable the cron job. Possibly a directory under /var/gluon where packages can drop a file when they need the offline status?

CodeFetch:
@rubo77 Is this really a problem?

rubo77:
We would need to change the sections in the site.conf again, then it will work.

The ips must be declared in the config block of each calling package

CodeFetch:
@rubo77 Yes, you're right... Maybe create a UCI section for the check-connection package and the upgrade scripts of the packages that use the connection check can add entries for it.

CodeFetch:
@rubo77 I've thought about this one... 

A uci-configurable micrond is something that would be a nice feature. I'm not sure whether the OpenWrt devs would like or use it. Cronjobs are one of [the few things](https://openwrt.org/docs/guide-user/base-system/notuci.config) that can't be configured with UCI. There is at least [one script](https://github.com/openwrt/packages/tree/master/net/wifischedule) in OpenWrt that writes UCI-crontab-configs to the crontab file. But I think it is out of scope for this PR. Thus I don't really have a clue how to do this properly in Gluon... The simplest solution would be to just add shell one-liners doing the ping commands and executing the trigger scripts depending on whether they were successful or not in the crontab file using the upgrade-scripts of the packages that need to do the connection checks. To me this is a messy approach, but everything else comes down to introducing something like a UCI-configurable micrond fork or adding support for additional UCI-format crontabs in micrond.

@NeoRaider Did you only choose to not use the UCI format for micrond to make it compatible with cron? Would you consider adding the ability to additionally read a UCI-formatted config file and switch Gluon scripts to use the UCI format instead? Do you know a better approach? I really don't have a clue what could be the right approach.

rotanid:
@rubo77 this needs a rebase.
how is testing of this coming along?

rubo77:
rebased.


You have to make a decision about how those cron jobs should be implemented. 

please add "help wanted"

rotanid:
a i see, this comment by @NeoRaider probably? https://github.com/freifunk-gluon/gluon/pull/1684#discussion_r278070768
maybe he can comment a bit more on a possible solution?

docs/package/gluon-scheduled-domain-switch.rst
```diff
@@ -21,18 +21,26 @@ domain_switch : optional (needed for domains to switch)
...
     switch_time :
         - UNIX epoch after which domain will be switched
-    connection_check_targets :
-        - array of IPv6 addresses which are probed to determine if the node is
-	  connected to the mesh
+
+check_connection :
+    targets :
+        - array of IPv6 addresses which are probed to determine if the node can
+        reach any time-servers to acquire the correct time
```
christf: please add that the check happens using the ICMP protocol via ping.

christf: please reword such that it gets clearer that a node is considered online whenever at least one of those hosts can be reached.

christf:
the ping check is perfectly fine - it is mesh-protocol agnostic which means we won'T need adjustments if we add other protocols.

Can we have the online status displayed on the statuspage?

rubo77:
Can someone take over this pr please? I won't find the time, because I have a child to feed for some time now ;)

rotanid:
i vote against too much configurability for a first approach.

so we need a file /var/gluon/need_cron_check_connection , which is created by e.g. gluon-scheduled-domain-switch or gluon-offline-ssid during their upgrade script run.

the gluon-check-connection package could either
a) check the existence of this file every minute and only do the ping if it exists or
b) it could check it during upgrade scripts and only install its cronjob if the file exists

i wouldn't want multiple sections for the addresses to check, we could use the new section but default to the NTP servers if the new section does not exist?

opinions on this @christf @rubo77  ? i maybe could implement the changes, but i need feedback on the plan.

rubo77:
Sounds good

rotanid:
> Sounds good

there's an a OR b question @rubo77 ;-)

rubo77:
> the gluon-check-connection package could either
> a) check the existence of this file every minute and only do the ping if it exists or
> b) it could check it during upgrade scripts and only install its cronjob if the file exists

i think the first a) is fine and more solid, in case you activate/deactivate e.g. the offline-ssid via uci, the file should be removed too

CodeFetch:
I think we should keep this as simple as possible. A cronjob is being set by an upgrade script and executes a Lua script once a minute. In site conf add a section check-connection with two arrays of hosts one called "global" (only accessible over the internet) and the other one "local" (only accessible over the local network). This package has a UCI section gluon-check-connection where one can add entries with the fields scope, script, event and interval.
If scope is empty/nil it is considered global+local. "Script" is the path to the executable to start on "event". "Event" can either be "offline", "online" or unset (both). And interval is used to only execute the trigger script once in a time.

The Lua script tries to ping each of the global hosts and if one is accessible writes a 1 into /tmp/check-connection-state-global else a 0 and starts trigger scripts. The same for local.

CodeFetch:
@rubo77 If you don't want to implement this I can take over, but my time is currently also very limited.

rubo77:
Yes, please take it over

mweinelt:
Hi @rubo77 

we decided that this changeset is a good candidate for the community repository that is about to be created. 

Please stay tuned.

rubo77:
How could this be stand-alone if the conflicting Package gluon-scheduled-domain-switch is still in Gluon?

mweinelt:
I'm not sure how they conflict.

rotanid:
> I'm not sure how they conflict.

no conflict, but  gluon-scheduled-domain-switch would depend on this new package and drop its own check, as this package is intended as an advanced version of the check inside the before-mentioned package.
this PR here was created so gluon-scheduled-domain-switch and gluon-offline-ssid don't need to implement the checks twice/themselves.

rotanid:
@rubo77 can you also write a comment with information about your latest force-push?
it's hard to see for others, what has changed...

rubo77:
i just rebased to current master and fixed an indention error, jankns complained about

docs/package/gluon-scheduled-domain-switch.rst
```diff
@@ -21,18 +21,26 @@ domain_switch : optional (needed for domains to switch)
         - amount of time without reachable gateway to switch unconditionally
     switch_time :
         - UNIX epoch after which domain will be switched
-    connection_check_targets :
-        - array of IPv6 addresses which are probed to determine if the node is
-	  connected to the mesh
+
```
mweinelt: The documentation should be part of your own package, not of the scheduled domain switch.

package/gluon-check-connection/luasrc/usr/bin/gluon-check-connection
```diff
@@ -4,17 +4,11 @@ local unistd = require 'posix.unistd'
 local util = require 'gluon.util'
 local site = require 'gluon.site'
 
-local offline_flag_file = "/tmp/gluon_offline"
+local offline_flag_file = "/var/gluon/check-connection/offline"
```
NeoRaider: With the changed path, `/var/gluon/check-connection` must be created before the file can be written.

package/gluon-check-connection/Makefile
```diff
@@ -0,0 +1,13 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-check-connection
```
mweinelt: A better name could be `gluon-online-status`.

docs/package/gluon-scheduled-domain-switch.rst
```diff
@@ -21,18 +21,26 @@ domain_switch : optional (needed for domains to switch)
...
     switch_time = 1546344000, -- 01.01.2019 - 12:00 UTC
-    connection_check_targets = {
+  },
+
+  check_connection = {
+    targets = {
       '2001:4860:4860::8888',
       '2001:4860:4860::8844',
     },
   },
```
NeoRaider: As `check_connection` is not specific for domain-switch, this section should be moved to the generic site conf docs.

package/gluon-check-connection/luasrc/usr/bin/gluon-check-connection
```diff
@@ -4,17 +4,11 @@ local unistd = require 'posix.unistd'
...
--- Check if domain-switch is scheduled
-if site.domain_switch() == nil then
-	-- Switch not applicable for current domain
-	os.exit(0)
-end
-
 -- Check reachability of pre-defined targets
-for _, ip in ipairs(site.domain_switch.connection_check_targets()) do
+for _, ip in ipairs(site.check_connection.targets()) do
 	local exit_code = os.execute("ping -c 1 -w 10 " .. ip)
```
christf: please add to the documentation that a node is considered online if any of the specified ip addresses can be reached via ping.

docs/package/gluon-scheduled-domain-switch.rst
rubo77: I just noticed, that the documentation wasnt clear there, should I make a separate PR for this docs change of the sheduled-domain-switch package?

package/gluon-check-connection/Makefile
rubo77: check-connection is more generic, It could also check other connections thatn just online later

package/gluon-check-connection/luasrc/usr/bin/gluon-check-connection
rubo77: I check for the path now and create it now before

docs/package/gluon-scheduled-domain-switch.rst
rubo77: Ah OK, so I'll add an extra docs page for just the check-connection package then

docs/package/gluon-scheduled-domain-switch.rst
rubo77: moved to  generic site conf docs

docs/package/gluon-scheduled-domain-switch.rst
rubo77: not an extra page, but moved to generic site conf docs

package/gluon-check-connection/Makefile
mweinelt: I'd rather see it as a service that provides the online status to other packages. The scope should probably end there.

docs/user/site.rst
```diff
@@ -419,6 +419,22 @@ autoupdater \: package
...
+
+check_connection \: optional
+
+    ``targets`` defines an array of IPv6 addresses which are probed for
+    connectivity, for example determine if the node can reach any time-servers
+    to acquire the correct time. This is e.g. used for the
+    *gluon-scheduled-domain-switch* package ::
+
+    check_connection = {
+      targets = {
```
mweinelt: Maybe call this `icmp_echo_targets` to clarify their usage.

package/gluon-check-connection/luasrc/usr/bin/gluon-check-connection
rubo77: done

rotanid:
the recent mini-reviews are not that useful, or am i mistaken that some bigger changes that were discussed have not yet been implemented?

package/gluon-check-connection/Makefile
```diff
@@ -0,0 +1,13 @@
...
+PKG_VERSION:=1
+
+include ../gluon.mk
+
+define Package/gluon-check-connection
+  TITLE:=Checks if a node can ping definable targets
```
rotanid: like in other contributions, i don't think the "Title" should be a sentence, a sentence is more fitting into the description - maybe i'm wrong?

NeoRaider: A sentence is fine for TITLE, I think about half of our packages do this. The naming is a bit off, it would be better if the field TITLE was called SUMMARY...

I generally avoid to set the description field, nobody reads that anyways.

CodeFetch:
Indeed some ideas are missing here.

> My idea was not to have additional cronjobs for the packages that rely on check-connections:
> check-connection should have a UCI section "gluon-check-connection" where one can add entries with the fields "scope", "script", "event", "onchange" and "interval".
> If "scope" is empty/nil it is considered "global&local". "script" is the path to the executable to start on "event". "event" can either be "offline", "online" or unset ("global&local"). And interval is used to only execute the trigger script once in a time (e.g. every 5 minutes).
> If "onchange" is set true the script is only being executed on a state change.

The current state of the PR is completely different and would introduce superfluous cronjobs.
And it does not distinguish between "local" and "remote" servers, but the scheduled-domain-switch only needs to check the "local" connection while the ssid-changer needs to check the "global" connection.

rubo77:
Sounds good, can you implement this?

CodeFetch:
@rubo77 I wish I could, but I've got too many projects to do at the moment. My schedule for OpenWrt/Gluon-things is:
- Making bootloaders for routers better and integrating an update mechanism in OpenWrt for bootloaders
- Reworking the OpenWrt documentation on "adding new device support" so that it gets more developers
- Integrating second-stage bootloader compilation in OpenWrt for supporting new TP-Link devices without hassle
- Giving the OpenWrt community a possibility to build community-inspired routers
- Finalizing NAT426 so that Babel is more useful
- Creating a layer 2 tunnel protocol based on WireGuard and getting it merged in the linux kernel
- Reworking Gluon's web interface based on the new ubus-API (some community already started working on it - anyone knows who?)
- Getting funding and looking for specialists for development of a "hidden node collision avoidance" algorithm and implementing it for ath10k
- Working on the "decentralized" Freifunk concept
- Finally world domination ;-)

Most of the things needed are already part of the ssid-changer or ffho's wifi fallback updater. You just need to put it in an own package.

package/gluon-check-connection/Makefile
rotanid: thanks for the explanation!

rotanid:
i think this is over-engineering and too much configuration.
i'd merge it with the smaller changes i suggested ( or maybe i find time for this some day myself )

CodeFetch:
@rotanid Okay, I'll look into it this weekend, too. When I look at the offline-ssid I see a lot of redundancy otherwise and I don't think that two, three or four (with fallback-autoupdater) cronjobs executed every minute will make the device more stable in any way. Thus I think merging it in the current state should not be considered.

CodeFetch:
@rubo77 Please have a look at #1930

rubo77:
@CodeFetch It seems like you took some of the changes in my branch and created a new history, is is difficult to merge... how do we proceed now?

should we close this PR and continue development in yours?

CodeFetch:
@rubo77 I wouldn't close this, yet as I don't know if the "complexity" of my proposal will be accepted. I didn't wrote it on top of your branch, but I think I've copied some text at first... What do you mean specifically?

AiyionPrime:
Hey everybody. We might need an interface like this for an ongoing corporation in hanover soon, so I'd like to put some effort in this. I hope "help appreciated" is still valid on this.

I'd like to work on this PR as a stepping stone.
Get it merged into master, work with it a little while and come back later with a second PR that adds, what we're missing most. (Likely some sort of hysteresis, or named sets of target hosts).
What I do not intend ist to get these features into this PR.

I contacted @rubo77 earlier this day and asked about a rebase on master.
There was one commit, that was not proper lua, I fixed it.
Technically there's no problem; I'm just afraid, things around this PR might have changed a lot in the last two years.

However, I compiled the [rebase](https://github.com/rubo77/gluon/pull/7) with a [merge of our nightly siteconf and rubo77s documented site-parameters](https://github.com/freifunkh/site/commit/8ba781a0a2c419d73f09c42be7955c06164c34c7).
The result is available on http://testbench-02.n.ffh.zone/ .
I put the node in a domain, that's not going to find mesh-neighbours.

My speedport promises to block it's internet connection from time to time;
while that's not the case all past reviewers of this PR have ssh-access to the node.

Other than that, calling `gluon-check-connection` does now create the directory `/var/gluon/check-connection` and puts an offline flag within, if there's no route to the specified set of ips.

Imo; there should be some sort of crontab, that exectues the test on a regular basis, as testing the remotes costs a few seconds.

Feedback is appreciated.

mweinelt:
We feel strongly that a package that pings out into the network to determine some state it can derive from local information has no place in the gluon core repository.

We are discussing a proposal on IRC right now with @blocktrron @AiyionPrime and @NeoRaider and will see what comes off it.

CodeFetch:
> We are discussing a proposal on IRC right now

That's the toxic nature of Gluon - Discussing things on IRC... This discussion was done already and the result was that we would like to add a possibility to perform global ping checks for the SSID changer...

CodeFetch:
The ax is already at the root of the trees, and every tree that does not produce good fruit will be cut down and thrown into the fire.

rotanid:
@mweinelt  what kind of behaviour is this? i have to agree somehow with @CodeFetch over at https://github.com/freifunk-gluon/gluon/pull/1930#issuecomment-856151779
some contribution open for a long while, someone tries to push it forward and the immediate reaction is to close it? in my eyes, the right way would have been to post the result of your  IRC discussion here so others can join the discussion. As i don't want to start some kind of "edit-war" i'll not reopen it for the moment...
Also, your argument may be wrong as we already have something that pings out in gluons "core" repo and this interface is simply meant to make this functionality better and modular and therefore usable for other packages, too. I have to admit though, i don't remember at the moment where we already use something like this - and my point may be invalid if it was only a PR that wasn't merged.
CC @blocktrron @AiyionPrime @NeoRaider

mweinelt:
> > We are discussing a proposal on IRC right now
> 
> That's the toxic nature of Gluon - Discussing things on IRC... This discussion was done already and the result was that we would like to add a possibility to perform global ping checks for the SSID changer...

It was clear from the long time this had been lying around, that none of the commiters felt this was happening. In fact while we added some review comments, some of us felt that it wasn't the right move going forward. Ultimately as maintainers of Gluon it is our perogative to decide what goes in, and what does not. It's why we usually ask to discuss approaches with us, before taking the chance to waste your time on something that we are not going to merge. I think it is fair to close pull requests, when that happens.

Our proposal stands. I tried to document it best I could in #2228. Please go read it, if you're interested.

CodeFetch:
@mweinelt I've tested it yesterday. It seems to work fine. There was only a tonumber missing. I'm going to make a PR for it in the community repository after it has been fully tested. I see your point of not doing local tests using ping, but for the SSID changer which I also want to take care of I see no better possibility than doing global pings.

Btw... I didn't feel motivated to do this as I haven't received any feedback whether it was something you would like that way. Furthermore I thought the code is so simple that it might be interesting for a newbie to take over.

rotanid:
i'm fine with #2228 - if it would have been created before closing the PRs, it would have been ideal ;-)

mweinelt:
> We are discussing a proposal on IRC right now with @blocktrron @AiyionPrime and @NeoRaider and will see what comes off it.

I tried to communicate that we were working on something and will come forward with it as it was ready. The order might not have been ideal, but I'd kindly ask for the benefit of the doubt that what I'm doing is not malicious.

PR #1686, opened 2019-03-24
rename DEVICES to GLUON_DEVICES

bobcanthelpyou:
The variable `DEVICES` is useful to build a firmware for a single device or a subset of a target.

Not sure if non-`GLUON_` vars could have a place in the docs, too.


bobcanthelpyou:
Should the variable `GLUON_DEBUG` be added, too?

mweinelt:
It's not clear from the documentation string that the DEVICES variable asks for Gluon profile names. I've verified this against my shell history, but this seems to be an OpenWrt variable. Useful either way.

NeoRaider:
In hindsight, the variable should have been called GLUON_DEVICES, but I don't think it's important enough to change the name now.

The docs should mention that the variable refers to Gluon image names, and how to find these names.

bobcanthelpyou:
Thanks for the suggestions. I will add some more information and rebase the commit.

bobcanthelpyou:
Added the `GLUON_` prefix

docs/user/getting_started.rst
```diff
@@ -186,6 +186,12 @@ GLUON_TARGET
 Special variables
 .................
 
+GLUON_DEVICES
+  List of devices to build. The list contain the Gluon profile name of a device, the profile
```
rotanid: *contains

mweinelt:
The commit message should begin with `build:`.

rotanid:
i adjusted the commit message and merged.

@bobcanthelpyou please leave a comment if you do changes, otherwise no one gets notified about your changes to the PR

bobcanthelpyou:
Sorry, thought that issue required a responds from NeoRaider.

PR #1693, opened 2019-04-04
docs: add notice for opkg repo change

bobcanthelpyou:
A diff might describe it better
https://github.com/freifunk-gluon/gluon/commit/210d97c53eb147ad9519118634c71e28eace4694#diff-0d44e6a3fa2aba2dc4a79afbfcc4ab1fR78

A cherrypick to v2018.2.x might be useful.

### DIFF
--- a/docs/releases/v2018.2.rst
+++ b/docs/releases/v2018.2.rst
@@ -127,8 +127,12 @@ adding ``-gluon-ebtables-limit-arp`` to *GLUON_SITE_PACKAGES*.
 Site changes
 ************
 
-No changes need to be made to *site.conf* or *site.mk* when upgrading from
-Gluon v2018.1.x.
+If an opkg repository for ``lede`` was configured the key needs to be migrated
+to ``openwrt``. ``lede`` is ignored and without an ``openwrt`` key the default
+OpenWrt repository is used.
+
+No other changes need to be made to *site.conf* or *site.mk* when upgrading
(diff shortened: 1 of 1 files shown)


docs/releases/v2018.2.rst
```diff
@@ -130,6 +130,10 @@ Site changes
 No changes need to be made to *site.conf* or *site.mk* when upgrading from
 Gluon v2018.1.x.
 
+Only the optional *opkg* configuration might need an update. The LEDE
+repository setting is ignored and a different OpenWrt repository need to set
```
rotanid: "need to set" sounds wrong, maybe "needs to be set" ?

bobcanthelpyou: Not sure, maybe something like:
> If an opkg repository for ``lede`` was configured the key needs to be migrated
> to ``openwrt``. ``lede`` is ignored and without an ``openwrt`` key the default
> OpenWrt repository is used. No other changes need to be made to *site.conf*
> or *site.mk* when upgrading from Gluon v2018.1.x.

mweinelt: Put 

> No other changes need to be made ...

on a new line, so it can be spotted without reading the paragraph.

rotanid:
merged and cherry-picked

bobcanthelpyou:
Thank you :)

PR #1697, opened 2019-04-09
ar71xx: add support for TP-Link TL-WR840N v2

bobcanthelpyou:
TP-Link TL-WR840N v2: https://openwrt.org/toh/hwdata/tp-link/tp-link_tl-wr840n_v2

  - [x] webinterface
  - [ ] tftp
  - [ ] other: <specify>
- wired network
- wifi (if applicable)
- led mapping
  - power/sys led
  - radio leds
    - [x] should show activity
  - switchport leds

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -290,6 +290,7 @@ ar71xx-tiny
   - TL-WR740N (v1, v3, v4, v5)
   - TL-WR741N/ND (v1, v2, v4, v5)
   - TL-WR743N/ND (v1, v2)
+  - TL-WR840N (v2)
   - TL-WR841N/ND (v3, v5, v7, v8, v9, v10, v11, v12)
   - TL-WR843N/ND (v1)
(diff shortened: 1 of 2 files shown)


belzebub40k:
Isn't it counterproductive to add new devices with 4MB flash and 32MB RAM if we want to get rid of such devices in the future?

rotanid:
@bobcanthelpyou so there's no problem with testing so far and only 2 things to test left, right?

@belzebub40k the current device integration policy allows 4/32 if no backport patch is necessary.
apart from that, the policy can be revisited at the next meeting or so.

belzebub40k:
@rotanid ok, sorry I thought this decision was already made.

bobcanthelpyou:
@rotanid I need to test `sysupgrade`, but already tested `autoupdate`.

bobcanthelpyou:
`sysupgrade` and `firstboot` seems to be supported, too.

PR #1700, opened 2019-04-11
Bugfix: gluon-config-mode-mesh-vpn: 0300-mesh-vpn.lua limit_ingress backport to v2018.2.x

2tata:
backport from #1682 to v2018.2.x

fix limit_ingress default value in case of default disabled simple-tc for tunneldigger.

In case if 'limit_bw_down' set for tunneldigger should simple-tc still be enabled?

If simple-tc enabled by default via site.conf the script '500-mesh-vpn' from 'gluon-mesh-vpn-core' will set 'limit_ingress' to 0 and also enabled simple-tc. Does 0 have a special meaning in this situation like "ignore it"?

Bug appears here: #1460
Related to #1447

Thanks to @lrnzo


rotanid:
i've cherry-picked the other commit from master.

PR #1703, opened 2019-04-13
gluon.mk: fix reference to LuaSrcDiet binary

bobcanthelpyou:
The luci modules where updated (c919f5fd29aeb37e6a7a50d6834fa338e7328dfd)
And there is a luasrcdiet update included `b5d5e5bf1 luci-base: update luasrcdiet` (https://github.com/openwrt/luci/commit/b5d5e5bf13e572af7f1589a01bfd338691d5f063)

I noticed some `bash: LuaSrcDiet: command not found` in the `make V=s` output and updated the luasrcdiet call

Could someone check a Gluon build if the result of the luasrcdiet output is as expected?

### DIFF
--- a/package/gluon.mk
+++ b/package/gluon.mk
@@ -58,7 +58,7 @@ define GluonSrcDiet
 	rm -rf $(2)
 	$(CP) $(1) $(2)
 	$(FIND) $(2) -type f | while read src; do \
-		if LuaSrcDiet --noopt-binequiv -o "$$$$src.o" "$$$$src"; then \
+		if luasrcdiet --noopt-binequiv -o "$$$$src.o" "$$$$src"; then \
 			chmod $$$$(stat -c%a "$$$$src") "$$$$src.o"; \
 			mv "$$$$src.o" "$$$$src"; \
 		fi; \


bobcanthelpyou:
Checked the file `status-page.lua` on a device without and with a source code diet.

Without the diet the lua file seem to be the same as in the Git repo.
With the diet the file is without unneeded white spaces and functions and var names are replaced with short names

My guess is works as expected, again.

mweinelt:
Good find! Minor nit below:

>  gluon.mk: update luasrcdiet binary 

No, you are likely not updating the binary, only the reference to it.

Also it's a fix, not just an update.

bobcanthelpyou:
@mweinelt sorry

Maybe `fix luasrcdiet reference` or `fix luasrcdiet command` would fit better?

mweinelt:
How about:

> `gluon.mk: fix reference to LuaSrcDiet binary`

bobcanthelpyou:
Updated the commit message as suggested

rubo77:
Are external packages afected? Do we have to change the Makefiles there too?

NeoRaider:
- In luci.mk, luasrcdiet is called like this now: `LUA_PATH="$(STAGING_DIR_HOSTPKG)/lib/lua/5.1/?.lua" luasrcdiet`. We should do the same.
- Or, as an alternative, revert the LuCI feed update (the LuCI feed will be removed from Gluon anyways eventually - we might be able to get away with just not updating it until that happens)

PR #1708, opened 2019-04-19
package/libgluonmath: add math lib for gauss jordan Ponit in Polygon algorithm

2tata:
As mentioned from @NeoRaider https://github.com/freifunk-gluon/gluon/pull/1393#discussion_r221433021

This MR split off the polygon math functions from the hoodselector MR because this math functions are not directly related to the hoodselector. So it can also use for other programs.

### DIFF
--- a/package/libgluonmath/Makefile
+++ b/package/libgluonmath/Makefile
@@ -1,0 +1,15 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=libgluonmath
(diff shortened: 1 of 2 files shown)


NeoRaider:
There is nothing Gluon-specific in this library anymore, so I think it should be named something like `lua-math-polygon`, and live in the packages repo.

2tata:
moved to https://github.com/freifunk-gluon/packages/pull/219

PR #1710, opened 2019-04-20
Add mechanism to reload domain related services

2tata:
This MR adds a mechanism to restart domain related services via `gluon-reload-daemons` services that want to be reloaded needs to be named in `/lib/gluon/reload.d`. `gluon-reload-daemons` will call the init.d scrips.

Recomended by @NeoRaider : https://github.com/freifunk-gluon/gluon/pull/1393#discussion_r221433257


package/gluon-core/files/usr/bin/gluon-reload-daemons
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+cd "/lib/gluon/reload.d" || exit 1
+
+err=0
+
+for service in *; do
+	echo "restarting: ${service} ..."
+	exec "/etc/init.d/$service restart" || err=1
```
NeoRaider: Uhm... did you test this?

"exec" will replace the calling shell, so if this command could succeed, only the first restart would happen. It cannot succeed though, as the shell will consider `/etc/init.d/$service restart` the name of the binary to run instead of a shell command with an argument due to the quotation marks.

Generally, I find this approach questionable. Restarting daemons is already an "expensive" operation, so it wouldn't hurt to make this more flexible by making `/lib/gluon/reload.d/*` shell scripts and just run the scripts (similar to our upgrade scripts).

Also, we don't always want to `restart` a daemon, a `reload` will sometimes suffice.

2tata:
On 2019-04-20 14:35, Matthias Schiffer wrote:
> NeoRaider requested changes on this pull request.
> 
> 
> 
>> @@ -0,0 +1,17 @@
> +#!/bin/sh
> +
> +cd "/lib/gluon/reload.d" || exit 1
> +
> +err=0
> +
> +for service in *; do
> +	echo "restarting: ${service} ..."
> +	exec "/etc/init.d/$service restart" || err=1
> 
> Uhm... did you test this?

ohh .. my mistake this obviously wont work..
I forgot this change on an another dir.

> Generally, I find this approach questionable. Restarting daemons is already an "expensive" operation, so it wouldn't hurt to make this more flexible by making `/lib/gluon/reload.d/*` shell scripts and just run the scripts (similar to our upgrade scripts).
> 
> Also, we don't always want to `restart` a daemon, a `reload` will sometimes suffice.
> 
Si, I will change it.

2tata:
@NeoRaider I have updated this MR

NeoRaider:
- Let's call the main script just `gluon-reload`
- Files in `/lib/gluon/reload.d` should be named after the Gluon packages that install them (so `core`, `mesh-vpn-fastd`, ...)
- A 3-digit prefix for ordering seems like a good idea (`500-core`, ...)
- To prevent accidental bridging of different domains or incorrect respondd announcements, the following ordering must be adhered to:
  - Stop all VPN daemons and respondd
  - Restart network
  - Start VPN daemons and respondd

2tata:
@NeoRaider done with my changes

package/gluon-respondd/files/lib/gluon/reload.d/009-respondd-stop
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+/etc/init.d/gluon-respondd stop
```
NeoRaider: Maybe have the respondd scripts at 100/900, and the VPN scripts at 200/800? No need to use the second and third digit with the first scripts we add.

package/gluon-core/files/usr/bin/gluon-reload
```diff
@@ -0,0 +1,17 @@
...
+
+err=0
+
+for script in *; do
+	echo "Reloading: ${script}"
+	./"$script" || err=1
+done
+
+if [ $err -eq 1 ]; then
+	echo 'One or more daemons failed to reload.'
```
NeoRaider: Add `>&2` to print to stderr

package/gluon-core/files/lib/gluon/reload.d/500-core
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+/etc/init.d/network reload
```
NeoRaider: Actually, a full network restart instead of a reload might be a good idea.

2tata:
@NeoRaider updated

PR #1714, opened 2019-04-22
gluon-mesh-batman-adv: drop packets from mesh with next-node source or destination address

### LINKED ISSUE #1659: Node-local DNS resolver often responds extremely slowly
### Bug report

**What is the problem?**

DNS queries to the DNS resolver on the node are often taking a long while to be answered.

I am repeatedly executing the following on a client connected to the node:
```
dig @fd4e:f2d7:88d2:ffff::1 ralfj.de
```
where `fd4e:f2d7:88d2:ffff::1 ` is the "next node" anycast IP. This command sometimes comes back immediately, sometimes it comes back after several seconds, and sometimes it times out after several seconds (all of these cases seem very roughly equally likely).
After the initial query, even when the answer takes several seconds, there is no DNS traffic on the tunnel between the node and the gateway. The link between the client and the node is a Linux bridge; both of these are just VMs on the same host and hence there should be 0 packet loss. I see the same behavior when talking to a real node over a WiFi link.

When I do `nslookup ralfj.de` on the node locally, the answer always comes back immediately. Also when I do `dig @fd4e:f2d7:88d2:ffff::101 ralfj.de`, which sends the query to one of our gateways, the answer comes back immediately -- so the network connection(s) all seem fine.

The DNS reply should come back with decent latency (<50ms, say).

**Gluon Version:**

v2018.2

**Site Configuration:**

https://git.hacksaar.de/FreifunkSaar/gluon-site/blob/e2a66b1a461f95a80ee9483c9dfca96fe6fb2b36/site.conf

**Custom patches:**

None.

NeoRaider:
Fixes #1659

@RalfJung Could you check if this is working as expected?

### DIFF
--- a/package/gluon-mesh-batman-adv/files/lib/gluon/ebtables/250-next-node
+++ b/package/gluon-mesh-batman-adv/files/lib/gluon/ebtables/250-next-node
@@ -7,6 +7,9 @@ local macaddr = client_bridge.next_node_macaddr()
 rule('FORWARD --logical-out br-client -i bat0 -o local-port -j DROP')
 rule('FORWARD --logical-out br-client -i local-port -o bat0 -j DROP')
 
+rule('PREROUTING --logical-in br-client -i bat0 -s ' .. macaddr .. ' -j DROP', 'nat')
+rule('PREROUTING --logical-in br-client -i bat0 -d ' .. macaddr .. ' -j DROP', 'nat')
+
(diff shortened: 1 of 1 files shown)


RalfJung:
"Unfortunately" we fixed the misbehaving node in our network to no longer send these "bad" packages... so I'm afraid I have no direct way to test this.

PR #1702, opened 2019-04-13
[WIP] add checks for obsolete settings

bobcanthelpyou:
This might fix #1698 

When running `make V=s GLUON_RELEASE="master" GLUON_TARGET="ar71xx-tiny" GLUON_DEVICES="tp-link-tl-wr840n-v2"` (site conf of Freifunk Hamburg and Gluon master) the following output is printed eight times:
```
site.conf obsolete: opkg.lede is obsolete and could be removed, check the release notes and documentation please
```

I am not sure if `print()` is a good choice or if something like `io.stderr:write()` should be used
e.g. `io.stderr:write(src .. ' obsolete: ' .. string.format(...) .. '\n')` (but it is visible with `V=s`, only)

Also not sure if a generic text is fine or if it should be possible to describe what is wrong, add a link to the documentation/release notes, ... what do you think?

PR #1716, opened 2019-04-28
gluon-core: disable legacy_rates by default, drop support for support…

mweinelt:
…ed and basic rates

All we ever wanted to achieve with supported and basic rates was disabling 802.11b. In the meantime OpenWrt provides¹ the `legacy_rates` option to accomplish exactly that. This obsoletes the need for configurable basic and supported rates, as neither 802.11a nor 802.11g are considered as harmful as 802.11b. Their "slowness" will instead be mitigated through airtime fairness in a future release.

This patchset makes use of the obsoletion check feature in #1702.

[1] https://github.com/openwrt/openwrt/commit/ed62d91f4b5296a4aa883ce975d76f590ef4e910

### DIFF
--- a/package/gluon-core/check_site.lua
+++ b/package/gluon-core/check_site.lua
@@ -27,28 +27,27 @@ need_string_array({'ntp_servers'}, false)
 
 need_string_match(in_domain({'prefix6'}), '^[%x:]+/64$')
 
-
+local supported_rates = {6000, 9000, 12000, 18000, 24000, 36000, 48000, 54000}
 for _, config in ipairs({'wifi24', 'wifi5'}) do
 	if need_table({config}, nil, false) then
 		need_string(in_site({'regdom'})) -- regdom is only required when wifi24 or wifi5 is configured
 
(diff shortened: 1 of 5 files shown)


package/gluon-core/check_site.lua
```diff
@@ -27,28 +27,31 @@ need_string_array({'ntp_servers'}, false)
 
 need_string_match(in_domain({'prefix6'}), '^[%x:]+/64$')
 
-
+local supported_rates = {6000, 9000, 12000, 18000, 24000, 36000, 48000, 54000}
```
mweinelt: This is used to validate mcast_rate entries, I've removed all 802.11b rates here, even when legacy_rates is disabled. Feels a bit weird in that case, but enabling legacy_rates would be equally as weird I suppose.

NeoRaider:
Do we even need a site.conf option for this? I'd prefer to just hardcode disabling legacy rates.

mweinelt:
Agreed, dropped.

Issue #1569, opened 2018-11-15
Support for "make dist"


yanosz:
Hello,

is:

* Gluon Builds are done using `make world` with some parameter are specified.
* Doing a gluon build is hardly done calling make world only one time
  * Different architectures require sucessive calls
  * Different side-configuration require sucessive calls (esp. for minor-changes - i.e. batman-adv-Version), that cannot be implemented in a multidomain setup

As an effect, different communities are creating different, elaborate build configurations.

Should:

* Gluon should support an additional target "make dist"
  * This target should build all available platforms (selecting explicit ones should be possible though)
  * This target should accept multiple site-configurations. **Validation** should be done **before** anything else (esp. sth. with  effort) is done
  * The target should create a folder called dist include the full tree of all images and packages # @

christf:
I think it would be helpful to standardize the builds on a gluon-project level instead of community level. Another idea was to specify a Jenkinsfile and have the parameters in place to customize.

For the make dist approach - how would parameters be stored?

yanosz:
Ok, thinking about this for a while, I think, that theses features are missing in Gluons Makefile
- [ ] Build more than one platform
- [ ] Use Envirnomental variables (i.e. yielding envsubst) for things, that cannot be hardcoded in the config.
- [ ] Validations before build

If `make dist` boils down to that, a jenkins job should be easy.

neocturne:
Hmm, I've been thinking about this for a bit, and while it would of course be possible to add an "all-targets" Make target, I don't think there are any usecases for it. Locally, I would never want to build all targets; in CI, you want a separate job for each target.

We are currently using Github's CI because it provides us with free runners. I can see a few things that could be improved here:
- Modularize Github Actions to make it easier for communities to reuse parts
- Implement templates for alternative CI that can be set up on owned infrastructure (Jenkins, Gitlab, ...)

Issue #1660, opened 2019-02-28
802.1x authentication for open infrastructure 802.11s networks


CodeFetch:
Using SAE with a pre-shared key does allow encryption, but not authentication of nodes if the PSK is publicly known. 
For open infrastructure networks a possibility to ensure at least trust on first use would be to initiate a connection between nodes using a publicly known PSK the first time, then generate a random PSK using that secure session and save it to flash on the peering nodes. A custom RADIUS server would be needed for that.

mweinelt:
What would happen if either side lost its state due to a reset?

CodeFetch:
They need to handle out new credentials. Actually I don't even know whether it is implemented open-source, yet. The goal of having something like that is to implement a trust model. If they lose their credentials, they lose their accumulated trust.

mweinelt:
How would you distinguish lost state versus an attack by someone malicious?

CodeFetch:
Distinguishing between a lost state and an attack is easy as the ID of the node should not rely on a MAC address, but a randomly initiated value. You could also use public key crypto for that. Trust models are the hard problem. They are ongoing research e.g. by the military. There are many papers about that. Some propose using neural networks other simple evidence based rating algorithms. For ordinary commercial networks you either use a shared PSK (which is prone to physical hacking) or a centralized trust agent which is a single point of failure and needs to be reachable all the time. Without having authenticated mesh connections, we have no possibility to implement trust models at all. Trust on first use would be a first step.

mweinelt:
Closing as stale and unlikely to be implemented.

Issue #1583, opened 2018-11-25
LXC virtualiziation for embedded devices


renne:
Hi,

more powerful embedded devices like the [Turris Omnia](https://omnia.turris.cz/en/) have enough resources to run Gluon in a [LXC-container](https://linuxcontainers.org/) but not enough resources and hardware to run VMWare or VirtualBox.

I suggest to implement support to run Gluon in [LXC-containers](https://linuxcontainers.org/) and offer target-specific containers via the [Linux Containers - Image server](https://uk.images.linuxcontainers.org/) (e.g. ARMv7L on [Turris Omnia](https://omnia.turris.cz/en/)).

mweinelt:
We are only a downstream project of OpenWrt, imo container support would need to happen upstream.

NeoRaider:
The host kernel would need to provide batman-adv support (and I'm not sure if all relevant debugfs interfaces have been replaced by Netlink yet - this is a prerequisite for using batman-adv with containers). Generally, I'm not convinced creating such dependencies between a Gluon container and its host is a good idea - at the very least, LXC would never be an officially supported target of Gluon with batman-adv.

We might look into containers when our support for alternative routing protocols like Babel has matured a bit, but for now, this is still considered experimental.

T-X:
Actually, since batman-adv v2018.2 the debugfs part in batman-adv is even disabled by default:

<pre><code>
    batman-adv: Disable CONFIG_BATMAN_ADV_DEBUGFS by default
    
    All tools which were known to the batman-adv development team are
    supporting the batman-adv netlink interface since a while. Also debugfs is
    not supported for batman-adv interfaces in any non-default netns. Thus
    disabling CONFIG_BATMAN_ADV_DEBUGFS by default should not cause problems on
    most systems. It is still possible to enable it in case it is still
    required in a specific setup.
    
    Signed-off-by: Sven Eckelmann
</code></pre>

https://git.open-mesh.org/batman-adv.git/commit/6f542489678248cc99131f2856158bd2c3989368

So you might actually be quite good to go now for a try to run Gluon in a container.

@renne: If you could build the Gluon x86 target, copy its rootfs, craft a matching LXC config yourself and if you could let us know whether that works for you already or what does not work yet, that would be very interesting.

I don't know what requirements OpenWrt would have to be able to run within an LX container. You might need to try with a very recent Linux kernel or with an OpenWrt (Linux kernel) on the host itself.

[...]

PR #1719, opened 2019-05-02
Added encryption support for mesh + managed networks

yanosz:
* This package arose from fieldtracks.org and addresses non-Freifunk setups, in which all routers are under a single administrative control.
* Unlike https://github.com/freifunk-gluon/gluon/pull/1658 the code is implemented outside of gluon's core scripts, reducing complexity and dependencies. IMHO it doesn't feel right to make it a core functionality.
* The code requires passwords to be present in the site.conf.  This is generally a bad practicie and replacing the password during build is basically a one-liner (`envsubst < site.conf.template > site.conf)`, that is not supported by gluon's build system #1569.

### DIFF
--- a/package/gluon-wlan-encryption-psk/Makefile
+++ b/package/gluon-wlan-encryption-psk/Makefile
@@ -1,0 +1,13 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-wlan-encryption-psk
+PKG_VERSION:=1
+
+include ../gluon.mk
+
+define Package/gluon-wlan-encryption-psk
+  TITLE:=Encrypt wlan using a pre-shared key (psk)
+  DEPENDS:=+gluon-core +wpad-mesh-wolfssl
+endef
+
+$(eval $(call BuildPackageGluon,gluon-wlan-encryption-psk))
--- a/package/gluon-wlan-encryption-psk/Readme.md
+++ b/package/gluon-wlan-encryption-psk/Readme.md
@@ -1,0 +1,15 @@
+# Gluon wlan encrypt
+Encrypt wireless networks in a non-Freifunk setup.
+
+This package allows to encrypt 802.11s and infrastructure networks using a pre shared key (psk).
+
+This setup results in passwords being stored in plain-text. Site-configuration, Firmware-downloads and
+Devices must be protected against unauthorized access
+
+*make sure to use exclude hostapd-mini site.mk*
+
+Example configuration:
+
+```lua
+wlan_encryption_psk = { mesh = '802.11s.passwd', ap = ' infrastructure.passwd', },
+```
--- a/package/gluon-wlan-encryption-psk/check_site.lua
+++ b/package/gluon-wlan-encryption-psk/check_site.lua
@@ -1,0 +1,1 @@
+need_table({'wlan_encryption_psk'})
--- a/package/gluon-wlan-encryption-psk/luasrc/lib/gluon/upgrade/390-gluon-wlan-encryption-psk
+++ b/package/gluon-wlan-encryption-psk/luasrc/lib/gluon/upgrade/390-gluon-wlan-encryption-psk
@@ -1,0 +1,21 @@
+#!/usr/bin/lua
+
+local site = require 'gluon.site'
+local uci = require('simple-uci').cursor()
+
+local ap_password = site.wlan_encryption_psk()["ap"]
+local mesh_password = site.wlan_encryption_psk()["mesh"]
+
+
+uci:foreach('wireless', 'wifi-iface', function(wifi_if)
+    local ifname = wifi_if['.name']
+    if wifi_if.mode == 'ap' and ap_password ~= nil and ap_password ~= '' then
+      uci:set('wireless', ifname, 'encryption', 'psk2')
+      uci:set('wireless', ifname, 'key', ap_password)
+    elseif wifi_if.mode == "mesh" and mesh_password ~= nil and mesh_password ~= '' then
+      uci:set('wireless', ifname, 'encryption', 'psk2/aes')
+      uci:set('wireless', ifname,  'key', mesh_password)
+    end
+end)
+
+uci:save('wireless')


package/gluon-wifi-encrypt/Makefile
```diff
@@ -0,0 +1,13 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-wifi-encrypt
```
CodeFetch: I know we often use the term "WiFi" like in the wifi-config package. The correct technical term is WLAN as WiFi is a brand. Yes, we say "googling" instead of "searching the internet", but that should just be slang.
Furthermore "encrypt" is too general as we can use other means of encryption like SAE.
Thus I'd like to call this package "gluon-wlan-encryption-psk"

package/gluon-wifi-encrypt/Makefile
```diff
@@ -0,0 +1,13 @@
...
+PKG_VERSION:=1
+
+include ../gluon.mk
+
+define Package/gluon-wifi-encrypt
+  TITLE:=Encrypt WiFi using psk
```
CodeFetch: WLAN PSK encryption

package/gluon-wifi-encrypt/Readme.md
```diff
@@ -0,0 +1,15 @@
+# Gluon WiFi encrypt
+Encrypt wireless networks in a non-Freifunk setup.
+
+This package allows to encrypt 802.11s and infrastructure networks using a pre shared key (psk).
+
+This setup results in passwords being stored in plain-text. Site-configuration, Firmware-downloads and
+Devices must be protected against unauthorized access
+
+*make sure to use exclude hostapd-mini site.mk*
```
CodeFetch: #WLAN PSK encryption
Encrypts client and 802.11s WLAN networks using pre-shared keys.

This package allows to encrypt 802.11s and infrastructure networks using pre-shared keys (PSK).

package/gluon-wifi-encrypt/Readme.md
```diff
@@ -0,0 +1,15 @@
...
+
+Example configuration:
+
+```lua
+encrypt_wifi = { mesh = '802.11s.passwd', ap = ' infrastructure.passwd', },
```
CodeFetch: `wlan_encryption = { mesh = { mode = 'psk', key = 'mesh-example-key' }, client = {  mode = 'psk', key = 'client-example-key' }}`

CodeFetch: Maybe better put this in the existing WLAN site.conf sections.

yanosz:
@CodeFetch: Thanks for your feedback. I like your ideas regarding naming. Most of them are rather minor .. most probably I'll fix 'em before building the fieltracks images .. that'll be before 09-2019 .. thus it may take a few weeks. I'd like to skip the effort for building and testing a version with these fixes, only.

* It's 390-gluon-wifi-encrypt and not 200-sth. because it has to be executed after 320-gluon-client-bridge-wireless (and scripts like this), since all infrastructure networks have to be present. A "late" 3x0-Number seems reasonable to me.
* It is outside the existing WLAN site.conf sections, beacuse they allow more flexibility (2.4, vs. 5, multiple networks). Aligning the sections with encryption somewhat likely results in patching a bunch of gluon's core scripts. I'd like to avoid that (complexity, effort), because I don't need that .. maybe later, if need frequency schedules or stuff like that.
* I don't get the point "I think this should keep #1658 in mind" - #1658 has a different treat model (and diffrent approach, thus). Also, the scope differers:  #1658 discusses an architecture including a prototype for privacy presevering community networks - #1719 is a hack for a private mesh network, that can be reused by others.

CodeFetch:
> * It's 390-gluon-wifi-encrypt and not 200-sth. because it has to be executed after 320-gluon-client-bridge-wireless (and scripts like this), since all infrastructure networks have to be present. A "late" 3x0-Number seems reasonable to me.

Sorry, I don't understand why... Don't you only change WLAN options?

> * It is outside the existing WLAN site.conf sections, beacuse they allow more flexibility (2.4, vs. 5, multiple networks). Aligning the sections with encryption somewhat likely results in patching a bunch of gluon's core scripts. I'd like to avoid that (complexity, effort), because I don't need that .. maybe later, if need frequency schedules or stuff like that.

You can add another check-site for that. You don't need to alter the core-modules.

> * I don't get the point "I think this should keep #1658 in mind" - #1658 has a different treat model (and diffrent approach, thus). Also, the scope differers:  #1658 discusses an architecture including a prototype for privacy presevering community networks - #1719 is a hack for a private mesh network, that can be reused by others.


> n = { mesh = { mode = 'psk', key = 'mesh-example-key' }, client = { mode = 'psk', key = 'client-example-key' }}

mode = 'psk' and mode = 'sae'

#1658 Does not really fit for community networks. It's the same use-case like yours. That's why I've opened #1660 (You need a special radius server for that and implement a trust model).

yanosz:
Hei,

> > * It's 390-gluon-wifi-encrypt and not 200-sth. because it has to be executed after 320-gluon-client-bridge-wireless (and scripts like this), since all infrastructure networks have to be present. A "late" 3x0-Number seems reasonable to me.
> 
> Sorry, I don't understand why... Don't you only change WLAN options?

yep - so does https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless, when creating a new AP-interface. I iterate over that.
 
> > * It is outside the existing WLAN site.conf sections, beacuse they allow more flexibility (2.4, vs. 5, multiple networks). Aligning the sections with encryption somewhat likely results in patching a bunch of gluon's core scripts. I'd like to avoid that (complexity, effort), because I don't need that .. maybe later, if need frequency schedules or stuff like that.
> 
> You can add another check-site for that. You don't need to alter the core-modules.

Well, in princible - yes. it's more complex, though.

> > * I don't get the point "I think this should keep #1658 in mind" - #1658 has a different treat model (and diffrent approach, thus). Also, the scope differers:  #1658 discusses an architecture including a prototype for privacy presevering community networks - #1719 is a hack for a private mesh network, that can be reused by others.
> 
> > n = { mesh = { mode = 'psk', key = 'mesh-example-key' }, client = { mode = 'psk', key = 'client-example-key' }}
> 
> mode = 'psk' and mode = 'sae'
> 
> #1658 Does not really fit for community networks. It's the same use-case like yours. That's why I've opened #1660 (You need a special radius server for that and implement a trust model).

Ah - ok, got it. The PR does - the discussion is rather philosophical.

yanosz:
Heiho,

I just changed names. My network has been running stable for a few days now.

package/gluon-wlan-encryption-psk/Readme.md
```diff
@@ -0,0 +1,15 @@
+# Gluon wlan encrypt
```
2tata: I think this should be remove here and add to the docs

2tata: I think this should be remove here and add to the docs

package/gluon-wlan-encryption-psk/Makefile
```diff
@@ -0,0 +1,13 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-wlan-encryption-psk
+PKG_VERSION:=1
+
+include ../gluon.mk
+
+define Package/gluon-wlan-encryption-psk
+  TITLE:=Encrypt wlan using a pre-shared key (psk)
```
CodeFetch: Title vs. description and "pre-shared key (PSK)" and WLAN

package/gluon-wlan-encryption-psk/Readme.md
CodeFetch: Yes, it better fits in the docs and "Gluon WLAN encryption"

package/gluon-wlan-encryption-psk/Readme.md
```diff
@@ -0,0 +1,15 @@
...
+Encrypt wireless networks in a non-Freifunk setup.
```
CodeFetch: Enables encryption for AP and 802.11s mesh STA wireless networks using pre-shared keys defined in the site.conf at firmware build time.
This could also be used as description of the package - Don't mention Freifunk in this context.

package/gluon-wlan-encryption-psk/Readme.md
```diff
@@ -0,0 +1,15 @@
...
+
+This package allows to encrypt 802.11s and infrastructure networks using a pre shared key (psk).
```
CodeFetch: Superfluous with the change above

package/gluon-wlan-encryption-psk/Readme.md
```diff
@@ -0,0 +1,15 @@
...
+
+This setup results in passwords being stored in plain-text. Site-configuration, Firmware-downloads and
```
CodeFetch: The pre-shared keys need to be kept secret. They are saved as plain text in the site.conf, the resulting firmware images and on the router. Thus access to these files and deployed routers needs to be restricted to authorized persons.

package/gluon-wlan-encryption-psk/Readme.md
```diff
@@ -0,0 +1,15 @@
...
+Devices must be protected against unauthorized access
+
+*make sure to use exclude hostapd-mini site.mk*
```
CodeFetch: Can you please explain this better?

CodeFetch: Can you please explain this better?

CodeFetch: Can you please explain this better?

CodeFetch: What I would wish to have here is a simple explanation about secure WPA passwords like:

Please be aware that the shortest allowed PSK's length is 8 characters and it can be up to 63 characters long. It is highly recommended to choose a randomly generated PSK with at least 40 characters consisting of alphanumeric, numeric and special characters.

CodeFetch: What I would wish to have here is a simple explanation about secure WPA passwords like:

Please be aware that the shortest allowed PSK's length is 8 characters and it can be up to 63 characters long. It is highly recommended to choose a randomly generated PSK with at least 40 characters consisting of alphanumeric, numeric and special characters.

CodeFetch:
Sorry... buggy Github?!

package/gluon-wlan-encryption-psk/Makefile
christf: and the title should mention that AP and mesh are going to be crypted using that key.

package/gluon-wlan-encryption-psk/Readme.md
christf: I like yanosz version better and would like to make a minor suggestion:

This setup requires in passwords being stored in the firmware images, site-configuration and on the devices in clear-text. The security of the network depends on the protection of these artifacts against unauthorized access. Do not rely on the devices ssh passwords for security.

package/gluon-wlan-encryption-psk/Readme.md
christf: I do not agree with this. It is clear that a WPA password is to be used. The implications of password security are also clear. I do not see the need to elaborate as part of the gluon project.

christf: > Can you please explain this better?

could the package exclude hostapd-mini?

yanosz:
@CodeFetch - @christf 

Thanks for your comments - I don't like to discuss changes to readmes - I guess, that both of your ideas are fine. I guess, the tricky is the site-effect driven configuration of different uci-default scripts. 

I.e. I don't how to write the needed, automated integration-tests making sure, that wifi is configured as defined.

Please provide a perspective on how to go one with this package - updating the readme is rather minor.

CodeFetch:
> I.e. I don't know how to write the needed, automated integration-tests making sure, that wifi is configured as defined.

Which tests? I think the code is good - I don't see any problems here.

rotanid:
> Please provide a perspective on how to go one with this package - updating the readme is rather minor.

maybe later we need a "conflicts" addition with #1658  ?
and changes to readme/docs should be addressed nevertheless.

as it is a separate package, i think we should/could merge it after the v2019.1 release, if i don't get any Veto from another maintainer until then.

yanosz:
> > I.e. I don't know how to write the needed, automated integration-tests making sure, that wifi is configured as defined.
> 
> Which tests? I think the code is good - I don't see any problems here.

Given a set of representative wifi configurations, I'd like to make sure, that all are resulting in correct configuration.

Afaik you mentioned in #1583 a KVM setup for doing so.

yanosz:
> > Please provide a perspective on how to go one with this package - updating the readme is rather minor.
> 
> maybe later we need a "conflicts" addition with #1658 ?
> and changes to readme/docs should be addressed nevertheless.
> 
> as it is a separate package, i think we should/could merge it after the v2019.1 release, if i don't get any Veto from another maintainer until then.

That's always the case; I guess, I wasn't clear about the context of this perspective.  ;-)
It's more like: What has to bested / changed in addation? 

I'm not interested in certain thime frame: neoraider encouraged me to file a PR a while ago and I guess that most likely, I'll get a comment from him or an another team member in the future. There's no need to hurry.

#1658 is a good point: Both need a additional software for WPA on 802.11s (is done using TLS-libs in software on some devices). I also expect conflicting side-effects from applying uci-scripts .. that's why I think automated integration tests are vital. But I don't like to work on this no, because #1658 is not merged yet and maybe subjected to changes (e.g. https://github.com/freifunk-gluon/gluon/pull/1658/commits/7b397029c249829664a51ca31e77ffe8954c7784 was before https://github.com/freifunk-gluon/gluon/commit/134a6460a7d562f74d2737c0d44197f05a04dc64)

package/gluon-wlan-encryption-psk/check_site.lua
```diff
@@ -0,0 +1 @@
+need_table({'wlan_encryption_psk'})
```
2tata: According to the [comment of CodeFetch](https://github.com/freifunk-gluon/gluon/pull/1719/files#r324689064) the min and may length should be checked by check_site.lua

yanosz:
Closed, based on the motivation outlined in #1721.

CodeFetch:
@yanosz I don't understand the problem now. If it works it's fine. The changes can be easily reviewed and I say ACK. There are no conflicts with the SAE approach except that both should not be installed at the same time. And if there are conflicts these have to be handled in the SAE pull-request if your's gets merged before (like removing an old configuration of the other package / making sure that the wireless configuration is clean). So why do you close it now?

rotanid:
i also don't get it, i thought we get the little things fixed and then merge it?

yanosz:
point is: It's neither fun nor beneficial to spent time on this PR, just 'cause github's notification say, that sby had reviewed it and then noticing its just a comment side issue .. there ain't know straight way for disussing PRs in gluon atm. 


I'm not going to waste time on this - do your own PR if you like a merge. Automated integration tests are missing: this PR ain't ready for primetime.

rotanid:
how is it Gluons fault that someone comments on your PR? that's the way GitHub works, everyone can comment.
But it's your decision of course.

2tata:
@rotanid shall I take that and finish the few changes?

rotanid:
> @rotanid shall I take that and finish the few changes?

your call

PR #1722, opened 2019-05-10
ramips: fix R6120 factory image

blocktrron:
The factory firmware omits the JFFS2 end-marker while flashing via
web-interface. Add a 64k padding after the marker fixes this problem.

When the end-marker is not present, OpenWRT won't save the overlayfs
after initial flash.

This patch is currently pending upstream (see https://patchwork.ozlabs.org/patch/1097561/)

This should probably also land in v2018.2.


rotanid:
commit 3000 to Gluon ;-)
upstream it got merged.

PR #1724, opened 2019-05-14
gluon-config-mode-mesh-vpn: display tc limits in Mbit/s

mweinelt:
untested as I currently lack a build env and it will be a few days before I will.

### DIFF
--- a/package/gluon-config-mode-mesh-vpn/luasrc/lib/gluon/config-mode/wizard/0300-mesh-vpn.lua
+++ b/package/gluon-config-mode-mesh-vpn/luasrc/lib/gluon/config-mode/wizard/0300-mesh-vpn.lua
@@ -24,54 +24,39 @@ return function(form, uci)
 	local o
 
 	local meshvpn = s:option(Flag, "meshvpn", pkg_i18n.translate("Use internet connection (mesh VPN)"))
-	meshvpn.default = uci:get_bool("fastd", "mesh_vpn", "enabled") or uci:get_bool("tunneldigger", "mesh_vpn", "enabled")
+	meshvpn.default = uci:get_bool("gluon", "mesh_vpn", "enabled")
 	function meshvpn:write(data)
-		if has_fastd then
-			uci:set("fastd", "mesh_vpn", "enabled", data)
-		end
-		if has_tunneldigger then
-			uci:set("tunneldigger", "mesh_vpn", "enabled", data)
-		end
+		uci:set("gluon", "mesh_vpn", "enabled", data)
 	end
 
 	local limit = s:option(Flag, "limit_enabled", pkg_i18n.translate("Limit bandwidth"))
 	limit:depends(meshvpn, true)
-	limit.default = uci:get_bool("simple-tc", "mesh_vpn", "enabled")
+	limit.default = uci:get_bool("gluon", "mesh_vpn", "limit_enabled")
 	function limit:write(data)
-		uci:set("simple-tc", "mesh_vpn", "interface")
-		uci:set("simple-tc", "mesh_vpn", "enabled", data)
-		uci:set("simple-tc", "mesh_vpn", "ifname", "mesh-vpn")
-		if not data and has_tunneldigger then
-			uci:delete("tunneldigger", "mesh_vpn", "limit_bw_down")
-		end
+		uci:set("gluon", "mesh_vpn", "limit_enabled", data)
 	end
 
 	o = s:option(Value, "limit_ingress", pkg_i18n.translate("Downstream (kbit/s)"))
 	o:depends(limit, true)
-	o.default = uci:get("simple-tc", "mesh_vpn", "limit_ingress")
-	if has_tunneldigger then
-		-- Check if limit_bw_down exists. If not, take the value from limit_ingress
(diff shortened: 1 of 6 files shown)


kaechele:
When you test please make sure values such as

- 0,5
- 0.5
- .5
- ,5

work as especially with ADSL you still see upstream rates of less than 1 Mbit/s.
From just looking through the code I'm pretty sure 0,5 and ,5 will fail at this point.

NeoRaider:
The datatype needs to be adjusted, so the correct validator is used. We don't support the comma as a decimal point in gluon-web, but the validator will correctly show that the entered value is unacceptable.

rubo77:
We could automagically detect that if a user enters a value above 1000 he / she seems to mean kilobyte instead of megabyte

mweinelt:
Switching out units is a no-no. Goes against principle of least surprise.

mweinelt:
If someone likes to pick this up, please do. I still have no development environment as NixOS and OpenWrt don't want to play nice together.

CodeFetch:
We should add a minimal bitrate. Less than 1 mbit/s does not make sense.

rotanid:
for a 16/1 ADSL line, less than 1mbit/s upload limit might make sense.

CodeFetch:
If you're the only user besides the gluon node in your private network and you and a node user try to upload a file, doesn't that mean that each one of you gets about half of the bandwidth/0.5 mbit/s? Is there really a need for doing throttling to less than 0.5 mbit/s?

rotanid:
ok, with 100kbit/s offload most freifunk communities won't work at all, especially without segmentation :D

Adorfer:
@rotanid we miss an indicator on the status page/respondd-data telling something like "50% of upstream packets and/or upstream bandwidth used by broadcast/background traffic". 
This would be similar to "airtime-indicator" on wifi: Telling users "sorry, but you have a link which is not very productive, consider action" (e.g. release the tc limit....)
(sorry for the derail)

rotanid:
derailing comments will be hidden soon, answer though: as you should know, this is not simple, because the node doesn't know about the available bandwidth - and it might even vary.

Adorfer:
I assume that the node knowth about the tc limit, because that's what we are discussing here!
off course the "knowledge" will be lost/not available by the time the router is in config mode.

in other words: i oppose to setting a hard lower limit as you propose for either direction, or saying "value must be MBit/s". **Perhaps a slider with progressive scaling could be a solution.**
a **warning** in the webui would be nice if a unreasonably low value is entered (like 100kBit/s).

But what's more important would be to have in respondd/statuspage a value/warning telling _"how much  of a link is used up by management of Batman plus Broadcasts incl. RA/ND/DHCP etc)

rubo77:
Isn't a reasonable low band with depending on the total size of the local Mesh Cloud ?

so in a very small cloud with less than fifty nodes such a band with would still be reasonable

Adorfer:
or in a babel network. therefor a non-linear scaled slider should help to reduce input error (but I have no idea if that feasible in the current luci)

rotanid:
@Adorfer ok, i didn't get before that you would assume "upstream bandwidth" is the configured limit (if any limit is set)
the other suggestion i don't understand, as the config mode can't depend on the size of the mesh as it isn't active at that time.

nevertheless, someone will need to step up and continue work on this (we have 97 open issues and they will only get more if people always report/ask for things but don't fix/do 'em)

package/gluon-config-mode-mesh-vpn/luasrc/lib/gluon/config-mode/wizard/0300-mesh-vpn.lua
```diff
@@ -46,31 +46,31 @@ return function(form, uci)
 		end
 	end
 
-	o = s:option(Value, "limit_ingress", pkg_i18n.translate("Downstream (kbit/s)"))
+	o = s:option(Value, "limit_ingress", pkg_i18n.translate("Downstream (Mbit/s)"))
 	o:depends(limit, true)
-	o.default = uci:get("simple-tc", "mesh_vpn", "limit_ingress")
+	o.default = uci:get("simple-tc", "mesh_vpn", "limit_ingress") / 1024
```
CodeFetch: The site.conf parameters for TC are optional. The need_number-function will return nil and the parameter will thus not be added to the uci configuration in /lib/gluon/upgrade/500-mesh-vpn in that case which will result in uci:get returning nil. A division of a nil value will result in an error.

You need to check the return value first.
`o.default = uci:get("simple-tc", "mesh_vpn", "limit_ingress") or 0
o.default = o.default / 1024`

Another option is to make a default value for TC mandatory in site.conf which makes sense as we need to implicitly set a default value here.

mweinelt:
Sorry, I currently lack motivation to finish this.

CodeFetch:
That's a pity. I thought the comments made clear that yours is the only way that makes sense and it could be merged as is after reviews and testing.

NeoRaider:
Will also need to be rebased onto #1743, but I can take care of that (as well as the fix for the nil setting).

NeoRaider:
Updated with the following changes:
- rebased onto #1743
- avoid error with nil defaults
- scale by 1000 instead of 1024, so old values that were entered in kbit/s do not end up as long decimal fractions like `2000 / 1024 = 1.953125`

PR #1725, opened 2019-05-20
gluon-core gluon-config-mode: fix two bugs in WiFi configuration

blocktrron:
**gluon-config-mode gluon-core: fix incorrect gluon.platform reference**

The is_outdoor function is placed inside the gluon.platform module, not
the platform_info module. Currently, the outdoor-mode wizard component
and the upgrade script fail due to nil-value calls.

**gluon-core: fix mesh radios not being created**

This commit fixes a bug where on first setup, mesh interfaces won't be
created for 2.4GHz radios.


rotanid:
according to my contact with @blocktrron via IRC it is build- and run-tested, therefore merging it now.

PR #1729, opened 2019-05-24
gluon-l3roamd: add missing uc dependency

christf:
uc is used in the initscript. It must be set as dependency


genofire:
It is needed here:
https://github.com/freifunk-gluon/gluon/blob/5f197067569920d91501f253481b46eb242da789/package/gluon-l3roamd/files/etc/init.d/gluon-l3roamd#L11

Issue #1446, opened 2018-06-24
significantly increased management traffic after upgrade to Gluon v2017.1.x


rotanid:
this week we upgraded around 75% of our nodes to Gluon v2017.1.8, while about 20% were using this version already.
before, the nodes were running Gluon v2016.2.7.
(we still have ~20 nodes running Gluon v2016.1.x - i have no control over those, no autoupdate and no SSH)
in total, we have about 540 online nodes.

starting with the upgrade, the mgmt_tx of a single idle node without clients increased from 153 kbit/s to 254 kbit/s (by 66%) and the mgmt_rx from 57 kbit/s to 90 kbit/s (58%)

see the attached graph which shows the increased traffic with lowered spikes over all nodes, also the single node graph. the "all nodes" graph is capped at 1000 kbit/s to be able to see something useful, the spikes would make that impossible without the cap.

@ecsv also reported similar data from his community.

![2018-06-24_15-39_gluon_network_mgmt_traffic_increase](https://user-images.githubusercontent.com/654617/41819710-5f71b426-77c5-11e8-8c77-60d725f26743.png)
![2018-06-24_15-41_gluon_node_v201718_traffic_rates](https://user-images.githubusercontent.com/654617/41819711-5fa53422-77c5-11e8-945b-3daf444bb1a3.png)

rubo77:
Any ideas already why this happens?

ecsv:
It is currently unknown why this happens. I have two main suspects

* batman-adv sends more ogms for an unknown reason
* the changes in the wireless layer increased the reliability of broadcasts (which are used to transport OGMs)

The effect can only be seen when a significant portion nodes are updated from 2016.2.x to 2017.1.x

Regarding the batman-adv first point: Freifunk Vogtland increased its batman-adv version during the update from 2017.0.1+norebroadcast to 2018.1+maint+norebroadcast patches. Rotanid's update changed batman-adv from 2016.2.x+maint patches+norebroadcast to 2017.2+maint patches. I would therefore guess that the culprit (when batman-adv is the reason) has to be searched between batman-adv v2017.0.1..v2017.2

----

[...]

PR #1733, opened 2019-05-29
batman-adv: update to current openwrt-routing / batman-adv v2019.2

T-X:
This updates the batman-adv OpenWrt package to the current version provided in the master branch of the openwrt-routing packages repository (minus config changes).

The second patch then enables the batman-adv multicast optimizations again.

### DIFF
--- a/package/gluon-mesh-batman-adv/luasrc/lib/gluon/upgrade/310-gluon-mesh-batman-adv-mesh
+++ b/package/gluon-mesh-batman-adv/luasrc/lib/gluon/upgrade/310-gluon-mesh-batman-adv-mesh
@@ -19,7 +19,7 @@ uci:section('batman-adv', 'mesh', 'bat0', {
 	gw_sel_class = gw_sel_class,
 	hop_penalty = 15,
 	routing_algo = routing_algo,
-	multicast_mode = false,
+	multicast_mode = true,
 })
 uci:save('batman-adv')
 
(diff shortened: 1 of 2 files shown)


T-X:
This partially applies #1357:

* multicast-to-multi-unicast was added in and released with batman-adv v2019.2
* reenables multicast optimizations

What is not added from #1357 yet:

* batman-adv noflood patch
* firewall changes/relaxations

Which in practice means that this should reduce ICMPv6 (and ARP, a few changes in that for batman-adv, too) overhead. But does not allow using multicast for arbitrary addresses yet.

rotanid:
thanks! have you run-tested this? if yes and i don't see a veto in the next days, i will merge it - it's master branch after all.

T-X:
I run-tested it with x86 Gluon via KVM. batman-adv started fine in those three VMs, they were talking to each other and settings were applied.

NeoRaider:
I started a similar update a while ago, with a few more changes. I've just pushed that to the branch wip/update-batadv.

My changes are currently blocked by the implementation of additional netlink interfaces in batadv-legacy, so the same batctl commands are usable for both compat versions.

rotanid:
@NeoRaider your comment doesn't say anything though about whether you prefer to finalize your own patches or are ok with the patches prepared here by TX? please comment on this ;-)

T-X:
@NeoRaider: Awesome! Yes, updating the full openwrt-routing to master is indeed another, good option.

Ideally I would like to have batman-adv v2019.2 in master sooner than later, due to the approaching BattleMesh. So just in case, if someone were to spot any tricky issues in their setup we would be able to investigate and fix it quickly in Paris.

So if openwrt-routing were updated to master in the next few days that would be great. Otherwise, would there be any cons of applying this PR now and then later, without hurry updating to openwrt-routing master?

NeoRaider:
As I don't see batadv-legacy getting the missing netlink calls in the next few days, I propose to go with the small update for now, and update to the routing master later.

rotanid:
@T-X do you need to include the fix from 9e00ecd8e42f3466566b8dd37444e1f02299d05d / #1446 here before we merge this PR ?

T-X:
Changelog v2:
* rebased to include fix from #1446

rotanid:
this change was reverted:
b41ee49aac31c850acdfd6ac1bded99aac99fb31
c1a7733956234417739b05ff269ef8b9f0154a42
reason according to commit message:
Fixes #1756

PR #1735, opened 2019-06-02
Add watchdog script written in Lua to the gluon-mesh-vpn-tunneldigger

MPW1412:
package.

The logic is based on the former watchdog written in (B)ash.

The code was written by Robin Weiligmann <admin@robwei.me> and me.

### DIFF
--- a/package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
+++ b/package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
@@ -1,21 +1,21 @@
 #!/usr/bin/lua
 
-function restartTunneldigger()
+function restart_tunneldigger()
 	io.popen('logger -t tunneldigger-watchdog "Restarting Tunneldigger."')
 	os.execute("/etc/init.d/tunneldigger restart")
 end
 
-function readPidFile()
+function read_pid_file()
 	local pidFile = io.open("/var/run/tunneldigger.mesh-vpn.pid", "r")
-	local returnPid = "0"
-	if pidFile~=nil then 
-		returnPid = pidFile:read("*l") 
-		io.close(pidFile)
+	if not pidFile then
+		return nil
 	end
+	returnPid = pidFile:read("*l") 
+	io.close(pidFile)
 	return returnPid
 end
 
-function checkForMeshVpnNeighbours()
+function check_for_mesh_vpn_neighbours()
 	local handleBatctl = io.popen("batctl o", 'r')
 	if handleBatctl~=nil then
 		for line in handleBatctl:lines() do
@@ -33,12 +33,12 @@ local uci = require('simple-uci').cursor()
 if uci:get_bool('tunneldigger', 'mesh_vpn', 'enabled') then
 	if io.popen("pgrep tunneldigger"):read("*l") ~= readPidFile() then
 		io.popen('logger -t tunneldigger-watchdog "Process-Pid does not match with pid-File."')
-		restartTunneldigger()
+		restart_tunneldigger()
 		return
 	end
-	if checkForMeshVpnNeighbours() ~= true then
+	if check_for_mesh_vpn_neighbours() ~= true then
 		io.popen('logger -t tunneldigger-watchdog "No vpn-mesh neighbours found."')
-		restartTunneldigger()
+		restart_tunneldigger()
 		return
 	end
 end
--- a/package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
+++ b/package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
@@ -1,5 +1,7 @@
 #!/usr/bin/lua
 
+local uci = require('simple-uci').cursor()
+
 function restart_tunneldigger()
 	io.popen('logger -t tunneldigger-watchdog "Restarting Tunneldigger."')
 	os.execute('/etc/init.d/tunneldigger restart')
@@ -30,7 +32,6 @@ function has_mesh_vpn_neighbours()
 	return false
 end
 
-local uci = require('simple-uci').cursor()
 if uci:get_bool('tunneldigger', 'mesh_vpn', 'enabled') then
 	if io.popen('pgrep tunneldigger'):read('*l') ~= read_pid_file() then
 		io.popen('logger -t tunneldigger-watchdog "Process-Pid does not match with pid-File."')
--- a/package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
+++ b/package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
@@ -1,0 +1,43 @@
+#!/usr/bin/lua
+
+function restartTunneldigger()
+	io.popen('logger -t tunneldigger-watchdog "Restarting Tunneldigger."')
+	os.execute("/etc/init.d/tunneldigger restart")
+end
+
+function readPidFile()
+	local pidFile = io.open("/var/run/tunneldigger.mesh-vpn.pid", "r")
+	local returnPid = "0"
+	if pidFile~=nil then 
+		returnPid = pidFile:read("*l") 
+		io.close(pidFile)
+	end
+	return returnPid
+end
+
+function getNeighbourCount()
+        local list = io.lines("/sys/kernel/debug/batman_adv/bat0/originators")
+        if list~=nil then
+                for line in list do
+                        if line:find('mesh%-vpn') then
+                                return 1
+                        end
+                end
+        end
+        return nil
+end
+
+local uci = require('simple-uci').cursor()
+local tunneldiggerEnabled = uci:get_bool('tunneldigger', 'mesh_vpn', 'enabled') 
+if tunneldiggerEnabled then
+	if io.popen("pgrep tunneldigger"):read("*l") ~= readPidFile() then
+		io.popen('logger -t tunneldigger-watchdog "Process-Pid does not match with pid-File."')
+		restartTunneldigger()
+		return
+	end
+        if getNeighbourCount() ~= 1 then
+		io.popen('logger -t tunneldigger-watchdog "No vpn-mesh neighbours found."')
+		restartTunneldigger()
+		return
+        end
+end
--- a/package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
+++ b/package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
@@ -16,20 +16,21 @@ function readPidFile()
 end
 
 function getNeighbourCount()
-        local list = io.lines("/sys/kernel/debug/batman_adv/bat0/originators")
-        if list~=nil then
-                for line in list do
+	local handleBatctl = io.popen("batctl o", 'r')
+        if handleBatctl~=nil then
+                for line in handleBatctl:lines() do
                         if line:find('mesh%-vpn') then
+				handleBatctl.close()
                                 return 1
                         end
                 end
         end
+	handleBatctl.close()
         return nil
 end
 
 local uci = require('simple-uci').cursor()
-local tunneldiggerEnabled = uci:get_bool('tunneldigger', 'mesh_vpn', 'enabled') 
-if tunneldiggerEnabled then
+if uci:get_bool('tunneldigger', 'mesh_vpn', 'enabled') then
 	if io.popen("pgrep tunneldigger"):read("*l") ~= readPidFile() then
 		io.popen('logger -t tunneldigger-watchdog "Process-Pid does not match with pid-File."')
 		restartTunneldigger()
--- a/package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
+++ b/package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
@@ -16,16 +16,16 @@ function read_pid_file()
 end
 
 function check_for_mesh_vpn_neighbours()
-	local handle_batctl = io.popen("batctl o", 'r')
-	if handle_batctl~=nil then
-		for line in handle_batctl:lines() do
+	local handle = io.popen("batctl o", 'r')
+	if handle~=nil then
+		for line in handle:lines() do
 			if line:find('mesh%-vpn') then
-				handle_batctl:close()
+				handle:close()
 				return true
 			end
 		end
 	end
-	handle_batctl:close()
+	handle:close()
 	return false
 end
 
--- a/package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
+++ b/package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
@@ -6,37 +6,37 @@ function restart_tunneldigger()
 end
 
 function read_pid_file()
-	local pidFile = io.open("/var/run/tunneldigger.mesh-vpn.pid", "r")
-	if not pidFile then
+	local pid_file = io.open("/var/run/tunneldigger.mesh-vpn.pid", "r")
+	if not pid_file then
 		return nil
 	end
-	returnPid = pidFile:read("*l") 
-	io.close(pidFile)
-	return returnPid
+	pid = pid_file:read("*l") 
+	pid_file:close()
+	return pid
 end
 
 function check_for_mesh_vpn_neighbours()
-	local handleBatctl = io.popen("batctl o", 'r')
-	if handleBatctl~=nil then
-		for line in handleBatctl:lines() do
+	local handle_batctl = io.popen("batctl o", 'r')
+	if handle_batctl~=nil then
+		for line in handle_batctl:lines() do
 			if line:find('mesh%-vpn') then
-				handleBatctl.close()
+				handle_batctl:close()
 				return true
 			end
 		end
 	end
-	handleBatctl.close()
+	handle_batctl:close()
 	return false
 end
 
 local uci = require('simple-uci').cursor()
 if uci:get_bool('tunneldigger', 'mesh_vpn', 'enabled') then
-	if io.popen("pgrep tunneldigger"):read("*l") ~= readPidFile() then
+	if io.popen("pgrep tunneldigger"):read("*l") ~= read_pid_file() then
 		io.popen('logger -t tunneldigger-watchdog "Process-Pid does not match with pid-File."')
 		restart_tunneldigger()
 		return
 	end
-	if check_for_mesh_vpn_neighbours() ~= true then
+	if not check_for_mesh_vpn_neighbours() then
 		io.popen('logger -t tunneldigger-watchdog "No vpn-mesh neighbours found."')
 		restart_tunneldigger()
 		return
--- a/package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
+++ b/package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
@@ -17,7 +17,7 @@ end
 
 function check_for_mesh_vpn_neighbours()
 	local handle = io.popen("batctl o", 'r')
-	if handle~=nil then
+	if not handle then
 		for line in handle:lines() do
 			if line:find('mesh%-vpn') then
 				handle:close()
--- a/package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
+++ b/package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
@@ -2,22 +2,22 @@
 
 function restart_tunneldigger()
 	io.popen('logger -t tunneldigger-watchdog "Restarting Tunneldigger."')
-	os.execute("/etc/init.d/tunneldigger restart")
+	os.execute('/etc/init.d/tunneldigger restart')
 end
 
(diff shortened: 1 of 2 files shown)


package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
```diff
@@ -0,0 +1,43 @@
...
+
+local uci = require('simple-uci').cursor()
+local tunneldiggerEnabled = uci:get_bool('tunneldigger', 'mesh_vpn', 'enabled') 
+if tunneldiggerEnabled then
+	if io.popen("pgrep tunneldigger"):read("*l") ~= readPidFile() then
+		io.popen('logger -t tunneldigger-watchdog "Process-Pid does not match with pid-File."')
+		restartTunneldigger()
+		return
+	end
+        if getNeighbourCount() ~= 1 then
```
mweinelt: indent

mweinelt: also whats the point here? this checks if neighbor count is not equal 1 and then goes on to log, that no mesh neighbours have been found. Do you possibly mean the following?

`if getNeighbourCount() < 1 then`

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
```diff
@@ -0,0 +1,43 @@
...
+		io.popen('logger -t tunneldigger-watchdog "No vpn-mesh neighbours found."')
+		restartTunneldigger()
+		return
+        end
```
mweinelt: indent

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
```diff
@@ -0,0 +1,43 @@
...
+                                return 1
+                        end
+                end
+        end
+        return nil
+end
+
+local uci = require('simple-uci').cursor()
+local tunneldiggerEnabled = uci:get_bool('tunneldigger', 'mesh_vpn', 'enabled') 
+if tunneldiggerEnabled then
```
mweinelt: there is no point to bind that name if you are only using it once.

mweinelt:
Please use tabs for indentation everywhere.

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/lua
+
+function restartTunneldigger()
+	io.popen('logger -t tunneldigger-watchdog "Restarting Tunneldigger."')
```
mweinelt: wouldn't a print statement suffice here to make it show up in logread?

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
```diff
@@ -0,0 +1,43 @@
...
+	local returnPid = "0"
+	if pidFile~=nil then 
+		returnPid = pidFile:read("*l") 
+		io.close(pidFile)
+	end
+	return returnPid
+end
+
+function getNeighbourCount()
+        local list = io.lines("/sys/kernel/debug/batman_adv/bat0/originators")
```
mweinelt: the debugfs is deprecated and not even present anymore on recent batman-adv versions.

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
MPW1412: I'm not sure, would it?

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
MPW1412: What do you suggest, accessing it via `batctl originators`?

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
MPW1412: There is, self explaining coding conventions to make code easily readable without comments. The only reason not to do this would be to save some space.

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
mweinelt: Nah, micron does not seem to log stdout/stderr to syslog.

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
mweinelt: Yes, use `batctl o -H` if you must. It still uses the debugfs because Gluon still runs on batman-adv 2018.1.

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
MPW1412: Jup, just tried it myself, doesn't work :)

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
MPW1412: Okay, will do.

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
MPW1412: I changed this. Let's not discuss about it :)

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
MPW1412: Changed in latest commit.

MPW1412:
Thanks for your feedback, @mweinelt. The latest commit does the requested changes.

MPW1412:
> Please use tabs for indentation everywhere.

Done in last commit. Had overseen it before.

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
```diff
@@ -0,0 +1,44 @@
...
+	local pidFile = io.open("/var/run/tunneldigger.mesh-vpn.pid", "r")
+	local returnPid = "0"
+	if pidFile~=nil then 
+		returnPid = pidFile:read("*l") 
+		io.close(pidFile)
+	end
+	return returnPid
+end
+
+function getNeighbourCount()
```
mweinelt: The code looks more like a `hasTunneldigerNeighbours` than a request for a count. You return one of (1, nil) curently but (true, false) would be more appropriate then.

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
```diff
@@ -0,0 +1,44 @@
...
+                        end
+                end
+        end
+	handleBatctl.close()
+        return nil
+end
+
+local uci = require('simple-uci').cursor()
+if uci:get_bool('tunneldigger', 'mesh_vpn', 'enabled') then
+	if io.popen("pgrep tunneldigger"):read("*l") ~= readPidFile() then
```
mweinelt: Did that actually happen in your tests? I suppose having a valid pidfile is necessary for the restart to work?

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
MPW1412: True, we had first counted the neighbours, before we noticed that that wastes cpu cycles.

nil and false is the same in lua, just like 1 (or any number) and true. I'll change it anyway.

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
MPW1412: Actually the init.d-script for the tunneldigger is broken:

```
root@Testknoten-MPW-TD-Watchdog-Master:/# /etc/init.d/tunneldigger restart
Stopping tunneldigger for interface mesh-vpn
[47430.044632] batman_adv: bat0: Interface deactivated: mesh-vpn
  waiting for tunneldigger to stop
[47430.051001] batman_adv: bat0: Removing interface: mesh-vpn
  tunneldigger stopped
Not starting tunneldigger - missing uuid
Starting tunneldigger on mesh-vpn
```

I wanted to look into that aswell. But one thing at a time.

MPW1412: To answer your question: We tested the code by manipulating the pid-file. But I don't know if this might actually happen. 

We didn't think about the logic, just copied it from the old (b)ash implementation.

mweinelt:
The Gluon codebase uses lowercase function names with underlines separating words, not camel casing as you do.

`readPidFile` vs `read_pid_file`

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/lua
+
+function restartTunneldigger()
+	io.popen('logger -t tunneldigger-watchdog "Restarting Tunneldigger."')
+	os.execute("/etc/init.d/tunneldigger restart")
+end
+
+function readPidFile()
```
mweinelt: I would turn the logic in this function around like this

```lua
function read_pid_file()
	local file = io.open("/var/run/tunneldigger.mesh-vpn.pid", "r")
	if not file:
		return nil
	pid = file:read("*l")
	file:close()
	return pid
end
```

MPW1412: Strongly disagree. But we're not the maintainers :).

mweinelt: I'm not a lua fanatic either, so I don't have a strong opinion about this.

MPW1412:
> The Gluon codebase uses lowercase function names with underlines separating words, not camel casing as you do.
> 
> `readPidFile` vs `read_pid_file`

✓

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
```diff
@@ -0,0 +1,44 @@
...
+	if not pidFile then
+		return nil
+	end
+	returnPid = pidFile:read("*l") 
+	io.close(pidFile)
+	return returnPid
+end
+
+function check_for_mesh_vpn_neighbours()
+	local handleBatctl = io.popen("batctl o", 'r')
```
mweinelt: camel case

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/lua
+
+function restart_tunneldigger()
+	io.popen('logger -t tunneldigger-watchdog "Restarting Tunneldigger."')
+	os.execute("/etc/init.d/tunneldigger restart")
+end
+
+function read_pid_file()
+	local pidFile = io.open("/var/run/tunneldigger.mesh-vpn.pid", "r")
```
mweinelt: camel case

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
```diff
@@ -0,0 +1,44 @@
...
+	if not pidFile then
+		return nil
+	end
+	returnPid = pidFile:read("*l") 
+	io.close(pidFile)
```
mweinelt: pidFile:close()

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
```diff
@@ -0,0 +1,44 @@
...
+	local handleBatctl = io.popen("batctl o", 'r')
+	if handleBatctl~=nil then
+		for line in handleBatctl:lines() do
+			if line:find('mesh%-vpn') then
+				handleBatctl.close()
+				return true
+			end
+		end
+	end
+	handleBatctl.close()
```
mweinelt: handleBatctl:close()

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
```diff
@@ -0,0 +1,44 @@
...
+end
+
+local uci = require('simple-uci').cursor()
+if uci:get_bool('tunneldigger', 'mesh_vpn', 'enabled') then
+	if io.popen("pgrep tunneldigger"):read("*l") ~= readPidFile() then
+		io.popen('logger -t tunneldigger-watchdog "Process-Pid does not match with pid-File."')
+		restart_tunneldigger()
+		return
+	end
+	if check_for_mesh_vpn_neighbours() ~= true then
```
mweinelt: if something ~= true then

equals

if not something then

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
MPW1412: Do you mean the variable name?

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
```diff
@@ -6,37 +6,37 @@ function restart_tunneldigger()
...
+	pid_file:close()
+	return pid
 end
 
 function check_for_mesh_vpn_neighbours()
-	local handleBatctl = io.popen("batctl o", 'r')
-	if handleBatctl~=nil then
-		for line in handleBatctl:lines() do
+	local handle_batctl = io.popen("batctl o", 'r')
+	if handle_batctl~=nil then
```
mweinelt: if not handle_batctl then, in fact I would call the variable something like `response`.

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
```diff
@@ -0,0 +1,44 @@
...
+	end
+	pid = pid_file:read("*l") 
+	pid_file:close()
+	return pid
+end
+
+function check_for_mesh_vpn_neighbours()
+	local handle = io.popen("batctl o", 'r')
+	if not handle then
+		for line in handle:lines() do
```
blocktrron: This loop (and all nested code) is only reached in case your `handle` is `nil`. This is most likely not what you wanted to achieve.

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
```diff
@@ -0,0 +1,44 @@
...
+	io.popen('logger -t tunneldigger-watchdog "Restarting Tunneldigger."')
+	os.execute("/etc/init.d/tunneldigger restart")
+end
+
+function read_pid_file()
+	local pid_file = io.open("/var/run/tunneldigger.mesh-vpn.pid", "r")
+	if not pid_file then
+		return nil
+	end
+	pid = pid_file:read("*l")
```
blocktrron: There should be no reason to declare `pid` as a global variable here. So declare it as a local variable.

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
```diff
@@ -0,0 +1,44 @@
...
+	local pid_file = io.open("/var/run/tunneldigger.mesh-vpn.pid", "r")
+	if not pid_file then
+		return nil
+	end
+	pid = pid_file:read("*l") 
+	pid_file:close()
+	return pid
+end
+
+function check_for_mesh_vpn_neighbours()
```
blocktrron: Oh, and renaming this function to something like `has_mesh_vpn_neighbours` would better reflect the return value IMHO.

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
MPW1412: Ah thanks. Must have happen during the refractoring :)

MPW1412:
Thanks for your feedback, @blocktrron. I tried to address all points in the last commit.

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
```diff
@@ -0,0 +1,44 @@
...
+	local handle = io.popen('batctl o', 'r')
+	if handle then
+		for line in handle:lines() do
+			if line:find('mesh%-vpn') then
+				handle:close()
+				return true
+			end
+		end
+	end
+	handle:close()
```
blocktrron: You can now end up calling `close` on a `nil` value here.

Structure it like

```
if not handle then
  return false
end
for line in handle:lines() do
  .....
end
handle:close()
return false
```

MPW1412: ✓

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
```diff
@@ -0,0 +1,45 @@
...
+		if line:find('mesh%-vpn') then
+			handle:close()
+			return true
+		end
+	end
+	handle:close()
+	return false
+end
+
+local uci = require('simple-uci').cursor()
```
mweinelt: please move this import to the top of the file

mweinelt:
Great progress, please test the latest version and report back.

MPW1412:
I just tested it in x86-64. Works :)

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
mweinelt: What about the tunneldigger init script? You said in https://github.com/freifunk-gluon/gluon/pull/1735#discussion_r289643359 that it was broken.

MPW1412: As you can see from the output, it behaves a little weird but does its job. The init script is not part of this PR. So we should discuss that elsewhere, maybe in an issue. I tried to understand what the script does, but didn't really get the idea behind the lines of codes. To me it just seems over engineered.

mweinelt: You said it was broken. If that were true the watchdog wouldn't work, hence why I asked. I assume it works sufficiently well?

MPW1412: Yes, for now it works.

MPW1412:
I already tested it...works. Do you want a second opinion?

mweinelt:
Yup, that would be nice. I'm not very comfortable yet reviewing Lua.

CodeFetch:
Looks good. Why did you use popen instead of execute for the logger? I'd use execute as it does not create a file handle in memory or at least directly closes it. Lua's garbage collection is not very clever and it might be possible that the file handle is only getting freed on garbage collection or program termination, but I'm not sure as you don't assign it to a variable at all... Just to be sure I'd use execute. This is still pedantic, as file handles don't really hurt. Your code is fine.

MPW1412:
Hello @CodeFetch,

thanks for your feedback. 

About the logging: Robin and I are not very familiar with Lua. When we first started this little project, we expected Lua to have some command or package to write to the syslog. But we couldn't find anything in the other gluon packages.

So we just used the first best command we knew. 

About edit(1): We don't know that the idea behind this is. We just ported the old (b)ash script to Lua as we were told (b)ash scripts wouldn't be accepted in Gluon anymore.

About edit(2): Okay, we'll stay with popen for now.

Best,
Matthias

package/gluon-mesh-vpn-tunneldigger/files/usr/bin/tunneldigger-watchdog
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/lua
+
+local uci = require('simple-uci').cursor()
+
+function restart_tunneldigger()
+	io.popen('logger -t tunneldigger-watchdog "Restarting Tunneldigger."')
```
mweinelt: Since we don't parse any return values we might use `os.execute` here as well, no?

Adorfer:
works as expected on 841: After killing the tunneldigger process, the checkscript detected and made a new tunnel.

mweinelt:
Squashe and merged with minor changes in 8e17635.

Thanks for your work!

PR #1738, opened 2019-06-06
Add minimal docker environment

mweinelt:
This will be a useful precursor for continuous integration.

I'm running NixOS at home and I use this container to build OpenWrt and Gluon.

Once merged we should set up automatic builds in the Docker Hub.

### DIFF
--- a/contrib/Dockerfile
+++ b/contrib/Dockerfile
@@ -1,0 +1,26 @@
+FROM debian:buster-slim
+
+RUN apt update && apt install -y --no-install-recommends \
+    ca-certificates \
+    file \
+    git \
+    subversion \
+    python \
+    build-essential \
+    gawk \
+    unzip \
+    libncurses5-dev \
+    zlib1g-dev \
+    libssl-dev \
+    libelf-dev \
+    wget \
+    time \
+    ecdsautils \
+    lua-check \
+  && rm -rf /var/lib/apt/lists/*
+
+RUN useradd -d /gluon gluon
+USER gluon
+
+VOLUME /gluon
+WORKDIR /gluon


bobcanthelpyou:
Why is Debian Buster (still testing) used and not Debian Stretch (stable)? Might it be useful to add the package `ecdsautils`?

mweinelt:
Buster is past its freeze, so it will not change much anymore. ecdsautils is a good catch, thanks

I'll probably need to move  the Dockerfile to a dedicated sub-directory, or else the whole of the Gluon directory will be considered build context:

```
$ docker build -t gluon --pull .
Sending build context to Docker daemon  13.75GB
Step 1/6 : FROM debian:buster-slim
[...]
```

NeoRaider:
> Buster is past its freeze, so it will not change much anymore. ecdsautils is a good catch, thanks
> 
> I'll probably need to move the Dockerfile to a dedicated sub-directory, or else the whole of the Gluon directory will be considered build context:
> 
> ```
> $ docker build -t gluon --pull .
> Sending build context to Docker daemon  13.75GB
> Step 1/6 : FROM debian:buster-slim
> [...]
> ```

Or add a .dockerignore just containing the line `*` to ignore everything. `contrib` sounds like a good place for the Dockerfile, but adding another subdir could be avoided.

mweinelt:
Okay, moved into `contrib/`.

mweinelt:
Added `libelf-dev` to satisfy

> warning: Cannot use CONFIG_STACK_VALIDATION=y, please install libelf-dev, libelf-devel or elfutils-libelf-devel

mweinelt:
Added the lua-check package, which will probably be part of our first ci job. :heart:

herbetom:
The following commands are helpfull for using the Dockerfile: 

`docker build -t gluon --pull contrib`
`docker run -it --rm -v $(pwd):/gluon gluon`

Thanks to @mweinelt for those!

PR #1744, opened 2019-06-15
check_site: change error title to something more unique

rubo77:
This small change will make those errors more easily distinguishable from other less important errors in the build-log.

### DIFF
--- a/scripts/check_site.lua
+++ b/scripts/check_site.lua
@@ -1,7 +1,7 @@
 local cjson = require 'cjson'
 
 local function config_error(src, ...)
-	error(src .. ' error: ' .. string.format(...), 0)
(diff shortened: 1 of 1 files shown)


NeoRaider:
The variable 'src' already contains the file name (either site.conf, or a specific domain config file name).

rubo77:
I was more thinking about a general uniqueness

NeoRaider:
The message literally starts with `site.conf error:`, you changed that to `site.conf - error in site.conf:`.

rubo77:
Do they all show up as "

mweinelt:
This error type is always visible right at the end of your build.

rubo77:
OK, so I added it to https://github.com/freifunk-gluon/gluon/wiki/Troubleshooting

maybe you know some more unambiguous strings, we could list there

PR #1749, opened 2019-06-16
gluon-core: fix outdoor board names

mweinelt:
The board names for CPE510, Nanostation Loco M5 and Nanostation M5 are incorrect.

I've personally verified the names against a CPE510 v1.1 and a Nanostation Loco M2 (XM).

Sorry for the mess.

### DIFF
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua
@@ -24,9 +24,16 @@ end
 
 function M.is_outdoor_device()
 	if M.match('ar71xx', 'generic', {
-		'cpe510-520-v1',
-		'ubnt-nano-m',
-		'ubnt-nano-m-xw',
+		'bullet-m',
+		'cpe510',
+		'lbe-m5',
+		'loco-m-xw',
+		'nanostation-m',
+		'nanostation-m-xw',
+		'rocket-m',
+		'rocket-m-ti',
+		'rocket-m-xw',
+		'unifi-outdoor',
 	}) then
 		return true
 


NeoRaider:
What about:
- rocket-m
- rocket-m-ti
- rocket-m-xw
- loco-m-xw
- lbe-m5 (not supported by Gluon, but we might as well add it)

blocktrron:
To fill this list with possibly all outdoor devices currently, the following are missing:

 - `rocket-m`
 - `rocket-m-ti`
 - `rocket-m-xw`
 - `loco-m-xw`

blocktrron:
Oh and while we are at it:
 - `unifi-outdoor`

The `unifi-outdoor-plus` is 2.4 GHz only.

mweinelt:
Thanks, that's what I was aiming for :+1:

mweinelt:
```
22:38 <@hexa-> DirkS: kannst du auf 64331-Dachgateway17 mal /tmp/sysinfo/board_name pasten?
09:08 <@DirkS> hexa-: cat /tmp/sysinfo/board_name
09:08 <@DirkS> nanostation-m
```

https://meshviewer.darmstadt.freifunk.net/#!/en/map/24a43cf41a21

So I think this is correct as it is.

CodeFetch:
Very strange... Another thing is the UniFi AC Mesh Lite which is an outdoor router, but shares its image with indoor routers. board_name shows unifiac-lite on this device...

mweinelt:
https://github.com/freifunk-gluon/gluon/blob/4c025ec3be1b5b9b0b1781e10996346a646a04ad/package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua#L41

CodeFetch:
Okay, confirmed this is correct. So I think it's all okay. Only thing I still wonder about is what a PowerBeam/NanoBeam shows, but I haven't deployed one. 

The reason why I'm interested is that I might need to use this logic to define the correct antenna gains. If there is no possibility to differ between a NanoStation and a NanoBeam there must be a missing mechanism that AirOS uses (likely subvendor device ID of radio module).

mweinelt:
Best to check how OpenWrt handles this Upstream. Thanks for checking the list!

CodeFetch:
@mweinelt The ar71xx.sh checks the subvendor device ID of the radio module as I've guessed, but there are many definitions missing. I'm going to make a pull-request and ping you, when it's merged...

PR #1751, opened 2019-06-17
fix luacheck warnings : fix comments accordingly

rubo77:
Due to the fixes in 4249d65af73ee9228d827af913b7caf2aac00809 the comments have to change too

### DIFF
--- a/package/gluon-web-mesh-vpn-fastd/luasrc/lib/gluon/config-mode/model/admin/mesh_vpn_fastd.lua
+++ b/package/gluon-web-mesh-vpn-fastd/luasrc/lib/gluon/config-mode/model/admin/mesh_vpn_fastd.lua
@@ -20,7 +20,7 @@ end
 function mode:write(data)
 	local site = require 'gluon.site'
 
-	-- methods will be recreated and filled with the original values from site.mesh_vpn.fastd.methods
+	-- site_methods will be recreated and filled with the original values from site.mesh_vpn.fastd.methods
 	-- if performance mode was selected, and the method 'null' was not present in the original table, it will be added
 	local site_methods = {}
 	if data == 'performance' then


package/gluon-web/luasrc/usr/lib/lua/gluon/web/http/protocol.lua
```diff
@@ -223,9 +223,7 @@ local function mimedecode_message_body(src, msg, filecb)
 			until not spos
 
 			if found then
-				-- We found at least some boundary. Save
-				-- the unparsed remaining data for the
-				-- next chunk.
+				-- We found at least some boundary.
```
NeoRaider: The comment is still correct, there is nothing to change here.

rubo77: But where is the remaining data used after that line? the if clause ends and else is another case.

NeoRaider: lchunk is used again in the next call of snk. Unlike data, lchunk is an upvalue.

rubo77: AH, ow I get it: the comment was misleading, I thought, with the remaining "data" they mean the variable "data", but the comment is not about the variable but just uses the word "data"

maybe we should rephrase this then?

> save the unparsed remaining data-part for the next chunk.

rubo77: changed to "the remaining string"

NeoRaider: Even if it refered to the "variable" data, the comment would still be accurate: The remaining (variable) "data" is saved (by assigning it to a different variable). Please just leave this comment unchanged.

rubo77: ok, done ;)

Issue #1756, opened 2019-06-17
deprecated messages of batman-adv 


bobcanthelpyou:
Deprecated messages of batman_adv via `dmesg` with the latest Gluon master.
```
[   10.164627] batman_adv: B.A.T.M.A.N. advanced openwrt-2019.2-1 (compatibility version 15) loaded
[   19.781858] batman_adv: [Deprecated]: sh (pid 1201) Use of sysfs file "mesh_iface".
[   19.781858] Use batadv genl family instead
[   22.481837] batman_adv: [Deprecated]: sh (pid 1660) Use of sysfs file "gw_mode".
[   22.481837] Use batadv genl family instead
[   22.634668] batman_adv: [Deprecated]: sh (pid 1660) Use of sysfs file "hop_penalty".
[   22.634668] Use batadv genl family instead
[   22.780563] batman_adv: [Deprecated]: sh (pid 1660) Use of sysfs file "multicast_mode".
[   22.780563] Use batadv genl family instead
[   22.893830] batman_adv: [Deprecated]: sh (pid 1660) Use of sysfs file "orig_interval".
[   22.893830] Use batadv genl family instead
[   26.258930] batman_adv: [Deprecated]: sh (pid 1989) Use of sysfs file "mesh_iface".
[   26.258930] Use batadv genl family instead
```

Might be introduced by https://github.com/open-mesh-mirror/batman-adv/commit/5bbd4721bfda46839addc9089135450d0e35f459

https://github.com/open-mesh-mirror/batman-adv/blob/master/Documentation/ABI/obsolete/sysfs-class-net-batman-adv
https://github.com/open-mesh-mirror/batman-adv/blob/master/Documentation/ABI/obsolete/sysfs-class-net-mesh

mweinelt:
This will probably be resolved through https://github.com/freifunk-gluon/gluon/pull/1752.

bobcanthelpyou:
Upgraded to https://github.com/freifunk-gluon/gluon/commit/3d3617ed2c85c24be1b8c2de876045381d9adf0e (v2018.2-138-g3d3617ed)

```
[   22.724641] batman_adv: B.A.T.M.A.N. advanced openwrt-2019.2-1 (compatibility version 15) loaded
[   34.145626] batman_adv: [Deprecated]: sh (pid 1377) Use of sysfs file "mesh_iface".
[   34.145626] Use batadv genl family instead
[...]

PR #1760, opened 2019-06-19
alfred: upgrade package to latest release 2019.2

T-X:
This upgrades alfred to v2019.2.

It does not fix the deprecation warnings in dmesg triggered by batadv-vis yet (also see #1756). But I opened a ticket for now for that at [open-mesh.org](https://www.open-mesh.org/issues/393).

### DIFF
--- a/patches/packages/routing/0005-alfred-upgrade-package-to-latest-release-2019.2.patch
+++ b/patches/packages/routing/0005-alfred-upgrade-package-to-latest-release-2019.2.patch
@@ -1,0 +1,156 @@
+From: Linus Lüssing <linus.luessing@c0d3.blue>
+Date: Wed, 19 Jun 2019 00:52:57 +0200
+Subject: alfred: upgrade package to latest release 2019.2
+
+Signed-off-by: Linus Lüssing <linus.luessing@c0d3.blue>
+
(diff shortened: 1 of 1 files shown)


NeoRaider:
I don't think this is important for Gluon v2019.1, and we should consider dropping gluon-alfred after that.

rotanid:
@NeoRaider does "not important" mean that it shouldnt be merged?

T-X:
@NeoRaider: Yes, you are right, it's not important. There is also pretty much no functional difference, the changes to alfred are all cosmetic.

My thinking was to stay consistent with the versions of batman-adv/batctl/alfred for the matter of form. And to avoid potential confusion (e.g. someone looking at "alfred -v" and then out of habit thinking he/she would be running batman-adv v2018.x or something stupid like that).

Does that make sense?

NeoRaider:
Running "alfred -v" to find out the batadv version would be quite creative ;) I don't even think there are many sites left that still use alfred, considering the packet size issues it had when respondd was created...

I'd like to avoid adding patches to Gluon that we don't need, so I'm closing this.

T-X:
Oki doki, but thanks for your consideration :-).

PR #1776, opened 2019-07-01
ramips-mt76x8: add support for TP-Link TL-MR3420 v5

rotanid:
  - [ ] webinterface
  - [x] tftp
  - [ ] other: <specify>
- wired network
- wifi (if applicable)
- led mapping
  - radio leds
    - [x] should show activity
  - switchport leds
    - [ ] should show link state and activity [not working in stock either]
- outdoor devices only

how to restore to stock:
use stock image without its first 512 bytes and do tftp

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -381,6 +381,7 @@ ramips-mt76x8
 
 * TP-Link
 
+  - TL-MR3420 v5 [#80211s]_
   - TL-WR841N v13 [#80211s]_
   - Archer C50 v3 [#80211s]_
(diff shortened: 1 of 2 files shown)


mweinelt:
Config mode blink sequence?

rotanid:
> Config mode blink sequence?

as marked in the checklist and discussed with you on IRC it is not working, but it's not the first device where it isn't working, e.g. all CPE210 versions.
and if someone finds a fix, which we didn't so far, it can be fixed by sysupgrades unlike some other LEDs as the diag led is not mentioned in UCI config.

PR #1782, opened 2019-07-06
ramips-rt305x: mark boards with 4M flash as broken

blocktrron:
This adds the broken flag to devices with 4MB flash. They tend to lose
their configuration because of missing space for the overlayfs. For this
reason, the boards are marked as broken.

This has been observed in the network of Freifunk Darmstadt with a D-Link DIR-615L which went offline after the update to Gluon 2018.2. https://meshviewer.darmstadt.freifunk.net/#/en/map/1cbdb9ae419a

This issue is also known upstream https://forum.openwrt.org/t/report-devices-here-with-18-06-0-provided-image-too-big-to-save-overlay/18161

### DIFF
--- a/targets/ramips-rt305x
+++ b/targets/ramips-rt305x
@@ -4,6 +4,7 @@ config '# CONFIG_KERNEL_KALLSYMS is not set'
 
 device('a5-v11', 'a5-v11', {
 	deprecated = true, -- 4/32
+	broken = true,
 })
 
 
@@ -11,6 +12,7 @@ device('a5-v11', 'a5-v11', {
 
 device('d-link-dir-615-h1', 'dir-615-h1', {
 	deprecated = true, -- 4/32
+	broken = true,
 })
 
 device('d-link-dir-615-d', 'dir-615-d', {
@@ -21,6 +23,7 @@ device('d-link-dir-615-d', 'dir-615-d', {
 		'd-link-dir-615-d4',
 	},
 	deprecated = true, -- 4/32
+	broken = true,
 })
 
 


skorpy2009:
Shouldn't this be than also backported to Gluon 2018.2 since it already appears there?

mweinelt:
Awesome, that leaves the target with vocores, that nobody anywhere actually uses, I've checked.

rotanid:
before we declare a previously supported device BROKEN, we should discuss how we could make it work, like it was done with ar71xx-tiny. i know it's not fun, but marking a supported devices as BROKEN is a serious step.
@mweinelt i have a vocore2, i don't use it often, but it works fine.

mweinelt:
@rotanid None is visible on multi.meshviewer.org. I also checked ~115 map/nodes json files (https://chaos.expert/freifunkranking.de/app/blob/master/communities.json), and not a single deployed vocore turned up.

Sure, the next steps are handling it like tiny first:

```lua
config 'CONFIG_GLUON_SPECIALIZE_KERNEL=y'

no_opkg()

packages {
	'-uboot-envtools',
	'-kmod-usb-core',
	'-kmod-usb-ohci',
	'-kmod-usb2',
	'-kmod-usb-ledtrig-usbport',
}

defaults {
	deprecated = true, -- 4/32
}
```

mweinelt:
FTR: I found ~8 DIR-615 out there.

blocktrron:
@rotanid Additional tiny-targets came up multiple times upstream, but were always rejected due to the low number of affected devices. We can try to mitigate the problem for now downstream, but this is (most likely) not a future-oriented solution.

bobcanthelpyou:
I own a DIR-615-H1 and a DIR-615-D1 and for me the devices are broken, too. Not cause of the space problem but there are some wireless issues, which makes the wireless connection broken after some time (but it could also be related). I will try to reproduce the upgrade space issue first.

rotanid:
> FTR: I found ~8 DIR-615 out there.

question is, are they running v2018.2.x or something older?
can the DIR-615 work if you don't put too many packages in the build?

mweinelt:
>  can the DIR-615 work if you don't put too many packages in the build?

Besides the domain director packages we have a pretty normal package selection for rt305x and it does not work.

```
GLUON_FEATURES := \
	autoupdater \
	config-mode-domain-select \
	config-mode-geo-location-osm \
	config-mode-outdoor \
	ebtables-filter-multicast \
	ebtables-filter-ra-dhcp \
	ebtables-limit-arp \
	ebtables-source-filter \
	mesh-batman-adv-15 \
	mesh-vpn-fastd \
	radvd \
	radv-filterd \
	respondd \
	status-page \
	web-advanced \
	web-ffda-domain-director \
	web-logging \
	web-private-wifi \
	web-wizard

# Additional packages to install on every image
GLUON_SITE_PACKAGES := \
	ffda-domain-director \
	iwinfo \
	haveged \
	respondd-module-airtime
```

rotanid:
ok, so we for example remove the airtime module from 4mb flash devices in -tiny, maybe that would help?
there's also config-mode-geo-location-osm and config-mode-outdoor which are optional.
of course, in your case you need the domain director so that is not optional.
if removing the above is not enough, web-logging or even web-private-wifi might be removed?

mweinelt:
Neither web-logging nor web-private-wifi require meaningful storage space.

rotanid:
airtime.so takes around 8KB.
also, you could remove USB packages from the device or even opkg - that should be sufficient for v2018.2.x and v2019.1.x Gluon branches.

Gluon v2019.2.x based on OpenWrt 19.07 will be something else of course.

mweinelt:
>  also, you could remove USB packages from the device or even opkg - that should be sufficient for v2018.2.x and v2019.1.x Gluon branches.

As was proposed in https://github.com/freifunk-gluon/gluon/pull/1782#issuecomment-508959693. I don't have access to affected devices, so I can't test that.

bobcanthelpyou:
Did an upgrade of a DIR-615 D1 and a DIR-615 H1 device, config is still fine, our package list is https://github.com/freifunk-ilmenau/site-ffil/blob/master/site.mk (but runtime was just some minutes).

mweinelt:
That is alot more constrained than our package set.

```Makefile
GLUON_FEATURES := \
	autoupdater \
	ebtables-filter-multicast \
	ebtables-filter-ra-dhcp \
	ebtables-limit-arp \
	mesh-batman-adv-14 \
	mesh-vpn-fastd \
	radvd \
	respondd \
	status-page \
	web-advanced \
	web-wizard

GLUON_SITE_PACKAGES := \
	iptables \
	iwinfo \
	haveged
```

Our v2018.2.2 images have these sizes:

```
gluon-ffda-1.6.2-d-link-dir-615-d-sysupgrade.bin   26-Jun-2019 23:41             3670200
gluon-ffda-1.6.2-d-link-dir-615-d1-sysupgrade.bin  26-Jun-2019 23:41             3670200
gluon-ffda-1.6.2-d-link-dir-615-d2-sysupgrade.bin  26-Jun-2019 23:41             3670200
gluon-ffda-1.6.2-d-link-dir-615-d3-sysupgrade.bin  26-Jun-2019 23:41             3670200
gluon-ffda-1.6.2-d-link-dir-615-d4-sysupgrade.bin  26-Jun-2019 23:41             3670200
gluon-ffda-1.6.2-d-link-dir-615-h1-sysupgrade.bin  26-Jun-2019 23:41             3575993
gluon-ffda-1.6.2-d-link-dir-615-rev-c1-sysupgra..> 26-Jun-2019 19:19             3473412
```

Adorfer:
> ok, so we for example remove the airtime module from 4mb flash devices in -tiny, maybe that would help?

i would sacrifice a lot of things, but not the airtime-module because it's essential for radio configuration in local meshes. 

i would prefer to kill the statuspage on those devices, because it lacks most relevant data (channel, htmode, ap disabled, vpn enabled etc) for meshnetworks anyway.

bobcanthelpyou:
Our DIR-615 D* have the same size as above. The size of H1 und C1 are different
```
3670200 gluon-ffil-2018.2.2.0-ffil~20190625-d-link-dir-615-d-sysupgrade.bin
3518649 gluon-ffil-2018.2.2.0-ffil~20190625-d-link-dir-615-h1-sysupgrade.bin
3407876 gluon-ffil-2018.2.2.0-ffil~20190625-d-link-dir-615-rev-c1-sysupgrade.bin
```

kpanic23:
Isn't the A5-V11 a ramips-rt305x device, too?
I haven't noticed any problems on upgrades. Running v2018.2.2 at the moment...

rotanid:
> i would sacrifice a lot of things, but not the airtime-module because it's essential for radio configuration in local meshes.


> i would prefer to kill the statuspage on those devices, because it lacks most relevant data (channel, htmode, ap disabled, vpn enabled etc) for meshnetworks anyway.

rotanid:
> i would sacrifice a lot of things, but not the airtime-module because it's essential for radio configuration in local meshes.
@mweinelt said something similar.

> i would prefer to kill the statuspage on those devices, because it lacks most relevant data (channel, htmode, ap disabled, vpn enabled etc) for meshnetworks anyway.
1) if removing usb+opkg is enough for @mweinelt , it probably will be enough for most others too.
2) Gluon can't decide which packages the communities remove. the only thing Gluon can say is: "we know this device still works if you don't put too many packages in its image, please remove bigger ones"

NeoRaider:
There is one thing we can fix: If it is possible to build images that can't format the overlay anymore, IMAGE_SIZE needs to be adjusted to account for the overlay size - so the build will fail rather than generating broken images.

rotanid:
so... we reject this PR in favor of adjusting the IMAGE_SIZE and maybe removing usb+opkg ?

rotanid:
so what's next?
the #2226 has been merged, do we need to backport https://github.com/openwrt/openwrt/pull/2226/files#diff-680cab00e1ace40c4fa9f72e474f73e9R406 ?
do we want to remove usb+opkg for this target?

rotanid:
i've just added it as a blocker for 2019.1, as we don't want build of a new release to fail.
we need to solve it by a backport and/or remove USB+opkg .
@NeoRaider @blocktrron what do you think about this?

blocktrron:
@NeoRaider @rotanid Generating the JFFS2 overlay (reserving 5*64k) might actually not be enough for our usecase.

A device might be able to save and retain it's configuration state, however it's still possible for the device to lose it's config by doing a sysupgrade to even the exact firmware already installed.

To be kind-of-sure, we should

a) drop OPKG for this target
b) remove USB packages

for the next release and mark the 4M devices as "to-be-removed after this release".

What do you think?

rotanid:
> To be kind-of-sure, we should
> 
> a) drop OPKG for this target
> b) remove USB packages
> 
> for the next release and mark the 4M devices as "to-be-removed after this release".
> 
> What do you think?

i agree.
c) add the DEPRECATED flag to the target
d) info about c) with reasoning in the release notes

rotanid:
i suggest to close this PR in favor of #1807

PR #1784, opened 2019-07-11
update i18n files via i18n-scan.pl and msgmerge

bobcanthelpyou:
Running `i18n-scan.pl` and `msgmerge` as suggested by the Gluon docs ([Internationalization support](https://gluon.readthedocs.io/en/latest/dev/web/i18n.html)) the *.pot and *.po files where updated.

e.g.:
```
cd package/gluon-web-mesh-vpn-fastd/i18n
../../../contrib/i18n-scan.pl ../files ../luasrc > gluon-web-mesh-vpn-fastd.pot
msgmerge -U fr.po gluon-web-mesh-vpn-fastd.pot
```

CHANGES:
* translated messages are sorted by msgid
* the default line length limit seems to be 80 characters
* the Content-Type string switches to a different line 
* gluon-web-wifi-config: msgid "Outdoor installation" changed to "Outdoor Installation"
* gluon-config-mode-outdoor: add msgstr "Content-Type: text/plain; charset=UTF-8"

QUESTION:
* ~How to handle `gluon-config-mode:` msgids which should be translated by the site i18n? Should these be added, too? e.g. `gluon-config-mode:hostname-help` of package/gluon-config-mode-hostname. Or how to ignore them via `i18n-scan.pl`?~
* ~How to handle the msgid `gluon-web-node-role:role:` of the gluon-web-node-role package?~
* ~How to handle the msgid `Enable PoE Passthrough` of gluon-web-network? Should it be replaced by `Enable %s` and `network.lua` modified? See https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-web-network/luasrc/lib/gluon/config-mode/model/admin/network.lua#L121~

TODO:
* ~update i18n of docs/user/site.rst~

Using the gettext package v0.19.8.1-9 via Debian Buster


package/gluon-config-mode-contact-info/i18n/fr.po
```diff
@@ -1,33 +1,38 @@
...
+"stockées sur ce nœud et vous pourrez les effacer par vous-même à tout moment."
 
 msgid "e.g. E-mail or phone number"
 msgstr "Ex : E-mail ou numéro de téléphone"
+
+msgid "gluon-config-mode:contact-help"
+msgstr ""
+
+msgid "gluon-config-mode:contact-note"
+msgstr ""
```
NeoRaider: The `gluon-config-mode:...` (and similar) entries should be omitted. We don't have a scripted way to do that at the moment.

bobcanthelpyou: Removed these msgids

bobcanthelpyou:
Updated the `gluon-site.pot` via a modified `i18n-scan.pl`-file, too.
```
diff --git a/contrib/i18n-scan.pl b/contrib/i18n-scan.pl
index 29abefcc..fd05a4f0 100755
--- a/contrib/i18n-scan.pl
+++ b/contrib/i18n-scan.pl
@@ -117,7 +117,7 @@ if( open C, "| msgcat -" )
 
        foreach my $key ( sort keys %stringtable )
        {
-               if( length $key )
+               if( (length $key) && ($key =~ /^gluon-config-mode:/) )
                {
                        $key =~ s/"/\\"/g;
                        printf C "msgid \"%s\"\nmsgstr \"\"\n\n", $key;
```

With a similar change it is possible to ignore the site only translation strings, which might be a fast workaround for Jenkins or similar to check for required translation file updates. But there might be a more suitable solution for that.

rubo77:
>translated messages are sorted by msgid

I think this doesn't make any sense in our project. it just adds unnecessary confusing changes to the good history.

A better solution would be to ignore this rule in the i18n linter.

bobcanthelpyou:
Why am i doing these changes? I am using the mention tools from the docs (it is simply a perl script and gettext) to check if there is any translation for a string missing, a translation string has changed and so on. I thought it could be helpful for the Gluon developers and everyone else to easily check for those issues. At the moment using these tools, the difference between before running the tools and after running them is quite confusing cause of those many changes, and at least for me it is not easily possible to tell if there is something missing or not (it takes quite some time). But there are also no Gluon rules for the i18n files itself that could be followed, at the moment just the gettext defaults apply. If there are different rules that should be applied for the i18n files, the script and the gettext parameters should be adjusted, to make it more comfortable for everyone. And at least make it possible to apply the rules to have a consistent base for new adjustments and so on.

NeoRaider:
>     * How to handle the msgid `Enable PoE Passthrough` of gluon-web-network? Should it be replaced by `Enable %s` and `network.lua` modified? See https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-web-network/luasrc/lib/gluon/config-mode/model/admin/network.lua#L121
I have an idea how to handle this better, will push something later.

NeoRaider:
With bab4af01ebf0709f951b9d16c993a0c9926e9b9e, gluon-web-network is now compatible with i18n-scan.pl.

NeoRaider:
I'm not sure what to do with the `gluon-config-mode:` strings - the clunky prefix comes from a time before i18n packages were namespaced. Now that we have the namespaces, the defining feature of the site-provided strings is that *translate* is called on a different i18n object, but there is no obvious way to extract this information from the Lua source.

NeoRaider:
Could you rebase this and remove the changes to the `gluon-config-mode:` strings for now? Then we can get the rest merged.

bobcanthelpyou:
* Rebased and re-run `i18n-scan.pl` and `msgmerge` to be up2date with latest changes.
* Removed all changes in `docs/site-example/i18n` containing the `gluon-config-mode:` strings.

NeoRaider:
Is there a reason this is still marked as a draft?

bobcanthelpyou:
No, i am just unfamiliar with drafts.

PR #1787, opened 2019-07-20
RFC: features: include gluon-radvd by default

### LINKED ISSUE #1775: /etc/init.d/gluon-respondd ignores br-client unless dhcpv6 is successful
Version: 2018.2.2
Device TP-Link 3600

## Is:
* When dhcpv6 has not yet succeeded for br-client, respondd is not bound to br-client

## Should:
* respondd should bind to br-client if it is there

## Cause:
Before binding `/etc/init.d/gluon-respondd` queries for usable client interfaces using 
```bash
local clientdevs=$(for dev in $(echo "$ifdump" | jsonfilter -e "@.interface[@.interface='$(cat /lib/gluon/respondd/client.dev 2>/dev/null)' && @.up=true].device"); do echo " -i $dev -t $MAXDELAY";done;)`
```
Testing the query manually boils down to: 
```bash
echo "$(ubus call network.interface dump)" | jsonfilter -e "@.interface[@.interface='$(cat /lib/gluon/respondd/client.dev 2>/dev/null)' && @.up=true].device"
```
As of 2018.2.2 br-client is configured as
```
config interface 'client'
        option igmp_snooping '1'
        option type 'bridge'
        option auto '1'
        option ra_holdoff '30'
        option multicast_querier '1'
        option macaddr '64:66:b3:4f:fc:ac'
        list ifname 'local-port'
        list ifname 'bat0'
        option ipv6 '1'
        option keep_ra_dnslifetime '1'
        option sourcefilter '0'
        option peerdns '1'
        option robustness '9'
        option reqprefix 'no'
        option query_interval '2000'
        option query_response_interval '500'
        option proto 'dhcpv6'
```
This configuration causes `ubus call network.interface dump` to consider as down until dhcpv6 has suceeded - but due to `option auto '1'` the interface is up anyway.

## Impact:
* If no configuration IPv6 mechanism is present, respondd will never bind to br-client (that happened to me)
* If dhcp6 is slow (e.g. radvd intervals) respondd takes a respawn interval to bind to br-client, i.e. slow start.
* If the network is unstable (i.e. mesh) respondd might discard the interface temporarily.
* So far, multicast traffic appears to be concencerned, only. I guess, that yanic's unicast ping for unreachable nodes might compensate the impact for a while.

## Remarks:
* It feels strange to query for br-client and exclude it (possibly). I'm not aware of any configurations requiring explicit client devices scanning.

NeoRaider:
Fixes #1775

@christf Should this also be included for gluon-mesh-babel by default?

### DIFF
--- a/docs/multidomain-site-example/site.mk
+++ b/docs/multidomain-site-example/site.mk
@@ -12,7 +12,6 @@ GLUON_FEATURES := \
 	ebtables-limit-arp \
 	mesh-batman-adv-15 \
 	mesh-vpn-fastd \
-	radvd \
 	respondd \
 	status-page \
 	web-advanced \
--- a/docs/site-example/site.mk
+++ b/docs/site-example/site.mk
@@ -12,7 +12,6 @@ GLUON_FEATURES := \
 	ebtables-limit-arp \
 	mesh-batman-adv-15 \
 	mesh-vpn-fastd \
-	radvd \
 	respondd \
 	status-page \
 	web-advanced \
--- a/package/features
+++ b/package/features
@@ -24,7 +24,11 @@ packages 'web-advanced & autoupdater' \
 	'gluon-web-autoupdater'
 
 packages 'status-page & (mesh-batman-adv-14 | mesh-batman-adv-15)' \
-        'gluon-status-page-mesh-batman-adv'
+	'gluon-status-page-mesh-batman-adv'
 
 packages 'mesh-batman-adv-14 | mesh-batman-adv-15' \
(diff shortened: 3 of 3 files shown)


christf:
indeed it makes sense to include it by default. It is required for ipv6-network.

christf:
indeed it makes sense to include it by default. It is required for ipv6-network.

rotanid:
@NeoRaider was his response the only reason for the RFC status of this PR and therefore it can be merged now?

PR #1789, opened 2019-07-30
ar71xx-nand: add support for Aerohive HiveAP 121

herrbett:
  - [ ] webinterface
  - [x] tftp
  - [x] other: serial
    `aerohive-hiveap-121`
- wired network
- wifi (if applicable)
- led mapping
  - radio leds - device does not have radio leds 
    - [ ] should show activity
  - switchport leds
- outdoor devices only - no outdoor device

### DIFF
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac
@@ -47,6 +47,8 @@ elseif platform.match('ar71xx', 'generic', {'archer-c5', 'archer-c58-v1',
                                             'archer-c59-v1', 'archer-c60-v1',
                                             'archer-c7'}) then
   table.insert(try_files, 1, '/sys/class/net/eth1/address')
+elseif platform.match('ar71xx', 'nand', {'hiveap-121'}) then
+  table.insert(try_files, 1, '/sys/class/net/eth0/address')
 elseif platform.match('ipq40xx', nil, {'avm,fritzbox-4040',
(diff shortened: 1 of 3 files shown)


blocktrron:
The patchthis PR adds to OpenWrt is currently not upstream (yet): https://patchwork.ozlabs.org/patch/1139378/

rotanid:
Thanks for your contribution. Until the patch is merged upstream, we will not merge it.

Do you have a link where flashing via TFTP without serial is explained? in the original upstream commit from 2017, there's only a description to flash it using a serial console combined with tftp.

according to our policy ( https://github.com/freifunk-gluon/gluon/wiki/Device-Integration ) we don't integrate new devices which can't be flashed without opening the device.

herrbett:
You are right, there is currently only one method of flashing the device, wich is a combined Serial and TFTP method described in the OpenWRT commit.

The Serial is a Cisco Style RJ45 Jack, accessible on the outside of the case.

rotanid:
> The Serial is a Cisco Style RJ45 Jack, accessible on the outside of the case.

interesting. i guess we haven't experienced this corner case so far :D

blocktrron:
The required patch was picked to 18.06, so it can be dropped from this PR. Then it could be merged, after modules are bumped.


https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=b7e7d220e7eff6ac1d73c5b97dec448704b26735

mweinelt:
Merged in c37644e004f559f29aee2a428b736475b2723fe7.

Thanks.

PR #1791, opened 2019-08-07
Switch to OpenWrt 19.07

blocktrron:
This PR is only for reference, what steeps need to be performed to migrate Gluon onto an OpenWrt 19.07 base, containing support for the new ath79 target which will replace ar71xx.


targets/ath79-tiny
```diff
@@ -0,0 +1,22 @@
...
+
+defaults {
+	deprecated = true, -- 4/32
+}
+
+
+-- TP-Link
+
+device('tp-link-tl-mr3020-v1', 'tplink_tl-mr3020-v1')
+device('tp-link-tl-wr-740n-v1', 'tplink_tl-wr740n-v1')
```
mweinelt: If we add these models in ath79 targets, shouldn't we drop them in the corresponding ar71xx target at the same time?

blocktrron: This decision should be based on whether or not the migration was tested or not.

The current selection of devices in the new targets is only based on what I've had lying around.

mweinelt:
I tried building ar71xx-generic and that errored out on the Alfa Hornet UB:

```
cp: cannot stat 'openwrt/bin/targets/ar71xx/generic/openwrt-ar71xx-generic-hornet-ub-squashfs-sysupgrade.bin': No such file or directory
```

Turns out I can't even build OpenWrt for that device, because it's kernel partition is only ~1.6 MB and it's located *behind* the rootfs.

```
alfa_mtdlayout_8M=mtdparts=spi0.0:256k(u-boot)ro,64k(u-boot-env)ro,6144k(rootfs),1600k(kernel),64k(nvram),64k(art)ro,7744k@0x50000(firmware)
```

This layout is also used for the Tube2H, so that device is also a dead end.

targets/ath79-tiny
mweinelt: So they were tested?

mweinelt:
I successfully built all targets in this pr.

targets/ath79-tiny
rotanid: good question. i think we should only enable fully tested devices until we have decided on something else.

blocktrron: I agree on this topic. As this PR is only an RFC (and someone else might want to try out the ath79 target), I've added targets such as lantiq or ath79. If we come close to migrating over to 19.07, we should definitely not carry these over to the master branch without sufficient testing.

CodeFetch:
As 19.07 will possibly not officially support ath79, we should wait until it's clear that nothing critical changes anymore.

rotanid:
> As 19.07 will possibly not officially support ath79, we should wait until it's clear that nothing critical changes anymore.

i agree, that this PR should not be merged before OpenWrt 19.07 has been released.
also, we already discussed to be careful about the targets.

mweinelt:
I disagree, as OpenWrt 19.07 has already been branched and our master branch can already follow it, so it gets proper testing.

CodeFetch:
Okay, but one should be aware that the sysupgrade process might change in a way which requires forced updating and that the migration scripts are still broken (don't detect whether the router was updated from ath79 or ar71xx) which can lead to unexpected behavior and inaccessibility of nodes in a Gluon setup when updating nodes twice with an ath79 firmware. So there will be breaking changes even if 19.07 was branched already and 19.07 still has some known bugs.

An ath79 release was not planned for 19.07, yet - 19.07 is intended to be an ar71xx release only. This means one needs to treat 19.07 ath79 like building an image with OpenWrt Master branch. What I wanted to say is: "Wait for ath79 support in Gluon until it is officially supported by an OpenWrt release, which is not planned for 19.07, yet.".

rotanid:
> An ath79 release was not planned for 19.07, yet - 19.07 is intended to be an ar71xx release only. This means one needs to treat 19.07 ath79 like building an image with OpenWrt Master branch. What I wanted to say is: "Wait for ath79 support in Gluon until it is officially supported by an OpenWrt release, which is not planned for 19.07, yet.".

this was already agreed on in August, please read other reviews ;)

@mweinelt hm.... the biggest problem i have with such rebases is, that contributions still waiting for a maintainers review may need to be rebased again. it might be a good policy to keep us from not taking care of someones contribution by not rebasing before taking care of long-waiting PRs.

@blocktrron do you need information from this branch for the rebase, regarding batman-adv changes? https://github.com/freifunk-gluon/gluon/tree/wip/update-batadv

blocktrron:
@CodeFetch This is not correct. The current state is that OpenWrt 19.07 was set to be released with ath79 "beta" support, which means limited support for this target. See https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=ce41f468e1491134353a189b4cdf169e32786d16

ath79 still has some shortcomings in terms of hardware-bug workarounds and especially the USB stack, for which no major backports will be done.

I have no strong opinion regarding migration of ar71xx to ath79 in Gluon for 2019.2, however i think it is important to have ath79 in the upcoming Gluon release.

@rotanid Thanks for the pointer to this branch, would have saved me a lot of work :smile: . I will integrate this into this PR and report back if it works as expected :+1:

blocktrron:
@rotanid rebased to latest OpenWrt 19.07 commit. batadv migration is run-tested on at71xx.

rotanid:
ok.
you are adding and removing devices with this PR.
1) removing
if we already know that devices won't work with 19.07, we should hurry and mark them as deprecated for 2019.1 and mention it in the release notes? #1815 @mweinelt 
removed devices also need to be removed from the docs.
2) adding
should not be done together with the rebase but via individual PRs, like with adding a new target.

mweinelt:
I've just verified that the ar71xx-generic devices in question still don't build with the latest state of this pr, so I'll deprecate them soon.

blocktrron:
@rotanid 

Regarding the removal: We have no choice, as the "kernel-after-rootfs" partitioning is something we can't change. If we already know this, we should mention it in our release-notes.

The devices I've added should be done as a separate PR, i will remove these commits as soon as this PR has it's RFC removed. I'm using them for testing this PR ;)

rubo77:
The whole target ath79-tiny will not be supported any more already.

That means half of the German wide routers will not be supported.

Isn't there a work around, at least for the 841?

rotanid:
> The whole target ath79-tiny will not be supported any more already.
> 
> That means half of the German wide routers will not be supported.
> 
> Isn't there a work around, at least for the 841?

we never supported ath79-tiny so far, nor ath79. may we will in the future, but unlikely for ath79-tiny.

regarding ar71xx-tiny, we haven't said it will be gone, if it's possible with the little time available to all of us, it will stay of course.

but that's off-topic, hence i will mark our comments as off-topic now.

CodeFetch:
Please set the broken flag for ath79 then. I don't think it is a good idea to release an "OpenWrt beta" version as "Gluon stable". If you've followed the discussions about the migration process on the mailing list, you know that it is still not stable. Even updating from ath79 to the same ath79 version can lead to severe problems. And regarding the sysupgrade process from ar71xx to ath79 there will likely be a "ar71xx migration release" which has a modified sysupgrade to support the new JSON metadata format of ath79 images. There were no final decisions on this, yet. Thus there might be heavy changes. Beta means beta in this case.

CodeFetch:
@rubo77 Freifunk Hannover will still maintain ath79 Gluon patches for hardware-upgraded routers.

mweinelt:
> Please set the broken flag for ath79 then. I don't think it is a good idea to release an "OpenWrt beta" version as "Gluon stable". If you've followed the discussions about the migration process on the mailing list, you know that it is still not stable. Even updating from ath79 to the same ath79 version can lead to severe problems. And regarding the sysupgrade process from ar71xx to ath79 there will likely be a "ar71xx migration release" which has a modified sysupgrade to support the new JSON metadata format of ath79 images. There were no final decisions on this, yet. Thus there might be heavy changes. Beta means beta in this case.

You are basically repeating what you said in https://github.com/freifunk-gluon/gluon/pull/1791#issuecomment-531508165.

We will, as always, act with due diligence and I'm honestly not sure why you think we wont.

CodeFetch:
> You are basically repeating what you said in [#1791 (comment)](https://github.com/freifunk-gluon/gluon/pull/1791#issuecomment-531508165).

Because the current state of the PR ignores this fact and treats ath79 as stable/does not set the broken flag and I feel you ignore my comment. 

> We will, as always, act with due diligence and I'm honestly not sure why you think we wont.

I don't question that you wouldn't do testing or something. I just remark that even if ath79 works in the current state, it is not in the final state and breaking changes are very likely. Thus ath79 should definitely have the broken mark set and the documentation needs a hint on that issue/that ath79 can not be considered to be in a state where devices can be deployed without physical access/ without taking the risk that manual forced flashing without keeping the configuration is needed or manual review and altering of the configuration and that migration issues can even lead to unresponsive devices which can only be accessed via failsafe mode.

Furthermore I wonder that you did not have problems with the migration from ar71xx to ath79, because for Gluon we will need additional migration scripts as otherwise some devices would have swapped network configurations due to a different device initialization order, because custom Gluon configurations are not covered by OpenWrt's migration script.

CodeFetch:
@mweinelt About which devices with kernel after rootfs-format are we talking? Because we could introduce a workaround for that if they are important/commonly used by changing U-Boot's bootcmd (trying to boot at the beginning of the fimware partition and if that fails at the ordinary kernel partition offset). Unfortunately this was not done for some older devices.

adrianschmutzler:
@CodeFetch 
Concerning image metadata: I haven't yet dived deeper into Rafał Miłecki's changes of the metadata/sysupgrade. Before that, ath79 image for upgrade were typically accepted without problems as long as SUPPORTED_DEVICES was set correctly (which I suppose should be the case for about 80-95 % of the devices in ath79). I haven't checked after Rafał Miłecki's changes though.

Concerning migration:
This splits into different subtopic:
1. Wifi path migration
Well, this accounts for changed Wifi paths between ath79/ar71xx. I have no special knowledge about this script except that it exists:
https://github.com/openwrt/openwrt/blob/master/target/linux/ath79/base-files/etc/hotplug.d/ieee80211/00-wifi-migration

2. LED migration
This account for changed LED names (e.g. archer-c59:green:wan to tp-link:green:wan). While those changes technically are not limited to the upgrade from ar71xx (one could also account for changes in ath79), most entry are based on it:
https://github.com/openwrt/openwrt/blob/master/target/linux/ath79/base-files/etc/uci-defaults/04_led_migration
Since this script is not widely known, I suspect many device supports just ignore name changes during upgrade and there is a bunch of devices missing there. I've recently sent a patch to add some:
https://patchwork.ozlabs.org/patch/1163376/
So, if you encounter other missing devices, please do the same.
This particular patch is currently on hold due to subtopic number 3:
The biggest problem at the moment is the swap of eth0/eth1 on several devices. This is not trivial to address, as this occurs at several places and user may have based scripts on corresponding uci node names etc.. We are currently discussing how to address this, and this topic is also where the multiple update problem was mentioned:
https://www.mail-archive.com/openwrt-devel@lists.openwrt.org/msg48181.html (See whole thread ...)
So, while 1. and 2. are easy to address, problem 3 is currently only at the beginning of being addressed. It also does not look like there is too much interest in the topic.
I personally am also not too keen on pushing this forward, as in Freifunk Franken firmware we do not rely on the OpenWrt upgrade script, so it does not really matter to me. I just started the discussion as I had the impression that nobody cared.

Concerning ath79 in openwrt-19.07:
Most development of ath79 takes place in master and backports are scarce. I personally would ignore ath79 in openwrt-19.07 at all, unless you really need some of the devices only added there. Otherwise it will mostly mean duplicate work.
If you want Gluon to get ready for ath79, it is easier to start developing that based on master.
However, since this is blocktrron's PR, he will be able to give more substantial advice than me anyway :-)

mweinelt:
> > You are basically repeating what you said in [#1791 (comment)](https://github.com/freifunk-gluon/gluon/pull/1791#issuecomment-531508165).
> 
> Because the current state of the PR ignores this fact and treats ath79 as stable/does not set the broken flag and I feel you ignore my comment.
> 

blocktrron is using these targets/devices to test, please reread https://github.com/freifunk-gluon/gluon/pull/1791#issuecomment-533108385.

mweinelt:
> @mweinelt About which devices with kernel after rootfs-format are we talking? Because we could introduce a workaround for that if they are important/commonly used by changing U-Boot's bootcmd (trying to boot at the beginning of the fimware partition and if that fails at the ordinary kernel partition offset). Unfortunately this was not done for some older devices.

They were ALFA devices, see https://github.com/freifunk-gluon/gluon/pull/1791/commits/5e38aadfa9b36c55a5cc1a8bcbaa7037e446726d.

CodeFetch:
@mweinelt It's not clear from the code.

T-X:
Just an organizational question: Would it make sense to have a batman-adv update + multicast feature activation in a dedicated release before pulling the OpenWrt update?

The diff for batman-adv itself would be small. But since it requires several mechanisms to work in tandem, it might be nicer for potential debugging to not mangle it with an OpenWrt 19.07 update?

rotanid:
> in a dedicated release

definitely not, dedicated commit maybe?

blocktrron:
I've removed the added device support from this PR. So this should be somewhat acceptable now.

Whether or not we wait for the upstream tag still seems to be an open question. So i leave this on RFC for now.

The batadv-legacy compatibility should be verified by someone who still runs legacy batman. Ping @oszilloskop @skorpy2009

skorpy2009:
I thought `batadv-legacy` compatibility would be irrelevant since it will be dropped latest when gluon is based on OpenWRT 19.07

rotanid:
> I thought `batadv-legacy` compatibility would be irrelevant since it will be dropped latest when gluon is based on OpenWRT 19.07

i don't think it makes sense to handle everything we want to do or want to remove with one single Pull Request.

rotanid:
> I've removed the added device support from this PR. So this should be somewhat acceptable now.

thanks.

> Whether or not we wait for the upstream tag still seems to be an open question. So i leave this on RFC for now.

this, and @T-X suggestion to update batman-adv before rebasing to 19.07

oszilloskop:
@blocktrron 
Whether irrelevant or not, a firmware built with your branch 'owrt1907' works well in our legacy batman network.  See: [https://map.ffm.freifunk.net/#!/de/map/e4956e43124e](https://map.ffm.freifunk.net/#!/de/map/e4956e43124e) (I'll turn off the node sometime)

blocktrron:
Rebased onto  #1825 to allow testing of the batadv-legacy removal.

blocktrron:
- Rebased to latest openwrt-19.07 head
 - Removed `0008-uboot-fritz4040-add-fixes-for-bootloader-image-generation.patch` as this fix is incorporated upstream now

Also removing the RFC now.

rotanid:
>        removed devices also need to be removed from the docs.

blocktrron:
@rotanid fixed

mweinelt:
Build tested on all targets but mvebu, run-tested on x86-64, ramips-mt7621, ar71xx-generic, mpc85xx-generic.

LGTM.

We're pushing this into our testing branch (~30 nodes) tonight.

mweinelt:
We saw the following issue on an Unifi AC-Mesh. A DFS event was seen, the interface had to reset and the firmware apparently pooped. 

```
[20018.959719] ath: phy0: DFS: radar found on freq=5700: id=4, pri=245, count=10, count_false=0
[20018.963740] device client0 left promiscuous mode
[20018.978675] br-client: port 4(client0) entered disabled state
[20024.092894] ath10k_pci 0000:00:00.0: failed to synchronize setup for vdev 0: -145
[20024.100629] ath10k_pci 0000:00:00.0: failed to stop vdev 0: -145
[20029.133946] ath10k_pci 0000:00:00.0: failed to flush transmit queue (skip 0 ar-state 1): 0
[20030.560949] ath10k_pci 0000:00:00.0: pdev param 0 not supported by firmware
```

We are testing now with an upgraded ath10k firmware for Wave1 radios and it looks like it improved the situation.

package/gluon-mesh-batman-adv/luasrc/lib/gluon/upgrade/310-gluon-mesh-batman-adv-mesh
```diff
@@ -1,33 +1,23 @@
...
-local site = require 'gluon.site'
 local util = require 'gluon.util'
 
 local uci = require('simple-uci').cursor()
 
 
-local gw_mode = uci:get('batman-adv', 'bat0', 'gw_mode') or 'client'
-local gw_sel_class = site.mesh.batman_adv.gw_sel_class()
-local routing_algo = site.mesh.batman_adv.routing_algo()
+local gw_mode = uci:get('network', 'gluon_bat0', 'gw_mode') or uci:get('batman-adv', 'bat0', 'gw_mode') or uci:get('batman-adv-legacy', 'bat0', 'gw_mode') or 'client'
```
mweinelt: >     package/gluon-mesh-batman-adv/luasrc/lib/gluon/upgrade/310-gluon-mesh-batman-adv-mesh:9:121: (W631) line is too long (166 > 120)

```suggestion
local gw_mode = uci:get('network', 'gluon_bat0', 'gw_mode') or uci:get('batman-adv', 'bat0', 'gw_mode') or
  uci:get('batman-adv-legacy', 'bat0', 'gw_mode') or 'client'
```

mweinelt:
```
diff --git a/package/gluon-core/luasrc/lib/gluon/upgrade/500-opkg b/package/gluon-core/luasrc/lib/gluon/upgrade/500-opkg
index eecedaf3..bf0d15dc 100755
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/500-opkg
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/500-opkg
@@ -14,7 +14,7 @@ local util = require 'gluon.util'
 local subst = {}
 
 
-local f = io.popen('. /etc/os-release; echo "$ID"; echo "$VERSION_ID"; echo "$LEDE_BOARD"; echo "$LEDE_ARCH"')
+local f = io.popen('. /etc/os-release; echo "$ID"; echo "$VERSION_ID"; echo "$OPENWRT_BOARD"; echo "$OPENWRT_ARCH"')
 subst['%%d'] = f:read()
 subst['%%v'] = f:read():gsub('-snapshot', '')
 subst['%%S'] = f:read()
```

patches/openwrt/0008-ath10k-firmware-bump-firmware-for-wave-1-chips.patch
```diff
@@ -0,0 +1,38 @@
+From: David Bauer <mail@david-bauer.net>
+Date: Thu, 26 Sep 2019 18:23:12 +0200
+Subject: ath10k-firmware: bump firmware for wave 1 chips
+
+When receiving DFS events on a Ubiquiti UniFi AC Mesh running OpenWrt
+using QCA ath10k & ath10k-firmware, we saw the radio feezing and a
```
skorpy2009:
```suggestion
using QCA ath10k & ath10k-firmware, we saw the radio freezing and a
```

patches/openwrt/0008-ath10k-firmware-bump-firmware-for-wave-1-chips.patch
```diff
@@ -0,0 +1,38 @@
...
+constant high load due to softirqs.
+
+This issue is not present when using the latest  QCA firmware for the
```
skorpy2009:
```suggestion
This issue is not present when using the latest QCA firmware for the
```

package/gluon-mesh-batman-adv/files/lib/netifd/proto/gluon_bat0.sh
```diff
@@ -36,15 +43,44 @@ proto_gluon_bat0_renew() {
...
 	local config="$1"
 
+	local routing_algo=$(lookup_site 'mesh.batman_adv.routing_algo' 'BATMAN_IV')
+
+	modprobe batman-adv
+
+	local gw_mode
+	json_get_vars gw_mode
+
+	batctl routing_algo "$routing_algo"  2>/dev/null
```
NeoRaider: The `2>/dev/null` can be removed, as we don't have batadv-legacy anymore.

package/gluon-mesh-batman-adv/files/lib/netifd/proto/gluon_bat0.sh
```diff
@@ -36,15 +43,44 @@ proto_gluon_bat0_renew() {
...
 	local primary0_mac="$(lua -e 'print(require("gluon.util").generate_mac(3))')"
 
 	ip link add primary0 type dummy
 	echo 1 > /proc/sys/net/ipv6/conf/primary0/disable_ipv6
 	ip link set primary0 address "$primary0_mac" mtu 1532 up
 
-	local routing_algo="$(lua -e 'print(require("gluon.site").mesh.batman_adv.routing_algo())')"
-	(echo "$routing_algo" >/sys/module/batman_adv/parameters/routing_algo) 2>/dev/null
-
 	echo bat0 > /sys/class/net/primary0/batman_adv/mesh_iface
```
NeoRaider: This line should be changed to batctl as well.

package/gluon-mesh-batman-adv/files/lib/netifd/proto/gluon_bat0.sh
```diff
@@ -56,8 +92,8 @@ proto_gluon_bat0_setup() {
 proto_gluon_bat0_teardown() {
 	local config="$1"
 
-	ip link del bat0
 	ip link del primary0
+	batctl interface destroy
```
NeoRaider: Why is the teardown order inverted here?

patches/openwrt/0008-ath10k-firmware-bump-firmware-for-wave-1-chips.patch
```diff
@@ -0,0 +1,38 @@
+From: David Bauer <mail@david-bauer.net>
+Date: Thu, 26 Sep 2019 18:23:12 +0200
+Subject: ath10k-firmware: bump firmware for wave 1 chips
```
NeoRaider: This patch should be pushed upstream.

patches/packages/routing/0002-batman-adv-disable-hardif-migration.patch
```diff
@@ -0,0 +1,24 @@
...
+index 258b7bd6a58892fd02e257c4679dc46dbcf7adf7..34da08f8ba173556a7f7b3fd458ffc12a09665fe 100755
+--- a/batman-adv/files/etc/uci-defaults/99-migrate-batadv_hardif
++++ b/batman-adv/files/etc/uci-defaults/99-migrate-batadv_hardif
+@@ -1,5 +1,10 @@
+ #!/bin/sh
+ 
++# Don't perform this migration. Gluon will take care of mesh configuration
++# and we need to preserve the gw_mode.
++
++exit 0
```
NeoRaider: Is this patch really necessary?  Does `99-migrate-batadv_hardif` not migrate `gw_mode` correctly?

package/gluon-mesh-batman-adv/files/lib/netifd/proto/gluon_bat0.sh
```diff
@@ -36,15 +43,44 @@ proto_gluon_bat0_renew() {
 proto_gluon_bat0_setup() {
 	local config="$1"
 
+	local routing_algo=$(lookup_site 'mesh.batman_adv.routing_algo' 'BATMAN_IV')
+
+	modprobe batman-adv
```
NeoRaider: modprobe is not needed anymore

patches/packages/routing/0002-batman-adv-disable-hardif-migration.patch
blocktrron: I have to test this. My intention was to not do this migration, as we do not rely on the interface bringup coming from the batman-adv package. Also, this would have split the configuration file from Batman v14.

However, as we do not have Batman v14 anymore (and an additional downstream patch is not nice), i will try to implement a different approach.

patches/openwrt/0008-ath10k-firmware-bump-firmware-for-wave-1-chips.patch
blocktrron: this commit was going into the wrong branch, i will remove it. We are currently bisecting our ath10k issues with newer firmware.

package/gluon-mesh-batman-adv/files/lib/netifd/proto/gluon_bat0.sh
blocktrron: This was probably done by mistake when rebasing. Will fix that.

patches/packages/routing/0002-batman-adv-disable-hardif-migration.patch
blocktrron: I think / hope I've found a better solution to this. Will have to verify it working on a running device.

blocktrron:
Okay, I've rebased the patches. I also think we mitigated the ath10k memleak we've seen on our AC Mesh.

However, as this issue seems to be within the firmware, we had to use the most recent ath10k-firmware, which is currently not available in linux-firmware.

I will try to kick off the upstreaming procedure, however we might need to carry these two patches downstream until the next OpenWrt release uses a newer firmware (or a linux-firmware bump finds it's way into 19.07)

patches/openwrt/0008-ath10k-firmware-retrieve-wave-1-firmware-from-kvalo.patch
```diff
@@ -0,0 +1,53 @@
+From: David Bauer <mail@david-bauer.net>
+Date: Sat, 5 Oct 2019 02:12:56 +0200
+Subject: ath10k-firmware: retrieve wave 1 firmware from kvalo
+
+This commit changes the source of the Wave 1 ath10k-firmware
+from linux-firmware to Kall Valos ath10k-firmware repository.
+
+This is neaessary as the firmware selected in linux-firmware produces
```
rotanid: "neaessary" - typo

targets/ar71xx-generic
```diff
@@ -3,9 +3,10 @@ config 'CONFIG_TARGET_SQUASHFS_BLOCK_SIZE=64'
 
 local ATH10K_PACKAGES = {}
 local ATH10K_PACKAGES_QCA9887 = {}
-if env.GLUON_WLAN_MESH == 'ibss' then
-	ATH10K_PACKAGES = {'-kmod-ath10k', 'kmod-ath10k-ct', '-ath10k-firmware-qca988x', 'ath10k-firmware-qca988x-ct'}
-	ATH10K_PACKAGES_QCA9887 = {'-kmod-ath10k', 'kmod-ath10k-ct', '-ath10k-firmware-qca9887', 'ath10k-firmware-qca9887-ct'}
+if env.GLUON_WLAN_MESH == '11s' then
+	ATH10K_PACKAGES = {'kmod-ath10k', '-kmod-ath10k-ct', 'ath10k-firmware-qca988x', '-ath10k-firmware-qca988x-ct'}
+	ATH10K_PACKAGES_QCA9886 = {'kmod-ath10k', '-kmod-ath10k-ct', 'ath10k-firmware-qca9888', '-ath10k-firmware-qca9888-ct'}
```
rotanid: is it on pupose to have qca9888 firmware in a variable called _QCA9886 ?

mweinelt:
```
Checking targets/ar71xx-generic                   5 warnings


    targets/ar71xx-generic:8:2: (W111) setting non-standard global variable ATH10K_PACKAGES_QCA9886

    targets/ar71xx-generic:303:13: (W113) accessing undefined variable ATH10K_PACKAGES_QCA9886

    targets/ar71xx-generic:308:13: (W113) accessing undefined variable ATH10K_PACKAGES_QCA9886

    targets/ar71xx-generic:313:13: (W113) accessing undefined variable ATH10K_PACKAGES_QCA9886

    targets/ar71xx-generic:318:13: (W113) accessing undefined variable ATH10K_PACKAGES_QCA9886
```

targets/ar71xx-generic
blocktrron: Yes, as QCA9888 and QCA9886 share the same firmware and i haven't seen (yet) a device using a QCA9888.

However, i think we should add definitions for every chip (and include "QCA9880" to the "ATH10k_PACKAGES" variable name)

blocktrron:
Oh and for everyone interested in the reason for the memleak: The firmware suddenly stops sending TX_DONE messages, leading to an overflowing TX queue in the Kernel. In this case you can see a stalling HTT Cookies pending counter while the other stats still indicate transmits happening.

It is a bit strange, as it also happens with the same firmware we used in 2018.2 / 2019.1. But as it's closed it's left to be a guessing game.

targets/ar71xx-generic
rotanid: ok, but with the big "K" to keep it consistent

mweinelt:
We currently suspect an issue with the cable powering our AC-Mesh, because using either one radio works just fine, but two radios at once does not.

CodeFetch:
@mweinelt A few months ago some of our AC Mesh started to become unstable in a similar way. It seems the buck converter is at the end of it's life time. So if your's is one of the first models, it could be possible that it has the same flaw. If you've used a passive/24V PoE before, you can get it running again with a proper 48V PoE adapter.

patches/openwrt/0009-ath10k-firmware-update-wave-1-firmware-to-10.2.4-1.0-00047.patch
```diff
@@ -0,0 +1,35 @@
+From: David Bauer <mail@david-bauer.net>
+Date: Sat, 5 Oct 2019 12:27:44 +0200
+Subject: ath10k-firmware: update wave 1 firmware to 10.2.4-1.0-00047
+
+This fixes frequent crashes observed on a UniFi AC Mesh using OpenWrt
+master and 19.07. 18.06 seems not affected from our testing.
+
+Signed-off-by: David Bauer <mail@david-bauer.net>
```
NeoRaider: Why is this patch still here, and not in upstream 19.07?

NeoRaider: Actually, both of the ath10k firmware patches.

blocktrron: I was still a bit unsure about upstreaming this patch, as upstream does not use qca-ath10k anyway. However (as it does not hurt), i will send them upstream too shortly. 

If it does not make it upstream, we have to carry them downstream anyway.

targets/ar71xx-generic
rotanid: after a rebase, this could be resolved

patches/openwrt/0009-ath10k-firmware-update-wave-1-firmware-to-10.2.4-1.0-00047.patch
rotanid: maybe that's the only thing keeping us from a merge?

blocktrron: I've started a Mailinglist Thread here. As no response came on this, i will submit the patches later today.

http://lists.infradead.org/pipermail/openwrt-devel/2019-October/019380.html

mweinelt: https://github.com/openwrt/openwrt/commit/98da3c8cdb6a352a5c74cbbbc948dcc91954fc99
https://github.com/openwrt/openwrt/commit/0e7113e6ece03c879bf99dff097dd868e3453f27

mweinelt:
The lua linter is unhappy about the unused `ATH10K_PACKAGES_QCA9886` definition. We should likely remove it until we have the first device that needs it.

NeoRaider:
Except for @mweinelt's last comment, LGTM

blocktrron:
@mweinelt I hope the lua linter is a happy lua linter now.

mweinelt:
Did another test build and all targets look good!

rubo77:
Can you please post the final consequences of this merge somewhere and link here?

blocktrron:
@rubo77 Which consequences are you referencing?

rubo77:
for example it sounded, like when we switch, the tiny-target will not be included in the master branch any more. Maybe some packages won't work any more? some other important things for collaborators to know?

blocktrron:
@rubo77 There is no hard impact on Gluon (yet), as OpenWrt 19.07 still includes ar71xx-tiny.

PR #1792, opened 2019-08-08
ar71xx-generic: Fix packages for GL.iNet AR750

CodeFetch:
The profile for GL.iNet AR750 currently selects ATH10K_PACKAGES, but this device actually has a QCA9887 which needs another driver.
This commit fixes the issue by setting ATH10K_PACKAGES_QCA9887 instead.


rotanid:
maybe this needs to be backported @CodeFetch , what do you think?

CodeFetch:
@rotanid Yes, WLAN-mesh-only routers of this type get lost forever without intervention.

mweinelt:
Backported as 92ecb316a5e1215dbdc57ced470dc8af69228466

Issue #1790, opened 2019-07-31
Traffic shaping using simple-tc does not work for downloads


SmithChart:
### Bug report

**What is the problem?**
A user in our community reported that taffic-shaping using the gluon-buildin simple-tc does not work for download-speed (but only for uplad speed). The community-member is using a WR941ND.

We were able to reproduce this on our desk using a newly flashed TL WR 841 v10:
The configuration was set using luci during the setup wizard. After reboot the config was:

```
root@freifunk60e327cf26e0-chrissi-test:~# uci show simple-tc
simple-tc.example=interface
simple-tc.example.enabled='0'
simple-tc.example.ifname='eth0'
simple-tc.example.limit_egress='1000'
simple-tc.example.limit_ingress='5000'
simple-tc.mesh_vpn=interface
simple-tc.mesh_vpn.enabled='1'
simple-tc.mesh_vpn.ifname='mesh-vpn'
simple-tc.mesh_vpn.limit_egress='500'
simple-tc.mesh_vpn.limit_ingress='500'
```

We tested the speed using iperf3 in both directions (but not at the same time). Results were:
Download: 10 Mbits/s, Upload: 500 Kbits/s

We removed the limits using:
`root@freifunk60e327cf26e0-chrissi-test:~# simple-tc mesh-vpn - -`
and the upload limit was removed as expected.

But: We were not able to set the download-limit using simple-tc:
```
root@freifunk60e327cf26e0-chrissi-test:~# simple-tc mesh-vpn - 500
root@freifunk60e327cf26e0-chrissi-test:~# simple-tc mesh-vpn 500 -
uradvd: error: netlink: No such file or directory
```
(Ignore the uradvd-name there. Jan Luebbe has already fixed that: https://github.com/freifunk-gluon/packages/commit/17e253563241eecd6e474c4ffd78adcd28744337 )

This restored the limits as they have been initially.

We dug into the gluon config and found out that the `KERNEL_NET_ACT_POLICE` is only marked as module for our kernel but the module is not installed onto our target. `simple-tc` depends (https://github.com/freifunk-gluon/packages/blob/master/net/simple-tc/files/etc/modules.d/30-simple-tc ) on that module but fails to load it.

We figure this happens because the package `simple-tc` has no dependency on this.

To prove our theory we added this dependency to the gluon core as follows:
```
$ git diff
diff --git a/package/gluon-core/Config.in b/package/gluon-core/Config.in
index caf4f4c6..1da7f294 100644
[...]

PR #1793, opened 2019-08-11
gluon-core: Add dependency for KERNEL_NET_ACT_POLICE

SmithChart:
This is a fix for the broken ingress traffic-shaping in gluon v2018.2.2
and possibly earlier.

For ingress traffic shaping the kernel option NET_ACT_POLICE is needed.
Before this patch there was no dependency to this. Neither in
gluon_core, gluon-mesh-vpn-core nor in the package.

This patch adds this dependency.

See: https://github.com/freifunk-gluon/gluon/issues/1790


mweinelt:
If this is required only for simple-tc we should try its inclusion to cases when the package is actually included into the firmware.

I'm not sure how to actually accomplish this, so I'll defer to @NeoRaider.

NeoRaider:
Thanks for looking into this. I've made a few adjustments to your patch and merged it in dc8149abc4c377d8718f1d8295472ba98c2208d4 and aa2c0dc3f9819de9deb902a0c36f098f49619c2a.

PR #1794, opened 2019-08-12
scripts/update.sh: add verbose error message when commit cannot be found

christf:
This displays the branch, repo and commitID that was attempted to update to but
failed. Users then can compare this output to the configuration they meant to
activate in modules and hopefully better find their typos.

### DIFF
--- a/scripts/update.sh
+++ b/scripts/update.sh
@@ -20,6 +20,9 @@ for module in $GLUON_MODULES; do
 
 	if ! git branch -f base "$commit" 2>/dev/null; then
 		git fetch "$repo" "$branch"
-		git branch -f base "$commit"
+		git branch -f base "$commit" || {
+		  echo "unable to find commit \"$commit\" on branch \"$branch\" in repo \"$repo\"." >&2
+		  exit 1
+		}
 	fi
 done


mweinelt:
Not sure what additional information this conveys, is the already existing output not enough?

```
--- Updating module 'openwrt' ---
Reinitialized existing Git repository in /gluon/openwrt/.git/
From https://git.openwrt.org/openwrt/openwrt
 * branch                  openwrt-18.06 -> FETCH_HEAD
fatal: Not a valid branch point: 'b7e7d220e7eff6ac1d73c5b97dec448704b26736'.
unable to find commit "b7e7d220e7eff6ac1d73c5b97dec448704b26736" on branch "openwrt-18.06" in repo "https://git.openwrt.org/openwrt/openwrt.git".
make: *** [Makefile:54: update] Error 1
```

christf:
I find it much easier to read. It is a proper error mesage instead of a status output of something that may or may not be related.

bobcanthelpyou:
Is it possible to remove/suppress the redundant old information above the new message somehow, so nobody has to read it twice?

rotanid:
> Is it possible to remove/suppress the redundant old information above the new message somehow, so nobody has to read it twice?

not specific for this error and we don't want to suppress any error, do we?

mweinelt:
I'm quite happy with the git style error message, it has a visual pattern that I recognize, hence I'd very much like to keep it.

PR #1797, opened 2019-08-20
docs: move supported devices to own page

belzebub40k:
Every now and then I forget where to find the list of supported devices at it takes me some time to find it under the table of contents. I think moving it to a dedicated page would be more user friendly.


rotanid:
thanks

PR #1798, opened 2019-08-23
gluon-core: implement generic generate_ip_from_prefix(), use it in mac_to_ip()

christf:
mac_to_ip() implements a special case, generating an ip6 address from, a
hwaddress. For wireguard we will use specially crafted addresses that
are not based on a mac address.

This commit allows to re-use the code by implementing
generate_ip_from_prefix() in a generic way, reusing it in mac_to_ip()
without breaking the existing interface.


NeoRaider:
We could reuse the existing `mac_to_ip` function name, adding two "optional" arguments:
```lua
local m4 = firstbyte or 0xff
local m5 = secondbyte or 0xfe
```

christf:
done.

PR #1800, opened 2019-08-24
gluon-config-mesh-vpn: use default value for mesh-vpn enabled from site

christf:
auto-enable mesh-vpn checkbox in config manager when enabled in site.conf

### DIFF
--- a/package/gluon-config-mode-mesh-vpn/luasrc/lib/gluon/config-mode/wizard/0300-mesh-vpn.lua
+++ b/package/gluon-config-mode-mesh-vpn/luasrc/lib/gluon/config-mode/wizard/0300-mesh-vpn.lua
@@ -21,10 +21,10 @@ return function(form, uci)
 
 	local s = form:section(Section, nil, msg)
 
-	local o
+	local site = require 'gluon.site'
 
 	local meshvpn = s:option(Flag, "meshvpn", pkg_i18n.translate("Use internet connection (mesh VPN)"))
-	meshvpn.default = uci:get_bool("gluon", "mesh_vpn", "enabled")
+	meshvpn.default = uci:get_bool("gluon", "mesh_vpn", "enabled") or site.mesh_vpn.enabled(false)
 	function meshvpn:write(data)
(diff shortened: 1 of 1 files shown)


rotanid:
hm, i don't get it?
we have it enabled in site.conf and our user's see a already checked checkbox in config-mode.
what is left that your PR changes?

christf:
for me the checkbox in the initial configuration was not enabled. This PR fixes that.

rotanid:
hm, this is interesting.... @mweinelt ?

NeoRaider:
This is definitely not the correct solution. We need to find out how the node can get into a state where gluon.mesh_vpn.enabled is not set correctly.

christf:
I installed without taking

rotanid:
> I installed a factory image not keeping settings.

via the config mode? via sysupgrade? i never used the factory image after there's no vendor image anymore... which device? maybe i can try to reproduce.

edeso:
that sounds familiar. have a WR841N/ND v9 that i tftpd a v2018.2.2 factory image 
https://images.ffkbu.de/BonnV2/stable/FastD/factory/gluon-ffkbu-V2-v2018.2.2-FastD-tp-link-tl-wr841n-nd-v9.bin
to (aka tftp recovery) ended up with a wifi meshing node but no vpn upstream.

a little check turned out that fastd was disabled (fastd.mesh_vpn.enabled='0') on the new node. enabling it worked fine.

 i visited the config mode to setup the node, so i might have disabled meshing there by accident, but i'm pretty sure i didn't because i never did that before.

NeoRaider:
I can't reproduce the issue with Gluon master on x86-64.

christf:
I saw this with factory installs on ar71xx and with factory images for x86-generic.

NeoRaider:
@christf Can you link such an x86 image?

rubo77:
And what is the site config, you used for your build?

christf:
hm. after fixing the boot hang issue with wireguard by moving the key generation into the init script I was not able to reproduce

NeoRaider:
Closing, please open an issue if it reoccurs.

PR #1803, opened 2019-08-31
Add lantiq devices as BROKEN

heini66:
Patch 1 of 2
Addd initial lantiq support

### DIFF
--- a/targets/targets.mk
+++ b/targets/targets.mk
@@ -26,5 +26,7 @@ ifneq ($(BROKEN),)
 $(eval $(call GluonTarget,ar71xx,mikrotik)) # BROKEN: no sysupgrade support
 $(eval $(call GluonTarget,brcm2708,bcm2710)) # BROKEN: Untested
 $(eval $(call GluonTarget,ipq806x)) # BROKEN: See target file for details
+$(eval $(call GluonTarget,lantiq,xrx200)) # BROKEN: No dedicated wan IF
+$(eval $(call GluonTarget,lantiq,xway)) # BROKEN: No dedicated wan IF
 $(eval $(call GluonTarget,mvebu,cortexa9)) # BROKEN: No AP+IBSS or 11s support
 endif


heini66:
flash sysupgrade via bootloader -> mtd1

rotanid:
you supply way too less information for two new targets and new devices.
please
a) check https://github.com/freifunk-gluon/gluon/wiki/Device-Integration
b) follow https://github.com/freifunk-gluon/gluon/wiki/Device-Integration-checklist
c) don't do PRs like this one here, which does not do anything with any effect on Gluon. This PR should have been split in half into your other two PRs, so the target and it's first device are added together. please do this together with b) , after you have read a). afterwards, we can close this PR here.

thank you for your contribution and for understanding my reasoning

heini66:
thanks for your comments. i'll try to get my next pr a little bit more ...
might you vistit up in https://forum.freifunk.net/t/alte-fritzbox-als-freifunk-repeater/21154 to clearify some things? ;)

blocktrron:
Regarding the State of AVM lantiq boards in 18.06:

1. Their naming is ... interesting. Describing boards with iterating marketing titles is not the kind of thing we want / need. However, conflicts have not arised (and the naming is fixed in https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=d4ba7bab25c84478ce3aa5ebd80d24b5a5624e7b )

2. The LED mappings were very inconsistent, as the Power LED had duplicate mappings for indicating running as well as DSL state on some devices, on others DSL state was mapped on the Info LED. For us, this means no indication of setup mode on these boards. However, i've tried to clean up this mess here: https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=3032bf7f89e8ad5b69ab2f031287f475e5d2b829

heini66:
https://www.ip-phone-forum.de/threads/versionen-der-7360.265609/
this might be helpfull for the 7360 versions ...

Am Di., 3. Sept. 2019 um 21:59 Uhr schrieb David Bauer <
notifications@github.com>:

mweinelt:
Let's postpone this until after v2019.1, so we can base this on OpenWrt v19.0[789012].

mweinelt:
@heini66 Thank you for you interest in getting more devices upstream. I'm closing these pulls (#1803 #1804 #1805) because the form given isn't actionable to the Gluon maintainers.

- Please create *one* pull request *per device* you want to add support for. 
- If a new target has to be added, do it in the commit where you add the first device to the target (cmp. e3952841567bb5afd4575336fb95be222f9357c9)
- For each device we need a test protocol (copy from https://github.com/freifunk-gluon/gluon/wiki/Device-Integration-checklist)

Also please wait until we've merged #1791 and base your changes on that. Thank you!

PR #1807, opened 2019-09-03
ramips-rt305x: remove opkg and USB packages from 4M flash boards

rotanid:
reduce size by removing dispensable packages,
to fix builds of this target in most cases.

result of discussion in #1782 

this PR has been build-tested

### DIFF
--- a/targets/ramips-rt305x
+++ b/targets/ramips-rt305x
@@ -1,9 +1,20 @@
 config '# CONFIG_KERNEL_KALLSYMS is not set'
+config 'CONFIG_GLUON_SPECIALIZE_KERNEL=y'
+
+no_opkg()
+
+local STRIP_USB_PACKAGES = {
+	'-kmod-usb-core',
+	'-kmod-usb-dwc2',
+	'-kmod-usb-ledtrig-usbport',
+}
+
 
 -- A5
 
 device('a5-v11', 'a5-v11', {
 	deprecated = true, -- 4/32
+	packages = STRIP_USB_PACKAGES,
 })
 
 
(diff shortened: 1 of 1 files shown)


targets/ramips-rt305x
```diff
@@ -1,4 +1,14 @@
 config '# CONFIG_KERNEL_KALLSYMS is not set'
+config 'CONFIG_GLUON_SPECIALIZE_KERNEL=y'
+
+no_opkg()
+
+packages {
+	'-kmod-usb-core',
+	'-kmod-usb2',
```
mweinelt: Are you sure rt305x has kmod-usb2?

NeoRaider: @mweinelt Many devices add kmod-usb2 in their DEVICE_PACKAGES.

rotanid:
> If we apply the same limitations as ar71xx-tiny I'd also want to see this target marked as deprecated.

the really problematic devices are deprecated for a while now in master branch.
technically not all devices in this target need the limitations, but afaik you can only apply them per-target atm.
therefore, the affected devices could be removed for the next but one release and the restrictions for the target removed?

targets/ramips-rt305x
rotanid: @mweinelt do you expect an answer from my side, too?
do you want me to "add" kmod-usb-dwc2 ?

rotanid: @mweinelt you were right, not kmod-usb2 has to be removed but kmod-usb-dwc2. i did this now and also changed the PR to only affect the 4mb devices.

PR #1811, opened 2019-09-07
drop unused node_client_prefix6

genofire:
@christf @rotanid We do not use anywhere `node_client_prefix6` why we need it ?


Adorfer:
I understood that this would be used for a future(!) "babel nodes having local dhcp-leasepools", like it's done for dhcpv4 already.

christf:
I really do not think this should be removed. Instead it should be used by the gluon-l3roamd initscript to configure the client_node_prefix (flag: -P).

rotanid:
> I really do not think this should be removed. Instead it should be used by the gluon-l3roamd initscript to configure the client_node_prefix (flag: -P).

but why has it been added without having any use (yet) ?
in its current state, it makes sense to me to remove it (not knowing much about babel and l3roamd...)

genofire:
@adorfer babel whould usw ddhcpd which use the prefix4 and the 464xlat use clat_range - which all work (would work) correct

Adorfer:
> @Adorfer babel whould usw ddhcpd which use the prefix4 and the 464xlat

that sounds more like legacy IP, but if that is used to get the prefix for local dhcp6 too, o.k.

christf:
please see patches in latest christf_next branch. Now it is used. Using fec0 nowadays is not a good idea anyways. This really should be configurable.

rotanid:
> please see patches in latest christf_next branch. Now it is used. Using fec0 nowadays is not a good idea anyways. This really should be configurable.

i don't doubt you have a plan for it, but at the moment it is not used / useless, right? so removing it before making a release should be better, one can always add it again when it is finally used somewhere.

christf:
it is a bug that it is not used. let's include the bugfix in the release instead of manifesting the crippled functionality.

genofire:
@rotanid why you close it? the PR #1813 reimplement it on the correct package. But we still have to remove it in this one

rotanid:
@christf can you comment on this?
also, can you use the docs which are removed here for your #1813 docs ?

PR #1813, opened 2019-09-09
redistribute routes toward internet, utilize missing values from site

christf:
this fixes two issues:
* redistribute the routes towards the internet
* correctly check for and use values from site.conf: prefix4 prefix6 and client_node_prefix6


package/gluon-l3roamd/check_site.lua
```diff
@@ -0,0 +1,3 @@
+need_string_match(in_domain({'prefix6'}), '^[%x:]+/64$', true)
+need_string_match(in_domain({'node_client_prefix6'}), '^[%x:]+/64$', false)
+need_string_match(in_domain({'prefix4'}), '^%d+.%d+.%d+.%d+$', false)
```
genofire: missing subnet here https://gluon.readthedocs.io/en/v2018.2.x/user/site.html

christf: indeed.

PR #1817, opened 2019-09-15
docs: remove Netgear WNDR3700 v5

bobcanthelpyou:
Reason:
* the device is part of the target `ramips-mt7621` (not `ar71xx-generic`)
* the device is marked as broken https://github.com/freifunk-gluon/gluon/blob/master/targets/ramips-mt7621#L18

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -55,7 +55,7 @@ ar71xx-generic
 
 * Netgear
 
-  - WNDR3700 (v1, v2, v5)
+  - WNDR3700 (v1, v2)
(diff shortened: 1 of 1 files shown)


bobcanthelpyou:
It might be removed from the v2018.2.x supported devices list, too.

mweinelt:
> It might be removed from the v2018.2.x supported devices list, too.

Yup, please add to this pull request.

mweinelt:
Fixed in v2018.2.x in 2d829a86028ef7a52a8fb7e9214f00e8744f3d83.

PR #1818, opened 2019-09-15
docs: mention ff05::2:1001 on the node monitoring page

bobcanthelpyou:


rubo77:
Shouldn't this be better?

It is a daemon listening on UDP port 1001 on the multicast address ff02::2:1001 on the mesh interface and on the multicast address ff05::2:1001 on the br-client interface.

Issue #1692, opened 2019-04-04
[RFC] Use zram-swap to save memory on 32 MiB devices


CodeFetch:
Has anyone tested it, yet? This might mitigate #1243 further.

bobcanthelpyou:
At some places zram is not suggested e.g. on devices with 4 MB flash, only.
```
Do not use zram-swap for 4MB flash devices as it increases the amount of firmware space used. It is listed here as it is helpful on machines with very little RAM memory. 
```
https://openwrt.org/docs/guide-user/additional-software/saving_space#excluding_packages

But i didn't test zram on Openwrt based devices, yet.

txt-file:
According to https://openwrt.org/packages/table/start?dataflt%5BName_pkg-dependencies*%7E%5D=zram we talk about 15 KB of flash space.

NeoRaider:
Including missing dependencies (kmod-lib-lz4, swap-utils, block-mount, libssp, possibly more I overlooked), I count about 90KB for the default zram-swap solution. It might be possible to work with less, but I haven't looked into it in detail.

CodeFetch:
Unfortunately I don't have any 8/32 MiB device deployed or I would test it.

@skorpy2009 As you're unable to phrase your opinion, I assume you're just a negative person.
While I do bootloader development and write an article for upgrading 4/32 MiB devices to 16/64 MiB for the long term, you're just trolling. I've already upgraded more than 50 WR841Ns and they work better than ever without memory pressure. I'm working on Layer 2 WireGuard which gives 40 MBit/s throughput with these devices. So if you are thinking of trashing your routers, give them to me, please and you're doing good for Freifunk and the environment.

christf:
I have deployed zram-swap on a few devices as well and did see positive effects on runtime behavior.

Given it needs quite a bit of space - I am not sure it should be on by default. How do you feel about building a package such that there is a choice to include it or not?

CodeFetch:
@christf Can't we just add `packages zram-swap` to the device definition of 8/32 MiB devices?

Adorfer:
@CodeFetch  Can this be done via site.mk for ar7xx-tiny-Target?

CodeFetch:
@Adorfer Yes, I think so, but I haven't tested it:
```
ifeq ($(GLUON_TARGET),ar71xx-tiny)
GLUON_SITE_PACKAGES += zram-swap
endif
```
[...]

PR #1819, opened 2019-09-15
ar71xx-generic: add zram-swap to boards with 8M flash but only 32M memory

rotanid:
see #1692 for tests and discussion.

build and run tested on a TL-WR1043ND v1


mweinelt:
I think after v2019.1 is a good time to try this out in master, then backport in time for v2019.1.1. Does that sound fine to everyone?

Issue #1415, opened 2018-06-04
gluon-status-page: show additional info


rotanid:
inspired by #1402 
related and maybe partly depending on #644
show the pretty names for additional data and hide the labels for unavailable fields

- ~~site~~ 21397ce678c32539247196135e3c30d914d5c592
- ~~domain~~ 21397ce678c32539247196135e3c30d914d5c592
- ~~roles~~ 490c522351e237ad43ab06c9d0bb8774feaca01e
- ~~wifi channels~~ 0ce961eb8c27df967e2dc3599ed2518a1d2e0faf
- bandwidth limit #2217 #2216
- ~~VPN type (active / fastd / l2tp / ...)~~ 16c5c64141ff0147943408cf7483ac64caa4479a
- ~~batman version / babel~~ 0ac3061225b255ecf81224e96ee3b80862f2d853
- ~~nexthop~~ #1878

mweinelt:
Domain pretty names is pretty expensive right now. We get the value in `nodeinfo.system.domain_code` and have to look at all files in `/lib/gluon/domains/*.json`.
A package that does this is `gluon-config-mode-domain-select` but that's because it shows all pretty names anyway.

rotanid:
showing the non-pretty names would still be an improvement ;)

mweinelt:
Really? `ffda_da_220_230` doesn't sound like anything I'd put on a status page.

rotanid:
more than nothing, although not ideal.
judging from our experience, no one except people with deep technical knowledge uses the status page anyway...

ambassador86:
I am preparing a patch for this in https://github.com/FreifunkChemnitz/gluon/tree/status-page-extension I will add a PR as soon as I am happy with it.

mweinelt:
If traffic limits are in place it would be great to publish them on the status-page as well.

rotanid:
added wifi channels, from #1727

lemoer:
@rotanid nexthop is done

AiyionPrime:
Something I lacked in the last years  and compensated by memorizing more and more router models was the base frequency of a nodes processor, as well as his installed ram. Is that something someone else missed as well?

@rotanid May you explain what `roles` is expected to contain?
Like whether a node is a supernode or not, based on `nodeinfo.vpn`?

AiyionPrime:
TIL: site.conf provides a 'roles' parameter:

[...]

PR #1402, opened 2018-05-23
gluon-status-page: show site_code & role in node details

achterin:
As suggested by @ambassador86 on irc, this PR adds _site_code_ and _node role_ information to status page.

PR #1820, opened 2019-09-15
gluon-status-page: add domain and site to status page

lemoer:
Partly implements #1415

### DIFF
--- a/package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
+++ b/package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
@@ -116,6 +116,8 @@
 					<% if nodeinfo.software.fastd then -%>
 						<dt><%:Mesh VPN%></dt><dd><%| enabled(nodeinfo.software.fastd.enabled) %></dd>
 					<%- end %>
+					<dt>Domain</dt><dd><%| site.domain_names[nodeinfo.system.domain_code]() %></dd>
+					<dt>Site</dt><dd><%| site.site_name() %></dd>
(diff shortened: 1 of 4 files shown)


rotanid:
@lemoer thanks. could you please look at #1402 and the discussion there and maybe also take "role" into your PR as in 1402?

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
```diff
@@ -116,6 +116,8 @@ <h2><%:Overview%></h2>
 					<% if nodeinfo.software.fastd then -%>
 						<dt><%:Mesh VPN%></dt><dd><%| enabled(nodeinfo.software.fastd.enabled) %></dd>
 					<%- end %>
+					<dt><%:Domain%></dt><dd><%| site.domain_names[nodeinfo.system.domain_code]() %></dd>
```
mweinelt: let's assume the firmware isn't multidomain, what happens then?

CodeFetch:
@lemoer Nice... Can you please try whether the design is still responsive? I did a similar change a while ago which lead to "jumps" in the viewports. I did a fixup of the CSS which I totally forgot about. If you experience similar problems I can have a look next week.

mweinelt:
Live with multidomain firmware here: http://[2001:67c:2ed8:1001:928d:78ff:fe23:af1c]/cgi-bin/status

I think site should appear before domain, other than that it lgtm.

rotanid:
i can confirm it works for a non-multidomain site.
note: i tested the commit before the swap of the order.

mweinelt:
Thanks!

PR #1822, opened 2019-09-19
docs, targets: mark and document several targets and devices as deprecated

mweinelt:
As per https://github.com/freifunk-gluon/gluon/pull/1791#issuecomment-533021677

https://shells.darmstadt.ccc.de/~hexa/gluon/1822/user/supported_devices.html

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -149,8 +149,8 @@ ar71xx-nand
 
   - NBG6716 [#ath10k]_
 
-ar71xx-tiny
------------
+ar71xx-tiny [#deprecated]_
+--------------------------
 
 * D-Link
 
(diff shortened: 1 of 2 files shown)


docs/user/supported_devices.rst
```diff
@@ -339,6 +339,10 @@ See also: :doc:`x86`
 Footnotes
 ---------
 
+.. [#deprecated]
+  The device or target is reaching its end of life soon. This means that we cannot guarantee
```
rotanid: maybe "means that there may be not support in the next release anymore"
why? we cannot guarantee any support, and the current sentence implicitly says that we guarantee it for all but the mentioned ones...

mweinelt: Updated the wording, please check again.

Issue #1538, opened 2018-09-26
Support for TP-Link WBS210 v2.0


datendelphin:
I got a new TP-link WBS210 and it reports v2.0 on the status page.

mweinelt:
Not supported in OpenWrt yet and no pull request yet. You might want to ask around and send your device to someone who is able to build the support.

Until that is resolved Gluon support is out of the question, sorry.

datendelphin:
I guess it is the same hardware as v1.2 I added following strings:

                       "WBS210(TP-LINK|UN|N300-2|00000000):2.0\r\n"
                       "WBS210(TP-LINK|US|N300-2|55530000):2.0\r\n"
                       "WBS210(TP-LINK|EU|N300-2|45550000):2.0\r\n"

to .support_list of tplink-safeloader.c, compiled openwrt and could upload the image and log in via ssh.

it looks similar to the patch included for CPE210 V2. I'll have a look at where I can report that to openwrt

mweinelt:
Support added to OpenWrt master in 5c5bf8b8658a588423f6ec445d7ef6a36f99a396. You'll probably also need 2524febf7927a1bf430d64b7790feb126023e3d1.

datendelphin:
I compiled openwrt with the dirty hack i mentioned above, and that works and did not brick my device. Now I'm trying to go the route with the mentioned patches, building dedicated v2 images, though I did not manage yet to build them, mostly due to lack of time and this is the first time tinkering with openwrt.

RalfJung:
> Support added to OpenWrt master in 5c5bf8b8658a588423f6ec445d7ef6a36f99a396. You'll probably also need 2524febf7927a1bf430d64b7790feb126023e3d1.

This looks like it is about the CPE210 v2, not the WBS210 v2?

datendelphin:
The expectation is that it is mainly the same hardware, except that the WBS has antenna connectors while the CPE has internal antennas.
My device in the mean time doesn't want to flash with the factory firmware properly. Still trying with tftp.

RalfJung:
FWIW, we have just successfully flashed a WBS v2 with our firmware, after patching it as described above.  Wifi meshing is working as well.

rotanid:
@RalfJung so you could go through our checklist and create a pull request maybe

RalfJung:
Well I don't have the device nor did I do the flashing, but maybe I can get the people that did to do that.^^

[...]

PR #1827, opened 2019-09-25
ar71xx-generic: Add support for TP-Link WBS210 v2

citronalco:
  - [ ] webinterface
  - [ ] tftp
  - [ ] other: <specify>
- [X] primary mac should match address on device label (or packaging) (https://gluon.readthedocs.io/en/latest/dev/hardware.html#notes)
- wired network
- wifi (if applicable)
- led mapping
  - radio leds
    - [X] should show activity
  - switchport leds
    - [X] should map to their respective port (or switch, if only one led present) 
- outdoor devices only
  - [X] added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -91,7 +91,7 @@ ar71xx-generic
   - TL-WR842N/ND (v1, v2, v3)
   - TL-WR1043N/ND (v1, v2, v3, v4, v5)
   - TL-WR2543N/ND (v1)
-  - WBS210 (v1.20)
+  - WBS210 (v1.20, v2)
   - WBS510 (v1.20)
 
 * Ubiquiti
--- a/patches/openwrt/0006-ar71xx-Add-support-for-TP-Link-WBS210-v2.patch
+++ b/patches/openwrt/0006-ar71xx-Add-support-for-TP-Link-WBS210-v2.patch
@@ -1,0 +1,148 @@
+From fa599cbdc721b445e6909b75af82620619cb6bce Mon Sep 17 00:00:00 2001
+From: Bernhard Geier <2651739+citronalco@users.noreply.github.com>
+Date: Wed, 25 Sep 2019 18:25:09 +0200
+Subject: [PATCH] Add support for TP-Links WBS210 v2
+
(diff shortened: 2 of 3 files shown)


citronalco:
Fixes #1538

patches/openwrt/0026-ar71xx-Add-support-for-TP-Link-WBS210-v2.patch
```diff
@@ -0,0 +1,148 @@
+From fa599cbdc721b445e6909b75af82620619cb6bce Mon Sep 17 00:00:00 2001
```
mweinelt: This patch needs to go upstream to OpenWrt first. Unfortunately they will not take new devices for ar71xx, so you would need to port it to ath79 first.

mweinelt: Think you've done it, it's now upstream. We'll add the ath79-generic target in #1848. After that is in we can take a backport of this commit.

https://github.com/openwrt/openwrt/commit/3930aab2cb067d04f5c9ebbf1af57744e19355eb

mweinelt:
@citronalco Hey, are you going to have time to finish this?

citronalco:
Thanks for the reminder....yes.

citronalco:
@mweinelt Would it be sufficient to simply rebase this PR and add the changes done in openwrt? (means: stick at ar71xx-generic)

WBS210v2 is in openwrt's master branch in target ath79 now, while Gluon uses openwrt branch 19.07. There were major changes in openwrt's master branch for ath79 lately, so backporting to 19.07 would be a lot of effort.

blocktrron:
I'd rather like the ath79 Backport instead of a ar71xx one. However, someone might have other opinions about this.

mweinelt:
I feel the same. The reason is that we don't have to care about an ar71xx-ath79 migration for every device we introduce directly in ath79.

citronalco:
I'm sorry, but looks like I don't have enough time to work on this.
As there's support for this device in OpenWRT it will be available in Gluon shortly anyway .

PR #1836, opened 2019-09-30
treewide: solve shellcheck warnings

2tata:


mweinelt:
rather treewide than contrib.

mweinelt:
The commit is still named that way, please remember when amending.

contrib/depdot.sh
```diff
@@ -10,7 +10,7 @@
 shopt -s nullglob
 
 
-pushd "$(dirname "$0")/.." >/dev/null
+pushd "$(dirname "$0")/.." >/dev/null || exit
```
christf: Let's exit with non-zero exit status on error.

mweinelt: In which case `set -e` would suffice.

rubo77:
Are all these changes really needed? why `command -v` instead of `which`.

those changes make the code unnecesary longer with no benefit, i guess, since those variables in quotes will probably never contain spaces

2tata:
> Are all these changes really needed? why `command -v` instead of `which`.
> 
> those changes make the code unnecesary longer with no benefit, i guess, since those variables in quotes will probably never contain spaces

command is an buildin command and which isn't .

See: https://github.com/koalaman/shellcheck/wiki/SC2230

rubo77:
Sure, but why do we have to change that? `Which` is built in in Gluon, isn't it?

NeoRaider:
Using `command` instead of `which` is definitely a good idea. I've seen some "interesting" build failures on hosts where the `which` in PATH had been replaced by some NodeJS-based implementation by NPM ...

@rubo77 I don't think you understand what a builtin command is.

contrib/depdot.sh
2tata: @mweinelt shellcheck still recommend the exit condition wether or not set -e is added.

2tata: @mweinelt applied

2tata:
All changes are done from my side :)

scripts/update-patches.sh
```diff
@@ -13,13 +13,13 @@ GLUONDIR="$(pwd)"
 for module in $GLUON_MODULES; do
 	echo "--- Updating patches for module '$module' ---"
 
-	rm -rf "${GLUON_PATCHESDIR}/$module"
+	rm -rf "${GLUON_PATCHESDIR:?}/$module"
```
mweinelt: Can you explain this?

scripts/features.sh
```diff
@@ -51,13 +51,13 @@ packages() {
 	shift
 
 	# We only allow variable names, parentheses and the operators: & | !
-	if [ "$(expr match "$cond" '.*[^A-Za-z0-9_()&|! ].*')" -gt 0 ]; then
+	if [[ "$cond" == '.*[^A-Za-z0-9_()&|! ].*' ]]; then
```
mweinelt: Will this automatically match -gt 0?

scripts/update-patches.sh
2tata: if ${GLUON_PATCHESDIR} and $module get emty at same time like:
${GLUON_PATCHESDIR} = ""
$module = ""

rm -rf "${GLUON_PATCHESDIR}/$module" will turn into rm -rf "/"

scripts/features.sh
2tata: fixed. expr match will return the counted matching characters so is at leas one character like . is matched it will be greater than 0. I change it to: `grep -q ".*[^A-Za-z0-9_()&|! ].*" <<< "$cond"`

contrib/depdot.sh
```diff
@@ -1,12 +1,11 @@
...
 
+set -e
 # Script to output the dependency graph of Gluon's packages
 # Limitations:
 #  * Works only if directory names and package names are the same (true for all Gluon packages)
 #  * Doesn't show dependencies through virtual packages correctly
 
-
-
 shopt -s nullglob
```
NeoRaider: As set and shopt are closely related, please move the `set -e` down below the comment.

package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
```diff
@@ -16,7 +16,7 @@ xor2() {
 
 interface_linklocal() {
         local macaddr="$(ubus call network.device status '{"name": "'"$1"'"}' | jsonfilter -e '@.macaddr')"
-        local oldIFS="$IFS"; IFS=':'; set -- $macaddr; IFS="$oldIFS"
+        local oldIFS="$IFS"; IFS=':'; set -- "$macaddr"; IFS="$oldIFS"
```
NeoRaider: This change is incorrect. `set` must be called without quotes, so the bytes of the MAC address are assigned to separate arguments.

scripts/features.sh
```diff
@@ -51,13 +51,13 @@ packages() {
 	shift
 
 	# We only allow variable names, parentheses and the operators: & | !
-	if [ "$(expr match "$cond" '.*[^A-Za-z0-9_()&|! ].*')" -gt 0 ]; then
+	if grep -q ".*[^A-Za-z0-9_()&|! ].*" <<< "$cond"; then
```
NeoRaider: - The leading and trailing `.*` are redundant
- Use single quotes for the regex

scripts/features.sh
```diff
@@ -51,13 +51,13 @@ packages() {
...
 		exit 1
 	fi
 
 	# Let will return false when the result of the passed expression is 0,
 	# so we always add 1. This way false is only returned for syntax errors.
-	local ret="$(env -i $vars bash --norc -ec "let _result_='1+($cond)'; echo -n \"\$_result_\"" 2>/dev/null)"
+	local ret="$(env -i "$vars" bash --norc -ec "let _result_='1+($cond)'; echo -n \"\$_result_\"" 2>/dev/null)"
```
NeoRaider: Again, we're relying on `$vars` being interpreted as multiple arguments, so the quotes break things.

scripts/filesize.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 check_command() {
-	which "$1" >/dev/null 2>&1
+	command -v "$1" >/dev/null 2>&1
```
NeoRaider: AFAICT, `command -v` doesn't print any message on errors, so the `2>&1` can be dropped (in multiple scripts)

scripts/modules.sh
```diff
@@ -1,7 +1,7 @@
 . ./modules
 [ ! -f "$GLUON_SITEDIR"/modules ] || . "$GLUON_SITEDIR"/modules
 
-FEEDS="$(echo $GLUON_FEEDS $GLUON_SITE_FEEDS | tr ' ' '\n')"
+FEEDS="$(echo "$GLUON_FEEDS" "$GLUON_SITE_FEEDS" | tr ' ' '\n')"
```
NeoRaider: Not using quotes here is preferable, as this will also drop extraneous whitespace.

scripts/update-patches.sh
```diff
@@ -13,13 +13,13 @@ GLUONDIR="$(pwd)"
...
 
 	cd "$GLUONDIR"/"$module"
 
 	n=0
 	for commit in $(git rev-list --reverse --no-merges base..patched); do
-		let n=n+1
+		(( n++ ))
```
NeoRaider: Which part doesn't shellcheck like, the `let` or the `n=n+1`? I prefer `let n++` to `(( n++ ))`.

scripts/update.sh
```diff
@@ -10,9 +10,9 @@ GLUONDIR="$(pwd)"
 for module in $GLUON_MODULES; do
 	echo "--- Updating module '$module' ---"
 	var=$(echo "$module" | tr '[:lower:]/' '[:upper:]_')
-	eval repo=\${${var}_REPO}
-	eval branch=\${${var}_BRANCH}
-	eval commit=\${${var}_COMMIT}
+	eval repo=\$\{"${var}"_REPO\}
+	eval branch=\$\{"${var}"_BRANCH\}
+	eval commit=\$\{"${var}"_COMMIT\}
```
NeoRaider: How about
```bash
eval "repo=\${${var}_REPO}"
```
or
```bash
eval 'repo=${'"${var}"'_REPO}'
```
etc.? I'm not sure which I like better, the second variant trades an ugly backslash for a few additional quotes...

package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
2tata: oh yes obviously ...

scripts/update-patches.sh
2tata: related to the let. See https://github.com/koalaman/shellcheck/wiki/SC2219

scripts/update.sh
2tata: I prefer last one. Applied

scripts/features.sh
2tata: `${vars[@]}` should also work or not?

NeoRaider: As long as you leave out the quotes, yes, `${vars[@]}` and `$vars` seem to be equivalent. Not  sure if there would be a difference if *vars* were declared as some kind of array?

2tata: hm, if I understood shellcheck docs right a quotes array should work:
https://github.com/koalaman/shellcheck/wiki/Sc2086#exceptions

NeoRaider: Yes, turning the string variable into an array sounds good.

2tata:
@NeoRaider I applied all your change request.

scripts/features.sh
```diff
@@ -31,33 +31,36 @@ sanitize() {
...
 	# We only allow variable names, parentheses and the operators: & | !
-	if [ "$(expr match "$cond" '.*[^A-Za-z0-9_()&|! ].*')" -gt 0 ]; then
+	if grep -q '[^A-Za-z0-9_()&|! ]' <<< "$cond"; then
 		exit 1
 	fi
 
 	# Let will return false when the result of the passed expression is 0,
 	# so we always add 1. This way false is only returned for syntax errors.
-	local ret="$(env -i $vars bash --norc -ec "let _result_='1+($cond)'; echo -n \"\$_result_\"" 2>/dev/null)"
+	set -A
```
NeoRaider: `set -A` doesn't exist, what is this supposed to do? Please at least build test your changes!

scripts/features.sh
```diff
@@ -31,33 +31,36 @@ sanitize() {
 	echo -n "$v"
 }
 
-vars=
+vars=()
 
 for feature in $1; do
-	if [ "$(type -t gluon_feature_nodefault_${feature})" != 'function' ]; then
+	if [ "$(type -t gluon_feature_nodefault_"${feature}")" != 'function' ]; then
```
NeoRaider: Please quote the whole string.

scripts/features.sh
2tata: Ah sorry that was ksh related. I added it locally for test reasons and accidentally committed it.

scripts/update-patches.sh
NeoRaider: Change `n++` to `++n`, otherwise this fails the `set -e` check when *n* is 0.

NeoRaider:
Squashed and merged with changes as commented, thanks.

PR #1838, opened 2019-10-02
ar71xx-generic: add ath10k packages to OCEDO Koala

blocktrron:
The OCEDO Koala was missing the correct package definition. Because of
this, firmware is potentially built with the wrong ath10k firmware /
driver.


rotanid:
maybe we need to backport this?

rotanid:
> maybe we need to backport this?

i cherry-picked it to v2019.1.x and backported it to v2018.2.x

Issue #1837, opened 2019-10-02
add Support for TP-Link Archer C60 v2


Marssl78:
Hi there!

I think, adding Archer C60 V2 would be easy, as it is supported in OpenWRT since 18.06.0. Further, it seems "functionally promising", because (other than V1) its migration from ar71xx to ath79 is completed, so might (also other than V1) run in 5-GHz-Band without OOM-issues...

edit: Forgotten to mention, that device is on the ["unofficial device wishlist"](https://md.darmstadt.ccc.de/gluon-device-wishlist#)

What I already did for testing:
I built a firmware with added profile for V2 and ran the testings as required. Primary MAC didn't match, so I built again with changed package for selecting MAC of interface 'eth0' as primary MAC, now primary MAC is matching the one labelled at the bottom of the case.

LEDs are working correctly, firmware is flashable through OEM WebInterface (will test TFTP-Method in addition this evening). Meshing over LAN / WLAN 2.4 / WLAN 5 and MeshVPN (fastd) is working, node shows up on map and clients get access to inet as expected.

I will let follow filled-out form of testings. Meanwhilest the device can be seen [on our map](https://bgl.map.ffgl.eu/#!v:m;n:d80d176dd712)

rotanid:
thank you for your work, we are looking forward to a pull request with the necessary changes and the filled-out checklist.

feel free to ask if you need any help/information!

Marssl78:
Just opened PR #1840 :-)

Testing checklist will go on tonight... ;-)

rotanid:
merged: b823b80356f63f4b52bec572cdafe30cf328e5f1

PR #1840, opened 2019-10-04
Add support archer c60v2

Marssl78:
Here are the neccessary changes to be made for #1837 (Add Support for TP-LINK Archer C60 v2)!

### DIFF
--- a/targets/ar71xx-generic
+++ b/targets/ar71xx-generic
@@ -326,7 +326,7 @@ device('tp-link-archer-c60-v1', 'archer-c60-v1', {
 })
 
 device('tp-link-archer-c60-v2', 'archer-c60-v2', {
-	broken = true, -- Testing in progress...
+	broken = false, -- Testing done
 })
 
 device('tp-link-re355', 're355-v1', {
--- a/targets/ar71xx-generic
+++ b/targets/ar71xx-generic
@@ -326,7 +326,7 @@ device('tp-link-archer-c60-v1', 'archer-c60-v1', {
 })
 
 device('tp-link-archer-c60-v2', 'archer-c60-v2', {
-	broken = false, -- Testing done
+	broken = true, -- OOM with 5GHz enabled in most environments
 })
 
 device('tp-link-re355', 're355-v1', {
--- a/targets/ar71xx-generic
+++ b/targets/ar71xx-generic
@@ -326,7 +326,6 @@ device('tp-link-archer-c60-v1', 'archer-c60-v1', {
 })
 
 device('tp-link-archer-c60-v2', 'archer-c60-v2', {
-	packages = ATH10K_PACKAGES,
 	broken = false, -- Testing done
 })
 
--- a/targets/ar71xx-generic
+++ b/targets/ar71xx-generic
@@ -325,6 +325,10 @@ device('tp-link-archer-c60-v1', 'archer-c60-v1', {
 	broken = true, -- OOM with 5GHz enabled in most environments
 })
 
+device('tp-link-archer-c60-v2', 'archer-c60-v2', {
+	broken = true, -- Testing in progress...
+})
+
 device('tp-link-re355', 're355-v1', {
 	broken = true, -- OOM with 5GHz enabled in most environments if device is 64M RAM variant
 })
--- a/targets/ar71xx-generic
+++ b/targets/ar71xx-generic
@@ -326,6 +326,7 @@ device('tp-link-archer-c60-v1', 'archer-c60-v1', {
 })
 
 device('tp-link-archer-c60-v2', 'archer-c60-v2', {
+	packages = ATH10K_PACKAGES,
 	broken = false, -- Testing done
 })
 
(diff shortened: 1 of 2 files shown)


Marssl78:
And here - finally - are the test results! B)

  - [X] webinterface (no renaming of gluon factory file in any way was neccessary flashing from Stock-Fw Ver. 161206)
  - [X] tftp (by renaming gluon factory image to "tp_recovery.bin" - notice the unusual break of file name scheme, as no model oder version prefixes are used)
  - [O] other: there is a built-in recovery function called "webpage firmware recovery" on newer TP-Link devices ([see here](https://www.tp-link.com/ca/support/faq/1482/)), but I didn't test this, as the only way to reach this seems to be to willingly (soft-)brick the device, which I didn't want to... ;-) If anyone knows how to activate this without risk, I will gladly want to know und test. ;-)
      *think `sysupgrade [-n]` or `firstboot`* firstboot as well as "sysupgrade -n sysupgrade-image.bin" lead me to a unconfiguered device, sysupgrade without -n option lead to a configuered device as expected (notice: in lack of "really newer" firmware I "upgraded" to the same version of my own firmware build)
    *usually means: gluon profile name must match image name* - Coudn't really test that, but since gluon profile name and image name match, I can at least say that the "usual condition" required in the comment is true
- [X] primary mac should match address on device label (or packaging) (https://gluon.readthedocs.io/en/latest/dev/hardware.html#notes) - with my first fw build this wasn't the case, so in second build I told package "010-primary-mac" to use mac of eth0, et voila! :-D
- wired network
- wifi (if applicable)
- led mapping
  - radio leds
    - [X] should show activity
  - switchport leds
    - [X] should map to their respective port (or switch, if only one led present) - only 1 LED which is lit as soon as cable is connected to any of the LAN ports
- outdoor devices only - it's not an outdoor device ;-)
  - [not applicable, since INdoor device!] added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`

Marssl78:
Unsure, if {packages = ATH10K_PACKAGES} should be included in device profile?
@rotanid

mweinelt:
According to https://openwrt.org/toh/hwdata/tp-link/tp-link_archer_c60_v2 the device has a QCA9882 radio. Therefore `ATH10K_PACKAGES` need to be included.

Marssl78:
> 
> 
> Therefore `ATH10K_PACKAGES` need to be included.

I also included it in the last commit, glad to have you confirm it's right! :-D

blocktrron:
@mweinelt This looks like a typo in the OpenWrt hwtable, as the FCC photos indicate a QCA9886 chip:

https://fccid.io/TE7C60/Internal-Photos/Internal-Photo-3222554

So no package definitions are needed (yet).

Marssl78:
> 
> 
> @mweinelt This looks like a typo in the OpenWrt hwtable, as the FCC photos indicate a QCA9886 chip:

To my knowledge, QCA9886 is in V1 of the device, while V2 has QCA9882. That makes a difference in ath79 migration: 5GHz in V1 is reported to be not supported by OpenWRT, while in V2 it is. Are packages also not needed for QCA9882?

Marssl78:
Did some internet research and [according to wikidevi](https://wikidevi.com/wiki/TP-LINK_Archer_C60_v2.0) @blocktrron is right. So revised to not use package definitions

targets/ar71xx-generic
```diff
@@ -325,6 +325,10 @@ device('tp-link-archer-c60-v1', 'archer-c60-v1', {
 	broken = true, -- OOM with 5GHz enabled in most environments
 })
 
+device('tp-link-archer-c60-v2', 'archer-c60-v2', {
+	broken = false, -- Testing done
```
rotanid: i think we need to set "broken = true" for the same reasons like c25/c58/c60v1

Marssl78: I think you're right. I was initially more hopeful, because in OpenWRT Techinfo the C60 is explicitly mentioned with "5GHz not supported" while in V2 there is no issue mentioned. Perhaps one should suggest changes do OpenWRT TOH/TechData as well, but that's another story. ;-)

Anyway: In lived life I am quite happy with my device, the neighbours torture it watching HD IPTV with it... :-D

rotanid: yes, in some cases 64m is enough, but reports about problems in more noisy environments lead to us marking ath10k-64m as broken
maybe newer ath10k versions are better, i don't know, but current gluon master branch is still based on OpenWrt 18.06 and therefore no changes to this rule for now.

Marssl78: So we're fine to go? Do I have to do anything? It's my very first GitHub / PR, so excuse the noob question! ;-D

rotanid: "i think we need to set "broken = true" for the same reasons like c25/c58/c60v1"
that's what you have to change, copy the broken-line from c25 or c58 or c60v1 instead of setting "broken = false"

Marssl78: I didn't know, that though you typed it in I would have to "double copy" it again. Between bringing children to bed etc. I wasn't very comprehensive in this :D 

Done. Done? :D

PR #1847, opened 2019-10-31
mpc85xx-p1020: add support for OCEDO Panda

blocktrron:
The OCEDO Panda is a SD-WAN Gateway originally sold by German vendor OCEDO. It is also known as Riverbed SteelConnect SDI-130w.

  - [ ] webinterface
  - [x] other: bootloader initramfs
     > root@64xxx-6c98eb000348:~# lua -e 'print(require("platform_info").get_image_name())'
     > ocedo-panda
- wired network
    - Device has two WAN ports - WAN Port 2 is unused with Gluon, WAN Port 1 works
- wifi (if applicable)
- led mapping
  - radio leds
    - [x] should show activity


rotanid:
what does "bootloader initramfs" mean?
does it match our requirement "installation possible without opening the device (e.g. to access serial console) " ?

blocktrron:
It's the same procedure as for the OCEDO Koala (which is already non-broken in Gluon).

In short: You will need to provide the OpenWrt Initramfs image using a TFTP server - hold the Reset button when powering on the Device and flash Gluon using sysupgrade.

No additional tools / opening of the case required.

rotanid:
> (which is already non-broken in Gluon).

which does not mean anything if rules would have changed in the meantime ;)

nevertheless, merging per your explanation.

PR #1849, opened 2019-11-03
modules: bump OpenWrt 19.07

blocktrron:
- Update to the latest OpenWrt 19.07 and associated feeds.
- Adjust ipq40xx target name, as otherwise the target won't built anymore.


blocktrron:
Run-tested: ipq40xx (Fritz!Box 4040)

https://meshviewer.darmstadt.freifunk.net/#!/en/map/989bcbd619e2

mweinelt:
Same for ipq806x?

rotanid:
> Same for ipq806x?
> 
> Also please split up the module bumps and add a shortlog.

+1

blocktrron:
@mweinelt @rotanid 

Updated the commit description. This now only bumps the OpenWrt feed, I will bump the packages seperately.

This change is only necessary for ipq40xx, ipq806x has no generic subtarget.

mweinelt:
> This change is only necessary for ipq40xx, ipq806x has no generic subtarget.

https://github.com/openwrt/openwrt/commit/7407e72399179a185c548c55aaa61a44af497f3f

blocktrron:
You are right, no idea on what i was looking at. Will include this too.

belzebub40k:
`docs/user/supported_devices.rst` need to be adjusted for ipq806x as well.

PR #1854, opened 2019-11-06
lantiq-xway:  support for AVM FRITZ!Box 7312

misanthropos:
  - [ ] webinterface
  - [ ] tftp
  - [x] other: EVA
  - [?] must have working autoupdate
- [x] xxxxx/xxx/phone button must return device into config mode
- [ ] primary mac should match address on device label (or packaging) (https://gluon.readthedocs.io/en/latest/dev/hardware.html#notes)
- wired network
- wifi (if applicable)
- led mapping
  - radio leds
    - [x] should show activity
  - switchport leds
- outdoor devices only
~

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -224,6 +224,13 @@ ipq806x-generic [#80211s]_
 
   - R7800
 
+lantiq-xway
+-----------
+
+* AVM
+
+  - FRITZ!Box 7312 [#avmflash]_
+
 mpc85xx-generic
 ---------------
 
--- a/targets/lantiq-xway
+++ b/targets/lantiq-xway
@@ -1,0 +1,3 @@
+device('avm-fritz-box-7312', 'avm_fritz7312', {
+	factory = false,
+})
--- a/targets/targets.mk
+++ b/targets/targets.mk
@@ -7,6 +7,7 @@ endif
 $(eval $(call GluonTarget,ar71xx,nand))
 $(eval $(call GluonTarget,brcm2708,bcm2708))
 $(eval $(call GluonTarget,brcm2708,bcm2709))
+$(eval $(call GluonTarget,lantiq,xway))
 $(eval $(call GluonTarget,mpc85xx,generic))
 $(eval $(call GluonTarget,mpc85xx,p1020))
 $(eval $(call GluonTarget,ramips,mt7621))


blocktrron:
Please fix the typo in the commit message.

docs/user/supported_devices.rst
```diff
@@ -21,6 +21,7 @@ ar71xx-generic
 * AVM
 
   - Fritz!Box 4020 [#avmflash]_
+  - Fritz!Box 7312 [#avmflash]_
```
blocktrron: This is not the correct place, as the 7312 is not part of the ar71xx-generic target. Add a new section for the lantiq-xway target and add the FRITZ!Box 7312 there.

docs/user/supported_devices.rst
```diff
@@ -224,6 +224,15 @@ ipq806x-generic [#80211s]_
 
   - R7800
 
+
+lantiq-xway
+-----------
+
+* AVM
+
+  - Fritz!Box 7312 [#avmflash]_
```
```suggestion
  - FRITZ!Box 7312 [#avmflash]_
```

blocktrron: Actually - we made this mistake at some places already - just like the inconsistent whitespacing. I would merge this as it is and clean up both points after - Is this okay with you?

mweinelt: Fine with me.

blocktrron:
I've made some changes to your commit: 

 - Remove the unnecessary whitespace newlines in the docs
 - added an `add` before `support for ...` to your title

Merging this now.

ecsv:
Which button is used here for the config mode? Because a user just complained that neither the dect nor the wlan button will do anything for him:

https://github.com/openwrt/openwrt/blob/e3eaf578082ee26ebc0021faed8ae42aca007f9e/target/linux/lantiq/files-5.4/arch/mips/boot/dts/lantiq/ar9_avm_fritz7312.dts#L30

misanthropos:
Assuming it runs gluon (meaning not OpenWRT) you press the DECT button
for 3 seconds.

See: https://gluon.readthedocs.io/en/latest/features/configmode.html

I believe the vanilla OpenWRT does not offer a config mode key.

ecsv:
Thanks for confirming this.

For some reason, it doesn't seem to work for him. But I don't have access to the device and trying remotely to assist him in getting the device up and running. And now it seems like he isn't happy about my word "complained". So I will stop for now.

PR #1856, opened 2019-11-06
lantiq-xway: enable multithreading

misanthropos:
By default SMP it is disabled / however I have not seen any problems with enabling it.


misanthropos:
wrong branch

Issue #1747, opened 2019-06-15
Drop support for IBSS Mesh


mweinelt:
The scheduled domain switching is now in place and v2019.1 will be the last Gluon major release with support for IBSS Mesh.

Support for IBSS Mesh will be dropped in master as soon as v2019.1 is branched.

kaechele:
If so target `mvebu-cortexa9` can be removed and #1730 can be closed.
No device in that target can currently do 802.11s.

belzebub40k:
@kaechele That target was also not capable to do IBSS. It was originally added because there was a statement they are going to work on 802.11s support which was later revoked. But I don't like the idea of dropping it as we have a few devices running in our community.

rubo77:
I checked with my site-conf scraper: https://github.com/rubo77/site-conf-scraper which uses all sites in the wiki site-conf page and found only 19 sites still using IBSS: ffruhr-site-ffalpen, ffruhr-site-ffher, ffruhr-site-ffhnx, ffruhr-site-ffniers, site-ff3l, site-ffa, site-ffac, site-ffbs, site-ffgoe, site-ffhb, site-ffhl, site-ffhmue, site-ffl, site-ffms, site-ffniers, site-ffrgb, site-ffrn, site-fftr, site-neuss

(not sure if any of them are still active)

rotanid:
@rubo77 your way to create this list may be broken, as for example we (site-ffa) don't use IBSS.

rubo77:
I just downloaded the default branches (mostly master) so, yes, the list is outdated, if they don't update their default branch. but it is the best we can get automatically so far, i guess. I am open for ideas, how to analyze all sites differently in my https://github.com/rubo77/site-conf-scraper

rotanid:
@rubo77 we have no branch "master" and the default branch uses 802.11s. check before writing please.
but this gets off-topic, feel free to contact me via IRC on this matter.

kpanic23:
@rubo77 Well, we (ff3l) _do_ use the master branch and have dropped IBSS support in March 2017 (https://github.com/ff3l/site-ff3l/commit/187f6834537ce0eb1a03cafa2113b2cc82046fea). I have no idea what your scraper does, but it definitively is wrong. Maybe it clashes with multi-domain sites having neither nor in site.conf?

yanosz:
Hello,

we're still using IBSS in cologne. We've had some discussions about moving to 802.11s but I'm not, if this is a good way way to go.

[...]

PR #1858, opened 2019-11-06
Drop IBSS support

NeoRaider:
As discussed before, drop IBSS support completely to simplify our setup.

The PR does not deal with the MAC address assignment yet; this will require some more thought and it can be handled in a separate PR.

I've also noticed that we were using `kmod-ath10k-ct` even for 11s on x86. ~I didn't touch it for now, but I assume this wasn't intended?~

Closes: #1747


docs/user/supported_devices.rst
```diff
@@ -359,13 +359,7 @@ Footnotes
...
 
 .. [#ath10k]
-  Device uses the ath10k WLAN driver. Images are built for 11s by default unless GLUON_WLAN_MESH
-  is set as described in :ref:`getting-started-make-variables`.
-
-.. [#80211s]
-  Device or target does not support IBSS. Images are built by default unless
-  GLUON_WLAN_MESH is explicitly set to something other than *11s*. Targets that
-  are affected as a whole may not be selected for build in this case.
+  Device uses the ath10k WLAN driver.
```
mweinelt: Isn't this useless information now?

NeoRaider: I guess we can drop this as well.

targets/x86.inc
```diff
@@ -37,6 +32,7 @@ packages {
 	'kmod-usb2',
 	'kmod-usb3',
 	'kmod-usb-serial',
+	'kmod-ath10k-ct',
```
mweinelt: Would replace this for reasons like this: https://github.com/freifunk-gluon/gluon/issues/1495

blocktrron: This would be a separate commit however imho. While we are at it, we can also add firmware for the QCA9886/9984.

blocktrron:
Even though we change this later, the ibss description of the generic_mac function should be 
dropped:

https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua#L175

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -202,20 +165,9 @@ local function configure_mesh_wireless(radio, index, config)
...
-			ibss_disabled,
-			mesh_disabled,
-			config.ibss.disabled(false)
-		)
-	)
 	configure_mesh(config.mesh(), radio, index, suffix,
 		first_non_nil(
-			mesh_disabled,
-			ibss_disabled,
+			is_disabled('mesh_' .. radio_name),
```
blocktrron: I think this is a corner case - But this way, the wireless enable-setting from IBSS is not transitioned to 11s, in case this is the first update to feature 11s. We need to save the enable-state for IBSS before deleting it's config.

NeoRaider: Users should migrate using scheduled-domain-switch on v2019.1.x, this just needs to be documented in the release notes.

blocktrron: Okay, this is a valid point. :+1:

blocktrron:
I've run-tested this changeset and on my 4040, everything still works as expected, whether i upgrade from a previous Gluon release or do a firstboot.

PR #1859, opened 2019-11-07
lantiq-xrx200: add support for AVM FRITZ!Box 7412

Dark4MD:
      
      * [ ]  webinterface
      * [ ]  tftp
      * [x]  other: eva_ramboot.py script in openwrt

      * [?] must have working autoupdate

    * [x]  wps button must return device into config mode

    * [n/a]  primary mac should match address on device label (or packaging) (https://gluon.readthedocs.io/en/latest/dev/hardware.html#notes)

    * wired network
      
    * wifi (if applicable)
      
    * led mapping
      
      * radio leds
        
        * [x]  should show activity
      * switchport leds
        
        * [n/a]  should map to their respective port (or switch, if only one led present)
        * [n/a]  should show link state and activity

    * outdoor devices only
      
      * [n/a]  added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`
        ~

HowTo Flash is described in the Commit for the Device in OpenWRT
https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=20f48c8ae334f8502e756282995cc84d89fa81c1


blocktrron:
> must have working autoupdate

Does `lua -e 'print(require("platform_info").get_image_name())'` match the Gluon platform name? In your case, this would be (per your definition) `avm-fritz-box-7412`

Dark4MD:
> lua -e 'print(require("platform_info").get_image_name())'

Yes it does. And i checked it again just now to be sure.

mweinelt:
Please create a new remark in the supported devices, that explains the need to use the eva_ramboot script and maybe even links to it.

blocktrron:
@mweinelt 
I think this is something we should do after creating the documentation (which is quite time-consuming - firtzfla.sh was a noticeable time-effort).

docs/user/supported_devices.rst
```diff
@@ -224,6 +224,14 @@ ipq806x-generic
...
   - R7800
 
+lantiq-xrx200
+-------------
+
+* AVM
+
+  - FRITZ!Box 7412
+
+
```
blocktrron: Drop this additional newline.

Dark4MD: done

docs/user/supported_devices.rst
```diff
@@ -224,6 +224,13 @@ ipq806x-generic
 
   - R7800
 
+lantiq-xrx200
+-------------
+
+* AVM
+
+  - FRITZ!Box 7412
```
```suggestion
  - FRITZ!Box 7412 [#eva_ramboot]_
```

PR #1861, opened 2019-11-07
lantiq-xway: enable multithreading

misanthropos:
Even though not officially supported, no problems have been encountered by enabling this feature so far.


blocktrron:
This is something to be decided by an upstream lantiq maintainer.
We do not want to diverge from OpenWrt in that downstream.

Any other objection on this @mweinelt @NeoRaider

mweinelt:
Agreed, we are not going to track this patch downstream. We are by no means experts on the lantiq-xway platform, so please submit this upstream. If it gets merged we can rediscuss backporting it.

Issue #1570, opened 2018-11-16
Migrate ar71xx target to ath79


blocktrron:
Upstream is planning to drop the ar71xx target after the 19.01 release. To allow us to smoothly upgrade to OpenWRT releases after that, we need to work on a migration path for existing devices.

To upgrade an ar71xx based device to ath79, the following adjustments have to be made:

 - Change WiSoC device path
 - Reinitialize LED configuration

Additionally all the downstream integration needs to be reverified, expecially with regards to

 - Errors (LEDs, Buttons, Network config)
 - Model naming changes (Autoupdater filename)
 - Primary MAC selection

This process has to be repeated for every device/revision individually. We have to basically treat each device as a if they were a new integration. With each error we risk breaking already deployed devices.

However not every device that is currently supported in Gluon will be ported over to ath79. Devices which rely on custom initialization code implemented in the machine-file are among those that are most likely not going to be ported.

Examples are:

 - Meraki MR18

I think we should focus our efforts on widely used and popular devices, among these:

## AVM
 - FRITZ!Box 4020
 - FRITZ!WLAN Repeater 450E

## TP-Link
 - Archer C5 v1
 - Archer C59 v1
 - Archer C7 v2
 - Archer C7 v4
 - Archer C7 v5
 - CPE210 v1
 - CPE210 v2
 - CPE510 v1
 - TL-WDR3600
 - TL-WDR4300
 - TL-WR842N v3
 - TL-WR1043ND v2
 - TL-WR1043ND v3
 - TL-WR1043ND v4
 - TL-WR1043N v5

## Ubiquiti
 - UniFi AC Lite
 - UniFi AC Pro
 - UniFi AC Mesh
 - UniFi AC Mesh Pro

CodeFetch:
What about WR841N(D)s and WR940Ns? They are about 98% of our deployed devices...

blocktrron:
@CodeFetch There are already discussions about that, but speaking about 4/32 devices i had a hard time getting my test-device (TL-WR740N v1) to work reliably in the ath79 target for Gluon (sloppy Config-mode was the smallest problem) some months ago on OpenWRT master.

As this change is going to hit Gluon probably somewhere in 2020 (19.01 supports ath79 and ar71xx in parallel) it's likely 4/32 devices have a hard time running Gluon by then. Also see this current discussion about this topic upstream: https://forum.openwrt.org/t/4-32-devices-vs-17-01-18-06/25203

[...]

PR #1862, opened 2019-11-09
ath79: add GLinet and AVM-devices

SvenRoederer:
This adds the following devices, which are already present in ar71xx:

* AVM Fritz!BOX 4020, WLAN-Repeater 300e
* GLinet AR150, GLinet AR300m (lite)
  * GLinet AR300m gets disabled, as done in OpenWRT

### DIFF
--- a/targets/ath79-generic
+++ b/targets/ath79-generic
@@ -2,6 +2,20 @@
 -- local ATH10K_PACKAGES_QCA9887 = {'kmod-ath10k', '-kmod-ath10k-ct', 'ath10k-firmware-qca9887', '-ath10k-firmware-qca9887-ct'}
 local ATH10K_PACKAGES_QCA9888 = {'kmod-ath10k', '-kmod-ath10k-ct', 'ath10k-firmware-qca9888', '-ath10k-firmware-qca9888-ct'}
 
+-- GL Innovations
+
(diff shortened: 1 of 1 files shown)


mweinelt:
Thanks for your contribution, but I'm not sure we are ready yet to accept devices in ath79 which are also in ar71xx.

- One issue will be colliding filenames when both targets are being built, so we will likely need to disable devices in ar71xx as soon as they're added in ath79.
- Another issue is the question how migration between ar71xx and ath79 will work. We've discussed this a bit in #1570.
- And third issue is … we're going to need device checklists (https://github.com/freifunk-gluon/gluon/wiki/Device-Integration-checklist) for every device that we accept/migrate into ath79.

SvenRoederer:
Esp. the issue with having the device in both targets I also had in mind. So this PR might be a good place to discuss on this.
As I had this commits prepared already, I published them to not get them lost.

blocktrron:
@SvenRoederer Until we have a migration strategy, ath79 and ar71xx is XOR in terms of device support. Also the checklist is missing as @mweinelt pointed out.

Regarding ath79 migration, there is already a tracking issue present in #1570 . So I'm in favor of closing this PR and discuss it there. Any objections against this?

SvenRoederer:
I'm fine with this.

PR #1863, opened 2019-11-09
targets: add Fritz!BOX specific packages

SvenRoederer:
During building a firmware with a stripped down package-list I realized that the Fritz!BOX boards depend on the fritz-tffs and fritz-caldata package.

### DIFF
--- a/targets/ar71xx-generic
+++ b/targets/ar71xx-generic
@@ -47,14 +47,17 @@ device('allnet-all0315n', 'all0315n', {
 
 device('avm-fritz-box-4020', 'fritz4020', {
 	factory = false,
+	packages = {'fritz-tffs'},
 })
 
 device('avm-fritz-wlan-repeater-300e', 'fritz300e', {
 	factory = false,
+	packages = {'fritz-tffs'},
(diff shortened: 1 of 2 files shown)


mweinelt:
They are already installed through OpenWrt's `DEVICE_PACKAGES` definition.

https://github.com/openwrt/openwrt/blob/openwrt-19.07/target/linux/ar71xx/image/generic.mk#L1316
https://github.com/openwrt/openwrt/blob/openwrt-19.07/target/linux/ipq40xx/image/Makefile#L128

blocktrron:
@SvenRoederer These packages are pulled in by OpenWrt - why should it be necessary to include them here?

SvenRoederer:
Looks like this is triggered by using "CONFIG_TARGET_PER_DEVICE_ROOTFS=y".

When disabling this during build the image creation fails as of this missing dependencies. In https://github.com/freifunk-berlin/firmware/commit/b633d4f153f229db869cede45ac08e8427888987  I enabled this CONFIG-option and the imagebuilder was able to finish.

PR #1864, opened 2019-11-09
ramips-mt7621: add support for Netgear R6220

blocktrron:
The NETGEAR R7800 was previously only supported as BROKEN in Gluon. With OpenWrt 19.07, the remaining bugs are ironed out, so the broken flag for target as well as device can be removed.

  - [x] webinterface
  - [x] other: nmrpflash
     > netgear-r6220
- wired network
- wifi (if applicable)
- led mapping
  - radio leds
    - [x] should show activity

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -280,6 +280,10 @@ ramips-mt7621
 
   - DIR-860L (B1)
 
+* NETGEAR
+
+  - R6220
+
 * Ubiquiti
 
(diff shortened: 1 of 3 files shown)


rotanid:
thanks.
can this be backported to OpenWrt 19.07 or are there too many changes regarding this device?

blocktrron:
It is backported to 19.07 and will be included in the final release.

rotanid:
thank you

PR #1866, opened 2019-11-10
ramips-mt7620: add support for tp-link archer c20 v1

ambassador86:
  * [x]  webinterface
    * [x]  must keep/forget configuration (if applicable) _think `sysupgrade [-n]` or `firstboot`_
  * [x]  must have working autoupdate   _usually means: gluon profile name must match image name_
'tp-link-archer-c20-v1'
* wired network  
  * [X]  should support all network ports on the device
  * [X]  must have correct port assignment (WAN/LAN)
* wifi
  * [X]  association with AP must be possible  
  * [X]  association with 802.11s mesh must be working  
  * [X]  ap/mesh mode must work in parallel 
* led mapping
  * power/sys led
    * [X]  lit while the device is on
    * [X]  should display config mode blink sequence
  * radio leds
    * [x]  should show activity
  * switchport leds
    * [x]  should show link state and activity 
* [X]  wps button must return device into config mode
* [X]  primary mac should match address on device label (or packaging) (https://gluon.readthedocs.io/en/latest/dev/hardware.html#notes)

### DIFF
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac
@@ -61,7 +61,7 @@ elseif platform.match('mpc85xx', 'p1020', {'aerohive,hiveap-330'}) then
   table.insert(try_files, 1, '/sys/class/net/eth0/address')
 elseif platform.match('mpc85xx', 'p1020', {'ocedo,panda'}) then
   table.insert(try_files, 1, '/sys/class/net/eth1/address')
-elseif platform.match('ramips', 'mt7620', {'miwifi-mini', 'tplink,c2-v1', 'c20i', 'c50'}) then
+elseif platform.match('ramips', 'mt7620', {'miwifi-mini', 'tplink,c2-v1', 'c20-v1', 'c20i', 'c50'}) then
   table.insert(try_files, 1, '/sys/class/net/eth0/address')
 elseif platform.match('ramips', 'mt7621', {'dir-860l-b1'}) then
   table.insert(try_files, 1, '/sys/class/ieee80211/phy1/macaddress')
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -303,6 +303,7 @@ ramips-mt7620
 * TP-Link  
 
   - Archer C2 v1
+  - Archer C20 (v1)
   - Archer C20i
   - Archer C50 v1
 
--- a/targets/ramips-mt7620
+++ b/targets/ramips-mt7620
@@ -31,7 +31,6 @@ device('nexx-wt3020-8m', 'wt3020-8M', {
 	},
 })
 
-
 -- TP-Link
 
 local tplink_region_suffix = ''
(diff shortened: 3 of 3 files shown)


package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac
```diff
@@ -61,6 +61,8 @@ elseif platform.match('mpc85xx', 'p1020', {'ocedo,panda'}) then
   table.insert(try_files, 1, '/sys/class/net/eth1/address')
 elseif platform.match('ramips', 'mt7620', {'miwifi-mini'}) then
   table.insert(try_files, 1, '/sys/class/net/eth0/address')
+elseif platform.match('ramips', 'mt7620', {'tplink,c20-v1'}) then
```
mweinelt: This condition should be unified with the one above.

ambassador86: Like now?

mweinelt: Yes, except for the spacing. Check how it's done above.

ambassador86:
I forgot the running test device: http://[2001:bc8:3f13:ffc2:c6e9:84ff:fe4f:34ee]

mweinelt:
Blocked because the WiFI-LED isn't working and needs to be fixed upstream.

txt-file:
@mweinelt Am I understanding it correct that you only want to integrate fully working devices? Then we could just drop any BROKEN device. Please explain if I am mistaken.

rotanid:
> @mweinelt Am I understanding it correct that you only want to integrate fully working devices? Then we could just drop any BROKEN device. Please explain if I am mistaken.

i haven't talked to mweinelt about this, but the problem is that such an LED issue can't be fixed by a sysupgrade later, it has to work on first configuration. (as far as i know)

also, integrating before where an upstream patch is available does not really help in getting it fixed upstream

blocktrron:
@ambassador86 We might have another problem here - Can you verify which MAC address is used on the 5GHz part? The MT7610 (from my observation) seems to have a generic Ralink MAC address, which will lead to problems as soon as two devices with the same MAC participate in a mesh.

Dark4MD:
When i build my Image for the C50v1 i also added the c20v1 because i have one myself laying aroung. For me the LEDs work on this device as expected as well. So only missing is removing the broken tag and adding it to supported_devices.

blocktrron:
```
 Only TFTP is supported at this moment.

Warning about bricking your device You need to use TFTP to install. Do not flash from the stock UI because you will get an lzma kernel decompression error and you’ll have to use serial console to recover from soft-brick. 
```

Does the Factory image work here from the TP-Link Web-UI?

ambassador86:
@blocktrron i haven't read this in the OpenWRT wiki and used the factory image with the web ui. And my router still works. I will try the latest gluon master to check the LEDs.

ambassador86:
Hello @blocktrron ,
with the latest gluon master the LEDs are working correctly. I was suprised that even the USB LED is lighting up when I plug in a pen drive.

PR #1870, opened 2019-11-19
docs: add mt76 to supported chipsets

blocktrron:
The docs previously only listed ath9k and ath10k based chipsets as fully
supported by Gluon, however we also support mt76.


mweinelt:
Much more to do on that page, we should likely overhaul the whole page.

rotanid:
we can always overhaul the page, but that's not a good reason to leave this change out IMO.

Issue #1728, opened 2019-05-24
BATMAN_V: Throughput on wired interfaces is not measured correctly


krombel:
### Bug report

**What is the problem?**
When there are two devices which are connected via e.g. vxlan and wifi the wifi connection is preferred.
batctl always responds "1.0" as speed while wifi has higher values (which is the reason for the preference

**What is the expected behaviour?**
vxlan shows a throughput near to the link capacity. At least sth bigger than "1.0"

**Gluon Version:**
v2018.2.1

**Custom patches:**
@ambassador86 pasted a link on #1726 to some patches from Freifunk Chemnitz which contains a patch which overwrites that value for batman-adv:
https://gitlab.com/FreifunkChemnitz/site-ffc/blob/b89e6568/patches/batman-v-wired-tpo.patch

This makes me assume that this is not an issue for our community (Freifunk München) but for BATMAN_V in general.

rotanid:
this looks like an upstream issue, but maybe it's better with batman-adv 2019.2 ? there's a PR for this, maybe you can test that, otherwise try to report it upstream please

mweinelt:
If not we can set throughput manually on MoW/MoL interfaces to something reasonable like 100 Mbit/s.

T-X:
One issue identified by @NeoRaider seems to be that the vxlan driver currently does not inherit/propagate the throughput of the underlying interface. A patch for the vxlan driver is work in progress (by @NeoRaider).

However I just noticed, that even with vxlan=false in the site.conf the BATMAN_V throughput is not determined correctly:

<pre><code>root@freifunk060464a439b1:/# batctl n
[B.A.T.M.A.N. adv openwrt-2018.1-8, MainIF/MAC: primary0/5e:75:6a:4b:97:ab (bat0/06:04:64:a4:39:b1 BATMAN_V)]
IF             Neighbor              last-seen
a6:73:e9:f5:86:48    0.380s (        1.0) [    br-wan]
f6:9b:e5:70:62:28    0.220s (        1.0) [    br-wan]
</code></pre>

Only using the eth1 (the only interface part of br-wan) directly seems to work:

<pre><code>root@freifunk060464a439b1:/# batctl if del br-wan
root@freifunk060464a439b1:/# brctl delif br-wan eth1
root@freifunk060464a439b1:/# batctl if add eth1
root@freifunk060464a439b1:/# batctl n
[...]

PR #1872, opened 2019-11-20
kernel: implement get_link_ksettings ethtool method for VXLAN and bridge

NeoRaider:
This allows BATMAN_V to determine the speed of the actual link underlying
VXLAN and bridge devices.

Closes: #1728

@krombel, @T-X, anyone with a BATMAN_V mesh: please test!

The included kernel patches have been accepted into net-next and they will be part of kernel 5.5. If this is working as expected, I'll also push them to the OpenWrt master, so we can get rid of the downstream patches with OpenWrt 19.07+1.

### DIFF
--- a/patches/openwrt/0006-kernel-implement-get_link_ksettings-ethtool-method-for-VXLAN-and-bridge.patch
+++ b/patches/openwrt/0006-kernel-implement-get_link_ksettings-ethtool-method-for-VXLAN-and-bridge.patch
@@ -1,0 +1,142 @@
+From: Matthias Schiffer <mschiffer@universe-factory.net>
+Date: Wed, 20 Nov 2019 00:07:49 +0100
+Subject: kernel: implement get_link_ksettings ethtool method for VXLAN and bridge
+
+Although the patch was accepted upstream, we add it in the "pending"
+directory to avoid modifying patch files provided by OpenWrt.
+
+Signed-off-by: Matthias Schiffer <mschiffer@universe-factory.net>
+
+diff --git a/target/linux/generic/pending-4.14/950-v5.5-0001-vxlan-implement-get_link_ksettings-ethtool-method.patch b/target/linux/generic/pending-4.14/950-v5.5-0001-vxlan-implement-get_link_ksettings-ethtool-method.patch
+new file mode 100644
+index 0000000000000000000000000000000000000000..6d8039ff704c0842eea900b1af2944fe05f586a3
+--- /dev/null
++++ b/target/linux/generic/pending-4.14/950-v5.5-0001-vxlan-implement-get_link_ksettings-ethtool-method.patch
+@@ -0,0 +1,50 @@
++From aa5f15831b6d112a5787c2e097f10da72b94caa6 Mon Sep 17 00:00:00 2001
++Message-Id: <aa5f15831b6d112a5787c2e097f10da72b94caa6.1574203789.git.mschiffer@universe-factory.net>
++From: Matthias Schiffer <mschiffer@universe-factory.net>
++Date: Tue, 12 Nov 2019 22:12:24 +0100
++Subject: [PATCH] vxlan: implement get_link_ksettings ethtool method
++
++Similar to VLAN and similar drivers, we can forward get_link_ksettings to
++the lower dev if we have one to get meaningful speed/duplex data.
++
++Signed-off-by: Matthias Schiffer <mschiffer@universe-factory.net>
++Signed-off-by: David S. Miller <davem@davemloft.net>
++(cherry picked from commit 36fe3a61aaca2be4ef7484187e7484d19d54553e)
++---
++ drivers/net/vxlan.c | 24 ++++++++++++++++++++++--
++ 1 file changed, 22 insertions(+), 2 deletions(-)
++
++--- a/drivers/net/vxlan.c
+++++ b/drivers/net/vxlan.c
++@@ -2831,9 +2831,29 @@ static void vxlan_get_drvinfo(struct net
++ 	strlcpy(drvinfo->driver, "vxlan", sizeof(drvinfo->driver));
++ }
++ 
+++static int vxlan_get_link_ksettings(struct net_device *dev,
+++				    struct ethtool_link_ksettings *cmd)
+++{
+++	struct vxlan_dev *vxlan = netdev_priv(dev);
+++	struct vxlan_rdst *dst = &vxlan->default_dst;
+++	struct net_device *lowerdev = __dev_get_by_index(vxlan->net,
+++							 dst->remote_ifindex);
+++
+++	if (!lowerdev) {
+++		cmd->base.duplex = DUPLEX_UNKNOWN;
+++		cmd->base.port = PORT_OTHER;
+++		cmd->base.speed = SPEED_UNKNOWN;
+++
+++		return 0;
+++	}
+++
+++	return __ethtool_get_link_ksettings(lowerdev, cmd);
+++}
+++
++ static const struct ethtool_ops vxlan_ethtool_ops = {
++-	.get_drvinfo	= vxlan_get_drvinfo,
++-	.get_link	= ethtool_op_get_link,
+++	.get_drvinfo		= vxlan_get_drvinfo,
+++	.get_link		= ethtool_op_get_link,
+++	.get_link_ksettings	= vxlan_get_link_ksettings,
++ };
++ 
++ static struct socket *vxlan_create_sock(struct net *net, bool ipv6,
+diff --git a/target/linux/generic/pending-4.14/950-v5.5-0002-bridge-implement-get_link_ksettings-ethtool-method.patch b/target/linux/generic/pending-4.14/950-v5.5-0002-bridge-implement-get_link_ksettings-ethtool-method.patch
+new file mode 100644
+index 0000000000000000000000000000000000000000..d56d9ee2849467c2757c99c86db6bc196dc58794
+--- /dev/null
++++ b/target/linux/generic/pending-4.14/950-v5.5-0002-bridge-implement-get_link_ksettings-ethtool-method.patch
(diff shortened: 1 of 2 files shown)


mweinelt:
Also poking @txt-file.

awlx:
We (myself or @krombel) are going to build it on the weekend for FFMUC :).

awlx:
We have build the firmware with the patch and it seems it doesn't work correctly at least on the Fritzbox 4040 it reports 1Mbit/s.
http://[2001:608:a01:109:9a9b:cbff:fee2:53d8]/cgi-bin/status
http://[2001:608:a01:109:e695:6eff:fe45:1ff6]/cgi-bin/status

Fritzbox 4040:
```
root@aw-fritzbox-4040:/sys# batctl n
[B.A.T.M.A.N. adv openwrt-2019.2-4, MainIF/MAC: primary0/6e:5c:cc:06:cc:e3 (bat0/98:9b:cb:e2:53:d8 BATMAN_V)]
IF             Neighbor              last-seen
52:98:0e:9b:5f:9c    0.060s (        1.0) [vx_mesh_lan]
c6:da:b6:13:5f:2f    0.070s (        1.0) [vx_mesh_lan]
ea:66:41:d5:31:86    0.060s (        1.0) [vx_mesh_lan]
52:98:0e:9b:5f:9d    0.470s (       45.2) [     mesh1]
52:98:0e:9b:5f:99    0.180s (       47.3) [     mesh0]
```
Gl.Inet 1300:
```
[B.A.T.M.A.N. adv openwrt-2019.2-4, MainIF/MAC: primary0/52:98:0e:9b:5f:9b (bat0/e4:95:6e:45:1f:f6 BATMAN_V)]
IF             Neighbor              last-seen
6e:5c:cc:06:cc:e4    0.210s (        1.0) [vx_mesh_lan]
c6:da:b6:13:5f:2f    0.280s (        1.0) [vx_mesh_lan]
ea:66:41:d5:31:86    0.340s (        1.0) [vx_mesh_lan]
6e:5c:cc:06:cc:e5    0.210s (       63.9) [     mesh1]
6e:5c:cc:06:cc:e1    0.100s (       46.5) [     mesh0]

```

NeoRaider:
Ugh... I just found this code in batadv:
```C
        /* Virtual interface drivers such as tun / tap interfaces, VLAN, etc
         * tend to initialize the interface throughput with some value for the
         * sake of having a throughput number to export via ethtool. This
         * exported throughput leaves batman-adv to conclude the interface
         * throughput is genuine (reflecting reality), thus no measurements
         * are necessary.
         *
         * Based on the observation that those interface types also tend to set
         * the link auto-negotiation to 'off', batman-adv shall check this
         * setting to differentiate between genuine link throughput information
         * and placeholders installed by virtual interfaces.
         */
        if (ret == 0 && link_settings.base.autoneg == AUTONEG_ENABLE) {
```

While this may make sense for tunnel interfaces like TAP, it usually doesn't for VLAN and VXLAN (which just forward the link info from the underlying device), and using the autoneg setting to differentiate between physical and virtual devices is very hacky... I'm not sure if the detection would be improved easily though: It should be possible to special-case VLAN and VXLAN, but then what about VLAN on TAP? And bridges are even worse...

I'd argue that using a placeholder value from a virtual interface is not significantly worse than a link speed from an actual Ethernet interface, which will also just be an upper limit of the available throughput: There may be a slower Ethernet switch behind the port (for example the builtin switch in most low-cost devices, where the CPU-switch connection is reported as 1000MBit/s, even though the external ports only support 100MBit/s); another case where the usable bandwidth is significantly lower would be a connection over a point-to-point wireless bridge.

@ecsv Any opinion on this?

NeoRaider:
FWIW, TUN/TAP seems to be the only tunnel-like interface type that implements get_link_ksettings() at all, and it has a very conservative default of 10MBit/s. All in all, the `link_settings.base.autoneg == AUTONEG_ENABLE` check doesn't seem very useful to me and should just be removed IMO.

ecsv:
I think @lindnermarek and @ordex are the correct persons to discuss this. They were working on the change https://github.com/open-mesh-mirror/batman-adv/commit/d60b8fc69ef24d56a10b85210685f6a7d15a8a08

NeoRaider:
I've reverted the batadv commit disabling the ethtool-based metrics for certain interface types for now, which should fix this issue.

In my opinion, ethtool-based metrics should be used for all non-wireless interface types that provide them for now. A better solution can be worked on when actual throughput measurement has been implemented in batadv.

awlx:
```
root@aw-glinet-b1300:/sys# cat ./devices/virtual/net/vx_mesh_lan/batman_adv/throughput_override
0.0 MBit
root@aw-glinet-b1300:/sys# batctl n
[B.A.T.M.A.N. adv openwrt-2019.2-4, MainIF/MAC: primary0/52:98:0e:9b:5f:9b (bat0/e4:95:6e:45:1f:f6 BATMAN_V)]
IF             Neighbor              last-seen
6e:5c:cc:06:cc:e4    0.140s (     1000.0) [vx_mesh_lan]
c2:66:1a:b4:55:a8    0.010s (     1000.0) [vx_mesh_lan]
b2:2d:f8:13:64:27    0.280s (     1000.0) [vx_mesh_lan]
6e:5c:cc:06:cc:e1    0.500s (       48.0) [     mesh0]
6e:5c:cc:06:cc:e5    0.230s (       63.9) [     mesh1]
```

```
root@aw-fritzbox-4040:/sys# cat ./devices/virtual/net/vx_mesh_lan/batman_adv/throughput_override
0.0 MBit
root@aw-fritzbox-4040:/sys# batctl n
[B.A.T.M.A.N. adv openwrt-2019.2-4, MainIF/MAC: primary0/6e:5c:cc:06:cc:e3 (bat0/98:9b:cb:e2:53:d8 BATMAN_V)]
IF             Neighbor              last-seen
52:98:0e:9b:5f:9c    0.130s (     1000.0) [vx_mesh_lan]
c2:66:1a:b4:55:a8    0.380s (     1000.0) [vx_mesh_lan]
b2:2d:f8:13:64:27    0.410s (     1000.0) [vx_mesh_lan]
52:98:0e:9b:5f:99    0.250s (       47.6) [     mesh0]
52:98:0e:9b:5f:9d    0.360s (       63.9) [     mesh1]
```

Now it looks good on both nodes :).

awlx:
Only flaw is, virtualised offloaders display 1.0 as throughput values:
```
root@ffplieninggbwkvmol:~# batctl n
[B.A.T.M.A.N. adv openwrt-2019.2-4, MainIF/MAC: primary0/66:13:7b:9d:b2:1b (bat0/52:54:00:9e:b3:58 BATMAN_V)]
IF             Neighbor              last-seen
8e:89:98:45:6c:70    0.150s (        1.0) [vx_mesh_lan]
a2:ca:66:65:30:40    0.480s (        1.0) [vx_mesh_lan]
9a:9b:66:30:d9:38    0.290s (        1.0) [vx_mesh_lan]
f2:00:29:03:00:00    0.250s (       10.0) [  mesh-vpn]
```

awlx:
Is it possible that it doesn't work for `vx_mesh_wan` interfaces? It seems CPE 210 and so on fallback to 1.0 throughput.
http://[2001:608:a01:10a:62e3:27ff:feb2:dc4e]/cgi-bin/status

mweinelt:
This seems to be because virtual ethernet interfaces have no link speed. Tested with `virtio-net-pci` devices and this patchset:

```
# ethtool eth0
Settings for eth0:
	Supported ports: [ ]
	Supported link modes:   Not reported
	Supported pause frame use: No
	Supports auto-negotiation: No
	Supported FEC modes: Not reported
	Advertised link modes:  Not reported
	Advertised pause frame use: No
	Advertised auto-negotiation: No
	Advertised FEC modes: Not reported
	Speed: Unknown!
	Duplex: Unknown! (255)
	Port: Other
	PHYAD: 0
	Transceiver: internal
	Auto-negotiation: off
	Link detected: yes
```
But virtio links with vhost-net can easily exceed 1 Gbps, so while the low throughput value isn't an issue when your APs only have this single link through this node, but as soon as they see another (shitty) link they will too easily switch to that.

awlx:
Yep, as long as there is only a single "uplink" node it's not a problem, but if we see another node via wifi with let's say 2Mbit/s throughput, all the traffic will switch to it which is horrible.

NeoRaider:
All in all, my VXLAN and bridge patches can only improve the situation in some setups - and as discussed with Marek on IRC, they may lead to new issues in others. It seems the only robust fix would be to finalize the patches for the tp-meter-based metric and stop relying on ethtool completely.

mweinelt:
Also assuming that the VXLAN throughput equals that of its lower link is flawed. But probably in regions that don't matter too much right now.

TP-Link Archer C7 v2 … raw throughput ~700 Mbit/s, vxlan troughput ~200 Mbit/s
Zyxel NBG6617 … raw throughput ~900 Mbit/s, vxlan throughput ~650 Mbit/s

awlx:
Problem will be, that even if the vxlan throughput is only 200mbit/s or below ... I still want to prefer it over any wireless connection.

So I see many ending up just hardcoding the wired connections to insane highspeed to always prefer them.

rotanid:
how do we proceed with this PR @awlx @NeoRaider @mweinelt  ?

awlx:
Well, in the current state it unfortunately makes some stuff better but we will still rely on the our manuel patches, because it caused too many not working scenarios so far.

mweinelt:
> All in all, my VXLAN and bridge patches can only improve the situation in some setups - and as discussed with Marek on IRC, they may lead to new issues in others. It seems the only robust fix would be to finalize the patches for the tp-meter-based metric and stop relying on ethtool completely.

neoraider in https://github.com/freifunk-gluon/gluon/pull/1872#issuecomment-562886514

Closing.

PR #1873, opened 2019-11-20
ath79-generic: add support for devolo WiFi pro 1750i

herrbett:
  - [ ] webinterface
  - [ ] tftp
  - [X] other: sysupgrade via SSH on the vendor firmware
`devolo-wifi-pro-1750i`
  - All MAC addresses are printed on the housing, it is identical to the MAC address of the 5Ghz Wifi.
- wired network
- wifi (if applicable)
- led mapping
  - radio leds
    - NO radio leds
  - switchport leds
    -  NO switchport leds

### DIFF
--- a/targets/ath79-generic
+++ b/targets/ath79-generic
@@ -11,6 +11,11 @@ device('devolo-wifi-pro-1200i', 'devolo_dvl1200i', {
 	factory = false,
 })
 
+device('devolo-wifi-pro-1750i', 'devolo_dvl1750i', {
+	packages = ATH10K_PACKAGES_QCA9880,
+	factory = false,
+})
+
 -- OCEDO
 
 device('ocedo-raccoon', 'ocedo_raccoon', {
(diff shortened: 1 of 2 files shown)


targets/ath79-generic
```diff
@@ -10,6 +10,9 @@ local ATH10K_PACKAGES_QCA9888 = {
 device('devolo-wifi-pro-1200i', 'devolo_dvl1200i', {
 	factory = false,
 })
+device('devolo-wifi-pro-1750i', 'devolo_dvl1750i', {
```
blocktrron: The seperating newline is missig here

targets/ath79-generic
```diff
@@ -10,6 +10,9 @@ local ATH10K_PACKAGES_QCA9888 = {
...
+	factory = false,
```
blocktrron: ath10k package reference is missing here - See https://github.com/freifunk-gluon/gluon/commit/e30b0092bb00ea51d3d08838b6ab0c401095fc3a

PR #1874, opened 2019-11-20
docs: add a note that hopglass-server needs an update to work with the new respondd address

RalfJung:
This remark would have saved us an hour or two of wondering why things do not work any more with the new address.

### DIFF
--- a/docs/releases/v2019.1.rst
+++ b/docs/releases/v2019.1.rst
@@ -15,6 +15,7 @@ possible.
 With Gluon v2019.1, nodes will not answer respondd queries on ``[ff02::2:1001]:1001`` anymore. Respondd
 querier setups still using this address must be updated to the new address ``[ff05::2:1001]:1001``
 (supported since Gluon v2017.1). This change was required due to cross-domain leakage of respondd data.
+If you are using hopglass-server to query respondd data, you need to update it to at least commit f0e2c0a5.
 
 If you are upgrading from a version prior to v2018.1, please note that the flash layout on some
 devices (TP-Link CPE/WBS 210/510) was changed. To avoid upgrade failures, make sure to upgrade


blocktrron:
Please wrap your commit message - also add a `docs:` prefix to your commit title.

Is there a specific revision from which on Hopglass Server works with the new address? This could be linked to.

docs/releases/v2019.1.rst
```diff
@@ -15,6 +15,7 @@ possible.
 With Gluon v2019.1, nodes will not answer respondd queries on ``[ff02::2:1001]:1001`` anymore. Respondd
 querier setups still using this address must be updated to the new address ``[ff05::2:1001]:1001``
 (supported since Gluon v2017.1). This change was required due to cross-domain leakage of respondd data.
+(hopglass-server needs an update to work with the new address, so make sure you use the latest version.)
```
mweinelt: If you are using hopglass to query respondd data you need to update hopglass to at least $VERSION.

Don't use brackets.

RalfJung:
I think I resolved all your comments.

blocktrron:
@RalfJung the commit title is still to long

RalfJung:
I wrapped it to 80 columns, what is the problem?

docs/releases/v2019.1.rst
```diff
@@ -15,6 +15,7 @@ possible.
 With Gluon v2019.1, nodes will not answer respondd queries on ``[ff02::2:1001]:1001`` anymore. Respondd
 querier setups still using this address must be updated to the new address ``[ff05::2:1001]:1001``
 (supported since Gluon v2017.1). This change was required due to cross-domain leakage of respondd data.
+If you are using hopglass-server to query respondd data, you need to update it to at least commit f0e2c0a5.
```
rubo77: there is no such commit f0e2c0a5

RalfJung: > there is no such commit f0e2c0a5

Yes there is: https://github.com/hopglass/hopglass-server/commit/f0e2c0a58b8947176f7c87680c2bcb27f302f14c

> github history only shows 7 characters

So what?

mweinelt: @petabyteboy Can you maybe create a tag including the fix?

petabyteboy: Yes, I will create a 1.0.0 tag including this commit soon.

RalfJung:
12 days later... can we land this without waiting for a hopglass release? We can always adjust the docs later when a release happens.

I can adjust the short-commit to 7 characters if you'd prefer that.

petabyteboy:
It turned out to be very difficult to reproduce this success, and I still have not seen a hopglass setup that works with Gluon 2019 nodes without patching around on the node. Because multiple users of the current master were still having unsolved problems I did not create the new release yet.

kpanic23:
TBH, I don't know why there should be a note in the gluon documentation. That note belongs in the hopglass repository.
@petabyteboy, We have hopglass-server running with Gluon 2019.1: https://map.ff3l.net

RalfJung:
I think the Gluon changelog should mention this as that's what people will read when bumping the firmware of their community. When you raise the minimum version you require some other software to have for smooth collaboration, that should be reflected in the changelog.

docs/releases/v2019.1.rst
mweinelt: @petabyteboy Ping.

RalfJung:
FWIW, besides patching hopglass-server, we also had to patch node-respondd to make gateways appear properly in hopglass again: https://github.com/hopglass/node-respondd/pull/2 (unfortunately, upstream seems rather dead these days)

PR #1876, opened 2019-11-23
gluon-private-wifi: add support for WPA3 encryption

### LINKED ISSUE #1826: Private WLAN should support WPA3
OpenWrt 19.07 has support for WPA3 and we should upgrade `web-private-wifi` to configure the private wifi in WPA2/WPA3 mixed mode.

- Upside is more secure connections for devices that support it.
- Downside is that we would need to bundle a cryptographic library.
  - This can therefore only be made available to devices with `>=8MB` of flash.

blocktrron:
This commit adds WPA3 encryption for Gluon's Private WiFi feature.

It allows node operators to select between three encryption modes:
 - WPA2 (old standard - no MFP)
 - WPA2/WPA3 mixed-mode (optional MFP)
 - WPA3 (mandatory MFP)

MFP is supported by all drivers which are currently fully supported by Gluon (rt2800, mt76, ath9k, ath10k).

As we need to include a OpenSSL library for WPA3, this PR also includes a mechanic to define a featureset. This way, packages can be included at compile-time as well as the featureset retrieved at run-time to decide, which functions can be supported in a devices featureset.

This Pull-Request still needs documentation written, so It's labeled as RFC.

fixes #1826

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,22 +112,30 @@ lint-sh: FORCE
 	@scripts/lint-sh.sh
 
 GLUON_DEFAULT_PACKAGES := hostapd-mini
-GLUON_CLASS_PACKAGES_standard :=
-GLUON_CLASS_PACKAGES_tiny :=
-
-GLUON_FEATURE_PACKAGES := $(shell scripts/features.sh '$(GLUON_FEATURES)' || echo '__ERROR__')
-ifneq ($(filter __ERROR__,$(GLUON_FEATURE_PACKAGES)),)
-$(error Error while evaluating GLUON_FEATURES)
-endif
-
-
-GLUON_PACKAGES :=
-define merge_packages
-  $(foreach pkg,$(1),
-    GLUON_PACKAGES := $$(strip $$(filter-out -$$(patsubst -%,%,$(pkg)) $$(patsubst -%,%,$(pkg)),$$(GLUON_PACKAGES)) $(pkg))
+define merge_lists
+  $(1) :=
+  $(foreach pkg,$(2),
+    $(1) := $$(strip $$(filter-out -$$(patsubst -%,%,$(pkg)) $$(patsubst -%,%,$(pkg)),$$(value $(1))) $(pkg))
   )
 endef
-$(eval $(call merge_packages,$(GLUON_DEFAULT_PACKAGES) $(GLUON_FEATURE_PACKAGES) $(GLUON_SITE_PACKAGES)))
+
+define feature_packages
+  $(1) := $(shell scripts/features.sh '$(2)' || echo '__ERROR__')
+endef
+
+$(eval $(call merge_lists,GLUON_FEATURE_LIST_standard,$(GLUON_FEATURES) $(GLUON_FEATURES_standard)))
+$(eval $(call merge_lists,GLUON_FEATURE_LIST_tiny,$(GLUON_FEATURES) $(GLUON_FEATURES_tiny)))
+
+$(eval $(call feature_packages,GLUON_FEATURE_PACKAGES_standard,$(GLUON_FEATURE_LIST_standard)))
+$(eval $(call feature_packages,GLUON_FEATURE_PACKAGES_tiny,$(GLUON_FEATURE_LIST_tiny)))
+
+ifneq ($(filter __ERROR__,$(GLUON_FEATURES_standard) $(GLUON_FEATURES_tiny)),)
+  $(error Error while evaluating features)
+endif
+
+$(eval $(call merge_lists,GLUON_DEFAULT_PACKAGES,$(GLUON_DEFAULT_PACKAGES) $(GLUON_SITE_PACKAGES)))
+$(eval $(call merge_lists,GLUON_CLASS_PACKAGES_standard,$(GLUON_FEATURE_PACKAGES_standard)))
+$(eval $(call merge_lists,GLUON_CLASS_PACKAGES_tiny,$(GLUON_FEATURE_PACKAGES_tiny)))
 
 
 LUA := openwrt/staging_dir/hostpkg/bin/lua
--- a/Makefile
+++ b/Makefile
@@ -133,8 +133,8 @@ ifneq ($(filter __ERROR__,$(GLUON_FEATURES_standard) $(GLUON_FEATURES_tiny)),)
 endif
 
 $(eval $(call merge_lists,GLUON_DEFAULT_PACKAGES,$(GLUON_DEFAULT_PACKAGES) $(GLUON_SITE_PACKAGES)))
-$(eval $(call merge_lists,GLUON_CLASS_PACKAGES_standard,$(GLUON_FEATURE_PACKAGES_standard)))
-$(eval $(call merge_lists,GLUON_CLASS_PACKAGES_tiny,$(GLUON_FEATURE_PACKAGES_tiny)))
+$(eval $(call merge_lists,GLUON_CLASS_PACKAGES_standard,$(GLUON_FEATURE_PACKAGES_standard) $(GLUON_SITE_PACKAGES_standard)))
+$(eval $(call merge_lists,GLUON_CLASS_PACKAGES_tiny,$(GLUON_FEATURE_PACKAGES_tiny) $(GLUON_SITE_PACKAGES_tiny)))
 
 
 LUA := openwrt/staging_dir/hostpkg/bin/lua
--- a/Makefile
+++ b/Makefile
@@ -112,6 +112,8 @@ lint-sh: FORCE
 	@scripts/lint-sh.sh
 
 GLUON_DEFAULT_PACKAGES := hostapd-mini
+GLUON_CLASS_PACKAGES_standard :=
+GLUON_CLASS_PACKAGES_tiny :=
 
 GLUON_FEATURE_PACKAGES := $(shell scripts/features.sh '$(GLUON_FEATURES)' || echo '__ERROR__')
 ifneq ($(filter __ERROR__,$(GLUON_FEATURE_PACKAGES)),)
@@ -144,12 +146,12 @@ config: $(LUA) FORCE
 	$(foreach conf,site $(patsubst $(GLUON_SITEDIR)/%.conf,%,$(wildcard $(GLUON_SITEDIR)/domains/*.conf)),$(call CheckSite,$(conf)))
 
 	@$(GLUON_CONFIG_VARS) \
-		$(LUA) scripts/target_config.lua '$(GLUON_TARGET)' '$(GLUON_PACKAGES)' \
+		$(LUA) scripts/target_config.lua '$(GLUON_TARGET)' '$(GLUON_DEFAULT_PACKAGES)' '$(GLUON_CLASS_PACKAGES_standard)' '$(GLUON_CLASS_PACKAGES_tiny)' \
 		> openwrt/.config
 	+@$(OPENWRTMAKE) defconfig
 
 	@$(GLUON_CONFIG_VARS) \
-		$(LUA) scripts/target_config_check.lua '$(GLUON_TARGET)' '$(GLUON_PACKAGES)'
+		$(LUA) scripts/target_config_check.lua '$(GLUON_TARGET)' '$(GLUON_DEFAULT_PACKAGES)' '$(GLUON_CLASS_PACKAGES_standard)' '$(GLUON_CLASS_PACKAGES_tiny)'
 
 
 all: config
--- a/Makefile
+++ b/Makefile
@@ -111,7 +111,6 @@ lint-lua: FORCE
 lint-sh: FORCE
 	@scripts/lint-sh.sh
 
-GLUON_DEFAULT_PACKAGES := hostapd-mini
 define merge_lists
   $(1) :=
   $(foreach pkg,$(2),
--- a/docs/dev/hardware.rst
+++ b/docs/dev/hardware.rst
@@ -9,6 +9,18 @@ Having an ath9k, ath10k or mt76 based WLAN adapter is highly recommended,
 although other chipsets may also work. VAP (multiple SSID) support
 is a requirement.
 
+.. _device-class-definition:
+
+Device classes
+--------------
+Gluon currently is aware of two device classes. Depending on the device class, different
+features can be installed onto the device.
+
+The ``tiny`` device-class contains devices with the following limitations:
+
+* All devices with less than 64 MB of system memory
+* All devices with less than 7 MB of usable firmware space
+* Devices using a single ath10k radio and less than 128MB of system memory
 
 .. _hardware-adding-profiles:
 
--- a/docs/features/private-wlan.rst
+++ b/docs/features/private-wlan.rst
@@ -4,6 +4,14 @@ Private WLAN
 It is possible to set up a private WLAN that bridges the WAN port and is separated from the mesh network.
 Please note that you should not enable ``mesh_on_wan`` simultaneously.
 
+The private WLAN is encrypted using WPA2 by default. On devices with enough flash and a supported radio,
+WPA3 or WPA2/WPA3 mixed-mode can be used instead of WPA2. For this to work, the ``wireless-encryption-wpa3``
+feature has to be added to ``GLUON_FEATURES``.
+
+It is recommended to enable IEEE 802.11w management frame protection for WPA2/WPA3 networks, however this
+can lead to connectivity problems for older clients. In this case, management frame protection can be
+made optional or completely disabled in the advanced settings tab.
+
 The private WLAN can be enabled through the config mode if the package ``gluon-web-private-wifi`` is installed.
 You may also enable a private WLAN using the command line::
 
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -545,6 +545,14 @@ GLUON_SITE_PACKAGES
     default package set. It is also possible to remove packages from the
     default set by prepending a minus sign to the package name.
 
+GLUON_SITE_PACKAGES_standard
+    Defines a list of additional packages to include or exclude for devices of
+    the standard device-class.
+
+GLUON_SITE_PACKAGES_tiny
+    Defines a list of additional packages to include or exclude for devices of
+    the tiny device-class.
+
 GLUON_RELEASE
     The current release version Gluon should use.
 
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -527,8 +527,18 @@ GLUON_DEPRECATED
     deployments of low-flash devices are required).
 
 GLUON_FEATURES
-    Defines a list of features to include. The feature list is used to generate
-    the default package set.
+    Defines a list of features to include. Depending on the device, the feature list
+    defined from this value is combined with the feature list for either the standard
+    or the tiny device-class. The resulting feature list is used to generate the default
+    package set.
+
+GLUON_FEATURES_standard
+    Defines a list of additional features to include or exclude for devices of
+    the standard device-class.
+
+GLUON_FEATURES_tiny
+    Defines a list of additional features to include or exclude for devices of
+    the tiny device-class.
 
 GLUON_SITE_PACKAGES
     Defines a list of packages which should be installed in addition to the
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -10,9 +10,9 @@ ar71xx-generic
 
 * ALFA Network
 
-  - AP121 [#deprecated]_
+  - AP121 [#deprecated]_  [#device-class-tiny]_
   - AP121F
-  - AP121U [#deprecated]_
+  - AP121U [#deprecated]_  [#device-class-tiny]_
 
 * Allnet
 
@@ -47,7 +47,7 @@ ar71xx-generic
 
 * Linksys
 
-  - WRT160NL
+  - WRT160NL  [#device-class-tiny]_
 
 * Netgear
 
@@ -82,7 +82,7 @@ ar71xx-generic
   - CPE220 (v1.1)
   - CPE510 (v1.0, v1.1)
   - CPE520 (v1.1)
-  - RE450 (v1)
+  - RE450 (v1)  [#device-class-tiny]_
   - TD-W8970 (v1) [#lan_as_wan]_
   - TL-WDR3500 (v1)
   - TL-WDR3600 (v1)
@@ -97,16 +97,16 @@ ar71xx-generic
 
 * Ubiquiti
 
-  - Air Gateway
-  - Air Gateway LR
-  - Air Gateway PRO
-  - Air Router
-  - Bullet M2/M5
-  - Loco M2/M5
+  - Air Gateway  [#device-class-tiny]_
+  - Air Gateway LR  [#device-class-tiny]_
+  - Air Gateway PRO  [#device-class-tiny]_
+  - Air Router  [#device-class-tiny]_
+  - Bullet M2/M5  [#device-class-tiny]_
+  - Loco M2/M5  [#device-class-tiny]_
   - Loco M2/M5 XW
-  - Nanostation M2/M5
+  - Nanostation M2/M5  [#device-class-tiny]_
   - Nanostation M2/M5 XW
-  - Picostation M2
+  - Picostation M2  [#device-class-tiny]_
   - Rocket M2/M5
   - Rocket M2/M5 Ti
   - Rocket M2/M5 XW
@@ -146,8 +146,8 @@ ar71xx-nand
 
   - NBG6716
 
-ar71xx-tiny [#deprecated]_
---------------------------
+ar71xx-tiny [#deprecated]_ [#device-class-tiny]_
+------------------------------------------------
 
 * D-Link
 
@@ -245,7 +245,7 @@ ipq40xx-generic
 * ZyXEL
 
   - NBG6617
-  - WRE6606
+  - WRE6606  [#device-class-tiny]_
 
 ipq806x-generic
 ---------------
@@ -372,8 +372,8 @@ ramips-mt76x8
 
   - VoCore2
 
-ramips-rt305x [#deprecated]_
-----------------------------
+ramips-rt305x [#deprecated]_  [#device-class-tiny]_
+---------------------------------------------------
 
 * A5-V11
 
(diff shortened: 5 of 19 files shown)


blocktrron:
As a sidenote: We've seen issues connecting with a MFP enabled Access Point (WPA2/WPA3) on a 2012 Apple MacBook Pro. So we might want to also allow configuration of 802.11w instead of doing hard defaults here.

scripts/target_lib.lua
```diff
@@ -44,6 +45,7 @@ local default_options = {
 	aliases = {},
 	manifest_aliases = {},
 	packages = {},
+	featureset = false,
```
NeoRaider: You could just define `featureset = "default"` here and undo all other changes in this file. Rather than adding a separate top-level function `featureset` for the defaults, this would be handled by the existing `defaults` function.

package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua
```diff
@@ -52,4 +52,8 @@ function M.is_outdoor_device()
 	return false
 end
 
+function M.get_featureset()
+	return M.readfile("/lib/gluon/featureset")
```
NeoRaider: The return value should be stripped to get rid of trailing whitespace.

package/gluon-web-private-wifi/i18n/de.po
```diff
@@ -16,6 +16,15 @@ msgstr "8-63 Zeichen"
...
 
+msgid "Disabled"
+msgstr "Deaktiviert"
+
+msgid "Optional"
+msgstr "Optional"
+
+msgid "Required"
+msgstr "Erforderlich"
+
```
NeoRaider: - `.pot` file should be updated using `contrib/i18n-scan.pl`
- All `.po` files should be updated using `msgmerge`

(see i18n docs)

docs/features/private-wlan.rst
```diff
@@ -4,6 +4,13 @@ Private WLAN
 It is possible to set up a private WLAN that bridges the WAN port and is separated from the mesh network.
 Please note that you should not enable ``mesh_on_wan`` simultaneously.
 
+The private WiFi is by default encrypted via WPA2. On devices with enough flash and a supported radio,
```
NeoRaider: Weird word order, better: "The private Wifi is encrypted via WPA2 by default."

docs/features/private-wlan.rst
```diff
@@ -4,6 +4,13 @@ Private WLAN
...
+WPA3 or WPA2/WPA3 mixed-mode can be used instead of WPA2.
+
+Additionally, IEEE 802.11w mangagement frame protection should be enforced for WPA2/WPA3, however this
+can lead to connectivity problems for older clients. In this case, mangagement frame protection can be
+set to optional or disabled completely using the config-mode.
```
NeoRaider: We usually refer to the web UI as "configuration wizard" or "advanced settings" instead of using the more technical term "config-mode" in the docs.

docs/user/supported_devices.rst
```diff
@@ -400,6 +400,10 @@ Footnotes
   The device or target is reaching its end of life soon. This means that support
   in the next major release of Gluon is doubtful.
 
+.. [#featureset-tiny]
+  These devices do not support the full featureset of Gluon. Namely, WPA3 support
+  is missing on these devices.
```
NeoRaider: Maybe we should mention that this is only relevant for the private-wifi feature at the moment.

blocktrron:
I've made the changes requested by @NeoRaider and also made the availability of WPA3 independent from the MFP support, as the support is now user-selectable anyhow.

rotanid:
maybe we should split up the featureset introduction in a separate PR?

blocktrron:
While this PR is mostly about the featuresets, i do not think that it is useful, as the usecase is directly connected with WPA3 support of any kind (OWE / SAE).

docs/features/private-wlan.rst
```diff
@@ -4,6 +4,13 @@ Private WLAN
 It is possible to set up a private WLAN that bridges the WAN port and is separated from the mesh network.
 Please note that you should not enable ``mesh_on_wan`` simultaneously.
 
+The private WiFi is encrypted via WPA2 by default. On devices with enough flash and a supported radio,
+WPA3 or WPA2/WPA3 mixed-mode can be used instead of WPA2.
+
+Additionally, IEEE 802.11w mangagement frame protection should be enforced for WPA2/WPA3, however this
+can lead to connectivity problems for older clients. In this case, mangagement frame protection can be
```
rotanid: two times: "mangagement" -> "management"

docs/user/supported_devices.rst
```diff
@@ -400,6 +400,10 @@ Footnotes
   The device or target is reaching its end of life soon. This means that support
   in the next major release of Gluon is doubtful.
 
+.. [#featureset-tiny]
+  These devices do not support the full featureset of Gluon. Namely, WPA3 support
+  for the Private-WiFi is missing on these devices.
```
rotanid: why using the "WiFi" trademark here?

docs/features/private-wlan.rst
```diff
@@ -4,6 +4,13 @@ Private WLAN
 It is possible to set up a private WLAN that bridges the WAN port and is separated from the mesh network.
 Please note that you should not enable ``mesh_on_wan`` simultaneously.
 
+The private WiFi is encrypted via WPA2 by default. On devices with enough flash and a supported radio,
```
rotanid: WiFi -> WLAN

package/gluon-web-private-wifi/i18n/de.po
```diff
@@ -1,38 +1,45 @@
...
 "Your node can additionally extend your private network by bridging the WAN "
 "interface with a separate WLAN. This feature is completely independent of "
 "the mesh functionality. Please note that the private WLAN and meshing on the "
 "WAN interface should not be enabled at the same time."
 msgstr ""
-"Dein Knoten kann zusätzlich die Reichweite deines privaten Netzes erweitern. "
-"Hierfür wird der WAN-Port mit einem separaten WLAN gebridged. Diese "
-"Funktionalität ist völlig unabhängig von den Mesh-Funktionen des Knotens. "
-"Beachte, dass du nicht gleichzeitig das Meshen über den WAN-Port aktiviert "
-"haben solltest."
```
rotanid: why is this section being removed?

rotanid:
as WiFi is a trademark, i suggest to use WLAN instead where possible, but that's just my opinion

package/gluon-web-private-wifi/i18n/de.po
blocktrron: Because I'm to stupid to use tools correctly. Fixed now.

docs/features/private-wlan.rst
blocktrron: DOne

docs/user/supported_devices.rst
blocktrron: Fixed

docs/features/private-wlan.rst
blocktrron: Fixed.

blocktrron:
@rotanid @NeoRaider Everything is (hopefully) addressed now.

package/gluon-wireless-encryption/Makefile
```diff
@@ -0,0 +1,33 @@
...
+
+PKG_NAME:=gluon-wireless-encryption
+PKG_VERSION:=1
+PKG_RELEASE:=1
+
+include ../gluon.mk
+
+define Package/gluon-wireless-encryption-wpa2
+  DEPENDS:=+hostapd-mini
+  TITLE:=Package for supporting WPA2 encrypted WiFi networks
```
rotanid: WiFi -> WLAN

package/gluon-wireless-encryption/Makefile
```diff
@@ -0,0 +1,33 @@
...
+endef
+
+define Package/gluon-wireless-encryption-wpa3
+  DEPENDS:=+hostapd-openssl
+  TITLE:=Package for supporting WPA3 encrypted WiFi networks
```
rotanid: WiFi -> WLAN

NeoRaider:
> it's not the scope of this PR, but we could think about renaming gluon-web-private-wifi to gluon-web-private-wlan

Or maybe *private-ap*?

rotanid:
> > it's not the scope of this PR, but we could think about renaming gluon-web-private-wifi to gluon-web-private-wlan
> 
> Or maybe _private-ap_?

also fine with me, but not wifi

package/gluon-wireless-encryption/Makefile
blocktrron: fixed

package/gluon-wireless-encryption/Makefile
blocktrron: fixed

blocktrron:
Fixed the two leftovers @rotanid pointed out and rebased.

The renaming of the `private-wifi` pacage is a good idea, but let's do this after this feature is finished ;)

Makefile
```diff
@@ -111,21 +111,30 @@ lint-lua: FORCE
...
-$(error Error while evaluating GLUON_FEATURES)
-endif
+GLUON_FEATURESET_STANDARD_FEATURE_PACKAGES := 
+GLUON_FEATURESET_TINY_FEATURE_PACKAGES := 
 
+define feature_packages
+  $(1) := $(shell scripts/features.sh '$(2)' || echo '__ERROR__')
+endef
+$(eval $(call feature_packages, GLUON_FEATURESET_STANDARD_FEATURE_PACKAGES, $(GLUON_FEATURES) $(GLUON_FEATURESET_STANDARD_FEATURES)))
+$(eval $(call feature_packages, GLUON_FEATURESET_TINY_FEATURE_PACKAGES, $(GLUON_FEATURES) $(GLUON_FEATURESET_TINY_FEATURES)))
```
NeoRaider: I think it would make sense to reuse `merge_packages` here (renamed to `merge_lists`), so the featureset featurelists can contain negative entries.

.luacheckrc
```diff
@@ -95,6 +95,7 @@ files["targets/*"] = {
 		"exec_capture_raw",
 		"exec_raw",
 		"factory_image",
+		"featureset",
```
NeoRaider: I think this can be dropped?

scripts/target_lib.lua
```diff
@@ -44,6 +44,7 @@ local default_options = {
 	aliases = {},
 	manifest_aliases = {},
 	packages = {},
+	featureset = "standard",
```
NeoRaider: Hmm, I wonder if "featureset" is really a good name for this - might get confusing, as we already have "features". And by itself, it categorizes devices rather than features - maybe something like "device class" (or just "class" in the target definitions) would be more appropriate?

docs/user/site.rst
```diff
@@ -527,8 +527,18 @@ GLUON_DEPRECATED
...
+    defined from this value is combined with the feature list for either the standard
+    or the tiny featureset. The resulting feature list is used to generate the default
+    package set.
+
+GLUON_FEATURESET_STANDARD_FEATURES
+    Defines a list of additional features to include for devices with the standard
+    featureset.
+
+GLUON_FEATURESET_TINY_FEATURES
+    Defines a list of additional features to include for devices with the standard
```
NeoRaider: "standard" -> "tiny"

docs/user/supported_devices.rst
```diff
@@ -414,6 +414,10 @@ Footnotes
   The device or target is reaching its end of life soon. This means that support
   in the next major release of Gluon is doubtful.
 
+.. [#featureset-tiny]
+  These devices do not support the full featureset of Gluon. Namely, WPA3 support
+  for the private WLAN is missing on these devices.
```
NeoRaider: Needs better explanation; should refer to the part of `site.rst` where setting the features for the featuresets is described.

NeoRaider: We also need to document the criteria for setting the featuresets of devices somewhere.

As this PR is about adding WPA3 and the size of the OpenSSL libraries, I'd expect "tiny" to be about the flash size only, but somehow in the target definitions below you also set `tiny` for 8/32?

blocktrron:
@NeoRaider Hopefully I've catches everything. Also implemented device-class based site packages.

blocktrron:
rebased

package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua
```diff
@@ -52,4 +53,29 @@ function M.is_outdoor_device()
 	return false
 end
 
+function M.supports_wpa3()
+	return unistd.access('/lib/gluon/features/wpa3')
```
NeoRaider: Commit message needs to be updated.

package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua
```diff
@@ -52,4 +53,29 @@ function M.is_outdoor_device()
...
+	local supports_mfp = true
+
+	if not M.supports_wpa3() then
+		return false
+	end
+
+	uci:foreach('wireless', 'wifi-device', function()
+		local phypath = '/sys/kernel/debug/ieee80211/phy' .. idx .. '/'
+
+		if not util.file_contains_line(phypath .. 'hwflags', 'MFP_CAPABLE') then
```
NeoRaider: Ugh, it sucks that there is no robust way to determine this without debugfs ...

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -35,6 +35,21 @@ function M.contains(table, value)
 	return false
 end
 
+function M.file_contains_line(path, value)
+	local file_content = M.readfile(path)
```
NeoRaider: - Use `io.lines()`: http://www.lua.org/manual/5.1/manual.html#pdf-io.lines
- Raise an error when the file can't be read (not sure if `io.lines()` already does that)

docs/features/private-wlan.rst
```diff
@@ -4,6 +4,14 @@ Private WLAN
 It is possible to set up a private WLAN that bridges the WAN port and is separated from the mesh network.
 Please note that you should not enable ``mesh_on_wan`` simultaneously.
 
+The private WLAN is encrypted through WPA2 by default. On devices with enough flash and a supported radio,
```
NeoRaider: Better: "through" -> "using"

package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua
NeoRaider: Also, there is a typo in the title.

docs/features/private-wlan.rst
```diff
@@ -4,6 +4,14 @@ Private WLAN
...
+WPA3 or WPA2/WPA3 mixed-mode can be used instead of WPA2. For this to work, the ``wireless-encryption-wpa3``
+feature has to be added to ``GLUON_FEATURES``.
+
+Additionally, IEEE 802.11w management frame protection should be enforced for WPA2/WPA3, however this
+can lead to connectivity problems for older clients. In this case, management frame protection can be
+made optional or completely disabled in the advanced settings tab.
+
```
NeoRaider: Please add `:` after "docs" in the commit title.

package/gluon-wireless-encryption/Makefile
```diff
@@ -0,0 +1,33 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-wireless-encryption
+PKG_VERSION:=1
```
NeoRaider: Drop `PKG_VERSION`, `PKG_RELEASE` is sufficient.

Makefile
```diff
@@ -111,7 +111,7 @@ lint-lua: FORCE
 lint-sh: FORCE
 	@scripts/lint-sh.sh
 
-GLUON_DEFAULT_PACKAGES := hostapd-mini
+GLUON_DEFAULT_PACKAGES :=
```
NeoRaider: The variable `GLUON_DEFAULT_PACKAGES` and the `merge_lists` using it can be removed as well.

package/gluon-wireless-encryption/Makefile
```diff
@@ -0,0 +1,33 @@
...
+
+define Package/gluon-wireless-encryption-wpa3
+  DEPENDS:=+hostapd-openssl
+  TITLE:=Package for supporting WPA3 encrypted wireless networks
+endef
+
+define Package/gluon-wireless-encryption-wpa2/install
+	$(INSTALL_DIR) $(1)/lib/gluon/features
+
+	touch $(1)/lib/gluon/features/wpa2
```
NeoRaider: Having a `wpa2` feature at all does not make sense to me; making it a part of the `wpa3` as well package even less.

Let's drop the flag and the package, and directly reference `hostapd-mini` in `package/features`.

Makefile
```diff
@@ -115,19 +115,32 @@ GLUON_DEFAULT_PACKAGES := hostapd-mini
...
+$(eval $(call feature_packages, GLUON_FEATURE_PACKAGES_standard, $(GLUON_FEATURES_standard)))
+$(eval $(call feature_packages, GLUON_FEATURE_PACKAGES_tiny, $(GLUON_FEATURES_tiny)))
+
+ifneq ($(filter __ERROR__,$(GLUON_FEATURES_standard) $(GLUON_FEATURES_tiny)),)
+  $(error Error while evaluating features)
+endif
+
+$(eval $(call merge_lists, GLUON_DEFAULT_PACKAGES, $(GLUON_DEFAULT_PACKAGES) $(GLUON_SITE_PACKAGES)))
+$(eval $(call merge_lists, GLUON_CLASS_PACKAGES_standard, $(GLUON_CLASS_PACKAGES_standard) $(GLUON_FEATURE_PACKAGES_standard)))
+$(eval $(call merge_lists, GLUON_CLASS_PACKAGES_tiny, $(GLUON_CLASS_PACKAGES_tiny) $(GLUON_FEATURE_PACKAGES_tiny)))
```
NeoRaider: - You can't use the same variable name as the target of `merge_lists` and as an input argument.
- Rather than initializing the different target variables as empty explicitly, this could be done in `merge_lists`.
- Remove all spaces around `,` in `$(call ...)` - probably not an issue in this case, but the spaces will be passed as parts of the arguments

scripts/target_config_lib.lua
```diff
@@ -7,6 +7,8 @@ return function(funcs)
 
 	local target = arg[1]
 	local extra_packages = arg[2]
+	local class_packages_standard = arg[3] or ''
+	local class_packages_tiny = arg[4] or ''
```
NeoRaider: Make this a table:
```lua
local class_packages = {
  standard = arg[3],
  tiny = arg[4],
}
```
Then the code below can be simplified to
```lua
handle_pkgs(class_packages[dev.options.class])
```
by moving the ` or ''` and `string.gmatch()` into `handle_pkgs` as well.

NeoRaider: It would be great it we could pass these lists in the environment instead of using separate arguments, but all ideas I have either limit future class names to characters that are valid in shell variable identifiers (disallowing `-` for example) or lead to more Makefile ugliness when generating the environment...

blocktrron:
Hopefully fixed everything. Will test my changes tomorrow and report back.

blocktrron:
Fixed two bugs with my changes, works as expected now.

Added the DIR-825 to the tiny device class, as I've missed it while researching the firmware partition sizes. Now all devices of the standard device class have at least 7M of flash space.

docs/dev/hardware.rst
```diff
@@ -9,6 +9,18 @@ Having an ath9k, ath10k or mt76 based WLAN adapter is highly recommended,
 although other chipsets may also work. VAP (multiple SSID) support
 is a requirement.
 
+.. _device-class-definition:
+
+Device classes
+--------------
+Gluon is currently aware of two device classes. Depending on the device class, different
```
mweinelt:
```suggestion
Gluon currently is aware of two device classes. Depending on the device class, different
```

docs/dev/hardware.rst
```diff
@@ -9,6 +9,18 @@ Having an ath9k, ath10k or mt76 based WLAN adapter is highly recommended,
...
+features can be installed onto the device.
+
+The ``tiny`` device-class contains the following type of devices:
```
mweinelt:
```suggestion
The ``tiny`` device-class contains devices with the following limitations:
```

docs/features/private-wlan.rst
```diff
@@ -4,6 +4,14 @@ Private WLAN
 It is possible to set up a private WLAN that bridges the WAN port and is separated from the mesh network.
 Please note that you should not enable ``mesh_on_wan`` simultaneously.
 
+The private WLAN is encrypted via WPA2 by default. On devices with enough flash and a supported radio,
```
mweinelt:
```suggestion
The private WLAN feature uses WPA2 encryption by default. On devices with enough flash and a supported radio,
```

docs/features/private-wlan.rst
```diff
@@ -4,6 +4,14 @@ Private WLAN
...
+WPA3 or WPA2/WPA3 mixed-mode can be used instead of WPA2. For this to work, the ``wireless-encryption-wpa3``
+feature has to be added to the list of your supported features.
+
+Additionally, IEEE 802.11w management frame protection should be enforced for WPA2/WPA3, however this
```
mweinelt:
```suggestion
It is recommended to require IEEE 802.11w management frame protection for WPA2/WPA3 networks, however this
```

docs/features/private-wlan.rst
```diff
@@ -4,6 +4,14 @@ Private WLAN
...
+can lead to connectivity problems for older clients. In this case, management frame protection can be
+set to optional or disabled completely using advanced settings.
```
mweinelt:
```suggestion
downgraded to be either optional or completely disabled using advanced settings. 
```

docs/user/supported_devices.rst
```diff
@@ -424,6 +424,12 @@ Footnotes
   The device or target is reaching its end of life soon. This means that support
   in the next major release of Gluon is doubtful.
 
+.. [#device-class-tiny]
+  These devices only support a subset of Gluons capabilities due to size or memory
```
```suggestion
  These devices only support a subset of Gluons capabilities due to flash or memory size
```

docs/user/supported_devices.rst
```diff
@@ -424,6 +424,12 @@ Footnotes
...
+  constraints. Devices are classified as tiny in they provide less than 7M of useable
```
mweinelt: I'd prefer usable over useable.

package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua
```diff
@@ -55,4 +56,33 @@ function M.is_outdoor_device()
 	return false
 end
 
+function M.device_supports_wpa3()
+	if M.match('ramips', 'rt305x') or M.match('ramips', 'mt7620') then
```
mweinelt: Is it possible to add a reference here why those exact targets are not supporting WPA3? How did you identify they don't?

package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua
mweinelt: Besides checking the debug we can screenscrape `iw` output.

```
iw phy phy0 info | grep 00-0f-ac:6
```

via https://kevinlocke.name/bits/2019/12/28/checking-802.11w-support/
https://kevinlocke.name/bits/2019/12/28/checking-802.11w-support/

package/gluon-web-private-wifi/i18n/de.po
```diff
@@ -13,18 +13,42 @@ msgstr ""
...
 msgstr "Name (SSID)"
 
+msgid "Optional"
+msgstr ""
+
 msgid "Private WLAN"
 msgstr "Privates WLAN"
 
+msgid "Required"
+msgstr "Aktiviert"
```
mweinelt: Erzwingen?

package/gluon-web-private-wifi/luasrc/lib/gluon/config-mode/model/admin/privatewifi.lua
```diff
@@ -34,16 +54,24 @@ function f:write()
...
 				mode       = 'ap',
-				encryption = 'psk2',
+				encryption = encryption.data,
 				ssid       = ssid.data,
 				key        = key.data,
 				macaddr    = macaddr,
 				disabled   = false,
 			})
+
+			-- hostapd-mini won't start in case 802.11w is found in
```
mweinelt:
```suggestion
			-- hostapd-mini won't start in case 802.11w is configured.
```

package/gluon-wireless-encryption/Makefile
mweinelt: `PKG_VERSION` is still here, did you miss this?

scripts/target_config_lib.lua
mweinelt: I'm not sure how this was resolved either.

package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua
blocktrron: I know the debugfs is not the must robust way, but i doubt screenscraping iw (which itself says you should not screenscrape it) is the better solution.

scripts/target_config_lib.lua
blocktrron: Faulty push, fixed.

package/gluon-wireless-encryption/Makefile
blocktrron: Faulty push, fixed.

package/gluon-web-private-wifi/i18n/de.po
blocktrron: Please not. MFP is a feature which can be *activated*, *deactivated* or flagged as *optional*. Erzwingen is not the correct term in this context.

package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua
blocktrron: done

docs/features/private-wlan.rst
blocktrron: Altered this part according to @NeoRaider s previous feedback (which was overwritten by accident).

docs/features/private-wlan.rst
blocktrron: done

package/gluon-web-private-wifi/luasrc/lib/gluon/config-mode/model/admin/privatewifi.lua
blocktrron: done

docs/user/supported_devices.rst
blocktrron: done

docs/user/supported_devices.rst
blocktrron: done

docs/features/private-wlan.rst
blocktrron: Altered this part according to @NeoRaider s previous feedback (which was overwritten by accident).

PR #1877, opened 2019-11-24
gluon-mesh-babel: Fix packets leaving wrong interface

christf:
when there is dual-stack on wan packets might leak out there that should be routed through the Freifunk. iNow that we have babeld 1.9 we can utilize the respective babeld feature and fix that issue.

### DIFF
--- a/package/gluon-mesh-babel/luasrc/lib/gluon/upgrade/300-gluon-mesh-babel-mkconfig
+++ b/package/gluon-mesh-babel/luasrc/lib/gluon/upgrade/300-gluon-mesh-babel-mkconfig
@@ -1,6 +1,8 @@
 #!/usr/bin/lua
 
 local site = require 'gluon.site'
+local uci = require('simple-uci').cursor()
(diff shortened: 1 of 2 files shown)


package/gluon-mesh-babel/luasrc/lib/gluon/upgrade/300-gluon-mesh-babel-mkconfig
```diff
@@ -1,6 +1,9 @@
 #!/usr/bin/lua
 
 local site = require 'gluon.site'
+local uci = require('simple-uci').cursor()
+nodeip = string.gsub(uci:get('network', 'loopback', 'ip6addr'), "/128", "")
```
NeoRaider: - Missing `local`
- Better use `uci:get('network', 'loopback', 'ip6addr'):match('^[^/]+')`

christf: thanks.

PR #1799, opened 2019-08-23
gluon-mesh-babel: Reduce load in respondd

christf:
This reduces the load by:
* querying the default route using netlink instead of parsing it from babel
* instead of dump, use monitor and a separate thread to obtain the babel neighbours

- [ ] resolve TODO in a different PR this on is meant to fix a problem: include traffic stats for all mesh interfaces

PR #1512, opened 2018-08-11
remove respondd data which is not mesh relevant

genofire:
Those data should not be provided by mesh protocol packages.
(Most of them is duplicated)

DO NOT MERGE without Merge of  https://github.com/freifunk-gluon/packages/pull/188

PR #1879, opened 2019-11-24
gluon-mesh-* respondd cleanup (part 1?)

NeoRaider:
Following up on #1512, this is a first step of implementing my suggestion to move common parts of the gluon-mesh-* respondd providers into gluon-respondd. Much more cleanup is necessary in the Babel respondd provider, but I ignored most of that for now to allow easier rebases of #1799 onto this.

In addition, this reverts some of #1676, which means that connected, but inactive wifi clients will disappear from the "total" client count again. One option would be to drop the "total" altogether and only let the mesh protocol provider add a "wired" client count, but I'm not sure if l3roamd can even distinguish wired and wireless clients at the moment.

I've also noticed that there are currently way too many different ways to iterate over mesh interfaces in these providers: UCI, ubus calls to netifd, and iteration over sysfs files. For IP addresses, both getifaddrs and a low-level netlink implementation exist. I haven't touched this for now, but it seems like we should unify a lot of this.

Finally, I haven't tested this much so far, and I wouldn't be surprised if I broke something...

### DIFF
--- a/package/gluon-mesh-batman-adv/src/Makefile
+++ b/package/gluon-mesh-batman-adv/src/Makefile
@@ -31,5 +31,8 @@ endif
 CFLAGS += $(LIBBATADV_CFLAGS)
 LDLIBS += $(LIBBATADV_LDLIBS)
 
-respondd.so: respondd.c
-	$(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) -shared -fPIC -D_GNU_SOURCE -o $@ $^ $(LDLIBS) -lgluonutil -luci
+
+SOURCES = respondd.c respondd-nodeinfo.c respondd-statistics.c respondd-neighbours.c
+
+respondd.so: $(SOURCES) respondd-common.h
+	$(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) -shared -fPIC -fvisibility=hidden -D_GNU_SOURCE -o $@ $(SOURCES) $(LDLIBS) -lgluonutil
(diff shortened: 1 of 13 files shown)


package/gluon-mesh-batman-adv/src/Makefile
```diff
@@ -31,5 +31,8 @@ endif
 CFLAGS += $(LIBBATADV_CFLAGS)
 LDLIBS += $(LIBBATADV_LDLIBS)
 
-respondd.so: respondd.c
-	$(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) -shared -fPIC -D_GNU_SOURCE -o $@ $^ $(LDLIBS) -lgluonutil -luci
+
+SOURCES = respondd.c respondd-nodeinfo.c respondd-statistics.c respondd-neighbours.c
```
christf: where doesrespondd.c come from?

NeoRaider: From the package's `src` directory, just like before?

rotanid:
@NeoRaider did you make progress? it seems to be blocking #1799 which was planned to be part of the release we want to do soon.

christf:
> One option would be to drop the "total" altogether and only let the mesh protocol provider add a "wired" client count, but I'm not sure if l3roamd can even distinguish wired and wireless clients at the moment.

l3roamd does not distinguish between wired or wifi clients but it knows which client is on which interface. Using that however is a stretch. The current implementation of the babel respondd provider uses the client count from l3roamd as total and uses iwinfo to provide the detailed information which wifi channel contains what amount of clients.
I do not think this distinction (wired or wifi) belongs into l3roamd either because I have not yet found a use-case that requires this knowledge from a roaming perspective.

rotanid:
> @NeoRaider did you make progress? it seems to be blocking #1799 which was planned to be part of the release we want to do soon.

bump @NeoRaider

christf:
These changes have been tested as part of the babel network in Magdeburg. No brekage was visible there.

package/gluon-mesh-batman-adv/src/Makefile
christf: yeah, I was tricked by the github gui. nevermind.

PR #1883, opened 2019-11-27
ath79-generic: add support for devolo WiFi pro 1750e

herrbett:
  - [ ] webinterface
  - [ ] tftp
  - [x] other: sysupgrade via SSH on the Vendor Firmware
- wired network
- wifi (if applicable)
- led mapping
  - radio leds
    - no radio led
  - switchport leds
    - no switchport leds

Has the same "issues" regarding the LAN Port mappings as the Devolo WiFi pro 1200e.
See #1882

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -187,6 +187,7 @@ ath79-generic
 
   - WiFi pro 1200i
   - WiFi pro 1750c
+  - WiFi pro 1750e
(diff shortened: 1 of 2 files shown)


blocktrron:
We've talked about this yesterday, and came to the conclusion, which is outlined in #1886. As the decision was made and the documentation PR is now open, this PR (just like the others) should be mergable.

PR #1886, opened 2019-12-03
docs: clarify WAN port on LAN-only devices

blocktrron:
At the December 2019 review day, the behavior for LAN only devices was defined. Such devices will use each LAN port as WAN.

### DIFF
--- a/docs/dev/wan.rst
+++ b/docs/dev/wan.rst
@@ -11,6 +11,11 @@ There are two cases in which the WAN port is used:
 After the VPN connection has been established, the node should be able to reach
 the mesh's DNS servers and use these for all other name resolution.
 
+If the device does not feature a WAN port, the LAN port is configured as WAN port.
+In case such a device has multiple LAN ports, all these can be used as WAN.
+Devices, which feature a "hybrid" port (labled as WAN/LAN), this port is used as WAN.
+
+This behavior can be reversed using the ``single_as_lan`` site.conf option.
 
 Routing tables
 ~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


mweinelt:
The "single_as_lan" option should be mentioned. (neoraider via IRC)

docs/dev/wan.rst
```diff
@@ -11,6 +11,11 @@ There are two cases in which the WAN port is used:
 After the VPN connection has been established, the node should be able to reach
 the mesh's DNS servers and use these for all other name resolution.
 
+If the device does not feature a WAN port, the LAN port is configured as WAN port.
+In case such a device has multiple LAN ports, all these can be used as WAN.
+Devices, which feature a "hybrid" port (labled as WAN/LAN), this port is used as WAN.
+
+This behavior can be reversed using the ``single_as_lan`` site-option.
```
mweinelt:
```suggestion
This behavior can be reversed using the ``single_as_lan`` site.conf option.
```

blocktrron:
@rubo77 CPE210 (but this is wrongly assumed to have a WAN port upstream, no one complained yet), DSL Routers and many enterprise wireless access points.

This is - strictly speaking - not about devices which are already in Gluon, but about devices which can be perspectively included into Gluon.

docs/dev/wan.rst
blocktrron: Thanks, adapted.

mweinelt:
There should be a remark added to the supported devices list, which devices are affected by this.

PR #1887, opened 2019-12-04
gluon-status-page: add gateway_nexthop to status page statistics [hexa, please don't push this one again accidently]

lemoer:


mweinelt:
Ok, I won't. Pinky promise!

mweinelt:
The unminified version works, the minified one does not.

PR #1890, opened 2019-12-12
lantiq-xway: add support for AVM FRITZ!Box 7320, 7330 and 7330 SL

Dark4MD:
      
      * [ ]  webinterface
      * [ ]  tftp
      * [x]  other: fritzfla.sh script or adam2 ftp

      * [?] must have working autoupdate

    * [x]  wps button must return device into config mode

    * [n/a]  primary mac should match address on device label (or packaging) (https://gluon.readthedocs.io/en/latest/dev/hardware.html#notes)

    * wired network
      
    * wifi (if applicable)
      
    * led mapping
      
      * radio leds
        
        * [x]  should map to their respective radio (with included patch)
        * [x]  should show activity (with included patch)
      * switchport leds
        
        * [n/a]  should map to their respective port (or switch, if only one led present)
        * [n/a]  should show link state and activity

    * outdoor devices only
      
      * [n/a]  added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`
        ~
All three versions share the same Hardware except no POTS for SL nariant. Tested on 7330 SL.
Flashing with modified fritzfla.sh script successfully tested.

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -261,6 +261,9 @@ lantiq-xway
 * AVM
 
   - FRITZ!Box 7312 [#avmflash]_
+  - FRITZ!Box 7320 [#avmflash]_ [#lan_as_wan]_
+  - FRITZ!Box 7330 [#avmflash]_ [#lan_as_wan]_
+  - FRITZ!Box 7330 SL [#avmflash]_ [#lan_as_wan]_
 
 mpc85xx-generic
 ---------------
--- a/targets/lantiq-xway
+++ b/targets/lantiq-xway
@@ -1,3 +1,8 @@
 device('avm-fritz-box-7312', 'avm_fritz7312', {
 	factory = false,
 })
+
+device('avm-fritz-box-7320', 'avm_fritz7320', {
+        factory = false,
+        aliases = {'avm-fritz-box-7330', 'avm-fritz-box-7330-sl'},
+})


heini66:
7320sl  could also be added here

Dark4MD:
> 
> 
> 7320sl could also be added here

Do you have something like a wiki page or details page about this variant? I already looked if there are devices that could also use the same image and this one didn't showed up.
If you happen to have this device maybe you could even test it.

docs/user/supported_devices.rst
```diff
@@ -261,6 +261,9 @@ lantiq-xway
 * AVM
 
   - FRITZ!Box 7312 [#avmflash]_
+  - FRITZ!Box 7320 [#avmflash]_ [#lan_as_wan]_
+  - FRITZ!Box 7330 [#avmflash]_ [#lan_as_wan]_
+  - FRITZ!Box 7330 SL [#avmflash]_ [#lan_as_wan]_
```
blocktrron: Have you tested AND verified it with every distinct model? I'm asking as the kernel might run on every device, but GPIO assignment / flags might differ between models.

patches/openwrt/0006-lantiq-xway-remove-redundant-wifi-led-on-fritz-box-7320.patch
```diff
@@ -0,0 +1,29 @@
+From 509894cffb7ec24ac21a381bed3c035efce42aeb Mon Sep 17 00:00:00 2001
+From: Manuel Kock <github.web@manu.li>
+Date: Mon, 11 Nov 2019 13:01:33 +0100
+Subject: [PATCH] lantiq: remove redundant WiFi LED on FRITZ!Box 7320
```
blocktrron: I'll pick this patch upstream so we can get rid of this patch. Will ping you when done.

docs/user/supported_devices.rst
Dark4MD: OpenWRT is for the 7320 and my 7330SL works fine with it, every button and led works so i guess the 7320 would work fine also. The 7330 has only POTS extra but is the same as the 7330SL. 
According to some threads in the IP Phone Forum it is even possible to mix the Stock firmwares for the 3 models with some hacks.
But if you want i can test with an 7320 to be sure, i just bid on it on ebay for cheap ;).

patches/openwrt/0006-lantiq-xway-remove-redundant-wifi-led-on-fritz-box-7320.patch
Dark4MD: Already saw the modules bump and removed the patch ;)

docs/user/supported_devices.rst
Dark4MD: My 7320 arrived today and works flawlessly with the image.

rubo77:
@heini66 said in chat: he has one and his test was successfull

blocktrron:
@rubo77 AVM does not list a 7320SL in their KB, so i would be very surprised this model ever existed apart from engineering samples or similar.

heini66:
@blocktrron take a look @ ebay / ebay-kleinanzeigen . there are several 7320sl ...

blocktrron:
@heini66 Could you send me a reference to such an offering?

All I've found was a single eBay listing which did not contain a Photo from the bottom label. So i assume the seller mistakenly offers a regular 7320 as 7320SL.

rotanid:
these devices were removed again in April 2020, see #1995 for the details and #1943 for the reason.

PR #1892, opened 2019-12-19
mpc85xx-p1020: add support for Enterasys WS-AP3710i

blocktrron:
The Enterasys / Extreme Networks WS-AP3710i is a ceiling / wall mount PoE powered access point.
It has a dual-core PowerPC processor and dualband WiFi radios.

It can be acquired for cheap (~15 Euros) second hand from IT redistributors.

  - [ ] webinterface
  - [ ] tftp
  - [x] other: sysupgrade via initramfs / serial console
    *enterasys-ws-ap3710i*
  - Reset button recessed at the bottom of the device
  - all mac addresses are printed on the label, matches Radio 1 (5 GHz) wifi mac
- wired network
- wifi (if applicable)
- led mapping
  - radio leds
    - [x]  should map to their respective radio
    - [x]  should show activity
  - switchport leds
    - [x]  should map to their respective port (or switch, if only one led present)
    - [x]  should show link state and activity

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -276,6 +276,10 @@ mpc85xx-p1020
 
   - HiveAP 330
 
+* Enterasys
+
+  - WS-AP3710i
+
 * OCEDO
 
   - Panda
--- a/targets/mpc85xx-p1020
+++ b/targets/mpc85xx-p1020
@@ -4,6 +4,14 @@ device('aerohive-hiveap-330', 'hiveap-330', {
         factory = false,
 })
 
+
+-- Enterasys
+
+device('enterasys-ws-ap3710i', 'ws-ap3710i', {
+	factory = false,
+})
+
+
 -- OCEDO
 
 device('ocedo-panda', 'panda', {


kpanic23:
Slightly offtopic, but does the device require 802.3af PoE or will a 24V passive injector suffice? I can't seem to find that information online.

blocktrron:
@kpanic23 The device requires 802.3af (Thankfully)

rotanid:
am i correct, that i always need to flash openwrt first?
my build got me only the sysupgrade image for the device, nothing else...

kpanic23:
Well, yes and no.
You need to boot OpenWRT from RAM (after TFTPing it there), then you can flash Gluon from OpenWRT.

blocktrron:
Correct. The initramfs can be obtained from the latest OpenWrt release build:

https://cdn.openwrt.org/releases/19.07.0/targets/mpc85xx/p1020/openwrt-19.07.0-mpc85xx-p1020-ws-ap3710i-initramfs-kernel.bin

rotanid:
thanks, i knew how to do it, i was just wondering if this is indeed intended.
to which you provided the answer now.

PR #1894, opened 2019-12-22
ipq40xx: add support for Aruba AP-303

blocktrron:
this commit adds support for the Aruba AP-303 Wireless access point. The device is as well amrketed as the Aruba InstantOn AP11 for around 70 Euro.

As the Device supports 802.11ac Wave 2 MU-MIMO, has a Quad-Core IPQ4029 ARM SoC and 512 MB (!) of DDR3 RAM, it is a compelling alternative to the Ubiquiti UniFi AC Lite.

Installation instructions are available at the OpenWrt Wiki: https://openwrt.org/toh/aruba/aruba_ap-303

-----

  - [ ] webinterface
  - [x] other: TFTP initramfs via U-Boot 
     > aruba-ap-303
- wired network
- wifi (if applicable)
- led mapping
  - radio leds
    - [x] should show activity

### DIFF
--- a/patches/openwrt/0007-ipq-wifi-add-BDF-for-Aruba-AP-303.patch
+++ b/patches/openwrt/0007-ipq-wifi-add-BDF-for-Aruba-AP-303.patch
@@ -1,0 +1,92 @@
+From: David Bauer <mail@david-bauer.net>
+Date: Sun, 15 Dec 2019 23:02:54 +0100
+Subject: ipq-wifi: add BDF for Aruba AP-303
+
+The BDF originates from the vendor-firmware.
+
+Signed-off-by: David Bauer <mail@david-bauer.net>
+(cherry picked from commit 4113d8a2554adf5ecee55cc07956eafad378eaff)
+
+diff --git a/package/firmware/ipq-wifi/Makefile b/package/firmware/ipq-wifi/Makefile
+index eb7c2df1aa36ded7774a772c8a7e02b2acb81b40..cc0505b97c6a04bafd88972cf6ce7890a637c33b 100644
+--- a/package/firmware/ipq-wifi/Makefile
++++ b/package/firmware/ipq-wifi/Makefile
+@@ -25,6 +25,7 @@ endef
+ 
+ ALLWIFIBOARDS:= \
+ 	alfa-network_ap120c-ac \
++	aruba_ap-303 \
+ 	asus_map-ac2200 \
+ 	avm_fritzbox-7530 \
+ 	avm_fritzrepeater-1200 \
+@@ -97,6 +98,7 @@ endef
+ # Add $(eval $(call generate-ipq-wifi-package,<devicename>,<display name>))
+ 
+ $(eval $(call generate-ipq-wifi-package,alfa-network_ap120c-ac,ALFA Network AP120C-AC))
++$(eval $(call generate-ipq-wifi-package,aruba_ap-303,Aruba AP-303))
+ $(eval $(call generate-ipq-wifi-package,asus_map-ac2200,ASUS MAP-AC2200))
+ $(eval $(call generate-ipq-wifi-package,avm_fritzbox-7530,AVM FRITZ!Box 7530))
+ $(eval $(call generate-ipq-wifi-package,avm_fritzrepeater-1200,AVM FRITZRepeater 1200))
+diff --git a/package/firmware/ipq-wifi/board-aruba_ap-303.qca4019 b/package/firmware/ipq-wifi/board-aruba_ap-303.qca4019
+new file mode 100644
+index 0000000000000000000000000000000000000000..4848115cfbe3a4a0ed6b17cac929731ecbd7968c
+GIT binary patch
+literal 24316
+zcmeHPdr(tX8b1l)p-UHTNDu)pAp}Se0tBi-!W$`%6c7~&un1I<M58<m4?`F1_@EFG
+zfm#qEXc?48Q$)t%1F%?IJEP40v48CBA8luLw|}fVw5!gvGrOyM?hO}02ql3q<#9ha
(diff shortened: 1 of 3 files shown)


rotanid:
is the serial console available without opening the device? i'm asking because according to openwrt link you provided, it is necessary to use the serial console to flash the device - and gluon's policy for non-broken devices is that they need to be flashable without opening a case.

mweinelt:
The console port is on the outside, but needs a custom cable with the model number JY728A.

> As the Device supports 802.11ac Wave 2 MU-MIMO, has a Quad-Core IPQ4029 ARM SoC and 512 MB (!) of DDR3 RAM, it is a compelling alternative to the Ubiquiti UniFi AC Lite.

Together with the current price point this qualifies for backporting the upstream commit.

rotanid:
custom cable, oof, i would vote against non-broken or even against merge...

rubo77:
I think, if you need an extra cable, it should not be listed as supported by Gluon

belzebub40k:
IMHO it doesn't make a real difference for a normal user if you need a "regular" serial cable like for the Ubiquiti Edgerouter X (not marked as broken) or if it is a special cable from the manufacturer. In both cases most users would need to buy the suitable cable/adapter.

blocktrron:
@rotanid The cable itself might be custom, however it is UART mapped on a micro-USB cable which is already used on some HP ProCurve switches. It can be soldered within minutes using a micro-USB cable and an FTDI. Also we already require serial cables for the ER-X (which is also a custom Cisco-Style cable).

Given the fact out current selection of ceiling-mount access points are scarce, i think this device is a welcome 2019-alternative to a UniFi AC Lite.

NeoRaider:
ER-X should not need any kind of serial cable, you can install that via SSH.

By our current definition of BROKEN, this device is indeed problematic. As discussed before, I'd like to reduce our use of "broken", and instead
- tag problematic devices with more specific "caveats"
- build problematic devices by default
- output some kind of machine-readable manifest of all images including caveats that could be used to build firmware download pages that include warnings etc.

blocktrron:
@NeoRaider To my knowledge, the SSH installation of the ER-X is void since Ubiquiti pushed out new Firmware some months ago, just like how they removed the necessary toolset from their UniFi series some months ago.

Actually, the Aruba AP-303 can also be flashed by similar method in case the AP is a fully-licensed Campus product (in this case you have CLI access), however this does not apply for the cheaper Instant or InstantON product line (which feature the interesting price-point).

My personal stand here is that i find the installation method for the Aruba more welcoming than the hassle which EVA Ramboot is (added bonus points for the interim-bootloader-step). It might be possible without additional tools, however it takes around 5 minutes without any visible feedback which step is currently happening.

As @mweinelt pointed out, this device offers really good hardware, built quality (seriously, a UniFi or AVM model feels like a child's toy  compared to this) and a hassle free way back to the vendor firmware. Given the price-point I'm more than pleased with what this device offers.

I will however leave the final decision about the inclusion to Gluon up to other people.

mweinelt:
Well, flashing some devices does require a certain kind of skillset - or the motivation to learn the required skillset. I see little to no issue, as the required pins are clearly available on the outside of the device.

We are fine with describing complex flash methods in well prepared guides as long as they can be done by and enduser sitting in front of a pc. Who says that people aren't crafty enough to fix up their own cable? Or yet buy the original one, because they want to flash a bunch of these and don't mind the extra cost. Why do we think it's necessary to patronize endusers in that way?

Let's just do our best to indicate how, and how hard it is, to install Gluon on these devices. It would be a waste not to include devices like this.

rotanid:
"Let's just do our best to indicate how, and how hard it is, to install Gluon on these devices"
how to achieve this *before* people buy the devices? as soon as we merge it non-broken, all communities are over time going to build the images - most without leaving any note anywhere about the added complexity compared to current supported devices.

blocktrron:
@rotanid I understand the point you are making here, however we have this case already. We have (apart from easy one-click web installation) countless ways of installation and Gluon should deliver Metadata about every device and references to instructions how to install it.

However, I'm with @mweinelt that we should not treat users like a child. If we give them the right instructions, they can also master a more complicated way of installation. See what we did with the FritzBox 4020 and 4040. Regarding purchasing decisions: I expect people to inform themselves upfront, but even if they didn't do it they can look up the information afterwards.

Maybe we can do a POC of metadata generation and deliver this not only because of this specific device but for all our other devices? This way we could also auto-generate everything the Gluon Firmware selector needs as soon as we have the more-detailed target-metadata from OpenWrt.

rotanid:
"See what we did with the FritzBox 4020 and 4040"
in our community, no one is buying the 4040 without one of the core people flashing it (and i don't see the point in buying the 4020)

what about merging it as broken, as long as we don't have a solution regarding metadata and such?

blocktrron:
@rotanid That's quite the opposite we've experienced in Darmstadt with people flashing their AVM devices themselves. As a sidenote.

Merging this as broken would violate our policy (and I'm still under the impression BROKEN does not apply here), but if we agree on making an exception in this case I'm fine with it.

awlx:
> "See what we did with the FritzBox 4020 and 4040"
> in our community, no one is buying the 4040 without one of the core people flashing it (and i don't see the point in buying the 4020)
> 
> what about merging it as broken, as long as we don't have a solution regarding metadata and such?

We (Freifunk München) have over 25 4040/4020 devices in our network and even a 7312. Two of the 4040 were flashed by the core team. And two people needed help after the purchase. So I think this is a very well working method and perfectly fine.

So I as well want to see the Aruba device being merged :).

rotanid:
a kind of vote among maintainers resulted in vote slightly for merging devices which need custom built cables.
before merging it, we should adapter our wiki accordingly.
we should define this very clearly and in this step we can also define more clearly which devices are meant when saying they must not need to be opened to get them working. on devices with screws (not clips) it's easiert to do the opening than the creation of a custom cable. on the other hand, the cable doesnt have to be done for each device again...

blocktrron:
@rotanid I've added a note to the device wiki page regarding this.

PR #1901, opened 2019-12-29
docs: site example: add validation script and travis config

rubo77:
# .travis.yml and test folder

The test folder contains a validation script, that can be called in your site folder if you want to validate the external packages, you defined in your modules file. Travis can be added to your git repository to automatically validate new commits to your site repository. It can also be called manually in your console with `bash tests/validate_site.sh`.

the validate_site.sh script will

- create a temporary test-folder
- clone the latest commit of gluon there
- check for valid repositories and commit-IDs in `modules`
- clone all defined repositories in their defined commit in `modules` 
- Lua linter check for all package feeds (call `luacheck`) in their checked-out state
- check if all packages defined in site.mk exist in either gluon or the checked out packages
- call `bash -n` on all (potentioally) existing `*.sh` scripts in the site folder

If used in GitHub you get those nice green hooks after each commit. Example: https://github.com/freifunk-kiel/site-ffki/commits/next

### DIFF
--- a/docs/site-example/tests/validate_site.sh
+++ b/docs/site-example/tests/validate_site.sh
@@ -1,0 +1,123 @@
+#!/bin/bash
+set -uo pipefail
+
+# validate_site.sh checks if the site.conf is valid json
+GLUON_REPO="https://github.com/rubo77/gluon"
+GLUON_BRANCH='ssid-changer'
+GLUON_PACKAGES_REPO="https://github.com/freifunk-gluon/packages"
+GLUON_PACKAGES_BRANCH='master'
+
+P="$(pwd)"
+echo "####### check if lua5.1 is installed ..."
+which lua5.1 
+if [ "$?" == 1 ]; then
+  echo lua5.1 not present!
+  echo install with sudo apt install lua5.1
+  exit 1
+fi
+
+CONFIGS="site.conf"
+if [ -d "domains" ]; then
+  CONFIGS="$CONFIGS "domains/*
+fi
+
+for c in $CONFIGS; do
+  echo "####### validating lua $c ..."
+  GLUON_SITEDIR="." GLUON_SITE_CONFIG="$c" lua5.1 tests/site_config.lua
+  if [ "$?" == 1 ]; then
+    exit 1
+  else
+    echo "OK: $c"
+  fi
+done
+
+for BASHFILE in "$P"/*.sh; do
+    [ -f "$BASHFILE" ] || continue
+  echo "####### validating $BASHFILE ..."
+  bash -n "$BASHFILE"
+  if [ "$?" == 0 ]; then
+    echo "OK: $BASHFILE"
+  fi
+done
+
+echo "####### validating $P/modules ..."
+GLUON_SITE_FEEDS="none"
+source "$P/modules"
+testpath=/tmp/site-validate
+rm -Rf "$testpath"
+mkdir -p "$testpath/packages"
+cd "$testpath/packages"
+if [ "$GLUON_SITE_FEEDS" != "none" ]; then
+  for feed in $GLUON_SITE_FEEDS; do
+    echo "####### checking PACKAGES_${feed^^}_REPO ..."
+    repo_var="PACKAGES_${feed^^}_REPO"
+    commit_var="PACKAGES_${feed^^}_COMMIT"
+    branch_var="PACKAGES_${feed^^}_BRANCH"
+    repo="${!repo_var}"
+    commit="${!commit_var}"
+    branch="${!branch_var}"
+    if [ "$repo" == "" ]; then
+      echo "repo $repo_var missing"
+      exit 1
+    fi
+    if [ "$commit" == "" ]; then
+      echo "commit $commit_var missing"
+      exit 1
+    fi
+  	if [ "$branch" == "" ]; then
+      echo "branch $branch_var missing"
+      exit 1
+    fi
+    git clone -b "$branch" --depth 1000 --single-branch "$repo" "$feed"
+    if [ "$?" != "0" ]; then exit 1; fi
+    cd "$feed"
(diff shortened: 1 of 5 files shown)


docs/user/site.rst
```diff
@@ -734,6 +734,16 @@ modules
 .. literalinclude:: ../site-example/modules
   :language: makefile
 
+.travis.yml and ``test`` folder
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The test folder contains a validation script, that can be called in your site
```
rotanid: this comma is obsolete in english

docs/user/site.rst
```diff
@@ -734,6 +734,16 @@ modules
...
+folder if you want to validate the external packages, you defined in your
```
rotanid: this comma is obsolete in english

lemoer:
I appreciate the idea, but why not use this:

https://github.com/freifunk-gluon/gluon/blob/master/Makefile#L98

rubo77:
Would that work without creating the whole 10gh building surroundings?

lemoer:
I do not understand your question?

rubo77:
If you just want to edit your site repository, you do not want to create a whole Gluon around it before you can commit

rubo77:
> I appreciate the idea, but why not use this:
> 
> https://github.com/freifunk-gluon/gluon/blob/master/Makefile#L98

the script uses this already.

But it does a lot more (I'll add it to the first post hier)

lemoer:
> Imagine, you work on a laptop with only a few GB free Disk space and on a slow (Freifunk-) internet connection.

A checkout of the gluon repository is 17MB, so I actually do not get your point...

Please place your script in `scripts/`. If people copy code from the gluon repo to their site repo, they will not get updates of your script as soon as there are updates...

rubo77:
yes, the download is 17MB, but if you use use `make` it starts building the whole buildchain which ends up in 10GB space. My script uses that `scripts/site_config.lua` without `make` 

If you have an Idea how to improve my script, you are welcome to add the changes, you think of. (maybe you create a PR on https://github.com/rubo77/gluon/tree/travis-site-check ) to be clear, what you mean.

If we put the .travis config in the `scripts` folder, I am afraid, noone of the users, that are not such erperts, will use it, but for exactly those users it would help the most

lemoer:
> but if you use use make it starts building the whole buildchain
That is not true. `make update-patches` or `make update` does not start building the toolchain.

rubo77:
Ah Ok, so I would be happy, if you improve the script

NeoRaider:
The bash script is extremely complex and mostly duplicates what a simple `make config` in a checked out Gluon can already do more reliably.

rubo77:
But gluon cannot check before you start building in the beginning, if all packages that are listed in site.mk are in the marked commits in the modules file. This check-script does this in seconds.

It saved me lots of build-time on our build-server by detecting misconfigurations before I started. It also finds out, if somone removes a used branch, that is linked in a linked user-repo, or such, before you start.

But if you have a fast alternative without lots of GB harddisk-usage for the build environment, I will be happy to change the script

rubo77:
When I call `make  update-patches`, it starts with

```
--- Updating module 'openwrt' ---
Reinitialized existing Git repository in /tmp/gluon/openwrt/.git/
```
which already takes hours on my mobile hotspot

NeoRaider:
I think any meaningful check will have to run at least `make update`. `make config` is indeed a bit more work, as Gluon will build some OpenWrt tools and the Lua binary, but it should still not take more than a few minutes from scratch - much less on subsequent runs (and a proper CI should be able to cache the build directory). It might even be possible to reduce the initial time further by using a precompiled Lua, but I'll have to look into that more, and I don't consider it essential for this PR.

I have no idea why you would bring up a mobile hotspot - for a CI runner, 250MB shouldn't be any issue, and when running such scripts locally, you only need to download the Gluon build environment once, which should obviously be done using a proper internet line...

rubo77:
This pr is not about a full check, which is already solved by Jenkins

This script would help users to just reddit the site locally, before they commit anything to their main repository, which uses aci

docs/site-example/tests/validate_site.sh
```diff
@@ -0,0 +1,123 @@
...
+GLUON_BRANCH='ssid-changer'
+GLUON_PACKAGES_REPO="https://github.com/freifunk-gluon/packages"
+GLUON_PACKAGES_BRANCH='master'
+
+P="$(pwd)"
+echo "####### check if lua5.1 is installed ..."
+which lua5.1 
+if [ "$?" == 1 ]; then
+  echo lua5.1 not present!
+  echo install with sudo apt install lua5.1
```
blocktrron, line 15: This is not platform-independent.

docs/site-example/tests/validate_site.sh
```diff
@@ -0,0 +1,123 @@
+#!/bin/bash
+set -uo pipefail
+
+# validate_site.sh checks if the site.conf is valid json
+GLUON_REPO="https://github.com/rubo77/gluon"
+GLUON_BRANCH='ssid-changer'
```
blocktrron, line 6: While this is just an example, a config-file or passing these variables via CLI might be the better solution.

docs/site-example/tests/validate_site.sh
```diff
@@ -0,0 +1,123 @@
...
+done
+
+echo "####### validating $P/modules ..."
+GLUON_SITE_FEEDS="none"
+source "$P/modules"
+testpath=/tmp/site-validate
+rm -Rf "$testpath"
+mkdir -p "$testpath/packages"
+cd "$testpath/packages"
+if [ "$GLUON_SITE_FEEDS" != "none" ]; then
```
blocktrron, line 50: Unlikely, but a Community might decide to name a feed "none".

PR #1904, opened 2020-01-02
docs: add Archer C60 to supported list

rubo77:


blocktrron:
The Archer C60 is not fully supported by Gluon (and never will be) due to the combination of 64M RAM and ath10k (see the BROKEN flag).

So the device should also not be listed in the supported devices.

rubo77:
Ah, OK, I didn't notice, that all devices with BROKEN flag are not listed in the Supported Devices list.

What is the problem with the C60, I have one and it runs fine

blocktrron:
It might run fine now, but this is only due to a hack-patch minimizing buffer-sizes for ath10k in OpenWrt. To not have another big "32M situation", they are considered BROKEN as of insufficient hardware.

blocktrron:
Closing as explanation was given. If you want to start the discussion about altering the ath10k 64M BROKEN definition, please open an issue / PR specifically for this topic as a whole.

rubo77:
Ok.

Can you please link to the explanation what exactly is the problem with that target?
Why do they have a buffer problem?

blocktrron:
See this thread on the openwrt-devel mailing list: http://lists.infradead.org/pipermail/openwrt-devel/2019-November/020430.html

(Continues in December)

PR #1905, opened 2020-01-04
ramips-mt7620: remove broken flag for Xiaomi MiWifi Mini

Dark4MD:
WiFi driver is stable now.
Router is running for 8 days now and both WiFi radios still work.
The running device:
https://hannover.freifunk.net/karte/#/de/map/6409807bbc0e

  - [ ] webinterface
  - [ ] tftp
  - [x] other: 
There are two ways to flash. Both are written in the OpenWRT Wiki. One via an Code Injection Bug and one via an Account on miwifi.com
https://openwrt.org/toh/xiaomi/mini#quick_openwrt_installation

autoupdater name is still the same when checked with the lua command.
After fresh reset with firstboot and Install of a fresh build it is still the correct mac address as seen on the device.
- wired network
- wifi (if applicable)
- led mapping
  - radio leds
    - [n/a] should map to their respective radio
    - [n/a] should show activity
device has still only one LED for power status
  - switchport leds
- outdoor devices only

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -300,6 +300,10 @@ ramips-mt7620
 
   - WT3020AD/F/H
 
+# Xiaomi
+
+  - MiWiFi Mini
+
 ramips-mt7621
 -------------
 
--- a/targets/ramips-mt7620
+++ b/targets/ramips-mt7620
@@ -36,5 +36,4 @@ device('nexx-wt3020-8m', 'wt3020-8M', {
 
 device('xiaomi-miwifi-mini', 'miwifi-mini', {
 	factory = false,
-	broken = true, -- 2.4GHz WiFi is unstable
 })


blocktrron:
Could you do th complete integration checklist? Especially the primary mac-address and autoupdater name would be interesting after such a long time.

Dark4MD:
> 
> 
> Could you do th complete integration checklist? Especially the primary mac-address and autoupdater name would be interesting after such a long time.

Everything checked and still working fine ;)

docs/user/supported_devices.rst
```diff
@@ -300,6 +300,10 @@ ramips-mt7620
 
   - WT3020AD/F/H
 
+# Xiaomi
+
+  - Mi WiFi Mini
```
blocktrron: Shouldn't this be `MiWiFi Mini` without the space?

Dark4MD: Depends on where you look. OpenWRT and most other sites say Mi Wifi with the Space. I guess because Mi is the sub brand and WiFi is the product type.

blocktrron: The runtime Model string is MiWifi Mini, so i would go for this.

Yes, Xiaomis naming is confusing but in this case consistency is key IMHO

Dark4MD: Done

PR #1908, opened 2020-01-05
Increase ROOTFS size when GLUON_DEBUG is set

christf:
this increases the size of the ROOTFS to 500MB when GLUON_DEBUG is set such that development tools can be included. Thanks to blocktron who prepared the patch.

### DIFF
--- a/targets/generic
+++ b/targets/generic
@@ -54,6 +54,7 @@ end
 if envtrue.GLUON_DEBUG then
 	config 'CONFIG_DEBUG=y'
 	config 'CONFIG_NO_STRIP=y'
+	config 'CONFIG_TARGET_ROOTFS_PARTSIZE=500'
 	config '# CONFIG_USE_STRIP is not set'
 	config '# CONFIG_USE_SSTRIP is not set'
 end


blocktrron:
Plese try the following patch:

```
From 038f6611a13a1ccae0ffdfe1eb3be91fa3a271fd Mon Sep 17 00:00:00 2001
From: David Bauer <mail@david-bauer.net>
Date: Sun, 5 Jan 2020 15:34:52 +0100
Subject: [PATCH] generic: increase RootFS size when GLUON_DEBUG is enabled

This increases the size of the root filesystem when GLUON_DEBUG is
enabled. Otherwise, the filesystem is too small.
---
 targets/generic | 1 +
 1 file changed, 1 insertion(+)

diff --git a/targets/generic b/targets/generic
index 0153386d..9d87ba4e 100644
--- a/targets/generic
+++ b/targets/generic
@@ -54,6 +54,7 @@ end
 if envtrue.GLUON_DEBUG then
 	config 'CONFIG_DEBUG=y'
 	config 'CONFIG_NO_STRIP=y'
+	config 'CONFIG_TARGET_ROOTFS_PARTSIZE=500'
 	config '# CONFIG_USE_STRIP is not set'
 	config '# CONFIG_USE_SSTRIP is not set'
 end
-- 
2.24.1
```

If it works, this should be the more appropriate way to fix the problem you're facing.

christf:
this patch works for me.

christf:
I took the liberty of pushing that patch into this PR

PR #1909, opened 2020-01-05
ipq40xx-generic: add alias for Aruba Instant On AP11

blocktrron:
The Aruba Instant On AP11 is the Aruba AP-303 with a stripped-down firmware.

Add an alias for the device to remove confusion about the different naming.

### DIFF
--- a/targets/ipq40xx-generic
+++ b/targets/ipq40xx-generic
@@ -25,6 +25,7 @@ defaults {
 
 device('aruba-ap-303', 'aruba_ap-303', {
 	factory = false,
+	aliases = {'aruba-instant-on-ap11'},
 })
 
 


mweinelt:
I don't think "Instant On" is worth mentioning in this context, do you?

blocktrron:
As it's (currently) the only AP11 marketed from Aruba, we can drop the instant-on in the image-name.

mweinelt:
aruba-ap-11 or aruba-ap11?

blocktrron:
I've renamed the AcessPoint to fit the naming with the description used in the Datasheet now: 
Screenshot from 2020-01-05 15-40-36

For comparison, the AP-303:
Screenshot from 2020-01-05 15-40-10

What do you think?

mweinelt:
Cool, works for me then.

PR #1910, opened 2020-01-06
ramips-mt7620: add support for TP-Link Archer C2 v1

Dark4MD:
  - [no] webinterface
  - [x] tftp: as described in the OpenWRT Wiki here https://openwrt.org/toh/tp-link/archer_c2_ac750
  - [no] other: <specify>
- wired network
- wifi (if applicable)
- led mapping
  - radio leds (only one)
    - [x] should show activity
  - switchport leds
- outdoor devices only

### DIFF
--- a/targets/ramips-mt7620
+++ b/targets/ramips-mt7620
@@ -32,6 +32,13 @@ device('nexx-wt3020-8m', 'wt3020-8M', {
 })
 
 
+-- TP-Link
+
+device('tp-link-archer-c2-v1', 'tplink_c2-v1', {
+        factory = false,
+})
+
+
 -- Xiaomi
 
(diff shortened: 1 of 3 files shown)


targets/ramips-mt7620
```diff
@@ -32,6 +32,11 @@ device('nexx-wt3020-8m', 'wt3020-8M', {
 })
 
 
+-- TP-Link
+
+device('tp-link-archer-c2-v1', 'tplink_c2-v1')
```
blocktrron: There's no installation process which utilizes the factory image. So we should not include it here.

Dark4MD: Good Point. No. I even tried the Snapshot OpenWRT Factory Image after reverting to stock and it wont work.

PR #1913, opened 2020-01-09
gluon-neighbour-info: use default port 1001 and default dest ::1

lemoer:


rubo77:
What are the consequences of these changes? Our what went wrong without?

rotanid:
as far as i understand it:
no consequences and nothing went wrong. it just adds default values to be able to use it without specifying a value explicitly

PR #1914, opened 2020-01-09
gluon-mesh-b*: name interface name in respondd statistics

lemoer:
This is a prepatory commit to allow showing the airtime statistics and the wifi frequency on the status page.

### DIFF
--- a/package/gluon-mesh-babel/src/respondd.c
+++ b/package/gluon-mesh-babel/src/respondd.c
@@ -620,6 +620,7 @@ static struct json_object * get_wifi_neighbours(const char *ifname) {
 		json_object_object_add(obj, "signal", json_object_new_int(entry->signal));
 		json_object_object_add(obj, "noise", json_object_new_int(entry->noise));
 		json_object_object_add(obj, "inactive", json_object_new_int(entry->inactive));
+		json_object_object_add(obj, "iface", json_object_new_string(ifname));
 
 		char mac[18];
 		snprintf(mac, sizeof(mac), "%02x:%02x:%02x:%02x:%02x:%02x",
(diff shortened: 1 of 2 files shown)


NeoRaider:
The interface name is an implementation detail; we usually identify interfaces by their BSSIDs in respondd data.

In addition, changes to these respondd providers should be based on #1879.

lemoer:
Is there a mapping between BSSID and interface name available in the status-page frontend?

NeoRaider:
I'm not sure, but if it isn't, we can add it.

christf:
> The interface name is an implementation detail; we usually identify interfaces by their BSSIDs in respondd data.
> 
> In addition, changes to these respondd providers should be based on #1879.

It should be based on #1799. The idea was to merge the refactoring then the load fix then everything else.

lemoer:
I updated the code to use bssid. A rebase is not done yet.

lemoer:
As neoraiders #1879 is now merged, should I wait till christfs #1799 is also merged?

package/gluon-mesh-babel/src/respondd.c
```diff
@@ -620,6 +620,7 @@ static struct json_object * get_wifi_neighbours(const char *ifname) {
 		json_object_object_add(obj, "signal", json_object_new_int(entry->signal));
 		json_object_object_add(obj, "noise", json_object_new_int(entry->noise));
 		json_object_object_add(obj, "inactive", json_object_new_int(entry->inactive));
+		json_object_object_add(obj, "iface", json_object_new_string(ifname));
```
NeoRaider: For Babel, the link-local IP address should be used.

NeoRaider:
> As neoraiders #1879 is now merged, should I wait till christfs #1799 is also merged?

I don't think so - #1799 is currently blocked on a major libbabelhelper rework (https://github.com/christf/libbabelhelper/pull/9) and will require significant changes afterwards. Feel free to get this PR in shape, and we can get it in first.

rotanid:
@lemoer ping, seems like something small we get still get into v2020.2

lemoer:
@rotanid I currently do not have a babel site config to test this with babel. That's the reason why this is stuck.

NeoRaider:
I have just checked the code and found that this is redundant: The neighbours are already nested in an object that has the interface BSSID as its key. No special handling is necessary for Babel after all.

PR #1918, opened 2020-01-14
ramips-mt76x8: add support for TP-Link TL-WA801ND v5

### LINKED ISSUE #1906: add Support for TP-Link WA801ND v5.0
hello, i want to support this Freifunk.
Since about 1 year this device laying here and waiting to be used as a freifunk-router. When are you finally creating an update for this?

nalxnet:
  - [ ] webinterface: does not work on v5 and will brick the device according to owrt wiki
  - [x] tftp: place renamed bootloader image (tp_recovery.bin) at tftp root, reachable at 192.168.0.66. Put device into tftp recovery by pressing reset button while booting until lock LED turns on. Device pulls firmware, flashes it and restarts. (see https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=ce31bdc20c3f2eb387a84c277dff76324a8e39c7)
  - [ ] other: none
    - [x] must keep/forget configuration (if applicable): sysupgrade w/ and w/o -n tested
  - [x] must have working autoupdate: checked device name: `tp-link-tl-wa801nd-v5`
- [x] reset/wps/phone button must return device into config mode: reset button does the trick
- wired network
  - [x] must have correct port assignment (WAN/LAN): only one WAN port
- wifi (if applicable)
- led mapping
  - radio leds
    - [x] should show activity
  - switchport leds
- no outdoor devices

Running test device: http://[2001:678:6e3:1110:724f:57ff:fe86:4c72]/cgi-bin/status

Not beautiful, but no functional impact: The lock LED (green/red) at the front stays red whenever the device is running. If there are objections I can look further into this, but I would need some help with that, at least pointing me into the right direction.

Fixes #1906

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -300,7 +300,7 @@ ramips-mt7620
 
   - WT3020AD/F/H
 
-* TP-Link  
+* TP-Link
 
   - Archer C2 v1
   - Archer C20i
@@ -350,6 +350,7 @@ ramips-mt76x8
 * TP-Link
 
   - TL-MR3420 v5
+  - TL-WA801ND v5
   - TL-WR841N v13
   - TL-WR902AC v3
   - Archer C50 v3
--- a/targets/ramips-mt76x8
+++ b/targets/ramips-mt76x8
@@ -36,6 +36,13 @@ device('tp-link-tl-mr3420-v5', 'tplink_tl-mr3420-v5', {
 	},
 })
 
+device('tp-link-tl-wa801nd-v5', 'tplink_tl-wa801nd-v5', {
+	factory = false,
+	extra_images = {
+		{'-squashfs-tftp-recovery', '-bootloader', '.bin'},
+	},
+})
+
 device('tp-link-tl-wr841n-v13', 'tl-wr841n-v13', {
 	factory = false,
 	extra_images = {


blocktrron:
> Not beautiful, but no functional impact: The lock LED (green/red) at the front stays red whenever the device is running. If there are objections I can look further into this, but I would need some help with that, at least pointing me into the right direction.

If you change the bridhtness value in the sysfs - does the LED state change?
```
echo 0 > /sys/class/led/LEDNAME/brightness
echo 1 > /sys/class/led/LEDNAME/brightness
```

That said - There is only one color for the LED defined in OpenWrt. Maybe look into the GPL dump of the device and see, if the other color is defined there.

nalxnet:
Thanks for the hints!

Brightness change works:
off: `echo 1 > /sys/class/leds/tl-wa801nd-v5\:orange\:wps/brightness`
on: `echo 0 > /sys/class/leds/tl-wa801nd-v5\:orange\:wps/brightness`

GPL sources show that the colors are configured as two LEDs, so I guess it needs to be patched in owrt to enable both colors. Sources show `GPIO43:WPS RED` which is `gpios = <&gpio1 11 GPIO_ACTIVE_LOW>;` in owrt. According to the other GPIO mappings the `GPIO42:WPS GREEN` from sources should be added as `gpios = <&gpio1 10 GPIO_ACTIVE_LOW>;` in owrt.

blocktrron:
@nalxnet This seems like the red LED is incorrectly flagged as ACTIVE_LOW and should rather be flagged as ACTIVE HIGH. writing 1 to brightness should turn of the LED, not off.

When you add the green color, make sure to add the `p1led_an` to the pinmux group, as otherwise the pin can't be used in GPIO mode.

nalxnet:
> @nalxnet This seems like the red LED is incorrectly flagged as ACTIVE_LOW and should rather be flagged as ACTIVE HIGH. writing 1 to brightness should turn of the LED, not off.

Indeed: I just flashed a new testbuild with a patched owrt, now the WPS LED (formerly known as lock LED) works as expected (red and green).

> When you add the green color, make sure to add the `p1led_an` to the pinmux group, as otherwise the pin can't be used in GPIO mode.

I don't know how to do this yet, this is my first contribution to gluon/owrt, no prior experience. I just added the GPIO definition in the device DTS and it works with no other modifications (see https://github.com/nalxnet/openwrt/commit/06ad8d141af4e8bfcd55c2bde032698135a55fd3).

blocktrron:
@nalxnet This because the bootloader leaves the `p1led_an` pin in this state. However, the complete state of the multiplexer has to be defined in the DTS.

Simply add the `p1led_an` here: https://github.com/openwrt/openwrt/blob/master/target/linux/ramips/dts/mt7628an_tplink_tl-wa801nd-v5.dts#L56

Details on the GPIO pinmux can be found in the MT7628 datasheet (Google) on page 23

nalxnet:
> Details on the GPIO pinmux can be found in the MT7628 datasheet (Google) on page 23

Thanks for your help @blocktrron , now I get where the `p1led_an` comes from. I'll also add the missing WPS button (guess this should also be added to the group as `uart1` - it's at GPIO46 according to the GPL sources), do a fresh owrt build and prepare a PR for owrt master.

How do we proceed here? Wait for the changes to get added to owrt, although this has no functional impact for gluon?

blocktrron:
@nalxnet Yes please. As soon as we hit OpenWrt master with this, we can backport this change (or it gets picked to OpenWrt 19.07).

LEDs are considered as an important component for Gluon, as it is the only way for the node to communicate it's status to the user.

nalxnet:
PR for owrt master is here: https://github.com/openwrt/openwrt/pull/2699

rotanid:
as the OpenWrt PR is merged now, you may rebase it onto openwrt 19.07 and format it for Gluon to be part of this PR

blocktrron:
@rotanid I've picked the commit to 19.07.

@nalxnet Please rebase your PR on the current Gluon master and retest.

PR #1924, opened 2020-01-26
ramips-mt7621: add support for Netgear EX6150

Noodlesalat:
  - [x] webinterface: webinterface was available under http://192.168.1.250/
  - [ ] tftp
  - [ ] other: <specify>
    *usually means: gluon profile name must match image name* model name: *netgear-ex6150*
- wired network
  - [x] must have correct port assignment (WAN/LAN): device has only one LAN-port
- wifi (if applicable)
- led mapping
  - radio leds
    - [x] should show activity
  - switchport leds
- outdoor devices only

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -324,6 +324,7 @@ ramips-mt7621
 
 * NETGEAR
 
+  - EX6150
   - R6220
 
 * Ubiquiti
--- a/targets/ramips-mt7621
+++ b/targets/ramips-mt7621
@@ -12,6 +12,10 @@ device('d-link-dir-860l-b1', 'dir-860l-b1')
 
 -- Netgear
 
+device('netgear-ex6150', 'netgear_ex6150', {
+        factory_ext = '.chk',
+})
+
 device('netgear-r6220', 'r6220', {
 	factory_ext = '.img',
 })


docs/user/supported_devices.rst
```diff
@@ -324,6 +324,7 @@ ramips-mt7621
 
 * NETGEAR
 
+  - EX6150
```
mweinelt: Should we explicitly call that v1?

Noodlesalat: "v1" is not mentioned anywhere and only "EX6150" is written on the device itself.
So i would say, no.

rotanid: i would say yes, as it would avoid confusion with the v2

Noodlesalat: I dont want to be a dealbreaker. If you want to add the "v1", go ahead. :)

rotanid: done

blocktrron: The situation with Netgear is a bit "problematic", as the v1 is not explicitly labled on the device. For the v2 it's not printed on the Box (v1 and v2 can be distinguished on the outside only by the MU-MIMO claim on the packaging) and on the device it's part of the model name.

So it's a compromise in some direction, regardless of the labling. Just wanted to share this insight here.

PR #1928, opened 2020-02-02
brcm2708-bcm2710: add manifest_alias for raspberry-pi-3-model-b-rev-1.2

Dark4MD:
It's the same Device as the others so it is still broken because of not working mesh on wifi but this rev is missing an alias so an nightly build could not update itself.

> autoupdater: warning: no matching firmware found (model raspberry-pi-3-model-b-rev-1.2)


NeoRaider:
It seems we should improve our `platforminfo_get_image_name` for the brcm2708-* targets rather than listing all the revision names...

PR #1930, opened 2020-02-08
gluon-check-connection proposal

CodeFetch:
It's untested and I won't have time in the near future to work on it.
It aims to simplify connectivity tests for other packages like ffho's autoupdater-wifi-fallback, scheduled-domain-switch and the offline-ssid package.

### DIFF
--- a/docs/package/gluon-check-connection.rst
+++ b/docs/package/gluon-check-connection.rst
@@ -1,0 +1,69 @@
+gluon-check-connection
+======================
+
+This package adds a script that checks if at least one connection to IPv6 hosts
+defined as *target groups* is working using the ping command. 
+The script is called once every minute by ``micrond``.
+For example one can define a group of *local* targets to check if a connection
+to hosts in the mesh network is possible (e.g. time or update servers) and 
+*global* targets for checking if a connection to the global internet is possible.
+Currently only IPv6 addresses are supported.
+This is e.g. used by the *gluon-scheduled-domain-switch* package
+
+site.conf
+---------
+
+Target groups can be pre-defined in the domain config.
+
+::
+
+    check_connection = {
+        targets = {
+	    local = {
+                'fe80::dead:c0de:1',
+                'fe80::bad:c0de:1',
+                'fe80::dead:c0de:2',
+                'fe80::bad:c0de:2',
+            },
+            global = {
+                '2620:0:ccc::2', -- OpenDNS
+                '2001:4860:4860::8888', -- Google DNS
+                '2600::1' -- Sprint DNS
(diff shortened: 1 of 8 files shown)


docs/package/gluon-check-connection.rst
```diff
@@ -0,0 +1,69 @@
...
+Defining target groups in the site.conf will overwrite existing ones with the same
+name when performing a *sysupgrade* or by triggering *gluon-reconfigure*.
+
+Configuration via UCI
+---------------------
+
+Packages can use gluon-check-connection to be triggered after connection checks.
+For this they can define the following *script* attributes:
+
+script : an entry for defining the ping target
```
CodeFetch: an entry for configuring and defining the triggered command

package/gluon-check-connection/luasrc/usr/bin/gluon-check-connection
```diff
@@ -0,0 +1,139 @@
...
+			if states[group] then
+				state_online = true
+			else
+				state_offline = true
+			end
+		end
+	end
+
+	if script['onchange'] and not state_changed then
+		return
```
CodeFetch: break instead of return

blocktrron:
Thanks for your proposal!

As this PR is not meant to be merged in it's current state I'm closing it, leaving the information there to be picked up by someone in the future.

CodeFetch:
@blocktrron I'm waiting for feedback and someone willing to test it. If you close it, it won't get any intention.

mweinelt:
> It's untested and I won't have time in the near future to work on it.

Sounds very much like it's a code dump without any intention to work on it.

CodeFetch:
It's a "proposal". As you don't comment on the concept itself, I see this as an ACK. So I'm going to test it when I have time. Yes, I don't really have an intention to work on it as I guess it's just fine as it is if it runs.

mweinelt:
We feel strongly that a package that pings out into the network to determine some state it can derive from local information has no place in the gluon core repository.

We are discussing a proposal on IRC right now with @blocktrron @AiyionPrime and @NeoRaider and will see what comes off it.

CodeFetch:
@mweinelt @blocktrron @AiyionPrime @NeoRaider  This package is supposed to be a generic test facility using ping. How can you test whether internet connection is available using "local" information only? If you would have looked at the PR you would have seen that this PR is both about performing local and global tests.

mweinelt:
We simply don't think it's reasonable for networks with 500-1500 nodes to basically ping flood any host on the internet. We think the local state, like a relatively fresh IPv6 default route, or a gateway announcement has sufficient informational value to derive a status for most use cases. 

That will be my last message on the matter.

PR #1932, opened 2020-02-12
POE on LAN, not on WAN

heis2201:
Devices with PoE on the **LAN** port will run Config Mode on the WAN port instead.

(This my first pull request, I may make some mistakes commiting this request. Also, to me it's unclear to change this for all affected releases)

### DIFF
--- a/docs/features/configmode.rst
+++ b/docs/features/configmode.rst
@@ -26,7 +26,7 @@ In general, Config Mode will be offered on the LAN ports. However, there
 are two practical exceptions:
 
 * Devices with just one network port will run Config Mode on that port.
-* Devices with PoE on the WAN port will run Config Mode on the WAN port instead.
+* Devices with PoE on the LAN port will run Config Mode on the WAN port instead.
 
 
 Accessing Config Mode


mweinelt:
What device are you talking about specifically?

heis2201:
Hi,
this is not regarding a specific device. I came across this when reading the documentation about Config Mode [1]. There is says: _Devices with PoE on the WAN port will run Config Mode on the WAN port instead._

However, the sentence seems to make little to me since because I am not aware of anything like PoE on a WAN port. I am assuming that the meaning is supposed to be that if you have POE-LAN ports, your config interface would be available on the WAN port.

Cheers!

[1]: https://gluon.readthedocs.io/en/v2018.2.1/features/configmode.html#port-configuration

blocktrron:
@heis2201 The CPE210 has PoE on the WAN port. 

The rationale being, that a CPE210 mounted on a tower can be brought to setup-mode by pressing the remote reset button.

heis2201:
I see... so suddenly it makes. Trash my request please!

blocktrron:
Glad we clarified this.

I will close this PR. Thanks for attempting to improve Gluon despite of the outcome :+1:

PR #1933, opened 2020-02-17
scripts: do not create gluon_base entry in feeds.conf

SvenRoederer:
We started to reuse some work of your framework in our framework, so we like to give something back, which finally makes contributing patches directly more easy.

The gluon-script "feeds.sh" creates by default a "gluon_base" entry in the feeds.conf
of OpenWrt. This links to the gluon-base packages directory and causes an error message
during build, as this directory does not exist in our non-gluon firmware.
Add a test for definiton of a "FOREIGN_BUILD" variable to the feeds.sh script, which
will no add the mentioned line to the feeds.conf, when defined.

This PR is taken from https://github.com/freifunk-berlin/firmware/commit/013a01b0c5bb34b4c46e1839f70cd07787379377

### DIFF
--- a/docs/user/getting_started.rst
+++ b/docs/user/getting_started.rst
@@ -229,3 +229,7 @@ GLUON_OUTPUTDIR
 
 GLUON_SITEDIR
   Path to the site configuration. Defaults to ``site``.
+
+FOREIGN_BUILD
+  If defined it changes the build-framework into a more generic mode, which will not necessarily 
+  generate a Gluon compatible firmware.
--- a/scripts/feeds.sh
+++ b/scripts/feeds.sh
@@ -11,7 +11,7 @@ rm -rf openwrt/feeds
 rm -rf openwrt/package/feeds
 
 (
-	echo 'src-link gluon_base ../../package'
+	[ -n "${FOREIGN_BUILD}" ] || echo 'src-link gluon_base ../../package'
 	for feed in $FEEDS; do
 		echo "src-link $feed ../../packages/$feed"
 	done


blocktrron:
I think this would be worth mentioning briefly in the documentation.

SvenRoederer:
That sound doable  ....

SvenRoederer:
added a note to getting_started.rst

SvenRoederer:
This works somehow in the direction @CodeFetch mentoined in https://github.com/freifunk-gluon/gluon/issues/1936#issuecomment-589549987

NeoRaider:
Sorry for taking so long to review such a tiny change - I've looked at it several times, and I'd really like to find a way to remove the special "gluon_base" handling from feeds.sh altogether, making your desired behaviour the default. In fact, I just got an idea how to do that...

In any case, getting-started is not a good doc page a describe an internal feature like this - it might even make sense to start a new page for such internals.

NeoRaider:
I've implemented a more generic solution that should cover your usecase in #1950.

SvenRoederer:
The overall rewrite of PR #1950 should do the some as intended by this PR. The new ```docs/dev/build.rst``` file seems more appropriate than than the ``getting-started.rst``` I also felt not well fitting.

Looking forward to see this merged.

mweinelt:
Hey @SvenRoederer, can you say if #1950 obsoleted this pull request?

SvenRoederer:
> Hey @SvenRoederer, can you say if #1950 obsoleted this pull request?

Should do. Thanks.

PR #1938, opened 2020-02-25
Dockerfile: Install squashfs-tools

geistesk:
Building Gluon 2020.1 failed due to the lack of the binary
mksquashfs-lzma. This occurred while building ar71xx-tiny.

### DIFF
--- a/contrib/Dockerfile
+++ b/contrib/Dockerfile
@@ -9,6 +9,7 @@ RUN apt update && apt install -y --no-install-recommends \
     build-essential \
     gawk \
     unzip \
+    squashfs-tools \
     libncurses5-dev \
     zlib1g-dev \
     libssl-dev \


NeoRaider:
Please post a full log of the failed build. OpenWrt brings its own variant of mksquashfs-lzma that must be used.

geistesk:
I no longer have a failed log, but I was able to fix the error of the missing binary by installing it. However, I can reproduce it again.

geistesk:
I reran my whole build within the old container and did not experience the same error, which I thought I had fixed with this PR. Thus, I cannot reproduce it and cannot give you any logging information. IIRC, the error was the absence of `mksquashfs-lzma` in the `gluon/openwrt/staging_dir/host/bin/` directory.

There seems to be some hardly reproducible bug, but my supposed fix has nothing to do with it. Perhaps the error occurred because of the parallel building by running `make` with `-j 4`.

Sorry for this and thank you very much for your efforts.

PR #1946, opened 2020-03-06
build: add support for OpenWrt-Imagebuilder and non-gluon style firmware-images

SvenRoederer:
For Freifunk Berlin we are in need of a way to define some device-specific packages (https://github.com/freifunk-berlin/firmware/issues/696). Instead of inventing a new way for such definition, a PR was made to reuse code of the gluon build-framework (https://github.com/freifunk-berlin/firmware/pull/765). This PR is about upstreaming the changes.

The Freifunk Brerlin firmware differs in some points from Freifunk Gluon. The major points are:
1. a completely different network-topology (OLSR-based meshing)
2. LuCI based web-frontend
3. Firmware-images are created with the OpenWrt Imagebuilder

Based on 1.) and 2.) there is (currently) no use of the gluon-packages, but some other external feeds. There are also some CONFIG-options of OpenWrt which are incompatible between both firmwares. For  this reason the file ```targets/generic_gluon``` was split from ```targets/generic``` and an additional ```targets/generic_ffberlin``` was added. The ```generic```-file contains just some very common settings and the ```generic_*```-files provide the build-type specific settings. With the  ```FOREIGN_BUILD```-flag the Makefile can control which ```generic-*```-file is included in addition to the ```generic```-file.

Imagebuilder-support:

Freifunk Berlin is using a two-phases based process (compilation, image-assembly) of creating the firmware-images. In the 1st phase all relevant packages are build and an imagebuilder.tar.xz is created. 

In the 2nd phase this OpenWrt-Imagebuilder is used to assemble the firmware-images by combining the previously built kernel and packages. As for Gluon the list of packages is generated from ```targets/generic*```, ```targets/$GLUON_TARGET```, the Makefile and the site-conf. In contrast to Gluon this list of packages is not fix stored in the OpenWrt-config on a per board-basis, as it needs to be provided to the imagebuilder later.

This is archived by the new ```scripts/target_config_profile.lua```, which generates a per GLUON_TARGET file containing the packagelist per board. This script uses the new function 'target_config_lib.lua.get_pkglist()' which is reusing the existing code of building the packagelist. 

---
The code of this PR is just cherry-picked from the above mentioned PR and currently a working prototype to be commented.It also still contains a lot of debug-output.

A welcome side-effect of the original PR is that it will be more easy to share code between both projects regarding the build-framework and device-support. As mentioned I only had to cherry-pick from the Freifunk-Berlin Repo without copy-paste of code.


rotanid:
@SvenRoederer because of #1950 you need to rebase this, please.
also please check first, if 1950 solves your request in #1933

SvenRoederer:
@rotanid  Rebasing is on my ToDo since some times, after having seen PR1950 was merged. But It seems the idea of this PR will get no veto, so it's worth spending my time on it again.

NeoRaider:
I think this PR does far too many things at once - I'd like to see it as inspiration of things we should support eventually, not as a concrete proposal for an implementation. So IMO rebasing this is rather low-priority, if useful at all.

CodeFetch:
> I think this PR does far too many things at once - I'd like to see it as inspiration of things we should support eventually, not as a concrete proposal for an implementation. So IMO rebasing this is rather low-priority, if useful at all.

@NeoRaider From the comment it is a request for comments and I think it is really well-thought-through, but I admit it's a bit much to grasp. Unfortunately I don't know the Makefiles that well, because I'd love to see this. Please have a look at it.

CodeFetch:
@SvenRoederer Can you please have a look at PR #2172? I think it will be cleaner if you could somehow adapt the Freifunk Berlin image builder to this approach. What other conflicts in the generic config file do you see? Can you make them configurable like GLUON_SPEZIALIZE_KERNEL with your packages so that you don't need a custom file for Freifunk Berlin anymore?

SvenRoederer:
@CodeFetch @lemoer  slowly keeping up with my time and happy to see PR #2172. Will have a close look at it, as it seem to provide half of the way this PR a going for.

mweinelt:
Marking as draft, given the fact that we can't, won't merge this as is.

blocktrron:
Closing as per successor https://github.com/freifunk-gluon/gluon/pull/2172

PR #1947, opened 2020-03-06
add common WiFi packages to relevant targets

blocktrron:
This commit adds packages for supporting common wireless adapters for Raspberry Pi and x86 platforms.

People often want to attach USB adapters to such platforms (whether or not this is a rational thing to do). Currently the packages have to installed after the initial boot process, which breaks Gluons initialization.

As those platforms have enough storage, include common USB adapter driver packages by default. Also extend the list of PCIe driver packages on x86 to include drivers and firmware for more WiFi cards available on the market.

This PR is not yet compile nor run-tested (as i do not own such adapters). So feel free to provide your experiences.

### DIFF
--- a/targets/x86.inc
+++ b/targets/x86.inc
@@ -34,5 +34,11 @@ packages {
 	'kmod-usb-serial',
 	'kmod-ath10k',
 	'ath10k-firmware-qca9887',
+	'ath10k-firmware-qca9888',
 	'ath10k-firmware-qca988x',
+	'ath10k-firmware-qca9984',
+	'kmod-mt76x0e',
+	'kmod-mt76x2',
+	'kmod-mt7603',
+	'kmod-mt7615e',
 }


mweinelt:
We might also need to add `usb-modeswitch` to fix quirks like this:

```
ID 057c:62ff AVM GmbH AVM Fritz!WLAN USB (in CD-ROM-mode)
```

I'm currently testing if adding the package suffices to make that work.

rotanid:
interesting, some of them we don't include in our community, but we include others:
https://github.com/tecff/site-ffa/blob/experimental/site.mk#L123
though it's mostly cable network cards, not wifi

NeoRaider:
USB wireless devices work poorly with Gluon at the moment, in particular because they are enumerated too late, so our upgrade scripts may not catch them. For that reason, we should leave out USB wireless drivers for now, so people don't break their configurations.

blocktrron:
In this case - I've removed the USB packages at all.

blocktrron:
Will merge this as soon as I've done a build-test.

targets/x86.inc
```diff
@@ -34,5 +34,11 @@ packages {
 	'kmod-usb-serial',
 	'kmod-ath10k',
 	'ath10k-firmware-qca9887',
+	'ath10k-firmware-qca9888',
 	'ath10k-firmware-qca988x',
+	'ath10k-firmware-qca9984',
+	'kmod-mt76x0e',
+	'kmod-mt76x2'
```
mweinelt: comma

mweinelt:
LGTM. Thanks!

PR #1948, opened 2020-03-06
ipq40xx-generic: add support for EnGenius ENS620EXT

nalxnet:
  - [x] webinterface: there are two factory images. you need to choose depending on the currently installed vendor firmware.
  - [ ] tftp
  - [ ] other: <specify>

```
root@ffka-lr-naturbursche:~# lua -e 'print(require("platform_info").get_image_name())'
engenius-ens620ext
```

- [x] primary mac should match address on device label (or packaging) (https://gluon.readthedocs.io/en/latest/dev/hardware.html#notes): matches mac of 2.4HGz wifi (3 mac addresses, lan/2.4/5, on label)
- wired network
- wifi (if applicable)
- led mapping
  - radio leds
    - [x] should show activity
  - switchport leds
- outdoor devices only

As I wrote above, there are two different factory images because the format changes with vendor firmware 3.5.
I could not find a similar case for current devices so my proposal: ~~Use image for versions below 3.5 as the default since the ap should come preinstalled with this version. Offer the other one as "other" image.~~ Let's offer both factory images as "other" images to avoid confusion.

### DIFF
--- a/targets/ipq40xx-generic
+++ b/targets/ipq40xx-generic
@@ -43,6 +43,17 @@ device('avm-fritz-repeater-1200', 'avm_fritzrepeater-1200', {
 })
 
 
+-- EnGenius
+
+device('engenius-ens620ext', 'engenius_ens620ext', {
+	factory = false,
+	extra_images = {
+		{'-squashfs-factory_30', '-factory_fw30', '.bin'},
+		{'-squashfs-factory_35', '-factory_fw35', '.bin'},
+	},
+})
+
+
 -- GL.iNet
 
 device('gl.inet-gl-b1300', 'glinet_gl-b1300', {
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -223,6 +223,10 @@ ipq40xx-generic
   - FRITZ!Box 4040 [#avmflash]_
   - FRITZ!Repeater 1200 [#eva_ramboot]_
 
+* EnGenius
+
+  - ENS620EXT
+
 * GL.iNet
 
   - GL-B1300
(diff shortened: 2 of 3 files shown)


targets/ipq40xx-generic
```diff
@@ -43,6 +43,17 @@ device('avm-fritz-repeater-1200', 'avm_fritzrepeater-1200', {
...
 
 
+-- EnGenius
+
+device('engenius-ens620ext', 'engenius_ens620ext', {
+	factory = '-squashfs-factory_30',
+	factory_ext = '.bin',
+	extra_images = {
+		{'-squashfs-factory_35', '-factory_35', '.bin'},
+	},
```
blocktrron: In tis special case, we should put both factory image with their indication as an extra image.

nalxnet: Thought it would be nice to offer a "factory" image, but I can see that this will be confusing to some degree. Changed it as suggested.

targets/ipq40xx-generic
```diff
@@ -43,6 +43,17 @@ device('avm-fritz-repeater-1200', 'avm_fritzrepeater-1200', {
 })
 
 
+-- EnGenius
+
+device('engenius-ens620ext', 'engenius_ens620ext', {
+	factory = false,
+	extra_images = {
+		{'-squashfs-factory_30', '-factory_30', '.bin'},
+		{'-squashfs-factory_35', '-factory_35', '.bin'},
```
mweinelt: Minor nit: Would it be helpful if we labeled them fw30 and fw35, to indicate that it's related to the firmware version?

herbetom: The need for multiple targets for diffrent firmware versions is thankfully not very common. So I would find it helpful to make it extra clear.

nalxnet: Sure, why not. Renamed to _fw30 and _fw35.

PR #1949, opened 2020-03-07
docs: add v2020.1.1 release notes

blocktrron:


docs/releases/v2020.1.1.rst
```diff
@@ -0,0 +1,17 @@
+Gluon 2020.1.1
+==============
+
+This is the first service release for the Gluon 2020.1.x line, fixing bugs reported by the community.
```
mweinelt:
```suggestion
This is the first service release for the Gluon 2020.1.x line, fixing regressions reported by the community.
```

docs/releases/v2020.1.1.rst
```diff
@@ -0,0 +1,17 @@
...
+
+- Fixed non-working LEDs on TP-Link Archer C5 v1 and Archer C7 v2 after an upgrade to Gluon 2020.1.
+
+- Fixed an issue which leads to AVM FRITZ!WLAN Repeater 450E devices being stuck in failsafe mode
+  ater an upgrade to Gluon 2020.1.
+
+Other changes
+-------------
+
+- Linux kernel has been updated to 4.14.171
```
mweinelt: Missing known issues.

blocktrron: Well, we missed these with the v2020.1 also. But I'll add them back here

mweinelt: Also add the regression we know about now.

blocktrron: I've included the ER-X one, are there more?

mweinelt: That depends on if we want to add the regressions fixed in this release.

rotanid: if we missed them in v2020.1 we should add them to the old release notes now too now that we noticed it

mweinelt: Fixed in c219d3c1dc6e5612273daf3f87437df526c400a3 and 9e4eb182908320cb94974663c7d201ce792eea78.

PR #1951, opened 2020-03-07
add .editorconfig

### LINKED ISSUE #1929: Add editorconfig to reflect our coding style
quick issue, no text yet.

mweinelt:
Resolves #1929

Not a big deal, mostly just codifying what we already agreed on.

Did I mess any exceptions?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+# Top-most EditorConfig file
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+indent_style = tab
+charset = utf-8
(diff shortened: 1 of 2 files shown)


.editorconfig
```diff
@@ -0,0 +1,8 @@
+# Top-most EditorConfig file
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = false
```
NeoRaider: I think final newlines should always be included?

mweinelt: I'm not opposed.

mweinelt:
I intend to merge this tonight and we can then iterate on it further when we notice oddities.

Issue #1952, opened 2020-03-08
tunneldigger-watchdog often says "Process-Pid does not match with pid-File."


RalfJung:
### Bug report

I noticed that I am seeing the tunneldigger-watchdog kick in with "Process-Pid does not match with pid-File" much more often that I would expect, and even when tunneldigger and the tunnel are in perfect health as far as I can see.

**What is the problem?**
tunneldigger-watchdog restarts tunneldigger even when there is no reason to do so, leading to unnecessary brief connectivity loss.

**What is the expected behaviour?**
tunneldigger-watchdog should not unnecessarily restart tunneldigger.

**Problem analysis:**
Looking at the watchdog, it seems to determine the current tunneldigger PID by looking at the first line of `pgrep tunneldigger`. However, that will show *all* processes with tunneldigger anywhere in the binary name. Another such process is `tunneldigger-watchdog` itself. My guess for what happens is that whenever the watchdog happens to have a PID lower than tunneldigger itself, it sees a wrong PID and then restarts tunneldigger.

I think the watchdog should switch to `pgrep -x /usr/bin/tunneldigger` to make sure that it only finds processes with that binary, not other processes with "tunneldigger" in their name.

**Gluon Version:** v2019.1.2

**Site Configuration:** https://git.hacksaar.de/FreifunkSaar/gluon-site/-/blob/master/site.conf

**Custom patches:** https://git.hacksaar.de/FreifunkSaar/gluon-ffsaar

mweinelt:
This change sounds sensible enough, can you create a pull request?

RalfJung:
Done: https://github.com/freifunk-gluon/gluon/pull/1953

PR #1953, opened 2020-03-08
tunneldigger-watchdog: only search for exactly the tunneldigger binary

RalfJung:
Fixes https://github.com/freifunk-gluon/gluon/issues/1952

### DIFF
--- a/package/gluon-mesh-vpn-tunneldigger/luasrc/usr/bin/tunneldigger-watchdog
+++ b/package/gluon-mesh-vpn-tunneldigger/luasrc/usr/bin/tunneldigger-watchdog
@@ -33,7 +33,7 @@ local function has_mesh_vpn_neighbours()
 end
 
 if uci:get_bool('tunneldigger', 'mesh_vpn', 'enabled') then
-	if io.popen('pgrep tunneldigger'):read('*l') ~= read_pid_file() then
+	if io.popen('pgrep -x /usr/bin/tunneldigger'):read('*l') ~= read_pid_file() then
 		os.execute('logger -t tunneldigger-watchdog "Process-Pid does not match with pid-File."')
 		restart_tunneldigger()
 		return


mweinelt:
Please use the package name as a prefix: gluon-mesh-vpn-tunneldigger

RalfJung:
@mweinelt  Done.

mweinelt:
Backported to v2020.1.x in c4d38a2ea9f4624315cf201bbabf6b56e996cf5d and v2019.1.x in 12221d442fdbbe245dbd270535b2bcb49dae1340.

package/gluon-mesh-vpn-tunneldigger/luasrc/usr/bin/tunneldigger-watchdog
```diff
@@ -33,7 +33,7 @@ local function has_mesh_vpn_neighbours()
 end
 
 if uci:get_bool('tunneldigger', 'mesh_vpn', 'enabled') then
-	if io.popen('pgrep tunneldigger'):read('*l') ~= read_pid_file() then
+	if io.popen('pgrep -x /usr/bin/tunneldigger'):read('*l') ~= read_pid_file() then
```
NeoRaider: I don't think this will ever match - without `-f`, `pgrep` only matches the process name, never its path (and whether the path is matches with `-f` depends on the way the binary was called).

NeoRaider: Two options:
- Use `pidof` instead of `pgrep`, which accepts the whole path (I have only checked a full-featured pidof though, not busybox)
- Check whether `/proc/$PID/exe` points to the right binary - which is also more efficient than using pgrep or pidof

RalfJung: I can `pgrep -x /usr/bin/tunneldigger` on two nodes and it finds 3 PIDs on each; the same it finds with `pgrep tunneldigger`.

RalfJung: I also adjusted the watchdog like here on one of the nodes and confirmed that it does not just restart tunneldigger.

So, to me it looks like this works as intended?

nudelkiste: > _Edit_: This only matches when the binary was called using the whole path.

I checked it out on several nodes with different hardware and tunneldigger was always started by using full path. In case of further trouble, I´d suggest to rename the watchdog to something different, e.g. tunneldigger-watchdog -> td-watchdog
This would help to avoid such grep-confusion, too

PR #1950, opened 2020-03-07
Build system improvements

NeoRaider:
- Improve handling of build environment
- Simplify manifest generation using .ONESHELL
- Increase control over `feeds.conf` generation (supersedes #1933)
- Add (very incomplete) docs

PR #1954, opened 2020-03-09
codestyle: replace leading spaces with tabs in target files

rotanid:


Makefile
```diff
@@ -6,10 +6,10 @@ export LC_ALL LANG
...
 # check for spaces & resolve possibly relative paths
 define mkabspath
- ifneq (1,$(words [$($(1))]))
-  $$(error $(1) must not contain spaces)
- endif
- override $(1) := $(abspath $($(1)))
+	ifneq (1,$(words [$($(1))]))
+		$$(error $(1) must not contain spaces)
+	endif
+	override $(1) := $(abspath $($(1)))
```
NeoRaider: You can't use tabs for indentation in Makefiles, tabs mark recipe lines.

NeoRaider: (see also #1950, where I touched these lines to use at least more than 1 space...)

AiyionPrime: Technically speaking, since make 3.82 one could :D
.RECIPEPREFIX is the corresponding buzzword.

But yeah, changing one space to many would be more adequate.

https://lists.gnu.org/archive/html/make-alpha/2010-07/msg00025.html

rotanid: ok i don't touch the Make file anymore in this PR now

PR #1955, opened 2020-03-09
lantiq-xway: add support for NETGEAR DGN3500B

rotanid:
not a great device, but someone had it laying around.

**TODO: the primary mac should be read from /sys/class/net/eth0.1/address which is currently not working, maybe someone has a clue?**

- [X] must be flashable from vendor firmware
  - [X] webinterface
  - [ ] tftp
  - [X] other: tftp via serial
- [X] must support upgrade mechanism
  - [X] must have working sysupgrade
    - [X] must keep/forget configuration (if applicable)
      *think `sysupgrade [-n]` or `firstboot`*
  - [X] must have working autoupdate
    *usually means: gluon profile name must match image name*
- [X] reset/wps/phone button must return device into config mode
- [ ] primary mac should match address on device label (or packaging) (https://gluon.readthedocs.io/en/latest/dev/hardware.html#notes)
- wired network
  - [X] should support all network ports on the device
  - [X] must have correct port assignment (WAN/LAN)
- wifi (if applicable)
  - [X] association with AP must be possible on all radios
  - [X] association with 802.11s mesh must be working on all radios 
  - [X] ap/mesh mode must work in parallel on all radios
- led mapping
  - power/sys led (_critical, because led definitions are setup on firstboot only_)
    - [X] lit while the device is on
    - [X] should display config mode blink sequence 
(https://gluon.readthedocs.io/en/latest/features/configmode.html)
  - radio leds
    - [X] should map to their respective radio
    - [X] should show activity
  - switchport leds
    - [X] should map to their respective port (or switch, if only one led present) 
    - [X] should show link state and activity
- outdoor devices only
  - [-] added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`

PR #1956, opened 2020-03-12
Primary MAC logic refactoring

NeoRaider:
- Remove redundant code by switching to a match table listing the targets and boards for each candidate for the primary MAC interface.
- Add some flexiblity by allow to switch out the sysfs file data source for the MAC address (needed for #1955)


rotanid:
i tested this PR with the Netgear DGN3500B in PR #1955 

also, i checked that all devices were correctly transferred from the old solution by comparing the code (obviously not by building and testing all the devices)

PR #1962, opened 2020-03-26
ath79-generic: add support for GL.iNet GL-AR750S

nalxnet:
  - [ ] webinterface: accepts only nand firmware. nand support comes with kernel 4.19 (next openwrt release)
  - [ ] tftp
  - [x] other: flash nor firmware from u-boot webinterface: reboot with reset button pressed and lan cable plugged in. 5G led will blink 5 times. then 2G led will turn on. release reset button and connect to u-boot web at http://192.168.1.1/index.html. upload sysupgrade image.
```
root@ffka-lr-minimesh:~# lua -e 'print(require("platform_info").get_image_name())'
gl.inet-gl-ar750s-nor
```
- wired network
- wifi (if applicable)
- led mapping
  - radio leds
    - [x] should show activity
  - switchport leds: device has only power and radio leds
    - n/a should map to their respective port (or switch, if only one led present) 
    - n/a should show link state and activity
- outdoor devices only

Running test device: http://[2001:678:6e3:1110:9683:c4ff:fe00:a7da]/cgi-bin/status

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -196,6 +196,7 @@ ath79-generic
 * GL.iNet
 
   - GL-AR300M-Lite
+  - GL-AR750S
 
 * OCEDO
 
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac
@@ -82,6 +82,7 @@ local primary_addrs = {
 			'hiveap-121',
 		}},
 		{'ath79', 'generic', {
+			'glinet,gl-ar750s-nor',
 			'ocedo,raccoon',
 		}},
 		{'ipq40xx', 'generic', {
--- a/patches/openwrt/0009-ath79-enable-GL-AR750S-NOR-variant-from-master.patch
+++ b/patches/openwrt/0009-ath79-enable-GL-AR750S-NOR-variant-from-master.patch
@@ -1,0 +1,72 @@
+From 0d5a95cac5aada71108e2178e0ec5b2a6efdc415 Mon Sep 17 00:00:00 2001
+From: Jan Alexander <jan@nalx.net>
+Date: Tue, 31 Mar 2020 21:50:28 +0200
+Subject: [PATCH] ath79: enable GL-AR750S NOR variant from master
+
+---
+ target/linux/ath79/base-files/etc/board.d/02_network          | 2 +-
+ .../ath79/base-files/etc/hotplug.d/firmware/11-ath10k-caldata | 2 +-
+ target/linux/ath79/dts/qca9563_glinet_gl-ar750s.dts           | 4 ++--
+ target/linux/ath79/image/generic.mk                           | 4 ++--
+ 4 files changed, 6 insertions(+), 6 deletions(-)
+
+diff --git a/target/linux/ath79/base-files/etc/board.d/02_network b/target/linux/ath79/base-files/etc/board.d/02_network
+index 8edd4e6c2e..2d2d4e33af 100755
+--- a/target/linux/ath79/base-files/etc/board.d/02_network
++++ b/target/linux/ath79/base-files/etc/board.d/02_network
+@@ -140,7 +140,7 @@ ath79_setup_interfaces()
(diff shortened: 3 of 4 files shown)


mweinelt:
Thank you for your pull request. Since then device in question is also availablle on the ath79 target we will only accept it there, since that saves us the effort of researching, planning and compiling a migration strategy for that device when ar71xx is finally deprecated.

So can you please rebase and test against ath79?

nalxnet:
> Since then device in question is also availablle on the ath79 target we will only accept it there, since that saves us the effort of researching, planning and compiling a migration strategy for that device when ar71xx is finally deprecated.

It's not as simple as that. Official support for the GL-AR750S in 19.07 is only for ar71xx, full support for ath79 will come with the next OpenWrt release (see https://github.com/openwrt/openwrt/commit/5fc63ec8a496066d101de41ae4f820c5c4e39c04). We could re-enable image generation for ath79, but as this was disabled for a reason in OpenWrt (painless and tested upgrade path), I would rather not do that and stick with ar71xx for now. There will be lots of other devices that need to be migrated anyway. Also someone already put a lot of work in testing possible migration scenarios (see comments in https://github.com/openwrt/openwrt/pull/2184).

mweinelt:
We are already actively using ath79 and the devices on that target pass our checklist, so there is no reason to invest any more time in ar71xx.

https://github.com/freifunk-gluon/gluon/blob/master/targets/ath79-generic

nalxnet:
Let's summarize:
- The device has 16 MB NOR and 128 MB NAND flash
- "official" support for the device in OpenWrt 19.07 is in ar71xx target (NOR flash only).
- "inofficial" support (disabled image generation) in OpenWrt 19.07 present for ath79 target (NOR flash only).
- Full support for both NOR and NAND (needs 4.19 kernel) flash in next OpenWrt release. Two separate images (root filesystem NOR or NAND, kernel is always in NOR flash) as two device profiles.

So we have three options:
1. Support the device in ar71xx now and migrate to ath79 when ar71xx is finally deprecated.
2. Support the device in ath79 now. This requires patching OpenWrt to re-enable image generation. Might also require some migration when updating OpenWrt to the next release (two device profiles).
3. Do not support the device now. Wait until we use the next OpenWrt release and support both profiles then.

As requested on IRC, please decide which is the way to go. Before this is clear I won't invest more time in this. (CC @mweinelt and @blocktrron)

blocktrron:
Okay, I've dug a bit deeper int into the state of the GL-AR750S in ath79. In fact, the NOR profile seems to match the ar71xx flavor in terms of flash usage. As the device has 16M of NOR flash (which is plenty), i would propose the following solution:

Patch the ath79 variant in 19.07 downstream to match the NOR variant from master. This way, we have a consistent upgrade path and can support the NOR / NAND target of the device as son as we support Kernel 4.19.

I do not think 1 is a good solution, as ar71xx is legacy already. I also don't see a reason not to support the device if we can get rid of the downstream patches later on.

Any thought on that @mweinelt

mweinelt:
WFM.

nalxnet:
I migrated device support to ath79 and patched OpenWrt 19.07 to enable image generation and to have the same profile name as the NOR variant currently has in OpenWrt master.

Re-tested everything and made sure it still works. I also successfully tested upgrading from this 4.14-kernel image to the NOR 4.19-kernel image using sysupgade. The NOR 4.19-kernel image was built using the next branch with reverted commit to switch to kernel 5.4.

Two notable things:
1. The device will move from ath79-generic to the new ath79-nand target when switching to current OpenWrt.
2. Flashing the GL-AR750S from u-boot web is limited to a 2 MB kernel size (see https://github.com/openwrt/openwrt/commit/e96bfaedf3a342750d92fbfa773017e40e34bec3). I was unable to build the next branch with kernel 5.4 because of the failed check (size exceeded). However, when switching gluon to kernel 4.19, we can support both NOR and NOR/NAND variants, meaning that when we switch gluon to kernel 5.4, we can only support NOR/NAND for flashing from factory.

rotanid:
Thanks.
Gluon will likely never use kernel 4.19 as upstream OpenWrt decided to go straight to kernel 5.4 with the next release... it was a tight vote, but unfortunately 5.4 won.

PR #1963, opened 2020-03-26
gluon-autoupdater: stop dropbear before stopping the network

### LINKED ISSUE #1957: Stop dropbear before shutting down network on autoupdate
### Bug report

Calling the autoupdater interactively might result in an SSH session that is stuck, because the dropbear couldn't terminate the connection in time before the network was shut down.

```
root@64289-rocket-probe:/tmp/run# autoupdater 
Retrieving manifest from http://updates.ffda.io/testing/sysupgrade/testing.manifest ...
Stopping cron...
Stopping haveged...
Stopping micrond...
Stopping sysntpd...
Stopping gluon-radvd...
Stopping uhttpd...
Stopping sse-multiplexd...
Stopping gluon-respondd...
vm.drop_caches = 3
Downloading image:  5248 / 5248 KiB
*STUCK*
```

I'm aware how to terminate an SSH connection from the client side using escape sequences, but that is besides the point.

**What is the problem?**

Dropbear should be stopped before the network gets stopped, so the TCP connection can be terminated.

**Gluon Version:**
v2020.1-26-g0f05fb5c+

**Site Configuration:**

n/a

**Custom patches:**

n/a

mweinelt:
Before active SSH sessions would get stuck, when the network was
suddenly stopped. If we instead stop dropbear before that and allow it
to terminate SSH sessions cleanly we improve the user experience.

Fixes #1957

Still untested.

### DIFF
--- a/package/gluon-autoupdater/files/usr/lib/autoupdater/abort.d/15start-dropbear
+++ b/package/gluon-autoupdater/files/usr/lib/autoupdater/abort.d/15start-dropbear
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+. /lib/gluon/autoupdater/lib.sh
+
+start_enabled dropbear
--- a/package/gluon-autoupdater/files/usr/lib/autoupdater/upgrade.d/05stop-dropbear
+++ b/package/gluon-autoupdater/files/usr/lib/autoupdater/upgrade.d/05stop-dropbear
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+. /lib/gluon/autoupdater/lib.sh
+
+stop dropbear


mweinelt:
Does not seem to be working as intended.


```
# autoupdater --force-version
Retrieving manifest from http://updates.ffda.io/testing/sysupgrade/testing.manifest ...
Stopping cron...
Stopping haveged...
Stopping micrond...
Stopping sysntpd...
Stopping gluon-radvd...
Stopping uhttpd...
Stopping sse-multiplexd...
Stopping gluon-respondd...
vm.drop_caches = 3
Downloading image:  5952 / 5952 KiB
Stopping dropbear...
Stopping network...
*STUCK*
```

NeoRaider:
Stopping the dropbear main process never kills active connections. I'm not aware of any sane way to achieve that at all.

mweinelt:
Actually `killall dropbear` gets the job done, the session terminates succesfully. But that also kills the autoupdater. *sigh*

rubo77:
Maybe start Autoupdater with `nohub`

belzebub40k:
Just did a update via upgrade via sysupgrade and saw this at the end.

```
# sysupgrade ...
...
Commencing upgrade. Closing all shell sessions.
Connection to 192.168.178.35 closed by remote host.
Connection to 192.168.178.35 closed.
```

Maybe take a look at how it is down in sysupgrade.

mweinelt:
Sysupgrade seems to exit after calling ubus, which handles the sysupgrade.

```
v "Commencing upgrade. Closing all shell sessions."

COMMAND='/lib/upgrade/do_stage2'

if [ -n "$FAILSAFE" ]; then
	printf '%s\x00%s\x00%s' "$RAM_ROOT" "$IMAGE" "$COMMAND" >/tmp/sysupgrade
	lock -u /tmp/.failsafe
else
	force_attr=""
	[ $FORCE -eq 1 ] && force_attr="\"force\": true,"
	backup_attr=""
	[ $SAVE_CONFIG -eq 1 ] && backup_attr="\"backup\": $(json_string $CONF_TAR),"
	ubus call system sysupgrade "{
		\"prefix\": $(json_string "$RAM_ROOT"),
		\"path\": $(json_string "$IMAGE"),
		$force_attr
		$backup_attr
		\"command\": $(json_string "$COMMAND"),
		\"options\": {
			\"save_partitions\": $SAVE_PARTITIONS
		}
	}"
fi
```

The shell script ends here.

NeoRaider:
sysupgrade does this by replacing procd (PID1) with upgraded, which kills all other processes, so it can unmount the root filesystem.

This doesn't help us here though: Normal sysupgrade doesn't stop anything in the kernel, so the network just stays up forever, until all processes are killed.

There are ways to "fix" this - `killall dropbear` together with ignoring certain signals like SIGHUP and SIGPIPE in the autoapdater might do the trick, but to me the issue feels very similar to a shutdown/reboot of a system that you're connected to via VPN: The VPN daemon will be stopped some point, and the SSH session will hang. IMO this is expected and not worth adding any hacks.

mweinelt:
I disagree with your assesment that this is to be expected and not a bug, but I don't see an easy way to fix this. Hence closed.

PR #1964, opened 2020-03-27
build: add class-packages for targets without devices

blocktrron:
When adding device classes, targets without devices such as x86 were not
handled. As site and feature packages are included on such a per-device
decision, x86 images ended up without most packages.

Include a class setting for a target and include the class-packages
target-wide when this setting is configured.

Fixes 9c52365077be ("build: introduce device classes")


blocktrron:
Fixed the luacheck warning

blocktrron:
Verified functionality on x86-64 as well as ar71xx-generic.

PR #1966, opened 2020-03-29
github: build-test all targets using GitHub actions

blocktrron:
This PR enables build-testing all targets using GitHub actions.

GitHub provides runners for public projects free of charge. This way, we are able to build-test all Targets simultaneously, which allows to have a PR built this way in under 2 hours.

The Goal of this PR is not to remove the existing Jenkins CI, as a lock-in to GitHub is not desirable. However we are not able to gather enough resources to host this functionality ourselves.

Demo runs can be found here: https://github.com/blocktrron/gluon/actions

### DIFF
--- a/.github/workflows/build-gluon.yml
+++ b/.github/workflows/build-gluon.yml
@@ -1,0 +1,310 @@
+
+# Update this file after adding/removing/renaming a target by running
+# `make list-targets BROKEN=1 | ./contrib/actions/generate-actions.py > ./.github/workflows/build-gluon.yml`
+
+name: Build Gluon
+on:
+  push:
+  pull_request:
+    types: [opened, synchronize, reopened]
+jobs:
+
+  ar71xx-generic:
+    name: ar71xx-generic
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Build
+        run: bash contrib/actions/run-build.sh ar71xx-generic
+      - name: Archive build output
+        uses: actions/upload-artifact@v1
+        with:
+          name: output
+          path: output
+
+  ar71xx-tiny:
+    name: ar71xx-tiny
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Build
+        run: bash contrib/actions/run-build.sh ar71xx-tiny
+      - name: Archive build output
+        uses: actions/upload-artifact@v1
+        with:
+          name: output
+          path: output
+
+  ar71xx-nand:
+    name: ar71xx-nand
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Build
+        run: bash contrib/actions/run-build.sh ar71xx-nand
+      - name: Archive build output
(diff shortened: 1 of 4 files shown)


contrib/actions/generate-actions.py
```diff
@@ -0,0 +1,56 @@
...
+      - name: Build
+        run: bash contrib/actions/run-build.sh {target_name}
+      - name: Archive build output
+        uses: actions/upload-artifact@v1
+        with:
+          name: output
+          path: output
+"""
+
+TARGETS= [
```
mweinelt: Can we use `make list-targets` instead?

contrib/actions/generate-actions.py
```diff
@@ -0,0 +1,56 @@
...
+	"ar71xx-generic",
+	"ar71xx-tiny",
+	"ar71xx-nand",
+        "ar71xx-mikrotik",
```
mweinelt: Weird indentation.

modules
```diff
@@ -1,6 +1,6 @@
 GLUON_FEEDS='packages routing gluon'
 
-OPENWRT_REPO=https://git.openwrt.org/openwrt/openwrt.git
+OPENWRT_REPO=https://github.com/openwrt/openwrt.git
```
mweinelt: This switch should be explained in the commit message.

contrib/actions/generate-actions.py
blocktrron: Yes. Altered so the script accepts the input from `make list-targets`

contrib/actions/generate-actions.py
blocktrron: It's obsolete now anyways.

modules
blocktrron: Done

contrib/actions/generate-actions.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python3
+
+import sys
+
+ACTIONS_HEAD = """
```
mweinelt:
```
# Update this file after adding/removing/renaming a target by running
#  `make list-targets | ./contrib/actions/generate-actions.py > .github/workflows/build-gluon.yml`
```

blocktrron: fixed

mweinelt:
I guess there is no way to spawn jobs from `make list-targets` dynamically at runtime, no?

blocktrron:
No, jobs cannot spawn other jobs.

We could build all targets in one job, but this way we would lose concurrency.

.github/workflows/build-gluon.yml
```diff
@@ -0,0 +1,310 @@
+
+# Update this file after adding/removing/renaming a target by running
+# `make list-targets BROKEN=1 | ./contrib/actions/generate-actions.py > .github/workflows/build-gluon.yml`
```
mweinelt, line 3: To be consistent:

```suggestion
# `make list-targets BROKEN=1 | ./contrib/actions/generate-actions.py > ./.github/workflows/build-gluon.yml`
```

mweinelt:
LGTM. Thanks!

PR #1968, opened 2020-04-05
docs: compress PNG for mtu diagram

rubo77:
I reduced the size from 200kb to 66kb (by https://tinypng.com)


blocktrron:
We would still have the previous image in our repository, so i see no benefit here.

NeoRaider:
It is a slight improvement for users viewing the docs on RTD, so I'm in favor of merging this.

mweinelt:
The compression looks acceptable, I don't see any issues with this. Thanks!

PR #1970, opened 2020-04-05
gluon-web-model: update inputs on form reset

Kasalehlia:
Register to 'reset' event on form element and make call to 'update' function
delayed in 'data-update' handler to allow the form values to update beforehand.

When using a form's 'reset' button, form field visibility was not updated.
This could lead to situations where a checkbox had to be toggled again
twice to display the detail text inputs. (Example taken from private
wifi package)


mweinelt:
Thanks for your contribution, it will take some time to test this as most of us aren't too really familiar with web stuff.

That being said; i hope I get the chance to testdrive this tonight.

package/gluon-web-model/javascript/gluon-web-model.js
```diff
@@ -517,7 +517,7 @@
 		for (var i = 0, node; (node = nodes[i]) !== undefined; i++) {
 			var events = node.getAttribute('data-update').split(' ');
 			for (var j = 0, event; (event = events[j]) !== undefined; j++) {
-				bind(node, event, update);
+				bind(node, event, function () {setTimeout(update, 0);});
```
mweinelt, line 520: Can you describe what happens here? This sounds like it introduces a delay in the update callback, like something here is racy.

package/gluon-web-model/files/lib/gluon/web/view/model/form.html
```diff
@@ -1,4 +1,4 @@
-<form method="post" enctype="multipart/form-data" action="<%|url(request)%>">
+<form method="post" enctype="multipart/form-data" action="<%|url(request)%>" data-update="reset">
```
mweinelt, line 1: Where can I find documentation on `data-update`, I feel lost.

package/gluon-web-model/javascript/gluon-web-model.js
NeoRaider: I think this is fine - the "update" function should be idempotent, so as long as we have a call after each update trigger, calling it "later" (at the next opportunity - basically still immediately, with certain caveats) shouldn't be a problem.

NeoRaider:
Ugh, a form reset doesn't trigger the change events of the changed input? How annoying...

package/gluon-web-model/files/lib/gluon/web/view/model/form.html
NeoRaider: We really need some documentation of the gluon-web{,-model} internals...

All gluon-web-model form elements specify a list of "interesting" events using the data-update attribute - usually containing the "change" event. Each listed event is bound to the *update* function in the global initialization code in `gluon-web-model.js`, which is responsible for updating the visibility of dependent elements.

NeoRaider:
Anyways, LGTM - I haven't tested this myself, but it should be fine if it works for @mweinelt.

package/gluon-web-model/javascript/gluon-web-model.js
Kasalehlia: `setTimeout` with time 0 basically puts the callback immediately at the end of the event queue, so the browser has time to do any other events and handlers. Since the form's inputs are updated with the reset data _after_ the `onreset` handlers on the form, we give the browser time to do that before we can process the new values.

NeoRaider: That was my interpretation of the change as well, thanks for the confirmation.

PR #1971, opened 2020-04-05
openwrt: Store Kernel Debug-Info

SmithChart:
This change stores a Kernel with Debug-Symbols for the current
architecture in a new output directory '<outputdir>/debug'.
This allows a developer or operator of a network to store the kernel
along with the actual images. In case of a kernel oops the debug
information can be used with the script
'scripts/decode_stacktrace.sh'  in the kernel source tree to get the
names to the symbols of the stack trace.

OpenWRT already provides the CONFIG_COLLECT_KERNEL_DEBUG -option that
creates a kernel with debug-symbols in the OpenWRT output directory.
This change enables this option and copies the generated kernel to the
gluon output directory.

We found this behavior very useful since we see a Kernel oops from time to time - especially during development and in continuous testing.
Since a gluon build already generates a lot of images adding a few 100MB of kernels does not seem like a problem.

What is your opinion on this chance? If you approve this chage I would - of course - add the missing documentation.


scripts/copy_output.lua
```diff
@@ -53,6 +54,11 @@ for _, image in ipairs(lib.images) do
 	end
 end
 
+local kernel_debug_source = string.format('openwrt/bin/targets/%s/kernel-debug.tar.bz2',
+	bindir)
+local kernel_debug_dest   = env.GLUON_IMAGEDIR..'/debug/'..env.BOARD..'-'..subtarget..'-kernel-debug.tar.bz2'
```
NeoRaider, line 59: Instead of `env.BOARD..'-'..subtarget`, using the Gluon target name - found in the variable *target* - would be more appropriate for a Gluon output file.

targets/generic
```diff
@@ -44,6 +44,8 @@ config 'CONFIG_PACKAGE_ATH_DEBUG=y'
 config '# CONFIG_KERNEL_IP_MROUTE is not set'
 config '# CONFIG_KERNEL_IPV6_MROUTE is not set'
 
+config 'CONFIG_COLLECT_KERNEL_DEBUG=y'
```
NeoRaider, line 47: Does this have an effect on the build time?

scripts/copy_output.lua
SmithChart: Ok, I'll give that a try.

targets/generic
SmithChart: The kernel is compiled in `openwrt/build_dir/target-<target>/linux-<arch>/linux-<version>/vmlinux` anyway. This file gets zipped and is placed into the output directory.
Impact on the build time should be below 1s.

docs/dev/debugging.rst
```diff
@@ -0,0 +1,40 @@
+Debugging
+=========
+
+
+.. _kerneloops:
+
+Kernel Oops
+-----------
+
+Sometimes a running Linux Kernel detects an error during runtime that can not
```
NeoRaider, line 10: "kernel" is usually written with a lowercase "k".

NeoRaider, line 10: "cannot" or "can't"

docs/dev/debugging.rst
```diff
@@ -0,0 +1,40 @@
...
+be corrected.
+This usually generates a stack trace that points to the location in the code
+that caused the oops.
+
+Linux Kernels in Gluon (and OpenWRT) are stripped.
```
NeoRaider, line 15: "OpenWrt"

docs/dev/debugging.rst
```diff
@@ -0,0 +1,40 @@
+Debugging
+=========
+
+
+.. _kerneloops:
```
NeoRaider, line 5: We usually add a prefixed based on the page path to make ref labels unique, for example `_dev-debugging-kernel-oops`

docs/user/getting_started.rst
```diff
@@ -113,6 +113,12 @@ symlinks, you can use the following command to resolve these links while copying
 
     cp -rL output/images /var/www
 
+The directory `output/images/debug` contains a compressed Kernel image for each
+architecture.
+These can be used for debugging and should be stored along with the images to
+allow debugging of Kernel problems on devices in the field.
+See :ref:`Debugging<kerneloops>` for more information.
```
NeoRaider, line 120: Add a space before `<`.

scripts/copy_output.lua
```diff
@@ -53,6 +54,11 @@ for _, image in ipairs(lib.images) do
 	end
 end
 
+local kernel_debug_source = string.format('openwrt/bin/targets/%s/kernel-debug.tar.bz2',
+	bindir)
+local kernel_debug_dest   = env.GLUON_IMAGEDIR..'/debug/'..target..'-kernel-debug.tar.bz2'
```
NeoRaider, line 59: Lining up the `=` characters doesn't make much sense when the line before is wrapped.

NeoRaider, line 59: Hmm, we should prefix the destination name including the site and release (and clean up old files using a wildcard, as it is done for images and the opkg repo). It should be possible to reuse some of the opkg repo code below.

docs/dev/debugging.rst
SmithChart: Fixed. (I left the heading as it is.)

SmithChart: Fixed.

docs/dev/debugging.rst
SmithChart: I've added the prefix as you suggested and changed the link in the getting started.

docs/dev/debugging.rst
SmithChart: Fixed.

docs/user/getting_started.rst
SmithChart: Done.

scripts/copy_output.lua
SmithChart: You are totally right. I have removed the alignment.

SmithChart: I have updated my branch to the naming scheme you suggested.

For example for  a build with the following make call:
```$ make all GLUON_TARGET=ar71xx-generic V=s -j10 GLUON_RELEASE=v2019.1.1-ffbs-19-g03876975-202004192147-parker_beta GLUON_BRANCH=parker_beta GLUON_PRIORITY=0```

I now the following files:
```
$ ls output/images/factory/gluon-ffbs-next-v2019.1.1-ffbs-19-g03876975-202004192147-parker_beta-* | head -n1
output/images/factory/gluon-ffbs-next-v2019.1.1-ffbs-19-g03876975-202004192147-parker_beta-alfa-network-ap121.bin

$ ls output/images/debug                                                                                                    
ffbs-next-v2019.1.1-ffbs-19-g03876975-202004192147-parker_beta-ar71xx-generic-kernel-debug.tar.bz2
```

Before copying new files I remove all `-kernel-debug.tar.bz2`for the given `target`.
So I you do multiple makes for different architectures you end up with a `kernel-debug.tar.bz2` for every architecture.

Is that what you intended?

**Comment by SmithChart** [comment]
Rebased on top of master.

**Comment by belzebub40k** [comment]
Wouldn't it be more suitable to put this into $(GLUON_OUTPUTDIR)/debug instead of $(GLUON_IMAGEDIR)/debug as this is not a (flashable) image?

**Comment by belzebub40k** [comment]
decode_stacktrace.sh also calls the script decodecode. I think this should also be mentioned in the docs. Apart from that this seems to work as described. Yesterday I did a build with these patches to debug #1982 and the [decoded stackstrace](https://github.com/freifunk-gluon/gluon/issues/1982#issuecomment-617590482) looks fine.

**Comment by SmithChart** [comment]
> Wouldn't it be more suitable to put this into $(GLUON_OUTPUTDIR)/debug instead of $(GLUON_IMAGEDIR)/debug as this is not a (flashable) image?

You are totally right. I guess I used `GLUON_IMAGEDIR` because it has been used right next to my code.
I added `GLUON_OUTDIR` to the `env` and updated my PR.

**Comment by SmithChart** [comment]
> decode_stacktrace.sh also calls the script decodecode. I think this should also be mentioned in the docs. Apart from that this seems to work as described. Yesterday I did a build with these patches to debug #1982 and the [decoded stackstrace](https://github.com/freifunk-gluon/gluon/issues/1982#issuecomment-617590482) looks fine.

OK, I can add this.
I guess it's also a good idea to mention `/sys/kernel/debug/crashlog` on ar71xx.

**Comment by mweinelt** [comment]
Build stage fails:

```
cp: cannot create regular file '/home/jenkins/workspace/gluon-ci_PR-1971/output/debug/ci-0.6+exp20200422-x86-64-kernel-debug.tar.bz2': No such file or directory
openwrt/staging_dir/hostpkg/bin/lua: scripts/target_lib.lua:114: assertion failed!
stack traceback:
	[C]: in function 'assert'
	scripts/target_lib.lua:114: in function <scripts/target_lib.lua:112>
	(tail call): ?
	scripts/copy_output.lua:69: in main chunk
	[C]: ?
make: *** [Makefile:196: all] Error 1
```

File: scripts/copy_output.lua [hunk 8]
```diff
@@ -53,6 +54,20 @@ for _, image in ipairs(lib.images) do
 	end
 end
 
+-- copy kernel image with debug symbols
+local kernel_debug_glob = string.format('%s/debug/\0-%s-kernel-debug.tar.bz2',
+        env.GLUON_OUTPUTDIR,
+        target)
```
NeoRaider, line 60: The `mkdir` creates `${GLUON_IMAGEDIR}/debug`, but this attempts to write to `${GLUON_OUTPUTDIR}/debug`.

NeoRaider: Either use `${GLUON_IMAGEDIR}` here, or define a separate variable `${GLUON_DEBUGDIR}` in the main Makefile to make this configurable, similar to `${GLUON_PACKAGEDIR}`.

targets/generic
NeoRaider: For large kernels like x86, the bzip2 takes ~18s on my machine. An increase of 10% for the incremental build time for a single device :/

scripts/copy_output.lua
SmithChart: > Build stage fails:
> 
> ```
> cp: cannot create regular file '/home/jenkins/workspace/gluon-ci_PR-1971/output/debug/ci-0.6+exp20200422-x86-64-kernel-debug.tar.bz2': No such file or directory
> openwrt/staging_dir/hostpkg/bin/lua: scripts/target_lib.lua:114: assertion failed!
> stack traceback:
> 	[C]: in function 'assert'
> 	scripts/target_lib.lua:114: in function <scripts/target_lib.lua:112>
> 	(tail call): ?
> 	scripts/copy_output.lua:69: in main chunk
> 	[C]: ?
> make: *** [Makefile:196: all] Error 1
> ```
This should not have happend. I guess my test environment wasn't clean enough.

I now have defined `${GLUON_DEBUGDIR}` in the Makefile and have exported it into the environment. And I am now using it to create the directory and to copy files there...
One can now override it like `${GLUON_IMAGEDIR}`.

SmithChart:
> > decode_stacktrace.sh also calls the script decodecode. I think this should also be mentioned in the docs. Apart from that this seems to work as described. Yesterday I did a build with these patches to debug #1982 and the [decoded stackstrace](https://github.com/freifunk-gluon/gluon/issues/1982#issuecomment-617590482) looks fine.
> 
> OK, I can add this.
> I guess it's also a good idea to mention `/sys/kernel/debug/crashlog` on ar71xx.

Done

docs/dev/debugging.rst
```diff
@@ -0,0 +1,51 @@
...
+
+.. note::
+   Make sure to use a kernel tree that matches the version and patches
+   that was used to build the kernel.
+   If in doubt just re-build the images for the architecture.
+
+Some more information on how to use this tool can be found at
+`LWN <https://lwn.net/Articles/592724/>`__.
+
+Obtaining Stacktraces on ar71xx
```
belzebub40k, line 45: AFAIK crashlog is not dedicated to ar71xx but it is disabled on some architectures/targets.

https://github.com/openwrt/openwrt/blob/openwrt-19.07/config/Config-kernel.in#L31

NeoRaider: With all those exclusions, that leaves us with... mips(el), I guess? Although it seems this is additionally enabled `target/linux/generic/config-4.14`, but I have no idea if the disabled OpenWrt .config setting has precedence over this defconfig file.

targets/generic
NeoRaider: I don't have a good idea how to proceed.
- Some kind of escape hatch for fast incremental builds during development? Might lead to lengthy rebuilds on CONFIG_COLLECT_KERNEL_DEBUG changes though, not sure
- Changing the compression method (-> OpenWrt patch; maybe make this configurable for upstream)

Makefile
```diff
@@ -55,7 +56,8 @@ endef
 GLUON_VARS = \
 	GLUON_RELEASE GLUON_REGION GLUON_MULTIDOMAIN GLUON_DEBUG GLUON_DEPRECATED GLUON_DEVICES \
 	GLUON_TARGETSDIR GLUON_PATCHESDIR GLUON_TMPDIR GLUON_IMAGEDIR GLUON_PACKAGEDIR \
-	GLUON_SITEDIR GLUON_RELEASE GLUON_BRANCH GLUON_LANGS GLUON_BASE_FEEDS BOARD SUBTARGET
+	GLUON_SITEDIR GLUON_RELEASE GLUON_BRANCH GLUON_LANGS GLUON_BASE_FEEDS BOARD SUBTARGET \
+	GLUON_DEBUGDIR
```
NeoRaider, line 60: Let's move `GLUON_DEBUGDIR` next to `GLUON_PACKAGEDIR`.

SmithChart:
Rebased onto current master

targets/generic
NeoRaider: Okay, I just tested a few compressions tools (standard Debian Buster packages) with a x86-64 build:
- kernel-debug.tar: 376M
- kernel-debug.tar.gz: 101M, compressed in ~12s
- kernel-debug.tar.bz2: 91M, compressed in ~15s
- kernel-debug.tar.xz: 57M, compressed in ~101s
- kernel-debug.tar.zst: 86M, compressed in ~1s

I think we have a clear winner - I'll look into bringing zstd support into OpenWrt soon.

NeoRaider: https://patchwork.ozlabs.org/project/openwrt/list/?series=176744

Makefile
SmithChart: Done. It's now on the same line.

docs/dev/debugging.rst
SmithChart: I would suggest to leave it with ar71xx for now? The hint where it may be stored is still in the docs. And if anyone wants to add other targets/architectures they are free to do so anytime.

SmithChart:
As far as I can see there is one discussion open here. Does anyone have an opinion on that?

docs/dev/debugging.rst
NeoRaider: At least in the kernel .config, the crashlog is enabled for all targets except for x86 - the OpenWrt Kconfig symbols only limits on which archs it is possible to override this via OpenWrt's .config. This seems like a bug to me, but to fix it, we should first determine on which archs the crashlog actually works... Also note that in OpenWrt master, crashlog support hasn't been ported to Kernel 5.4 yet.

For this doc file, I would just remove the target from the title, and write something like "On many targets, ..." in the text.

rotanid:
> As far as I can see there is one discussion open here. Does anyone have an opinion on that?

i see two open ones, and one got an answer after your comment. please ask the related people via @ in the corresponding discussion-threads so we know what exactly is open.
if it's done, mark it as resolved.

NeoRaider:
I have pushed the zstd patches to the Gluon master now - please update

docs/dev/debugging.rst
SmithChart: Done. Documentation now reads:
```
Obtaining Stacktraces
.....................

On many targets stacktraces can be read from the following
location after reboot: ::

    /sys/kernel/debug/crashlog

```

SmithChart:
> I have pushed the zstd patches to the Gluon master now - please update

I am working on an update. With the changed build system I'll need to do some extra testing. I will let you know when I am ready.

mweinelt:
The netlify errors will go away as soon as you'll rebase against master. Sorry for the noise.

SmithChart:
> I have pushed the zstd patches to the Gluon master now - please update

@NeoRaider my branch is updated to use your zstd-compressed kernels.

mweinelt:
FWIW: The docs changes are visible here https://deploy-preview-1971--epic-jang-844eca.netlify.app/dev/debugging.html

scripts/copy_output.lua
```diff
@@ -66,6 +67,20 @@ for _, images in pairs(lib.images) do
...
 end
 
+-- copy kernel image with debug symbols
+local kernel_debug_glob = string.format('%s/\0-%s-kernel-debug.tar.zst',
+        env.GLUON_DEBUGDIR,
+        target)
+lib.exec {'rm', '-f', kernel_debug_glob}
+local kernel_debug_source = string.format('openwrt/bin/targets/%s/kernel-debug.tar.zst',
+        bindir)
+local kernel_debug_dest = string.format('%s/%s-%s-%s-kernel-debug.tar.zst',
```
NeoRaider, line 77: The glob and dest patterns should have `gluon-` before the site code.

docs/dev/debugging.rst
```diff
@@ -0,0 +1,51 @@
...
+that caused the oops.
+
+Linux kernels in Gluon (and OpenWrt) are stripped.
+That means they do not contain any debug symbols.
+On one hand this leads to a smaller binary and faster loading times on the
+target.
+On the other hand this means that in a case of a stack trace the unwinder
+can only print memory locations and no further debugging information.
+
+Gluon stores a compressed kernel with debug symbols for every architecture
```
NeoRaider, line 22: "architecture" -> "target"

scripts/copy_output.lua
SmithChart: Done.
with:
```
$ make all GLUON_TARGET=ar71xx-generic V=s -j10 GLUON_RELEASE=v2020.1-182-g93cdd052-20200603xxxx-parker_beta
```

I now get an output of 
```
$ ls output/debug
gluon-ffbs-next-v2020.1-182-g93cdd052-20200603xxxx-parker_beta-ar71xx-generic-kernel-debug.tar.zst
```

An of course it is cleand up before the next build of this target.

SmithChart:
@mweinelt 
> FWIW: The docs changes are visible here https://deploy-preview-1971--epic-jang-844eca.netlify.app/dev/debugging.html
Thanks for the link. I can say: It looks exactly like my local build :)

SmithChart:
> Thanks, this is working nicely. I have two more nitpicks, then we're ready to go.

@NeoRaider: I've done both requested changes. Please re-review.

PR #1972, opened 2020-04-05
Add gluon-logging package

### LINKED ISSUE #1845: site.conf configuration for gluon-web-logging
hello folks,

it would be very nice, if you could set default value for `system.@system[0].log_ip` in the site.conf.

cheers

mweinelt:
Allows configuration of remote syslog from site.conf.

Conflicts with `gluon-web-logging` as only one of them can control the remote syslog destination.

Resolves #1845. <s>Untested.</s>

### DIFF
--- a/package/gluon-logging/check_site.lua
+++ b/package/gluon-logging/check_site.lua
@@ -1,0 +1,3 @@
+need_string({'syslog', 'ip'}, true)
+need_number_range({'syslog', 'port'}, 1, 65535, false)
+need_one_of({'syslog', 'proto'}, {'tcp', 'udp'}, false)
--- a/package/gluon-logging/luasrc/lib/gluon/upgrade/040-syslog
+++ b/package/gluon-logging/luasrc/lib/gluon/upgrade/040-syslog
@@ -1,0 +1,19 @@
+#!/usr/bin/lua
+local site = require('gluon.site')
+local uci = require('simple-uci').cursor()
+
+local system = uci:get_first('system', 'system')
+
+if not site.syslog.ip() then
+	uci:delete('system', system, 'log_remote')
+	uci:delete('system', system, 'log_ip')
+	uci:delete('system', system, 'log_port')
+	uci:delete('system', system, 'log_proto')
+else
+	uci:set('system', system, 'log_remote', 1)
+	uci:set('system', system, 'log_ip', site.syslog.ip())
+	uci:set('system', system, 'log_port', site.syslog.port(514))
+	uci:set('system', system, 'log_proto', site.syslog.proto('udp'))
+end
+
+uci:save('system')
--- a/scripts/check_site.lua
+++ b/scripts/check_site.lua
@@ -289,6 +289,19 @@ function M.need_number(path, required)
 	return need_type(path, 'number', required, 'be a number')
 end
 
+function M.need_number_range(path, min, max, required)
+	local val = need_type(path, 'number', required)
+	if not val then
+		return nil
+	end
+
+	if val < min or val > max then
+		var_error(path, val, "be in range [" .. min .. ", " .. max .. "]")
+	end
+
+	return val
+end
+
 function M.need_boolean(path, required)
 	return need_type(path, 'boolean', required, 'be a boolean')
 end
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -32,6 +32,7 @@ files["package/**/check_site.lua"] = {
 		"need_chanlist",
 		"need_domain_name",
 		"need_number",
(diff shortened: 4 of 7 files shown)


mweinelt:
Tested and found to be working.

scripts/check_site.lua
```diff
@@ -289,6 +289,19 @@ function M.need_number(path, required)
 	return need_type(path, 'number', required, 'be a number')
 end
 
+function M.need_number_range(path, min, max, required)
```
blocktrron, line 292: `need_number_range` has to be added to .luacheckrc for check_site.lua files.

mweinelt: Ah, that makes sense. Thanks!

mweinelt:
@lrnzo Can you confirm that this fits your use case?

package/gluon-logging/luasrc/lib/gluon/upgrade/040-syslog
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/lua
+local site = require('gluon.site')
+local uci = require('simple-uci').cursor()
+
+local system = uci:get_first('system', 'system')
+
+if not site.syslog.ip() then
+	uci:delete_all('system', system, function(entry)
+		return entry['.name']:find('^log_')
```
blocktrron, line 9: IMHO we should explicitly delete only the 4 entries this package creates.

package/gluon-logging/check_site.lua
```diff
@@ -0,0 +1,3 @@
+need_string({'syslog', 'ip'}, false)
```
blocktrron, line 1: I haven't checked, but does this have to be an IP address or can it also be a hostname?

mweinelt: According to https://openwrt.org/docs/guide-user/base-system/log.essentials#network_logging it's a "<Destination IP>".

package/gluon-logging/luasrc/lib/gluon/upgrade/040-syslog
mweinelt: Whats the most elegant way to accomplish that?

mweinelt: Maybe like this:
```suggestion
		return entry['.name']:find('^log_(remote|ip|port|proto)')
```

NeoRaider: `|` does not exist in Lua patterns, and parentheses are only used fore captures, they don't affect the matching itself.

NeoRaider: `util.contains()` with a list of section names would be an option, but just using `uci:delete` seems like the simpler approach.

lrnzo:
> @lrnzo Can you confirm that this fits your use case?

yes, i built a firmware with the following in my site.conf. i had the gluon-feature `web-logging` disabeld to avoid any possible conflicts.

```	
syslog = {
		ip = "2a01:4f8:161:5205:2::1",
		port = 514,
		proto = "udp",
	},
```
and i get this on my testing router:

```
root@ffnw-10feed9c2618:~# uci show | grep log
system.@system[0].ttylogin='0'
system.@system[0].log_size='64'
system.@system[0].log_remote='1'
system.@system[0].log_ip='2a01:4f8:161:5205:2::1'
system.@system[0].log_port='514'
system.@system[0].log_proto='udp'
```
i will built another firmware with both `web-logging` and `logging` enabled to see how the configmode looks like. but so far i would say, yes, this fits my needs. thanks a lot :)

mweinelt:
> i will build another firmware with both `web-logging` and `logging` enabled to see how the configmode looks like.

You can't. I set the logging package to conflict with web-logging, as logging will set the site.conf given values on every upgrade and therefore overwrite the values set in web-logging.

lrnzo:
> You can't. I set the logging package to conflict with web-logging, as logging will set the site.conf given values on every upgrade and therefore overwrite the values set in web-logging.

maybe this is why i got the following error?

```
make[1]: *** [/home/lrnzo/no-backup/gluon/openwrt/include/toplevel.mk:227: world] Error 1
make[1]: Leaving directory '/home/lrnzo/no-backup/gluon/openwrt'
make: *** [Makefile:195: all] Fehler 2
```

mweinelt:
Rebuilding with `V=s` will show you the collision.

lrnzo:
i prefere saving the cpu-time :) i just checked the syslog on the syslog server. it works!

PR #1973, opened 2020-04-06
x86-geode: add target class

blocktrron:
x86-geode does not include the common x86 target-settings. Thus we need
to specify the device class in order to build images with all necessary
packages included.

### DIFF
--- a/targets/x86-geode
+++ b/targets/x86-geode
@@ -1,3 +1,5 @@
+class 'standard'
+
 packages {
 	'kmod-3c59x',
 	'kmod-8139cp',


mweinelt:
Why not instead move 

```
config 'CONFIG_VDI_IMAGES=y'
config 'CONFIG_VMDK_IMAGES=y'
```
into x86-64 and x86-generic and include x86.inc in the x86-geode target?

blocktrron:
I'm not sure about any attached consequences, do I've opted to "just make it work" instead of flaming a fire on another end.

mweinelt:
The geode factory image is ~23M in size, so it depends on whether all those drivers fit into that. I think that would be a pretty reasonable assumption.

Issue #1960, opened 2020-03-22
status-page: graph table is not responsive


tackin:
Node is on gluon 2020.1.1
As shown in the picture, the headline to the graph is outside of the layout on my mobile device.
My guess, it is a minor formaterror in the status-page html.

![photo_2020-03-22_10-18-35](https://user-images.githubusercontent.com/15726279/77246405-2811c900-6c27-11ea-8eab-2848ba5ce34d.jpg)

[http://[2001:bf7:fc0f:2:b24e:26ff:fe6c:e166]/cgi-bin/status](http://[2001:bf7:fc0f:2:b24e:26ff:fe6c:e166]/cgi-bin/status)
upps! Github with url encode-error? .. :-)

nalxnet:
The current data tables do not resize or restructure according to the responsive screen layout. The data tables need to react to such a small screen width. One approach could be to display each table row as it's own block/card, as with a such a small screen with there is barely enough space even for the table header.

PR #1975, opened 2020-04-07
gluon-status-page: set fixed layout and responsive view for data tables

nalxnet:
This commit sets a fixed layout for the data tables to prevent the table from overflowing the parent frame. Column widths are chosen to fit the current column labels, except for the first (hostname) column, which takes the leftover width.

To get the needed width to render a useful neighbor node table, the corresponding frame width is doubled.

For smaller frame widths (screen width < 1250px with 3-column layout, screen width <700px with 1-column layout), the table cannot be displayed in a usable way that does not overflow. This commit also changes the data table rendering for these screen widths to display table rows as "cards" (repeated table header, similar to tables in other columns).

You can see it in action here, might need to reload without cache to get the new stylesheet/js: http://[2001:678:6e3:1110:9683:c4ff:fe00:a7da]/cgi-bin/status

Tested with Chromium 80 on latest Arch Linux, Safari on iOS 13 and Chrome 80 / Firefox 68 / Safari 13 on macOS 10.14.

Fixes https://github.com/freifunk-gluon/gluon/issues/1960


blocktrron:
Thanks for tackling this issue :+1: 

Can you provide screenshots showing the result? It does not have to be on every platform you've listed.

nalxnet:
Sure, here are screenshots for desktop, tablet and mobile viewport. But since the new table layout is responsive, I recommend to open the link to the running device in your favourite browser and play around with the viewport width.

Desktop:
desktop

Tablet:
tablet

Mobile:
mobile

mweinelt:
Awesome!

PR #1977, opened 2020-04-08
docs: examples: remove haveged from GLUON_SITE_PACKAGES

jluebbe:
OpenWRT 19.07 enables urngd by default, so haveged is redundant.

### DIFF
--- a/docs/multidomain-site-example/site.mk
+++ b/docs/multidomain-site-example/site.mk
@@ -29,7 +29,7 @@ GLUON_MULTIDOMAIN=1
 #		chosen feature flags
 
 
-GLUON_SITE_PACKAGES := haveged iwinfo
+GLUON_SITE_PACKAGES := iwinfo
 
 ##	DEFAULT_GLUON_RELEASE
 #		version string to use for images
--- a/docs/site-example/site.mk
+++ b/docs/site-example/site.mk
@@ -23,7 +23,7 @@ GLUON_FEATURES := \
 #		selection that would be enabled by default or due to the
 #		chosen feature flags
 
-GLUON_SITE_PACKAGES := haveged iwinfo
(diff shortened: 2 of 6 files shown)


mweinelt:
There are a few more places where haveged is mentioned, not entirely sure removing is as easy as removing it here. Also we need to make urngd is started early enough in the setup mode to have a working key generation for fastd et al.

```
❯ rg haveged
docs/site-example/site.mk
26:GLUON_SITE_PACKAGES := haveged iwinfo

docs/multidomain-site-example/site.mk
32:GLUON_SITE_PACKAGES := haveged iwinfo

package/gluon-autoupdater/files/usr/lib/autoupdater/abort.d/90gluon-autoupdater
7:start_enabled haveged

package/gluon-autoupdater/files/usr/lib/autoupdater/download.d/10gluon-autoupdater
7:stop haveged

package/gluon-setup-mode/files/lib/gluon/setup-mode/rc.d/S13haveged
3:if [ -x /etc/init.d/haveged ] && /etc/init.d/haveged enabled; then
4:	. /etc/init.d/haveged
```

jluebbe:
I left the other places as they are, as my understanding is that they are needed to handle it correctly if a user decides they still want haveged.

mweinelt:
The autoupdater scripts probably need a check, whether the init script even exists. There might also be the need to add such scripts for urngd.

NeoRaider:
I think we should replace the references to haveged wuith urngd in the autoupdater code, I see no reason to keep supporting haveged.

jluebbe:
> I think we should replace the references to haveged with urngd in the autoupdater scripts, I see no reason to keep supporting haveged.

I've replaced haveged with urngd  in gluon-setup-mode and gluon-autoupdater as well.

PR #1978, opened 2020-04-09
github: actions: only build certain branches

blocktrron:
This enables build-testing only on master as well as next and release branches.

Pull requests done from source branches of this repository are still built.
Previously, those builds were redundant.


mweinelt:
Indeed, that halves the amount of checks. Nice!

PR #1980, opened 2020-04-11
Gluon reload.d add ebtables and radvd

2tata:
add missing reload scripts for ebtabels and uradvd.


mweinelt:
Maybe we also need to stop/start gluon-radv-filtered *before* gluon-ebtables:

```
root@64283-cccda-w:~# /etc/init.d/gluon-ebtables stop
Can't delete the chain 'RADV_FILTER', it's referenced in chain 'FORWARD', rule 21.
Sorry, rule does not exist.
root@64283-cccda-w:~# /etc/init.d/gluon-ebtables start
Chain RADV_FILTER already exists.
```

mweinelt:
Another unit that needs restarting is sysntpd. That is because the ntp servers can be overriden per domain.

```
 2471 root      1216 S<   /usr/sbin/ntpd -n -N -S /usr/sbin/ntpd-hotplug -p ntp.ffda.io
```

```
need_string_array({'ntp_servers'}, false)
```

mweinelt:
Also we should likely stick to the startup order we already have in /etc/rc.d.

package/gluon-ebtables/files/lib/gluon/reload.d/6000-gluon-ebtables-start
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
```
mweinelt, line 1: Probably meant 600, not 6000.

2tata:
> Maybe we also need to stop gluon-radv-filtered _before_ gluon-ebtables:
> 
> ```
> root@64283-cccda-w:~# /etc/init.d/gluon-ebtables stop
> Can't delete the chain 'RADV_FILTER', it's referenced in chain 'FORWARD', rule 21.
> Sorry, rule does not exist.

this appear now on start

> root@64283-cccda-w:~# /etc/init.d/gluon-ebtables start
> Chain RADV_FILTER already exists.
> ```

this appear now on stop

mweinelt:
I think this would be a good candidate for a CI test. Should be similar to https://github.com/freifunk-gluon/gluon/blob/master/tests/test_gluon-reconfigure.py, except you might need to wait until the system is booted up completely. What would be the canonical way to do this @lemoer?

NeoRaider:
At least the following initscripts need a restart (I might have overlooked something though):
- S19dnsmasq
- S19firewall
- S19gluon-ebtables
- S20network
- S50gluon-radvd
- S50gluon-radv-filterd
- S50gluon-respondd
- S90tunneldigger
- S95fastd

All of these should first be stopped in reverse order, and then started again. I think using the same numbering as the initscripts makes sense (maybe 3xx for stop scripts and 7xx for start scripts, and stop scripts need the indices subtracted from 100, so 19 -> 381).

Also, you add some empty files in this commit, this must be fixed. You can just do everything in a single commit, there is no value in splitting this one tiny commit for each initscript.

2tata:
tested, it works now ping on IPv4/v6 works after a few seconds when gluon-reload applied on a new domain.

NeoRaider:
- `network` has the wrong index (should stop on 80 and start on 20)
- `gluon-wan-dnsmasq` has the wrong index (should stop on 40 and start on 60); I assume `gluon-wan-dnsmasq` doesn't strictly require a restart, but it doesn't hurt either
- I overlooked the following two services, which should be restarted as well:
  - `S20gluon-arp-limiter`
  - `S99alfred`
- Let's name all start/stop scripts after the init scripts they call, so
  - `mesh-vpn-fastd` -> `fastd`
  - `respondd` -> `gluon-respondd`
  - etc.

2tata:
> * `network` has the wrong index (should stop on 80 and start on 20)
> * `gluon-wan-dnsmasq` has the wrong index (should stop on 40 and start on 60); I assume `gluon-wan-dnsmasq` doesn't strictly require a restart, but it doesn't hurt either
> * I overlooked the following two services, which should be restarted as well:
>   
>   * `S20gluon-arp-limiter`
>   * `S99alfred`
> * Let's name all start/stop scripts after the init scripts they call, so
>   
>   * `mesh-vpn-fastd` -> `fastd`
>   * `respondd` -> `gluon-respondd`
>   * etc.

done. seems like I ann the reload mechanism for gluon-wan-dnsmasq instead for dnsmasq. I have add dnsmasq now, shall I remove gluon-wan-dnsmasq?

NeoRaider:
Thanks. I have made a few final adjustments and merged this:
- Fixed switched start/stop order of *alfred*, *dnsmasq* and *gluon-arp-limiter*
- Moved *dnsmasq* scripts from `gluon-setup-mode` to `gluon-core`
- Made *alfred*, *dnsmasq* and *gluon-arp-limiter* script executable (please always test your changes!)

PR #1983, opened 2020-04-14
gluon-core: rewrite MAC-address generation logic

blocktrron:
This rewrites the MAC address generation logic to distinguish between
MAC addresses assigned to a wireless interface or virtual / wired
interfaces.

Each radio is now assigned a range of SSIDs, allowing for up to 8 SSIDs
per radio. Previosuly, there were 8 MAC addresses total for the entire
device.

To keep compatibility on already deployed nodes, MAC addresses for wired
/ virtual interfaces stay unchanged.

This change is required to support nodes with more than 2 radios.
It also allows us to increase the number of VAPs per radio.

Run-tested: WDR3600


NeoRaider:
I don't think there are any significant compatibility concerns - we can complete redefine all MAC addresses as long as we make sure our update scripts don't miss any. The problem I see is the probability for MAC address collisions:

The old logic is giving us a MAC address prefix with 43 bits of entropy (48 bit minus locally-administered, multicast, and 3 bits for counting) -> 8796093022208 prefixes in total. With the new logic, a single-band node uses two prefixes, and a dual-band node three (note that is doesn't make a difference whether the prefixes are selected independently, or consecutively like implemented by this PR). The following table shows the collision probabilities depending on the number of nodes in a domain and prefixes per node:

| Prefixes/node / Nodes/domain | 1000 | 2000 | 5000 | 10000
| --- | --- | --- | --- | --- |
| 1 | 5.679e-08 | 2.273e-07 | 1.421e-06 | 5.684e-06 |
| 2 | 2.273e-07 | 9.092e-07 | 5.684e-06 | 2.274e-05 |
| 3 | 5.114e-07 | 2.046e-06 | 1.279e-05 | 5.116e-05 |

I think we should try to keep the current scheme of 4 addresses / radio and overlapping radio and non-radio ranges going as long as possible, and just extend to a second prefix when we need 3 or 4 radios.

As a long-term goal, we might try to get rid of the fixed mapping completely and assign MAC addresses dynamically depending on the configuration:
- Only AP and mesh VIFs? Cram up to 4 radios into a single 8 address prefix
- Need more than 4 VIFs on a single radio? Each radio gets a separate prefix, as implemented here

NeoRaider:
> note that is doesn't make a difference whether the prefixes are selected independently, or consecutively like implemented by this PR

Well, this is not entirely correct either - it shouldn't make a different whether you select two independent prefixes, or two consecutive ones, as long as the first of the two can be on any index. If you only select from the even indices and additionally use the subsequent odd index, effectively removing one bit of entropy, collisions get slightly less likely, but you need to reserve twice as many addresses for each node even when a single 8-address prefix would suffice.

blocktrron:
Totally forgot about this.

To be fair, I'm aware of the reduced entropy. For supporting triple radio devices, your proposal would be sufficient in the scope of current Gluon.

However (and that's why i came up with this solution), the additional addresses per radio allow communities to use them for their custom packages.

NeoRaider:
> However (and that's why i came up with this solution), the additional addresses per radio allow communities to use them for their custom packages.

I think this is an even stronger argument for a more dynamic* assignment of MAC addresses, as I proposed as a future extension - otherwise we would need to have some kind of "MAC address index registry", so independently developed community packages don't run into address conflicts.

[*] "dynamic" meaning at `gluon-reconfigure` time I guess - I don't think we want to have real runtime assignments

blocktrron:
This needs to be reworked to not unnecessary reserve MAC addresses

Issue #1961, opened 2020-03-25
config-mode, autoupdater: Names of branches are ignored


citronalco:
According to the gluon docs it is possible to give the different firmware branches names, like in this site.conf excerpt:

```
autoupdater = {
    branch = 'stable',
    branches = {
        stable = {
            name = 'Maximum stability',
            [...]
        },
        experimental = {
            name = 'Bleeding edge',
            [...]
```
I would expect that the "name" would show up in Luci/Config Mode's Autoupdater module, so that  the users could choose between "Maximum stability" and "Bleeding edge". But the web interface shows only "stable" and "experimental" - what's in "name" seems to be competely irrelevant - or is it shown anywhere?

Additionally the order in which the branches are shown in the web interface seems to be pretty random and neither in the same order as configured in site.conf (what I would expect) nor alphabetically.

This affects latest Gluon (2020.1), 2019 and probably all olders, too.

blocktrron:
The list is not sorted, as the site generates uci entries which in return are used to generate the dropdown in the web view.

For using the name setting in the web-view you can try this patch:

```
From cd2fbfce9a0dd29516ac8c006b39c3d81f864632 Mon Sep 17 00:00:00 2001
From: David Bauer <mail@david-bauer.net>
Date: Wed, 25 Mar 2020 17:31:43 +0100
Subject: [PATCH] gluon-web-autoupdater: use name key for prettyname

---
 .../luasrc/lib/gluon/config-mode/model/admin/autoupdater.lua    | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/package/gluon-web-autoupdater/luasrc/lib/gluon/config-mode/model/admin/autoupdater.lua b/package/gluon-web-autoupdater/luasrc/lib/gluon/config-mode/model/admin/autoupdater.lua
index adf7cbd6..7cd1e13e 100644
--- a/package/gluon-web-autoupdater/luasrc/lib/gluon/config-mode/model/admin/autoupdater.lua
+++ b/package/gluon-web-autoupdater/luasrc/lib/gluon/config-mode/model/admin/autoupdater.lua
@@ -24,7 +24,7 @@ end
 o = s:option(ListValue, "branch", translate("Branch"))
 uci:foreach("autoupdater", "branch",
 	function (section)
-		o:value(section[".name"])
+		o:value(section[".name"], section["name"])
 	end
 )
 o.default = uci:get("autoupdater", autoupdater, "branch")
[...]

PR #1984, opened 2020-04-18
gluon-web-autoupdater: use human-readable names as branch labels, sort by label

NeoRaider:
Use the value of the `name` site.conf field as label (it was
accidentally unused before).

Our site.conf currently doesn't define a specific order for the branch
entries. To avoid changing branch orders, sort entries by this label.

Fixes: #1961


mweinelt:
@citronalco Can you test this pull request?

citronalco:
Tested, works.

Thank you!

PR #1989, opened 2020-04-24
targets/ramips-mt76x8: add support for Cudy WR1000

AiyionPrime:

  * [x]  flashing via web works
 
  * [x]  must have working sysupgrade <-- just tested once, looked fine.
        
* [x]  reset/wps/phone button must return device into config mode

* [x]  primary mac should match address on device label (or packaging) 

* wired network
      
* wifi (if applicable)
      
* led mapping
      
          _Power LED Does never blink, maybe hardwired_
        
  *   [x]  **should display config mode blink sequence** <-- Not entirely sure, as power led never blinks, but the WPS LED does it's job during config mode
  * radio leds
        
    * [x]  should show activity
  * switchport leds
        
    * [x]  should show link state and activity

---------------------
https://hannover.freifunk.net/karte/#/en/map/b44bd6224ebc

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -354,6 +354,10 @@ ramips-mt7621
 ramips-mt76x8
 -------------
 
+* Cudy
+
+  - WR1000 (v1)
+
 * GL.iNet
 
   - GL-MT300N v2
--- a/targets/ramips-mt76x8
+++ b/targets/ramips-mt76x8
@@ -1,3 +1,8 @@
+-- Cudy
+
+device('cudy-wr1000', 'cudy_wr1000')
+
+
 -- GL.iNet
 
 device('gl-mt300n-v2', 'gl-mt300n-v2', {


blocktrron:
What is the problem with sysupgrade on the device?

AiyionPrime:
There is likely none. I just have not tested, whether sysupgrade -n or firstboot work, yet.

AiyionPrime:
Tested, works like a charm.

nalxnet:
Seems like there is a v2 hardware out there in the wild that's indistinguishable from v1 on the outside and unsupported in OpenWRT: https://forum.openwrt.org/t/v2-of-cudy-wr1000/56965

This might get frustrating for someone buying a new device that ends up not being supported.

AiyionPrime:
I'm not sure, whether it's not supported.
The only source appears to be the author of the thread, that realizes he must not flash sysupgrade images on stock firmware [near the end of the thread](https://forum.openwrt.org/t/v2-of-cudy-wr1000/56965/33).

I can confirm that gluon works on devices with the description 

> EU/1.0

nalxnet:
> The only source appears to be the author of the thread, that realizes he must not flash sysupgrade images on stock firmware [near the end of the thread](https://forum.openwrt.org/t/v2-of-cudy-wr1000/56965/33).

There's also the author of the original v1 commit that claims v2 uses a different image header format.

AiyionPrime:
True.

AiyionPrime:
I'll add a hint in the docs, that only v1 is supported, that should suffice; as I just went through my waste paper and can confirm the revision is in fact readable on the outside of the packaging.

AiyionPrime:
@nalxnet are you fine with this?

```
diff --git a/docs/user/supported_devices.rst b/docs/user/supported_devices.rst
index 85104975..3485cf9f 100644
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -356,7 +356,7 @@ ramips-mt76x8
 
 * Cudy
 
-  - WR1000
+  - WR1000 v1
 
 * GL.iNet
```

AiyionPrime:
I think everything beside that is an upstream issue in openwrt, as
`lua -e 'print(require("platform_info").get_image_name())'` identifies the device as `cudy-wr1000` and not `cudy-wr1000-v1` or something like that.

nalxnet:
> @nalxnet are you fine with this?

Looks good. Maybe we should decide how to consistently identify the hardware revision in the docs, some write the version in parentheses, others do not. But that's another topic...

AiyionPrime:
Well then. I don't expect the remaining three builds to fail. And do not think my additions are related to the failing ar71xx-generic.

mweinelt:
> Maybe we should decide how to consistently identify the hardware revision in the docs, some write the version in parentheses, others do not.

Please put the revision in between two brackets. It allows for clearer visibility of the revision part of the model specification..

> And do not think my additions are related to the failing ar71xx-generic.

The CI build for ar71xx-generic has been failing alot, the builder is probably running out of available diskspace. We probably need to backport a63638904dadbf5e522c34e52af5e8d68882b7b0 to resolve that.

nalxnet:
> > Maybe we should decide how to consistently identify the hardware revision in the docs, some write the version in parentheses, others do not.
> 
> Please put the revision in between two brackets. It allows for clearer visibility of the revision part of the model specification..

I've created a PR to normalize the hardware revision references for the other devices: https://github.com/freifunk-gluon/gluon/pull/1991

docs/user/supported_devices.rst
```diff
@@ -354,6 +354,10 @@ ramips-mt7621
 ramips-mt76x8
 -------------
 
+* Cudy
+
+  - WR1000 v1
```
mweinelt, line 359: Revision in brackets please.

AiyionPrime: Sure, right? This would be the first device in the target with the bracket notation.

> ramips-mt76x8
> -------------
> 
> * Cudy
> 
>   - WR1000 v1
> 
> * GL.iNet
> 
>   - GL-MT300N v2
>   - VIXMINI
> 
> * NETGEAR
> 
>   - R6120
> 
> * TP-Link
> 
>   - Archer C50 v3
>   - Archer C50 v4
>   - TL-MR3020 v3
>   - TL-MR3420 v5
>   - TL-WA801ND v5
>   - TL-WR841N v13
>   - TL-WR902AC v3
> 
> * VoCore
> 
>   - VoCore2

Please consider writing that down as a general rule in the docs.

AiyionPrime: is done since 05432d8dde23da0a84a4d95a57e0c6b306605e4a

blocktrron:
Can you squash your commits into one and rename the resulting commit to `ramips-mt76x8: add support for Cudy WR1000`?

This is the last nitpick (i promise) :smile:

AiyionPrime:
Sure. Two minutes.

PR #1990, opened 2020-04-24
scripts: move output files instead of copy

mweinelt:
Moving images instead of copying them saves some disk space and there is probably no reason to keep the images in the OpenWrt buildroot.

Additionally the hope is that this will save enough disk space to have ar71xx-generic build through GitHub Actions.

### DIFF
--- a/scripts/copy_output.lua
+++ b/scripts/copy_output.lua
@@ -36,20 +36,22 @@ local function clean(image, name)
 	lib.exec {'rm', '-f', dir..'/'..file}
 end
 
-for _, image in ipairs(lib.images) do
-	clean(image, image.image)
+for _, images in pairs(lib.images) do
+	for _, image in ipairs(images) do
+		clean(image, image.image)
 
-	local destdir, destname = image:dest_name(image.image)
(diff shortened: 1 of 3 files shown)


scripts/copy_output.lua
```diff
@@ -64,5 +64,5 @@ if lib.opkg and (env.GLUON_DEVICES or '') == '' then
 	lib.exec {'rm', '-f', dest_dir('\0')..'/\0'}
 	lib.exec({'rmdir', '-p', dest_dir('\0')}, true, '2>/dev/null')
 	mkdir(dest_dir(package_prefix))
-	lib.exec {'cp', 'openwrt/bin/targets/'..bindir..'/packages/\0', dest_dir(package_prefix)}
+	lib.exec {'mv', 'openwrt/bin/targets/'..bindir..'/packages/\0', dest_dir(package_prefix)}
```
NeoRaider, line 67: Removing this directory from the OpenWrt dir should be avoided, it will significantly increase incremental build time with `GLUON_DEVICES` during development (see `clean_output.lua`).

mweinelt:
Neat, ar71xx-generic passed!

PR #1992, opened 2020-04-25
mac80211: ath10k: increase rx buffer size to 2048

ecsv:
Before, only frames with a maximum size of 1528 bytes could be transmitted between two 802.11s nodes.

For batman-adv for instance, which adds its own header to each frame, we typically need an MTU of at least 1532 bytes to be able to transmit without fragmentation.

This patch now increases the maxmimum frame size from 1528 to 1656 bytes.

Tested with two ath10k devices in 802.11s mode, as well as with batman-adv on top of 802.11s with forwarding disabled.

Fix originally found and developed by @greearb and is already part of ath10k-ct

Link: https://github.com/greearb/ath10k-ct/issues/89
Link: https://github.com/greearb/ath10k-ct/commit/9e5ab25027e0971fa24ccf93373324c08c4e992d
Forwarded: https://github.com/openwrt/openwrt/pull/2959


mweinelt:
Tested and LGTM. Allows ping with payloads > 1466 Bytes.

Issue #1982, opened 2020-04-14
Frequent Reboots under 2020.1 with ar71xx-*


margau:
In Frankfurt, there are reports of nodes often rebooting since 2020.1.
Some nodes are collected in https://md.margau.net/ffffm-nodereboot, and it seems that the problem is related to ar71xx-*. Our stats, e.g. https://freifunk.fail/d/000000002/nodes?orgId=1&from=now-90d&to=now&var-name=ORPLID-SEE, suggest that it not an memory leak.

### Bug report

Nodes are frequently rebooting without any visible cause.

**What is the expected behaviour?**
Nodes should not reboot. Was no problem before 2020.1

**Gluon Version:**
Tag 2020.1 and Tag 2020.1.1

**Site Configuration:**
https://github.com/freifunk-ffm/site-ffffm

**Custom patches:**
-

blocktrron:
Can you acquire a crashtrace from the routers console? Otherwise it's pure guesswork what is going on here.

mweinelt:
I can confirm more frequent reboots since v2020.1.

https://stats.darmstadt.freifunk.net/dashboard/snapshot/cb8TAPiqn16keAicL3eVp32z7hyFXlIB?orgId=1

rotanid:
how big are the layer2 segments (domains, hoods, ...) in which these nodes reside?
in darmstadt's case, they aren't big as far as i know...

jluebbe:
We (Braunschweig) are seeing this as well. It's happening on mostly idle nodes with a few batman neighbors (~3) as well. I've got only one Oops captured so far, but will try collect more:
```
[ 9037.708855] Data bus error, epc == 8012b1c0, ra == 8012b1b4                                                                                                       
[ 9037.714630] Oops[#1]:                                                                                                                                                                      
[ 9037.716987] CPU: 0 PID: 11688 Comm: logger Not tainted 4.14.171 #0                                                                                                                         
[ 9037.723371] task: 818f2d00 task.stack: 807b2000                                                                                                                                            
[ 9037.728042] $ 0   : 00000000 fffffff8 80a3d000 81042070                                                                                                                                    
[ 9037.733454] $ 4   : 00000001 014000c0 00000000 00000004                                                                                                                                    
[ 9037.738856] $ 8   : 00000014 80070a24 8011bb3c 00000048                                                                                                                                    
[ 9037.744259] $12   : 7f99a180 77f422a0 00000000 00000000                                                                                                                                    
[...]

PR #1994, opened 2020-04-25
kernel: backport slub prefetch fix from v4.19

NeoRaider:
This patch fixes a regression introduced in kernel v4.14. While the
commit message only mentions a performance penalty, the issue is
suspected to be the cause of spurious data bus errors on MIPS CPUs
(ar71xx target).

Fixes: #1982

My test device has completed more than 350M cycles of the test program posted in #1982 without issues. Needs more testing on a wider variety of hardware.

### DIFF
--- a/patches/openwrt/0010-kernel-backport-slub-prefetch-fix-from-v4.19.patch
+++ b/patches/openwrt/0010-kernel-backport-slub-prefetch-fix-from-v4.19.patch
@@ -1,0 +1,63 @@
+From: Matthias Schiffer <mschiffer@universe-factory.net>
+Date: Sat, 25 Apr 2020 19:27:34 +0200
+Subject: kernel: backport slub prefetch fix from v4.19
+
+This patch fixes a regression introduced in kernel v4.14. While the
+commit message only mentions a performance penalty, the issue is
+suspected to be the cause of spurious data bus errors on MIPS CPUs
+(ar71xx target).
+
+Signed-off-by: Matthias Schiffer <mschiffer@universe-factory.net>
+
+diff --git a/target/linux/generic/backport-4.14/190-v4.19-mm-slub-restore-the-original-intention-of-prefetch_f.patch b/target/linux/generic/backport-4.14/190-v4.19-mm-slub-restore-the-original-intention-of-prefetch_f.patch
+new file mode 100644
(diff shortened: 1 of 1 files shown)


jluebbe:
I've now run the test program on several devices:
TP-Link TL-WR841N/ND v9 726M cycles
TP-Link TL-WR940N v6 644M cycles
TP-Link TL-WDR4300 v1 358M cycles
TP-Link Archer C7 v2 482M cycles

Before, they only one of them even reached 37M. So I'd say that this fixes the problem.

belzebub40k:
Unifi AC Lite: more than 1,700,000,000 iterations and no crash
WR-842v3: more than 500,000,000 and no crash

I'll role it out to our experimental branch to have some real world results.

rubo77:
For info:

- The slub patch was added 2 years ago in https://gitlab.freedesktop.org/lyudess/linux/commit/2482ddec670f
- landed in Kernel 4.19 on 22. Okt. 2018
- 4.19 didn't land in any Openwrt release, 19.07.0 is still with Kernel 4.14

In Am nort sure, when kernel 4.19 was added in gluon, I find only this commit: https://github.com/openwrt/openwrt/commit/545bfbc3a922facfe52e9397da56605be33ff49a

blocktrron:
@rubo77 Gluon master follows OpenWrt 19.07 which follows kernel 4.14.

belzebub40k:
Stability of the 18 nodes running the new version looks much better. Only 3 restarts in the last 30h and 2 of them were 4/32 devices.

SvenRoederer:
just a side-note: the patch is changing kernels "mm/slub.c" it will also affect ath79-target, right? As it's the same platform, the problem might be seen there also.

mweinelt:
LGTM.

mweinelt:
> just a side-note: the patch is changing kernels "mm/slub.c" it will also affect ath79-target, right? As it's the same platform, the problem might be seen there also.

In principle yes, but we also didn't see this on every device in ar71xx, and since we have far fewer ath79 devices at this point the issue might exist, but may not be visible yet.

NeoRaider:
The patch was backported to kernel v4.14.178, so we get to drop it with the next kernel bump.

PR #1995, opened 2020-04-26
lantiq-xway: remove support for AVM FRITZ!Box 7320 7330 7330SL

### LINKED ISSUE #1943: Fritzbox 7330 ethernet connection only when connected to LAN1 before power up
To be able to connect to the Router (Config-Mode / normal mode) the ethernet cable has to be plugged in LAN1 upon booting. If it's not connected at the moment of powering up the device, no connection is possible (even the ports of a switch don't light up then). LAN2 doesn't work at all. Same behaviour in Gluon and OpenWRT.

Connection should be established when you plug in the ethernet cable, no matter when.

**Gluon Version:**

v2020.1

**Site Configuration:**

https://github.com/ffka/site-ffka/blob/master/site.conf

**LInks**

Same Bug in [OpenWRT](https://bugs.openwrt.org/index.php?do=details&task_id=2240&string=fritz&type%5B0%5D=&sev%5B0%5D=&pri%5B0%5D=&due%5B0%5D=&reported%5B0%5D=&cat%5B0%5D=&status%5B0%5D=open&percent%5B0%5D=&opened=&dev=&closed=&duedatefrom=&duedateto=&changedfrom=&changedto=&openedfrom=&openedto=&closedfrom=&closedto=)

blocktrron:
This device has broken Ethernet on both ports.

Remove support for those devices. for now, as there was no feedback from
the original author.

Closes #1943

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -271,9 +271,6 @@ lantiq-xway
 * AVM
 
   - FRITZ!Box 7312 [#avmflash]_
-  - FRITZ!Box 7320 [#avmflash]_ [#lan_as_wan]_
-  - FRITZ!Box 7330 [#avmflash]_ [#lan_as_wan]_
-  - FRITZ!Box 7330 SL [#avmflash]_ [#lan_as_wan]_
 
 * NETGEAR
 
--- a/targets/lantiq-xway
+++ b/targets/lantiq-xway
@@ -2,11 +2,6 @@ device('avm-fritz-box-7312', 'avm_fritz7312', {
 	factory = false,
 })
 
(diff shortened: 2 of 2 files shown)


mweinelt:
What's the discussion between removing and marking it as broken, why have you decided for remval?

Trivia: multi.meshviewer.org sees four 7320, no 7330/7330SL, so the fallout is minimal either way.

blocktrron:
I think there's no point in keeping the device as broken when the ethernet seems to work very unreliable. That's why I've opted for removing the device altogether.

mweinelt:
Does that affect the WAN port as well?

PR #1999, opened 2020-04-27
docs: add Aruba AP-303 and Instant On AP11 to supported devices list

nalxnet:
I notices these two are missing in the docs.


mweinelt:
Thanks!

PR #2000, opened 2020-04-29
tests: add batctl tpmeter test case

mweinelt:
In https://github.com/freifunk-gluon/gluon/commit/a070e688498b7ab79cba5c9e8126d2a787a92f68 batman-adv was updated to emit correct exit codes when running tpmeter, which should make this test case work.

cc @lemoer


lemoer:
thanks! :)

PR #2004, opened 2020-05-02
Jenkinsfile: conditionally build additional targets

mweinelt:
Let's build more targets through our Jenkins CI and decide more intelligently what to build.


mweinelt:
While a good idea in general I don't think I'll continue here.

PR #2007, opened 2020-05-03
Fix package selection precedence

### LINKED ISSUE #2005: negative package selection is broken
### Bug report

**What is the problem?**
- What is not working as expected?

In hanover we made use of the negative package selection [like this](https://github.com/freifunkh/site/blob/5052b05dafccba04fd6d26c510cd0e23f241e151/site.mk#L33)
`-gluon-web-autoupdater`
this does not work anymore, as we intended to.

- How is it misbehaving?

Building a recent gluon (latest) with our site results in the build-process trying to add the marked package, though its functionality (and files) are provided by a different one.
This results in breaking the build, as the step `check_data_file_clashes` does correctly break, as both packages provide nearly the same files.
The point is, the negative marked one should not only not provide no files, but not be part of the build at all.

- When did the problem first start showing up?

Our nightly build system was broken for a while, due to broken patches on our end, so we were not immediately aware of it.
But after bisecting gluon 1df243d1b3257177888c9cddae516abf737c32e5 (merged in #1876) appears to be the first bad commit.
 
- What were you doing when you first noticed the problem?

Repairing our buildsystem in order to provide new device-support.

- On which devices (vendor, model and revision) is it misbehaving?

I bisected this only for the target `ramips-mt76x8` but the issue should show up on every target we got.

- Does the issue appear on multiple devices or targets?

Yes, I suppose it does on every single one of them.

- How do you think it should work instead?

The negative marked package should not be automatically added to the build.
In the best scenario, the automatic process would ensure, that something _provides_ the functionality, but not adding a package on it's own.
The latest gluon  should be able to build with the latest site of hanover ( freifunkh/site@5052b05dafccba04fd6d26c510cd0e23f241e151 )

- Did it work like that before?

Yes, before the merge of #1876 .

**Gluon Version:**
the current latest gluon commit has the issue:
daf8a6c18c2bc23efa075daf43c255915fb294b3

the bug exists since this commit:
1df243d1b3257177888c9cddae516abf737c32e5

**Site Configuration:**

freifunkh/site@5052b05dafccba04fd6d26c510cd0e23f241e151

**Additional info**
the script evaluated by git bisect:
```bash
#!/bin/bash
make update
target="ramips-mt76x8"; make V=s GLUON_TARGET=$target &> logs/$(date -Is)_$target.log

latest=$(ls -t logs | head -n1)
echo reading $latest
! cat "logs/$latest" | grep "check_data_file_clashes"
```

the replacement package in ff-Hochstifts package repo:
https://git.ffho.net/FreifunkHochstift/ffho-packages/src/master/ffho-web-autoupdater

Warning: there is an **unrelated** warning about the ffh-district-migrate package.
It's save to ignore it, for this issue.

**Custom patches:**

Only the ones specified in hanovers patch directory.
They are most likely unrelated to this problem.

https://github.com/freifunkh/site/blob/master/patches/0001-fix-when-vlan-crc-is-00000000.patch
https://github.com/freifunkh/site/blob/master/patches/0002-patches-uradvd-change-ip-lifetimes-to-150-300-second.patch
https://github.com/freifunkh/site/blob/master/patches/0003-added-TP-Link-TL-WR841ND-N-Devices-for-8M-and-16M-Va.patch

NeoRaider:
The precedence of different package lists was broken since #1876,
disallowing removal of GLUON_FEATURES packages via GLUON_SITE_PACKAGES.

Including all package selections, both implicit defaults and explicit
handling in Gluon, the order of precedence is now the following:

1.  OpenWrt defaults (including target-specific defaults)
2.  Device-specific packages from OpenWrt
3.  Generic default packages (from target/generic)
4.  Target default packages (target/$(GLUON_TARGET))
5.  Removal of opkg for tiny targets
6.  Packages derived from GLUON_FEATURES + GLUON_FEATURES_$(class)
7.  GLUON_SITE_PACKAGES
8.  GLUON_SITE_PACKAGES_$(class)
9.  Device-specific packages from target/$(GLUON_TARGET)
10. Device-specific packages from GLUON_$(device)_SITE_PACKAGES

This also contains various pieces of cleanup:

- No hardcoded order of device classes for target_config.lua arguments
  anymore (in fact, the Makefile doesn't know anything about device
  classes now)
- target_conifg_lib.lua only hardcodes the fallback class for x86, no
  other occurences of specific class names
- Feature -> package list mapping is moved from Makefile to the Lua code
  as well (still implemented in Shell though)

Fixes #2005

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,8 @@ endef
 GLUON_VARS = \
 	GLUON_RELEASE GLUON_REGION GLUON_MULTIDOMAIN GLUON_DEBUG GLUON_DEPRECATED GLUON_DEVICES \
 	GLUON_TARGETSDIR GLUON_PATCHESDIR GLUON_TMPDIR GLUON_IMAGEDIR GLUON_PACKAGEDIR \
-	GLUON_SITEDIR GLUON_RELEASE GLUON_BRANCH GLUON_LANGS GLUON_BASE_FEEDS BOARD SUBTARGET
+	GLUON_SITEDIR GLUON_RELEASE GLUON_BRANCH GLUON_LANGS GLUON_BASE_FEEDS \
+	GLUON_TARGET BOARD SUBTARGET
 
 unexport $(GLUON_VARS)
 GLUON_ENV = $(foreach var,$(GLUON_VARS),$(var)=$(call escape,$($(var))))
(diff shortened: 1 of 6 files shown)


AiyionPrime:
From #gluon:

> 23:56:36          aiyion | hexa-, neoraider: my test is done, with #2007 fe88bb5e i was able to build through without any changes to our site.

AiyionPrime:
And the new last commit:

> 08:31:23          aiyion | hexa-, neoraider: dirclean build of 53690d83 succeeded at 03:11.

AiyionPrime:
And the resulting image is flashable:
https://hannover.freifunk.net/karte/#/en/map/b44bd6224ebc

Anything left to test on my end?

PR #2008, opened 2020-05-03
add hardware: Linksys EA6350 v3

ctr49:
successfully built, flashed and ran a device based on this

### DIFF
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac
@@ -105,6 +105,9 @@ local primary_addrs = {
 			'archer-c60-v1',
 			'archer-c7',
 		}},
+		{'ipq40xx', 'generic', {
+			'linksys,ea6350v3',
+		}},
 		{'ipq806x', 'generic', {
 			'netgear,r7800',
 		}},
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac
@@ -105,9 +105,6 @@ local primary_addrs = {
 			'archer-c60-v1',
 			'archer-c7',
 		}},
-		{'ipq40xx', 'generic', {
-			'linksys,ea6350v3',
-		}},
 		{'ipq806x', 'generic', {
 			'netgear,r7800',
 		}},
@@ -128,6 +125,11 @@ local primary_addrs = {
 	{board('lan'), {
 		{'lantiq', 'xway', {
 			'netgear,dgn3500b',
+		}},
+	}},
+	{board('wan'), {
+		{'ipq40xx', 'generic', {
+			'linksys,ea6350v3',
 		}},
 	}},
 	-- phy0 default
--- a/targets/ipq40xx-generic
+++ b/targets/ipq40xx-generic
@@ -61,6 +61,11 @@ device('gl.inet-gl-b1300', 'glinet_gl-b1300', {
 })
 
 
+-- Linksys
+
+device('linksys-ea6350v3', 'linksys_ea6350v3')
+
+
 -- NETGEAR
 
 device('netgear-ex6100v2', 'netgear_ex6100v2', {


mweinelt:
Thank you for yout contribution.

For adding device support we require the [device checklist](https://github.com/freifunk-gluon/gluon/wiki/Device-Integration-checklist) to be copied and filled out. This is to ensure that the core functionality of Gluon works on this device.

ctr49:
  - [x] webinterface
  - [ ] tftp _(not tested yet)_
    - [ ] must keep/forget configuration (if applicable) _(not tested yet)_
- [ ] reset/wps/phone button must return device into config mode _(not tested yet)_
- [p] primary mac should match address on device label label shows: eth1 mac matches label, primary_mac is off by 0x10
- wired network
- wifi (if applicable)
- led mapping
  - radio leds 
    - [ ] should map to their respective radio (no radio LEDs)
    - [ ] should show activity (no radio LEDs)
  - switchport leds
- outdoor devices only

mweinelt:
> [p] primary mac should match address on device label label shows: eth1 mac matches label, primary_mac is off by 0x10

This means you need to modify `package/gluon-core/luasrc/lib/gluon/upgrade.d/010-primary-mac` to have it read the correct label mac from the correct interface/phy.

ctr49:
I would have expected that the update would actually pick the eth1 mac as primary, but it picks eth0 mac now. So instead of being 0x10 off, it's 1-off. Still not tas intended. Any suggestions?

mweinelt:
Did you `sysupgrade -n <image.bin>`, as in forget all settings? If not try `firstboot` and `reboot`.

ctr49:
Yes and I also tried on a second device (using the factory image coming from stock firmware), same behavior: eth0 mac being used instead of the eth1 mac.

NeoRaider:
It seems the `eth1` MAC address is set "too late" for the upgrade script to notice it on this device. Try adding the device match to the `board('lan')` block instead.

blocktrron:
I might be ruining the day here, but the device has a fixed 3M kernel limit which we already hit 2.7M of (OpenWrt Master - Kernel 5.4). Oh and the bootloader does not support LZMA.

So this raises the question for me if this device should be flagged as BROKEN from the beginning for this reason.

ctr49:
board('lan') didn't help either.
What's the OpenWRT equivalent of primary_mac? I'd like to verify if plain OpenWRT does it "right"...

ctr49:
> the device has a fixed 3M kernel limit which we already hit 2.7M of (OpenWrt Master - Kernel 5.4). Oh and the bootloader does not support LZMA.
> 
> So this raises the question for me if this device should be flagged as BROKEN from the beginning for this reason.

There are other devices with a 3M kernel limit (ubnt-erx) and even some with a smaller ones down to 2M (Zyxel NBG6616) that are not marked as broken. Is not supporting native lzma (vs zImage) such a huge overhead to justify a 3M limit as cause for flagging this BROKEN?

As far as I can see this would be the very first device that would be marked broken due to kernel size limit (just grepping through "broken" in targets real quick, correct me if I'm wrong).

blocktrron:
The ER-X might be circumventable as we can probably combine both kernel partitions to allow up to 6M kernel. The NBG6616 however is in the ar71xx target, which will never see a kernel newer than 4.14. So this is a non-issue.

For me the question is whether or not we want to add a device we know has to be dropped 2-3 releases into the future. This is something we should decide on before merging this as non-broken.

NeoRaider:
> board('lan') didn't help either.
> What's the OpenWRT equivalent of primary_mac? I'd like to verify if plain OpenWRT does it "right"...

OpenWrt doesn't have such a concept so far (there is the idea of a "label-mac" in the device tree, not sure if that is merged yet), but this will only work when the MAC address can be determined by the kernel via the device tree - which is not the case on this device. Please post the contents of `/etc/board.json` - that *should* contain the right MAC address somewhere.

Missing LZMA support in the bootloader doesn't seem like a problem, zImage should work fine (but of course it may still be an issue if even the LZMA-compressed kernel is close to 3M). lzma-loader-like solutions (or just a second/third-stage U-Boot) would probably work, but I'm not sure if we want to go that route.

NeoRaider:
@ctr49 Ah, is the MAC address 1 too high or too low? I think `board('wan')` might be the correct handler after all.

rotanid:
> For me the question is whether or not we want to add a device we know has to be dropped 2-3 releases into the future. This is something we should decide on before merging this as non-broken.

we do have devices which will have problems earlier than this one which aren't flagged as broken.
in my opinion BROKEN is wrong for this device, if the kernel size limit is the only point here.

maybe we need more/different steps for deprecating devices, depending on the time we estimate until a device can not be supported anymore?

the one thing which stands against supporting EA6350 without any flag - for me - is that it is still available in many online-shops, so we could encourage people to spend 90€ on a device which will only last maybe 2 years. my opinion to support it without any flag would be much stronger if it wouldn't be available brand new...

ctr49:
> Ah, is the MAC address 1 too high or too low? I think `board('wan')` might be the correct handler after all.

Yes, that did the trick! primary_mac is now the one from the label.

> the one thing which stands against supporting EA6350 without any flag - for me - is that it is still available in many online-shops, so we could encourage people to spend 90€ on a device which will only last maybe 2 years. my opinion to support it without any flag would be much stronger if it wouldn't be available brand new...

The device is neither new nor expensive. You can get it for a bargain from time to time. The three devices I've got here were £ 27.50 each from Amazon UK ... in July 2019. Not too bad for a device with those specs.

NeoRaider:
To reduce confusion about different ways to specify the interface to use the address from, and different times when these become valid in the boot process, I'm working on a patch to migrate all eth0/eth1 matches to `board('lan')` and `board('wan')`.

ctr49:
Is there anything else I can contribute to get this merged?

NeoRaider:
From my side, this looks good, while @blocktrron has raised a concern. I don't think we are encouraging people to buy a device just by supporting it - lots of devices we support are not advisable at all, for various reasons.

blocktrron:
Fine with me, i just wanted to raise awareness for this issue.

Issue #2009, opened 2020-05-05
Kernel panic in 2020.1.2 when using pre-Pentium 4


ChristophFranzen:
Hallo zusammen, ein [_Gluon 2020.1.2_ (x86; Freifunk Trier)](https://github.com/freifunktrier/firmware_store/blob/master/firmware/stable/factory/gluon-fftr-0.13.4%2Btackin-x86-generic.img.gz) fällt hier auf die Nase. Es haut sich weg in dem Augenblick, wo der Kernel init starten will:
```
traps: init[1] trap invalid opcode ip:b7f39508 sp:bfdafb30 error:0 in libc.so[b7f05000+4b000]
Kernel panic - not syncing: Attempted to kill init! exitcode=0x00000004
```
Der Prozessor ist ein P-III, ältere Linux-Kernel (_Jessie_, ein altes _Gluon v2016.1.6-3-g871460b_, M0n0wall/_pfSense_ [BSD]) gehen auf dem Gerät. Das war neu aufgesetzt, weil Sysupgrade von dem alten Gluon nicht ging.
### Hardware:
[NORCO_MB-X67.pdf](https://github.com/freifunk-gluon/gluon/files/4580999/2015.07.16.22.52.01.NORCO_MB-X67.pdf) Es handelt sich um eine _Watchguard Firebox X Core_ [deren Modelle X500…X2500 sich NUR in der Original-Software unterscheiden]. Darin steckt abweichend von der PDF-Datei ein _WG-X66A_, einziger Unterschied: kein Gigabit-Ethernet, alle 6 Schittstellen sind _RTL8139C+_; LAN7 ist nicht bestückt.
> Intel Pentium III/Celeron Processor in Socket 370 (133 MHz), support B-stepping
> Intel 815EP chipset, IT8712F I/O Chipset
> ~~Two Intel 82540 Ethernet controller Four Intel 82551, One on-chip MAC + Intel 82562~~
> One 168pin DIMM Socket, support PC-100/133 MHz SDRAM
> One 40-pin (2x20, 2mm) IDE connector that support Ultra ATA100, build in secondary master IDE Channel
> One 44-pin (2x22, 2mm) IDE connector that support Ultra ATA100, build in secondary slave IDE Channel
> One CompactFlash type II socket, support ATA mode, build in primary master IDE Channel
> AMI licensed BIOS Seven RJ-45 external connectors One RS-232 DB-9 connector
### Ganzes Log
[...]

PR #2010, opened 2020-05-05
RFC: add x86-legacy subtarget

txt-file:
As x86-generic is compiled to pentium4 (and newer) there is a need for
a subtarget for older devices. The x86-legacy subtarget is set to
compile to pentium (and newer) and should support even very old devics.

Comments welcome.

Closes: #2009

### DIFF
--- a/targets/x86-legacy
+++ b/targets/x86-legacy
@@ -1,0 +1,10 @@
+include 'x86.inc'
+
+packages {
+        '-kmod-gpio-nct5104d',
+        '-kmod-leds-gpio',
+        '-kmod-leds-apu2',
+}
+
+factory_image('x86-legacy', 'combined-squashfs', '.img.gz')
+sysupgrade_image('x86-legacy', 'combined-squashfs', '.img.gz')
--- a/.github/workflows/build-gluon.yml
+++ b/.github/workflows/build-gluon.yml
@@ -412,6 +412,27 @@ jobs:
           name: x86-geode_output
           path: output
 
+  x86-legacy:
+    name: x86-legacy
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Install Dependencies
+        run: sudo contrib/actions/install-dependencies.sh
+      - name: Build
+        run: contrib/actions/run-build.sh x86-legacy
+      - name: Archive build logs
+        if: ${{ !cancelled() }}
+        uses: actions/upload-artifact@v1
+        with:
+          name: x86-legacy_logs
+          path: openwrt/logs
+      - name: Archive build output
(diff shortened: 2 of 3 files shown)


txt-file:
compiled successfully.

targets/x86-legacy
```diff
@@ -0,0 +1,10 @@
+include 'x86.inc'
+
+packages {
+        '-kmod-gpio-nct5104d',
+        '-kmod-leds-gpio',
+        '-kmod-leds-apu2',
+}
```
NeoRaider, line 7: Should these package removals also be included upstream? Or are they not relevant anymore because of the move to the new image build code in OpenWrt master?

txt-file: Without these excludes I get:
```
Configuration failed:
 * unable to enable package 'kmod-gpio-nct5104d'
 * unable to enable package 'kmod-leds-gpio'
 * unable to enable package 'kmod-leds-apu2'
make: *** [Makefile:177: config] Fehler 1
```
As these packages are available for other subtargets it would be bad to exclude them for all other x86-subtargets.

NeoRaider: Ah, these are Gluon additions from `x86.inc`. I think we should just not include `x86.inc`, as we already do for `x86-geode`.

rotanid: @NeoRaider have you intentionally marked your remark as resolved?

NeoRaider: Yes, this was discussed on IRC, and I'm okay with the current solution.

txt-file: @rotanid IMHO opinion we do not need to look carefully after the image size for the x86 targets. Therefore I see no problem to have much more packages than actually useful on x86 targets. On the other hand I see a problem to have multiple different package lists like currently with x86-geode not using x86.inc. If there is a need for some additional package it currently needs to be added to x86.inc and x86-geode.

mweinelt:
Please update the GitHub Actions workflow using

```
make list-targets BROKEN=1 | ./contrib/actions/generate-actions.py > ./.github/workflows/build-gluon.yml
```

and squash that into the commit.

txt-file:
seems to be recognized as there is a x86-legacy check

blocktrron:
The GHA update requested by @mweinelt does not seem to be contained  in your latest push

txt-file:
I runtime tested this yesterday on a WatchGuard FireBox X1250e. Flashed gluon to a CF-card, it booted. Tested sysupgrade from USB flash, worked also.
I found one problem: The bootloader gets overwritten by sysupgrade. If any special boot parameter is needed something needs to be invented to keep it. But this is an upstream issue.

PR #2014, opened 2020-05-08
gluon-web-model: cleanup backend error handling

NeoRaider:
So I'm not entirely sure what to do with this. This removes code that has been dead for a very long time... but maybe we do need a bit more error feedback after all?


rotanid:
i haven't found any typo at least, but haven't checked or tested the code's functionality

PR #2021, opened 2020-05-10
gluon-mesh-vpn-fastd: clean up peers and groups on update

NeoRaider:
The 'preserve' flag can be used to mark a peer so it is not removed or
modified on upgrades. In addition, groups containing preserved peers are
not removed.
    
Fixes: #557

### DIFF
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -302,6 +302,12 @@ mesh_vpn
 
     You can set syslog_level from verbose (default) to warn to reduce syslog output.
 
+    fastd allows to configure a tree of peer groups and peers. By default, the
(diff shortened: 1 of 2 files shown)


CodeFetch:
@NeoRaider As you've changed the default behavior (keeping old peer groups if none with the same name overwrites it) this should be mentioned in the release notes, I guess... Thank you for taking care of it. That's nearly what I was planning to do and I'm happy with it.

NeoRaider:
> @NeoRaider As you've changed the default behavior (keeping old peer groups if none with the same name overwrites it) this should be mentioned in the release notes, I guess...

That's why this PR has the "meta: release notes" label :)

PR #2023, opened 2020-05-10
scripts: Add push_pkg.sh to enhance package development workflow (v2)

lemoer:
v2: In contrast to the last patches, this is now built on top of ssh
    only, without using e.g. 9pfs. Furthermore it works also with
    arbitary remote hosts on any target/architecture. Also the
    scripts were renamed and moved to /scripts.

The aim of this commit is to allow fast rebuild cycles during the
development of gluon packages.

Currently the following workflow can be used:

	# start a local qemu instance
	scripts/run_qemu.sh output/images/factory/[...].img

	# do your changes in the file you want to patch
	vi package/gluon-ebtables/files/etc/init.d/gluon-ebtables

	# rebuild and update the package
	scripts/push_pkg.sh package/gluon-ebtables/

	# test your changes
	...

	# do more changes
	...

	# rebuild and update the package
	scripts/push_pkg.sh package/gluon-ebtables/

	# test your changes
	...

	(and so on...)

Implementation details:
- Currently this is based on ssh/scp.
- Opkg is used to install/update the packages in the remote machine.

Benefits:
- This works with compiled and non-compiled packages.
- This works with native OpenWrt and Gluon packages.
- This even performs the check_site.lua checks as they are integrated
  as post_install scripts into the openwrt package.
- It works for all architectures/targets.

EDIT:
- 2020-12-20: I updated this comment to the description of the new patch.

### DIFF
--- a/contrib/push_pkg.sh
+++ b/contrib/push_pkg.sh
@@ -1,0 +1,149 @@
+#!/bin/sh
+
+set -e
+
+topdir="$(realpath "$(dirname "${0}")/../openwrt")"
+
+# defaults to qemu run script
+ssh_host=localhost
+build_only=0
+preserve_config=1
+
+print_help() {
+	echo "$0 [OPTIONS] PACAKGE_DIR [PACKAGE_DIR] ..."
+	echo ""
+	echo " -h          print this help"
+	echo " -r HOST     use a remote machine as target machine. By default if this"
+	echo "             option is not given, push_pkg.sh will use a locally"
+	echo "             running qemu instance started by run_qemu.sh."
+	echo " -p PORT     use PORT as ssh port (default is 22)"
+	echo " -b          build only, do not push"
+	echo " -P          do not preserve /etc/config. By default, if a package"
+	echo "             defines a config file in /etc/config, this config file"
+	echo "             will be preserved. If you specify this flag, the package"
+	echo "             default will be installed instead."
+	echo ""
+	echo ' To change gluon variables, run e.g. "make config GLUON_MINIFY=0"'
+	echo ' because then the gluon logic will be triggered, and openwrt/.config'
+	echo ' will be regenerated. The variables from openwrt/.config are already'
+	echo ' automatically used for this script.'
+	echo
+}
+
+while getopts "p:r:hbP" opt
+do
+	case $opt in
+		P) preserve_config=0;;
+		p) ssh_port="${OPTARG}";;
+		r) ssh_host="${OPTARG}"; [ -z "$ssh_port" ] && ssh_port=22;;
+		b) build_only=1;;
+		h) print_help; exit 0;;
+		*) ;;
+	esac
+done
+shift $(( OPTIND - 1 ))
+
+[ -z "$ssh_port" ] && ssh_port=2223
+
+if [ "$build_only" -eq 0 ]; then
+	remote_info=$(ssh -p "${ssh_port}" "root@${ssh_host}" '
+		source /etc/os-release
+		printf "%s\t%s\n" "$OPENWRT_BOARD" "$OPENWRT_ARCH"
+	')
+	REMOTE_OPENWRT_BOARD="$(echo "$remote_info" | cut -f 1)"
+	REMOTE_OPENWRT_ARCH="$(echo "$remote_info" | cut -f 2)"
+
+	# check target
+	if ! grep -q "CONFIG_TARGET_ARCH_PACKAGES=\"${REMOTE_OPENWRT_ARCH}\"" "${topdir}/.config"; then
+		echo "Configured OpenWrt Target is not matching with the target machine!" 1>&2
+		echo
+		printf "%s" "    Configured architecture: " 1>&2
+		grep "CONFIG_TARGET_ARCH_PACKAGES" "${topdir}/.config" 1>&2
+		echo "Target machine architecture: ${REMOTE_OPENWRT_ARCH}" 1>&2
+		echo 1>&2
+		echo "To switch the local with the run with the corresponding GLUON_TARGET:"  1>&2
+		echo "  make GLUON_TARGET=... config" 1>&2
+		exit 1
+	fi
+fi
+
+if [ $# -lt 1 ]; then
+	echo ERROR: Please specify a PACKAGE_DIR. For example:
+	echo
+	echo " \$ $0 package/gluon-core"
+	exit 1
+fi
+
+while [ $# -gt 0 ]; do
+
+	pkgdir="$1"; shift
+	echo "Package: ${pkgdir}"
+
+	if ! [ -f "${pkgdir}/Makefile" ]; then
+		echo "ERROR: ${pkgdir} does not contain a Makefile"
+		exit 1
+	fi
+
+	if ! grep -q BuildPackage "${pkgdir}/Makefile"; then
+		echo "ERROR: ${pkgdir}/Makefile does not contain a BuildPackage command"
+		exit 1
+	fi
+
+	opkg_packages="$(make TOPDIR="${topdir}" -C "${pkgdir}" DUMP=1 | awk '/^Package: / { print $2 }')"
+
+	search_package() {
+		find "$2" -name "$1_*.ipk" -printf "%f\n"
+	}
+
+	make TOPDIR="${topdir}" -C "${pkgdir}" clean
+	make TOPDIR="${topdir}" -C "${pkgdir}" compile
+
+	if [ "$build_only" -eq 1 ]; then
+		continue
+	fi
+
+	# IPv6 addresses need brackets around the ${ssh_host} for scp!
+	if echo "${ssh_host}" | grep -q :; then
+		BL=[
+		BR=]
+	fi
+
+	for pkg in ${opkg_packages}; do
+
+		for feed in "${topdir}/bin/packages/${REMOTE_OPENWRT_ARCH}/"*/ "${topdir}/bin/targets/${REMOTE_OPENWRT_BOARD}/packages/"; do
+			printf "%s" "searching ${pkg} in ${feed}: "
+			filename=$(search_package "${pkg}" "${feed}")
+			if [ -n "${filename}" ]; then
+				echo found!
+				break
+			else
+				echo not found
+			fi
+		done
+
+		if [ "$preserve_config" -eq 0 ]; then
+			opkg_flags=" --force-maintainer"
+		fi
+
+		# shellcheck disable=SC2029
+		if [ -n "$filename" ]; then
+			scp -P "${ssh_port}" "$feed/$filename" "root@${BL}${ssh_host}${BR}:/tmp/${filename}"
+			ssh -p "${ssh_port}" "root@${ssh_host}" "
+				set -e
+				echo Running opkg:
+				opkg install --force-reinstall ${opkg_flags} '/tmp/${filename}'
+				rm '/tmp/${filename}'
+				gluon-reconfigure
+			"
+		else
+			# Some packages (e.g. procd-seccomp) seem to contain BuildPackage commands
+			# which do not generate *.ipk files. Till this point, I am not aware why
+			# this is happening. However, dropping a warning if the corresponding
+			# *.ipk is not found (maybe due to other reasons as well), seems to
+			# be more reasonable than aborting. Before this commit, the command
+			# has failed.
+			echo "Warning: ${pkg}*.ipk not found! Ignoring." 1>&2
+		fi
+
+	done
+done
--- a/contrib/run_qemu.sh
+++ b/contrib/run_qemu.sh
@@ -1,0 +1,15 @@
+#!/bin/sh
+
+# Note: You can exit the qemu instance by first pressing "CTRL + a" then "c".
+#       Then you enter the command mode of qemu and can exit by typing "quit".
+
+qemu-system-x86_64 \
+    -d 'cpu_reset' \
+    -enable-kvm \
+    -gdb tcp::1234 \
+    -nographic \
+    -netdev user,id=wan,hostfwd=tcp::2223-10.0.2.15:22 \
+    -device virtio-net-pci,netdev=wan,addr=0x06,id=nic1 \
+    -netdev user,id=lan,hostfwd=tcp::6080-192.168.1.1:80,hostfwd=tcp::2222-192.168.1.1:22,net=192.168.1.100/24 \
+    -device virtio-net-pci,netdev=lan,addr=0x05,id=nic2 \
+    "$@"
--- a/docs/dev/packages.rst
+++ b/docs/dev/packages.rst
@@ -3,6 +3,85 @@ Package development
 
 Gluon packages are OpenWrt packages and follow the same rules described at https://openwrt.org/docs/guide-developer/packages.
 
+Development workflow
+====================
+
+When you are developing packages, it often happens that you iteratively want to deploy
+and verify the state your development. There are two ways to verify your changes:
+
+1) One way is to rebuild the complete firmware, flash it, configure it and verify your
+   development then. This usually takes at least a few minutes to get your changes
+   working so you can test them. Especially if you iterate a lot, this becomes tedious.
+2) Another way is to rebuild only the package you are currently working on and
+   to deploy this package to your test system. Here not even a reboot is required.
+   This makes iterating relatively fast. Your test system could be real hardware or
+   even a qemu in most cases.
+
+Gluon provides scripts to enhance workflow 2). Here is an example illustrating
+the workflow using these scripts:
+
+.. code-block:: shell
+
+  # start a local qemu instance
+  contrib/run_qemu.sh output/images/factory/[...]-x86-64.img
+
+  # apply changes to the desired package
+  vi package/gluon-ebtables/files/etc/init.d/gluon-ebtables
+
+  # rebuild and push the package to the qemu instance
+  contrib/push_pkg.sh package/gluon-ebtables/
+
+  # test your changes
+  ...
+
+  # do more changes
+  ...
+
+  # rebuild and push the package to the qemu instance
+  contrib/push_pkg.sh package/gluon-ebtables/
+
+  # test your changes
+  ...
+
+  (and so on...)
+
+  # see help of the script for more information
+  contrib/push_pkg.sh -h
+  ...
+
+Features of ``push_pkg.sh``:
+
+* Works with compiled and non-compiled packages.
+
+  * This means it can be used in the development of C-code, Lua-Code and mostly any other code.
+
+* Works with native OpenWrt and Gluon packages.
+* Pushes to remote machines or local qemu instances.
+* Pushes multiple packages in in one call if desired.
+* Performs site.conf checks.
+
+Implementation details of ``push_pkg.sh``:
+
+* First, the script builds an opkg package using the OpenWrt build system.
+* This package is pushed to a *target machine* using scp:
+
+  * By default the *target machine* is a locally running x86 qemu started using ``run_qemu.sh``.
+  * The *target machine* can also be remote machine. (See the cli switch ``-r``)
+  * Remote machines are not limited to a specific architecture. All architectures supported by gluon can be used as remote machines.
+
+* Finally opkg is used to install/update the packages in the target machine.
+
+  * While doing this, it will not override ``/etc/config`` with package defaults by default. (See the cli switch ``-P``).
+  * While doing this, opkg calls the ``check_site.lua`` from the package as post_install script to validate the ``site.conf``. This means that the ``site.conf`` of the target machine is used for this validation.
+
+Note that:
+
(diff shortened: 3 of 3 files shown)


contrib/start_qemu.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+p=$(dirname $0)
+
+qemu-system-x86_64 \
+    -d 'cpu_reset' \
+    -enable-kvm \
+    -s \
```
mweinelt, line 8:
```
-s              shorthand for -gdb tcp::1234
```

mweinelt: FWIW: I'm opting for better readability here.

contrib/start_qemu.sh
```diff
@@ -0,0 +1,17 @@
...
+    -nographic \
+    -virtfs local,id=feeds,path=${p}/../openwrt/bin/packages/x86_64/,security_model=mapped-xattr,mount_tag=feeds \
+    -virtfs local,id=kmods,path=${p}/../openwrt/bin/targets/x86/64/packages/,security_model=mapped-xattr,mount_tag=kmods \
+    -netdev user,id=wan \
+    -device virtio-net-pci,netdev=wan,addr=0x06,id=nic1 \
+    -netdev user,id=lan,hostfwd=tcp::6080-192.168.1.1:80,hostfwd=tcp::2222-192.168.1.1:22,net=192.168.1.100/24 \
```
mweinelt, line 14: This works well for the config mode, but not for when the normal runmode.

lemoer: Of course. This could be resolved later. :-)

lemoer:
I would rather like to discuss this from a conceptual point of view first. If we decide that we want to have such a feature and which technologies should be used, we can go into detail later.

lemoer:
I updated the patch to remove the ugly `9pfs` stuff and completely rely on ssh. Please review and comment.

lemoer:
This commit significantly reduces iteration time for package changes:
``` shell
# time sh scripts/push_pkg.sh package/gluon-mesh-vpn-wireguard
...

________________________________________________________
Executed in    1.57 secs   fish           external 
   usr time  740.34 millis  1032.00 micros  739.31 millis 
   sys time  336.54 millis  197.00 micros  336.34 millis 
```
This time includes rebuilding the package and deploying it.

AiyionPrime:
Just tested, works very well.

NeoRaider:
I think it would make sense to put this in the `contrib` directory, where other scripts that aren't directly involved in Gluon's build process live.

scripts/push_pkg.sh
```diff
@@ -0,0 +1,129 @@
...
+		echo "ERROR: ${pkgdir} does not contain a Makefile"
+		exit 1
+	fi
+
+	if ! grep BuildPackage "${pkgdir}/Makefile" > /dev/null; then
+		echo "ERROR: ${pkgdir}/Makefile does not contain a BuildPackage command"
+		exit 1
+	fi
+
+	opkg_packages="$(grep BuildPackage "${pkgdir}/Makefile" | cut -d',' -f 2 | tr -d ')' | tr -d '\n')"
```
NeoRaider, line 78: A better way find out what packages a Makefile contains would be:
```sh
make TOPDIR="${topdir}" -C "${pkgdir}" DUMP=1 | awk '/^Package: / { print $2 }'
```
The DUMP feature is also what OpenWrt uses internally for enumerate packages.

scripts/push_pkg.sh
```diff
@@ -0,0 +1,129 @@
...
+		fi
+
+		# shellcheck disable=SC2029
+		if [ -n "$filename" ]; then
+			scp -P "${ssh_port}" "$feed/$filename" "root@${BL}${ssh_host}${BR}:/tmp/${filename}"
+			echo Running opkg:
+			ssh -p "${ssh_port}" "root@${ssh_host}" opkg install --force-reinstall "/tmp/${filename}"
+			echo ok
+			ssh -p "${ssh_port}" "root@${ssh_host}" rm "/tmp/${filename}"
+			ssh -p "${ssh_port}" "root@${ssh_host}" gluon-reconfigure
```
NeoRaider, line 117: How about putting all these commands into a single 'ssh' run? Would just need a `set -e` to abort on the first error.

package/gluon.mk
```diff
@@ -20,12 +20,31 @@ shell-unescape = $(subst @1,@,$(subst @2,$(newline),$(1)))
...
+END__GLUON__CHECK__SITE__LIB
+lua /tmp/check_site_lib.lua <<'END__GLUON__CHECK__SITE'
+$(call shell-verbatim,cat '$(1)')
+END__GLUON__CHECK__SITE
+rm /tmp/check_site_lib.lua
+echo ok
+endef
+
+else
+
```
NeoRaider, line 39: Hmm, I'm don't like this change... I'll see if I can come up with an alternative solution.

scripts/push_pkg.sh
```diff
@@ -0,0 +1,129 @@
...
+	echo ' will be regenerated. The variables from openwrt/.config are already'
+	echo ' automatically used for this script.'
+	echo
+}
+
+while getopts "p:r:hb" opt
+do
+	case $opt in
+		p) ssh_port="${OPTARG}";;
+		r) ssh_host="${OPTARG}"; ssh_port=22;;
```
NeoRaider, line 31: The behaviour of the option handling is somewhat confusing... If `-r` is used after `-p`, the `-p` will be ignored.

How about selecting the port after all options have been parsed, using the following logic?

- If `-p` was given, always use that port
- Else-if `-r` was given, default to port 22
- Else, default to port 2223

scripts/push_pkg.sh
```diff
@@ -0,0 +1,129 @@
...
+	esac
+done
+shift $(( OPTIND - 1 ))
+
+if [ "$build_only" -eq 0 ]; then
+	OPENWRT_BOARD="$(ssh -p "${ssh_port}" "root@${ssh_host}" sh -c \''. /etc/os-release; echo "$OPENWRT_BOARD"'\')"
+	OPENWRT_ARCH="$(ssh -p "${ssh_port}" "root@${ssh_host}" sh -c \''. /etc/os-release; echo "$OPENWRT_ARCH"'\')"
+
+	# check target
+	if ! grep "CONFIG_TARGET_ARCH_PACKAGES=\"${OPENWRT_ARCH}\"" "${topdir}/.config" > /dev/null; then
```
NeoRaider, line 44: Use `grep -q` instead of `grep >/dev/null`

scripts/run_qemu.sh
```diff
@@ -0,0 +1,12 @@
...
+
+qemu-system-x86_64 \
+    -d 'cpu_reset' \
+    -enable-kvm \
+    -s \
+    -nographic \
+    -netdev user,id=wan,hostfwd=tcp::2223-10.0.2.15:22 \
+    -device virtio-net-pci,netdev=wan,addr=0x06,id=nic1 \
+    -netdev user,id=lan,hostfwd=tcp::6080-192.168.1.1:80,hostfwd=tcp::2222-192.168.1.1:22,net=192.168.1.100/24 \
+    -device virtio-net-pci,netdev=lan,addr=0x05,id=nic2 \
```
NeoRaider, line 11: In my qemu script I also have the following lines:
```
        -chardev stdio,id=char0,signal=off \
        -serial chardev:char0 \
```
This connects the serial console of the VM to the terminal running the script. `signal=off` might be controversial - it makes `Ctrl-C` etc. work as expected in the serial console instead of terminating qemu, but in consequence you need a second terminal if you actually want to kill qemu.

lemoer: The proper way to exit qemu with `signal=off` is to press `Ctrl-a`, `Ctrl-c` and type `quit`.

NeoRaider: Good to know, I wasn't aware qemu had an escape key like that.

lemoer: I added a note to `run_qemu.sh`, so hopefully other users will stumble across that hint aswell.

package/gluon.mk
lemoer: Please let me know when you have decided what to do here.

Or maybe you could clarify why you don't like the approach? Then I could think about a solution too.

mweinelt:
The walkthrough in the initial post sounds good. It would be good to have documented.

package/gluon.mk
mweinelt: It has been a month. Is this something that we can address another time or does it prevent merging this?

lemoer:
@mweinelt done.

docs/dev/packages.rst
```diff
@@ -3,6 +3,72 @@ Package development
...
+   to deploy this package to your test system. Here not even a reboot is required.
+   This makes iterating relatively fast. Your test system could be real hardware or
+   even a qemu in most cases.
+
+Gluon provides scripts to enhance workflow 2):
+
+.. code-block:: shell
+
+  # start a local qemu instance
+  contrib/run_qemu.sh output/images/factory/[...].img
```
mweinelt, line 25:
```suggestion
  contrib/run_qemu.sh output/images/factory/[...]-x86-64.img
```

docs/dev/packages.rst
```diff
@@ -3,6 +3,72 @@ Package development
...
+
+  # do your changes in the file you want to patch
```
mweinelt, line 27:
```suggestion
  # apply changes to the desired package
```

docs/dev/packages.rst
```diff
@@ -3,6 +3,72 @@ Package development
...
+  # rebuild and push the package
+  contrib/push_pkg.sh package/gluon-ebtables/
+
+  # test your changes
+  ...
+
+  # do more changes
+  ...
+
+  # rebuild and push the package
```
mweinelt, line 39:
```suggestion
  # rebuild and push the package to the running vm
```

docs/dev/packages.rst
```diff
@@ -3,6 +3,72 @@ Package development
...
+
+  (and so on...)
+
+  # see help of the script for more information
+  contrib/push_pkg.sh -h
+  ...
+
+Implementation details of ``push_pkg.sh``:
+
+* Currently this is based on ssh/scp.
```
mweinelt, line 53:
```suggestion
* File transfers use the scp utility.
```

docs/dev/packages.rst
```diff
@@ -3,6 +3,72 @@ Package development
...
+* Opkg is used to install/update the packages in the remote machine.
```
mweinelt, line 54:
```suggestion
* Packages are installed/updated using ``opkg`` on the virtual machine.
```

docs/dev/packages.rst
```diff
@@ -3,6 +3,72 @@ Package development
...
+* It does neither build nor push dependencies of the packages automatically.
+* It will not override /etc/config with package defaults by default. (See the cli switch ``-P``).
```
mweinelt, line 56:
```suggestion
* It will not override ``/etc/config`` with package defaults by default. (See the cli switch ``-P``).
```

docs/dev/packages.rst
```diff
@@ -3,6 +3,72 @@ Package development
...
+
+Benefits:
+
+* This works with compiled and non-compiled packages.
+* This works with native OpenWrt and Gluon packages.
+* This even performs the ``check_site.lua`` checks as they are integrated as post_install scripts into the openwrt package.
+* It works for all architectures/targets.
```
mweinelt, line 63: Except `run_qemu.sh` uses KVM on x86_64. The relationship between `run_qemu.sh` and developing on actual hardware isn't clear from this.

docs/dev/packages.rst
```diff
@@ -3,6 +3,72 @@ Package development
...
+
+Note that:
+
+* If you add new packages, you must run ``make update config GLUON_TARGET=...``.
+* You can change the target via ``make config GLUON_TARGET=...``.
+* Sometimes when things break, you can heal them by compiling a package with its dependencies: ``cd openwrt; make package/gluon-ebtables/clean; make package/gluon-ebtables/compile; cd ..``.
+* You can exit qemu by pressing ``CTRL + a`` and ``c`` afterwards.
+* ``push_pkg.sh`` can also be used to push packages to other targets than qemu or x86-64.
```
mweinelt, line 71:
```suggestion
* ``push_pkg.sh`` can also be used to push packages to other targets than your native arch.
```

Is this the intended meaning?

contrib/run_qemu.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+# Note: You can exit the qemu instance by first pressing "CTRL + a" then "c".
+#       Then you enter the command mode of qemu and can exit by typing "quit".
+
+qemu-system-x86_64 \
+    -d 'cpu_reset' \
+    -enable-kvm \
+    -s \
```
mweinelt, line 9:
```
-s              shorthand for -gdb tcp::1234
```

No harm in using the longer option, since it is much clearer what it does.

contrib/push_pkg.sh
```diff
@@ -0,0 +1,153 @@
...
+build_only=0
+preserve_config=1
+
+print_help() {
+	echo "$0 [OPTIONS] PACAKGE_DIR"
+	echo ""
+	echo " -h          print this help"
+	echo " -r HOST     use ssh host (default port will be 22)"
+	echo " -p PORT     use ssh port"
+	echo " -b          build only, do not push anywhere"
```
mweinelt, line 18:
```suggestion
	echo " -b          build only, do not push"
```

contrib/push_pkg.sh
```diff
@@ -0,0 +1,153 @@
...
+		*) ;;
+	esac
+done
+shift $(( OPTIND - 1 ))
+
+[ -z "$ssh_port" ] && ssh_port=2223
+
+if [ "$build_only" -eq 0 ]; then
+	OPENWRT_BOARD="$(ssh -p "${ssh_port}" "root@${ssh_host}" sh -c \''. /etc/os-release; echo "$OPENWRT_BOARD"'\')"
+	OPENWRT_ARCH="$(ssh -p "${ssh_port}" "root@${ssh_host}" sh -c \''. /etc/os-release; echo "$OPENWRT_ARCH"'\')"
```
mweinelt, line 48: What is the idea behind the OPENWRT_ scope here? I'd say `REMOTE_BOARD` and `REMOTE_ARCH` makes more sense, but I think we'd be fine without the qualifier.

mweinelt (changes requested):
Looks pretty good, a few remarks.

lemoer:
@mweinelt Thanks, I implemented your suggestions.

package/gluon.mk
mweinelt: And two more weeks went by. :disappointed:

AiyionPrime: @neoraider As this becomes a rather stuck feedback-loop; how about meeting up in mumble for a few minutes tomorrow evening? Maybe that makes it easier to brainstorm and clearer for @lemoer what benefits the alternative solution should provide.

NeoRaider: Sorry for the delay.

The issue with this piece of code is that `GluonCheckSite` is put into the ipk packages as a postinst script; by conditionally changing the definition, you are also changing the package files (which may not be rebuilt when a regular build is run later?).

Optimally, we wouldn't need different logic in `GluonCheckSite` for regular and `push_pkg.sh` builds at all. If it can't be avoided, the condition must be checked at package install time, not build time.

lemoer: Thanks for the feedback.

> Optimally, we wouldn't need different logic in GluonCheckSite for regular and push_pkg.sh builds at all.

I didn't find a way to do this.

> If it can't be avoided, the condition must be checked at package install time, not build time.

I implemented this one.

scripts/check_site.lua
```diff
@@ -1,4 +1,3 @@
-local cjson = require 'cjson'
```
NeoRaider, line 1: I'm currently evaluating whether we can switch to `lua-jsonc` unconditionally, removing the need to support two different loaders in `check_site.lua`.

NeoRaider: https://github.com/freifunk-gluon/gluon/pull/2234

package/gluon.mk
NeoRaider: > > Optimally, we wouldn't need different logic in GluonCheckSite for regular and push_pkg.sh builds at all.
> 
> I didn't find a way to do this.

I *think* I have an idea how to achieve this, but let's get https://github.com/freifunk-gluon/gluon/pull/2234 merged first.

NeoRaider: I have implemented this in #2237.

lemoer:
This PR is now implicitly also "level-2 blocked" by #2271, because #2271 is now blocking/influencing #2237. I would really like to get this merged soon.

NeoRaider:
#2237 is merged, please rebase!

lemoer:
@NeoRaider Done.

I also squashed it.

contrib/push_pkg.sh
```diff
@@ -0,0 +1,156 @@
...
+		h) print_help; exit 0;;
+		*) ;;
+	esac
+done
+shift $(( OPTIND - 1 ))
+
+[ -z "$ssh_port" ] && ssh_port=2223
+
+if [ "$build_only" -eq 0 ]; then
+	REMOTE_OPENWRT_BOARD="$(ssh -p "${ssh_port}" "root@${ssh_host}" sh -c \''. /etc/os-release; echo "$OPENWRT_BOARD"'\')"
```
NeoRaider, line 49:
```suggestion
	REMOTE_OPENWRT_BOARD="$(ssh -p "${ssh_port}" "root@${ssh_host}" '. /etc/os-release; echo "$OPENWRT_BOARD"')"
```
SSH already runs the passed command through a shell.

It would also be nice to merge both SSH commands into one and split the resulting string afterwards to improve performance.

contrib/push_pkg.sh
```diff
@@ -0,0 +1,156 @@
...
+				echo not found
+			fi
+		done
+
+		# shellcheck disable=SC2029
+		if [ -n "$filename" ]; then
+			scp -P "${ssh_port}" "$feed/$filename" "root@${BL}${ssh_host}${BR}:/tmp/${filename}"
+			ssh -p "${ssh_port}" "root@${ssh_host}" "
+				set -e
+				if [ \"$preserve_config\" -eq 1 ]; then
```
NeoRaider, line 125:
```suggestion
				if [ '$preserve_config' -eq 1 ]; then
```
Use single quotes for the remote commands whenever you don't need variables from the remote side interpreted.

contrib/push_pkg.sh
lemoer: Done.

contrib/push_pkg.sh
lemoer: Done.

lemoer:
@NeoRaider Updated once again.

contrib/push_pkg.sh
```diff
@@ -0,0 +1,160 @@
...
+					echo \"Restoring /etc/config\"
+					for f in \$(find /etc/config.bak/ -mindepth 1); do
+						echo \"Restoring /etc/config/\$(basename \"\$f\")\"
+						mv -f \"\$f\" /etc/config/
+					done
+					rmdir /etc/config.bak
+				fi
+				rm \"/tmp/${filename}\"
+				gluon-reconfigure
```
NeoRaider, line 147: There are a lot more `\"` in this section that could be replaced with `'`.

lemoer: Done.

lemoer:
@NeoRaider And updated once again.

contrib/push_pkg.sh
```diff
@@ -0,0 +1,160 @@
...
+				opkg install --force-reinstall '/tmp/${filename}'
+				echo Running opkg finished: ok.
+				if [ '$preserve_config' -eq 1 ]; then
+					echo 'Restoring /etc/config'
+					for f in \$(find /etc/config.bak/ -mindepth 1); do
+						echo \"Restoring /etc/config/\$(basename \"\$f\")\"
+						mv -f \"\$f\" /etc/config/
+					done
+					rmdir /etc/config.bak
+				fi
```
NeoRaider, line 145: Hmm, thinking about this section again, you have actually found a bug in Gluon's package Makefiles!

opkg should never overwrite config files on package upgrades. The only reason the `preserve_config` flag is even necessary is that Gluon doesn't correctly specify `conffiles` in its Makefiles.

I'll prepare a PR to fix our Makefiles, and then we can get rid of another few lines of this script :)

NeoRaider: #2276

You can keep the `preserve_config` flag, but it should map to the `--force-maintainer` option of opkg instead of handling config files manually.

lemoer: Implemented.

lemoer:
I implemented several changes now. Let's just freeze this now and merge it once #2276 is merged. If new ideas come, we can add them later in new PRs or directly on the master branch.

PR #2024, opened 2020-05-11
build: ignore deactivated feeds

SvenRoederer:
The OpenWrt feeds.conf.defaults contains some feeds that are commented out and not active. 
But this feeds are returned by the default_feeds.sh script anyway. Limit the script to only return
active feeds, by filtering out lines starting with '#'.

### DIFF
--- a/scripts/default_feeds.sh
+++ b/scripts/default_feeds.sh
@@ -1,1 +1,2 @@
-DEFAULT_FEEDS="$(awk '{print $2}' openwrt/feeds.conf.default)"
+# list feeds which don't start with #
+DEFAULT_FEEDS="$(awk '!/^#/ {print $2}' openwrt/feeds.conf.default)"


mweinelt:
```
❯ cat feeds.conf.default
src-git packages https://git.openwrt.org/feed/packages.git
src-git luci https://git.openwrt.org/project/luci.git
src-git routing https://git.openwrt.org/feed/routing.git
src-git telephony https://git.openwrt.org/feed/telephony.git
#src-git video https://github.com/openwrt/video.git
#src-git targets https://github.com/openwrt/targets.git
#src-git management https://github.com/openwrt-management/packages.git
#src-git oldpackages http://git.openwrt.org/packages.git
#src-link custom /usr/src/openwrt/custom-feed
```

scripts/default_feeds.sh
```diff
@@ -1 +1,2 @@
-DEFAULT_FEEDS="$(awk '{print $2}' openwrt/feeds.conf.default)"
+# only list feeds which don't start with #
+DEFAULT_FEEDS="$(grep -v ^# openwrt/feeds.conf.default | awk '{print $2}')"
```
NeoRaider, line 2: Combining grep and awk in a single line is rarely useful. Shorter:
```sh
awk '!/^#/ {print $2}' openwrt/feeds.conf.default
```
Also, remove the word "only" from the comment.

NeoRaider:
The commit message should mention that this only affects OpenWrt master, as release branches never have commented feed entries.

scripts/default_feeds.sh
SvenRoederer: @NeoRaider  your awk skills are better than mine ... :-)

changed as suggested

PR #2025, opened 2020-05-12
Github Actions improvements

NeoRaider:
During debugging of a spurious build failure (that hasn't reoccurred yet...) I found some potential for improvement in our Github Actions setup.

- Set CONFIG_AUTOREMOVE to avoid running out of disk space
  - Do our Jenkins jobs build incrementally or from scratch? If the latter is the case, the flag should be set for Jenkins as well
- Set BUILD_LOG=1 and save logs as separate artifact
  - This step is run even when the build fails
- Some minor script cleanup


mweinelt:
> Set CONFIG_AUTOREMOVE to avoid running out of disk space
> 
>     * Do our Jenkins jobs build incrementally or from scratch? If the latter is the case, the flag should be set for Jenkins as well

They do incremental builds.

contrib/actions/generate-actions.py
```diff
@@ -25,13 +25,19 @@
...
-        run: bash contrib/actions/install-dependencies.sh
+        run: sudo contrib/actions/install-dependencies.sh
       - name: Build
-        run: bash contrib/actions/run-build.sh {target_name}
+        run: contrib/actions/run-build.sh {target_name}
+      - name: Archive build logs
+        if: ${{{{ !cancelled() }}}}
+        uses: actions/upload-artifact@v1
+        with:
+          name: {target_name}-logs
```
mweinelt, line 35:
```suggestion
          name: {target_name}/logs
```

IMO improves readability

contrib/actions/generate-actions.py
```diff
@@ -25,13 +25,19 @@
...
+          path: openwrt/logs
       - name: Archive build output
         uses: actions/upload-artifact@v1
         with:
-          name: {target_name}
+          name: {target_name}-output
```
mweinelt, line 40:
```suggestion
          name: {target_name}/output
```

docs/user/getting_started.rst
```diff
@@ -208,6 +208,12 @@ GLUON_TARGET
 Special variables
 .................
 
+GLUON_AUTOREMOVE
+  Setting ``GLUON_AUTOREMOVE=1`` enable the ``CONFIG_AUTOREMOVE`` OpenWrt setting, which will delete package build
```
mweinelt, line 212:
```suggestion
  Setting ``GLUON_AUTOREMOVE=1`` enables the ``CONFIG_AUTOREMOVE`` OpenWrt setting, which will delete package build
```

contrib/actions/generate-actions.py
NeoRaider: Ah, I didn't know this was valid.

NeoRaider:
```
Artifact name is not valid: ar71xx-tiny/logs. It cannot contain '\', '/', "', ':', '<', '>', '|', '*', and '?'
```
Options:
- `{target_name}.logs`
- `{target_name}@logs`
- `{target_name}+logs`
- `{target_name}_logs`
- `{target_name}~logs`
- ...

Also, it looks like they failed to escape `'` and/or `"` correctly in their error message ;)

blocktrron: I would sick with `-` then

NeoRaider: I'm leaning towards `{target_name}_logs` or maybe `{target_name}--logs`. @mweinelt @rotanid ?

mweinelt: I'm leaning towards anything but `-`. it needs to be visually different from `target-subtarget`.

blocktrron: Go with `_` then

PR #2026, opened 2020-05-15
x86: enable /dev/mem

belzebub40k:
PC Engine APU Boards are quite popular and it would be nice if we could update the BIOS directly in Gluon. In general this is already possible via flashrom but for it to work `CONFIG_KERNEL_DEVMEM` has to be enabled.

### DIFF
--- a/targets/x86.inc
+++ b/targets/x86.inc
@@ -1,5 +1,6 @@
 config 'CONFIG_VDI_IMAGES=y'
 config 'CONFIG_VMDK_IMAGES=y'
+config 'CONFIG_KERNEL_DEVMEM=y'
 
 packages {
 	'kmod-3c59x',


txt-file:
In my opinion this should be done upstream (in OpenWrt).

rotanid:
@belzebub40k i also have some APU2 out there with Gluon and tried to discuss this topic in September 2019 on the Gluon IRC Channel.
In the end i ended up building a special image with your suggested change in order to update the nodes' coreboot firmware - but as the disabled DEVMEM access seems to be a security feature, i didn't enable it in out default image.

slightly off-topic: is there an important coreboot firmware change for APU2 which makes an update necessary? i last updated in autumn 2019

belzebub40k:
> @belzebub40k i also have some APU2 out there with Gluon and tried to discuss this topic in September 2019 on the Gluon IRC Channel.
> In the end i ended up building a special image with your suggested change in order to update the nodes' coreboot firmware - but as the disabled DEVMEM access seems to be a security feature, i didn't enable it in out default image.

Ok, didn't knew that this feature was already discussed and rejected.

> slightly off-topic: is there an important coreboot firmware change for APU2 which makes an update necessary? i last updated in autumn 2019

None that I'm aware of. I just did some tests on an APU and noticed the BIOS was outdated and wanted to update it to the latest version.

rotanid:
> rejected

not really, i didn't really get feedback on IRC when talking about it :D

Issue #1974, opened 2020-04-07
RFC: Multidomain: Primary vs. Pretty domain names


mweinelt:
Currently with multidomain builds all domain names within a domain are treated equally and only the one selected is announced, even though that is most likely just be a pretty name. We however do need the technical domain name to be able to link them logically to our gateways, who simply cannot announce themselves as part of all the possible pretty domain codes.

In practice that means that in tools like meshviewer we have the multiple domain codes (pretty and logical) for a single L2 domain. Consider the following configuration:

```lua
        domain_names = {
                dom1 = 'Domain 1',
                ffda_da_110 = 'Darmstadt: Stadtzentrum',
                ffda_da_120 = 'Darmstadt: Mollerstadt',
                ffda_da_130 = 'Darmstadt: Hochschulviertel',
                ffda_da_210 = 'Darmstadt: Johannesviertel',
                ffda_da_220_230 = 'Darmstadt: Martinsviertel',
                ffda_da_270 = 'Darmstadt: Bürgerparkviertel',
                ffda_da_310 = 'Darmstadt: Am Oberfeld',
                ffda_da_320 = 'Darmstadt: Mathildenhöhe'
        },
        hide_domains = { 'dom1' },
```

Multiple nodes will happily announce all of the above. But gateways can only be part of *one* of these. 

In my opinion we need to clean up the meaning of the domain_code variable to be more technically helpful. Any way I can imagine would require some migration strategy, that I have no clear idea of yet.

I'd imagine the following would be a more proper configuration structure for what is actually being required:

```lua
        domain_name = "dom1",  -- this is the logical domain name, that is used for configuration and for the respondd domain_code
        domain_pretty_names = {
                -- these are the names being offered in the dropdown
                ffda_da_110 = 'Darmstadt: Stadtzentrum',
                ffda_da_120 = 'Darmstadt: Mollerstadt',
                ffda_da_130 = 'Darmstadt: Hochschulviertel',
                ffda_da_210 = 'Darmstadt: Johannesviertel',
                ffda_da_220_230 = 'Darmstadt: Martinsviertel',
                ffda_da_270 = 'Darmstadt: Bürgerparkviertel',
                ffda_da_310 = 'Darmstadt: Am Oberfeld',
                ffda_da_320 = 'Darmstadt: Mathildenhöhe'
        },
        hide_domains = {}  -- domain_name wouldn't be part of the dropdown, but other domain names might still need hiding, idk
```

[...]

PR #2027, opened 2020-05-16
Emit primary_domain_code via respondd

lemoer:
Implements #1974

### DIFF
--- a/package/gluon-respondd/src/respondd-nodeinfo.c
+++ b/package/gluon-respondd/src/respondd-nodeinfo.c
@@ -64,6 +64,10 @@ static struct json_object * get_site_code(void) {
 	return ret;
 }
 
+static struct json_object * get_primary_domain_code(void) {
+	return gluonutil_wrap_and_free_string(gluonutil_get_primary_domain());
+}
+
 static struct json_object * get_domain_code(void) {
 	return gluonutil_wrap_and_free_string(gluonutil_get_domain());
 }
@@ -124,8 +128,10 @@ struct json_object * respondd_provider_nodeinfo(void) {
 
 	struct json_object *system = json_object_new_object();
 	json_object_object_add(system, "site_code", get_site_code());
-	if (gluonutil_has_domains())
+	if (gluonutil_has_domains()) {
 		json_object_object_add(system, "domain_code", get_domain_code());
+		json_object_object_add(system, "primary_domain_code", get_primary_domain_code());
+	}
 	json_object_object_add(ret, "system", system);
 
 	return ret;
--- a/package/libgluonutil/src/libgluonutil.c
+++ b/package/libgluonutil/src/libgluonutil.c
@@ -225,21 +225,24 @@ char * gluonutil_get_primary_domain(void) {
 	const char *domain_path_fmt = "/lib/gluon/domains/%s.json";
 	char domain_path[strlen(domain_path_fmt) + strlen(domain_code)];
 	snprintf(domain_path, sizeof(domain_path), domain_path_fmt, domain_code);
-	free(domain_code);
 
 	char primary_domain_path[PATH_MAX+1];
 	char *primary_domain_code;
 	ssize_t len = readlink(domain_path, primary_domain_path, PATH_MAX);
 	if (len < 0) {
 		// EINVAL = file is not a symlink = the domain itself is the primary domain
-		if (errno != EINVAL)
+		if (errno != EINVAL) {
+			free(domain_code);
 			return NULL;
+		}
 
-		primary_domain_code = basename(domain_path);
-	} else {
-		primary_domain_path[len] = '\0';
-		primary_domain_code = basename(primary_domain_path);
+		return domain_code;
 	}
+
+	free(domain_code);
+
+	primary_domain_path[len] = '\0';
+	primary_domain_code = basename(primary_domain_path);
 
 	char *ext_begin = strrchr(primary_domain_code, '.');
 	if (!ext_begin)
--- a/package/libgluonutil/src/libgluonutil.c
+++ b/package/libgluonutil/src/libgluonutil.c
@@ -227,21 +227,25 @@ char * gluonutil_get_primary_domain(void) {
 	snprintf(domain_path, sizeof(domain_path), domain_path_fmt, domain_code);
 	free(domain_code);
 
-	char primary_domain_path[PATH_MAX] = "";
-	if ((readlink(domain_path, primary_domain_path, PATH_MAX) < 0) && (errno == EINVAL)) {
+	char primary_domain_path[PATH_MAX+1] = "";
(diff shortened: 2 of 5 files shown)


package/libgluonutil/src/libgluonutil.c
```diff
@@ -212,6 +214,36 @@ char * gluonutil_get_domain(void) {
...
+
+	char *primary_domain_code = basename(primary_domain_path);
+	char *ext_begin = strrchr(primary_domain_code, '.');
+	if (ext_begin) {
+		// strip .json from filename
+		*ext_begin = '\0';
+		return strdup(primary_domain_code);
+	}
+
+	return NULL;
```
NeoRaider, line 244: Let's move the success return to the end of the function, and turn this error case into a `if (!ext_begin)` branch instead.

package/libgluonutil/src/libgluonutil.c
```diff
@@ -212,6 +214,36 @@ char * gluonutil_get_domain(void) {
...
+	char *domain_code = gluonutil_get_domain();
+	if (!domain_code)
+		return NULL;
+
+	const char *domain_path_fmt = "/lib/gluon/domains/%s.json";
+	char domain_path[strlen(domain_path_fmt) + strlen(domain_code)];
+	snprintf(domain_path, sizeof(domain_path), domain_path_fmt, domain_code);
+	free(domain_code);
+
+	char primary_domain_path[PATH_MAX] = "";
```
NeoRaider, line 230: Let's make this buffer `PATH_MAX+1`, but still pass `PATH_MAX` to `readlink`, so it is 0-terminated even for oversized paths (however that is supposed to happen). It is not necessary to initialize the whole array: Setting byte `len` for `len` returned by readlink is sufficient.

package/libgluonutil/src/libgluonutil.c
```diff
@@ -212,6 +214,36 @@ char * gluonutil_get_domain(void) {
...
+	if ((readlink(domain_path, primary_domain_path, PATH_MAX) < 0) && (errno == EINVAL)) {
```
NeoRaider, line 231: Missing error handling for `errno != EINVAL`.

package/libgluonutil/src/libgluonutil.c
```diff
@@ -212,6 +214,36 @@ char * gluonutil_get_domain(void) {
...
+		// EINVAL = file is not a symlink = the domain itself is the primary domain
+		strncpy(primary_domain_path, domain_path, PATH_MAX);
```
NeoRaider, line 233: This strncpy can be avoided: just call `basename` either on `domain_path` or `primary_domain_path` depending on readlink success.

package/gluon-respondd/src/respondd-nodeinfo.c
```diff
@@ -124,8 +128,10 @@ struct json_object * respondd_provider_nodeinfo(void) {
 
 	struct json_object *system = json_object_new_object();
 	json_object_object_add(system, "site_code", get_site_code());
-	if (gluonutil_has_domains())
+	if (gluonutil_has_domains()) {
 		json_object_object_add(system, "domain_code", get_domain_code());
+		json_object_object_add(system, "primary_domain_code", get_primary_domain_code());
```
NeoRaider, line 133: Hmm, `primary_domain_code` or `domain_code_primary`? I kinda like the latter, as the fields will be next to each other when sorted by keys.

lemoer: I would rather prefer the first. The second one might be confusing. It could also suggest, that there is a "primary" object with a special meaning or so...

lemoer:
@NeoRaider Updated and tested.

lemoer:
I added the field to the staus-page as well.

NeoRaider:
> I added the field to the staus-page aswell.

IMO this is too technical for the status page. It is only interesting for software that wants to group nodes by their actual mesh domain, independently of the user-visible domain name, but it should not be shown to the user.

lemoer:
In theory, I would agree with you. But in practice it happens very often that you want to know the underlying layer 2 network. E.g. if you want to go to the gateway server and figure out which bat device you have to look at. In our setup, we have the scheme `bat12 = "Domäne 12"` for the primary domains. Having yet another source of truths here is cumbersome.

mweinelt:
IMO adding it to the status page is a benefit, since we otherwise hide important technical details from users. The real domain name is a good indicator if whether two nodes in different "pretty name"-domains can mesh with each other.

mweinelt:
But please don't make it a separate value, just add it as a detail to the already existing domain field.

Maybe something like this:
```
**Domain**
Darmstadt: Waldkolonie (dom4)
```

lemoer:
@mweinelt Good idea! I implemented it like this.

mweinelt:
Live @ http://[2001:67c:2ed8:1001:e695:6eff:fe41:f7a0]/cgi-bin/status

package/libgluonutil/src/libgluonutil.c
```diff
@@ -212,6 +214,40 @@ char * gluonutil_get_domain(void) {
...
+	char *domain_code = gluonutil_get_domain();
+	if (!domain_code)
+		return NULL;
+
+	const char *domain_path_fmt = "/lib/gluon/domains/%s.json";
+	char domain_path[strlen(domain_path_fmt) + strlen(domain_code)];
+	snprintf(domain_path, sizeof(domain_path), domain_path_fmt, domain_code);
+	free(domain_code);
+
+	char primary_domain_path[PATH_MAX+1] = "";
```
NeoRaider, line 230: Remove this initialization, only set the terminating 0 at the length returned by `readlink`.

lemoer:
@NeoRaider done.

package/libgluonutil/src/libgluonutil.c
```diff
@@ -212,6 +214,42 @@ char * gluonutil_get_domain(void) {
...
+
+	char primary_domain_path[PATH_MAX+1];
+	char *primary_domain_code;
+	ssize_t len = readlink(domain_path, primary_domain_path, PATH_MAX);
+	if (len < 0) {
+		// EINVAL = file is not a symlink = the domain itself is the primary domain
+		if (errno != EINVAL)
+			return NULL;
+
+		primary_domain_code = basename(domain_path);
```
NeoRaider, line 238: Hmm, I guess we could just return `domain_code` in this branch after all if we move the `free` below this point.

The following lines could be moved out of the `else` branch then.

NeoRaider:
It seems you have accidentally pushed a few unrelated commits into this branch.

lemoer:
Oh, sorry. I fixed this.

NeoRaider:
Thanks! I squashed the fixups and merged this.

Issue #1178, opened 2017-07-07
Feature Request: use 802.11s features for better control of meshlinks


A-Kasper:
For better performance it could be nice to have control about some Mesh-Links
802.11s features some options we could use to get control about the mesh-links of a node. 

For example we could provide white- or blacklisting of nodes the node is meshing with:
https://github.com/o11s/open80211s/wiki/HOWTO#advanced-tinkering

other possible 802.11s parameters are:
Possible mesh parameters are:
 - mesh_retry_timeout
 - mesh_confirm_timeout
 - mesh_holding_timeout
 - mesh_max_peer_links
 - mesh_max_retries
 - mesh_ttl
 - mesh_element_ttl
 - mesh_auto_open_plinks
 - mesh_hwmp_max_preq_retries
 - mesh_path_refresh_time
 - mesh_min_discovery_timeout
 - mesh_hwmp_active_path_timeout
 - mesh_hwmp_preq_min_interval
 - mesh_hwmp_net_diameter_traversal_time
 - mesh_hwmp_rootmode
 - mesh_hwmp_rann_interval
 - mesh_gate_announcements
 - mesh_fwding
 - mesh_sync_offset_max_neighor
 - mesh_rssi_threshold
 - mesh_hwmp_active_path_to_root_timeout
 - mesh_hwmp_root_interval
 - mesh_hwmp_confirmation_interval
 - mesh_power_mode
 - mesh_awake_window
 - mesh_plink_timeout

It would also be nice if iw dev mesh0 station dump parameters like the bitrates and expected throughput could be shown at the statuspage. (btw: I think the beacon interval of 100 could be longer. )

NeoRaider:
I think most 11s options aren't very interesting for us, as we do all the routing decisions in a higher-layer routing protocol and only use 11s as a replacement for adhoc mode.

Some work on allowing to blacklist links was done in freifunk-gluon/packages#118, but it is not finished.

Increasing the beacon interval is probably a good idea, we just need to decide on a good value (also, all VIFs use the same beacon interval, so changes will affect the AP interfaces as well).

CodeFetch:
[...]

PR #2028, opened 2020-05-21
gluon-core: add support for configuring the beacon interval

blocktrron:
This adds support for the beacon interval to be set on a per-band base.
This has the potential to reduce the amount of airtime used up for
sending beacon frames.

### DIFF
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -119,6 +119,10 @@ regdom \: optional
 wifi24 \: optional
     WLAN configuration for 2.4 GHz devices.
     ``channel`` must be set to a valid wireless channel for your radio.
+    ``beacon_interval`` can be specified to set a custom beacon interval in
+    time units (TU). A time unit is equivalent to 1024 microseconds.
+    If not set, the default value of 100 TU (=102.4 ms) is used.
+
 
     There are currently two interface types available. You may choose to
     configure any subset of them:
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
@@ -191,6 +191,7 @@ wireless.foreach_radio(uci, function(radio, index, config)
 
 	local channel = get_channel(radio, config)
 	local htmode = get_htmode(radio)
+	local beacon_interval = config.beacon_interval()
 
 	uci:delete('wireless', radio_name, 'disabled')
 
@@ -225,6 +226,8 @@ wireless.foreach_radio(uci, function(radio, index, config)
 			configure_mesh_wireless(radio, index, config)
 		end
 	end
+
+	uci:set('wireless', radio_name, 'beacon_int', beacon_interval)
 
 	fixup_wan(radio, index)
 end)
--- a/package/gluon-core/check_site.lua
+++ b/package/gluon-core/check_site.lua
@@ -32,6 +32,7 @@ local supported_rates = {6000, 9000, 12000, 18000, 24000, 36000, 48000, 54000}
 for _, config in ipairs({'wifi24', 'wifi5'}) do
 	if need_table({config}, nil, false) then
 		need_string(in_site({'regdom'})) -- regdom is only required when wifi24 or wifi5 is configured
+		need_number({config, 'beacon_interval'}, false)
 
 		if config == "wifi24" then
 			local channels = {1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14}


rotanid:
thanks, are there also downsides besides the mentioned advantage/potential?

blocktrron:
@rotanid devices at the edge of the coverage zone might not display the network immediately / at all if there are a lot of packets lost.

Whether or not it produces issues in close proximity depends on the onfigured value. I was testing it with ~ 1 beacon per second and did not experience issues at all.

NeoRaider:
Is this something a community can reasonable decide, or should we just hardcode values in Gluon?

mweinelt:
Pretty sure this should default to a well-tested and understood value. No point in trying to save airtime if it results in clients having a hard time seeing you.

blocktrron:
@NeoRaider From my POV, this is a setting we can expose without issues. Values up to 1000 ms should be fine for most use cases.

@mweinelt We already use the preset from OpenWrt (which is 100ms). I think exposing this setting brings a huge benefit for "Event" domains (as some communities have). When there are many devices broadcasting beacons, the difference becomes less noticeable even at larger intervals.

Remember, it's only big of a deal if there's already a large amount loss between your end-device and the access point. The access point will still send a beacon frame in case your device sends a probe request.

mweinelt:
I don't seen an issue with exposing, I'm just not sure I have an immediate use case while lacking experience with regard to this setting.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -225,6 +226,13 @@ util.foreach_radio(uci, function(radio, index, config)
 		end
 	end
 
+
+	if beacon_interval then
+		uci:set('wireless', radio_name, 'beacon_int', beacon_interval)
+	else
+		uci:delete('wireless', radio_name, 'beacon_int')
+	end
```
NeoRaider, line 234: The if/else can be removed - unlike the old *luci.model.uci*, *simple-uci*'s `uci:set()` will delete the option when `nil` is passed.

docs/multidomain-site-example/site.conf
```diff
@@ -8,12 +8,14 @@
...
 
   wifi24 = {
+    beacon_interval = 1000,
     mesh = {
       mcast_rate = 12000,
     },
   },
 
   wifi5 = {
+    beacon_interval = 500,
```
NeoRaider, line 18: - As long as we don't have any experience with this setting, it should not appear in our example configs, or only as a comment
- If we leave this as a comment, the non-multidomain example should be adjusted as well.

blocktrron: Removed this setting

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
blocktrron: This is good to know. I assume I've did this for a long time on multiple occasions.

Removed the redundant block.

docs/user/site.rst
```diff
@@ -119,6 +119,8 @@ regdom \: optional
 wifi24 \: optional
     WLAN configuration for 2.4 GHz devices.
     ``channel`` must be set to a valid wireless channel for your radio.
+    ``beacon_interval`` can be specified to set a custom beacon interval in
+    milliseconds. If this is not set, the default value of 100 ms is used.
```
mweinelt, line 123: Pretty sure this is wrong. It's not milliseconds, but time units. Let me check hostapd.sh.

NeoRaider: I just had a look at what LuCI does, and it has a very simple solution: The field is just called "Beacon Interval", without any explanation of units...

mweinelt, line 123:
```suggestion
    time units (TU). A time unit is equivalent to 1024 microseconds. If not set, the default value of 100 TU (=102.4 ms) is used.
```

docs/user/site.rst
```diff
@@ -119,6 +119,10 @@ regdom \: optional
 wifi24 \: optional
     WLAN configuration for 2.4 GHz devices.
     ``channel`` must be set to a valid wireless channel for your radio.
+    ``beacon_interval`` can be specified to set a custom beacon interval in
+    time units (TU). A time unit is equivalent to 1024 microseconds.
+    If not set, the default value of 100 TU (=102.4 ms) is used.
```
rotanid, line 124: @blocktrron  sorry i'm late, but writing ms as milliseconds would have been better to differentiate from the microseconds, no?

mweinelt: I'd rather replace microseconds with µs if we want to go there.

rotanid: hm, ok, both better than currently

mweinelt: Fixed in 97e5434b3267ec48953d16fb92de084da116f1c9.

CodeFetch:
@blocktrron I don't have a problem with the "event" use-case itself, but I think that it does not work this way and your statement is misleading. While there is a foreign transmission the AP won't send beacons either way because of CSMA/CA if it senses this transmission. Don't suggest to try to fix a bad network setup with that. It won't reduce the majority of collisions in the hidden node case and if there is no hidden node, the mentioned problem does not exists except the network setup was planned very very badly  (e.g. using too many APs at one place).

blocktrron:
Reducing the beacon rate allows more time to to be used for client traffic. I did not claim it reduces the collision probability on the shared medium.

The point with the event domain was about reducing the regular beacon interval as clients will perform periodic probe requests nevertheless.

CodeFetch:
@blocktrron I understood your intention, but as far as I know it's a myth. A beacon ordinarily fits into less than 4 TUs (which really is the worst case). As we have 2 SSIDs [I assumed](https://github.com/freifunk-gluon/gluon/issues/1178#issuecomment-345446388) a maximum of 10% bandwidth consumption.  As we have set 54 Mbps as a minimum rate now, the time decreases drastically. Here's someone who actually measured it: http://wifinigel.blogspot.com/2013/08/its-well-known-rule-of-thumb-when.html 

For getting the air clear buffered packets should be delivered as fast as possible and power-saving stations will have a higher probability to request them timely when you have a standard beacon interval. Otherwise you'll have all power-saving stations contend for PS-polling in the worst case causing a higher collision probability.

blocktrron:
I feel this is becoming arguing about a solution which clearly does not fit all (hence it's optional).

The traffic indication map is often sent with beacons directed to the STA, thus the collision probability is reduced (and the beacon can be transmitted at a much higher rate). How this is handled is not consistent, as fullmac drivers implement their own handling. Possible timeouts are still determined using the beacon interval.

So configuring the beacon interval might or might not give you a benefit here, ultimately it depends on your environment.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -226,6 +227,8 @@ wireless.foreach_radio(uci, function(radio, index, config)
 		end
 	end
 
+	uci:set('wireless', radio_name, 'beacon_int', beacon_interval)
```
rotanid, line 230: what happens here if beacon_interval is nil because it's not defined in the site?

rotanid: ok, it seems if "nil" is supplied, uci:set does simply delete the setting from the config.

PR #2030, opened 2020-05-23
docs: enhance OWE documentation

rubo77:
This adds a better hint how to configure an OWE network in the example site.conf and adds a hint to the `wireless-encryption-wpa3` package in the documentation

### DIFF
--- a/docs/site-example/site.conf
+++ b/docs/site-example/site.conf
@@ -47,9 +47,21 @@
       -- ssid = 'alpha-centauri.freifunk.net', (optional - SSID for open client network)
       -- disabled = true, -- (optional)
 
-      -- Configuration for a non-backward compatible OWE network below.
-      -- owe_ssid = 'owe.alpha-centauri.freifunk.net', -- (optional - SSID for OWE client network)
-      -- owe_transition_mode = true, -- (optional - enables transition-mode - requires ssid as well as owe_ssid)
+      -- `owe_ssid` sets the SSID for OWE client network and
+      -- `owe_transition_mode` enables transition-mode, both settings are
+      -- optional, the transition-mode requires `ssid` as well as `owe_ssid` set.
+      -- Note: some devices cannot connect with transition mode enabled
+      -- There are two backward compatible configurations possible: 
+      
+      -- 1. a (mostly) backward compatible OWE network, where those clients
+      --    that are able automatically connect to the OWE SSID:
+      -- owe_ssid = 'owe.alpha-centauri.freifunk.net', 
+      -- owe_transition_mode = true,
+
+      -- 2. a securely backward compatible OWE network, where you have to
+      --    manually choose the SSID you want to connnect to:
+      -- owe_ssid = 'owe.alpha-centauri.freifunk.net',
+      -- owe_transition_mode = false, -- (default)
     },
 
     mesh = {
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -130,15 +130,18 @@ wifi24 \: optional
     This will only affect new installations.
     Upgrades will not change the disabled state.
 
-    ``ap`` holds the client network configuration.
-    To create an unencrypted client network, a string named ``ssid`` which sets the
-    interface's ESSID is required. This is the wireless network clients connect to.
-    For an OWE secured network, the ``owe_ssid`` string has to be set. It sets the
-    SSID for the opportunistically encrypted wireless network, to which compatible
-    clients can connect to.
-    To utilize the OWE transition mode, ``owe_transition_mode`` has to be set to true.
-    Note that for the transition mode to work, both ``ssid`` as well as ``owe_ssid``
-    have to be enabled.
+    ``ap`` holds the client network configuration. To create an unencrypted
+    client network, a string named ``ssid`` which sets the interface's ESSID is
+    required. This is the wireless network clients connect to. For an OWE
+    secured network, the ``owe_ssid`` string has to be set which sets the SSID
+    for the opportunistically encrypted wireless network (to which compatible
+    clients can connect to) and the ``wireless-encryption-wpa3`` feature has to
+    be added to ``GLUON_FEATURES_standard`` in ``site.mk`` (which will enable it
+    only for non-tiny devices). To utilize the OWE transition mode,
+    ``owe_transition_mode`` has to be set to true. Note that for the transition
+    mode to work, both ``ssid`` as well as ``owe_ssid`` have to be enabled. Also
+    note that some devices get problems connecting, when transition mode is
+    enabled.
 
     ``mesh`` requires a single parameter, a string, named ``id`` which sets the
     mesh id, also visible as an open WiFi in some network managers. Usually you


docs/site-example/site.conf
```diff
@@ -68,6 +68,10 @@
     outdoor_chanlist = '100-140',
     ap = {
       ssid = 'alpha-centauri.freifunk.net',
+      
+      -- Configuration for a non-backward compatible OWE network below.
+      -- owe_ssid = 'owe.alpha-centauri.freifunk.net', -- (optional - SSID for OWE client network)
+      -- owe_transition_mode = false, -- (optional - false disables transition-mode - otherwise the ssid as well as owe_ssid are required)
```
blocktrron, line 74: This was left out intentionally. See the comment above the `wifi5` block

docs/user/site.rst
```diff
@@ -133,12 +133,14 @@ wifi24 \: optional
...
-    Note that for the transition mode to work, both ``ssid`` as well as ``owe_ssid``
-    have to be enabled.
+    For an OWE secured network, the ``owe_ssid`` string has to be set which sets the
+    SSID for the opportunistically encrypted wireless network (to which compatible
+    clients can connect to) and the ``wireless-encryption-wpa3`` feature has to be
+    added to ``GLUON_FEATURES_standard`` in ``site.mk`` (which will enable it only
+    for non-tiny devices).
+    To disable the OWE transition mode, ``owe_transition_mode`` has to be set to
+    ``false``. Note that for the transition mode to work, both ``ssid`` as well
+    as ``owe_ssid`` have to be enabled.
```
blocktrron, line 143: Why is this rewrite necessary? The default value (if absent) for the transition mode is false anyways, while this rewrite implies it's true by default.

Just add the required feature or explain why you feel this is necessary.

package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
```diff
@@ -84,7 +84,7 @@ local function configure_owe_transition_mode(config, radio_name)
 		return
 	end
 
-	if not ap() or not ap.owe_transition_mode() then
+	if not ap() or (ap.owe_transition_mode() ~= nil and not ap.owe_transition_mode()) then
```
blocktrron, line 87: Both nil as well as false result in a boolean false in this expression. So why you feel this is necessary?

docs/user/site.rst
rubo77: Yes, I changed the default to true

package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
rubo77: I wanted to change the default to true here. How would be the correct syntax?

rubo77:
I think it is important to change the default to true, because without transition mode all Android 9 Devices would not be able to connect any more at all.

mweinelt:
Transition mode is a fickle business right now, as only very few modern devices actually support that. So the general recommendation is to leave it off for now and just have a pure OWE SSID for testing purposes. Also remember to name it differently, since it will require a non-compatible WLAN profile on e.g. Android.

blocktrron:
@rubo77 Your claim about Android 9 is not true. enabling transition mode breaks Android 9 connectivity, while disabled transition mode with separate SSIDs for OWE & Open is the safest bet for now. Or with which device did you make this discovery?

package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
blocktrron: and --> or

NeoRaider: `not ap() or not ap.owe_transition_mode()` is already unnecessarily verbose. For default `false` use:
```lua
if not ap.owe_transition_mode(false) then
```
To make the default true, just change that to:
```lua
if not ap.owe_transition_mode(true) then
```
Our `site` library allows arbitrary indexing into non-existing site.conf keys - `ap.this.key.is.nonsense('foo')` would just return `'foo'` without throwing an error.

Unrelated to this PR, the same simplification is possible few other places in this file.

docs/user/site.rst
rubo77: I reverted the default change

package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
rubo77: OK, I will do an extra PR for that. and I reverted the default change

rubo77:
> Please give a detailed rationale about the intent of your change

Sorry, I misunderstood your explanation in the [Forum](https://forum.freifunk.net/t/firmware-bauen-mit-wpa3-owe-support/22051/4?u=rubo77).
I reverted the default change and this PR is only about the documentation now.

package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
rubo77: #2031

blocktrron:
@rubo77 The transition mode is (by design) backwards compatible. So a non-backwards compatible network would consist of

 - No open SSID
 - A OWE SSID
 - Transition Mode set to false

Yes, some (!) Android 9 devices will have a hard time connecting, however this is due to a bug on the User device side, not a design decision.

So the site-example actually describes `Configuration for a backward compatible OWE network below.` instead of `Configuration for a non-backward compatible OWE network below.`.

rubo77:
So the "non-backward compatible" shouldn't refer to those Android 9 devices?

I think it is an essential problem, that we shouldn't ignore, since probably 50% of all phones are on Android 9 right now.
This might change in the future, then we can rephrase the docs.

blocktrron:
@rubo77 The breakage with some (!) Android 9 devices is not a design fault and (to my knowledge) no other implementation is flawed this way. It's also dependent on the WiFi firmware, so a lot of variables leading to breakage there.

However, the OWE transition method is there to allow OWE functionality without leaving non-supporting devices in the dust. You imply with your proposed change the OWE transition mode creates a non-backwards compatible network which is simply not true.

We can mention the issue separately, but calling the transition-mode non-backwards compatible due to some client vendors having broken implementation is not helpful at all. Explain the problems one might face instead of making false claims about the design of the transition mode.

rubo77:
OK, so I added the two possible backward configurations as example now:

```
-- 1. a (mostly) backward compatible OWE network, where those clients that are able automatically connect to the OWE SSID:
-- owe_ssid = 'owe.alpha-centauri.freifunk.net', 
-- owe_transition_mode = true, -- note: some devices cannot connect with transition mode enabled

-- 2. a securely backward compatible OWE network, where you have to manually choose the SSID you want to connnect to:
-- owe_ssid = 'owe.alpha-centauri.freifunk.net',
-- owe_transition_mode = false, -- (default)
```
I hope, this is correct now

docs/user/site.rst
```diff
@@ -130,24 +130,27 @@ wifi24 \: optional
...
     ``mesh`` requires a single parameter, a string, named ``id`` which sets the
     mesh id, also visible as an open WiFi in some network managers. Usually you
     don't want users to connect to this mesh-SSID, so use a cryptic id that no
     one will accidentally mistake for the client WiFi.
 
     ``mesh`` also accepts an optional ``mcast_rate`` (kbit/s) parameter for
-    setting the multicast bitrate. Increasing the default value of 1000 to something
-    like 12000 is recommended.
+    setting the multicast bitrate. Increasing the default value of 1000 to
+    something like 12000 is recommended.
```
blocktrron, line 153: This cahnge is not related to OWE?

docs/site-example/site.conf
```diff
@@ -47,9 +47,16 @@
       -- ssid = 'alpha-centauri.freifunk.net', (optional - SSID for open client network)
       -- disabled = true, -- (optional)
 
-      -- Configuration for a non-backward compatible OWE network below.
-      -- owe_ssid = 'owe.alpha-centauri.freifunk.net', -- (optional - SSID for OWE client network)
-      -- owe_transition_mode = true, -- (optional - enables transition-mode - requires ssid as well as owe_ssid)
+			-- `owe_ssid` sets the SSID for OWE client network and `owe_transition_mode` enables transition-mode, both settings are optional, the transition-mode requires ssid as well as owe_ssid set.
```
blocktrron, line 50: Wrap this line

docs/site-example/site.conf
```diff
@@ -47,9 +47,16 @@
...
+			-- There are two backward compatible configurations possible:
+			
+			-- 1. a (mostly) backward compatible OWE network, where those clients that are able automatically connect to the OWE SSID:
```
blocktrron, line 53: Wrap this line

docs/site-example/site.conf
```diff
@@ -47,9 +47,16 @@
...
+			-- owe_ssid = 'owe.alpha-centauri.freifunk.net', 
+			-- owe_transition_mode = true, -- note: some devices cannot connect with transition mode enabled
```
blocktrron, line 55: Move the explanation above

docs/user/site.rst
rubo77: no, it is just a correction of the max line length my editor suggested. should I remove it?

blocktrron: yes

docs/site-example/site.conf
```diff
@@ -47,9 +47,21 @@
...
+			
+			-- 1. a (mostly) backward compatible OWE network, where those clients
+			--    that are able automatically connect to the OWE SSID:
+			-- owe_ssid = 'owe.alpha-centauri.freifunk.net', 
+			-- owe_transition_mode = true,
+
+			-- 2. a securely backward compatible OWE network, where you have to
+			--    manually choose the SSID you want to connnect to:
+			-- owe_ssid = 'owe.alpha-centauri.freifunk.net',
+			-- owe_transition_mode = false, -- (default)
```
blocktrron, line 64: indentation

blocktrron:
Merged with changed. Thanks!

PR #2031, opened 2020-05-24
gluon-client-bridge: simplify boolean expressions

rubo77:
This will remove some unnecessary conditions, and save some characters in an encapsulated and-statement for a consecutive or-statements

### DIFF
--- a/package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
+++ b/package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
@@ -25,7 +25,7 @@ local function configure_ap(radio, index, config, radio_name)
 
 	local macaddr = util.get_wlan_mac(uci, radio, index, 1)
 
-	if not ap() or not ap.ssid() or not macaddr then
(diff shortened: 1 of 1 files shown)


package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
```diff
@@ -25,7 +25,7 @@ local function configure_ap(radio, index, config, radio_name)
 
 	local macaddr = util.get_wlan_mac(uci, radio, index, 1)
 
-	if not ap() or not ap.ssid() or not macaddr then
+	if not ap.ssid(false) or not macaddr then
```
NeoRaider, line 28: Remove the `false` here - site accessors will default to `nil`, which has the same effect when used in a condition.

As `ssid` isn't a boolean, `false` isn't a semantically meaningful value. Generally, only pass `false` for booleans, and leave out the default value for other types when you just want to check whether the option exists (like it was done for `ap()` before).

NeoRaider: (for `owe_ssid` as well)

PR #2034, opened 2020-05-25
gluon-core: move wireless-related utils to gluon-wireless

blocktrron:
Seperate wireless-related helper methods from the util package into a
new wireless package. This keeps them seperated, as the amount of
wireles helpers increased in the past, justifying a seperate package.


rotanid:
commit message
 three times "seperated" --> separated
 wireles --> wireless

Issue #2032, opened 2020-05-25
significantly increased memory usage after upgrade to Gluon v2020.1.x


rotanid:
After updating from v2019.1.2 to v2020.1.2 memory usage increased.
The increase in our network is from average ~60% to ~70% on an idle node and at least 5% more on an active node. (with 32M memory)
On devices with only 32M memory, this leads to frequent reboots.
On devices with more memory, the increase can mostly also be seen but is not problematic.

Our firmware for these devices did not change in other ways with this Gluon version bump which could affect the memory usage.

please find attached the memory graphs of 4 32M devices and 2 64M devices before and after the firmware upgrade.

![memory_v202012_18a6f7c5cdb6_841v11](https://user-images.githubusercontent.com/654617/82846449-3612ed00-9ee9-11ea-8bde-59b86797d31a.png)
![memory_v202012_18a6f7c5c3bc_841v11](https://user-images.githubusercontent.com/654617/82846452-37441a00-9ee9-11ea-99a5-ebefb10a9e55.png)
![memory_v202012_18a6f7c5aaec_841v11](https://user-images.githubusercontent.com/654617/82846455-37441a00-9ee9-11ea-9e54-c022370cbd5a.png)
![memory_v202012_60e3275773ac_841v9](https://user-images.githubusercontent.com/654617/82846456-37dcb080-9ee9-11ea-8e92-f3a7cd5ad9a7.png)
![memory_v202012_ec086ba4d066_1043v3](https://user-images.githubusercontent.com/654617/82846451-36ab8380-9ee9-11ea-87f4-22c17e17d294.png)
![memory_v202012_18d6c75152c2_1043v4](https://user-images.githubusercontent.com/654617/82846453-37441a00-9ee9-11ea-8b07-0883873a7110.png)

as far as i know, others also observed the issues as reported on the Gluon IRC channel for weeks.

rotanid:
i can confirm that #2036 fixes the reported issue, 9 of our 10 test devices have an uptime of around 2 days by now.

PR #2036, opened 2020-05-27
Fix increased RAM usage since OpenWrt 19.07

NeoRaider:
The squashfs block size patch has the most significant effect (~2.6MB), the other two patches are only included for completeness - they presumably save around 10KB.

Fixes: #2032

While implementing this, I noticed an interesting effect in our config generation: the `CONFIG_TARGET_SQUASHFS_BLOCK_SIZE=64` in ar71xx-generic is only able to override `CONFIG_TARGET_SQUASHFS_BLOCK_SIZE=256` from the generic config because the string `64` is sorted after `256`. I have implemented a fix for this as well and will submit it as a separate PR when this is merged.


belzebub40k:
I just upgrade a wr841v11 with this patch and the memory usage has gone down by ~10% compared to the previous reboot.

rotanid:
@blocktrron it would have been nice to wait for feedback from the author of the issue before merging.
also, i don't see any reason why this could not have waited a while longer no matter my observation, why the rush?

blocktrron:
@rotanid As explained on IRC, the squashfs change is obvious when looking at the 18.06 <--> 19.07 diff. As multiple people (including myself) approved it reduces the memory load, we are good to got with that from my POV.

PR #2037, opened 2020-05-28
Config generation fixes

NeoRaider:
As noted in #2036, precedence of different `.config` settings in generic and target-specific config is resolved in a quite surprising way: Rather than using the value that is set last, the "greatest" value when sorted as strings is used. This PR fixes the behaviour.

See also the commit message of the first patch for further details.

With all current target configurations, the generated .config is unchanged with this PR.


NeoRaider:
I have pushed 2 more commits to this branch: one minor cleanup, and one that will be used to provide VMDK and VDI images for x86 in `next`.

rotanid:
this needs to be backported to v2020.2.x in order to be able to backport #2051

NeoRaider:
@rotanid This PR was merged before v2020.2.x was branched.

rotanid:
@NeoRaider d'oh, sorry.

PR #2041, opened 2020-05-30
Add all mesh interfaces to firewall rule wired_mesh

### LINKED ISSUE #1479: Multiple mesh_lan interfaces. Anything named different than mesh_lan won't be able to mesh
I am in situation where lan mesh needs to be provided into multiple seperate L2 networks like legacy did.
We use the new vxlan based gluon_wired proto already.
I therefore defined in the /etc/config/network

```
config interface 'mesh_lan'        
        option ifname 'eth0.97'    
        option mesh 'bat0'         
        option transitive '1'      
        option index '4'           
        option proto 'gluon_wired' 
        option macaddr '02:19:99:67:8a:fa'
        option disabled '0'               
                                          
config interface 'mesh_test'              
        option ifname 'eth0.98'           
        option mesh 'bat0'                
        option transitive '1'             
        option index '4'                  
        option proto 'gluon_wired'        
        option macaddr '02:19:99:67:8a:fb'
        option disabled '0'
```

eth0.97 and eth0.98 are my vlans in wich batman should be available in. This should not be a bridge because in this seperate vlans are different dhcp instances running. This creates vx_mesh_lan and vx_mesh_test but mesh_test is not meshing at all with the nodes behind eth0.98.

```
~# ip link
...
14: mesh-vpn: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1312 qdisc tbf master bat0 state UNKNOWN qlen 1000
    link/ether 2a:da:d6:07:f3:ef brd ff:ff:ff:ff:ff:ff
15: vx_mesh_test: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1430 qdisc noqueue master bat0 state UNKNOWN qlen 1000
    link/ether 2a:da:d6:07:f3:ec brd ff:ff:ff:ff:ff:ff
16: vx_mesh_lan: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1430 qdisc noqueue master bat0 state UNKNOWN qlen 1000
    link/ether 2a:da:d6:07:f3:ec brd ff:ff:ff:ff:ff:ff
```

To confirm eth0.98 has connectivity to the nodes I swaped eth0.97 and eth0.98. In this case only nodes behind eth0.98 appear in the mesh. If I rename mesh_lan to say for instance mesh_first no mesh on ether interface would work.

I also tested a single mesh_x interface. Anything named different than mesh_lan does not mesh at all.

We are running v2018.1

Sprinterfreak:
This allows for multiple mesh legs out of one box. Useful for offloaders
and networks using vlan for separate mesh transport.

This fixes #1479

### DIFF
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/140-firewall-rules
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/140-firewall-rules
@@ -80,9 +80,18 @@ for _, zone in ipairs({ 'wan', 'local_client', 'mesh' }) do
 end
 
 
--- We can't put mesh_wan into this zone, as mesh_wan is the same
--- interface as wan, which has its own zone
-uci:set('firewall', 'wired_mesh', 'network', {'mesh_lan'})
+local wired_mesh_ifaces = {}
+uci:foreach('network', 'interface',
+	function(iface)
+		-- Select all interfaces with proto gluon_wired except for
+		-- mesh_wan into this zone, as mesh_wan is the same
+		-- interface as wan, which has its own zone
+		if iface['proto'] == 'gluon_wired' and iface['.name'] ~= 'mesh_wan' then
+			table.insert(wired_mesh_ifaces, iface['.name'])
+		end
+	end
+)
+uci:set('firewall', 'wired_mesh', 'network', wired_mesh_ifaces)
 
 -- VXLAN for wired meshing
 for _, zone in ipairs({'wired_mesh', 'wan'}) do


package/gluon-core/luasrc/lib/gluon/upgrade/140-firewall-rules
```diff
@@ -81,8 +81,17 @@ end
 
 
 -- We can't put mesh_wan into this zone, as mesh_wan is the same
--- interface as wan, which has its own zone
-uci:set('firewall', 'wired_mesh', 'network', {'mesh_lan'})
+-- interface as wan, which has its own zone but allow any other defined
+-- gluon_wired interface addicted to bat0 to communicate
+mil = {}
```
NeoRaider, line 86: - Missing `local` (if you have `lualint` installed, you can `make lint-lua` to find such issues)
- Let's name this `wired_mesh_ifaces`

package/gluon-core/luasrc/lib/gluon/upgrade/140-firewall-rules
```diff
@@ -81,8 +81,17 @@ end
...
+uci:foreach('network', 'interface',
+	function(i)
```
NeoRaider, line 88: `i` -> `iface`

package/gluon-core/luasrc/lib/gluon/upgrade/140-firewall-rules
```diff
@@ -81,8 +81,17 @@ end
...
+		if i['proto'] == 'gluon_wired' and i['.name'] ~= 'mesh_wan' then
```
NeoRaider, line 89: This could use a comment explaining that `mesh_wan` uses the same underlying interface as `wan` and is therefore not assigned to a firewall zone itself.

Sprinterfreak: The comment is above the block and remained from the previous version. Maybe we could move it directly to the line where the actual filtering happens

NeoRaider: Sounds good

rotanid:
this obviously needs to be merged into one commit and it needs a decent commit message ;-)

NeoRaider:
Thanks - I slightly modified the commit message when merging.

Issue #1783, opened 2019-07-06
respondd doesn't determine interface type of L2TP and Wireguard as "tunnel"


christf:
After a discussion in #wireguard, it seems like respondd is not correctly determining whether a  mesh interfaces is a tunnel or not. When considering the pointtopoint flag on each interface, it would consider wireguard interfaces as well.

 I cannot think of any gotchas, what are your thoughts?

NeoRaider:
It seems preferable to me to actually check the interface type string using netlink.

Currently we check whether an interface is a TUN/TAP interface or not; I assume this will not work for L2TP either.

christf:
I am assuming you are referring to the unsighned short ifi_type in struct ifinfomsg? For this I have not been able to find a definition of its contents.

According to the docs there is also a rtattr attached called IFLA_LINK that contains the "link type". The kernel header says:
```
  /* IFLA_LINK.
     For usual devices it is equal ifi_index.
     If it is a "virtual interface" (f.e. tunnel), ifi_link
     can point to real physical interface (f.e. for bandwidth calculations),
     or maybe 0, what means, that real media is unknown (usual
     for IPIP tunnels, when route to endpoint is allowed to change)
   */
``` 
That would require to figure out whether the ifindex is equal to the IFLA_LINK value and if it isn'T then deduce that the current interface must be a tunnel of some sort. Imho this seems to be our best shot.

NeoRaider:
@christf  I was refering to the `IFLA_INFO_KIND` field (nested in `IFLA_LINKINFO`), which contains the link kind as a string (like "vxlan").

NeoRaider:
You can also look at the `iproute2` code. The kind is the first word in the third line printed by `ip -d link` for virtual interfaces.

NeoRaider:
Hmm, this issue is much more cumbersome than I anticipated. There are way too many ways to ask the kernel what "type" an interface is, and every driver implements a different subset:

- `ifi_itype` just tells us whether an interface uses Ethernet headers or something else (so it has the same value for all interface types we support for meshing). The same information can be found in `/sys/class/net/*/type`
- IFLA_LINK is unset for most interface types (but there is `/sys/class/net/*/iflink`, which contains this information, but it is just set to the ifindex for tunnel interfaces, and also not netns-aware; all in all, useless)
[...]

PR #2043, opened 2020-06-01
Fix tunnel interface detection in respondd nodeinfo

NeoRaider:
As noted in #1783, Wireguard and L2TP interfaces are not correctly detected as tunnels in our current respondd nodeinfo providers. To fix this, use the DEVTYPE value of the interface `uevent` as an additional data source. I did not check my code in an actual tunneldigger or wireguard setup (as I don't have access to one), but I checked the DEVTYPE of l2tpeth interfaces created using `ip l2tp`, and looked at the wireguard kernel code.

To avoid code duplication between the batadv and babel providers, this function is implemented in libgluonutil. A helper to get the base interface of VLAN/VXLAN/... interfaces  is added as well.

The first patch of this series is some cleanup of the link-local address function in the mesh-babel code. The mesh-babel changes are compile-tested only - a test in an actual Babel setup would be highly appreciated (but I guess I can try to build something myself if necessary).

### DIFF
--- a/package/gluon-mesh-babel/src/respondd.c
+++ b/package/gluon-mesh-babel/src/respondd.c
@@ -62,38 +62,40 @@
 
 static struct babelhelper_ctx bhelper_ctx = {};
 
-static void obtain_if_addr(const char *ifname, char *lladdr) {
+static bool get_linklocal_address(const char *ifname, char lladdr[INET6_ADDRSTRLEN]) {
 	struct ifaddrs *ifaddr, *ifa;
-	int family, n;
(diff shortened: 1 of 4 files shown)


NeoRaider:
I gave this PR a quick test run on an isolated gluon-mesh-babel node as well now.

NeoRaider:
Should we just merge this, as it seems to work as expected?

rotanid:
> Should we just merge this, as it seems to work as expected?

lets's do it ;)

PR #2044, opened 2020-06-01
actions: build documentation

blocktrron:
This adds a new workflow for building the Gluon documentation.

In contrast to the existing build workflow, the documentation
is built for every branch, as it only takes around two minutes
to render the HTML documentation.

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -1,0 +1,20 @@
+name: Build Documentation
+on:
+  push:
+  pull_request:
+    types: [opened, synchronize, reopened]
+jobs:
+  build-documentation:
+    name: docs
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Install Dependencies
+        run: sudo pip3 install sphinx-rtd-theme
+      - name: Build documentation
+        run: make -C docs html
+      - name: Archive build output
+        uses: actions/upload-artifact@v1
+        with:
+          name: docs_output
+          path: docs/_build/html


mweinelt:
> In contrast to the existing build workflow, the documentation
> is built for every branch

Wait what? So GHA will build the documentation 20 times per Commit?

blocktrron:
Images are only built for release branches, next and master. The documentation is built regardless of the branches name.

mweinelt:
Ok, silly me. It was late apparently.

mweinelt:
Ok, now that I reread it … it's clear. Not sure what I thought.

The artifact would be really helpful if it was published to e.g. netlify.

https://github.com/netlify/actions

blocktrron:
Hmm netlify seems to have costs attached to it (or do they have a special OSS plan)?

mweinelt:
> Hmm netlify seems to have costs attached to it (or do they have a special OSS plan)?

Hm, the free plan is for individuals, *sigh*. I was sure ffrgb/meshviewer was using that, so I thought it might be an option. 

Maybe @xf- has an idea what might be other (similiar?) options.

NeoRaider:
Hmm, I'm not sure if we really need to deploy the docs for every PR - knowing that the docs build without warnings is a big improvement already, and for master and all release branches we have RTD.

mweinelt:
For most people writing Restructured Text is uncommon, hence the idea to deploy it so it can be browsed and visually inspected.

I don't think we can match the diff stat to wait for changes to `/docs/**`,so it's probably this way or not at all right now.

PR #2045, opened 2020-06-02
Rewrite features.sh in Lua

NeoRaider:
This was left as a TODO in #2007.

The `features` file is converted to a Lua-based DSL.
    
A helper function `_` is used in the DSL; this will return the original string for enabled features, and nil for disabled features. This allows to use boolean operations on features without making the code too verbose.
    
Besides having more readable and robust code, this also fixes the bug that all files `packages/*/features` were evaluated instead of only using the feature definitions of currently active feeds.


package/features
```diff
@@ -1,37 +1,67 @@
-nodefault 'web-wizard'
+-- The _ prefix allows to perform boolean logic on feature names
+--
+-- Do not use the 'nodefault' flag in feature specifications that  have
+-- such a boolean match expression rather than a single feature name
```
blocktrron, line 4: I think a pointer to the documentation would be great here. Without context, the nodefault flag is not self-explanatory.

NeoRaider: Ah, I forgot we already have a doc page about the feature spec format, I'll have to update that as well.

PR #2050, opened 2020-06-07
lantiq-xrx200: add support for AVM Fritz!Box 3370

dzzinstant:
Based on Dark4MD's patch (https://github.com/freifunk-gluon/gluon/commit/2360924e4a43754b5a3c24e5797d94c6a9451a7f).
Further changes, might need discussion: 
1. Long press (>3s) the power button to get to setup mode. There is the problem of incidentally powering off the device by releasing the button too soon. But this shouldn't cause real damage (except for the loss of time), because the operator was planning to shut down the device anyway.
And the opposite problem of incidentally putting the device into setup mode, when the intention was to switch it off.
2. Patch to show LAN activity on LED, but IMHO that's not an important feature.
3. Don't know the reason for the ordering of LAN ports, so didn't change anything there (https://github.com/dzzinstant/gluon/issues/4), probably not relevant for gluon?

The flashing process is a bit more involved than for other AVM devices, see hints in follow-up to this message.

### Device Integration Checklist
  - [ ] webinterface
  - [ ] tftp
  - [x] other: ftp [EVA bootloader, separate images for kernel and filesystem partitions] -> https://github.com/dzzinstant/gluon/issues/3
  ```
  :~# lua -e 'print(require("platform_info").get_image_name())'
  avm-fritz-box-3370-rev-2-hynix-nand
  ```
- [ ] reset/wps/phone button must return device into config mode [no such button => use 'power' button instead] -> https://github.com/dzzinstant/gluon/commit/7ba5577eac5146fc8df3b669cd3c10a96e0ad0a1
- wired network
  - [ ] must have correct port assignment (WAN/LAN) [numbering of LAN ports is mixed up, does that matter?] -> https://github.com/dzzinstant/gluon/issues/4
- wifi (if applicable)
- led mapping
  - radio leds
    - [x] should show activity
  - switchport leds
    - [ ] should show link state and activity [shows only state, maybe apply patch to also show activity] -> https://github.com/dzzinstant/gluon/commit/84fe9fc1a0d9642b84f7bda9077f4b8038a1dfc4
- outdoor devices only
  ( ) added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`

**Additional / device-specific**
- [x] wan-radio/private wlan [can be enabled in parallel with client + mesh wlan]
- [ ] 'WLAN' button enables/disables radio -> https://github.com/openwrt/openwrt/pull/3042
- [ ] also works in the 5 Ghz band [3370 only has a single radio for both bands]
  - [x] association with AP must be possible
  - [x] association with 802.11s mesh must be working (only indoor mode)
  - [x] ap/mesh mode must work in parallel (only indoor mode)
  - [ ] outdoor mode with DFS [basically works, but some clients cause false radar detections]

Successfully tested on:
```
$ grep "^HW.*Revision\s\|^bootloader\|NAND device:" support\ FRITZ.Box\ WLAN\ 3370*.txt 
HWRevision	175
HWSubRevision	2
bootloaderVersion	1.1186
[    1.560000] NAND device: Manufacturer ID: 0x20, Chip ID: 0xdc (ST Micro NAND04GW3B2DN6), 512MiB, page size: 2048, OOB size: 64
```
(using the `..-hynix-nand-..` images; there was no automatic reboot after 30s)

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -270,6 +270,7 @@ lantiq-xrx200
 
 * AVM
 
+  - FRITZ!Box 3370 (Hynix Nand, Micron Nand) [#lan_as_wan]_
   - FRITZ!Box 7360 (v1, v2) [#avmflash]_ [#lan_as_wan]_
   - FRITZ!Box 7360 SL [#avmflash]_ [#lan_as_wan]_
   - FRITZ!Box 7362 SL [#eva_ramboot]_ [#lan_as_wan]_
--- a/package/gluon-config-mode-outdoor/luasrc/lib/gluon/config-mode/wizard/0250-outdoor.lua
+++ b/package/gluon-config-mode-outdoor/luasrc/lib/gluon/config-mode/wizard/0250-outdoor.lua
@@ -25,5 +25,5 @@ return function(form, uci)
 		end
 	end
 
-	return {'gluon', 'wireless'}
+	return {'gluon', 'network', 'wireless'}
 end
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
@@ -162,12 +162,13 @@ local function fixup_wan(radio, index)
 	uci:set('wireless', name, 'macaddr', macaddr)
 end
 
-local function configure_mesh_wireless(radio, index, config)
+local function configure_mesh_wireless(radio, index, config, disabled)
 	local radio_name = radio['.name']
 	local suffix = radio_name:match('^radio(%d+)$')
 
 	configure_mesh(config.mesh(), radio, index, suffix,
 		first_non_nil(
+			disabled,
 			is_disabled('mesh_' .. radio_name),
 			config.mesh.disabled(false)
 		)
@@ -215,7 +216,7 @@ wireless.foreach_radio(uci, function(radio, index, config)
 			util.add_to_set(hostapd_options, 'country3=0x4f')
 			uci:set_list('wireless', radio_name, 'hostapd_options', hostapd_options)
 
-			uci:delete('wireless', 'mesh_' .. radio_name)
+			configure_mesh_wireless(radio, index, config, true)
 		else
 			uci:delete('wireless', radio_name, 'channels')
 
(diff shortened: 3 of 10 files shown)


dzzinstant:
## Installing gluon on the Fritzbox 3370
Gluon can be installed using the EVA FTP method as described in [OpenWRT: AVM FRITZ!Box WLAN 3370](https://openwrt.org/toh/avm/fritz.box.wlan.3370#installation_lan_and_serial_optional). Just use your community's gluon images instead of the stock OpenWRT ones.
However, a few things should be kept in mind:

### Differences to other AVM boxes, possible difficulties
1. Device should be at least `HWSubRevision 2` (see first lines of file obtained through `http://fritz.box/support.lua`)
2. The kernel and filesystem reside on different flash partitions, so you have to flash two images:
`gluon-avm-fritz-box-3370-rev-2-..-eva-kernel.bin` -> `mtd1`
`gluon-avm-fritz-box-3370-rev-2-..-eva-filesystem.bin` -> `mtd0`
3. The device is available with different types of flash memory, you have to use the corresponding set of images:
  Micron flash chip -> use `..-micron-nand-..` set of images
  ST Micro or Hynix flash chip -> use `..-hynix-nand-..` set of images
  Unfortunately, it's not easy to determine which flash chip is installed, so:
   - While Fritz!OS is still installed, call `http://fritz.box/support.lua` in your web browser, save the support file to disk (it may take ~one minute to gather the data, be patient).
    The resulting file should contain a debug line starting with `NAND device:`, which identifies the flash chip.
   - Installing the wrong set of images seems to be non-fatal (the device will either bootloop or remain at the bootloader stage). If you picked the wrong ones, just use the EVA/ftp procedure again to install the other set of images.
4. Some versions of the bootloader automatically reboot after ~30 seconds. In this case be really fast or copy-paste the complete sequence of commands from the clipboard to the terminal.
### References
[OpenWRT: AVM FRITZ!Box WLAN 3370](https://openwrt.org/toh/avm/fritz.box.wlan.3370)
[Commit message: lantiq: complete AVM FRITZ!Box 3370 support](https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=0b62fe5ed87ecac52301096b15abb69f96117c8c)
[Michael Kuron: OpenWRT on AVM Fritz!Box 3370](https://blog.michael.kuron-germany.de/2018/12/openwrt-on-avm-fritzbox-3370/)

dzzinstant:
Merge has gone wrong. I'll close and resubmit this pull request.

NeoRaider:
You can just force-push your branch, no need to close it to fix something.

dzzinstant:
[Pointer] Couldn't reopen, therefore resubmitted as https://github.com/freifunk-gluon/gluon/pull/2056

PR #2051, opened 2020-06-10
build: target_config_lib: do not build unused packages for targets without opkg

NeoRaider:
Normally, we build all nonshared packages (which includes all kernel
modules) to generate an opkg feed for later package installations by
users. On targets without opkg, this just wastes time - disable it.

This also reduced the size of the base kernel by a lot, as interfaces only needed by optional modules can be removed as well - saving more than 64KB on ar71xx-tiny!

This is something I wanted for some time now, and with #2037 it's finally easily possible.
I propose we also backport this to v2020.2.x.

### DIFF
--- a/scripts/target_config_lib.lua
+++ b/scripts/target_config_lib.lua
@@ -139,6 +139,7 @@ lib.check_devices()
 if not lib.opkg then
 	lib.config('SIGNED_PACKAGES', false)
 	lib.config('CLEAN_IPKG', true)
+	lib.config('ALL_NONSHARED', false)
 	lib.packages {'-opkg'}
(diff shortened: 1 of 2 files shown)


mweinelt:
The actions artifact goes down from 264 to 211 MB. That's a win!

rotanid:
nice! to we want to backport #2037 and #2051 before doing the release of v2020.2 ?
i think this would be good, as it could be announced as part of the initial release notes as an improvement and could also serve as a reason to deprecate v2020.1.x early.

mweinelt:
We do, just let it rest on master for a few days.

PR #2052, opened 2020-06-10
mac80211: rt2800: enable MFP support unconditionally

blocktrron:
With this backported patch, rt2x00 will automatically fall back to software crypto in case MFP is used on the link.

Enable WPA3 features for the mt7620 target, as WPA3-SAE as well as OWE are working now.

This might also be possible for the rt305x subtarget, however I'm not in the possession of such a device.


blocktrron:
@rotanid Did have a short talk with neoraider on this subject and I'll merge this patch with a follow-up.

Issue #2053, opened 2020-06-11
5 GHz Mesh still disabled after disabling outdoor option on outdoor device


rotanid:
### Bug report

**What is the problem?**
In #2048 it was reported, that on an outdoor device with automatically enabled outdoor mode it does not work as expected if you disabled outdoor mode on the starting page of the config mode.
A necessary config section in /etc/config/network was missing, which was fixed in #2049 
Now the remaining issue is, that the section is there but the 5 GHz Wifi mesh interface stays disabled, you need to manually enable it.

**What is the expected behaviour?**
5 GHz Mesh Interface is enabled if Outdoor Mode gets disabled (or rather: was never enabled on a new device)

**Gluon Version:**
v2020.1.x commit e285ccd2e95d683c3e2b3703d6e2e6434c6d73ce
very likely also v2019.1.x and v2020.2.x and master branch affected

**Site Configuration:**
https://github.com/tecff/site-ffa/blob/experimental/site.conf

PR #2054, opened 2020-06-11
Enable 5GHz mesh interfaces when disabling outdoor mode, fix default status from site.conf

NeoRaider:
In both places where the outdoor mode can be disabled, we would leave the mesh interfaces disabled by default after #2049. This is a problem for the wizard in particular, as there is no way to enable the mesh interfaces at all without switching to the advanced settings.

While fixing this, I also noticed that we were ignoring the `disabled` site.conf flag for mesh interfaces since Gluon v2017.1... This is fixed as well.

Fixes: #2053

Completely untested so far, @rotanid please test! I'll give it a few test runs tomorrow.

### DIFF
--- a/package/gluon-config-mode-outdoor/luasrc/lib/gluon/config-mode/wizard/0250-outdoor.lua
+++ b/package/gluon-config-mode-outdoor/luasrc/lib/gluon/config-mode/wizard/0250-outdoor.lua
@@ -21,6 +21,24 @@ return function(form, uci)
 		if data ~= outdoor_mode then
 			uci:set('gluon', 'wireless', 'outdoor', data)
 			uci:save('gluon')
+
+			if data == false then
+				local mesh_ifaces_5ghz = {}
+				uci:foreach('wireless', 'wifi-device', function(config)
+					if config.hwmode ~= '11a' and config.hwmode ~= '11na' then
+						return
+					end
+
+					local radio_name = config['.name']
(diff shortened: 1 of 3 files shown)


rotanid:
i tested you PR/branch on a Unifi AC Mesh.
the issue #2053 is gone.
re-activating outdoor mode via config mode also results in what is expected (enabled outdoor mode, disabled 5ghz mesh interface)

i haven't tested your second fix regarding site.conf with disabled mesh interface

NeoRaider:
I have tested this on a CPE510 now and fixed two issues:
- Inverted logic in `gluon-web-wifi-config` (the default setting must be changed when outdoor mode is enabled, so it becomes visible when the user disables it)
- Invalid use of `uci:foreach` in `gluon-config-mode-outdoor` leading to segfaults (no sections may be added to or removed from a UCI config while iterating over it)

NeoRaider:
Okay, merged and backports to v2020.2.x, v2020.1.x and v2019.1.x.

Only the 3c2593b68436479d10cfa09def81b6f8b6e3546b is backported to v2019.1.x, as we didn't backport the prerequisites of d2052431dd4ab4c1c5e0ce406219e281df297715 (and having to reload the enable the mesh wifi in the advances settings was considered acceptable), and the issue fixed by 17bc6228929b38356d4209af64f807432daf713b should not occur in v2019.1.x, as we didn't backport b15b008e6951084531b69583076931be844fe075 either.

PR #2055, opened 2020-06-12
ipq40xx: essedma: Disable TCP segmentation offload for IPv6

ecsv:
It was noticed that the the whole MAC can hang when transferring data from one ar40xx port (WAN ports) to the CPU and from the CPU back to another ar40xx port (LAN ports).

Usually, the problem first starts with a simple data corruption:

  $ wget https://cdimage.debian.org/debian-cd/current/amd64/iso-cd/debian-10.4.0-amd64-netinst.iso -O /dev/null
  ...
  Connecting to saimei.ftp.acc.umu.se (saimei.ftp.acc.umu.se)|2001:6b0:19::138|:443... connected.
  ...
  Read  error at byte 48807936/352321536 (Decryption has failed.). Retrying.

But after a short while, the whole MAC will stop to react. No traffic can be transported anymore from the CPU port from/to the AR40xx PHY/switch and the MAC has to be resetted.

Disabling the TCP segmentation offload for IPv6 seems to prevent this hardware problem.

Forwarded: https://patchwork.ozlabs.org/project/openwrt/patch/20200609133504.6563-1-sven@narfation.org/

### DIFF
--- a/patches/openwrt/0012-ipq40xx-essedma-Disable-TCP-segmentation-offload-for-IPv6.patch
+++ b/patches/openwrt/0012-ipq40xx-essedma-Disable-TCP-segmentation-offload-for-IPv6.patch
@@ -1,0 +1,77 @@
+From: Sven Eckelmann <sven@narfation.org>
+Date: Tue, 9 Jun 2020 15:23:04 +0200
+Subject: ipq40xx: essedma: Disable TCP segmentation offload for IPv6
+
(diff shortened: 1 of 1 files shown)


mweinelt:
Thanks for preparing this, I agree with @blocktrron.

blocktrron:
Bumped OpenWrt. This will also go to v2020.2.

PR #2056, opened 2020-06-13
lantiq-xrx200: add support for AVM Fritz!Box 3370 [resubmitted]

dzzinstant:
(This is a resubmission of https://github.com/freifunk-gluon/gluon/pull/2050, couldn't reopen. Sorry for inconvenience.)
Based on Dark4MD's patch (https://github.com/freifunk-gluon/gluon/commit/2360924e4a43754b5a3c24e5797d94c6a9451a7f).
Further changes, might need discussion: 
1. Long press (>3s) the power button to get to setup mode. There is the problem of incidentally powering off the device by releasing the button too soon. But this shouldn't cause real damage (except for the loss of time), because the operator was planning to shut down the device anyway.
And the opposite problem of incidentally putting the device into setup mode, when the intention was to switch it off.
2. Patch to show LAN activity on LED, but IMHO that's not an important feature.
3. Don't know the reason for the ordering of LAN ports, so didn't change anything there (https://github.com/dzzinstant/gluon/issues/4), probably not relevant for gluon?

The flashing process is a bit more involved than for other AVM devices, see hints in follow-up to this message.

### Device Integration Checklist
  - [ ] webinterface
  - [ ] tftp
  - [x] other: ftp [EVA bootloader, separate images for kernel and filesystem partitions] -> https://github.com/dzzinstant/gluon/issues/3
  ```
  :~# lua -e 'print(require("platform_info").get_image_name())'
  avm-fritz-box-3370-rev-2-hynix-nand
  ```
- [ ] reset/wps/phone button must return device into config mode [no such button => use 'power' button instead] -> https://github.com/dzzinstant/gluon/commit/7ba5577eac5146fc8df3b669cd3c10a96e0ad0a1
- wired network
  - [ ] must have correct port assignment (WAN/LAN) [numbering of LAN ports is mixed up, does that matter?] -> https://github.com/dzzinstant/gluon/issues/4
- wifi (if applicable)
- led mapping
  - radio leds
    - [x] should show activity
  - switchport leds
    - [ ] should show link state and activity [shows only state, maybe apply patch to also show activity] -> https://github.com/dzzinstant/gluon/commit/84fe9fc1a0d9642b84f7bda9077f4b8038a1dfc4
- outdoor devices only
  ( ) added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`

**Additional / device-specific**
- [x] wan-radio/private wlan [can be enabled in parallel with client + mesh wlan]
- [ ] 'WLAN' button enables/disables radio -> https://github.com/openwrt/openwrt/pull/3042
- [ ] also works in the 5 Ghz band [3370 only has a single radio for both bands]
  - [x] association with AP must be possible
  - [x] association with 802.11s mesh must be working (only indoor mode)
  - [x] ap/mesh mode must work in parallel (only indoor mode)
  - [ ] outdoor mode with DFS [basically works, but some clients cause false radar detections]

Successfully tested on:
```
$ grep "^HW.*Revision\s\|^bootloader\|NAND device:" support\ FRITZ.Box\ WLAN\ 3370*.txt 
HWRevision	175
HWSubRevision	2
bootloaderVersion	1.1186
[    1.560000] NAND device: Manufacturer ID: 0x20, Chip ID: 0xdc (ST Micro NAND04GW3B2DN6), 512MiB, page size: 2048, OOB size: 64
```
(using the `..-hynix-nand-..` images; there was no automatic reboot after 30s)


dzzinstant:
## Installing gluon on the Fritzbox 3370
Gluon can be installed using the EVA FTP method as described in [OpenWRT: AVM FRITZ!Box WLAN 3370](https://openwrt.org/toh/avm/fritz.box.wlan.3370#installation_lan_and_serial_optional). Just use your community's gluon images instead of the stock OpenWRT ones.
However, a few things should be kept in mind:

### Differences to other AVM boxes, possible difficulties
1. Device should be at least `HWSubRevision 2` (see first lines of file obtained through `http://fritz.box/support.lua`)
2. The kernel and filesystem reside on different flash partitions, so you have to flash two images:
`gluon-avm-fritz-box-3370-rev-2-..-eva-kernel.bin` -> `mtd1`
`gluon-avm-fritz-box-3370-rev-2-..-eva-filesystem.bin` -> `mtd0`
3. The device is available with different types of flash memory, you have to use the corresponding set of images:
  Micron flash chip -> use `..-micron-nand-..` set of images
  ST Micro or Hynix flash chip -> use `..-hynix-nand-..` set of images
  Unfortunately, it's not easy to determine which flash chip is installed, so:
   - While Fritz!OS is still installed, call `http://fritz.box/support.lua` in your web browser, save the support file to disk (it may take ~one minute to gather the data, be patient).
    The resulting file should contain a debug line starting with `NAND device:`, which identifies the flash chip.
   - Installing the wrong set of images seems to be non-fatal (the device will either bootloop or remain at the bootloader stage). If you picked the wrong ones, just use the EVA/ftp procedure again to install the other set of images.
4. Some versions of the bootloader automatically reboot after ~30 seconds. In this case be really fast or copy-paste the complete sequence of commands from the clipboard to the terminal.
### References
[OpenWRT: AVM FRITZ!Box WLAN 3370](https://openwrt.org/toh/avm/fritz.box.wlan.3370)
[Commit message: lantiq: complete AVM FRITZ!Box 3370 support](https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=0b62fe5ed87ecac52301096b15abb69f96117c8c)
[Michael Kuron: OpenWRT on AVM Fritz!Box 3370](https://blog.michael.kuron-germany.de/2018/12/openwrt-on-avm-fritzbox-3370/)

Adorfer:
i see the missing features "clear type labeling" and "no other IO input apart from power" as "not solvable by software". 
we are not in an ideal world. x86 and raspi work fine without any button (last one even with freifunk-disfunctional wifi). And UBNT-NS2M has build a universe of incompatible mainboards, all labled as "Nanostation M2". 
In other word's is't not nice to live with those issues, but for those who can work around them, it would be nice to have direct gluon support.

blocktrron:
@Adorfer I don't understand what you want to say with your comment here.

After all, this device does not tick the boxes without the button patch. I don't like the idea of accessing setup-mode with a power-typed button as this might have unforeseen consequences. Also, the next person comes with a device only sporting a SD button and so forth.

If we drop the patch adding Power button setup-mode, I'm fine with supporting this as broken. But someone might have another opinion on this.

patches/openwrt/0010-lantiq-xrx200-show_LAN_activity_on_AVM_3370.patch
```diff
@@ -0,0 +1,24 @@
...
+@@ -55,7 +55,7 @@ netgear,dm200)
+ 	;;
+ avm,fritz3370-rev2-hynix|\
+ avm,fritz3370-rev2-micron)
+-	ucidef_set_led_switch "lan" "LAN" "fritz3370:green:lan" "switch0" "0x17"
++	ucidef_set_led_netdev "lan" "LAN" "fritz3370:green:lan" "eth0"
+ 	;;
+ zyxel,p-2812hnu-f1|\
+ zyxel,p-2812hnu-f3)
+
```
blocktrron, line 24: This patch can be dropped. We do not want to carry patches with us that are not upstreamable.

Also only the link-state is required, as activity is not always a given and depends on the hardware implementation of the switch.

dzzinstant:
> Patches for this device were already rejected due to the fact of the device not having the required version displayed on the outside.
The [OpenWRT device page](https://openwrt.org/toh/avm/fritz.box.wlan.3370) gives the impression that the only way to identify the device is either trial and error or opening the box. I assumed that was the basic problem. Using the debug output from Fritz!OS makes this much easier.
 
> Apart from that, using the power button as a way to enter setup mode is something i don't like.
> How do you explain the constraints a power button has to have in order to enter setup mode?
Using the same button for two purposes isn't very elegant, but I think it's similar to how the reset button is employed (or used to be?), you either get a normal reboot (short press) or to setup-mode (long press). Also /etc/rc.button/wps has some code which would be invoked by a short press.

I have to admit the use of the power button for setup would be fairly obvious for the AVM 3370 (because there are only 2 buttons anyway), but maybe not so much for people with other devices. Still, in the context of gluon the only use pattern involving a long-press is to get to setup mode (AFAIK). OpenWRT is taking care of the regular power-off, so there shouldn't be a device dependent problem.

OTOH, I also see the problem with possible clutter, and so far, the AVM 3370 would be the only device benefitting from this patch. (Grepping through the .dts* files, it seems all other devices with `KEY_POWER` also have `KEY_RESTART|KEY_WPS_BUTTON|KEY_PHONE`.)

patches/openwrt/0010-lantiq-xrx200-show_LAN_activity_on_AVM_3370.patch
dzzinstant: I agree. It's not vital, and other devices (e.g. AVM 4040) do well without it.

blocktrron:
Reset button handling is limited to setup-mode in Gluon, there's a patch which disabled the OpenWrt default. The WPS code execution also is not a problem, as we do not have WPS enabled on any VAP.

The issue from my side is from the documentation side. WPS and Reset buttons have to be accessible from the CPU while APs which use an additional Power management IC with a power button  (which makes sense) do not always report this as a clearly distinguishable interrupt to the CPU (if at all). In case of an overvoltage / over-temperature, they might signal a poweroff to the OS this way and cut off power after a certain amount of time.

In Gluon, this would (for example) affect the OCEDO Panda and possibly other enterprise APs (Aruba for sure, possibly Meraki).

Still, this is not easy to document, as it's not always obvious whether or not the Power button will end up signaling the interrupt to the CPU. Also, there are devices using a pure-hardware switch labeled power. 

I'm really impressed with the work you've already did with this device! But this device was already rejected on IRC when @Dark4MD first proposed the patches. Back than, we've figured out the device is kind of niche anyway so this would not have a lot of impact anyways (Oddly enough, @rotanid and myself own a unit).

blocktrron:
Thanks to @mweinelt , here are the original IRC logs:

```
22:06 <DarkAMD> skuril, grad mal die buttons an ner fritzbox 3370 testen wollen zwecks anstehendem pr und dabei festgestellt, abgesehen das die wlan taste wifi in openwrt heist und nichts tut, dass die box statt dect oder so (hat es ja nicht) ne power taste hat, die genau das tut. die box ausschalten und anschalten. war mir nicht bewusst das openwrt das kann, andererseits kann ich nen x86 image ja auch herunterfahren aber an nem router noch nie gesehen
22:14 <blocktrron> DarkAMD: 3370 macht halt wieder ein neues fass auf mit von außen nicht erkennbarer versionierung
08:41 <rotanid> DarkAMD: ja bei der 3370 gibts ja leider 2 unterschiedliche flash chips die unterschiedliche images brauchen, und wie blocktrron schreibt steht das nirgends auf gerät oder verpackung, welche version es ist... IMHO nimmt gluon die daher eher nicht auf... oder nur mit BROKEN
```

Dark4MD:
It's possible to get the NAND Type via the OEM Web Interface described here:
https://www.bachmann-lan.de/openwrt-auf-einer-avm-fritzbox-3370-installieren/
Have tested it and it works this way. So its possible to get the NAND type without opening device at least.
I think it would be nice to have it at least as broken because its the best use by now for this device because of the dropped official support.

blocktrron:
Well, this went cold.

To reach a consensus, i would propose to remove the added button handling, patch and docs entry and add them as broken.

Any opinions?

dzzinstant:
Thanks for picking this up, I almost had forgotten about it. I agree with your objections, I'll send a corrected submission tomorrow.

Adorfer:
@blocktrron 
> @Adorfer I don't understand what you want to say with your comment here.

there are other devices with shortcomings too. 
and we still use RPI (no reset/setup-button + no working wifi) or NanostationM (pcb board variant not detectable by label) in gluon.

PR #2057, opened 2020-06-13
gluon-mesh-batman-adv: do not delete bat0 during hardif teardown

NeoRaider:
With very bad timing, it is possible that the teardown script of a
gluon_mesh interface runs when bat0 was just created, but primary0 is not
yet added to it. Although there is no hardif to remove in this case,
bat0 will still be deleted, because there is no hardif in bat0.

Disable the interface removal logic by passing `-M` to `batctl interface`.

This was reported by @rotanid and @mbaumga


NeoRaider:
Backported to v2020.1.x and v2020.2.x; v2019.1.x is not affected.

PR #2006, opened 2020-05-03
 targets/lantiq-xrx200: added device Arcadyan VGV7510KW22 aka o2 Box 6431

4ludwig4:
Can't yet answer all questions but wanted to get this started to get feedback and gradually push this when I have time and the skills to do so.

The device is already in usage as a node (https://map.ffmuc.net/#!/en/map/1883bfbda4c8) and now after seeing it "works" I wanted to contribute by bringing this upstream.

- [x] must be flashable from vendor firmware
  - [x] webinterface
  - [ ] tftp
  - [x] other: serial (some people connect the serial without opening the case) in combination with tftp worked for me (updated openwrt wiki for this device: https://openwrt.org/toh/arcadyan/vgv7510kw22)  but there are also reports that people got it working without using the serial port.
- [ ] must support upgrade mechanism
  - [x] must have working sysupgrade
    - [x] must keep/forget configuration (if applicable)
      *think `sysupgrade [-n]` or `firstboot`*
  - [ ] must have working autoupdate
    *usually means: gluon profile name must match image name* (I guess an alias which has the right name is also okay as the openwrt image is called different than the output of the get_image_name() function)
- [x] reset/wps/phone button must return device into config mode (reset and wps buttons work)
- [x] primary mac should match address on device label (or packaging) (https://gluon.readthedocs.io/en/latest/dev/hardware.html#notes)
- wired network
  - [ ] should support all network ports on the device
  - [ ] must have correct port assignment (WAN/LAN)
  - This is a tricky one as already on OpenWRT it seems like my WAN port was broken. It could be that this is due to me opening the case not very carefully or an already damaged WAN port from before as I didn't test the router beforehand. Via vlan I mapped one of the LAN ports to the WAN port.
- wifi (if applicable)
  - [x] association with AP must be possible on all radios
  - [x] association with 802.11s mesh must be working on all radios
  - [x] ap/mesh mode must work in parallel on all radios
- led mapping
  - power/sys led (_critical, because led definitions are setup on firstboot only_)
    - [x] lit while the device is on
    - [x] should display config mode blink sequence
(https://gluon.readthedocs.io/en/latest/features/configmode.html)
  - radio leds
    - [x] should map to their respective radio
    - [ ] should show activity (if i remember correctly this was triggered by `phy0radio` which didn't work which seems to be an issue not just for me [https://openwrt.org/docs/guide-user/base-system/led_configuration#wifi_activity_triggers] so I switched to `phy0tx` manually)
  - switchport leds
[...]

PR #2058, opened 2020-06-14
Use LAN/WAN interfaces from board.json, add lantiq special case

NeoRaider:
Use the WAN port of switch0 instead of `dsl0` on lantiq devices where available. It would be great if we could test this on a few of our existing lantiq devices - I don't have any.

See also: https://github.com/freifunk-gluon/gluon/pull/2006#issuecomment-643479894

### DIFF
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/020-interfaces
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/020-interfaces
@@ -11,9 +11,12 @@ end
 local platform = require 'gluon.platform'
 local site = require 'gluon.site'
 
+local json = require 'jsonc'
 local uci = require('simple-uci').cursor()
 local unistd = require 'posix.unistd'
 
+local board_data = json.load('/etc/board.json')
+local network_data = (board_data or {}).network
 
 local function iface_exists(ifaces)
 	if not ifaces then return nil end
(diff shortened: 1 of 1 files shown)


blocktrron:
I'll be able to test this with a 7412 and 7312. However this might not be ideal, as they only have single LAN port.

NeoRaider:
@blocktrron Well, it should at least not break anything for these devices then.

4ludwig4:
@NeoRaider I can test this with the [Arcadyan VGV7510KW22](https://openwrt.org/toh/arcadyan/vgv7510kw22) if it makes sense? Don't know so how to test this yet.

NeoRaider:
@4ludwig4 Just rebase your change onto this branch and check if the WAN interface is assigned correctly after `sysupgrade -n` or `firstboot`.

NeoRaider:
@blocktrron Did you have a chance to test this?

4ludwig4:
@NeoRaider I wrote a fix in #2006 for this which I'm currently building then testing as your code didn't work for my Arcadyan VGV7510KW22 (o2 Box 6431).

`elseif platform.match('lantiq') then`
Don't know if I should just limit this to my device instead of doing this for all lantiq devices. Or how do you usually know that this change is device wide?

NeoRaider:
@4ludwig4 Thanks, your fix looks correct - I have updated my PR with a slightly modified version.

The whole lantiq platform uses the same code to set up the WAN interface in `board.json`, so the code should fine like this.

PR #2060, opened 2020-06-20
gluon-wan-dnsmasq: allow passing interfaces to update

blocktrron:
This commit adds the ability to pass a variable amount of interfaces to the update script.

This way, external custom packages which add a custom VPN WAN interface can update the DNS server gluon is using for WAN resolving.


package/gluon-wan-dnsmasq/luasrc/lib/gluon/wan-dnsmasq/update.lua
```diff
@@ -41,8 +41,11 @@ if type(static) == 'table' and #static > 0 then
     append_server(server)
   end
 else
-  pcall(append_interface_servers, 'wan6')
-  pcall(append_interface_servers, 'wan')
+  for idx, interface in ipairs(arg) do
+    if idx > 0 then
```
NeoRaider, line 45: `ipairs` starts at index 1, so this condition is redundant.

blocktrron:
@NeoRaider Thanks for the quick review.

I like the idea of using flag-files. I've updated the PR (till need to do a complete test-build). What do you think?

package/gluon-wan-dnsmasq/files/lib/gluon/wan-dnsmasq/interface.d/wan-bridge
```diff
@@ -0,0 +1,2 @@
+wan6
+wan
```
NeoRaider, line 2: Let's make the file name match the package name.

package/gluon-wan-dnsmasq/files/etc/hotplug.d/iface/50-gluon-wan-dnsmasq
```diff
@@ -1,3 +1 @@
-if [ "$INTERFACE" = 'wan' -o "$INTERFACE" = 'wan6' ]; then
-   /lib/gluon/wan-dnsmasq/update.lua
-fi
+/lib/gluon/wan-dnsmasq/update.lua "$INTERFACE"
```
NeoRaider, line 1: Removing the `$INTERFACE` checks will cause way more Lua processes to be spawned when setting up the network (or when tearing it down), so I'd like to keep it.

package/gluon-wan-dnsmasq/luasrc/lib/gluon/wan-dnsmasq/update.lua
```diff
@@ -36,13 +49,20 @@ end
 
 local static = uci:get_first('gluon-wan-dnsmasq', 'static', 'server')
 
+local handled_interfaces = handled_interfaces()
+
+if not util.contains(handled_interfaces, arg[1]) then
+  os.exit(0)
+end
```
NeoRaider, line 56: I'm not sure if this check is really useful - while checking inside instead of outside the script can avoid some redundancy, it is generally less efficient, and the current version always expects exactly one interface to be passed, which might not be sufficient for all usecases.

package/gluon-wan-dnsmasq/files/lib/gluon/wan-dnsmasq/interface.d/wan-bridge
NeoRaider: Hmm, and should we prepend a number to the filename to specify a precedence?

package/gluon-wan-dnsmasq/files/etc/hotplug.d/iface/50-gluon-wan-dnsmasq
blocktrron: My intention was to allow additional packages to only drop a file to `/lib/gluon/wan-dnsmasq/interface.d/`. But i see the performance impact this implies.

Will restore the previous behavior.

package/gluon-wan-dnsmasq/files/lib/gluon/wan-dnsmasq/interface.d/wan-bridge
blocktrron: FWIW, this specifies the order in which nameservers are added to resolv.conf. So this sounds like a good idea.

package/gluon-wan-dnsmasq/luasrc/lib/gluon/wan-dnsmasq/update.lua
blocktrron: From my POV this made a lot of sense if the behavior of `50-gluon-wan-dnsmasq` would be to call the script every time an interface comes up. But when we revert this, we can also drop this change.

blocktrron:
@NeoRaider Addressed your comments. Anything else left?

Issue #2063, opened 2020-06-28
Brief cross-domain link when switching domains


RalfJung:
### Bug report

To reproduce, make sure you have two nodes in the same domain that are meshing via wifi.
Then I run
```
uci set gluon.core.domain="<otherdomain>"
gluon-reconfigure
# wait a bit
reboot
```
on one of the devices.

**What is the problem?**
After "reconfigure" and before the reboot, router advertisements leak from one mesh to the other. So suddenly all nodes in one domain also have the IPv6 prefix from the other domain.

**What is the expected behaviour?**
Cross-domain links like this should be prevented.

**Gluon Version:**
Gluon 5696be72

**Site Configuration:**
https://git.hacksaar.de/FreifunkSaar/gluon-site

**Custom patches:**
none

blocktrron:
Just to be sure - this affects only the prefixes gluon-radvd advertises on the switching nodes?

RalfJung:
I am not sure -- we saw all 5 prefixes on the other side:
```
    inet6 2a03:2260:3009:1400:32b5:c2ff:fe6e:9d6b/64 scope global dynamic 
       valid_lft 1772sec preferred_lft 272sec
    inet6 2a03:2260:3009:1200:32b5:c2ff:fe6e:9d6b/64 scope global dynamic 
       valid_lft 1771sec preferred_lft 271sec
    inet6 2a03:2260:3009:1300:32b5:c2ff:fe6e:9d6b/64 scope global dynamic 
       valid_lft 1764sec preferred_lft 264sec
    inet6 2a03:2260:3009:1100:32b5:c2ff:fe6e:9d6b/64 scope global dynamic 
       valid_lft 1782sec preferred_lft 282sec
    inet6 fd4e:f2d7:88d2:fffd:32b5:c2ff:fe6e:9d6b/64 scope global dynamic 
       valid_lft 86364sec preferred_lft 14364sec
    inet6 2a03:2260:3009:400:32b5:c2ff:fe6e:9d6b/64 scope global dynamic 
       valid_lft 1792sec preferred_lft 292sec
    inet6 2a03:2260:3009:300:32b5:c2ff:fe6e:9d6b/64 scope global dynamic 
       valid_lft 1765sec preferred_lft 265sec
    inet6 2a03:2260:3009:200:32b5:c2ff:fe6e:9d6b/64 scope global dynamic 
       valid_lft 1738sec preferred_lft 238sec
[...]

PR #2065, opened 2020-06-30
Domain switch during reboot/gluon-reload

NeoRaider:
When changing gluon.core.domain, not all scripts and services switch to the new settings at the same time:

- Some parts of Gluon will use the new settings as soon as gluon.core.domain is set
- Some parts of Gluon will use the new settings as soon as gluon-reconfigure is run
- Some services need a restart to use the new settings (reboot or `gluon-reload`)

Fix this by adding the options `gluon.core.switch_domain` and `gluon.core.reconfigure` that allow to do the actual switch during reboot or `gluon-reload`.

This also changes the following packages that are affected by this issue, and which need testing:
- [ ] gluon-scheduled-domain-switch @blocktrron 
- [ ] gluon-hoodselector @2tata 

I currently don't have a setup to test these packages.

The domain change + gluon-reconfigure in `gluon-config-mode-domain-select` is not changed, so domain-dependent information on the reboot page is shown correctly.

Closes: #2063 @RalfJung


blocktrron:
@NeoRaider I'll test the scheduled domain switch. However i think it is quite obsolete now, as we've dropped IBSS and batman-adv 14.

Maybe it's best to move it to the community packages?

NeoRaider:
@blocktrron Might still be useful for migrations BATMAN_IV -> BATMAN_V, or IP range switches.

NeoRaider:
Oops, I just noticed that there is a major bug in this PR that breaks networking in the setup mode. I'm not sure how I missed that during testing...

RalfJung:
With this PR, what is the precedure for switching domains that I should test? Do I still set `gluon.core.domain`, or do I set the two new options instead?

Also I am a bit confused about why there are two options. Isn't the presence of `switch_domain` enough to indicate that a reconfigure should happen during reboot?

RalfJung:
Oh, the PR adds documentation explaining exactly that. Never mind :)

RalfJung:
Hm, to test this I tried to reproduce the problem on my virtual node without a mesh partner, but that does not seem to have worked... I switched the domains and did `gluon-reconfigure`, but then did not do the reboot. According to `tcpdump`, the local uradvd still sends out the announcements from the old domain:

Any idea what I need to do to see the old buggy behavior without having two nodes that mesh?

NeoRaider:
I assume that some respondd providers will start sending out new domain data as soon as `gluon.core.domain` is modified, as these are querying the site configuration directly. Other respondd providers will query UCI, seeing the new values as soon as `gluon-reconfigure` has run. `gluon-neighbour-info -r nodeinfo` or `gluon-neighbour-info -r statistics` will likely show some inconsistent data.

Outside of respondd providers, any package accessing the site/domain config or UCI at runtime may be affected, but I haven't looked at every package in detail.

RalfJung:
The main problem we saw was other nodes getting RAs from the wrong domain, so I thought I should see wrong RAs with `tcpdump` on the switching node. But I cannot see those. Strange.

I'll check the other indicators you mentioned later.

RalfJung:
In my VM, switching domains seems to work fine with the new command, even when using `--no-reboot`.

I cannot easily test if https://github.com/freifunk-gluon/gluon/issues/2063 is still around, that needs multiple wifi-capable devices and a build of the firmware for them.

rotanid:
> I cannot easily test if #2063 is still around, that needs multiple wifi-capable devices and a build of the firmware for them.

@RalfJung  but a test would help to get this solved ;-) as you can see it's just laying around...

docs/features/multidomain.rst
```diff
@@ -88,18 +88,25 @@ domain of a router, if and only if one of the above conditions matches.
...
 
-**via config mode:**
+When the node is not in config mode, ``gluon-switch-domain`` will automatically
+reboot the node by default. This can be suppressed by passing ``--no-reboot``::
 
-To allow switching the domain via config mode, ``config-mode-domain-select``
-has to be added to GLUON_FEATURES in the site.mk.
+    gluon-switch-domain --no-reboot 'newdomaincode'
+
+Switching the domain without reboot is currently **experimental**.
```
mweinelt, line 103: I think this statement is key as it makes very clear that we currently recommend rebooting to safely change the domain. This also expresses the need for more testing in the wild, which is apparently somewhat involved or @RalfJung would've already done so.

mweinelt:
@rotanid Unless you see field-testing as a requirement for this merge I would go with my sentiment in https://github.com/freifunk-gluon/gluon/pull/2065#discussion_r470995182 and merge this today.

rotanid:
> @rotanid Unless you see field-testing as a requirement for this merge I would go with my sentiment in [#2065 (comment)](https://github.com/freifunk-gluon/gluon/pull/2065#discussion_r470995182) and merge this today.

maybe we can make an issue or something to set field-testing as a blocker before 2020.3 release?

RalfJung:
> > I cannot easily test if #2063 is still around, that needs multiple wifi-capable devices and a build of the firmware for them.
> 
> @RalfJung but a test would help to get this solved ;-) as you can see it's just laying around...

I only own one device, so I cannot test this. I am not sure when we'll have the next in-person Freifunk hackathon here where others could also bring hardware, but we are not doing many in-person meetings these days.

So, sorry, not much I can do currently.

docs/features/multidomain.rst
RalfJung: I am completely fine with requiring a reboot. We saw the previous problem even with the reboot method, just because I waited a minute during the 3-command sequence that changes the domain.  But then gluon extensions (hoodselector, domain-director) should also do these reboots, and I am not sure if they do.

NeoRaider: The domain director should definitely be moved to the new commands as soon as this PR is merged.

The scheduled-domain-switch is changed as part of this PR and reboots for the change. The hoodselector uses `--no-reboot`, but as the whole hoodselector is experimental, using an experimental command seems fine to me.

mweinelt: > The domain director should definitely be moved to the new commands as soon as this PR is merged.

I can take care of that.

mweinelt: Proposed the change here https://git.darmstadt.ccc.de/ffda/firmware/packages/-/merge_requests/4

Issue #2064, opened 2020-06-30
2018.2.4+2020.1.3 compatible Ubiquity AP and AP LR (long range) are not distinguished


smoe:
Separate versions of the firmware are offered for these two devices that are just symlinks for real:
```
lrwxrwxrwx ...    59 Jun 14 02:08 gluon-ffoh-0.9.21~exp20200614-ubiquiti-unifi-ap-lr-sysupgrade.bin -> gluon-ffoh-0.9.21~exp20200614-ubiquiti-unifi-sysupgrade.bin
```
On the map and the node info page, the device describes itself as
```
Geräte-Modell | Ubiquiti UniFi
```
which tends to confuse users, for whom the "LR" likely is important to distinguish their installed devices. And one is tempted to wonder if the installed firmware was the correct one if not aware of the symlink.

Is there something that could be done about it? I likely can help with access to such a device if that is of any help.

Cheers,
Steffen

NeoRaider:
It's likely that the devices can be distinguished by some value in their EEPROM partition. Please provide us with dumps of the EEPROM partition (as listed in `/proc/mtd`) of at least 2 devices of each model (preferably more).

smoe:
We have 7 installed, but the only one I am confident about and have access to at the very moment is:
```
root@Carpe-Haus2-UG-west-K6-UbiquityAP_LR:~# cat /proc/mtd
dev:    size   erasesize  name
mtd0: 00040000 00010000 "u-boot"
mtd1: 00010000 00010000 "u-boot-env"
mtd2: 00760000 00010000 "firmware"
mtd3: 001c0000 00010000 "kernel"
mtd4: 005a0000 00010000 "rootfs"
mtd5: 00350000 00010000 "rootfs_data"
mtd6: 00040000 00010000 "cfg"
mtd7: 00010000 00010000 "EEPROM"
```
Another two LR are still boxed up. I hope to activate them temporarily over the upcoming days to provide this data. Others following this thread please help out for confirmed non-LR Unify APs.

NeoRaider:
@smoe Okay, so EEPROM is mtd7. You can just run `cat /dev/mtd7 > /tmp/eeprom` to read the contents of the partition into a file.

rotanid:
woah, old devices... we have only one of those in the field.
here's mtd7 eeprom from a non-LR non-AC "Unifi AP": https://upthis.de/dl/hnfcku/unifi_ap_eeprom
note: link is valid for only 7 days!

[...]

PR #2066, opened 2020-07-01
ar71xx: make Ubiquiti UniFi image naming more consistent, add detection for UniFI AP-LR

NeoRaider:
These patches are
- build-tested only
- not upstream yet (will submit after initial testing)

I can runtime test at least on UniFi AP Outdoor+ to make sure the rename doesn't break board detection, but I don't have any other UniFi or AirMax devices that might be affected by these changes.

Optimally, we can get test coverage on all of the following devices to make sure I haven't broken anything:
- [x] UniFi AP
- [x] UniFi AP-LR
- [ ] UniFi AP-PRO
- [ ] UniFi AP Outdoor
- [x] UniFi AP Outdoor+
- [ ] AirMax M2/M5 (Bullet/NS/NS Loco/Rocket rev. XM)

To test:
- Model name is displayed correctly
- Correct upgrade image name is chosen

Closes: #2064

### DIFF
--- a/patches/openwrt/0014-ar71xx-make-Unifi-AP-naming-more-consistent.patch
+++ b/patches/openwrt/0014-ar71xx-make-Unifi-AP-naming-more-consistent.patch
@@ -1,0 +1,177 @@
+From: Matthias Schiffer <mschiffer@universe-factory.net>
+Date: Wed, 1 Jul 2020 20:45:38 +0200
+Subject: ar71xx: make Unifi AP naming more consistent
+
+- Make device titles and model strings match
+- More consistent word splitting:
+  - UniFiAP -> UniFi AP
+  - UAP -> UniFi AP
+- Make "UniFi AP-PRO" formatting match official name
+
+- The image names are adjusted as well:
+  - uap -> unifiap
+  - unifi-outdoor -> unifiap-outdoor
+
+Signed-off-by: Matthias Schiffer <mschiffer@universe-factory.net>
+
+diff --git a/target/linux/ar71xx/base-files/lib/ar71xx.sh b/target/linux/ar71xx/base-files/lib/ar71xx.sh
+index 044ef4eae5e1cd2f04bacc5b0c18a151f2fe1253..9e3819d56ee4384dcf2444f345a3fd35a437cd10 100755
+--- a/target/linux/ar71xx/base-files/lib/ar71xx.sh
++++ b/target/linux/ar71xx/base-files/lib/ar71xx.sh
+@@ -1424,10 +1424,10 @@ ar71xx_board_detect() {
+ 	*"Tube2H")
+ 		name="tube2h"
+ 		;;
+-	*"UniFi")
++	*"UniFi AP")
+ 		name="unifi"
+ 		;;
+-	*"UniFi AP Pro")
++	*"UniFi AP-PRO")
+ 		name="uap-pro"
+ 		;;
+ 	*"UniFi-AC-LITE/MESH")
+@@ -1438,10 +1438,10 @@ ar71xx_board_detect() {
+ 		name="unifiac-pro"
+ 		ubnt_unifi_ac_board_detect
+ 		;;
+-	*"UniFiAP Outdoor")
++	*"UniFi AP Outdoor")
+ 		name="unifi-outdoor"
+ 		;;
+-	*"UniFiAP Outdoor+")
++	*"UniFi AP Outdoor+")
+ 		name="unifi-outdoor-plus"
+ 		;;
+ 	*"WAM250")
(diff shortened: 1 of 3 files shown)


NeoRaider:
It would be great if someone could at least give this PR a spin on an UniFi AP-LR before I submit these patches for OpenWrt upstream. @smoe?

smoe:
Will do. I presume your patch also works against the OpenWrt in 2020.1.3? Well, I'll find out.

rotanid:
do we also need to reflect these changes in the OpenWrt ath79 target? after all, the ar71xx target will not be part of the next OpenWrt release...

NeoRaider:
> do we also need to reflect these changes in the OpenWrt ath79 target? 

None of these devices have been ported to ath79 so far.

> after all, the ar71xx target will not be part of the next OpenWrt release...

ar71xx will not be part of the next OpenWrt *release*, but it will still be available in the release *branch* to be built from source, so this does not really affect Gluon. While it is likely that the code will eventually bitrot and be removed, I don't see this happening for the next OpenWrt release.

smoe:
> It would be great if someone could at least give this PR a spin on an UniFi AP-LR before I submit these patches for OpenWrt upstream. @smoe?

It worked just fine: "Ubiquiti UniFi AP-LR" it reads. What do you want me to do to check the "Correct upgrade image name is chosen"?

rotanid:
> > do we also need to reflect these changes in the OpenWrt ath79 target?
> 
> None of these devices have been ported to ath79 so far.

this is wrong, see here:
https://git.openwrt.org/?p=openwrt/openwrt.git;a=blob_plain;f=target/linux/ath79/dts/ar7241_ubnt_unifi.dts;hb=refs/heads/master

smoe:
Binaries (factory and sysupgrade) for this unifi-branch are on http://www.ffoh.de/firmware/ubiquiti-unify-20200705/ for additional volunteers/resources.  I sadly don't have access to the other nodes of this kind on our network.

NeoRaider:
> > > do we also need to reflect these changes in the OpenWrt ath79 target?
> > 
> > 
> > None of these devices have been ported to ath79 so far.
> 
> this is wrong, see here:
> https://git.openwrt.org/?p=openwrt/openwrt.git;a=blob_plain;f=target/linux/ath79/dts/ar7241_ubnt_unifi.dts;hb=refs/heads/master

Hmm, I see. For ath79, the model name is always taken from the DTS at the moment, so we will need a somewhat different solution.

rotanid:
for "UniFi AP" the output using the image from @smoe looks OK:
```
root@freifunk44d9e7b96366:~# # With vendor and hardware version (example: "tp-li
nk-tl-wdr4300-v1")
root@freifunk44d9e7b96366:~# lua -e 'print(require("platform_info").get_image_name())'
ubiquiti-unifi-ap
root@freifunk44d9e7b96366:~# # With vendor and hardware version (human readable, example: "TP-Link TL-WDR4300 v1")
root@freifunk44d9e7b96366:~# lua -e 'print(require("platform_info").get_model())'
Ubiquiti UniFi AP
root@freifunk44d9e7b96366:~# # Only type of router (example: "tl-wdr4300")
root@freifunk44d9e7b96366:~# lua -e 'print(require("platform_info").get_board_name())'
unifi
```
@smoe btw, it's "Unifi" not "Unify" ;-)

@NeoRaider ok , how could the solution for ath79 look?

i think especially the AirMax M2 stuff should get a test before we merge this... unfortunately i don't have such a device available for testing.

rotanid:
@NeoRaider as we are probably nearing our last major Gluon release based on OpenWrt 19.07 which also is the last to include ar71xx let me raise the question if you want to work on this PR before this release or if you want to bury it and only fix it some time in ath79 ?

smoe:
https://coco.ffoh.de/meshviewer/#!/de/map/24a43c7dd022
I just installed 2020.2.2 and did not find the patch in there. :(

rotanid:
> https://coco.ffoh.de/meshviewer/#!/de/map/24a43c7dd022
> I just installed 2020.2.2 and did not find the patch in there. :(

no wonder, as this PR is not merged and marked as DRAFT

rotanid:
@NeoRaider as we're now in the process of doing the last major Gluon release including ar71xx, this PR has either to be done know or being re-opened then doing changes for ath79 ;-)

NeoRaider:
I think we can defer this to the next release, so we only have to deal with one target.

rotanid:
@NeoRaider but @blocktrron mentioned on IRC, we might need to tackle ar71xx, too because of upgrades? maybe he can share his thoughts.

blocktrron:
This depends on the concept how we want to design the update process, so this is TBD.

Issue #1922, opened 2020-01-25
gluon-radv-filterd spams log when there are network issues


RalfJung:
### Bug report

**What is the problem?**
When there are network issues on the node, "logread" prints hundreds of lines of gluon-radv-filterd not being able to find the originator for certain MAC addresses (or something along those lines). Due to log rotation, this means "logread" only shows approximately the last hour of logs, making it impossible to debug issue that started more than an hour ago.

The log looks like this:
```
Mon Nov 25 15:15:10 2019 daemon.err gluon-radv-filterd[1515]: Unable to find TQ for originator ca:fe:ba:be:00:01 (router ee:ee:ee:ee:00:01)
Mon Nov 25 15:15:10 2019 daemon.err gluon-radv-filterd[1515]: Unable to find TQ for originator ca:fe:ba:be:00:04 (router ee:ee:ee:ee:00:04)
Mon Nov 25 15:15:10 2019 daemon.err gluon-radv-filterd[1515]: Unable to find TQ for originator ca:fe:ba:be:00:02 (router ee:ee:ee:ee:00:02)
Mon Nov 25 15:15:10 2019 daemon.err gluon-radv-filterd[1515]: Unable to find TQ for originator ca:fe:ba:be:00:03 (router ee:ee:ee:ee:00:03)
Mon Nov 25 15:15:26 2019 daemon.err gluon-radv-filterd[1515]: Unable to find TQ for originator ca:fe:ba:be:00:01 (router ee:ee:ee:ee:00:01)
Mon Nov 25 15:15:26 2019 daemon.err gluon-radv-filterd[1515]: Unable to find TQ for originator ca:fe:ba:be:00:04 (router ee:ee:ee:ee:00:04)
Mon Nov 25 15:15:26 2019 daemon.err gluon-radv-filterd[1515]: Unable to find TQ for originator ca:fe:ba:be:00:02 (router ee:ee:ee:ee:00:02)
Mon Nov 25 15:15:26 2019 daemon.err gluon-radv-filterd[1515]: Unable to find TQ for originator ca:fe:ba:be:00:03 (router ee:ee:ee:ee:00:03)
```

**What is the expected behaviour?**
gluon-radv-filterd shouldn't flood the log.

**Gluon Version:** v2019.1.1

**Site Configuration:** https://git.hacksaar.de/FreifunkSaar/gluon-site

**Custom patches:** https://git.hacksaar.de/FreifunkSaar/gluon-ffsaar

[...]

PR #2068, opened 2020-07-01
gluon-radv-filterd: minor cleanup

NeoRaider:
Closes: #1922 @RalfJung


blocktrron:
How about including this in v2020.2?

NeoRaider:
> How about including this in v2020.2?

Sounds good to me.

Issue #1653, opened 2019-02-18
gluon-config-mode-geo-location-osm: Providing a TileLayer URL


yanosz:
### Bug report

**What is the problem?**
There is no configuration option for a TileLayer URL in gluon-config-mode-geo-location-osm.
Roughly following https://wiki.openstreetmap.org/wiki/Tile_usage_policy using custom TileLayers (e.g. from Freifunk Aachen) appears to be a good idea

**What is the expected behaviour?**
Having a setting like tilelayer_url to specifiy the tile-layer.

**Gluon Version:**
2018.2

yanosz:
imho this is a minor enhancement

mweinelt:
The passing of the XYZ tile url would need to happen in `gluon-config-mode-geo-location-osm` but we would also need to patch `gluon-web-osm` here (https://github.com/freifunk-gluon/gluon/blob/0d4188d40b4d406feb0f9224b39acb93c50e052b/package/gluon-web-osm/javascript/gluon-web-osm.js#L50) and in the function signature.

PR #2072, opened 2020-07-04
gluon-config-mode-geo-location-osm, gluon-web-osm: add support for custom tile layers

NeoRaider:
Allow replacing the default OSM layer with a custom XYZ layer in site.conf.

Closes: #1653

I wonder if the "OSM" in the package names etc. even still makes sense like this... but I guess it can be adjusted in a separate PR if necessary.


docs/user/site.rst
```diff
@@ -488,6 +493,11 @@ config_mode \: optional
               },
               zoom = 13,
               -- openlayers_url = 'http://ffac.example.org/openlayer',
+              -- tile_layer = {
+              --   type = 'XYZ',
+              --   url = 'https://{a-c}.tile.openstreetmap.org/{z}/{x}/{y}.png',
```
2tata, line 498: Maybe we can use the preexsiting url parameter to reduce the amount of variables in the siteconf?

NeoRaider: @2tata What preexisting URL parameter? The `openlayers_url` is where the JS and CSS code for OpenLayers is loaded from, it has nothing to do with the layer data.

2tata: ah, I was not entirely sure.

PR #2074, opened 2020-07-04
gluon-web: don't display outdoor mode on preserve_channels

blocktrron:
This will hide the outdoor mode setting on compatible devices in case
the defined channels should be preserved.

Otherwise a user might be under the impression their device is compliant
with out door operation when in reality it still uses prohibited
channels.

### DIFF
--- a/package/gluon-config-mode-outdoor/luasrc/lib/gluon/config-mode/wizard/0250-outdoor.lua
+++ b/package/gluon-config-mode-outdoor/luasrc/lib/gluon/config-mode/wizard/0250-outdoor.lua
@@ -1,8 +1,14 @@
 return function(form, uci)
 	local platform = require 'gluon.platform'
+	local wireless = require 'gluon.wireless'
 
 	if not (platform.is_outdoor_device() and platform.device_uses_11a(uci)) then
 		-- only visible on wizard for outdoor devices
+		return
+	end
+
+	if wireless.preserve_channels(uci) then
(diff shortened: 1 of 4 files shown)


blocktrron:
There's room for improvement in terms of the `preserve_channels` setting and it's impacts.

However, this is not directly related with the change introduced in this PR. I'll merge this and create a new PR for documentation improvements for the `preserve_channels` setting.

Adorfer:
i'd prefer to have a 
"preserve_channels" reserved for autoupdater-use only, as it was introduced (and set by most users)
in other words
1) outdoormode overrule "preserve-channels"
2) introduce a "force-outdoorchannels" (which overrules outdoor-mode)
3) perhaps rename "preserve-channels" to something like  "preseve-indoor-channels" or "preserve_channels_during_autoupdate"

blocktrron:
@Adorfer I don't see how this is related to this change. This change is about improving the status-quo to not let users under the wrong impression their device is now ready for outdoor operation (which it might not be).

As already explained in the forums, your proposal requires major reconstructions of core parts, as the current implementation handles setting the wireless configuration as an update-script. This operation is currently stateless in terms of whether it being called from the config-mode reboot or on a software upgrade.

Also this will bring more edge cases which need to be addressed (e.g. tracking custom channels in both states).

rotanid:
@blocktrron maybe also backport this PR to v2020.2.x ?

blocktrron:
will do.

Issue #1832, opened 2019-09-28
Android does not respond to multicast listener queries in power saving mode


lemoer:
We figured out that Android does not respond to multicast listener queries when it is in sleeping mode.

When the device wakes up and the multicast entry in the router has expired, it takes up to 20 seconds till the router sends the next multicast listener query and the device responds. Only after this IPv6 neighbour discovery is possible again and the device has restored its IPv6 connectivity.

T-X:
This sounds like an Android bug. Either an IGMP/MLD query should wake up the device, so that it could receive it. Or it should send an unsolicited IGMP/MLD report upon waking up.

I'm not familiar with the Android ecosystem. @lemoer, could you file a bug for Android and reference it here?

lemoer:
https://issuetracker.google.com/issues/141785896

lemoer:
Currently I do not plan to reopen this issue by myself:

1. I do not really know how to answer their question of reproduction of the issue.
2. I do not have enough time to investigate here.

So best would be someone takes this issue in his responsibility or we revert multicast optimization for now.
-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

CodeFetch:
It takes years for the buggy android devices to vanish...

T-X:
I updated the Si6 Networks ipv6toolkit and wrote a script around it to more easilly reproduce and evaluate the issues with MLD in an isolated setup:

* https://github.com/T-X/icmpv6-test
* https://github.com/T-X/ipv6toolkit/tree/fixes%2Bmldq6-tool

And I opened a new issues on the Android bugtracker here with additional information on how to reproduce the issue and added the results from the script:

* https://issuetracker.google.com/issues/149630944

T-X:
What I found quite interesting when playing with the ICMPv6 test script was:

* Everything looks fine when the phone is connected to a charger
* When on battery, Echo Requests-Replies still work fine
* When on battery, Neighbor Advertisements to Neighbor Solicitations seem rate-limited?
* When on battery, MLD Queries are filtered completely?
* When on battery, when sending MLD Queries via unicast (not quite standard compliant) there are at least some responses, similar to NDISC (now heavily rate-limited instead of filtered completely?)

Especially because ICMPv6 Echo Requests always work it currently seems to me that they are actively filtering based on the ICMPv6 type. And the people who implemented that might have overlooked the importance of MLD.

T-X:
[...]

PR #2076, opened 2020-07-05
kernel: bridge: Implement MLD Querier wake-up calls / Android bug workaround

T-X:
Implement a configurable MLD Querier wake-up calls "feature" which
works around a widely spread Android bug in connection with IGMP/MLD
snooping.

Currently there are mobile devices (e.g. Android) which are not able
to receive and respond to MLD Queries reliably because the Wifi driver
filters a lot of ICMPv6 when the device is asleep - including
MLD. This in turn breaks IPv6 communication when MLD Snooping is
enabled. However there is one ICMPv6 type which is allowed to pass and
which can be used to wake up the mobile device: ICMPv6 Echo Requests.

If this bridge is the selected MLD Querier then setting
"multicast_wakeupcall" to a number n greater than 0 will send n
ICMPv6 Echo Requests to each host behind this port to wake
them up with each MLD Query. Upon receiving a matching ICMPv6 Echo
Reply an MLD Query with a unicast ethernet destination will be sent
to the specific host(s).

Link: https://issuetracker.google.com/issues/149630944
Link: https://github.com/freifunk-gluon/gluon/issues/1832

### DIFF
--- a/patches/openwrt/0016-kernel-bridge-Implement-MLD-Querier-wake-up-calls-Android-bug-workaround.patch
+++ b/patches/openwrt/0016-kernel-bridge-Implement-MLD-Querier-wake-up-calls-Android-bug-workaround.patch
@@ -1,0 +1,2605 @@
+From: Linus Lüssing <linus.luessing@c0d3.blue>
+Date: Tue, 30 Jun 2020 18:01:56 +0200
+Subject: kernel: bridge: Implement MLD Querier wake-up calls / Android bug workaround
+
+Implement a configurable MLD Querier wake-up calls "feature" which
+works around a widely spread Android bug in connection with IGMP/MLD
+snooping.
+
+Currently there are mobile devices (e.g. Android) which are not able
+to receive and respond to MLD Queries reliably because the Wifi driver
+filters a lot of ICMPv6 when the device is asleep - including
+MLD. This in turn breaks IPv6 communication when MLD Snooping is
+enabled. However there is one ICMPv6 type which is allowed to pass and
+which can be used to wake up the mobile device: ICMPv6 Echo Requests.
+
+If this bridge is the selected MLD Querier then setting
+"multicast_wakeupcall" to a number n greater than 0 will send n
+ICMPv6 Echo Requests to each host behind this port to wake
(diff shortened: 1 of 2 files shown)


rotanid:
thanks. i've added some labels, including the "blocked" label which stays until this has been merged upstream, as we don't like to carry patches "forever" as you know

rotanid:
@T-X have you submitted the patches upstream, maybe you can put the links in the initial text?
at least in the google thread i can't find the patches, i did not search in Linux Kernel patchwork...

T-X:
Changelog v2:
* rebased to master
* fixed the ICMPv6 header check, was a no-op due to a typo ("if (ipv6_mc_check_icmpv6 < 0)" vs. "if (ipv6_mc_check_icmpv6(skb) < 0)" ) in br_multicast_wakeupcall_check()

T-X:
@rotanid I've submitted the Linux kernel as an RFC [here](https://patchwork.ozlabs.org/project/netdev/patch/20200816202424.3526-1-linus.luessing@c0d3.blue/), but no one was really happy about it.

And I couldn't really spot an alternative suggestion which everyone would be happy about either.

So not sure if that means we'd have to deal with this issue on our own for now. Or if I should retry submitting that Linux bridge patch to OpenWrt later.

rotanid:
@T-X generally speaking we have a policy to not carry patches which aren't upstream, if we wanted to make exceptions to this i think we should allow every maintainer a veto right or at least it would need to be a majority call.

mweinelt:
I see the usability benefit of carrying these patches, as well as the downside with having to rebase it every once in a while. I want to have it and will gladly switch to a better solution when one comes by.

blocktrron:
I would propose to go with these patches for now, in order to give it a substantial amount of testing.

We can still drop them later on, if rebasing becomes unbearable or a better solution is worked out.

mweinelt:
@T-X Please rebase, we have no routing patches and this series is introducing patchset 0002.

T-X:
Rebase to current master done (no changes).

mweinelt (approved):

PR #2077, opened 2020-07-05
openwrt: backport DFS fixes

blocktrron:
This backports two fixes related to operation on DFS-required
channels.

When a DFS-required channel was selected as the regular
(non-outdoor-mode) 5 GHz channel, hostapd would switch to a non-DFS
channel as OpenWrt did not pass a chanlist of allowed ACS channels.

When hostapd is given a single channel for the chanlist or there's no
available channel left (all allowed channels are in the no-occupancy
period), hostapd prints "no DFS channels left, waiting for NOP to
finish" to the syslog but never stopped transmitting on this channel,
still sending out beacon frames and allowing client data transfer.

### DIFF
--- a/patches/openwrt/0014-mac80211-create-channel-list-for-fixed-channel-operation.patch
+++ b/patches/openwrt/0014-mac80211-create-channel-list-for-fixed-channel-operation.patch
@@ -1,0 +1,28 @@
+From: David Bauer <mail@david-bauer.net>
+Date: Sat, 4 Jul 2020 13:20:02 +0200
+Subject: mac80211: create channel list for fixed channel operation
+
+Currently a device which has a DFS channel selected using the UCI
+channel setting might switch to a non-DFS channel in case no chanlist is
+provided (UCI setting "channels") when the radio detects a DFS event.
+
+Automatically add a chanlist consisting of the configured channel when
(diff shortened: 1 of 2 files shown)


mweinelt:
Isn't this also very relevant for OpenWrt 19.07 and would warrant a backport there?

blocktrron:
@mweinelt I'm hesitant backporting it to 19.07 until the patch is accepted upstream.

The mac80211 patch is something i would backport, even though it changes hostapds behavior.

blocktrron:
Thinking about this again, i would propose to add these two patches for now to the Gluon tree and i will backport both after ~2 weeks to 19.07.

mweinelt:
I have no issue with testdriving them for two weeks, although I doubt we will gain much feedback in that timeframe.

mweinelt:
https://lists.infradead.org/pipermail/hostap/2020-July/038615.html

mweinelt:
With the v2020.2 released planned for this weekend and these being edge cases we decided to unblock the release. We do plan to revisit this in time for v2020.2.1.

blocktrron:
Merging this now.The hostapd patch is most likely the most important to us, while the other one only brings a benefit in case a community selects a outdoor channel as their regular (non-outdoor) channel.

PR #2084, opened 2020-07-24
actions: cache OpenWrt DL folder

blocktrron:
This adds the ability to cache OpenWrt dependencies on a per-target
base. Artifacts over 10MB are excluded, as GitHub imposes a limit of 5G
of available space per repository cache. This affects mostly
linux(-firmware) and gcc / gdb.

The goal is to reduce the total amount of requests necessary to fetch
dependencies.


blocktrron:
Merging this, as it reduces the build time significantly ATM, as the OpenWrt sources CDN is unavailable.

PR #2085, opened 2020-08-08
gluon-status-page: add wireless client count

blocktrron:
This adds the wireless client count for 2.4GHz and 5 GHz radios to the status page.

Previously, only the total client count advertised by the mesh protocol was visible.


rotanid:
looks good, only question i have is if we want to list "other" separately so one does not have to substract 2.4+5 from total manually?

blocktrron:
@rotanid I'm not a big fan of this, as this number is not precise.

blocktrron:
Anybody against backporting this to v2020.2?

rotanid:
ok

PR #2092, opened 2020-08-14
build: check for unsynced feeds before build

NeoRaider:
Forgetting to `make update` or leaving uncommitted changes in the
repositories managed by Gluon is a recurring cause of confusion, even
for experienced developers. Let's print an obvious warning message in
this case.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -114,13 +114,6 @@ define CheckTarget
 	fi
 endef
 
-define CheckExternal
-	if [ ! -d openwrt ]; then
-		echo "You don't seem to have obtained the external repositories needed by Gluon; please call \`make update\` first!"
-		exit 1
-	fi
-endef
-
 define CheckSite
 	if ! GLUON_SITEDIR='$(GLUON_SITEDIR)' GLUON_SITE_CONFIG='$(1).conf' $(LUA) -e 'assert(dofile("scripts/site_config.lua")(os.getenv("GLUON_SITE_CONFIG")))'; then
 		echo 'Your site configuration ($(1).conf) did not pass validation'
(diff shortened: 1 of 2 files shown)


mweinelt:
Can we consider modules in `$GLUON_SITEDIR/modules` as well?

NeoRaider:
> Can we consider modules in `$GLUON_SITEDIR/modules` as well?

This should already work, as we include scripts/modules.sh to get the list.

SvenRoederer:
For the Freifunk-Berlin firmware we are using a regular Makefile which automatically takes care of the required steps. As some time ago we adapted some of the gluon-scripts, this Makefile have been updated too. So it should be relatively easy to use it for Gluon too.
Do you have some policy to not use such Makefile to it's full extend or why Gluon sticks to the "call the required buildsteps manually" philosophy?

NeoRaider:
Hmm, I guess having control over the different build steps makes the most sense from the POV of developer experience, but it might not always be the optimum for users...

IMO, `make update` should always stay a separate step (it's conceptually a very different action from the actually running the build - this matches how dependency updates are handled in other embedded build systems, for example OpenWrt feed updates, or OE/Yocto layer management).

A 'build all targets' command would be a nice extension, but it's difficult to do without adding another layer of Make, which I would like to avoid (our Makefile is complex enough...)

It might be an option to replace our toplevel Makefile with a Python script when updating to the next OpenWrt version (as that version will switch its Python dependency from Python 2 to 3). I'd expect that such a move would make extensions to the build logic significantly easier. It's just a half-baked idea at the moment though, so we might not go through with this plan any time soon, if ever.

PR #2101, opened 2020-08-19
package/gluon-core: 200-wireless simplify if conditions

2tata:


blocktrron:
Remove the `package/` from the prefix. Otherwise LGTM.

PR #2102, opened 2020-08-20
mesh-vpn: fully abstract VPN methods

blocktrron:
This fully abstracts VPN methods, making gluon-mesh-vpn-fastd and
gluon-mesh-vpn-tunneldigger completely self-contained.

Provide a LUA interface for generic interacting with VPN methods in
gluon-mesh-vpn-core and web packages.

This also adds the ability to install tunneldigger and fastd to the same
image, selecting the VPN method based on the selected domain.

### DIFF
--- a/package/gluon-config-mode-mesh-vpn/luasrc/lib/gluon/config-mode/reboot/0100-mesh-vpn.lua
+++ b/package/gluon-config-mode-mesh-vpn/luasrc/lib/gluon/config-mode/reboot/0100-mesh-vpn.lua
@@ -1,19 +1,13 @@
 local site_i18n = i18n 'gluon-site'
 
 local uci = require("simple-uci").cursor()
-local unistd = require 'posix.unistd'
 
 local platform = require 'gluon.platform'
 local site = require 'gluon.site'
 local sysconfig = require 'gluon.sysconfig'
-local util = require "gluon.util"
+local vpn = require 'gluon.mesh-vpn'
 
 local pretty_hostname = require 'pretty_hostname'
-
-
-local has_fastd = unistd.access('/lib/gluon/mesh-vpn/fastd')
-local has_tunneldigger = unistd.access('/lib/gluon/mesh-vpn/tunneldigger')
-
 
 local hostname = pretty_hostname.get(uci)
 local contact = uci:get_first("gluon-node-info", "owner", "contact")
@@ -22,19 +16,18 @@ local pubkey
 local msg
 
 
-if has_tunneldigger then
-	local tunneldigger_enabled = uci:get_bool("tunneldigger", "mesh_vpn", "enabled")
-	if not tunneldigger_enabled then
-		msg = site_i18n._translate('gluon-config-mode:novpn')
+if vpn.enabled() then
+	local _, active_vpn = vpn.get_active_provider()
+
+	if active_vpn ~= nil then
+		pubkey = active_vpn.public_key()
 	end
-elseif has_fastd then
-	local fastd_enabled = uci:get_bool("fastd", "mesh_vpn", "enabled")
-	if fastd_enabled then
-		pubkey = util.trim(util.exec("/etc/init.d/fastd show_key mesh_vpn"))
+
+	if pubkey ~= nil then
 		msg = site_i18n._translate('gluon-config-mode:pubkey')
-	else
-		msg = site_i18n._translate('gluon-config-mode:novpn')
 	end
+else
+	msg = site_i18n._translate('gluon-config-mode:novpn')
 end
 
 if not msg then return end
--- a/package/gluon-config-mode-mesh-vpn/luasrc/lib/gluon/config-mode/wizard/0300-mesh-vpn.lua
+++ b/package/gluon-config-mode-mesh-vpn/luasrc/lib/gluon/config-mode/wizard/0300-mesh-vpn.lua
@@ -1,10 +1,8 @@
-local unistd = require 'posix.unistd'
-
-local has_fastd = unistd.access('/lib/gluon/mesh-vpn/fastd')
-local has_tunneldigger = unistd.access('/lib/gluon/mesh-vpn/tunneldigger')
+local vpn = require 'gluon.mesh-vpn'
+local _, active_vpn = vpn.get_active_provider()
 
 return function(form, uci)
-	if not (has_fastd or has_tunneldigger) then
+	if active_vpn == nil then
 		return
 	end
 
--- a/package/gluon-mesh-vpn-core/luasrc/lib/gluon/mesh-vpn/update-config
+++ b/package/gluon-mesh-vpn-core/luasrc/lib/gluon/mesh-vpn/update-config
@@ -1,14 +1,7 @@
 #!/usr/bin/lua
 
 local uci = require('simple-uci').cursor()
-local unistd = require 'posix.unistd'
-
-local vpn
-if unistd.access('/lib/gluon/mesh-vpn/fastd') then
-	vpn = 'fastd'
-elseif unistd.access('/lib/gluon/mesh-vpn/tunneldigger') then
-	vpn = 'tunneldigger'
-end
+local vpn_name, vpn = require('gluon.mesh-vpn').get_active_provider()
 
 local vpn_config = {
 	enabled = uci:get_bool('gluon', 'mesh_vpn', 'enabled'),
@@ -17,32 +10,20 @@ local vpn_config = {
 	limit_ingress = uci:get('gluon', 'mesh_vpn', 'limit_ingress'),
 }
 
-uci:delete('simple-tc', 'mesh_vpn')
-uci:section('simple-tc', 'interface', 'mesh_vpn', {
-	ifname = 'mesh-vpn',
-	enabled = vpn_config.limit_enabled,
-	limit_egress = vpn_config.limit_egress,
-})
-
-if vpn == 'fastd' then
-	uci:set('fastd', 'mesh_vpn', 'enabled', vpn_config.enabled)
-	uci:set('simple-tc', 'mesh_vpn', 'limit_ingress', vpn_config.limit_ingress)
-else
+if vpn_name ~= 'fastd' then
 	uci:set('fastd', 'mesh_vpn', 'enabled', false)
+	uci:save('fastd')
 end
-uci:save('fastd')
 
-if vpn == 'tunneldigger' then
-	uci:set('tunneldigger', 'mesh_vpn', 'enabled', vpn_config.enabled)
-
-	if vpn_config.limit_enabled then
-		uci:set('tunneldigger', 'mesh_vpn', 'limit_bw_down', vpn_config.limit_ingress)
-	else
-		uci:delete('tunneldigger', 'mesh_vpn', 'limit_bw_down')
-	end
-else
+if vpn_name ~= 'tunneldigger' then
 	uci:set('tunneldigger', 'mesh_vpn', 'enabled', false)
+	uci:save('tunneldigger')
 end
-uci:save('tunneldigger')
 
-uci:save('simple-tc')
+vpn.enable(vpn_config.enabled)
+if vpn_config.limit_enabled then
+	vpn.set_limit(vpn_config.limit_ingress, vpn_config.limit_egress)
+else
+	vpn.set_limit(nil, nil)
+end
+
--- a/package/gluon-mesh-vpn-core/luasrc/lib/gluon/upgrade/500-mesh-vpn
+++ b/package/gluon-mesh-vpn-core/luasrc/lib/gluon/upgrade/500-mesh-vpn
@@ -5,11 +5,11 @@ local users = require 'gluon.users'
 local util = require 'gluon.util'
 
 local uci = require('simple-uci').cursor()
-local unistd = require 'posix.unistd'
 
+local vpn_core = require 'gluon.mesh-vpn'
 
 uci:section('network', 'interface', 'mesh_vpn', {
-	ifname = 'mesh-vpn',
+	ifname = vpn_core.get_interface(),
 	proto = 'gluon_mesh',
 	transitive = true,
 	fixed_mtu = true,
@@ -35,12 +35,7 @@ uci:save('firewall')
 
 -- VPN migration
 if not uci:get('gluon', 'mesh_vpn') then
-	local vpn
-	if unistd.access('/lib/gluon/mesh-vpn/fastd') then
-		vpn = 'fastd'
-	elseif unistd.access('/lib/gluon/mesh-vpn/tunneldigger') then
-		vpn = 'tunneldigger'
-	end
+	local vpn, _ = vpn_core.get_active_provider()
 
 	local fastd_enabled = uci:get('fastd', 'mesh_vpn', 'enabled')
 	local tunneldigger_enabled = uci:get('tunneldigger', 'mesh_vpn', 'enabled')
--- a/package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn.lua
+++ b/package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn.lua
@@ -1,7 +1,47 @@
+local uci = require('simple-uci').cursor()
+
+local util = require 'gluon.util'
+
 local M = {}
 
-function M.get_mesh_vpn_interface()
-  return 'mesh-vpn'
+function M.enabled()
+	return uci:get_bool('gluon', 'mesh_vpn', 'enabled')
+end
+
+function M.enable(val)
+	return uci:set('gluon', 'mesh_vpn', 'enabled', val)
+end
+
+function M.get_interface()
+	return 'mesh-vpn'
+end
+
+function M.get_provider(name)
+	return require('gluon.mesh-vpn.provider.' .. name)
+end
+
+function M.get_provider_names()
+	local out = {}
+
+	for _, v in ipairs(util.glob('/lib/gluon/mesh-vpn/provider/*')) do
+		table.insert(out, v:match('([^/]+)$'))
+	end
+
+	return out
+end
+
+function M.get_active_provider()
+	-- Active provider is the provider in use
+	-- by the currently active site / domain
+
+	for _, name in ipairs(M.get_provider_names()) do
+		local provider = M.get_provider(name)
+		if provider.active() then
+			return name, provider
+		end
+	end
+
+	return nil, nil
 end
 
 return M
--- a/package/gluon-mesh-vpn-tunneldigger/luasrc/lib/lua/gluon/mesh-vpn/provider/tunneldigger.lua
+++ b/package/gluon-mesh-vpn-tunneldigger/luasrc/lib/lua/gluon/mesh-vpn/provider/tunneldigger.lua
@@ -1,0 +1,42 @@
+local uci = require('simple-uci').cursor()
+
+local site = require 'gluon.site'
+local vpn_core = require 'gluon.mesh-vpn'
+
+local M = {}
+
+function M.public_key()
+	return nil
+end
+
+function M.enable(val)
+	uci:set('tunneldigger', 'mesh_vpn', 'enabled', val)
+	uci:save('tunneldigger')
+end
+
+function M.active()
+	return site.mesh_vpn.tunneldigger() ~= nil
+end
+
+function M.set_limit(ingress_limit, egress_limit)
+	if ingress_limit ~= nil then
+		uci:set('tunneldigger', 'mesh_vpn', 'limit_bw_down', ingress_limit)
+	else
+		uci:delete('tunneldigger', 'mesh_vpn', 'limit_bw_down')
+	end
+
+	if egress_limit ~= nil then
+		uci:section('simple-tc', 'interface', 'mesh_vpn', {
+			ifname = vpn_core.get_interface(),
+			enabled = true,
+			limit_egress = egress_limit,
+		})
+	else
+		uci:delete('simple-tc', 'mesh_vpn')
+	end
+
+	uci:save('tunneldigger')
+	uci:save('simple-tc')
+end
+
+return M
--- a/package/gluon-web-admin/files/lib/gluon/config-mode/view/admin/info.html
+++ b/package/gluon-web-admin/files/lib/gluon/config-mode/view/admin/info.html
@@ -6,14 +6,19 @@
 	local sysconfig = require 'gluon.sysconfig'
 	local platform = require 'gluon.platform'
 	local util = require "gluon.util"
+	local has_vpn, vpn = pcall(require, 'gluon.mesh-vpn')
 
 	local _ = translate
 
 
 	local pubkey
-	local meshvpn_enabled = uci:get_bool("fastd", "mesh_vpn", "enabled")
-	if meshvpn_enabled then
-		pubkey = util.trim(util.exec('/etc/init.d/fastd show_key mesh_vpn'))
+	if has_vpn and vpn.enabled() then
+		local _, active_vpn = vpn.get_active_provider()
+
+		if active_vpn ~= nil then
+			pubkey = active_vpn.public_key()
+		end
+
 		if pubkey == '' then
 			pubkey = nil
 		end
--- a/package/gluon-mesh-vpn-fastd/luasrc/lib/gluon/upgrade/400-mesh-vpn-fastd
+++ b/package/gluon-mesh-vpn-fastd/luasrc/lib/gluon/upgrade/400-mesh-vpn-fastd
@@ -2,6 +2,7 @@
 
 local site = require 'gluon.site'
 local util = require 'gluon.util'
+local vpn_core = require 'gluon.mesh-vpn'
 
 local uci = require('simple-uci').cursor()
 
@@ -37,7 +38,7 @@ end
 uci:section('fastd', 'fastd', 'mesh_vpn', {
 	group = 'gluon-mesh-vpn',
 	syslog_level = syslog_level,
-	interface = 'mesh-vpn',
+	interface = vpn_core.get_interface(),
 	mode = 'tap',
 	mtu = site.mesh_vpn.mtu(),
 	secure_handshakes = true,
--- a/package/gluon-mesh-vpn-fastd/luasrc/usr/lib/lua/gluon/mesh-vpn/provider/fastd.lua
+++ b/package/gluon-mesh-vpn-fastd/luasrc/usr/lib/lua/gluon/mesh-vpn/provider/fastd.lua
@@ -1,0 +1,36 @@
+local uci = require('simple-uci').cursor()
+
+local site = require 'gluon.site'
+local util = require 'gluon.util'
+local vpn_core = require 'gluon.mesh-vpn'
+
+local M = {}
+
+function M.public_key()
+	return util.trim(util.exec('/etc/init.d/fastd show_key mesh_vpn'))
+end
+
+function M.enable(val)
+	uci:set('fastd', 'mesh_vpn', 'enabled', val)
+	uci:save('fastd')
+end
+
+function M.active()
+	return site.mesh_vpn.fastd() ~= nil
+end
+
+function M.set_limit(ingress_limit, egress_limit)
+	uci:delete('simple-tc', 'mesh_vpn')
+	if ingress_limit ~= nil and egress_limit ~= nil then
+		uci:section('simple-tc', 'interface', 'mesh_vpn', {
+			ifname = vpn_core.get_interface(),
+			enabled = true,
+			limit_egress = egress_limit,
+			limit_ingress = ingress_limit,
+		})
+	end
+
+	uci:save('simple-tc')
+end
+
+return M
--- a/package/gluon-mesh-vpn-tunneldigger/luasrc/lib/gluon/upgrade/400-mesh-vpn-tunneldigger
+++ b/package/gluon-mesh-vpn-tunneldigger/luasrc/lib/gluon/upgrade/400-mesh-vpn-tunneldigger
@@ -2,6 +2,7 @@
 
 local site = require 'gluon.site'
 local util = require 'gluon.util'
+local vpn_core = require 'gluon.mesh-vpn'
 
 local uci = require('simple-uci').cursor()
 
@@ -23,7 +24,7 @@ end
 uci:section('tunneldigger', 'broker', 'mesh_vpn', {
 	enabled = enabled,
 	uuid = util.node_id(),
-	interface = 'mesh-vpn',
+	interface = vpn_core.get_interface(),
 	bind_interface = 'br-wan',
 	group = 'gluon-mesh-vpn',
 	broker_selection = 'usage',


package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn/core.lua
```diff
@@ -0,0 +1,69 @@
+local util = require "gluon.util"
+
+local M = {}
+
+function M.get_interface()
+  return 'mesh-vpn'
+end
+
+function M.get_method(name)
+  return require("gluon.mesh-vpn.method." .. name)
```
NeoRaider, line 10: Maybe we should find another term here, as "method" is already used in fastd configuration (and also appears in site.conf for this reason).

NeoRaider:
I see a few conceptual issues that need to be discuss:
- Should the chosen VPN implementation only be a matter of site/domain config, or should is be possible for the user to change this (for example turning the security/performance mode of `gluon-web-mesh-vpn-fastd` into a generic switch hat selects between fastd and tunneldigger)?
- There are a few places in this PR that query a specific method whether it is enabled. This doesn't make sense to me - VPN enable status should be a single value in `/etc/config/gluon`. The `enabled` flags of the individual VPN implementation config files are dependent values that are populated based on the global enable status and the selected VPN implementation; they should only be set, but never be queried by Gluon's scripts (except for migration purposes, in case the global enable status doesn't exist yet)

package/gluon-mesh-vpn-fastd/luasrc/usr/lib/lua/gluon/mesh-vpn/method/fastd.lua
```diff
@@ -0,0 +1,38 @@
...
+function M.enabled()
+	return uci:get_bool("fastd", "mesh_vpn", "enabled")
+end
+
+function M.enable(val)
+	local enabled = '0'
+	if val then
+		enabled = '1'
+	end
+	uci:set("fastd", "mesh_vpn", "enabled", enabled)
```
NeoRaider, line 21: Just pass `true` or `false` to `uci:set`, `simple-uci` will turn that into `'0'` and `'1'`.

package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn/core.lua
```diff
@@ -0,0 +1,69 @@
...
+  local methods = {}
+
+  for _, name in ipairs(M.get_inactive_method_names()) do
+    table.insert(methods, M.get_method(name))
+  end
+
+  return methods
+end
+
+return M
```
NeoRaider, line 69: Indentation should use tabs

package/gluon-mesh-vpn-fastd/luasrc/usr/lib/lua/gluon/mesh-vpn/method/fastd.lua
```diff
@@ -0,0 +1,38 @@
...
+function M.set_limit(_, _)
+	-- handled by simple-tc
+	return nil
+end
+
+function M.uci_sections()
+	return {"fastd"}
+end
+
+return M
```
NeoRaider, line 38: Missing final newline in multiple files

package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn/core.lua
blocktrron: How about `vpn_proto`?

package/gluon-mesh-vpn-fastd/luasrc/usr/lib/lua/gluon/mesh-vpn/method/fastd.lua
blocktrron: Missed this. Fixed.

package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn/core.lua
blocktrron: Fixed.

package/gluon-mesh-vpn-fastd/luasrc/usr/lib/lua/gluon/mesh-vpn/method/fastd.lua
blocktrron: Fixed.

blocktrron:
@NeoRaider 

For the first point, I would propose to leave it to the site configuration for now. The goal of this PR is not to add this functionality but to remove the core VPN package dependencies on the specific VPN implementation packages.

Your second remark is correct. I've simply didn't correctly interpret the context and it's goal for which the specific VPN states are extracted. I'll fix that.

package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn/core.lua
mweinelt: Protocol is the correct identifier if you want to differentiate between fastd, tunneldigger and wireguard. They're different protocols.

NeoRaider: I'm thinking about adding L2TP support to fastd though, making it an alternative to tunneldigger...

I think a more general term would be something like "feature provider" or "function provider" - unfortunately this could get mixed up with service providers :/

package/gluon-mesh-vpn-core/luasrc/lib/gluon/mesh-vpn/update-config
```diff
@@ -22,27 +15,14 @@ uci:section('simple-tc', 'interface', 'mesh_vpn', {
...
-		uci:delete('tunneldigger', 'mesh_vpn', 'limit_bw_down')
-	end
+vpn.enable(vpn_config.enabled)
+if vpn_config.limit_enabled then
+	vpn.set_limit(vpn_config.limit_ingress, vpn_config.limit_egress)
 else
-	uci:set('tunneldigger', 'mesh_vpn', 'enabled', false)
+	vpn.set_limit(nil, nil)
 end
-uci:save('tunneldigger')
```
NeoRaider, line 46: You are removing the code that disables the unavailable VPNs, this is not correct.

NeoRaider: Hmm, I wrote the code this way so that even VPNs that are not installed anymore get disabled (for the case when for example `fastd` is installed, but `gluon-mesh-vpn-fastd` isn't).

What we really want is a generic facility that allows removed packages to clean up after themselves on firmware upgrades, that way we wouldn't need a full list of packages that could exist. I have a few ideas how to solve this...

package/gluon-mesh-vpn-fastd/luasrc/usr/lib/lua/gluon/mesh-vpn/method/fastd.lua
```diff
@@ -0,0 +1,34 @@
...
+
+local M = {}
+
+function M.public_key()
+	return util.trim(util.exec("/etc/init.d/fastd show_key mesh_vpn"))
+end
+
+function M.enabled()
+	return uci:get_bool("fastd", "mesh_vpn", "enabled")
+end
```
NeoRaider, line 14: The impl-specific `enabled()` methods can be removed.

package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn/core.lua
```diff
@@ -0,0 +1,79 @@
...
+	for _, name in ipairs(M.get_method_names()) do
+		local method = M.get_method(name)
+
+		if method.available() then
+			return name
+		end
+	end
+
+	return nil
+end
```
NeoRaider, line 43: As `get_active_method_name()` needs to load the method modules anyways, it would be nicer to merge it into `get_active_method()` (which can then have two return values, the name and the implementation).

package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn/core.lua
```diff
@@ -0,0 +1,79 @@
...
+
+function M.get_inactive_methods()
+	local methods = {}
+
+	for _, name in ipairs(M.get_inactive_method_names()) do
+		table.insert(methods, M.get_method(name))
+	end
+
+	return methods
+end
```
NeoRaider, line 77: Again, `get_inactive_methods()` and `get_inactive_method_names()` both iterate over all methods and load all impls. Let's have a single `get_inactive_methods()` function that returns a table of impls keyed by the names.

Also, this is currently unused, but I assume it might become used in the future?

package/gluon-config-mode-mesh-vpn/luasrc/lib/gluon/config-mode/wizard/0300-mesh-vpn.lua
```diff
@@ -64,5 +62,11 @@ return function(form, uci)
...
 	end
 
-	return {'gluon', 'fastd', 'tunneldigger', 'simple-tc'}
+	local uci_sections = {'gluon', 'simple-tc'}
+
+	for _, section in ipairs(active_vpn.uci_sections()) do
+		table.insert(uci_sections, section)
+	end
+
+	return {'gluon', 'simple-tc'}
```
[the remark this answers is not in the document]
NeoRaider: Maybe we should first get rid of all these useless return values first and simply do a global `uci commit` at the end of the config wizard... It will only write files that need to change anyways.

NeoRaider (changes requested):

CodeFetch:
@NeoRaider
> Should the chosen VPN implementation only be a matter of site/domain config, or should is be possible for the user to change this (for example turning the security/performance mode of gluon-web-mesh-vpn-fastd into a generic switch hat selects between fastd and tunneldigger)?

I've been working on IPSec using L2TP where I'd like to be able to let the user configure the encryption this way. Please make it as generic as possible or we will rip it out again, later.

blocktrron:
Fixed all outstanding comments.

Regarding the teardown of old VPN methods - I tend to leave it as - is for now. This might not be perfect, but we remain consistent in the behavior.

For the future, we can have a teardown manager which manages teardown in case a package does no exist anymore. The goal of allowing a third-party VPN method from a community feed is still possible this way, they have to take care of teardown themselves though. 

@AiyionPrime Not build / ru ntested yet although the linter is happy.

AiyionPrime:
Thanks!
I just compiled and installed it on one of my cudy wr1000 (ramips-mt76x8).

You five got access to it, welcome to my kitchen; obviously leave the rest of the network untouched.

AiyionPrime:
I rebuilt the firmware;  http://meshvpnabstracttest.n.ffh.zone/cgi-bin/status runs now on 251d550eed8290a78ecadcea2f5326d436e4b95f.

In case our dns is mesed up:

package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn.lua
```diff
@@ -1,7 +1,66 @@
...
+
+function M.enable(val)
+	return uci:set("gluon", "mesh_vpn", "enabled", val)
+end
+
+function M.get_interface()
+	return 'mesh-vpn'
+end
+
+function M.get_method(name)
```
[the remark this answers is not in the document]
AiyionPrime: In that case the same would apply to the other methods of `M`, as well as wide parts of this PR including the title.

Though I see where this is going; `proto`is currently used mostly for the distinction of network protocols like udp, tcp and icmp in gluon.
Whether this lines up enough with 'fastd', 'babel', 'tunneldigger', 'wireguard' is up for you to decide.

`method` on the other hand, is used less frequently but in the context of (fastd) cypher-methods.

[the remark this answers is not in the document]
mweinelt: > In that case the same would apply to the other methods of `M`, as well as wide parts of this PR including the title.

Yeah, that was kinda implied. I wasn't going to spam every occurrence, so here we are.

package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn.lua
```diff
@@ -1,7 +1,66 @@
...
+	local out = {}
+
+	for _, v in ipairs(util.glob("/lib/gluon/mesh-vpn/method/*")) do
+		table.insert(out, v:match('([^/]+)$'))
+	end
+
+	return out
+end
+
+function M.get_active_method()
```
lemoer, line 33:
```suggestion
function M.get_active_method()
	-- The "active" method is the (lexically) last of the "available" methods.
```

package/gluon-mesh-vpn-tunneldigger/luasrc/lib/lua/gluon/mesh-vpn/method/tunneldigger.lua
```diff
@@ -0,0 +1,33 @@
...
+function M.public_key()
+	return nil
+end
+
+function M.enable(val)
+	uci:set("tunneldigger", "mesh_vpn", "enabled", val)
+	uci:save("tunneldigger")
+end
+
+function M.active()
```
AiyionPrime, line 16: As there are multiple vpn methods, that are available but only one that is returned using `M.get_active_method()` this function might better be called `available()` to avoid a false affordance.
```suggestion
function M.available()
```

package/gluon-mesh-vpn-fastd/luasrc/usr/lib/lua/gluon/mesh-vpn/method/fastd.lua
```diff
@@ -0,0 +1,39 @@
...
+function M.public_key()
+	return util.trim(util.exec("/etc/init.d/fastd show_key mesh_vpn"))
+end
+
+function M.enable(val)
+	uci:set("fastd", "mesh_vpn", "enabled", val)
+	uci:save("fastd")
+end
+
+function M.active()
```
AiyionPrime, line 17: Same as in https://github.com/freifunk-gluon/gluon/pull/2102/files#r626032365
```suggestion
function M.available()
```

package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn.lua
```diff
@@ -1,7 +1,66 @@
...
+	local name = nil
+	local method = nil
+
+	for _, n in ipairs(M.get_method_names()) do
+		method = M.get_method(n)
+
+		if method.active() then
```
AiyionPrime, line 40: Related to https://github.com/freifunk-gluon/gluon/pull/2102/files#r626032365
```suggestion
		if method.available() then
```

package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn.lua
```diff
@@ -1,7 +1,66 @@
...
+	return nil, nil
+end
+
+function M.get_inactive_methods()
+	local methods = {}
+
+	for _, name in ipairs(M.get_method_names()) do
+		local method = M.get_method(name)
+
+		if not method.active() then
```
AiyionPrime, line 58: Related to https://github.com/freifunk-gluon/gluon/pull/2102/files#r626032365
```suggestion
		if not method.available() then
```

AiyionPrime:
@NeoRaider we just saw, you asked for the invert of the changes I proposed;
why do you think is `active` is the better choice in in `package/gluon-mesh-vpn-fastd`and `package/gluon-mesh-vpn-tunneldigger`?

lemoer:
After thinking about this again, I do not get the point of this whole `active()` (or `available()`) function...

If the package `gluon-mesh-vpn-fastd` is selected for the build, the build will fail due to the [check_site.lua script here](https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-mesh-vpn-fastd/check_site.lua#L2) anyway if `mesh_vpn.fastd` is undefined in any of the sites or domains. This means `active()` (or `available()`) would always return true for a valid firmware build, doesn't it? Or did I miss a point here?

blocktrron:
Rationale for `available` is to allow for firmware which packs multiple VPN methods and selects them on per-domain basis.

lemoer:
Ah, ok. Then `check_site.lua` has to be altered, doesn't it?

blocktrron:
Where exactly does a check_site script require to be modified?

lemoer:
[This line](https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-mesh-vpn-fastd/check_site.lua#L2) will fail, if it isn't

blocktrron:
Okay, i see.

The mesh-vpn abstraction is only one step towards the flexible VPN provider goal, so i would go for that in a follow-up PR. This Task is for adapting the integration to the core VPN service, which should initially keep the behavior while trying to achieve an extensible interface we do not shoot ourselves in the foot with.

package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn.lua
blocktrron: Clarified below, added altered description.

package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn.lua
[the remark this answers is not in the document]
blocktrron: proto is a better fit, thanks for that suggestion.

package/gluon-mesh-vpn-tunneldigger/luasrc/lib/lua/gluon/mesh-vpn/method/tunneldigger.lua
blocktrron: As explained to @lemoer , this is intended to have different methods for different domains but one vpn method active per domain. SO I'd stick with active here.

blocktrron:
Adressed the comments.

I'd stick with active in this context, as available would imply a multitude of methods to be available for a site.

package/gluon-mesh-vpn-tunneldigger/luasrc/lib/lua/gluon/mesh-vpn/proto/tunneldigger.lua
```diff
@@ -0,0 +1,33 @@
+local uci = require("simple-uci").cursor()
+
+local site = require 'gluon.site'
```
AiyionPrime, line 3: This is a small thing left regarding code-quality;
the require-statements with their used brackets and quotes are somewhat inconsistent.
But other than that the code looks fine now.

package/gluon-mesh-vpn-tunneldigger/luasrc/lib/lua/gluon/mesh-vpn/proto/tunneldigger.lua
```diff
@@ -0,0 +1,33 @@
...
+end
+
+function M.active()
+	return site.mesh_vpn.tunneldigger() ~= nil
+end
+
+function M.set_limit(ingress_limit, _)
+	if ingress_limit ~= nil then
+		uci:set("tunneldigger", "mesh_vpn", "limit_bw_down", ingress_limit)
+	else
```
AiyionPrime, line 23: Same nitpick regarding quotes would apply here.

package/gluon-mesh-vpn-fastd/luasrc/usr/lib/lua/gluon/mesh-vpn/proto/fastd.lua
```diff
@@ -0,0 +1,39 @@
...
+	return site.mesh_vpn.fastd() ~= nil
+end
+
+function M.set_limit(ingress_limit, egress_limit)
+	uci:delete('simple-tc', 'mesh_vpn')
+	if ingress_limit ~= nil and egress_limit ~= nil then
+		uci:section('simple-tc', 'interface', 'mesh_vpn', {
+			ifname = 'mesh-vpn',
+			enabled = 1,
+			limit_egress = ingress_limit,
```
AiyionPrime, line 27: Why is this swapped?

package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn.lua
```diff
@@ -1,7 +1,54 @@
+local uci = require("simple-uci").cursor()
+
+local util = require "gluon.util"
```
AiyionPrime, line 3: just a small inconsistency

AiyionPrime (commented):
The only thing that matters to me is the question in fastd.lua.
Everything else is just nitpicking.
Looks good.

package/gluon-mesh-vpn-fastd/luasrc/usr/lib/lua/gluon/mesh-vpn/proto/fastd.lua
blocktrron: Not intentional. Fixed.

AiyionPrime (approved):
The code is still compiling but as the last diff was quite clear, I expect no issues in compiling and running it. The test device will be updated in about 13 minutes.

package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn.lua
```diff
@@ -1,7 +1,54 @@
...
+		proto = M.get_proto(n)
+
+		if proto.active() then
+			name = n
+		end
+	end
+
+	if proto ~= nil then
+		return name, proto
+	end
```
lemoer, line 49:
```suggestion
	for _, name in ipairs(M.get_proto_names()) do
		local proto = M.get_proto(n)

		if proto.active() then
			return name, proto
		end
	end
```

package/gluon-mesh-vpn-core/luasrc/lib/gluon/upgrade/500-mesh-vpn
```diff
@@ -5,8 +5,8 @@ local users = require 'gluon.users'
 local util = require 'gluon.util'
 
 local uci = require('simple-uci').cursor()
-local unistd = require 'posix.unistd'
 
+local vpn_core = require 'gluon.mesh-vpn'
 
 uci:section('network', 'interface', 'mesh_vpn', {
 	ifname = 'mesh-vpn',
```
lemoer, line 12: It seems like you missed to replace `'mesh-vpn'` here with your newly introduced method `get_interface()`. I already made another commit, where I touched all the instances of this string. See here: https://github.com/freifunk-gluon/gluon/pull/2168/commits/67cd4e0e07b127557b329df56f143462d3c45658 . You can use this commit to cross check.

package/gluon-web-admin/files/lib/gluon/config-mode/view/admin/info.html
```diff
@@ -6,14 +6,15 @@
 	local sysconfig = require 'gluon.sysconfig'
 	local platform = require 'gluon.platform'
 	local util = require "gluon.util"
+	local vpn = require 'gluon.mesh-vpn'
```
lemoer, line 9: I suppose this will break things, if `gluon-mesh-vpn-core` is not baked into the firmware.

lemoer (commented):
The rest looks good to me.

package/gluon-web-admin/files/lib/gluon/config-mode/view/admin/info.html
```diff
@@ -6,14 +6,15 @@
...
 
 	local _ = translate
 
 
 	local pubkey
-	local meshvpn_enabled = uci:get_bool("fastd", "mesh_vpn", "enabled")
-	if meshvpn_enabled then
-		pubkey = util.trim(util.exec('/etc/init.d/fastd show_key mesh_vpn'))
+	if vpn.enabled() then
```
lemoer, line 15:
```suggestion
	if has_vpn and vpn.enabled() then
```

package/gluon-web-admin/files/lib/gluon/config-mode/view/admin/info.html
blocktrron: IMHO depending on the core VPN package is better, given it is only pulling in some minified scripts.

blocktrron: It already depends on `gluon-config-mode-core`

package/gluon-web-admin/files/lib/gluon/config-mode/view/admin/info.html
blocktrron: Depends on config-mode-core.

package/gluon-web-admin/files/lib/gluon/config-mode/view/admin/info.html
lemoer: As of now, this is not a good idea since `gluon-mesh-vpn-core` checks the site.conf with its `check_site.lua`. If we add this dependency, this would mean that even if you want to build an image without vpn, you would have to define e.g. `mesh_vpn.mtu` in your site.conf. (For reference: [Link](https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-mesh-vpn-core/check_site.lua#L2))

lemoer: `gluon-config-mode-core` != `gluon-mesh-vpn-core`

AiyionPrime:
The current state is built and deployed on

http://meshvpnabstracttest.n.ffh.zone/
2001:678:978:114:b64b:d6ff:fe22:4ebc

AiyionPrime:
Latest push compiles, and runs on http://meshvpnabstracttest.n.ffh.zone/
(2001:678:978:114:b64b:d6ff:fe22:4ebc)

@lemoer `gluon-reconfigure` works without problems.

lemoer:
We have two approvals here. So can we merge this?

package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn.lua
```diff
@@ -1,7 +1,47 @@
...
+function M.enable(val)
+	return uci:set('gluon', 'mesh_vpn', 'enabled', val)
+end
+
+function M.get_interface()
+	return 'mesh-vpn'
+end
+
+function M.get_proto(name)
+	return require('gluon.mesh-vpn.proto.' .. name)
```
NeoRaider, line 20: The term "protocol" seems very inaccurate to me. How about "provider"?

package/gluon-mesh-vpn-fastd/luasrc/usr/lib/lua/gluon/mesh-vpn/proto/fastd.lua
```diff
@@ -0,0 +1,40 @@
...
+function M.active()
+	return site.mesh_vpn.fastd() ~= nil
+end
+
+function M.set_limit(ingress_limit, egress_limit)
+	uci:delete('simple-tc', 'mesh_vpn')
+	if ingress_limit ~= nil and egress_limit ~= nil then
+		uci:section('simple-tc', 'interface', 'mesh_vpn', {
+			ifname = vpn_core.get_interface(),
+			enabled = 1,
```
NeoRaider, line 27: With `simple-uci`, you can use a boolean here.

package/gluon-mesh-vpn-tunneldigger/luasrc/lib/lua/gluon/mesh-vpn/proto/tunneldigger.lua
```diff
@@ -0,0 +1,33 @@
...
+	return site.mesh_vpn.tunneldigger() ~= nil
+end
+
+function M.set_limit(ingress_limit, _)
+	if ingress_limit ~= nil then
+		uci:set('tunneldigger', 'mesh_vpn', 'limit_bw_down', ingress_limit)
+	else
+		uci:delete('tunneldigger', 'mesh_vpn', 'limit_bw_down')
+	end
+	uci:save('tunneldigger')
```
NeoRaider, line 26: Tunneldigger also needs simple-tc for upload limit.

blocktrron:
Addressed comments and rebased on #2223

AiyionPrime:
There are currently 17 occurrences of `proto`, are they meant to stay?

AiyionPrime:
Looks better, I'll test later, when the dust has settled a little.

blocktrron:
Works from my testing.

AiyionPrime:
Compiles and gluon-reconfigure does it's job. I'll close blocktrron/gluon#6 then.

package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn.lua
```diff
@@ -1,7 +1,47 @@
...
+
+function M.get_provider(name)
+	return require('gluon.mesh-vpn.provider.' .. name)
+end
+
+function M.get_provider_names()
+	local out = {}
+
+	for _, v in ipairs(util.glob('/lib/gluon/mesh-vpn/provider/*')) do
+		table.insert(out, v:match('([^/]+)$'))
```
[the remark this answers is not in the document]
AiyionPrime: Apparently not, nevermind. https://github.com/freifunk-gluon/gluon/blob/a2be901803469d378d6469551a65c20f744aefff/scripts/check_site.lua#L22

AiyionPrime (changes requested):
Just reread the code once more.

package/gluon-config-mode-mesh-vpn/luasrc/lib/gluon/config-mode/reboot/0100-mesh-vpn.lua
```diff
@@ -1,40 +1,30 @@
...
-	if not tunneldigger_enabled then
-		msg = site_i18n._translate('gluon-config-mode:novpn')
-	end
-elseif has_fastd then
-	local fastd_enabled = uci:get_bool("fastd", "mesh_vpn", "enabled")
-	if fastd_enabled then
-		pubkey = util.trim(util.exec("/etc/init.d/fastd show_key mesh_vpn"))
+if vpn.enabled() then
+	local _, active_vpn = vpn.get_active_provider()
+	pubkey = active_vpn.public_key()
```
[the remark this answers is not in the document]
blocktrron: Yes

package/gluon-web-admin/files/lib/gluon/config-mode/view/admin/info.html
```diff
@@ -6,14 +6,15 @@
...
 
 	local _ = translate
 
 
 	local pubkey
-	local meshvpn_enabled = uci:get_bool("fastd", "mesh_vpn", "enabled")
-	if meshvpn_enabled then
-		pubkey = util.trim(util.exec('/etc/init.d/fastd show_key mesh_vpn'))
+	if has_vpn and vpn.enabled() then
+		local _, active_vpn = vpn.get_active_provider()
```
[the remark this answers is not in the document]
blocktrron: Yes

package/gluon-mesh-vpn-tunneldigger/luasrc/lib/lua/gluon/mesh-vpn/provider/tunneldigger.lua
```diff
@@ -0,0 +1,46 @@
...
+	else
+		uci:delete('simple-tc', 'mesh_vpn')
+	end
+
+	uci:save('tunneldigger')
+	uci:save('simple-tc')
+end
+
+function M.uci_sections()
+	return {'tunneldigger'}
```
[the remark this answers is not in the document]
blocktrron: These methods are now obsolete anyways, as uci::commit is called globally on reconfigure anyways. I removed them now.

package/gluon-web-admin/files/lib/gluon/config-mode/view/admin/info.html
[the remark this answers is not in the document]
AiyionPrime: resolved in 98a1c196ed8d76c0449ed85f291b4fac1ace9467

package/gluon-mesh-vpn-tunneldigger/luasrc/lib/lua/gluon/mesh-vpn/provider/tunneldigger.lua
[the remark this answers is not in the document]
AiyionPrime: resolved in 98a1c196ed8d76c0449ed85f291b4fac1ace9467

package/gluon-config-mode-mesh-vpn/luasrc/lib/gluon/config-mode/reboot/0100-mesh-vpn.lua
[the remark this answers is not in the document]
AiyionPrime: resolved in 98a1c196ed8d76c0449ed85f291b4fac1ace9467

AiyionPrime (approved):
Looks good to me.
@NeoRaider or @mweinelt may you review once more?

AiyionPrime:
I compiled the code before the weekend; and just did so again. It's deployed on http://meshvpnabstracttest.n.ffh.zone/cgi-bin/status.
You all still should have access to it.

blocktrron:
Verified initial setup & upgrades are working in a fastd environment.

Let's do this.

PR #2103, opened 2020-08-22
gluon-ebtables-filter-multicast: allow respondd queries

T-X:
This allows running a respondd querier and map server behind a Gluon
node.

For instance at Freifunk Lübeck we now moved the map server
behind a Gluon VM and removed batman-adv and fastd from the
map server VM to reduce the maintenance work.

Increased multicast overhead should be minimal / non existent, as it is
unlikely to accidentally have respondd queriers running behind a Gluon
node.

### DIFF
--- a/package/gluon-ebtables-filter-multicast/luasrc/lib/gluon/ebtables/110-mcast-allow-respondd
+++ b/package/gluon-ebtables-filter-multicast/luasrc/lib/gluon/ebtables/110-mcast-allow-respondd
@@ -1,0 +1,1 @@
+rule 'MULTICAST_OUT -p IPv6 --ip6-protocol udp --ip6-destination-port 1001 --ip6-dst ff05::2:1001 -j RETURN'


package/gluon-ebtables-filter-multicast/luasrc/lib/gluon/ebtables/110-mcast-allow-respondd
```diff
@@ -0,0 +1,2 @@
+rule 'MULTICAST_OUT -p IPv6 --ip6-protocol udp --ip6-destination-port 1001 --ip6-dst ff02::2:1001 -j RETURN'
```
NeoRaider, line 1: The first rule doesn't seem useful - current Gluon nodes don't bind to `ff02::2:1001` over the mesh interface anyways.

T-X:
Changelog v3:
* removed rule for the legacy ff02::2:1001 respondd address

package/gluon-ebtables-filter-multicast/luasrc/lib/gluon/ebtables/110-mcast-allow-respondd
T-X: Fine for me, FFHL has converged to ff05::2:1001. Less legacy to carry along in Gluon without that rule.

rubo77:
> at Freifunk Lübeck we now moved the map server
> behind a Gluon VM and removed batman-adv and fastd from the
> map server VM to reduce the maintenance work.

What is the benefit against running the map on a gateway?

T-X:
@rubo77 We've never run the map server on a gateway. To be able to (re)move any gateway VM at any time, without affecting other services.

kpanic23:
We have a split setup. hopglass-servers run on one of our gateways. The map is hosted on our services server (which also hosts website, wiki, email, mailing lists, mattermost, etc...) which collects the map data directly from the gateway every minute via cronjob. Grafana and prometheus also run on the services server. The services server is not connected to the Freifunk network.

rubo77:
We have connected or services server with the map via gre tunnel to a gateway, so it is connected to the Freifunk network to collect the data directly

Issue #522, opened 2015-10-12
Redesign API for respondd


tcatm:

jplitza:
I suppose this is the continuation of the discussion we had in #465?

NeoRaider:
@jplitza, yes. We'd like to get this done before 2015.2, so we can get rid of the old API as soon as possible.

NeoRaider:
Suggestion: use JSON for the requests. Draft:

```
{
  "keys": ["nodeinfo"],
  "single": true,
  "compress": true
}
```

* `keys`: lists the keys to get (so multiple types can be requested with a single request, as supported by the current GET API)
* `single`: can be set to true if `keys` contains only a single key to send a reply without the key wrapper (see the current `GET nodeinfo` vs. `nodeinfo` requests to see the difference). Defaults to `false`. Do we even need this? The only users of the old API are the old statuspage (which won't support the new API anyways) and piping the data to alfred.
* `compress`: obvious. Defaults to `false`?

jplitza:
I am getting the impression we are re-inventing (read-only) HTTP over UDP here.

JSON would have the advantage of being forward-compatible: We could add further flags in the future without breaking old applications. But it is also very verbose: What you posted is four times longer than "GETZ nodeinfo". As you mentioned, "single" isn't really required, so what advantage does JSON give us right now?

(also: did you abandon the idea of partitioning the answers?)

tcatm:
Well, let's start by looking at the current state. There are two types of requests:

**a)** `GET id [id...]` for requesting a bunch of records as compressed json
**b)** `id` for requesting a single record, uncompressed, raw json

These requests are used like this:

b) is used by the old status page just with the id set to "nodeinfo". We'll need to keep this around for backward compatibility a while.
b) is also used locally as a replacement for `collect.lua id`. This time, other ids like "statistics" and "neighbours" are also queried. This interface can be changed at any time so it shouldn't pose a problem.
a) was introduced in 2015.1 as a stable API. There may be users relying on it so we should keep it for a few releases, too.
a) is used by the new status-page. We could change that before releasing 2015.2.

If we were to introduce a new API in 2015.2, we still must support the old APIs for a while, giving us a total of three public APIs.

Therefore, I'd like to propose not changing anything in 2015.2 but to deprecate the public part of b) now and schedule its removal for 2015.2+1. A few releases later we could replace a) the same way (or transition to NetJSON rightaway).

NeoRaider:
[...]

Issue #496, opened 2015-09-13
gluon-0.7.2-x86-vmware does not get dhcp on wan interface


nmaas87:
As written here (https://forum.freifunk.net/t/bug-in-gluon-0-7-2-x86-vmware/7801), I got an 0.7.2. version of the Freifunk Trier VMWare Image (https://github.com/freifunktrier/firmware_store/blob/master/firmware/stable/factory/gluon-fftr-0.7.2-x86-vmware.vmdk) - which does, after inital setup - not get an dhcp lease on the wan (eth1) interface. I left everything in the configuration on the website on default, activated mesh_via_vpn and checked in the vpn key - which was done successfully. however, my appliance can't get an ip on the wan interface and does not connect to the internet. same thing with the image from ggrz. Any ideas?
Can reproduce the error on Workstation 11 as well as on VMWare ESXi 4.1 - with multiple gluon 0.7.2 Versions. Wan Interface, as descibred in the core config of gluon is correctly configured to be eth1, udhcpc eht1 does not give back an ip. Using an gluon 0.3.3 image from Freifunk Westpfalz (https://www.it-kl.eu/2015/08/gluon-x86-unter-vmware/ ) does work correctly in the same configuration and enviroment.

jplitza:
Could you verify the layer 2 connectivity of that interface, i.e. use something like `ping6 -I eth1 ff02::1` and see if your host responds? Also, more details about the network setup of your VMWare setup would probably be needed.
Also, what Gluon versions do those community version numbers correspond to?

nmaas87:
Hi there,

ping6 -I eth1 ff02::1does only give me an ping6: sendto: Operation not permitted.
Same for eth0. However, an ping6 on br-client, br-wan and bat0 give me back an answer.
The system also gives out IPv6 addresses on eth0, the client network - but none ipv4, due to the fact that it is not online.

Network:
Freifunk VM, eth0 (E1000) for client port to vlan 5
Freifunk VM, eth1 (E1000) for wan port to vlan 1
Internal pfSense, which gives LAN and dhcp on vlan 1
Internal pfSense, which takes WAN on vlan 2
VMWare ESXi 4.1, attached with trunk port to switch
Switch with trunk port to VMWare
Switch with vlan 2, WAN access is attached here
Switch with vlan 1, LAN port, Freifunk is attached here with Uplink
Switch with vlan 5, Client port of Freifunk

[...]

PR #2108, opened 2020-08-24
docs: fixes for v2020.2 release notes

rotanid:
we forgot issues marked as known issue in the last release's release notes, forgot to mention the issue with upgrades for ER-X and can remove an issue that seems to be obsolete starting with v2020.1

### DIFF
--- a/docs/releases/v2020.2.rst
+++ b/docs/releases/v2020.2.rst
@@ -173,8 +173,9 @@ Build system
 Known issues
 ------------
 
-* Out of memory situations with high client count on ath9k.
-  (`#1768 <https://github.com/freifunk-gluon/gluon/issues/1768>`_)
+* Upgrading EdgeRouter-X from versions before v2020.1.x may lead to a soft-bricked state due to bad blocks on the
+  NAND flash which the NAND driver before this release does not handle well.
+  (`#1937 <https://github.com/freifunk-gluon/gluon/issues/1937>`_)
 
 * The integration of the BATMAN_V routing algorithm is incomplete.
 
@@ -189,3 +190,9 @@ Known issues
   (`#94 <https://github.com/freifunk-gluon/gluon/issues/94>`_)
 
   Reducing the TX power in the Advanced Settings is recommended.
+
+* In configurations not using VXLAN, the MAC address of the WAN interface is modified even when Mesh-on-WAN is disabled
+  (`#496 <https://github.com/freifunk-gluon/gluon/issues/496>`_)
+
+  This may lead to issues in environments where a fixed MAC address is expected (like VMware when promiscuous mode is
+  disallowed).


docs/releases/v2020.2.rst
```diff
@@ -189,3 +190,13 @@ Known issues
   (`#94 <https://github.com/freifunk-gluon/gluon/issues/94>`_)
 
   Reducing the TX power in the Advanced Settings is recommended.
+
+* The MAC address of the WAN interface is modified even when Mesh-on-WAN is disabled
+  (`#496 <https://github.com/freifunk-gluon/gluon/issues/496>`_)
+
+  This may lead to issues in environments where a fixed MAC address is expected (like VMware when promiscuous mode is
+  disallowed).
```
blocktrron, line 198: This is fixed when using vxlan

docs/releases/v2020.2.rst
```diff
@@ -189,3 +190,13 @@ Known issues
...
+
+* Inconsistent respondd API (`#522 <https://github.com/freifunk-gluon/gluon/issues/522>`_)
+
+  The current API is inconsistent and will be replaced eventually. The old API will still be supported for a while.
```
blocktrron, line 202: We're dragging this with us since years. I'm not sure whether or not we want to continue doing so.

rotanid: but simply leaving it out from release notes while still having it marked and not discussing it is not the way to go

docs/releases/v2020.2.rst
rotanid: i will add this to this PR and the other PRs in the evening

mweinelt:
On 25.08.20 14:52, David Bauer wrote:
> @blocktrron commented on this pull request.
> 
> 
> 
>> +* The MAC address of the WAN interface is modified even when Mesh-on-WAN is disabled
> +  (`#496 <https://github.com/freifunk-gluon/gluon/issues/496>`_)
> +
> +  This may lead to issues in environments where a fixed MAC address is expected (like VMware when promiscuous mode is
> +  disallowed).
> 
> This is fixed when using vxlan
> 
>> +* Inconsistent respondd API (`#522 <https://github.com/freifunk-gluon/gluon/issues/522>`_)
> +
> +  The current API is inconsistent and will be replaced eventually. The old API will still be supported for a while.
> 
> We're dragging this with us since years. I'm not sure whether or not we want to continue doing so.
> 

Indeed, that remark isn't really helpful to anyone.

docs/releases/v2020.2.rst
mweinelt: As long as we don't have a plan for a replacement it makes no sense to carry that remark.

NeoRaider (approved):

PR #2111, opened 2020-08-28
Fix two feature handling bugs

NeoRaider:
This fixes two major bugs in the new feature definition handling, both leading to non-existent packages like `gluon-web-advanced` to be added to the package list when the corresponding feature flags are set in GLUON_FEATURES. In particular, it was more or less impossible to define a custom `features` file in a feed because of the second issue.

Reported by @T-X

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -110,5 +110,6 @@ files["package/features"] = {
 	read_globals = {
 		"_",
 		"feature",
+		"when",
 	},
 }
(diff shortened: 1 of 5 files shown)


T-X:
Seems to work fine here. Tested with the Freifunk Vogtland site and Gluon master branch on a GL.iNet GL-USB150 (ar71xx-generic).

blocktrron (approved):

PR #2113, opened 2020-08-31
ramips-mt7620: add support for Netgear EX3700/EX3800

ambassador86:
  - [x] webinterface
  - [x] other: nmrpflash
    - [x] must keep/forget configuration (if applicable)  *think `sysupgrade [-n]` or `firstboot`*
  - [x] must have working autoupdate *usually means: gluon profile name must match image name*
- [x] wps button must return device into config mode
- wired network
  - [x] must have correct port assignment (WAN)
- wifi (if applicable)
- led mapping
    - [x] should display config mode blink sequence (https://gluon.readthedocs.io/en/latest/features/configmode.html)

### DIFF
--- a/targets/ramips-mt7620
+++ b/targets/ramips-mt7620
@@ -20,6 +20,14 @@ device('gl-mt750', 'gl-mt750', {
 	factory = false,
 })
 
+-- Netgear
+
+device('netgear-ex3700', 'ex3700-ex3800', {
+	aliases = {'netgear-ex3800',
+		'netgear-ex3700-ex3800',
+	},
+	factory_ext = '.chk',
+})
 
 -- Nexx
 
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac
@@ -111,6 +111,7 @@ local primary_addrs = {
 			'c20i',
 			'c50',
 			'tplink,c2-v1',
+			'ex3700'
 		}},
 		{'x86'},
 	}},


targets/ramips-mt7620
```diff
@@ -20,6 +20,14 @@ device('gl-mt750', 'gl-mt750', {
 	factory = false,
 })
 
+-- Netgear
+
+device('netgear-ex3700', 'ex3700-ex3800',{
+    aliases = {
+               'netgear-ex3800',
+       },
```
rotanid, line 28: indentation error ;)

targets/ramips-mt7620
```diff
@@ -20,6 +20,14 @@ device('gl-mt750', 'gl-mt750', {
 	factory = false,
 })
 
+-- Netgear
+
+device('netgear-ex3700', 'ex3700-ex3800',{
```
rotanid, line 25: missing space before the opening curly bracket ;)

rotanid:
apart from the minor nitpicks i wrote, please check if this leads to images whose gluon profile name matches the image name
what about the primary mac matching the printed one?

ambassador86:
I already added another alias (which is not pushed yet), to ensure the imagename for the update, as the device itself will use the name ex3700-ex3800. The primary mac is not the printed one. It is off by one and the printed one matches the mac adress for the wan interface. I also tried to the the correct primary mac assigned.
Unfortunately I can only check this with an EX3800 at the moment because my EX3700 was DOA.

ambassador86:
The device is working now and the MAC address matches. Internally it does not distinguish between the EX3700 and the EX3800. Test device: http://map.chemnitz.freifunk.net/#!v:m;n:3c37867c5e6e (an EX3800). I also have an EX3700 now to test it.

targets/ramips-mt7620
```diff
@@ -20,6 +20,14 @@ device('gl-mt750', 'gl-mt750', {
 	factory = false,
 })
 
+-- Netgear
+
+device('netgear-ex3700', 'ex3700-ex3800', {
+	aliases = {'netgear-ex3800',
```
blocktrron, line 26: Add a newline here

blocktrron (approved):

blocktrron:
Did the outlined change myself. Merged, thanks!

Issue #1761, opened 2019-06-23
v2018.2.x wifi mesh link loss


2tata:
I recognize that wifi meshing functionality of gluon-v2018.2.1-13 (eed810aac1b0f6795622907b0de7dbc0fbfadc9d) seems partly to be broken.
I've update some devices from gluon-v2017.1.8-7 (e968a225be0efafbd2d4638d0bd530e2d4e841f3) to gluon-v2018.2.1-13 (eed810aac1b0f6795622907b0de7dbc0fbfadc9d) and the meshing with 11s seems broken to older firmware as well as to gluon-v2018.2.1-13 partly. Are there kind of problems with the wifi driver?

### Bug report

**What is the problem?**

The node [TataWilh210-SO(gluon-v2018.2.1-13)](https://map.ffnw.de/#!/en/map/98ded0889a7c) was able to see [ffnw-galaxy2(gluon-v2017.1.8-7)](https://map.ffnw.de/#!/en/map/10feed568d22) with nearly 80-90% but now its lower that 10% same for [TataWilh510-NW(gluon-v2018.2.1-13)](https://map.ffnw.de/#!/en/map/98ded0727ce8) (got bricked while trying to downgrade it from v2018.2.1-13 to v2017.1.8-7) to [OldenburgRaiffeisenLappan5(gluon-v2018.2.1-13)](https://map.ffnw.de/#!/en/map/802aa86e503e) but before the link wars almost 60-70% and now its gone.

A more extreme case both [OldenburgRaiffeisenMosle5(gluon-v2018.2.1-13)](https://map.ffnw.de/#!/en/map/802aa834090f) and [OldenburgRaiffeisenLappan5(gluon-v2018.2.1-13)](https://map.ffnw.de/#!/en/map/802aa86e503e) were able to see [Oldenburg-Lappan-kismet(gluon-v2017.1.8-7)](https://map.ffnw.de/#!/en/map/802aa8d5ceb5) with almost 100% bidirectional. Now its 96% - 0%. The zero (or 1%) in one direction is striking.

Affected devices were:
TP-Link TL-WR1043N/ND v1
Ubiquiti Loco M XW
TP-Link TL-WR741N/ND v4
Ubiquiti UniFi-AC-PRO
TP-Link CPE210 v1.1
TP-Link CPE510 v1.1

**Site Configuration:**

[...]

Issue #130, opened 2014-07-04
ath9k wifi stops working occasionally


lost-tty:
This issue tracks the sporadic loss of wireless functionality on ath9k based nodes running **Gluon 2014.2**. Before reporting new information from hung nodes, make sure the node is actually running a firmware based on this specific Gluon version.
# What to report

If you notice this particular problem on your node, please report at least the following information by posting a comment on this issue:
- Gluon version (tag or commit) the node is running.
- output of `cat /tmp/sysinfo/model`
- uptime
- contents of these files:
  - `/sys/kernel/debug/ieee80211/phy0/ath9k/reset`
  - `/sys/kernel/debug/ieee80211/phy0/ath9k/queues`
- Check the contents of `/sys/kernel/debug/ieee80211/phy0/ath9k/interrupt` twice with about 10 seconds inbetween. Report whether the `RXLP`, `RXHP` and `TX` values are increasing. 
- Check the last few lines of `dmesg`. Report anything mentioning `ath9k` or `phy0` with timestamps close to the event (compare with uptime). Do not post repeated messages.
# Symptoms

We are aware of the following indicators related to the loss of wifi functionality.
## Best Effort queue stopped

In the output of `cat /sys/kernel/debug/ieee80211/phy0/ath9k/queues` the `(BE)` entry will read `stopped: 1`. This seems to be a very strong indicator.

If possible, let your node continue running in this state for up to 12 hours, while checking regularly whether for the queue to be still stopped. In some cases this problem will resolve itself. Should your node recover on its own, please _edit_ your original report instead of adding a new comment.

If you can't wait, running `wifi` is enough to get your node working again. There's no need to reboot.
## Mesh partners lost

In some cases `batctl o|grep wlan0` will not show mesh partners anymore, yet the Best Effort queue is **not stopped**. If you notice this problem, please file a complete report.
# Misc.
- ath9k is instable since http://git.openwrt.org/?p=12.09/openwrt.git;a=commit;h=d29f21c84e2d8af8a5c18ceae80426c026ddac07
- http://kbu.freifunk.net/wiki/index.php?title=Ath9k-Probleme

rotanid:
as we (Freifunk Altdorf) and also Freifunk München (derkalle) are experiencing problems which seem to be closely related to this one i'd like to reopen this issue.
for a start i have a first report with the data you asked for in the starting post of this issue:

gluon version:
v2016.1
model:
TP-Link TL-WR1043N/ND v2
uptime:
[...]

Issue #1768, opened 2019-06-25
ath9k devices run out of memory above certain client count


mweinelt:
#### Welcome to "Debugging out of memory situations" 2019 Edition!

**What is the problem?**

Devices with ath9k radios seem to be quickly running out of memory as soon as they reach a certain client count.

Running `wifi` causes the memory to be temporarily freed again, but does not prevent the device from leaking.

Example: Archer C7 v2 looses ~68 MB of normally free RAM in roughly 20 minutes, then OOM reboots, multiple times

https://stats.darmstadt.freifunk.net/d/000000021/router-meshviewer-export?orgId=1&var-node=f4f26d707c53&refresh=15m

![c7v2-clients](https://user-images.githubusercontent.com/131599/60113371-ecb69c80-9771-11e9-899f-b3202ca1e3c6.png)
![c7v2-loadmemory](https://user-images.githubusercontent.com/131599/60113372-ecb69c80-9771-11e9-85e4-768b5bd49a57.png)
![c7v2-uptime](https://user-images.githubusercontent.com/131599/60113373-ed4f3300-9771-11e9-9306-2cb33da90c03.png)

Other affected nodes are:
- 1043N v3
  https://stats.darmstadt.freifunk.net/d/000000021/router-meshviewer-export?orgId=1&var-node=f4f26d3e9cac&refresh=15m
- CPE210 v1.1
  https://stats.darmstadt.freifunk.net/d/000000021/router-meshviewer-export?orgId=1&var-node=8416f96a33dc&refresh=15m

**What is the expected behaviour?**

Memory usage should increase at a reasonable rate per client

**Gluon Version:**

v2018.2.1-7-geed810aa

**Site Configuration:**

https://git.darmstadt.ccc.de/ffda/site/tree/v2018.2.x

**Custom patches:**

none

blocktrron:
We were experiencing this issue since last year, so probably everything based on LEDE 17.01 onwards is affected.

awlx:
Could it be related to this? 
https://forum.openwrt.org/t/page-table-leak-on-tp-link-td-w8980-with-openwrt-18-06-2/39356

CodeFetch:
A log would be great. This should really not happen...

mweinelt:
So far we have not seen anything meaningful in dmesg/logread. Errors/warnings in there are only symptomatic, as they happen late when little memory is left.

We have not found any issue in userspace and therefore continued looking into the kernel space.

[...]

Issue #93, opened 2014-04-07
2.4GHz WLAN dies after some time


jplitza:
In Bremen we have multiple reports of the (2.4 GHz) WLAN dying after some time (6 days, give or take a few).

This concerns both meshing and clients. I tried to debug the problem with clients:
`tcpdump -i br-client` on the node shows the DHCP requests and replies, while `tcpdump -i wlan0` only shows the requests, not the replies (and they never reach the client). ebtables are empty, when inserting a logging rule into each possible chain the output was

```
[1235303.240000] eb_brt_brt IN=bat0 OUT= MAC source = c6:b6:a5:52:3a:9b MAC dest = 98:0c:82:c2:ad:0a proto = 0x0800
[1235303.250000] eb_nat_pre IN=bat0 OUT= MAC source = c6:b6:a5:52:3a:9b MAC dest = 98:0c:82:c2:ad:0a proto = 0x0800
[1235303.260000] eb_flt_fwd IN=bat0 OUT=wlan0 MAC source = c6:b6:a5:52:3a:9b MAC dest = 98:0c:82:c2:ad:0a proto = 0x0800
[1235303.270000] eb_nat_pst IN= OUT=wlan0 MAC source = c6:b6:a5:52:3a:9b MAC dest = 98:0c:82:c2:ad:0a proto = 0x0800
[1235303.280000] eb_flt_inp IN=bat0 OUT= MAC source = c6:b6:a5:52:3a:9b MAC dest = 98:0c:82:c2:ad:0a proto = 0x0800
```

As you can see, the bridge found the correct OUT interface (wlan0) but the packet never goes out that interface.

To make things even more confusing, when using 5 GHz (wlan1), everything worked fine. The ebtables log was the same.

I didn't debug the meshing interface wlan0-1 any further, but the meshing with my other (2.4 GHz-only) node was broken at that time as well.

Executing `wifi` on the failing node fixed both the client connection and the meshing connection.

Other people reported similar issues. They didn't come up during the nightly phase, presumably because the nodes were rebooting regularly (nearly each night after an update).

Any ideas on how to debug this any further when it strikes again? Relevant logs?

PR #2114, opened 2020-09-05
WIP: patches/openwrt: add 0016-ath9k-check-for-deaf-rx-path-state.patch to…

2tata:
… fix 0% bug on meshlinks

This Merge Request is related to #1761  and #130 

I would like to test this patch in our community Nordwest Freifunk because currently we have approximately 500 affected nodes, who regularly lose there receive capability.

Patch: https://github.com/openwrt/openwrt/commit/f9022964cfa0a44984d2efa2dafe3452899acfdf#diff-aec5f1d1f9e40133d2d8abdbd0151dcf

### DIFF
--- a/patches/openwrt/0016-ath9k-check-for-deaf-rx-path-state.patch
+++ b/patches/openwrt/0016-ath9k-check-for-deaf-rx-path-state.patch
@@ -1,0 +1,126 @@
+From: Jan-Tarek Butt <tarek@ring0.de>
+Date: Sat, 5 Sep 2020 11:18:59 +0200
+Subject: ath9k: check for deaf rx path state
+
+Various chips occasionally run into a state where the tx path still
+appears to be working normally, but the rx path is deaf.
+
+There is no known register signature to check for this state explicitly,
+so use the lack of rx interrupts as an indicator.
+
+This detection is prone to false positives, since a device could also
+simply be in an environment where there are no frames on the air.
+However, in this case doing a reset should be harmless since it's
+obviously not interrupting any real activity. To avoid confusion, call
+the reset counters in this case "Rx path inactive" instead of something
+like "Rx path deaf", since it may not be an indication of a real
+hardware failure.
+
+Signed-off-by: Jan-Tarek Butt <tarek@ring0.de>
+
+diff --git a/package/kernel/mac80211/patches/ath/359-ath9k-check-for-deaf-rx-path-state.patch b/package/kernel/mac80211/patches/ath/359-ath9k-check-for-deaf-rx-path-state.patch
+new file mode 100644
+index 0000000000000000000000000000000000000000..347d06ec888c56102523ed40b38aa4b5083d4c9c
+--- /dev/null
++++ b/package/kernel/mac80211/patches/ath/359-ath9k-check-for-deaf-rx-path-state.patch
+@@ -0,0 +1,100 @@
++From: Felix Fietkau <nbd@nbd.name>
++Date: Wed, 25 Jan 2017 13:00:58 +0100
++Subject: [PATCH] ath9k: check for deaf rx path state
++
++Various chips occasionally run into a state where the tx path still
++appears to be working normally, but the rx path is deaf.
++
++There is no known register signature to check for this state explicitly,
++so use the lack of rx interrupts as an indicator.
++
++This detection is prone to false positives, since a device could also
++simply be in an environment where there are no frames on the air.
++However, in this case doing a reset should be harmless since it's
++obviously not interrupting any real activity. To avoid confusion, call
++the reset counters in this case "Rx path inactive" instead of something
++like "Rx path deaf", since it may not be an indication of a real
++hardware failure.
++
++Signed-off-by: Felix Fietkau <nbd@nbd.name>
++---
++
++--- a/drivers/net/wireless/ath/ath9k/ath9k.h
+++++ b/drivers/net/wireless/ath/ath9k/ath9k.h
++@@ -1027,6 +1027,7 @@ struct ath_softc {
++ 
++ 	u8 gtt_cnt;
++ 	u32 intrstatus;
+++	u32 rx_active;
++ 	u16 ps_flags; /* PS_* */
++ 	bool ps_enabled;
++ 	bool ps_idle;
++--- a/drivers/net/wireless/ath/ath9k/debug.c
+++++ b/drivers/net/wireless/ath/ath9k/debug.c
++@@ -763,6 +763,7 @@ static int read_file_reset(struct seq_fi
++ 		[RESET_TYPE_BEACON_STUCK] = "Stuck Beacon",
++ 		[RESET_TYPE_MCI] = "MCI Reset",
++ 		[RESET_TYPE_CALIBRATION] = "Calibration error",
+++		[RESET_TYPE_RX_INACTIVE] = "Rx path inactive",
++ 		[RESET_TX_DMA_ERROR] = "Tx DMA stop error",
++ 		[RESET_RX_DMA_ERROR] = "Rx DMA stop error",
++ 	};
++--- a/drivers/net/wireless/ath/ath9k/debug.h
+++++ b/drivers/net/wireless/ath/ath9k/debug.h
++@@ -50,6 +50,7 @@ enum ath_reset_type {
++ 	RESET_TYPE_BEACON_STUCK,
++ 	RESET_TYPE_MCI,
++ 	RESET_TYPE_CALIBRATION,
+++	RESET_TYPE_RX_INACTIVE,
++ 	RESET_TX_DMA_ERROR,
++ 	RESET_RX_DMA_ERROR,
++ 	__RESET_TYPE_MAX
++--- a/drivers/net/wireless/ath/ath9k/link.c
+++++ b/drivers/net/wireless/ath/ath9k/link.c
++@@ -53,13 +53,27 @@ reset:
++ 
++ }
++ 
+++static bool ath_rx_active_check(struct ath_softc *sc)
+++{
+++	if (sc->rx_active) {
+++		sc->rx_active = 0;
+++		return true;
+++	}
+++
+++	ath_dbg(ath9k_hw_common(sc->sc_ah), RESET,
+++		"rx path inactive, resetting the chip\n");
+++	ath9k_queue_reset(sc, RESET_TYPE_RX_INACTIVE);
+++	return false;
+++}
+++
++ void ath_hw_check_work(struct work_struct *work)
++ {
++ 	struct ath_softc *sc = container_of(work, struct ath_softc,
++ 					    hw_check_work.work);
++ 
++ 	if (!ath_hw_check(sc) ||
++-	    !ath_tx_complete_check(sc))
+++	    !ath_tx_complete_check(sc) ||
+++	    !ath_rx_active_check(sc))
++ 		return;
++ 
++ 	ieee80211_queue_delayed_work(sc->hw, &sc->hw_check_work,
++--- a/drivers/net/wireless/ath/ath9k/main.c
+++++ b/drivers/net/wireless/ath/ath9k/main.c
++@@ -269,6 +269,7 @@ static bool ath_complete_reset(struct at
++ 	}
++ 
++ 	sc->gtt_cnt = 0;
+++	sc->rx_active = 1;
++ 
++ 	ath9k_hw_set_interrupts(ah);
++ 	ath9k_hw_enable_interrupts(ah);
++@@ -452,6 +453,7 @@ void ath9k_tasklet(unsigned long data)
++ 			ath_rx_tasklet(sc, 0, true);
++ 
++ 		ath_rx_tasklet(sc, 0, false);
+++		sc->rx_active = 1;
++ 	}
++ 
++ 	if (status & ATH9K_INT_TX) {


T-X:
There were similar patches from Simon a while ago, before the one from Felix:

https://patchwork.kernel.org/patch/9433619/
https://patchwork.kernel.org/patch/9433621/

I would tend to start with the first one of these two. Because then with Grafana measuring TQ values on a per device basis we can learn more about this issue and which ath9k chips are affected by this particular hang.

And then later add the second and learn which ath9k chips get deaf even without returning a 0xdeadbeaf.

T-X:
Btw. if you suspect that such patches might help I would expect to see a gradual decrease in average TQ values. Or an increase of TQ values after rolling out an update. Because it would be one device after another over time turning deaf. Does anyone see such a pattern in their Grafana?

2tata:
I have applied this patch to one of my local routers "Harbard" [0] before I notice more or less daily stalls for hours, where the wifi functionality get lost. These affected mesh links but also clients for the freifunk STA and also for the private STA:

```
client0   ESSID: "nordwest.freifunk.net"
          Access Point: 12:7F:6B:73:39:E0
          Mode: Master  Channel: 6 (2.437 GHz)
          Tx-Power: 14 dBm  Link Quality: 36/70
          Signal: -74 dBm  Noise: -95 dBm
          Bit Rate: 39.0 MBit/s
          Encryption: none
          Type: nl80211  HW Mode(s): 802.11bgn
          Hardware: unknown [Generic MAC80211]
          TX power offset: unknown
          Frequency offset: unknown
          Supports VAPs: yes  PHY name: phy0

mesh0     ESSID: "ffnw-mesh_02:00:0a:12:e0:00"
          Access Point: 12:7F:6B:73:39:E1
          Mode: Mesh Point  Channel: 6 (2.437 GHz)
          Tx-Power: 14 dBm  Link Quality: 47/70
          Signal: -63 dBm  Noise: -95 dBm
          Bit Rate: 72.2 MBit/s
          Encryption: none
          Type: nl80211  HW Mode(s): 802.11bgn
          Hardware: unknown [Generic MAC80211]
          TX power offset: unknown
          Frequency offset: unknown
          Supports VAPs: yes  PHY name: phy0

owe0      ESSID: "owe.nordwest.freifunk.net"
          Access Point: 12:7F:6B:73:39:E2
          Mode: Master  Channel: 6 (2.437 GHz)
          Tx-Power: 14 dBm  Link Quality: unknown/70
          Signal: unknown  Noise: -95 dBm
          Bit Rate: unknown
          Encryption: WPA3 OWE (CCMP)
          Type: nl80211  HW Mode(s): 802.11bgn
          Hardware: unknown [Generic MAC80211]
          TX power offset: unknown
          Frequency offset: unknown
          Supports VAPs: yes  PHY name: phy0

wlan0-1   ESSID: "Harbard"
          Access Point: 12:7F:6B:73:39:E3
          Mode: Master  Channel: 6 (2.437 GHz)
          Tx-Power: 14 dBm  Link Quality: 62/70
          Signal: -48 dBm  Noise: -95 dBm
          Bit Rate: 117.0 MBit/s
          Encryption: WPA2 PSK (CCMP)
          Type: nl80211  HW Mode(s): 802.11bgn
          Hardware: unknown [Generic MAC80211]
          TX power offset: unknown
          Frequency offset: unknown
          Supports VAPs: yes  PHY name: phy0
```

@T-X  I don't have a Gravana where I can see that but I would expect the around 500 affected nodes with 0% links should disappear in our network.

[0] https://map.ffnw.de/#!/en/map/ec086ba4dd84

rotanid:
have you also seen the commit which removed this again?
https://github.com/openwrt/openwrt/commit/5f2a1ac59aa996cf749e27cea30aca4270a4ef9c

blocktrron:
Has anyone tested this and if so, what where your observations?

CodeFetch:
I wonder why only some communities have these issues... Please post the ath9k debug information when you have an affected device. Maybe we can find out the trigger this way or at least some reliable indicators.

rotanid:
@CodeFetch 
> I wonder why only some communities have these issues... 

i could ask: why do you assume that only some communities have these issues?

there were plenty of issues in the past which lead many communities (including ours) to integrate "workaround scripts" which recover nodes with such problems without anyone noticing it.
i definitely understand that solving a problem is better than integrating a workaround - as i have shown in the past in several cases where i reported issues or helped debugging - but in the end it is often necessary to put "working freifunk" above "instable freifunk waiting for a fix that may never come"

@2tata as blocktrron asked, have you deployed this patch in your community broadly already and can report about the results?

CodeFetch:
@rotanid 
> > I wonder why only some communities have these issues...
> 
> i could ask: why do you assume that only some communities have these issues?

Because our network is rock-solid at the moment and I've never experienced the deaf RX bug since 2015. I think we are overlooking something. My read is that special circumstances (e.g. like using tunneldigger) make the issue appear much more likely. Ath9k is timing critical... Maybe the deaf RX bug appears when there is too much contention in the kernel-space or something. What do the affected nodes/communities have in common? I'd rather like to know how devices get into the deaf state than how to detect it...

rotanid:
no tunneldigger here.

2tata:
The Patch seems not working. I will try the patch set @T-X linked here.

2tata:
how I can apply this patch set?

mweinelt:
They are against linux-wireless, so apply them against the OpenWrt kernel using quilt.

https://openwrt.org/docs/guide-developer/build-system/use-patches-with-buildsystem

T-X:
Another option, if you are not that used to quilt, there is also the option to use git. You can take a snapshot of the build dir (build_dir/target-mips_24kc_musl/linux-ar71xx_generic/backports-4.19.137-1/), git-ify it (git init; git add *; git commit). And then apply the patch to this snapshot. Then export it again with "git format-patch" and copy it to the OpenWrt patches directory for mac80211.

ecsv:
The less aggressive deaf and deadbeef patch can be found in https://github.com/FreifunkVogtland/gluon/tree/v2020.2.x-hwtest/patches/openwrt as 0018-ath9k-work-around-AR_CFG-0xdeadbeef-chip-hang.patch and 0019-ath9k-Reset-chip-on-potential-deaf-state.patch

rotanid:
@2tata  (or anyone else)
> @2tata as blocktrron asked, have you deployed this patch in your community broadly already and can report about the results?

especially with such a patch, we would need communities deploying this to a greater number of nodes to see if it helps and doesn't come with disadvantages, so we're waiting for volunteers and feedback here...

CodeFetch:
Maybe OWE is causing the problem? Who has got this problem in their community and what do they do different than those who don't have issues?

argonym:
I am runnig an OpenWrt-only (no Gluon) custom master build at home (only a few clients, not much traffic) for ~5 weeks.
It includes the original [patch](https://patchwork.kernel.org/project/linux-wireless/patch/20161117083614.19188-2-sven.eckelmann@open-mesh.com/) and Felix's patch (so the one included in this PR).
However, I modified both to only log the event and _not_ do the reset.

- Unfortunately, the issue itself has not appeared meanwhile, while before I had observed it maybe once in 1-2 weeks - might be pure chance. (I am just assuming that I have been suffering from the "deaf rx". No proof. When it happens, wifi clients observe around 50% packet loss, and new stations cannot associate anymore, until I reboot or restart the interface.)
- I do see the "rx path inactive, resetting the chip" from Felix's patch being logged maybe every other day on average, so it certainly does cause false-positives.
- I never saw the "RX deaf hang is detected. Schedule chip reset" from the original patch being logged. So at least this one doesn't seem to cause false positives.

I have not yet tried [Sven's modified patch](https://github.com/FreifunkVogtland/gluon/blob/v2020.2.x-hwtest/patches/openwrt/0020-ath9k-Reset-chip-on-potential-deaf-state.patch).

Does anybody know how to trigger the problem?

T-X:
I think there are multiple bugs causing issues. I hope I can somehow piece together what I tested a while ago.

The setup I used to (sort of) reproduce the issue every few days was a three nodes line topology setup. Where the outer nodes would just not or only barely see each other. So kind of like a typical hidden node setup. And then we had some iperf UDP traffic from node 1 over node 2 to node 3, using batman-adv.

At some point we would observe that TX queues got stuck. And that ar9003_hw_proc_txdesc() would report a lot of -EINPROGRESS here: https://elixir.bootlin.com/linux/latest/source/drivers/net/wireless/ath/ath9k/ar9003_mac.c#L363

There is this 512 elements "ts_ring" which ath9k shares with the "Queue Control Unit" of the Wifi chip. My guess is that these two somehow get out of sync or that it somehow gets holey at some point. A "iw scan" typically fixes this as it calls ath9k_queue_reset() which resets the wifi chip. One way to verify that you are in this situation is that a normal "ping6 fe80::.." with fixed neighbor entries would not work, but when adding a "-Q 24" to ping6, so that the VO (voice) WMM queue instead of the BE (best effort) WMM queue things would work perfectly fine, without any issues. Another point to check that you have this issue is /sys/kernel/debug/ieee80211/phy0/ath9k/queues. As @tcatm mentioned here: https://github.com/freifunk-gluon/gluon/issues/93#issuecomment-47979144. The "stopped" field was removed but you should see the qdepth stuck at >= 8, even when the link is idle. At which point ath9k will not accept any new frames for the specific WMM queue / QCU queue.

The MAC has 10 QCUs where four of them are dedicated to the four WMM queues. Citing from the datasheets, QCUs have the following two tasks: "1) Managing the Tx descriptor chain processing for frames pushed to the QCU from the CPU by traversing the linked list of Tx descriptors and transferring frame data from the host to the targeted DCU. 2) Managing the queue transmission policy to determine when the frame at the head of the queue should be marked as available for transmission."

However when resetting too often, like with a fixed 4 seconds interval, we also observed that things got stuck in other weird ways... At that point an "iw scan" / ath9k_queue_reset() would not help and only unloading and reloading the ath9k kernel module completely would fix it. At some point I even observed qdepth for the BE queue would show -1. Which even after going through the ath9k source code didn't make much sense to me, I couldn't figure out how that underun might have happened. So there might potentially also still be some miscounting bug in ath9k for the qdepth.

And yeah, the third issue is that the RX side is sometimes deaf as well.

---

Anyway, this issue is probably in some ways a duplicate of #93 and #130. So for anyone observing similar issues, I highly recommend reading through #130, especially.

CodeFetch:
@T-X Hm... I remember that we had this problem with bad temperature compensation. Could it possibly be the hardware which thinks that CSMA/CA is being triggered or something? What do you think is getting out of sync? So is it definitely the hardware that gets stuck? That the RX gets deaf would speak for that. In that case would Sven's patch work? Because I think Sven's patch is totally legit as the return value of the chip indicates a failure. The question is if we maybe cause this failure ourselves by feeding the chip with incorrect data.

CodeFetch:
@T-X I also remember that we observed this problem when the powersave features were enabled. Sven's patch resets the chip if it still has the deadbeef state after being waked up from powersave. My guess is that we misuse the powersave in some way and that disabling powersave might already stabilize it.

oszilloskop:
From the point of view from an user this PR patch and the patches from Simon are very helpful. I had the biggest improvement when all are included.

In my setups the issue occurs mainly with routers that have a very high traffic rate. I support a big setup with 17 nodes (WLAN and LAN Mesh) and approx. 60 connected clients (traffic rate up to 70 Mbit/s).
I tested the patches of this PR and of Simon in several combinations there.

CodeFetch:
> I had the biggest improvement when all are included.

@oszilloskop It would be nice if you could include logging mechanisms as @argonym did so that we can see at which paths your device really gets resetted.

oszilloskop:
> It would be nice if you could include logging mechanisms as @argonym
Sorry, but this is far beyond my capabilities.

CodeFetch:
@argonym Can you describe how you did the logging or show us your patches?

CodeFetch:
@T-X When looking through the commits in ath9k it seems that there are different reasons for the TX queue to stuck including calibration. But what really catches my eye is https://github.com/torvalds/linux/commit/2b4a66980217332d91ab1785e1750857d6d52bc8#diff-f1cbfae0c2d5da82493f2827245f85bd2a4aa5e181d7cacb17d46c737ad6c6d7
The bad handling of A-MPDU was also a problem with issue #1768. 

Do we really have a new problem here? #130 was a stuck-beacon issue. I remember that, but I don't believe we still have a miscounting bug in ath9k itself.

CodeFetch:
@T-X 
Hm... Should this can become negative?
https://github.com/torvalds/linux/blob/88bb507a74ea7d75fa49edd421eaa710a7d80598/drivers/net/wireless/ath/ath9k/xmit.c#L289

mweinelt:
Converted to draft, as this is not in a state that can be merged.

argonym:
Possibly related openwrt bug: https://bugs.openwrt.org/index.php?do=details&task_id=2176
A userspace-based workaround: https://github.com/libremesh/lime-packages/tree/master/packages/wifi-unstuck-wa

2tata:
On 24/06/2021 15:34, argonym wrote:
> Possibly related openwrt bug: https://bugs.openwrt.org/index.php?do=details&task_id=2176
> A userspace-based workaround: https://github.com/libremesh/lime-packages/tree/master/packages/wifi-unstuck-wa
> 

It seems not related to wifi but more on batman-adv.

ecsv:
> It seems not related to wifi but more on batman-adv.

Hu?

tecoboot:
What is the status of this patch?
Is the ATH_RX_STUCK_WAR_TIMEOUT   15000 correct? Or should it be 15 (seconds)?
In 5.4.142, I don's see the reset trigger, even if the RX is deaf. Maybe this patch is broken, between 4.19 and 5.4.

2tata:
On 19/08/2021 22:28, Teco Boot wrote:
> What is the status of this patch?
> Is the ATH_RX_STUCK_WAR_TIMEOUT   15000 correct? Or should it be 15 (seconds)?
> In 5.4.142, I don's see the reset trigger, even if the RX is deaf. Maybe this patch is broken, between 4.19 and 5.4.
> 
Its seems not an ath bug at all.

mabe more a batman-adv bug?

Lg
Tarek

ecsv:
> On 19/08/2021 22:28, Teco Boot wrote: What is the status of this patch? Is the ATH_RX_STUCK_WAR_TIMEOUT 15000 correct? Or should it be 15 (seconds)? In 5.4.142, I don's see the reset trigger, even if the RX is deaf. Maybe this patch is broken, between 4.19 and 5.4.
> Its seems not an ath bug at all. mabe more a batman-adv bug? Lg Tarek

Again?

The ath9k HW gets stuck - even when no batman-adv is run over it. So why is batman-adv to blame here?

tecoboot:
> On 19/08/2021 22:28, Teco Boot wrote: What is the status of this patch? Is the ATH_RX_STUCK_WAR_TIMEOUT 15000 correct? Or should it be 15 (seconds)? In 5.4.142, I don's see the reset trigger, even if the RX is deaf. Maybe this patch is broken, between 4.19 and 5.4.
> Its seems not an ath bug at all. mabe more a batman-adv bug? Lg Tarek

It has nothing to do with batman-adv.

ecsv:
> What is the status of this patch?
> Is the ATH_RX_STUCK_WAR_TIMEOUT 15000 correct? Or should it be 15 (seconds)?
> In 5.4.142, I don's see the reset trigger, even if the RX is deaf. Maybe this patch is broken, between 4.19 and 5.4.

You are talking about https://github.com/freifunk-gluon/gluon/commit/fa0740cca4683f90bbf05157fd80109d2c40aa84 ? The code itself would need at least 15000 x ATH_HW_CHECK_POLL_INT (1000msecs) to detect a stuck state. Which should be 25 minutes. And it doesn't match the commit description - so yeah, you might be on to something. Maybe @blocktrron can say more about it.

But the PR here is about another version which only used 1s of RX inactivity.

tecoboot:
> You are talking about [fa0740c](https://github.com/freifunk-gluon/gluon/commit/fa0740cca4683f90bbf05157fd80109d2c40aa84) ?
Yes. Commit:
```
| It triggers a chip reset when there is no RX activity for more than 15 seconds.
```
15 seconds != 25 minutes

My question is: why isn't sc->rx_active decremented anymore, since 5.4.142 (or far before...)?
@blocktrron?

blocktrron:
@tecoboot which patch are you using and with which gluon versions? Did the patch work for you in the past and up to which base?

The timeout claim for my commit is an oversight which needs to be fixed.

tecoboot:
> @tecoboot which patch are you using and with which gluon versions? 

Not gluon. I'm cherrypicking...
I've my own debian build with self-compiled kernel.
I use mesh & olsrd, now with wpa_supplicant SAE .
If I'm stuck, I'll test on openwrt and report findings.

>Did the patch work for you in the past and up to which base?

I'm facing problems with update, from kernel 4.14 (with authsae) to 5.4. (wpa_supplicant).
4.19 with wpa_supplicant has the problems. But here, (adjusted 15s) patch  [fa0740c](https://github.com/freifunk-gluon/gluon/commit/fa0740cca4683f90bbf05157fd80109d2c40aa84) seems to work, it shows the countdown.
My patch has countdown debug code, just for learning what is happening.

> The timeout claim for my commit is an oversight which needs to be fixed.
OK :-)

Is Felix Fietkau @nbd168 involved? He came up with this patch.

2tata:
On 20/08/2021 08:19, Sven Eckelmann wrote:
>> On 19/08/2021 22:28, Teco Boot wrote: What is the status of this patch? Is the ATH_RX_STUCK_WAR_TIMEOUT 15000 correct? Or should it be 15 (seconds)? In 5.4.142, I don's see the reset trigger, even if the RX is deaf. Maybe this patch is broken, between 4.19 and 5.4.
>> Its seems not an ath bug at all. mabe more a batman-adv bug? Lg Tarek
> 
> Again?

sorry, didn't saw the batman-adv relation post before.

I just notice the 0 link quality issue also on LAN links where no ath
driver is working at all.

Lg
Tarek

tecoboot:
> sorry, didn't saw the batman-adv relation post before. I just notice the 0 link quality issue also on LAN links where no ath driver is working at all. Lg Tarek

Then open another bug. 
Here, we discuss 0016-ath9k-check-for-deaf-rx-path-state.patch

2tata:
> > sorry, didn't saw the batman-adv relation post before. I just notice the 0 link quality issue also on LAN links where no ath driver is working at all. Lg Tarek
> 
> Then open another bug.
> Here, we discuss 0016-ath9k-check-for-deaf-rx-path-state.patch

As you may notice this Merge Request is related to the 0% bug issue. That the reason why I tried thous patches. 
OK, so I'll close this issue, because the 0% bug is not or not only wifi related and did not get fixed by this patches.

PR #2115, opened 2020-09-05
treewide: rename local_client zone

blocktrron:
This renames the local_client zone to loc_client, as local_clint exceeds
the maximum zone length allowed for firewall3, which is 11 bytes.

This worked previously due to firewall3 using unsafe string operations.
Now creation of the chain fails (latest OpenWrt master).

### DIFF
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/300-firewall-rules
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/300-firewall-rules
@@ -18,7 +18,7 @@ end
 
 uci:foreach('firewall', 'zone', reject_input_on_wan)
 
-for _, zone in ipairs({'mesh', 'local_client', 'wired_mesh'}) do
+for _, zone in ipairs({'mesh', 'loc_client', 'wired_mesh'}) do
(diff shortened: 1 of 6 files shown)


package/gluon-core/luasrc/lib/gluon/upgrade/300-firewall-rules
```diff
@@ -57,8 +57,16 @@ for _, zone in ipairs({'mesh', 'local_client', 'wired_mesh'}) do
...
 
-uci:section('firewall', 'rule', 'local_client_ICMPv4_in', {
-	src = 'local_client',
+-- ToDo Remove in v2022.x
+uci:delete('firewall', 'local_client_ICMPv6_in')
+local local_client_networks = uci:get_list('firewall', 'local_client', 'network')
+if local_client_networks ~= nil then
+	uci:set_list('firewall', 'loc_client', 'network', local_client_networks)
+	uci:delete('firewall', 'local_client')
+end
```
NeoRaider, line 66: This part can be removed. `network` is completely rewritten on each update (see line 25 of this file), so we don't need to migrate anything.

PR #2116, opened 2020-09-07
actions: add editorconfig checker

mweinelt:
Instead of pointing out those issues manually and finding them by chance, let's run the `editorconfig-checker`.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -35,6 +35,9 @@ indent_size = 2
 
 [*.lua]
 
+[{Makefile,*.mk}]
+indent_style = unset
+
 [*.md]
 indent_style = space
 indent_size = 4
--- a/.editorconfig
+++ b/.editorconfig
@@ -61,3 +61,7 @@ indent_size = 2
 [CMakeLists.txt]
 indent_style = space
 indent_size = 2
+
+[{docs,contrib/ci}/*site*/**/*.conf]
+indent_style = space
+indent_size = 2
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -30,3 +30,25 @@ jobs:
         run: ln -s ./docs/site-example ./site
       - name: Lint shell code
         run: make lint-sh
+
+  editorconfig:
+    name: Editorconfig
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Install Dependencies
+        run: sudo apt install curl tar
+      - name: Install editorconfig-checker
+        env:
+          VERSION: 2.6.0
+          OS: linux
+          ARCH: amd64
+        run: |
+          curl -O -L -C - https://github.com/editorconfig-checker/editorconfig-checker/releases/download/$VERSION/ec-$OS-$ARCH.tar.gz
+          tar xzf ec-$OS-$ARCH.tar.gz
+          sudo mv ./bin/ec-$OS-$ARCH /usr/bin/editorconfig-checker
+          sudo chmod +x /usr/bin/editorconfig-checker
+      - name: Install example site
+        run: ln -s ./docs/site-example ./site
+      - name: Lint editorconfig
+        run: make lint-editorconfig
--- a/scripts/lint-editorconfig.sh
+++ b/scripts/lint-editorconfig.sh
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+set -e
+
+editorconfig-checker .github contrib docs package scripts targets tests ./*.* .luacheckrc .editorconfig
--- a/Makefile
+++ b/Makefile
@@ -151,7 +151,10 @@ list-targets: FORCE
 		echo "$$target"
 	done
 
-lint: lint-lua lint-sh
+lint: lint-editorconfig lint-lua lint-sh
+
+lint-editorconfig: FORCE
+	@scripts/lint-editorconfig.sh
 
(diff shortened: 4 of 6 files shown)


mweinelt:
Okay, the linter runs. Now I need input on how to standardize several file types. Particularly
- `*{.h,.c}`
- `Makefile`
- `CMakeLists.txt`

Also we do have a healthy amount of mixed indentation in `.rst` files.

I've reported two issues upstream with the checker.
- We need to be able to skip certain paths like `docs/_build` or `openwrt/` https://github.com/editorconfig-checker/editorconfig-checker/issues/122
- The checker tries to replace the cwd inside the absolute path to print a relative one. If I run the check from `/gluon` it somehow formats `package/gluon.mk` as `package..mk`. https://github.com/editorconfig-checker/editorconfig-checker/issues/121

mweinelt:
We have `*.c` files that include a LICENSE header that introduces mixed (space) indentation into files that otherwise use tabs. E.g. `package/libgluonutil/src/libgluonutil.c`. We could use the occasion to migrate to SPDX identifiers, no? @NeoRaider

blocktrron:
+1 for migrating to SPDX headers

AiyionPrime:
Talked to @mweinelt this afternoon, I'll be kicking this PR for a while now.

For starters I rebased it to master, fixed several problems the linter called out,
but left others out for now, so the action can run through.

`make lint-editorconfig` runs with exit code `0` on my machine, but this PR is obviously still a work in progress. I'll be extracting commits from it and open them as seperate PR's to make this easier to review.

On thing I'd like to propose, is marking file extensions as explicitly rule-less like I did with `.pl` in fdb86ef46f2c5c64bd6e3b3b556f3d24ff500d8d to mark that we thought about it.
It will still be covered by the global setting but might help contributors to see, that their fileextension is already been taken into account.

Anyway. I hope this will get @mweinelt's work into master and hopefully end most of our indentation nitpicks.

AiyionPrime:
bumped ec to `2.6.0`

.editorconfig
```diff
@@ -23,6 +43,10 @@ indent_size = 2
 indent_style = space
 indent_size = 2
 
+# FIXME this needs to be gone eventually
```
AiyionPrime, line 46: If I understand it correctly, the Makefile syntax requires tabs and the precompiler breaks on lines when they do contain tabs, but spaces are fine?

AiyionPrime:
This is what I'm currently at. Got rid of every exception but the ones for Makefiles (#2749).

Shell files (#2743), html files (#2746) and c as well as header files (both #2747) do work properly.

Most of lua as well as js, rst, dia, Dockerfile and templates have already been merged.

AiyionPrime:
I think this is roughly it. I'd opt for leaving makefiles as they are for now or use some more sophisticated tool to lint it.

Once #2747 and #2751 are merged we can go on with the remaining few commits in here.

AiyionPrime:
Now solely blocked by #2751.

To verify this PR is working, checkout the branch, introduce an error or omit the last commit, that fixed spaces in ramips:

```console
$ git checkout lint-editorconfig
$ make lint-editorconfig
$ echo $?
0
$ git checkout HEAD~
$ make lint-editorconfig
targets/ramips-mt7621:
	69: Wrong indentation type(spaces instead of tabs)

1 errors found
echo $?
2
```

AiyionPrime:
rebased and unblocked, this is ready for review

AiyionPrime (approved):

.github/workflows/lint.yml
```diff
@@ -30,3 +30,25 @@ jobs:
...
+      - name: Install editorconfig-checker
+        env:
+          VERSION: 2.6.0
+          OS: linux
+          ARCH: amd64
+        run: |
+          curl -O -L -C - https://github.com/editorconfig-checker/editorconfig-checker/releases/download/$VERSION/ec-$OS-$ARCH.tar.gz
+          tar xzf ec-$OS-$ARCH.tar.gz
+          sudo mv ./bin/ec-$OS-$ARCH /usr/bin/ec
+          sudo chmod +x /usr/bin/ec
```
mweinelt, line 50:
```suggestion
          sudo mv ./bin/ec-$OS-$ARCH /usr/bin/editorconfig-checker
          sudo chmod +x /usr/bin/editorconfig-checker
```

At least that's the name for `ec` in my distro.

scripts/lint-editorconfig.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+set -e
+
+ec .github contrib docs package scripts targets tests ./*.* .luacheckrc .editorconfig
```
mweinelt, line 5:
```suggestion
editorconfig-checker .github contrib docs package scripts targets tests ./*.* .luacheckrc .editorconfig
```

.github/workflows/lint.yml
AiyionPrime: resolved in c93eb80bbe44f290d2506d14cc7d47b5f15fa82c and 48b70a9ea5fc37da9accae81f20d1f624ab6e8e6

scripts/lint-editorconfig.sh
AiyionPrime: resolved in c93eb80bbe44f290d2506d14cc7d47b5f15fa82c and 48b70a9ea5fc37da9accae81f20d1f624ab6e8e6

PR #2118, opened 2020-09-15
ar71xx-generic: remove USB package for builds of d-link-dir-825-rev-b1

rotanid:
without removing the USB packages i couldn't add any custom package to this device, i'm not sure where we draw the line here so i'm just proposing this change, not directly committing it

EDIT: build tested now

### DIFF
--- a/targets/ar71xx-generic
+++ b/targets/ar71xx-generic
@@ -101,6 +101,7 @@ device('d-link-dir-825-rev-b1', 'dir-825-b1', {
 	profile = 'DIR825B1',
 	factory = false,
 	class = 'tiny', -- Only 6M of usable Firmware space
+	packages = { '-kmod-usb-core', '-kmod-usb2', '-kmod-usb-ledtrig-usbport' },
 })
 
 


targets/ar71xx-generic
```diff
@@ -101,6 +101,7 @@ device('d-link-dir-825-rev-b1', 'dir-825-b1', {
 	profile = 'DIR825B1',
 	factory = false,
 	class = 'tiny', -- Only 6M of usable Firmware space
+	packages = { -kmod-usb-core -kmod-usb2 -kmod-usb-ledtrig-usbport },
```
blocktrron, line 104: `packages = { '-kmod-usb-core', '-kmod-usb2', '-kmod-usb-ledtrig-usbport' },`

blocktrron (changes requested):

targets/ar71xx-generic
rotanid: thanks, marked it as draft because not build-tested ;)
fixed.

blocktrron:
If I recall correctly, you've included multiple USB packages via your site, is this correct? Are the USB packages from OpenWrt really the issue here?

If so, the better option from my POV is to exclude USB packages for all devices with the tiny flag. This can then be overridden by the communities site. The USB packages alone do not support much use cases (at least none i can think of).

rotanid:
not sure why, but i couldn't reproduce it.
therefore giving up the upstreaming attempt and leaving in place a site-local solution (we don't have a device of this type anyway in our community)

PR #2119, opened 2020-09-15
clean up old cleanup code lines

rotanid:
this PR removes parts of two files, which were marked as "can be removed" a few years ago, one of which even wasn't part of any Gluon release.
also, it removes a single file marked as deletable a while ago.

### DIFF
--- a/package/gluon-client-bridge/luasrc/lib/gluon/upgrade/300-gluon-client-bridge-network
+++ b/package/gluon-client-bridge/luasrc/lib/gluon/upgrade/300-gluon-client-bridge-network
@@ -40,9 +40,6 @@ uci:section('network', 'interface', 'client', {
 
 uci:save('network')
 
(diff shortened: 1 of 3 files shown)


blocktrron:
What a coincidence - I've prepared similar patches just days ago :smile: .

Have a look at https://github.com/blocktrron/gluon/commits/remove-old-migrations , there are a bit more paths to clean up. The ones already in your PR look good.

rotanid:
@blocktrron :D you  have in addition the file package/gluon-core/luasrc/lib/gluon/upgrade/800-migrate-batadv  and the gluon users stuff.
i think these should be separate commits, not like in your branch.
feel free to add these to my PR here, you can push to this branch.

as we don't really support upgrades from before v2018.2 afterwards, we should somehow make sure to announce this very clear at least.
best would be to make it impossible to do such an upgrade, but i don't think this is possible...

blocktrron (approved):
Removing more stuff can be done afterwards. Content looks good as is.

PR #2123, opened 2020-09-28
Create 0027-kernel-add-cpe210-v320-support.patch

Parabol1337:


blocktrron:
I assume this is a mistake? The linked commit does not seem to add CPE510 support but instead add a MTD chip definition.

Anyways, I'm closing this, as discussion can be followed up in #2124

PR #2124, opened 2020-09-28
Add Support for CPE210 V3.20

Parabol1337:


blocktrron:
Is this a backport for an upstream commit? Gluon tries to keep as close as possible to upstream OpenWrt, thus it is preferred to not include patches which are not backports

SmithChart:
I had a chat with @Parabol1337 at our Freifunk meetup. As far as I understand he wants to backport these two patches from recent OpenWRT to v2019.1.x:

https://github.com/openwrt/openwrt/commit/4a2380a1e778a8f8e0bfb0a00f2996ed0aab58d8
https://github.com/openwrt/openwrt/commit/359f5e539036db4f7ac69a6d1c3fb7fe70266ffd

These two patches should be the same changes as #2123 and #2124 (this PR). I will test these patches on v2019.1.x and get back to this PR.

@blocktrron: What is the preferred way to add these backports to Gluon so that you can still track where they are from?

blocktrron:
@SmithChart Picking both with the original upstream commit name / message is fine.

Parabol1337:
new pull 
V2019.1.x/cpe210 3.20 #2136

PR #2128, opened 2020-10-06
autoupdater: document behaviour if manifest DATE is in the future

tokudan:
We were surprised when we saw most nodes update overnight to a new firmware that was supposed to be released slowly over a 7 day window in the future, so I believe that the behaviour should be documented.

I do not see any obvious way to build the documentation, so I hope the output is fine, but I was not able to verify it.

### DIFF
--- a/docs/features/autoupdater.rst
+++ b/docs/features/autoupdater.rst
@@ -30,6 +30,27 @@ in ``site.mk``, care must be taken to pass the same ``GLUON_RELEASE`` to ``make 
 as otherwise the generated manifest will be incomplete.
 
 
+Manifest format
+------------------------
+
+The manifest starts with a short header, followed by the list of firmwares and signatures.
+The header contains the following information:
+
+.. code-block:: sh
+
+    BRANCH=stable
+    DATE=2020-10-07 00:00:00+02:00
+    PRIORITY=7
+
+- ``BRANCH`` is the autoupdater branch name that needs to match the nodes configuration.
+- ``DATE`` specifies when the time period for the update begins. Nodes will do their regular update during a random minute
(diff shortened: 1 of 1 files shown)


mweinelt:
Please check the netlify checks, they fail for some reason.

```
8:41:15 PM: /opt/build/repo/docs/features/autoupdater.rst:38: WARNING: Unexpected indentation.
```

tokudan:
I added an empty line to match the code block in Automated nightly builds, but if that doesn't help I have no idea what's the issue.

docs/features/autoupdater.rst
```diff
@@ -30,6 +30,27 @@ in ``site.mk``, care must be taken to pass the same ``GLUON_RELEASE`` to ``make
...
+.. code-block:: sh
+
+    BRANCH=stable
+    DATE=2020-10-07 00:00:00+02:00
+    PRIORITY=7
+
+- ``BRANCH`` is the autoupdater branch name that needs to match the nodes configuration.
+- ``DATE`` specifies when the update time period for the update begins. Nodes might not always have a working NTP
+  configuration, which is why a fallback mechanism exists, that automatically updates if ``DATE`` is at least 24h
+  in the past.
```
rotanid, line 48: i'm not sure if this is completely correct or might be misleading, as it doesn't mention that regular updates only happen between 04:00 and 04:59 , see code comment here:
https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-autoupdater/luasrc/lib/gluon/upgrade/500-autoupdater#L54-L57

mweinelt: Please go ahead and make a suggestion.

blocktrron (approved):

PR #2130, opened 2020-10-08
ath79: migrate OCEDO Koala from ar71xx

blocktrron:
This provides the basic migration procedures necessary for most boards to be ported from ar71xx to ath79.

Handling of the WiFi path migration is handled upstream. We handle recreation of LED configuration, which will not be preserved.

Additionally, there's already a the necessary migration for devices where the GMAC iterator is swapped for ath79 when compared to ar71xx. This is not the case for the board in question.

### DIFF
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/002-ath79-migration
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/002-ath79-migration
@@ -10,6 +10,7 @@ local migration_name = 'ar71xx-ath79'
 
 local migration_target_config = {
 	-- {DEVICE, ETH_SWAP}
+	{'ocedo,koala', false}
 }
 
 -- Create a list for use with platform matching
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/002-ath79-migration
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/002-ath79-migration
@@ -1,0 +1,98 @@
+#!/usr/bin/lua
+
+local platform = require 'gluon.platform'
+local sysconfig = require 'gluon.sysconfig'
+local migration = require 'gluon.migration'
+
+local uci = require('simple-uci').cursor()
+
+local migration_name = 'ar71xx-ath79'
+
+local migration_target_config = {
+	-- {DEVICE, ETH_SWAP}
+}
+
+-- Create a list for use with platform matching
+local migration_targets = {}
+for _, v in ipairs(migration_target_config) do
+	table.insert(migration_targets, v[0])
+end
+
+-- Check if device is a migration target
+if not platform.match('ath79', 'generic', migration_targets) then
+    return
+end
+
+-- Check if this is an initial installation
+if not sysconfig.gluon_version then
+    os.execute('touch /lib/gluon/core/migrations/ar71xx-ath79')
+    return
+end
+
+-- Check if migration has already been done
+if migration.is_migration_done(migration_name) then
+    return
+end
+
+-- Reinit board.json
+migration.reinit_board_json()
+
+migration.reinit_led_config(uci)
+
+local function sysconfig_ifname_swap(input_ifnames)
+	local output_ifnames = {}
+	for v in input_ifnames:gmatch('%S+') do
+		if string.match(v, 'eth0') then
+			table.insert(output_ifnames, v:gsub('eth0', 'eth1'))
+		elseif string.match(v, 'eth1') then
+			table.insert(output_ifnames, v:gsub('eth1', 'eth0'))
+		end
+	end
+
+	return table.concat(output_ifnames, ' ')
+end
+
+-- For devices with embedded FE switch, eth0 and eth1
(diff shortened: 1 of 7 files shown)


package/gluon-core/luasrc/lib/gluon/upgrade/002-ath79-migration
```diff
@@ -0,0 +1,63 @@
...
+
+local unistd = require 'posix.unistd'
+local uci = require('simple-uci').cursor()
+
+
+if not platform.match('ath79') then
+    return
+end
+
+if platform.match('ath79', 'generic', {'devolo,dvl1200e', 'devolo,dvl1200i',
```
mweinelt, line 14: Add a comment why these are excluded.

package/gluon-core/luasrc/lib/gluon/upgrade/002-ath79-migration
```diff
@@ -0,0 +1,63 @@
...
+os.execute('board_detect')
+
+-- Remove LED configuration
+uci:foreach('system', 'rssid', function(config)
+    uci:delete('system', config['.name'])
+end)
+uci:foreach('system', 'led', function(config)
+    uci:delete('system', config['.name'])
+end)
+uci:commit('led')
```
mweinelt, line 41:
```suggestion
uci:commit('system')
```

package/gluon-core/luasrc/lib/gluon/upgrade/002-ath79-migration
```diff
@@ -0,0 +1,63 @@
...
+
+-- Create LED configuration
+os.execute('/lib/gluon/core/migration/ath79-create-config.sh')
+
+local function swap_eth0_eth1()
```
mweinelt, line 46: unused?

mweinelt (commented):
Thanks for getting this started, I have a few remarks.

package/gluon-core/luasrc/lib/gluon/upgrade/002-ath79-migration
blocktrron: Yes, It's not needed for this specific board, but included for future use.

patches/openwrt/0005-base-files-allow-sourcing-of-config_generate.patch
```diff
@@ -0,0 +1,17 @@
...
+diff --git a/package/base-files/files/bin/config_generate b/package/base-files/files/bin/config_generate
+index 0dca5549e8a91749217d959b543a16f1258b0af0..cc838d25e8acfa7cc0652c9801de795e497d3033 100755
+--- a/package/base-files/files/bin/config_generate
++++ b/package/base-files/files/bin/config_generate
+@@ -5,7 +5,7 @@ CFG=/etc/board.json
+ . /usr/share/libubox/jshn.sh
+ 
+ [ -s $CFG ] || /bin/board_detect || exit 1
+-[ -s /etc/config/network -a -s /etc/config/system ] && exit 0
++[ "$1" != "source-only" ] && [ -s /etc/config/network -a -s /etc/config/system ] && exit 0
```
NeoRaider, line 14: How about splitting the script into a sourceable helper library and the actual executable script?

Also, "source-only" doesn't seem to be set anywhere?

blocktrron: If we pull this off upstream we can do that. Rebasing this split downstream is PITA and that's why I've switched to sourcing config_generate literally minutes ago (It's still WIP, so It's currently broken anyways).

NeoRaider: How to perform the patch at runtime rather than compile time:
```sh
cat /lib/functions/config-generate.sh | sed 's/Foo/Bar/' | (
. /dev/stdin
# Add code using the helpers here
)
```
(very cursed, don't actually to this)

blocktrron: I first saw your tweet and immediately knew it's context :smile:

package/gluon-core/luasrc/usr/lib/lua/gluon/migration.lua
```diff
@@ -0,0 +1,37 @@
...
+end
+
+function M.reinit_led_config(uci)
+	-- Remove LED configuration
+	uci:foreach('system', 'rssid', function(config)
+		uci:delete('system', config['.name'])
+	end)
+	uci:foreach('system', 'led', function(config)
+		uci:delete('system', config['.name'])
+	end)
```
NeoRaider, line 25: This should use `uci:delete_all()` - IIRC, deleting entries in a `uci:foreach()` is always wrong and can lead to all kinds of funky results, including segfaults (which should really be fixed, if it isn't by now...)

blocktrron:
CLosing as @NeoRaider will propose a solution for intent-based runtime config migration.

PR #2131, opened 2020-10-11
build: improve refreshing patches

blocktrron:
This PR adds a new `refresh-patches` step to the project Makefile which allows to apply and update all patches to feeds and OpenWrt without installing feed packages to the OpenWrt buildsystem.

This reduces the time necessary for refreshing patches when updating feeds or the OpenWrt base.

This PR will also print the commit title for every patch created when updating patches. This was previously only done when applying    patches.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,6 +90,13 @@ update-patches: FORCE
 	scripts/update-patches.sh
 	scripts/patch.sh
 
+refresh-patches: FORCE
+	@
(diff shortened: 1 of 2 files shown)


NeoRaider (approved):

mweinelt:
Unfortunately no update for the docs at `docs/dev/basics.rst`.

PR #2134, opened 2020-10-16
disable overwriting wan ifname on sysupdate

Parabol1337:
When you change the wan ifname and update gluon, the update will override this config.

### DIFF
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/110-network
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/110-network
@@ -5,13 +5,16 @@ local sysconfig = require 'gluon.sysconfig'
 
 
 uci:section('network', 'interface', 'wan', {
-	ifname = sysconfig.wan_ifname,
 	type = 'bridge',
 	igmp_snooping = true,
 	multicast_querier = false,
 	peerdns = false,
 	auto = true,
 })
+
+if not uci:get('network', 'wan', 'ifname') then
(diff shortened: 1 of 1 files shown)


T-X:
I think this is intended, that the wan ifname in UCI is updated on sysupgrade. I think if you swap the contents of these files it should be persistent and sysupgrade safe:

* /lib/gluon/core/sysconfig/lan_ifname
* /lib/gluon/core/sysconfig/wan_ifname

Currently I can only find these two files documented here though:

https://gluon.readthedocs.io/en/latest/dev/hardware.html?highlight=wan_ifname#package-adjustments

That way things should also stay consistent and work for enabling and disabling mesh-on-lan, as documented here:

https://gluon.readthedocs.io/en/latest/features/wired-mesh.html?highlight=lan_ifname#wired-mesh-mesh-on-wan-lan

NeoRaider:
Thanks for your contribution, but I don't think this change is desirable. The currently recommended way to change the interface name is to edit `/lib/gluon/core/sysconfig/wan_ifname` and then run `gluon-reconfigure`.

Note that modifying the generated network configuration at all is not officially supported and might break with a future update (as a major rework of Gluon's network setup is planned to happen at some point), but it has worked for long enough that such a change would be noted prominently in Gluon's release notes.

A note for future contributions: Unless fixing an issue specific to an older Gluon version or backporting a fix, all pull requests must target the `master` branch.

PR #2140, opened 2020-10-20
github: add workflow for checking patch status

blocktrron:
Add a workflow to verify all patches to OpenWrt and the packages feeds
are refreshed.

If they are not, the workflow fails.


mweinelt (approved):
LGTM. Thanks!

PR #2141, opened 2020-10-21
package/libgluonutil: add function get_domains

2tata:
refactor duplicated code.

### DIFF
--- a/package/gluon-config-mode-domain-select/luasrc/lib/gluon/config-mode/wizard/0200-domain-select.lua
+++ b/package/gluon-config-mode-domain-select/luasrc/lib/gluon/config-mode/wizard/0200-domain-select.lua
@@ -2,7 +2,6 @@
 return function(form, uci)
 	local site_i18n = i18n 'gluon-site'
 
-	local json = require 'jsonc'
 	local site = require 'gluon.site'
 	local util = require 'gluon.util'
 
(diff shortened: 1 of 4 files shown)


package/gluon-config-mode-domain-select/luasrc/lib/gluon/config-mode/wizard/0200-domain-select.lua
```diff
@@ -22,10 +21,7 @@ return function(form, uci)
 
 	local function get_domain_list()
 		local list = {}
-		for _, domain_path in ipairs(util.glob('/lib/gluon/domains/*.json')) do
-			local domain_code = domain_path:match('([^/]+)%.json$')
-			local domain = assert(json.load(domain_path))
-
+		for domain_code, domain in pairs(util.get_domains()) do
```
NeoRaider, line 24: This looks wrong, was this even tested?

- `get_domains()` returns an integer-indexed table of entries of the form `{ domain_code = ..., domain = ... }`
- Using `pairs()` you are iterating over the table key-value pairs (in unspecified order), so in this loop `domain_code` will be the integer index of the domain, and `domain` will be the `{ domain_code = ..., domain = ... }` entry

For `gluon-config-mode-domain-select` the optimal solution would be change `get_domains()` to return a table mapping the domain code to the loaded JSON - in that case iterating over it using `pairs()` would do the correct thing (and the unspecified iteration order is fine, as we sort the list afterwards).

The hoodselector would require more changes if `get_domains()` were modified this way (and it can't depend on the alphabetical order of the entries returned by `glob()`).

2tata:
I am now even not sure if we should merge this change. Thus this code change will also appear on non domain setups. Which will be deadcode. It is possible to use kind of a precompiler to remove code parts at compile time?

CodeFetch:
@2tata The size of your changes are negligible. This should compress pretty well... Please don't hold your innovative spirit back, because of such restrictions as storage capacity and memory size. We have plenty of it without the 4 MiB devices...

blocktrron:
No progress and questioned by the author if the change is justified.

Feel free to reopen once these points are cleared up.

PR #2142, opened 2020-10-25
generalize the build-framework 

SvenRoederer:
This change adds the option "GLUON_BUILDTYPE", to support building other types of firmwireswith the build-framework.
Currently several gluon-specific settings are hard-coded and can't be changed easily, esp. "CONFIG_TARGET_DEVICE_*" and most of the things from targets/generic.

As gluon has some nice features which are missing in other build-frameworks but also some incompatible settings I'm looking for a way of reusing most of the gluon-work. My initial idea can be seen in #1946 and this PR is a up-to-date piece of the old cold. 
As the old PR seems to be welcome, but to complex to review, I'm doing this piece by piece now.

So looking for some feedback.


SvenRoederer:
@NeoRaider @CodeFetch as you have been the last commenting on #1946, what are your thoughts on this?

CodeFetch:
This is a very clean solution. Maybe call the variable FLAVOR instead...

SvenRoederer:
I also thought of any combination of [GLUON|BUILD]_[FLAVOR|TYPE] - so I'll update as you like (per convention?)
In addition I was thinking of moving the generic_* files out of targets/ or into a subfolder. As the more TYPEs will come, the more it bloats the folder.

NeoRaider:
Adding a way to reference alternative configurations is acceptable, but I don't want these alternative configurations as part of the Gluon repo. The split of the main `generic` target file is also a maintenance burden, as it is not clear which changes should be done in `generic` and which in `generic_gluon`.

If you still want to make this change, I propose the following:
- Do not split `generic`
- Maintain `generic_ffberlin` somehwere else, include what you left in `generic`

I'm not sure what you gain by using this part of the Gluon build system - a lot of this code's complexity stems from per-device package handling and the `GLUON_FEATURES` system. For the static configuration that `generic_ffberlin` contains, using OpenWrt's `scripts/diffconfig.sh` to store a config seed seems like a much more appropriate solution.

CodeFetch:
What about just skipping the generic configuration if a "flavor.mk" file exists in the site directory and adding the changes there?

SvenRoederer:
> If you still want to make this change, I propose the following:
>     * Do not split `generic`
>     * Maintain `generic_ffberlin` somehwere else, include what you left in `generic`
> 

Splitting this file was driven by making handling of upstream changes (gluon) automatically work with downstream forks (e.g. Freifnuk-berlin). But I completely agree to your concerns of the maintenance hassle that this change might cause here. Pushing this downstream is completely to understand, even not liked from my "downstream" perspective :-) .

> 
> I'm not sure what you gain by using this part of the Gluon build system - a lot of this code's complexity stems from per-device package handling and the `GLUON_FEATURES` system. For the static configuration that `generic_ffberlin` contains, using OpenWrt's `scripts/diffconfig.sh` to store a config seed seems like a much more appropriate solution.

At the end all my effort is driven by PR #1946;  to combine the flexible way Gluon offers to define packagelists, features and per-device packagelists and integrate this with using the OpenWrt imagebuilder.

SvenRoederer:
> What about just skipping the generic configuration if a "flavor.mk" file exists in the site directory and adding the changes there?

Also a nice idea. But modifying the generic in the downstream fork causes merge-conflicts, which will help to detect incompatible upstream (gluon) changes automatically. 

> 
> @SvenRoederer I agree with NeoRaider that you could actually make more use of the build system which could profit you, but I see that it takes much effort...

As mentioned, my intention is based on #1946 - to make gluon a more modular firmware build-framework. The outcome of this PR was, that adding support for imagebuilder based operation is welcome I'm happy to sped time in this for gluon, rather that just "stealing" your code and modifying it for our "downstream usage".

> EDIT2: I've just looked more closely at your code. It is impressing how simple it is. I'll try to get a grasp on your network configuration and my make thoughts on how to create a compatible Gluon-version.

thanks, looking forward to it.

CodeFetch:
We are currently looking into the issue in Hanover and try to find a good way for building images similar to the image builder... First step we are working on is to build an OpenWrt toolchain using CONFIG_MAKE_TOOLCHAIN and make it reusable with CONFIG_EXTERNAL_TOOLCHAIN. Then we need to add a possibility to re-use the tools etc. And then in the end we could create a menuconfig interface and adapt LibreMesh's chef to build Gluon images for example.

From my personal perspective I'd like to keep Gluon rather clean in a "uniformity"/modularity sense and independent. Therefore I'd like to keep the UI etc. independent and rather integrate features of other mesh firmwares like LibreMesh into the existing Gluon framework. OpenWrt gives us "device support" and that's all we need I think...

CodeFetch:
I totally forgot this... There was a bug which prevented using the external toolchain. It has been fixed in https://github.com/openwrt/openwrt/pull/10179. Maybe @lemoer is still interested.

AiyionPrime:
@SvenRoederer Are you still interested in investing time in this?
If so, a good start would be to address neoraiders suggestions.
If it does not have priority for you anymore, I'd kindly ask you to close this, until you'd like to pick it up again.

AiyionPrime:
@SvenRoederer Any answer is fine, but none will result in this being closed in the next days.
That does not need to be permanent; whenever you or somebody else wants to continue working on this, we can reopen it.

AiyionPrime:
@SvenRoederer I'll be closing this here than, so others can take up your work if wanted.

If you'll ever intend to continue working on this, just let me know, I'll reopen it in a blink.
Thanks for the effort and have a nice weekend.

PR #2144, opened 2020-11-04
ramips-mt7620: use manifest alias for EX3700/EX3800

blocktrron:
Use a manifest alias for the autoupdater image string of the Netgear
EX3700 / EX3800. This way, no unnecessary symlink is created and the
autoupdater functionality is preserved.


ambassador86:
Thanks for the hint. The next time when I encounter such devices I will use this instead of the symlink.

PR #2145, opened 2020-11-09
ath79-generic: add support TP-Link Archer C2 v3

mweinelt:
https://openwrt.org/toh/tp-link/tp-link_archer_c2_ac900

64M RAM, guess we'll mark it as TINY/BROKEN.

  - [x] webinterface
  - [ ] tftp
  - [ ] other: <specify>
    ```
    # lua -e 'print(require("platform_info").get_image_name())'
    tp-link-archer-c2-v3
    ```
- wired network
- wifi (if applicable)
- led mapping
  - radio leds
    - [x] should show activity
  - switchport leds
- <s>outdoor devices only</s>
  - [ ] <s>added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`</s>

### DIFF
--- a/targets/ath79-generic
+++ b/targets/ath79-generic
@@ -73,6 +73,13 @@ device('ocedo-raccoon', 'ocedo_raccoon', {
 
 -- TP-Link
 
+device('tp-link-archer-c2-v3', 'tplink_archer-c2-v3', {
+	packages = ATH10K_PACKAGES_QCA9887,
+	class = 'tiny',
+	broken = true,  -- 64M ath9k + ath10k
+
+})
+
 device('tp-link-archer-c6-v2', 'tplink_archer-c6-v2', {
 	packages = ATH10K_PACKAGES_QCA9888,
 })
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac
@@ -95,6 +95,7 @@ local primary_addrs = {
 		{'ath79', 'generic', {
 			'glinet,gl-ar750s-nor',
 			'ocedo,raccoon',
+			'tplink,archer-c2-v3',
 		}},
 		{'brcm2708'},
 		{'ipq40xx', 'generic', {


targets/ath79-generic
```diff
@@ -73,6 +73,10 @@ device('ocedo-raccoon', 'ocedo_raccoon', {
 
 -- TP-Link
 
+device('tp-link-archer-c2-v3', 'tplink_archer-c2-v3', {
+	packages = ATH10K_PACKAGES_QCA9887,
```
blocktrron, line 77: `class = 'tiny',`

Device has only 64M with ath10k.

targets/ath79-generic
```diff
@@ -73,6 +73,11 @@ device('ocedo-raccoon', 'ocedo_raccoon', {
...
+  class = 'tiny',  -- 64M ath9k + ath10k
```
blocktrron, line 78: Well, the indentation looks off here. And in addition to tiny, it should also be marked as broken. 

Missed that in the first round. sorry.

blocktrron (changes requested):

targets/ath79-generic
mweinelt: > 64M RAM, guess we'll mark it as BROKEN.

I said as much in the OP.

docs/user/supported_devices.rst
```diff
@@ -205,6 +205,7 @@ ath79-generic
 
 * TP-Link
 
+  - Archer C2 (v3)
```
blocktrron, line 208: Broken devices are also not listed in the docs.

blocktrron (changes requested):

docs/user/supported_devices.rst
mweinelt: Fixed

mweinelt:
https://meshviewer.darmstadt.freifunk.net/#!/en/map/b04e265c1ef1

Runs with 5 GHz disabled, because the owner wants to keep that band for themselves.

PR #2149, opened 2020-11-20
[2019.1.x] scripts: download.pl: retry download using filename

blocktrron:
This fixes OpenWrts handling of some assets in case they have to be pulled from OpenWrt's sources mirror.

This ensures that images can still be built in case the original mirror goes down.

```
With this commit, the download script will try downloading source files
using the filename instead of the url-filename in case the previous
download attempt using the url-filename failed.

This is required, as the OpenWrt sources mirrors serve files using the
filename files might be renamed to after downloading. If the original
mirror for a file where url-filename and filename do not match goes
down, the download failed prior to this patch.

Further improvement can be done by performing this only for the
OpenWrt sources mirrors.
```

### DIFF
--- a/patches/openwrt/0027-scripts-download.pl-retry-download-using-filename.patch
+++ b/patches/openwrt/0027-scripts-download.pl-retry-download-using-filename.patch
@@ -1,0 +1,51 @@
+From: David Bauer <mail@david-bauer.net>
+Date: Wed, 18 Nov 2020 16:02:23 +0100
+Subject: scripts: download.pl: retry download using filename
+
+With this commit, the download script will try downloading source files
+using the filename instead of the url-filename in case the previous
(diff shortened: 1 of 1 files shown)


SvenRoederer:
It's just a backport from OpenWrt master (https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=d36999389890fb952fc7cc8c0db8e1bbb671af12), right?

rotanid:
@blocktrron is your PR obsoleted via https://github.com/freifunk-gluon/gluon/commit/99e1e52740187874edb5ec58e2d01bccaea91a5d
?

rotanid:
looks like it, backport for v2019.1.x (OpenWrt 18.06)

PR #2153, opened 2020-11-24
ath79: Add support for Plasma Cloud PA300(E)

ecsv:
* Check the separate patches for the Forwarded: information
* Two devices are in this PR do avoid the conflicts between two independent PRs which depend on the same base modifications

  - [ ] webinterface
  - [ ] tftp
  - [x] other: factory image with ap51-flash
- wired network
- wifi (if applicable)
- led mapping
  - radio leds (**none available**)
    - [ ] should show activity
  - switchport leds (**none available**)
- outdoor devices only

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -206,6 +206,7 @@ ath79-generic
 * Plasma Cloud
 
   - PA300
+  - PA300E
 
 * TP-Link
 
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -203,6 +203,10 @@ ath79-generic
 
   - Raccoon
 
+* Plasma Cloud
+
+  - PA300
+
 * TP-Link
 
   - Archer C6 (v2)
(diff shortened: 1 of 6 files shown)


mweinelt (commented):
The gluon side of things LGTM. I cannot speak to backporting these patches and would like to defer to @blocktrron or @NeoRaider on those.

package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac
```diff
@@ -125,6 +125,10 @@ local primary_addrs = {
 			'om2pv4',
 			'om2p-hsv4',
 		}},
+		{'ath79', 'generic', {
+			'plasmacloud,pa300',
+			'plasmacloud,pa300e',
+		}},
```
blocktrron, line 131: I've just merged #2170 - As both device-trees define the label-mac-device, this is now obsolete.

blocktrron (commented):
Same goes for the other Plasma Cloud device. IMHO a backport is fine as long as these boards are already supported in OpenWrt master.

package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac
ecsv: thanks for the info. I have now dropped it.

PR #2154, opened 2020-11-24
Mention names of opkg-keys in doc

yanosz:
@mweinelt  Regarding our discussion on IRC, hoping not to have identified a temporary or second hand location.

### DIFF
--- a/docs/user/getting_started.rst
+++ b/docs/user/getting_started.rst
@@ -159,6 +159,14 @@ to sign the generated package repository).
 OpenWrt will handle the generation and handling of the keys itself.
 When making firmware releases based on Gluon, it might make sense to store
 the keypair, so updating the module repository later is possible.
+In fact you should take care to reuse the same opkg keypair, so you don't pollute the key
+store (see ``/etc/opkg/keys``) on the node.
+
+The signing-key is stored at ``openwrt/key-build.pub``, ``openwrt/key-build``,
+``key-build.ucert`` and  ``key-build.ucert.revoke``.
+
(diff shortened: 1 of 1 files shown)


docs/user/getting_started.rst
```diff
@@ -160,6 +160,9 @@ OpenWrt will handle the generation and handling of the keys itself.
 When making firmware releases based on Gluon, it might make sense to store
 the keypair, so updating the module repository later is possible.
```
mweinelt, line 161: In fact it might make sense to keep your keys as to not unnecessarily pollute `/etc/opkg/keys`, as new keys will accumulate there.

```
/etc/opkg/keys# ls
079c0a37b972c1a1  5151f69420c3f508  7f11d80444aab12d  b5043e70f9a75cde  c10b9afab19ee428  dd6de0d06bbd3d85  f94b9dd6febac963
0b26f36ae0f4106d  5b49f5281035bda8  81872ee93ddb1598  b5d1310b4ed4f371  c121aa1b252ca6be  e060f96eb975b1d1  fa70faf16b7ea02a
1035ac73cc4e59e3  72a57f2191b211e0  9ef4694208102c43  bbde64923f153391  dace9d4df16896bf  ebe8d5417501206c
1a3d48e69b4f1ab8  792d9d9b39f180dc  b2d571e0880ff617  bfdf8d1f965eeca8  dcfc4b885ef61fdd  f74ce42dab2d5fe2
```

 Can we add that?

mweinelt (changes requested):
Thanks for taking care of this. I think we should more strongly mention that people should backup and restore those, and even reuse an existing keypair for all of their builds, if possible.

PR #2156, opened 2020-12-02
[2020.2.x] kernel: bridge: Implement MLD Querier wake-up calls / Android bug workaround

T-X:
Implement a configurable MLD Querier wake-up calls "feature" which
works around a widely spread Android bug in connection with IGMP/MLD
snooping.

Currently there are mobile devices (e.g. Android) which are not able
to receive and respond to MLD Queries reliably because the Wifi driver
filters a lot of ICMPv6 when the device is asleep - including
MLD. This in turn breaks IPv6 communication when MLD Snooping is
enabled. However there is one ICMPv6 type which is allowed to pass and
which can be used to wake up the mobile device: ICMPv6 Echo Requests.

If this bridge is the selected MLD Querier then setting
"multicast_wakeupcall" to a number n greater than 0 will send n
ICMPv6 Echo Requests to each host behind this port to wake
them up with each MLD Query. Upon receiving a matching ICMPv6 Echo
Reply an MLD Query with a unicast ethernet destination will be sent
to the specific host(s).

Link: https://issuetracker.google.com/issues/149630944
Link: https://github.com/freifunk-gluon/gluon/issues/1832

(cherry picked from commit b2add48d2a7fa86c05eff3a2e6935168d60b2bf8)

### DIFF
--- a/patches/openwrt/0015-kernel-bridge-Implement-MLD-Querier-wake-up-calls-Android-bug-workaround.patch
+++ b/patches/openwrt/0015-kernel-bridge-Implement-MLD-Querier-wake-up-calls-Android-bug-workaround.patch
@@ -1,0 +1,2605 @@
+From: Linus Lüssing <linus.luessing@c0d3.blue>
+Date: Tue, 30 Jun 2020 18:01:56 +0200
+Subject: kernel: bridge: Implement MLD Querier wake-up calls / Android bug workaround
+
+Implement a configurable MLD Querier wake-up calls "feature" which
+works around a widely spread Android bug in connection with IGMP/MLD
+snooping.
+
+Currently there are mobile devices (e.g. Android) which are not able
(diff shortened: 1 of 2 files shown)


T-X:
As we have multicast snooping enabled on the br-lan for quite a while, Android devices should be affected by this issue already, even though the batman-adv multicast-to-unicast feature is not enabled yet.

This patch has been running fine at Freifunk Vogtland for some months now, with no issues reported so far.

As this is an annoying bug for Android users, maybe a backport for the v2020.2.2 release might be appropriate?

rotanid:
we ( @blocktrron @NeoRaider @mweinelt ) discussed this PR on IRC today and neither of us was fond of a backport to a stable branch of changes with possible high impact with the small field test coverage  this change has so far.

my opinion:
if  there are reports of other communities rolling out this change to multi-hundred nodes and with >1 month time since the rollout, we can discuss this again

Issue #2155, opened 2020-11-29
Multiple B.A.T.M.A.N. Adv. patches for better performance


awlx:
As @FreifunkMUC we tracked down several issues of batman-adv when it runs in another encapsulation (eg VXLAN). The Batman project now has several patches for that. What is the best way to get them to Gluon?

Background:
https://gist.github.com/awlx/27c383511182f9bfb9d9e098abb823fe

Patches:
https://patchwork.open-mesh.org/project/b.a.t.m.a.n./patch/20201127173849.19208-2-sw@simonwunderlich.de/

mweinelt:
They should honestly come via the routing feed. CCing @ecsv for his opinion.

rotanid:
as far as i can see, the patches aren't even in upstream batman-adv yet?
(and i guess the patches are for batman-adv, not the protocol B.A.T.M.A.N. itself?)

awlx:
They got pulled into net today as far as I can see @rotanid.
https://lists.open-mesh.org/mailman3/hyperkitty/list/b.a.t.m.a.n@lists.open-mesh.org/message/Y7FYBSSUJO4MVDEDJBMBVNFRXO7TWLP6/

rotanid:
ah right sorry, 27 hours ago in upstream at git.open-mesh.org

awlx:
Also I am not saying this should happen _right_ now. I just wanted to open the issue to keep track of the changes :).

rotanid:
as it seems to be improvement, not bugfixing, i'd say we wait for the batman-adv release including these patches, then they will also be part of openwrt-routing and we would get them anyway by bumping the routing feed :)
other maintainers may have other opinions :)

ecsv:
These patches will be backported on the next release (2021.0) and put into the openwrt-routing feed. If it is really urgent for someone then I can also ask Simon to merge an early PR with these changes.

But also keep in mind that you also have to fix vxlan:

* https://patchwork.ozlabs.org/project/netdev/patch/20201126125247.1047977-1-sven@narfation.org/
* https://patchwork.ozlabs.org/project/netdev/patch/20201126125247.1047977-2-sven@narfation.org/

CodeFetch:
We somehow did not really measure a big improvement with the patches.
You can find the patches for the Gluon master here:
https://github.com/freifunkh/site/tree/pr_wireguard_vpn/patches

awlx:
Maybe it's not really measurable on the nodes :/.

[...]

PR #2163, opened 2020-12-21
Generalize proto "gluon_wired.sh"

lemoer:
These are two small patches to allow using the `gluon_wired.sh` proto for the `mesh_vpn` interface.

This works for layer 2 and layer 3 vpns:
- In case of a layer 2 vpn, this just adds the `gluon_mesh.sh` proto as before.
- In case of a layer 3 vpn, creates an intermediate vxlan interface on top of the interface before applying the `gluon_mesh.sh` proto.

Furthermore an additional setting `vxpeer6addr` is added to the gluon_wired.sh proto, which is only used, if a vxlan layer is added. It can be used to control the peer address of vxlan.

This is preparatory to support wireguard in a clean way.

### DIFF
--- a/package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
+++ b/package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
@@ -8,6 +8,7 @@ proto_gluon_wired_init_config() {
         proto_config_add_boolean transitive
         proto_config_add_int index
         proto_config_add_boolean vxlan
+        proto_config_add_string vxpeer6addr
 }
 
 xor2() {
@@ -15,8 +16,24 @@ xor2() {
         echo -n "${1:1:1}" | tr '0123456789abcdef' '23016745ab89efcd'
 }
 
+is_layer3_device () {
+        local addrlen="$(cat "/sys/class/net/$1/addr_len")"
+        test "$addrlen" -eq 0
+}
+
 # shellcheck disable=SC2086
 interface_linklocal() {
+        if is_layer3_device "$1"; then
+                if ! ubus call network.interface dump | \
+                     jsonfilter -e "@.interface[@.l3_device='$1']['ipv6-address'][*].address" | \
+                     grep -e '^fe[89ab][0-9a-f]' -m 1; then
+                        proto_notify_error "$config" "MISSING_LL_ADDR_ON_LOWER_IFACE"
+                        proto_block_restart "$config"
+                        exit 1
+                fi
+                return
+        fi
+
         local macaddr="$(ubus call network.device status '{"name": "'"$1"'"}' | jsonfilter -e '@.macaddr')"
         local oldIFS="$IFS"; IFS=':'; set -- $macaddr; IFS="$oldIFS"
 
@@ -29,11 +46,12 @@ proto_gluon_wired_setup() {
 
         local meshif="$config"
 
-        local transitive index vxlan
-        json_get_vars transitive index vxlan
+        local transitive index vxlan vxpeer6addr
+        json_get_vars transitive index vxlan vxpeer6addr
 
         # default args
         [ -z "$vxlan" ] && vxlan=1
+        [ -z "$vxpeer6addr" ] && vxpeer6addr='ff02::15c'
 
         proto_init_update "$ifname" 1
         proto_send_update "$config"
@@ -46,8 +64,9 @@ proto_gluon_wired_setup() {
                 [ -n "$index" ] && json_add_string macaddr "$(lua -e "print(require('gluon.util').generate_mac($index))")"
                 json_add_string proto 'vxlan6'
                 json_add_string tunlink "$config"
+                # ip6addr (the lower interface ip6) is used by the vxlan.sh proto
                 json_add_string ip6addr "$(interface_linklocal "$ifname")"
-                json_add_string peer6addr 'ff02::15c'
+                json_add_string peer6addr "$vxpeer6addr"
                 json_add_int vid "$(lua -e 'print(tonumber(require("gluon.util").domain_seed_bytes("gluon-mesh-vxlan", 3), 16))')"
                 json_add_boolean rxcsum '0'
                 json_add_boolean txcsum '0'
--- a/package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
+++ b/package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
@@ -7,6 +7,7 @@ init_proto "$@"
 proto_gluon_wired_init_config() {
         proto_config_add_boolean transitive
         proto_config_add_int index
+        proto_config_add_boolean vxlan
 }
 
 xor2() {
@@ -28,15 +29,16 @@ proto_gluon_wired_setup() {
 
         local meshif="$config"
 
-        local vxlan="$(lua -e 'print(require("gluon.site").mesh.vxlan(true))')"
+        local transitive index vxlan
+        json_get_vars transitive index vxlan
 
-        local transitive index
-        json_get_vars transitive index
+        # default args
+        [ -z "$vxlan" ] && vxlan=1
 
         proto_init_update "$ifname" 1
         proto_send_update "$config"
 
-        if [ "$vxlan" = 'true' ]; then
+        if [ "$vxlan" -eq 1 ]; then
                 meshif="vx_$config"
 
                 json_init
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/220-interface-lan
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/220-interface-lan
@@ -15,6 +15,7 @@ uci:section('network', 'interface', 'mesh_lan', {
 	igmp_snooping = false,
 	proto         = 'gluon_wired',
 	index         = 4,
+	vxlan         = site.mesh.vxlan(true),
 })
 
 if sysconfig.lan_ifname:match(' ') then
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/210-interface-wan
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/210-interface-wan
@@ -7,6 +7,7 @@ uci:section('network', 'interface', 'mesh_wan', {
 	ifname = 'br-wan',
 	proto  = 'gluon_wired',
 	index  = 0,
+	vxlan  = site.mesh.vxlan(true),
 })
 
 local enable = site.mesh_on_wan(false)


package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
```diff
@@ -7,15 +7,29 @@ init_proto "$@"
...
 }
 
 xor2() {
         echo -n "${1:0:1}"
         echo -n "${1:1:1}" | tr '0123456789abcdef' '23016745ab89efcd'
 }
 
+is_layer3_device () {
+        local addrlen="$(cat "/sys/class/net/$1/addr_len")"
+        test "$addrlen" -eq 0
```
mweinelt, line 21: Only L2 devices have MAC addresses, hence an `addr_len` of 0 bytes indicates a L3 device.

```
What:		/sys/class/net/<iface>/addr_len
Date:		April 2005
KernelVersion:	2.6.12
Contact:	netdev@vger.kernel.org
Description:
		Indicates the hardware address size in bytes.
		Values vary based on the lower-level protocol used by the
		interface (Ethernet, FDDI, ATM, IEEE 802.15.4...). See
		include/uapi/linux/if_*.h for actual values.
```

package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
```diff
@@ -28,10 +42,22 @@ proto_gluon_wired_setup() {
...
+
+        ( proto_add_host_dependency "$config" '' "$ifname" )
 
-        local transitive index
-        json_get_vars transitive index
+        # default args
+        [ -z "$is_tunnel" ] && is_tunnel=0
+        [ -z "$vxpeer6addr" ] && vxpeer6addr='ff02::15c'
+
+        if [ "$is_tunnel" -eq 1 ] && is_layer3_device "$ifname"; then
```
mweinelt, line 54: Can you explain what `is_tunnel` represents? I'm not sure I follow the reasoning here.

lemoer: It is possible that `vxlan` for *mesh_on_lan* and *mesh_on_wan* is disabled in the site/domain.

However for vpn we want to have vxlan if the interface is a layer 3 vpn and disabled if we have a layer 2 vpn. There is no correlation to the `vxlan` setting from site.

As I was not able to get the information whether an interface is a tunnel from somewhere else, I introduced the proto variable `is_tunnel`.

package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
lemoer: That's correct.

package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
mweinelt: I wonder if we can get read `IFF_TUN` and `IFF_TAP` from `/sys/class/net/$ifname/tun_flags`.

https://elixir.bootlin.com/linux/v4.14.213/source/include/uapi/linux/if_tun.h#L62

lemoer: This does not seem to exist for wireguard devices.

mweinelt: No sensible way I see to reliably detect a tunnel interface in a consistent fashion.

package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
```diff
@@ -7,15 +7,29 @@ init_proto "$@"
...
+is_layer3_device () {
+        local addrlen="$(cat "/sys/class/net/$1/addr_len")"
+        test "$addrlen" -eq 0
+}
+
 # shellcheck disable=SC2086
 interface_linklocal() {
+        if is_layer3_device "$1"; then
+                # this is probably a layer 3 device
+                ubus call "network.interface.$1" status  |  \
```
mweinelt, line 28: Where is the status command coming from?

```
# ubus -v list network
'network' @4aeb533b
	"restart":{}
	"reload":{}
	"add_host_route":{"target":"String","v6":"Boolean","interface":"String"}
	"get_proto_handlers":{}
	"add_dynamic":{"name":"String"}
```

package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
```diff
@@ -7,15 +7,29 @@ init_proto "$@"
...
 }
 
+is_layer3_device () {
+        local addrlen="$(cat "/sys/class/net/$1/addr_len")"
+        test "$addrlen" -eq 0
+}
+
 # shellcheck disable=SC2086
 interface_linklocal() {
+        if is_layer3_device "$1"; then
```
mweinelt, line 26: What converts `mesh-vpn` (from https://github.com/freifunk-gluon/gluon/pull/2163/files#diff-44dd4d020d9ae0de557dc06405de0e7ce445eeb893d001cf1914a932f8bbedf6R11) to `mesh_vpn` needed for the ubus call?

lemoer: Netifd does that. `$ifname` (fetched from `$2`) in this script is **not** `ifname` from the uci section. `$ifname` is already resolved to `mesh_vpn` in this script.

I really dislike the naming conventions in netifd. It's just unnecessarily confusing. Scripts seem to name this argument `$2` differently:
```
# awk '/^proto_.*_setup/ { P=1; printf $1; } /^}/ {if (P) printf "\n"; P=0} /.*\$2.*/ { if(P) printf $0 }' /lib/netifd/proto/*
proto_batadv_setup()
proto_batadv_hardif_setup()	local iface="$2"
proto_batadv_vlan_setup()	local iface="$2"
proto_dhcp_setup()	local iface="$2"
proto_dhcpv6_setup()	local iface="$2"
proto_gluon_bat0_setup()
proto_gluon_mesh_setup()	export IFNAME="$2"
proto_gluon_wired_setup()        local ifname="$2"
proto_vxlan_setup()
proto_vxlan6_setup()
```

Naming it `iface` (like e.g. *proto_batadv_hardif_setup()*) seems to be better than naming it `ifname`. However still *proto_batadv_setup()* defines `iface="$config"` which means that `$iface=$1`:

```
root@lemoer-wgtest:/lib/netifd/proto# awk '/^proto_.*_setup/ { P=1; printf $1; } /^}/ {if (P) printf "\n"; P=0} /.*iface=.*/ { if(P) printf $0 }' /lib/netifd/proto/*
proto_batadv_setup()	local iface="$config"
proto_batadv_hardif_setup()	local iface="$2"
proto_batadv_vlan_setup()	local iface="$2"
proto_dhcp_setup()	local iface="$2"
proto_dhcpv6_setup()	local iface="$2"
proto_gluon_bat0_setup()
proto_gluon_mesh_setup()
proto_gluon_wired_setup()
proto_vxlan_setup()
proto_vxlan6_setup()
```

It's a mess in my oppinion. I don't know why netifd doesn't introduce something like "lower interface" or so. But maybe I am just not getting the concept...

mweinelt: Yet the call in L19/L33 only works when `$1` is `mesh-vpn`.

```sh
# ubus call network.device status '{"name": "mesh_vpn"}'
Command failed: Invalid argument
```

vs.

```sh
# ubus call network.device status '{"name": "mesh-vpn"}'
{
	"external": false,
	"present": true,
	"type": "Network device",
	"up": true,
	"carrier": true,
	"link-advertising": [
		
	],
	"link-partner-advertising": [
		
	],
	"link-supported": [
		
	],
	"speed": "10F",
	"autoneg": false,
	"mtu": 1312,
	"mtu6": 1312,
	"macaddr": "42:b6:0e:6d:16:c7",
	"txqueuelen": 1000,
	"ipv6": true,
	"promisc": false,
	"rpfilter": 0,
	"acceptlocal": false,
	"igmpversion": 0,
	"mldversion": 0,
	"neigh4reachabletime": 30000,
	"neigh6reachabletime": 30000,
	"neigh4gcstaletime": 60,
	"neigh6gcstaletime": 60,
	"neigh4locktime": 100,
	"dadtransmits": 1,
	"multicast": true,
	"sendredirects": true,
	"statistics": {
		"collisions": 0,
		"rx_frame_errors": 0,
		"tx_compressed": 0,
		"multicast": 0,
		"rx_length_errors": 0,
		"tx_dropped": 0,
		"rx_bytes": 0,
		"rx_missed_errors": 0,
		"tx_errors": 0,
		"rx_compressed": 0,
		"rx_over_errors": 0,
		"tx_fifo_errors": 0,
		"rx_crc_errors": 0,
		"rx_packets": 0,
		"tx_heartbeat_errors": 0,
		"rx_dropped": 0,
		"tx_aborted_errors": 0,
		"tx_packets": 10560199,
		"rx_errors": 0,
		"tx_bytes": 1723304086,
		"tx_window_errors": 0,
		"rx_fifo_errors": 0,
		"tx_carrier_errors": 0
	}
}
```

lemoer: Ok, well you are right, I am wrong. `$2` to `proto_*_setup()` is acutally a device name, not an interface name (speaking in netifd terminology). So we have a problem here.

lemoer:
```
ubus call network.interface dump | \
   jsonfilter -e "@.interface[@.l3_device='$1']['ipv6-address'][*].address" | \
   grep -e '^fe[89ab]' | head -n 1
```

lemoer: Applied this.

lemoer:
I rebased and removed the `proto_add_host_dependency` from *gluon_wired* again. It is not necessary, even worse, it has bricked things.

This PR is now tested with mesh_on_wan and with fastd.

package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
NeoRaider: In libgluonutil we have this code: https://github.com/freifunk-gluon/gluon/blob/master/package/libgluonutil/src/libgluonutil.c#L156-L202

NeoRaider:
Extending `gluon_wired` seems to add some complexity. Maybe we want to introduce a separate `gluon_vpn` proto? Both protos could source a file with common code though.

I might be overlooking something, but I don't understand why the fastd config is modified by this PR. Can't that just use `gluon_mesh` as before, and we only use a different proto for L3 VPNs?

lemoer:
I experimented a bit and with introducing new proto. The drawback of this approach is that we would have to complicate the firewall rule generation, as it is not that simple anymore to select all interfaces with proto `gluon_wired` into the zone `gluon_wired`. We would have to rename the zone and add interfaces matching the new proto aswell, or we would have to duplicate the section.

In my understanding, `gluon_wired` is just a proto, which *can* add a vxlan layer between `gluon_mesh` and the actual interface. It makes sense that this would also work for layer 3 devices.

However, some of the complexity is added due to the fact that the decision if a vxlan layer is added in between, is dependent to the interface. I would suggest therefore to move this decision out of the proto by exposing a boolean config parameter `add_vxlan_layer` from the proto.

Regarding fastd: I just modified the general *mesh_vpn* proto as the section was in the *gluon-mesh-vpn-core* package and `gluon_wired` supports leaving out the vxlan layer. That was the easiest option without the need to move the *mesh_vpn* proto into the other packages. I can differentiate here, if that's desired.

lemoer:
I implemented the suggested changes.

I kept changing the proto of the vpn interface to `gluon_wired`, as I currently see no benefit in staying with `gluon_mesh` here, if we would want to introduce `gluon_wired` for wireguard devices anyways and `gluon_wired` supports omitting the creation of the vxlan layer.

package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
lemoer: Not necessary anymore.

lemoer:
I just pushed a new version of this patch. I generalized it without adding too much extra glue. It's ready for review (and afterwards merge).

package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
```diff
@@ -7,15 +7,29 @@ init_proto "$@"
 proto_gluon_wired_init_config() {
         proto_config_add_boolean transitive
         proto_config_add_int index
+        proto_config_add_boolean add_vxlan_layer
```
NeoRaider, line 10: Please move the addition and handling of this option to a separate commit. I think a shorter name like `vxlan` would look better.

The option would probably not exist for long, as non-VXLAN setups should be phased out eventually (but we don't have a schedule for that yet - will depend on the patches mentioned in #2155 I guess).

package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
```diff
@@ -7,15 +7,29 @@ init_proto "$@"
...
+        local addrlen="$(cat "/sys/class/net/$1/addr_len")"
+        test "$addrlen" -eq 0
+}
+
 # shellcheck disable=SC2086
 interface_linklocal() {
+        if is_layer3_device "$1"; then
+                ubus call network.interface dump | \
+                        jsonfilter -e "@.interface[@.l3_device='$1']['ipv6-address'][*].address" | \
+                        grep -e '^fe[89ab]' | head -n 1
```
NeoRaider, line 29: Hmm, I assume this is about L3 VPN tunnels like Wireguard? How will link-local addresses be assigned for that?

Maybe we should enforce the configuration of an explicit `vxip6addr` in addition to `vxpeer6addr` in such cases, and only support L2 interfaces when no local address is configured.

package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
```diff
@@ -28,24 +42,31 @@ proto_gluon_wired_setup() {
...
+        json_get_vars transitive index vxpeer6addr add_vxlan_layer
 
-        local transitive index
-        json_get_vars transitive index
+        # default args
+        if [ -z "$add_vxlan_layer" ]; then
+                proto_notify_error "$config" "MISSING_OPTION_ADD_VXLAN_LAYER"
+                proto_block_restart "$config"
+                exit
+        fi
```
NeoRaider, line 53: No need to emit an error here, we can just let the option default to 1.

package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
```diff
@@ -56,6 +77,7 @@ proto_gluon_wired_setup() {
         json_init
         json_add_string name "${config}_mesh"
         json_add_string ifname "@${meshif}"
+        [ -n "$index" ] && json_add_string macaddr "$(lua -e "print(require('gluon.util').generate_mac($index))")"
```
NeoRaider, line 80: Remove this, it should be meaningless in VXLAN mode (as the MAC address is set above), and incorrect otherwise (will modify the MAC address of an underlying interface, but we already set all MAC addresses as intended).

NeoRaider (changes requested):
I'm not sure how long-lived this solution will be. I'd like to get rid of the static index assignment in Gluon's network setup eventually, but I don't have a design for a replacement yet.

package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
lemoer: It's dynamically added via `proto_add_ipv6_address` in a netifd script.

lemoer: Please let me know when you have decided if you want to introduce `vxip6addr` instead.

NeoRaider: I think using the LL address of the underlying interface is fine for now, but the code will need error handling to make the setup fail when no such address exists.

lemoer:
Just updated the PR. All changes suggested by @NeoRaider are now implemented.

I implemented the last change as extra commit to make changes clear. This means that the commits should be squashed during the merge.

AiyionPrime:
@mweinelt or @NeoRaider would one of you like to review again?

package/gluon-core/luasrc/lib/gluon/upgrade/220-interface-lan
```diff
@@ -15,6 +15,7 @@ uci:section('network', 'interface', 'mesh_lan', {
 	igmp_snooping = false,
 	proto         = 'gluon_wired',
 	index         = 4,
+	vxlan = site.mesh.vxlan(true),
```
blocktrron, line 18: indentation of equals.

package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
```diff
@@ -8,15 +8,28 @@ proto_gluon_wired_init_config() {
...
+        local addrlen="$(cat "/sys/class/net/$1/addr_len")"
+        test "$addrlen" -eq 0
+}
+
 # shellcheck disable=SC2086
 interface_linklocal() {
+        if is_layer3_device "$1"; then
+                ubus call network.interface dump | \
+                        jsonfilter -e "@.interface[@.l3_device='$1']['ipv6-address'][*].address" | \
+                        grep -e '^fe[89ab]' | head -n 1
```
mweinelt: What does this pattern match? Why `9ab`?

mweinelt (approved):
Looks pretty solid to me

package/gluon-core/luasrc/lib/gluon/upgrade/220-interface-lan
AiyionPrime: resolved in 8fdc8e70740e38fd6c1946fa9a9fd61d0bdfceed

blocktrron (approved):

rotanid:
two approvals here, could be merged, can you please squash the indention fix commit @AiyionPrime @lemoer ?

lemoer:
@rotanid thx, done

blocktrron:
Just so this is noted somewhere - custom created mesh interface will require to have a vxlan property assigned to them.

package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
```diff
@@ -28,24 +46,27 @@ proto_gluon_wired_setup() {
...
         local meshif="$config"
 
-        local vxlan="$(lua -e 'print(require("gluon.site").mesh.vxlan(true))')"
+        local transitive index vxlan vxpeer6addr
+        json_get_vars transitive index vxlan vxpeer6addr
 
-        local transitive index
-        json_get_vars transitive index
+        # default args
+        [ -z "$vxlan" ] && vxlan=1
```
blocktrron, line 53: maybe set the old default value using "$(lua -e 'print(require("gluon.site").mesh.vxlan(true))')"? This way it does not break custom interfaces w/o vxlan on sysupgrade.

lemoer: This was never officially supported. Of course, we could do this and (very few) people would benefit. But I think we should not start introducing migration code for unsupported configuration ways.

rotanid:
> Just so this is noted somewhere - custom created mesh interface will require to have a vxlan property assigned to them.

so we need this in docs and/or release notes, does it break existing stuff?

blocktrron:
It's breaking custom modifications which can break on any upgrade anyway.

So release notes (if at all) should be fine.

lemoer:
Thanks :)

rotanid:
> It's breaking custom modifications which can break on any upgrade anyway.
> 
> So release notes (if at all) should be fine.

if we know something, we should warn ;)

PR #1534, opened 2018-09-22
Add Wireguard Support to gluon

christf:
This patchset add wireguard integration into gluon. As usual, I am raising this early to get some feedback. 

Open tasks:
- [x] Improve documentation for server-side broker
- [x] assign ll-address based on nodeID to wg-interface on client-side

should we make use of 228cdef3584b9e8f99efc83c8251a34be85e5273 and extend that functionality? => not for now. In case tatas PR gets merged before this one, then the functionality to obtain the mesh interfaces must be extended.

PR #2165, opened 2020-12-29
packages: introduce syslog function in gluon.util

CodeFetch:
This commit introduces a simple function for writing to the syslog and replaces the custom function used by gluon-hoodselector.


2tata (approved):

mweinelt (approved):

mweinelt:
This was extracted from #1534 AFAIR, so we now have two users, so I see the purpose of extracting it.

PR #2166, opened 2020-12-29
Github Actions: Use matrix avoid repeats

awlx:
I simplified the Github workflow to make use of a matrix to spawn multiple build jobs.

See example here:
https://github.com/freifunkMUC/site-ffm/runs/1621859564?check_suite_focus=true


mweinelt:
If that replaces the generator scripts in `contrib/actions` please delete it, otherwise modify it.

PR #2167, opened 2020-12-30
packages: introduce toboolean function in gluon.util

CodeFetch:
This commit introduces a simple function to convert standard values which
are regularly being used as booleans like numbers and strings to booleans
to mimic the behavior of other programming languages.


rotanid:
Thanks for your contribution.
Although i don't see why we should blow up the code with a function which is not used anywhere in Gluon - maybe i'm missing something, please enlighten me in this case.

mweinelt:
Yeah, unless we have a consumer this makes no sense at all.

PR #2168, opened 2021-01-01
Add package gluon-mesh-vpn-wireguard using wgpeerselector

lemoer:
These patches implement wireguard vpn based on #2163 and the wgpeerselector in https://github.com/freifunk-gluon/packages/pull/240 .

Docs have not been written yet. This can be made up once the PR is approved. Some documentation regarding the wgpeerselector is available in https://github.com/freifunk-gluon/packages/pull/240 .

Related Work:
- https://github.com/freifunk-gluon/gluon/pull/1534
- https://github.com/freifunk-gluon/gluon/pull/2102
- https://github.com/freifunkMUC/community-packages/blob/gluon-mesh-vpn-wireguard-vxlan/gluon-mesh-vpn-wireguard-vxlan/files/lib/gluon/gluon-mesh-wireguard-vxlan/checkuplink

Feel free to discuss.

### DIFF
--- a/package/gluon-mesh-vpn-core/files/lib/netifd/proto/gluon_wireguard.sh
+++ b/package/gluon-mesh-vpn-core/files/lib/netifd/proto/gluon_wireguard.sh
@@ -1,0 +1,86 @@
+#!/bin/sh
+
+PROTO_DEBUG=1
+
+. /lib/functions.sh
+. ../netifd-proto.sh
+init_proto "$@"
+
+WG=/usr/bin/wg
+
+proto_gluon_wireguard_init_config() {
+	proto_config_add_int index
+	proto_config_add_int mtu
+}
+
+interface_linklocal_from_wg_public_key() {
+	# We generate a predictable v6 address
+	local macaddr="$(printf "%s" "$1"|md5sum|sed 's/^\(..\)\(..\)\(..\)\(..\)\(..\).*$/02:\1:\2:\3:\4:\5/')"
+	local oldIFS="$IFS"; IFS=':';
+	# shellcheck disable=SC2086
+	set -- $macaddr; IFS="$oldIFS"
+	echo "fe80::$1$2:$3ff:fe$4:$5$6"
+}
+
+proto_gluon_wireguard_setup() {
+	local config="$1"
+	local ifname="$2"
+
+	local index mtu
+	json_get_vars index mtu
+
+	local public_key="$(/lib/gluon/mesh-vpn/wireguard_pubkey.sh)"
+
+	# The wireguard proto itself can not be moved here, as the proto does not
+	# allow add_dynamic.
+
+	wireguard_ip=$(interface_linklocal_from_wg_public_key "$public_key")
+
+	## Add IP
+
+	proto_add_host_dependency "$config" '' "$ifname"
+	proto_init_update "$ifname" 1
+	proto_add_data
+	json_add_string zone 'wired_mesh'
+	proto_close_data
+	proto_add_ipv6_address "$wireguard_ip" "128"
+	proto_send_update "$ifname"
+
+	## wgpeerselector
+
+	json_init
+	json_add_string name "${ifname}_peerselector"
+	json_add_string ifname "$ifname"
+	json_add_string proto 'wgpeerselector'
+	json_add_string unix_group 'gluon-mesh-vpn'
+	json_add_boolean transitive 1
+	json_close_object
+	ubus call network add_dynamic "$(json_dump)"
+
+	## vxlan
+
+	json_init
+	json_add_string name "mesh-vpn"
+	json_add_string proto 'vxlan6'
+	json_add_string tunlink "$ifname"
+	# ip6addr (the lower interface ip6) is used by the vxlan.sh proto
+	json_add_string ip6addr "$wireguard_ip"
+	json_add_string peer6addr "fe80::1"
+	json_add_int vid "$(lua -e 'print(tonumber(require("gluon.util").domain_seed_bytes("gluon-mesh-vxlan", 3), 16))')"
+	json_add_boolean rxcsum '0'
+	json_add_boolean txcsum '0'
+	json_close_object
+	ubus call network add_dynamic "$(json_dump)"
+
+	proto_init_update "$ifname" 1
+	proto_send_update "$config"
+}
+
+proto_gluon_wireguard_teardown() {
+	local config="$1"
+
+	proto_init_update "*" 0
+	proto_send_update "$config"
+}
+
+add_protocol gluon_wireguard
--- a/package/gluon-mesh-vpn-core/luasrc/lib/gluon/upgrade/500-mesh-vpn
+++ b/package/gluon-mesh-vpn-core/luasrc/lib/gluon/upgrade/500-mesh-vpn
@@ -54,6 +54,8 @@ if not uci:get('gluon', 'mesh_vpn') then
 	else
 		enabled = site.mesh_vpn.enabled(false)
 	end
+	-- wireguard is not listed here, as it didn't exist before the uci section
+	-- gluon.mesh_vpn was introduced. Therefore no migration is necessary.
 
 
 	local limit_enabled = tonumber((uci:get('simple-tc', 'mesh_vpn', 'enabled')))
--- a/package/gluon-mesh-vpn-wireguard/Makefile
+++ b/package/gluon-mesh-vpn-wireguard/Makefile
@@ -1,0 +1,13 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-mesh-vpn-wireguard
+PKG_VERSION:=1
+
+include ../gluon.mk
+
+define Package/gluon-mesh-vpn-wireguard
+  TITLE:=Support for connecting meshes via wireguard
+  DEPENDS:=+gluon-core +libgluonutil +gluon-mesh-vpn-core +wireguard-tools +@GLUON_SPECIALIZE_KERNEL:KERNEL_TUN +wgpeerselector +libubus
+endef
+
+$(eval $(call BuildPackageGluon,gluon-mesh-vpn-wireguard))
--- a/package/gluon-mesh-vpn-wireguard/check_site.lua
+++ b/package/gluon-mesh-vpn-wireguard/check_site.lua
@@ -1,0 +1,9 @@
+local function check_peer(k)
+	need_alphanumeric_key(k)
+
+	need_string_match(in_domain(extend(k,
+		{'public_key'})), "^" .. ("[%a%d+/]"):rep(42) .. "[AEIMQUYcgkosw480]=$")
+	need_string(in_domain(extend(k, {'endpoint'})))
+end
+
+need_table({'mesh_vpn', 'wireguard', 'peers'}, check_peer)
--- a/package/gluon-mesh-vpn-wireguard/files/lib/gluon/mesh-vpn/wireguard_pubkey.sh
+++ b/package/gluon-mesh-vpn-wireguard/files/lib/gluon/mesh-vpn/wireguard_pubkey.sh
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+INCLUDE_ONLY=1
+. /lib/netifd/proto/wireguard.sh
+
+ensure_key_is_generated wg_mesh
+uci get "network.wg_mesh.private_key" | /usr/bin/wg pubkey
--- a/package/gluon-mesh-vpn-wireguard/luasrc/lib/gluon/upgrade/400-mesh-vpn-wireguard
+++ b/package/gluon-mesh-vpn-wireguard/luasrc/lib/gluon/upgrade/400-mesh-vpn-wireguard
@@ -1,0 +1,39 @@
+#!/usr/bin/lua
+
+local uci = require('simple-uci').cursor()
+local site = require 'gluon.site'
+
+local private_key = uci:get("network", 'wg_mesh', "private_key")
+
+if not private_key or not private_key:match("^" .. ("[%a%d+/]"):rep(42) .. "[AEIMQUYcgkosw480]=$") then
+  private_key = "generate"
+end
+
+uci:section('network', 'interface', 'wg_mesh', {
+  proto = 'wireguard',
+  fwmark = 1,
+  private_key = private_key,
+})
+
+uci:section('network', 'interface', 'mesh_wg_mesh', {
+	ifname = 'wg_mesh',
+	proto = 'gluon_wireguard'
+})
+
+-- Clean up previous configuration
+uci:delete_all('wgpeerselector', 'peer', function(peer)
+  return peer.preserve ~= '1'
+end)
+
+for name, peer in pairs(site.mesh_vpn.wireguard.peers()) do
+  uci:section("wgpeerselector", "peer", name, {
+    enabled = true,
+    endpoint = peer.endpoint,
+    public_key = peer.public_key,
+    allowed_ips = { "fe80::1/128" },
+    ifname = 'wg_mesh',
+  })
+end
+
+uci:save("wgpeerselector")
+uci:save("network")
--- a/package/gluon-mesh-vpn-wireguard/luasrc/usr/lib/lua/gluon/mesh-vpn/provider/wireguard.lua
+++ b/package/gluon-mesh-vpn-wireguard/luasrc/usr/lib/lua/gluon/mesh-vpn/provider/wireguard.lua
@@ -1,0 +1,37 @@
+local uci = require('simple-uci').cursor()
+
+local site = require 'gluon.site'
+local util = require 'gluon.util'
+local vpn_core = require 'gluon.mesh-vpn'
+
+local M = {}
+
+function M.public_key()
+	return util.trim(util.exec("/lib/gluon/mesh-vpn/wireguard_pubkey.sh"))
+end
+
+function M.enable(val)
+	uci:set('network', 'wg_mesh', 'disabled', not val)
+	uci:save('network')
+end
+
+function M.active()
+	return site.mesh_vpn.wireguard() ~= nil
+end
+
+function M.set_limit(ingress_limit, egress_limit)
+	-- TODO: Test that limiting this via simple-tc here is correct!
+	uci:delete('simple-tc', 'mesh_vpn')
+	if ingress_limit ~= nil and egress_limit ~= nil then
+		uci:section('simple-tc', 'interface', 'mesh_vpn', {
+			ifname = vpn_core.get_interface(),
+			enabled = true,
+			limit_egress = egress_limit,
+			limit_ingress = ingress_limit,
+		})
+	end
+
+	uci:save('simple-tc')
+end
+
+return M
--- a/package/gluon-mesh-vpn-wireguard/src/respondd.c
+++ b/package/gluon-mesh-vpn-wireguard/src/respondd.c
@@ -1,0 +1,231 @@
+/*
+  Copyright (c) 2020, Leonardo Mörlein <me@irrelefant.net>
+  Copyright (c) 2016, Matthias Schiffer <mschiffer@universe-factory.net>
+  All rights reserved.
+
+  Redistribution and use in source and binary forms, with or without
+  modification, are permitted provided that the following conditions are met:
+
+    1. Redistributions of source code must retain the above copyright notice,
+       this list of conditions and the following disclaimer.
+    2. Redistributions in binary form must reproduce the above copyright notice,
+       this list of conditions and the following disclaimer in the documentation
+       and/or other materials provided with the distribution.
+
+  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+  IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+  DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+  FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+  DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+  SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+  OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+
+#include <respondd.h>
+
+#include <json-c/json.h>
+#include <libgluonutil.h>
+#include <uci.h>
+
+#include <ctype.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+
+#include <sys/socket.h>
+#include <sys/un.h>
+
+#include "libubus.h"
+
+static struct json_object * stdout_read(const char *cmd, const char *skip, bool oneword) {
+	FILE *f = popen(cmd, "r");
+	if (!f)
+		return NULL;
+
+	char *line = NULL;
+	size_t len = 0;
+	size_t skiplen = strlen(skip);
+
+	ssize_t read_chars = getline(&line, &len, f);
+
+	pclose(f);
+
+	if (read_chars < 1) {
+		free(line);
+		return NULL;
+	}
+
+	if (line[read_chars-1] == '\n')
+		line[read_chars-1] = '\0';
+
+	const char *content = line;
+	if (strncmp(content, skip, skiplen) == 0)
+		content += skiplen;
+
+	if (oneword) {
+		for (int i = 0; i < len; i++) {
+			if (isspace(line[i])) {
+				 line[i] = 0;
+			}
+		}
+	}
+
+	struct json_object *ret = gluonutil_wrap_string(content);
+	free(line);
+	return ret;
+}
+
+static struct json_object * get_wireguard_public_key(void) {
+	return stdout_read("exec /lib/gluon/mesh-vpn/wireguard_pubkey.sh", "", false);
+}
+
+static struct json_object * get_wireguard_version(void) {
+	return stdout_read("exec wg -v", "wireguard-tools ", true);
+}
+
+static bool wireguard_enabled(void) {
+	bool enabled = true;
+
+	struct uci_context *ctx = uci_alloc_context();
+	if (!ctx)
+		goto disabled_nofree;
+	ctx->flags &= ~UCI_FLAG_STRICT;
+
+	struct uci_package *p;
+	if (uci_load(ctx, "network", &p))
+		goto disabled;
+
+	struct uci_section *s = uci_lookup_section(ctx, p, "wg_mesh");
+	if (!s)
+		goto disabled;
+
+	const char *disabled_str = uci_lookup_option_string(ctx, s, "disabled");
+	if (!disabled_str || !strcmp(disabled_str, "1"))
+		enabled = false;
+
+disabled:
+	uci_free_context(ctx);
+
+disabled_nofree:
+	return enabled;
+}
+
+static bool get_pubkey_privacy(void) {
+	bool ret = true;
+	struct json_object *site = NULL;
+
+	site = gluonutil_load_site_config();
+	if (!site)
+		goto end;
+
+	struct json_object *mesh_vpn;
+	if (!json_object_object_get_ex(site, "mesh_vpn", &mesh_vpn))
+		goto end;
+
+	struct json_object *pubkey_privacy;
+	if (!json_object_object_get_ex(mesh_vpn, "pubkey_privacy", &pubkey_privacy))
+		goto end;
+
+	ret = json_object_get_boolean(pubkey_privacy);
+
+end:
+	json_object_put(site);
+
+	return ret;
+}
+
+static struct json_object * get_wireguard(void) {
+	bool wg_enabled = wireguard_enabled();
+
+	struct json_object *ret = json_object_new_object();
+	json_object_object_add(ret, "version", get_wireguard_version());
+	json_object_object_add(ret, "enabled", json_object_new_boolean(wg_enabled));
+	if (wg_enabled && !get_pubkey_privacy())
+		json_object_object_add(ret, "public_key", get_wireguard_public_key());
+	return ret;
+}
+
+static struct json_object * respondd_provider_nodeinfo(void) {
+	struct json_object *ret = json_object_new_object();
+
+	struct json_object *software = json_object_new_object();
+	json_object_object_add(software, "wireguard", get_wireguard());
+	json_object_object_add(ret, "software", software);
+
+	return ret;
+}
+
+static json_object *blobmsg_attr2json(struct blob_attr *attr, int type)
+{
+	int len = blobmsg_data_len(attr);
+	struct blobmsg_data *data = blobmsg_data(attr);
+	struct blob_attr *inner_attr;
+	json_object *res = NULL;
+	switch(type) {
+		case BLOBMSG_TYPE_STRING:
+			return gluonutil_wrap_string(blobmsg_get_string(attr));
+		case BLOBMSG_TYPE_BOOL:
+			return json_object_new_boolean(blobmsg_get_bool(attr));
+		case BLOBMSG_TYPE_INT16:
+			return json_object_new_double(blobmsg_get_u16(attr));
+		case BLOBMSG_TYPE_INT32:
+			return json_object_new_double(blobmsg_get_u32(attr));
+		case BLOBMSG_TYPE_INT64:
+			return json_object_new_double(blobmsg_get_u64(attr));
+		case BLOBMSG_TYPE_DOUBLE:
+			return json_object_new_double(blobmsg_get_double(attr));
+		case BLOBMSG_TYPE_TABLE:
+			res = json_object_new_object();
+			__blob_for_each_attr(inner_attr, data, len) {
+				json_object_object_add(res, blobmsg_name(inner_attr), blobmsg_attr2json(inner_attr, blobmsg_type(inner_attr)));
+			};
+			break;
+		case BLOBMSG_TYPE_ARRAY:
+			res = json_object_new_array();
+			__blob_for_each_attr(inner_attr, data, len) {
+				json_object_array_add(res, blobmsg_attr2json(inner_attr, blobmsg_type(inner_attr)));
+			}
+			break;
+	}
+
+	return res;
+}
+
+static void cb_wgpeerselector_vpn(struct ubus_request *req, int type, struct blob_attr *msg)
+{
+	json_object_object_add(req->priv, "mesh_vpn", blobmsg_attr2json(msg, type));
+}
+
+static struct json_object * respondd_provider_statistics(void) {
+	struct json_object *ret = json_object_new_object();
+	struct ubus_context *ctx = ubus_connect(NULL);
+	uint32_t ubus_path_id;
+
+	if (!ctx) {
+		fprintf(stderr, "Error in gluon-mesh-vpn-wireguard.so: Failed to connect to ubus.\n");
+		goto err;
+	}
+
+	if (ubus_lookup_id(ctx, "wgpeerselector.wg_mesh", &ubus_path_id)) {
+		goto err;
+	}
+
+	ubus_invoke(ctx, ubus_path_id, "status", NULL, cb_wgpeerselector_vpn, ret, 1000);
+
+err:
+	if (ctx)
+		ubus_free(ctx);
+	return ret;
+}
+
+
+const struct respondd_provider_info respondd_providers[] = {
+	{"nodeinfo", respondd_provider_nodeinfo},
+	{"statistics", respondd_provider_statistics},
+	{}
+};
--- a/patches/openwrt/0005-package-uci-backport-cli-add-option-for-changing-save-path.patch
+++ b/patches/openwrt/0005-package-uci-backport-cli-add-option-for-changing-save-path.patch
@@ -1,0 +1,83 @@
+From: Leonardo Mörlein <me@irrelefant.net>
+Date: Sat, 16 Jan 2021 23:11:01 +0100
+Subject: package/uci: backport: "cli: add option for changing save path"
+
+This is a backport of
+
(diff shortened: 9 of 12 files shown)


package/gluon-mesh-vpn-wireguard/Makefile
```diff
@@ -0,0 +1,13 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-mesh-vpn-wireguard
+PKG_VERSION:=1
+
+include ../gluon.mk
+
+define Package/gluon-mesh-vpn-wireguard
+  TITLE:=Support for connecting meshes via wireguard
+  DEPENDS:=+gluon-core +libgluonutil +gluon-mesh-vpn-core +wireguard +@GLUON_SPECIALIZE_KERNEL:KERNEL_TUN +wgpeerselector +libubus
```
lemoer, line 10: Seems like wireguard doesn't need tun. 

Can I just `GLUON_SPECIALIZE_KERNEL:KERNEL_TUN` by `GLUON_SPECIALIZE_KERNEL:KERNEL_WIREGUARD` ?

AiyionPrime:
This peace of code works pretty well in hanover, as long as nobody reboots the server running screen...

Jokes aside, I just flashed to devices with two different build of the PR in question and ramped up the throughput of the cudy from 30-40Mbi/t to 60Mbit/s.

Hopefully they'll live throughout the day:

http://aiyion-jt-or750i.n.ffh.zone
http://wireguard-cudy-hannover.n.ffh.zone

package/gluon-mesh-vpn-wireguard/Makefile
blocktrron: If I'm not mistaken, this is not possible, as wireguard is not in-tree for this OpenWrt release, therefor it can't be built-in.

package/gluon-mesh-vpn-wireguard/luasrc/lib/gluon/upgrade/400-mesh-vpn-wireguard
```diff
@@ -0,0 +1,53 @@
...
+local function calc_address (public_key)
+  local s = hash.md5(public_key)
+  return 'fe80::02'..s:sub(1,2)..':'..s:sub(3,4)..'ff:fe'..s:sub(5,6)..':'..s:sub(7,10)
+end
+
+local private_key = uci:get("network", "vpn", "private_key")
+
+if not private_key or not private_key:match("^" .. ("[%a%d+/]"):rep(43) .. "=$") then
+  -- generate one
+  private_key = util.trim(util.exec("/usr/bin/wg genkey"))
```
lemoer, line 17: This is a bad idea, since the upgrade scripts seem to run before haveged is started. Not sure, whether this is intentional, but fastd seems to handle this by storing "generate" in the uci setting.

Leaving it like this is no option as this ocasionally bricks devices.

lemoer: Seems like the `wireguard` proto does not have an option like generate.

I don't know in which step we could do the generation otherwise. Ideas welcome.

lemoer:
Rebased on #2163.

package/gluon-mesh-vpn-wireguard/luasrc/lib/gluon/upgrade/400-mesh-vpn-wireguard
lemoer: I submitted this patch upstream: https://patchwork.ozlabs.org/project/openwrt/patch/20210108012435.175345-1-me@irrelefant.net/

T-X:
Hi @lemoer. Do I understand correctly that this is an alternative implementation for Wireguard in Gluon? If so, could you summarize the differences to #1534 and freifunk-gluon/community-packages/pull/6?

Would it make sense to coordinate/collaborate with @awlx, @krombel and @christf?

lemoer:
Hi @T-X,

you are right. This seems to be useful, as a lot of work has been done here and there. First of all, I would like to thank @awlx, @krombel, @christf and everyone else who contributed for their work. It was very helpful to rely on what you have already accomplished.

## Comparison to freifunk-gluon/community-packages/pull/6

- This pull-request is derived from their pull-request. So all concerns like ntp and so, which were discussed there, are included here already.
- We removed the *checkuplink* script and split it into reusable and better debuggable parts:
    - The main part is the [*wgpeerselector*](https://github.com/freifunk-gluon/packages/pull/240).
        - It's a daemon written in lua taking over most of tasks, which were handled by *checkuplink* before. Adding peers to wireguard, sync ntp, ....
        - One of the major additional features is proper logging.
        - Another new feature is, that this daemon keeps track of how long connections are already established.
        - While most of the algorithm here remains similar to *checkuplink*, we removed the broker part. I will add some words on that later on.
        - Furthermore the setup of the wireguard interface itself is done by the existing upstream *wireguard.sh* proto.
        - It has been written in such a manner, that it is independent from gluon and could be reused/extended in other projects. Therefore I moved it to the packages repo.
    - To set up the vxlan, we rely on existing code in *gluon_wired.sh* proto. (See #2163)
- We added a respondd provider.

## Comparison to #1534 

- #1534 is a about adding wireguard support for babel only, as it has no vxlan layer in mind.
- In #1534 one interface for each peer is created.
    - If I understand it correctly, this is because they want the nodes to connect to multiple peers at the same time.
    - As routing and wireguards AllowedIPs are two different things and routing decisions should be done by babeld, the AllowedIPs are set to `::/0` and one interface per peer is created. Babeld can then select the proper peer based on the interface only, without wireguard interfering.
- As we noted that in most cases gluon communities let their node connect to only one gateway at a time anyways, we did not implement having multiple interfaces managed by the same *wgpeerselector* instance.
    - However, as the code in *wgpeerselector* is written in a modular way, I suspect this might not be such a big change, once this is necessary/desired in future.
    - Not having multiple interfaces still allows using *wgepeerselector* for babeld.
- In contrast to freifunk-gluon/community-packages/pull/6, the *wgepeerselector* in its current state can be used by the layer 3 approach, as it does not rely on vxlan.

## Brokerless Architecture

- As the link local IPs used for the wireguard interfaces are derived by the public key anyways, it is not necessary to have a broker.
    - On the server, there just needs to be a small glue, which periodically checks `wg show` if there are new established wireguard connections and adds the vxlan subscriptions for the respective calculated IPs.
        - We wrote/adapted a small python daemon for this task: https://github.com/freifunkh/wireguard-vxlan-glue
    - Peers are added statically to the wireguard interface on the server side.
- The main benefit is simplicity.
- The main drawback is, that this only works for communities allowing vpn connections based on whitelisting peer keys. To allow connections from arbitrary untrusted nodes with wireguard, a broker will always be necessary.
- Adding broker support to *wgpeerselector*:
    - As the *wgpeerselector* is written in a modular way, it should not be too much work to add a connection to a broker.
    - Because the broker is unnecessary for communities with a whitelisting approach, the broker should always remain an optional part of the *wgpeerselector*.
- Of course, a broker has other benefits like the ability to distribute and balance nodes across supernodes. I find it interesting, but to KISS, I think, we should start with this feature set only.

To finish up, I would be very happy to receive opinions from @awlx, @krombel and @christf and of course any other contributor. I am also up for discussion in the IRC or in Mumble. Just ping me in the IRC.

awlx:
Hey, first of all nice work! 

But removing the broker part can also lead to slow performance on your gateways. Especially when you have many many nodes connecting. We analyzed that adding keys to wireguard and not cleaning up those when no connecting is established leads to high cpu usage on the long run.  So we cleanup all not connected nodes periodically and just add what is really needed.

Especially on gateways with 800 or more peers the approach of calling wg show periodically is not very performant.

AiyionPrime:
We stripped the cleanup routine quite drastically;
the routes are removed on a regular basis when the last handshake is about one minute above wg's refresh interval of two minutes, just the peers keys stay in place.
I hoped, that'd be sufficient to keep the load low; but maybe we were wrong on that?

awlx:
Hm, good question gonna verify that on one of our gateways :). The issue appears after like 12 hours so I setup the cleanup job to just delete the vxlan routes and report back :).

awlx:
But just as a comment, I would like to have the broker implemented anyways so people can turn it off or on. :)

AiyionPrime:
Though I really like the broker; I'm even more satisfied by splitting of such features if not strictly necessary for wg in gluon from a wg-pull request, in order to grant a higher likeliness of diversity in how communities will setup their wireguard infrastructure in the future.

I'd really like to hear about your upcoming verification :)

rotanid:
(blocked label added because of the dependency on #2163 )
i like the modular approach, if i'm right this solution can be extended to also work with babel or a  wg broker, so from a longterm/gluon-upstream perspective this should be the way to go in my opinion, including this and afterwards adding the optional parts for a broker and/or babel-based networks.

lemoer:
I updated this PR. We have been running the current state with about ~30 routers in Hannover for quite a while. Currently there are no known issues. It was actively tested for 3 weeks by about 15 different people.

Together with #2163 this is ready for review.

lemoer:
I updated this PR once again to add a contribution from @AiyionPrime to split up the "software" parts in respondd.

AiyionPrime:
> Hm, good question gonna verify that on one of our gateways :). The issue appears after like 12 hours so I setup the cleanup job to just delete the vxlan routes and report back :).

_@awlx_

Have you already found time for your verification?

Furthermore, at @freifunkh we're quite happy with the pr,
currently we're freezing our infrastructure for a week to find unconspicuous bugs.
But for now this looks pretty good, and delivers quite some bandwith on the user end.

I can only encourage other communities to test this, it's a mere weekend project to recreate the setup, even if on has not done anything with wireguard before. Furthermore @freifunkMUC documented their not so different approach very well, this should help to recognize patterns.

awlx:
> _@awlx_
> 
> Have you already found time for your verification?

No sorry, with new job and #FFMEET in full swing we just kept the old cleanup job.

lemoer:
Once #2163 is merged, I'll rebase this onto #2163. 

Steps to rebase onto #2163:
- [ ] Rename `add_vxlan_layer` to `vxlan`.

AiyionPrime:
> Once #2163 is merged, I'll rebase this onto #2163.

Good idea @lemoer :D


Furthermore, this should no longer be marked as blocked, right @rotanid ?
> (blocked label added because of the dependency on #2163 )

mweinelt:
I think this looks pretty good overall, but I have not tested it. I believe we should merge this right after the v2021.1 branch-off and give it a round of testing in master before actually releasing it.

I'm especially interested right now into the the scaling of lots of keys on the gateway side, as that seems to be a limiting factor as to how well this will work on the gateways. Is it routes - or keys? Because we currently hold roughly 1600 keys in fastd, and if that would be detrimental to performance that should be made well known. @AiyionPrime @awlx 

We'll also need some documentation (`feature/vpn` and `user/site#mesh_vpn`) to help possible tester get this configured, especially with a mention and explanation of the gateway glue that you have written to manage vxlan fdb entries and allowedroutes(?). @AiyionPrime @lemoer

awlx:
We see no performance issues at all anymore on the gateways which correlate with numbers of keys/peers. And we install all 1500ish keys on all gateways.

AiyionPrime:
Same in Hanover. Though we're at a mere 100 wireguard clients still.

package/gluon-mesh-vpn-wireguard/files/lib/gluon/mesh-vpn/wireguard_pubkey.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+INCLUDE_ONLY=1
+. /lib/netifd/proto/wireguard.sh
+
+ensure_key_is_generated vpn
+uci get network.vpn.private_key | /usr/bin/wg pubkey
```
zx2c4, line 7: `/usr/bin/wg` -> `wg`?  Why wouldn't /usr/bin be in PATH here? I assume `uci` is in /usr/bin too.

package/gluon-mesh-vpn-wireguard/luasrc/lib/gluon/upgrade/400-mesh-vpn-wireguard
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/lua
+
+local uci = require('simple-uci').cursor()
+local site = require 'gluon.site'
+
+local private_key = uci:get("network", "vpn", "private_key")
+
+if not private_key or not private_key:match("^" .. ("[%a%d+/]"):rep(43) .. "=$") then
```
zx2c4, line 8: `^[A-Za-z0-9+/]{42}[A|E|I|M|Q|U|Y|c|g|k|o|s|w|4|8|0]=$` is more accurate: https://lists.zx2c4.com/pipermail/wireguard/2020-December/006222.html

package/gluon-mesh-vpn-wireguard/src/respondd.c
```diff
@@ -0,0 +1,262 @@
...
+		free(line);
+		line = NULL;
+	}
+
+	const char *content = line;
+	if (strncmp(content, skip, skiplen) == 0)
+		content += skiplen;
+
+	if (oneword) {
+		for (int i = 0; i < len; i++){
```
zx2c4, line 74: Rather than start at `0` here, why not actually start at `skiplen` in the event it matches?

package/gluon-mesh-vpn-wireguard/src/respondd.c
```diff
@@ -0,0 +1,262 @@
...
+static struct json_object * get_wireguard_public_key(void) {
+	return stdout_read("exec /lib/gluon/mesh-vpn/wireguard_pubkey.sh", "", false);
+}
+
+static struct json_object * get_wireguard_version(void) {
+	return stdout_read("exec wg -v", "wireguard-tools ", true);
+}
+
+static bool wireguard_enabled(void) {
+	FILE *fd = popen("grep '^wireguard ' /proc/modules", "r");
```
zx2c4, line 99: This is brittle and won't do what you want. Firstly, shelling out to grep from C seems a bit needless when you could just fopen this file. But actually, that's the wrong file, because it won't contain "wireguard" when wireguard is built as a built-in rather than a module (which might eventually happen). A simpler way is to replace this whole function with one line:

```c
return access("/sys/module/wireguard", F_OK) == 0;
````

That will return true if wireguard is a module and is loaded, or if wireguard is a built-in. (It won't tell you whether wireguard is a module but isn't yet loaded, though.)

File: patches/openwrt/0025-wireguard-tools-allow-generating-private_key-v2.patch [hunk 7]
```diff
@@ -0,0 +1,62 @@
...
+ 	fi
+ }
+ 
++ensure_key_is_generated() {
++	local private_key
++	private_key="$(uci get network."$1".private_key)"
++
++	if [ "$private_key" == "generate" ]; then
++		local ucitmp
++		ucitmp="$(mktemp -d)"
```
zx2c4, line 43: If you're writing a private key into that directory, you might want to first run `umask 077`.

AiyionPrime:
I'm currently building the rebased PR,
will look for changes git missed tomorrow noon
and address your proposals around 3pm.

Thanks for your support!

package/gluon-mesh-vpn-core/files/lib/netifd/proto/gluon_wireguard.sh
```diff
@@ -0,0 +1,86 @@
...
+proto_gluon_wireguard_init_config() {
+	#no_device=1
+	#available=1
+	proto_config_add_int index
+	proto_config_add_int mtu
+}
+
+interface_linklocal_from_wg_public_key() {
+	# We generate a predictable v6 address
+	local macaddr="$(printf "%s" "$1"|md5sum|sed 's/^\(..\)\(..\)\(..\)\(..\)\(..\).*$/02:\1:\2:\3:\4:\5/')"
```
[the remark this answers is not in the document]
AiyionPrime: Can't change this lighthearted; the serverside would need adjustments as well.
@lemoer Your thoughts?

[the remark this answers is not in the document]
zx2c4: Looks like you're really only using 40 bits from it anyway (not the full 128), so it probably doesn't matter a lot either way.

package/gluon-mesh-vpn-wireguard/src/respondd.c
```diff
@@ -0,0 +1,262 @@
...
+
+end:
+	json_object_put(site);
+
+	return ret;
+}
+
+static bool wgpeerselector_enabled(void) {
+	bool enabled = true;
+	struct json_object *ret = json_object_new_object();
```
AiyionPrime, line 136:
```suggestion
```

package/gluon-mesh-vpn-core/files/lib/netifd/proto/gluon_wireguard.sh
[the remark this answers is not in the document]
lemoer: As we all agree that this doesn't matter much, I'll leave it as it is. We already have deployed nodes and causing breaking changes here doesn't seem necessary to me.

lemoer:
In order to prepare the rebase on master, I assessed what has to be done. Especially the changes from "mesh-vpn: Fully abstract VPN methods" (https://github.com/freifunk-gluon/gluon/pull/2102) raised the need to change a lot of stuff here. In order to prepare this, I split some commits here in *(Part A)* and *(Part B)*.

*(Part A)* = These parts are mostly unaffected by #2102.
*(Part B)* = These parts are affected by #2102 a lot. Most of this has to be rewritten.

I am not sure, whether this was 100% precise, but it gives an overview of how much code has to be rewritten.

package/gluon-mesh-vpn-wireguard/files/lib/gluon/mesh-vpn/wireguard_pubkey.sh
lemoer: No, it's in `/sbin/uci`.

package/gluon-mesh-vpn-wireguard/src/respondd.c
lemoer: True, this would be a bit more efficient. But this code is not that performance critical here.

lemoer:
I did some rebase efforts today. Now this PR ...
- ... is rebased on master.
- ... uses the infrastructure from #2102.
- ... is rebased on the merged version of #2163.
- ... is rebased on the current state of the wgpeerselector branch.
- ... contains most of the suggestions from @zx2c4. Some of his suggestions are left to be implemented. (Thanks your suggestions!)
- ... contains the memory leak fix suggested by @AiyionPrime.

There are some things left to be done:
- [ ] The patches for wireguard-tools and uci do not apply anymore.
- [ ] Some suggestions from@zx2c4 are left to be implemented.
- [ ] As I had to rewrite some of the patches, exhaustive testing is definitely needed.

package/gluon-mesh-vpn-wireguard/src/respondd.c
```diff
@@ -0,0 +1,248 @@
...
+	char *line = NULL;
+	size_t len = 0;
+	size_t skiplen = strlen(skip);
+
+	ssize_t r = getline(&line, &len, f);
+
+	pclose(f);
+
+	if (r >= 0) {
+		len = strlen(line); /* The len given by getline is the buffer size, not the string length */
```
NeoRaider, line 59: The `strlen` is unnecessary - while `len` is the buffer length, `r` contains the correct string length.

One missing case is that `line` may be NULL when `r` is 0, which happens when the command doesn't output anything (I think there are other places in Gluon where this isn't handled correctly).

package/gluon-mesh-vpn-wireguard/check_site.lua
```diff
@@ -0,0 +1,8 @@
+local function check_peer(k)
+	need_alphanumeric_key(k)
+
+	need_string_match(in_domain(extend(k, {'public_key'})), "^" .. ("[%a%d+/]"):rep(43) .. "=$")
```
lemoer, line 4:
```suggestion
	need_string_match(in_domain(extend(k, {'public_key'})), "^" .. ("[%a%d+/]"):rep(42) .. "[A|E|I|M|Q|U|Y|c|g|k|o|s|w|4|8|0]=$")
```

AiyionPrime (commented):
> wireguard-**tools**
Ah, stimmt ^^

AiyionPrime:
@rotanid this one currently has no merge conflict

package/gluon-mesh-vpn-wireguard/check_site.lua
```diff
@@ -0,0 +1,9 @@
+local function check_peer(k)
+	need_alphanumeric_key(k)
+
+	need_string_match(in_domain(extend(k,
+		{'public_key'})), "^" .. ("[%a%d+/]"):rep(42) .. "[A|E|I|M|Q|U|Y|c|g|k|o|s|w|4|8|0]=$")
```
NeoRaider, line 5:
```suggestion
		{'public_key'})), "^" .. ("[%a%d+/]"):rep(42) .. "[AEIMQUYcgkosw480]=$")
```

package/gluon-mesh-vpn-wireguard/src/respondd.c
```diff
@@ -0,0 +1,245 @@
...
+static struct json_object * get_wireguard_public_key(void) {
+	return stdout_read("exec /lib/gluon/mesh-vpn/wireguard_pubkey.sh", "", false);
+}
+
+static struct json_object * get_wireguard_version(void) {
+	return stdout_read("exec wg -v", "wireguard-tools ", true);
+}
+
+static bool wireguard_enabled(void) {
+	return access("/sys/module/wireguard", F_OK) == 0;
```
NeoRaider, line 92: `gluon-mesh-vpn-wireguard` (indirectly) depends on the kernel module, so this will always return true.

package/gluon-iptables-clamp-mss-to-pmtu/files/lib/gluon/mesh-vpn/iptables-mss.rules
```diff
@@ -1,3 +1,3 @@
 *mangle
--A FORWARD -o mesh-vpn+ -p tcp -m tcp --tcp-flags SYN,RST SYN -j TCPMSS --clamp-mss-to-pmtu
+-A FORWARD -o vpn+ -p tcp -m tcp --tcp-flags SYN,RST SYN -j TCPMSS --clamp-mss-to-pmtu
```
NeoRaider, line 2: `vpn` is not very descriptive. I don't think we should change `mesh-vpn`, let's fix the script instead.

package/gluon-mesh-vpn-core/luasrc/lib/gluon/upgrade/500-mesh-vpn
```diff
@@ -7,13 +7,15 @@ local util = require 'gluon.util'
...
 
 uci:section('network', 'interface', 'mesh_vpn', {
 	ifname = vpn_core.get_interface(),
-	proto = 'gluon_mesh',
+	proto = active_vpn.proto(),
 	transitive = true,
 	fixed_mtu = true,
-	macaddr = util.generate_mac(7),
+	macaddr = util.generate_mac(7), -- needed for proto gluon_mesh
+	index = 7,                      -- needed for proto gluon_wireguard
```
NeoRaider, line 18: We wouldn't need to change anything in this UCI section if `mesh-vpn` is already the VXLAN interface.

package/gluon-mesh-vpn-core/files/lib/netifd/proto/gluon_wireguard.sh
NeoRaider, line 20: The `md5sum` here feels extremely weird to me. Could the first bytes of the public key be used directly, without hashing, after decoding the base64? Are they distributed uniformly?

package/gluon-mesh-vpn-core/files/lib/netifd/proto/gluon_wireguard.sh
```diff
@@ -0,0 +1,86 @@
...
+	json_init
+	json_add_string name "vpn_wired"
+	json_add_string proto "gluon_wired"
+	json_add_string ifname "$ifname"
+	json_add_int index "$index"
+	json_add_boolean transitive 1
+	json_add_boolean vxlan 1
+	json_add_int mtu "$mtu"
+	json_add_boolean fixed_mtu 1
+	json_add_string vxpeer6addr 'fe80::1'
```
NeoRaider, line 71: For the sake of symmetry, I would prefer to derive this address from the peer's public key.

package/gluon-mesh-vpn-wireguard/files/lib/gluon/mesh-vpn/wireguard_pubkey.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+INCLUDE_ONLY=1
+. /lib/netifd/proto/wireguard.sh
+
+ensure_key_is_generated vpn
+iface="$(lua -e "vpn = require 'gluon.mesh-vpn'; print(vpn.get_interface())")"
+uci get "network.$iface.private_key" | /usr/bin/wg pubkey
```
NeoRaider, line 8: You are using a Linux ifname as the UCI section name here - this is just wrong. The UCI section is always called `mesh_vpn`.

NeoRaider: Oops, I see you are also introducing a new section based on `vpn_core.get_interface()` in a new upgrade script. Please don't do this, use static section names instead.

lemoer: The wireguard.sh proto determines it's interface name based on the uci section name:

See here: https://github.com/openwrt/openwrt/blob/master/package/network/utils/wireguard-tools/files/wireguard.sh#L117

package/gluon-mesh-vpn-core/files/lib/netifd/proto/gluon_wireguard.sh
awlx: This is also used on the server side to generate a predictable v6 address so I would refrain from changing this.

awlx: https://github.com/freifunkMUC/wgkex/blob/6cd3db50d2c133eea818a3631108cf676b121f17/wgkex/worker/netlink.py#L39

For example.

package/gluon-mesh-vpn-wireguard/files/lib/gluon/mesh-vpn/wireguard_pubkey.sh
NeoRaider: Let's call that interface something like `wg_mesh` then.

package/gluon-mesh-vpn-core/files/lib/netifd/proto/gluon_wireguard.sh
NeoRaider: Collisions don't matter, these are link-local addresses - so only the two addresses of a single tunnel must not collide.

Making this symmetrical by deriving the server address using the same algorithm (which I'd really like) requires changing the server side anyways.

lemoer: I think, collisions actually do matter.

IIRC, the vxlan code in linux can not handle link-local addresses, which do not have "a kernel wide" route. At least in *wireguard-vxlan-glue* we add such a route ([see here](https://github.com/freifunkh/wireguard-vxlan-glue/blob/master/netlink.py#L165-L171)). Also the vxlan.sh proto establishes a kernel wide route for `fe80::1` with the configuration from this PR:

```
root@crashtester:~# ip -6 route show fe80::1
fe80::1 dev vpn proto static metric 1024 pref medium
```

As we need "kernel wide routes" for the link-local addresses, the link-local addresses must be unique.

package/gluon-mesh-vpn-core/files/lib/netifd/proto/gluon_wireguard.sh
lemoer: In contrast to changing the address schema of the routers, this will not cause much headache w.r.t. migration of existing devices, because I guess, that we can simply configure both ip addresses (old and new) on the server-side to support old and new versions here.

So I will try to implement this.

package/gluon-mesh-vpn-wireguard/files/lib/gluon/mesh-vpn/wireguard_pubkey.sh
lemoer: This would imply interface names that are mesh-vpn provider dependent. While this would not cause much problems for most of the code, I have no idea yet how to solve this for the script [/lib/gluon/mesh-vpn/iptables-mss.rules](https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-iptables-clamp-mss-to-pmtu/files/lib/gluon/mesh-vpn/iptables-mss.rules#L2).

However, I would rather suggest to use `mesh_vpn` as interface name. This could be used for all mesh-vpn providers and would be descriptive as well.

package/gluon-mesh-vpn-core/files/lib/netifd/proto/gluon_wireguard.sh
lemoer: cc: @AiyionPrime

AiyionPrime: seen

package/gluon-mesh-vpn-core/luasrc/lib/gluon/upgrade/500-mesh-vpn
lemoer: I think, this will actually be a smart solution.

I suspect, this will become the third rewrite of the gluon_wireguard.sh netifd proto. And tbh, developing netifd protos is kind of pain. -.-

package/gluon-mesh-vpn-wireguard/luasrc/lib/gluon/upgrade/400-mesh-vpn-wireguard
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/lua
+
+local uci = require('simple-uci').cursor()
+local site = require 'gluon.site'
+local vpn_core = require 'gluon.mesh-vpn'
+
+local private_key = uci:get("network", vpn_core.get_interface(), "private_key")
+
+if not private_key or not private_key:match("^" .. ("[%a%d+/]"):rep(42) .. "[A|E|I|M|Q|U|Y|c|g|k|o|s|w|4|8|0]=$") then
```
lemoer, line 9:
```suggestion
if not private_key or not private_key:match("^" .. ("[%a%d+/]"):rep(42) .. "[AEIMQUYcgkosw480]=$") then
```

package/gluon-mesh-vpn-core/files/lib/netifd/proto/gluon_wireguard.sh
awlx: It's also the IP which is included in the `allowed-ips` of wireguard so it _must_ not collide otherwise wireguard doesn't work.

NeoRaider: Ah, it is very unfortunate that WG requires you to care about the addresses in the tunnel, and doesn't just create an interface per peer.

The VXLAN issue sounds weird, I'll look into that. Such a route should usually not make a difference.

awlx: It's actually the neat thing that we don't require a gazillion interfaces but just one for all peers :). Which makes handling the server side way easier. Because you can pre-create all interfaces via systemd, ifupdown or whatever and just add peers via netlink. So no dynamic interface creation thus no unmanaged devices.

NeoRaider: Yeah, I can see benefits for both configurations - OpenVPN does the same, requiring to configure addresses to be routed through the tunnel on the OpenVPN side in TUN mode.

I usually prefer to have this information in the Linux routing table, requiring a separate interface for each L3 tunnel (and if I get around to adding WG support to fastd, that is likely how I'm going to implement it, assigning a separate WG instance and UDP port to each tunnel).

That being said, I'm okay with the current implementation for wgpeerselector in Gluon, but the md5sum still looks weird...

package/gluon-iptables-clamp-mss-to-pmtu/files/lib/gluon/mesh-vpn/iptables-mss.rules
lemoer: Done.

package/gluon-mesh-vpn-core/luasrc/lib/gluon/upgrade/500-mesh-vpn
lemoer: Done.

package/gluon-mesh-vpn-core/files/lib/netifd/proto/gluon_wireguard.sh
lemoer: @NeoRaider I thought about this idea again. While it would make sense from a symmetry point of view, it would also complicate the code a lot.

The peers and their keys are currently only handled by wgpeerselector and the vxlan interfaces are created separately. The wgpeerselector doesn't even know about the upper vxlan interfaces. Therefore deriving the vxpeer6addr from the public_key of the currently active peer is not easy.

I would really like to keep the wgpeerselector code as simple as possible.

package/gluon-mesh-vpn-wireguard/files/lib/gluon/mesh-vpn/wireguard_pubkey.sh
lemoer: Done.

package/gluon-mesh-vpn-wireguard/src/respondd.c
lemoer: Done.

package/gluon-mesh-vpn-core/files/lib/netifd/proto/gluon_wireguard.sh
lemoer: I would like to keep the md5sum avoid ugly migrations of existing devices. Is this acceptable?

NeoRaider: Yes, let's keep it - we have other places in Gluon where we use md5sum to derive MAC addresses.

package/gluon-mesh-vpn-core/files/lib/netifd/proto/gluon_wireguard.sh
NeoRaider: Okay

AiyionPrime (commented):
I do not like the two unrelated software paths being merged like this.
As you said, wireguard is always enabled. Having wireguard.enabled reflect whether wgpeerselector is enabled does not seem intuitive to me.

NeoRaider:
@AiyionPrime 
> I do not like the two unrelated software paths being merged like this.
> As you said, wireguard is always enabled. Having wireguard.enabled reflect whether wgpeerselector is enabled does not seem intuitive to me.

As far as I can tell, this PR is the one "gluon-mesh-vpn-wireguard" implementation that is likely to be merged eventually; I'm not sure what "two unrelated software paths" you're referring to.

Generally a package `gluon-mesh-vpn-X` contains a VPN provider "X", which creates a respondd reply section "X". So having `wireguard.enabled` reflect whether `gluon-mesh-vpn-wireguard` is enabled makes sense to me.

If the package were called `gluon-mesh-vpn-wgpeerselector`, this should be reflected in the overall naming of the files and data structures; is the question of package naming what you want to discuss?

AiyionPrime:
For both fastd and tunneldigger the wrapping code is pretty straightforward.
`wgpeerselector` is complex enough to even have an acknowledged name.
We considered wireguard and `wgpeerselector` two different pieces of code; I don't see what has changed with that.

> I'm not sure what "two unrelated software paths" you're referring to.
In my eyes the two pieces of software this PR adds could be very well reflected in different respondd paths.

The package naming is not what I want to discuss.

NeoRaider:
Hmm, I see what you mean. Putting the field "enabled" under "software" together with the version number isn't great; what is enabled is not the software, but some Gluon VPN provider which internally uses the software.

However, this is how it's been implemented for fastd, and we don't want to change that. Having the `software.wireguard.enabled` field reflect whether the `gluon-mesh-vpn-wireguard` VPN provider is enabled or not isn't *too* bad IMO, and there is some value in symmetry.

Considering this, and also the naming of the other providers, renaming the whole package might actually be a good idea. `gluon-mesh-vpn-tunneldigger` was named after the software after all, and not `gluon-mesh-vpn-l2tp` - which was a good call, as there is now fastd with L2TP support as well. Similarly, `wgpeerselector` is a software establishing Wireguard tunnels, but there may be other software doing the same in the future (I'm considering adding this to fastd as well).

So my choice would be to rename this package to `gluon-mesh-vpn-wgpeerselector`, but I'd like to hear a few more opinions on this before someone actually does the work.

lemoer:
For marketing reasons I would prefer `gluon-mesh-vpn-wireguard`. It is more concise and self-explanatory.

However, I would also agree with `gluon-mesh-vpn-wgpeerselector`.

mweinelt:
I support the `gluon-mesh-vpn-wireguard` name.

Whether you name it `-wireguard` or `-wgpeerselector` is IMO a moot discussion, since both software packages are used, and ideally wgpeerselector can simply be considered an implementation detail to get us wireguard support.

NeoRaider:
Okay, `gluon-mesh-vpn-wireguard` it is. I'll have another look at the latest changes in the next days.

awlx:
Btw, we did some heavy lifting on the broker site of things the last days:
https://github.com/freifunkMUC/wgkex/releases

More features, less hardcoded stuff etc. and finally we are able to switch domains without a cleanup first. Also automatic cleanups run now every hour.

package/gluon-mesh-vpn-wireguard/luasrc/usr/lib/lua/gluon/mesh-vpn/provider/wireguard.lua
```diff
@@ -0,0 +1,37 @@
...
+	uci:set('network', 'wg_mesh', 'disabled', not val)
+	uci:save('network')
+end
+
+function M.active()
+	return site.mesh_vpn.wireguard() ~= nil
+end
+
+function M.set_limit(ingress_limit, egress_limit)
+	-- TODO: Test that limiting this via simple-tc here is correct!
```
lemoer, line 23: Check this.

patches/openwrt/0005-package-uci-backport-cli-add-option-for-changing-save-path.patch
```diff
@@ -0,0 +1,83 @@
+From: Leonardo Mörlein <me@irrelefant.net>
+Date: Sat, 16 Jan 2021 23:11:01 +0100
+Subject: package/uci: backport: "cli: add option for changing save path"
+
+This is a backport of
+
+https://git.openwrt.org/?p=project/uci.git;a=commit;h=4b3db1179747b6a6779029407984bacef851325c
```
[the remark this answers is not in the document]
lemoer: No, not yet.

PR #2169, opened 2021-01-02
actions: run tasks based on set of modified paths

mweinelt:
Builds are often unnecessarily queuing up, so this is basically what I had in mind in #2004 ported to actions.

### DIFF
--- a/.github/workflows/build-gluon.yml
+++ b/.github/workflows/build-gluon.yml
@@ -9,8 +9,24 @@ on:
       - master
       - next
       - v20*
+    paths:
+      - "modules"
+      - "Makefile"
+      - "scripts/**"
+      - "package/**"
+      - "patches/**"
+      - "targets/**"
+      - ".github/workflows/build-gluon.yml"
   pull_request:
     types: [opened, synchronize, reopened]
+    paths:
+      - "modules"
+      - "Makefile"
+      - "scripts/**"
+      - "package/**"
+      - "patches/**"
+      - "targets/**"
+      - ".github/workflows/build-gluon.yml"
 jobs:
   build_firmware:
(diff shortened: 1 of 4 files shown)


.github/workflows/build-gluon.yml
```diff
@@ -9,8 +9,24 @@ on:
...
   pull_request:
     types: [opened, synchronize, reopened]
+    paths:
+      - "modules"
+      - "Makefile"
+      - "scripts/**"
+      - "package/**"
+      - "patches/**"
+      - "targets/**"
+      - ".github/workflows/build-gluon.yml"
```
blocktrron, line 29: This needs to be added to the generator script,

.github/workflows/check-patches.yml
```diff
@@ -1,8 +1,17 @@
+---
```
blocktrron, line 1: Not a full-time YAML wizard, is this added for a reason or is it by accident?

mweinelt: This is the usual YAML header.

.github/workflows/build-gluon.yml
mweinelt: Ah, yeah. This was a quick amend after the initial push, since I missed it. And then I did it wrong.

mweinelt: Except I did dd it here: https://github.com/freifunk-gluon/gluon/pull/2169/files#diff-c73615de12add5f5e9f1be4721edeaf5999c2e198b04fafd2baf636c92cbcec5. Did you miss that?

blocktrron: obviously.

PR #2171, opened 2021-01-04
gluon-config-mode-core: serve config mode in normal mode

lemoer:
This commit starts a second uhttpd instance bound to localhost:81 which
serves the config mode during normal runtime. This is useful to change
settings via ssh port forwarding:

    ssh routername -L 12345:127.0.0.1:81

Then you can access the config mode on your computer using
http://localhost:12345/.

### DIFF
--- a/package/gluon-config-mode-core/files/etc/init.d/gluon-config-mode
+++ b/package/gluon-config-mode-core/files/etc/init.d/gluon-config-mode
@@ -1,0 +1,14 @@
+#!/bin/sh /etc/rc.common
+
+START=50
+
+USE_PROCD=1
+
+UHTTPD_BIN="/usr/sbin/uhttpd"
+
+start_service() {
+	procd_open_instance
+	procd_set_param respawn
+	procd_set_param command "$UHTTPD_BIN" -f -h /lib/gluon/config-mode/www -x /cgi-bin -A 1 -R -p 127.0.0.1:81
+	procd_close_instance
+}
--- a/package/gluon-config-mode-core/files/usr/lib/autoupdater/abort.d/70gluon-config-mode
+++ b/package/gluon-config-mode-core/files/usr/lib/autoupdater/abort.d/70gluon-config-mode
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+. /lib/gluon/autoupdater/lib.sh
+
+
+start_enabled gluon-config-mode
--- a/package/gluon-config-mode-core/files/usr/lib/autoupdater/download.d/30gluon-config-mode
+++ b/package/gluon-config-mode-core/files/usr/lib/autoupdater/download.d/30gluon-config-mode
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+. /lib/gluon/autoupdater/lib.sh
+
+
+stop gluon-config-mode


rotanid:
so the additional instance is running all the time?
i find that a bit too much in terms of memory consumption, considering 32MB devices and 99,9% (just a guess) of people probably not using this

Adorfer:
just enable it for 64/128MB-RAM devices then.

CodeFetch:
>  considering 32MB devices

Reminds me of:
```
QUESTION: I read in a newspaper that in 1981 you said, ``640K of memory should
be enough for anybody.'' What did you mean when you said this?

ANSWER: I've said some stupid things and some wrong things, but not that. No
one involved in computers would ever say that a certain amount of memory is
enough for all time.

The need for memory increases as computers get more potent and software gets
more powerful. In fact, every couple of years the amount of memory address
space needed to run whatever software is mainstream at the time just about
doubles. This is well-known.

When IBM introduced its PC in 1981, many people attacked Microsoft for its
role. These critics said that 8-bit computers, which had 64K of address space,
would last forever. They said we were wastefully throwing out great 8-bit
programming by moving the world toward 16-bit computers.

We at Microsoft disagreed. We knew that even 16-bit computers, which had 640K
of available address space, would be adequate for only four or five years. (The
IBM PC had 1 megabyte of logical address space. But 384K of this was assigned
to special purposes, leaving 640K of memory available. That's where the
now-infamous ``640K barrier'' came from.)

A few years later, Microsoft was a big fan of Intel's 386 microprocessor chip,
which gave computers a 32-bit address space.

Modern operating systems can now take advantage of that seemingly vast
potential memory. But even 32 bits of address space won't prove adequate as
time goes on.

Meanwhile, I keep bumping into that silly quotation attributed to me that says
640K of memory is enough. There's never a citation; the quotation just floats
like a rumor, repeated again and again. 
```

mweinelt:
I don't think this is a good investment of our resources. People who can `ssh -L` can usually also executes scripts like `start-config-mode`. So how about a script that starts it up on demand?

lemoer:
I am currently working on a web service, which uses ssh tunnels to provide remote access to the config-mode of gluon nodes. While this is relatively easy when I just need to set up an ssh tunnel, it gets much more complicated when I also need to run commands. Therefore I do not like the suggestion of @mweinelt . Also the suggestion of @Adorfer is not very convenient, because this would mean that the web service can not be used to administrate 32MB devices.

To put a number on this: We are talking about 160K memory footprint, if I interpreted the numbers correctly.

Another way to save the memory would be to use the existing uhttpd instance. Currently this instance is only used for the status-page. We could move the config-mode into the same path. As the status-page is relatively public, there is no access control yet. To avoid leaking the config-mode to the public internet, we could add an early check into its lua cgi scripts to only deliver/work, when the requesting ip is `::1`.

What do you think?

mweinelt:
I think such a solution with its external service dependency is probably better kept in our community-packages.

CodeFetch:
@mweinelt It does not depend on an external service.
My idea was that one could also create a python program which offers a Webserver and SSH to HTTP proxy automatically even for Windows as a standalone executable using py2exe. This way one can use the controller securely even without HTTPS. But I don't think this is a problem at all as a user could install his or her own DynDNS daemon on the router and setup his domains without any security implications.

CodeFetch:
@mweinelt By the way the solution lemoer suggests uses a decentral controller solution running on the node itself. There is really no external service needed.

mweinelt:
> I am currently working on a web service, which uses ssh tunnelling to provide remote access to the config-mode of gluon nodes.

That does sound very much like an external dependency, that will be its primary consumer.

CodeFetch:
@mweinelt lemoer is working on a decentral controller running on a Gluon node: https://github.com/lemoer/gluon/tree/gluon-controller
To securely use his controller either a PSK-secured secondary WLAN network needs to be used or HTTPS needs to be set up using a valid certificate to access the controller login.

T-X:
Thanks for the edit @CodeFetch, that makes it easier to understand.

But I also don't understand, if the controller Gluon node has all the SSH keys and therefore in theory full root access, why can't it just enable the localhost config mode on those nodes on demand? For 160 kB it seems worth the effort, as we still have Gluon nodes with 32 MB RAM running at around 85% memory usage today.

blocktrron:
Discussed with @lemoer at the RC3 meetup - the focus for his goal will shift towards an API, thus serving the config mode at runtime is not necessary.

mkg20001:
I've made this a community package for the time being, since graz is using this (ffgraz-config-mode-at-runtime)

rubo77:
@mkg20001 wehre?

PR #2172, opened 2021-01-07
Allow Users to Customize Firmwares

lemoer:
The idea of this PR is to intruduce a way for users that are normally not
familiar with all the details of the site configuration to preconfigure
images before flashing. This will tackle common usecases like baking in
authorized ssh keys, skipping config mode, enabling mesh on lan, and so on.

The new feature can be made possible in a safe way, since some of our site
options can be used in such a manner, that the resulting firmware is not broken
if they are misused. Or at least, defects are not that harmful and do not cause
silent bugs. Still the PR includes a mechanism that uses our
established *check_site* framework to verify this custom configuration.

Some technical details:
- A modified OpenWrt Imagebuilder is generated as gluon output.
- The necessary packages are built into the imagebuilder, so it is standalone.
- Users can download the imagebuilder.
- A `custom.json` can be added in the main dir of the imagebuilder.
    - This config may contain some selected *site* options.
    - It will be built into the image as */lib/gluon/custom.json*
    - It is merged with site and domain config.
    - It has to pass the site checks.
    - Only varibles classified with `in*_custom` are allowed.
- The build is invoked with *./build_gluon.sh PROFILE*.
    - Only profiles allowed in gluon can be built.
    - Calling *./build_gluon.sh* without arguments lists all available profiles.

Currently the following site options are allowed in the *custom.json*:
- default_domain
- authorized_keys
- mesh_on_wan
- mesh_on_lan
- poe_passthrough
- setup_mode.skip
- roles.default

As development progresses, web applications similar to http://meshkit.freifunk.net/
could be created, that allow users, without even downloading the imagebuilder,
to obtain customized firmwares that include their ssh keys or skip config mode.

Currently building a single profile takes only a couple of seconds:

    time sh build_gluon.sh tl-wdr4300-v1
    ...
    Calculating checksums...

    ________________________________________________________
    Executed in   10.52 secs   fish           external
       usr time    8.62 secs  584.00 micros    8.62 secs
       sys time    3.22 secs   46.00 micros    3.22 secs

State of this PR:
- It is already usable as a proof of concept.
- Suggestions and improvements welcome.
- Tested on *x86-64*.
- Compile tested on *ar71xx-generic*.

Note:
- This is not about building gluon with the upstream OpenWrt Imagebuilder.
- Also, adding arbitrary additional files or packages as normally used in
  the imagebuilder is explicitly unsupported.


lemoer:
It's no longer visible why the check "Lint / Shell" failed. The check is not failing on my laptop, so I think it's fine...

NeoRaider:
The hacked IB approach is something Gluon did a few years ago, and I was extremely happy to get rid of it. Let's not go there again - the current build code is *already* a maintenance nightmare. The fact that you need to pass `DEVICE_TYPE=other` also hints at the fact that the resulting images will not have the correct selection of packages at all (I'm not sure how the per-device rootfs of the OpenWrt buildsystem interacts with the IB).

That said, I like the idea of allowing a simplified preconfig of images. It's something I've wanted for a long time, but I haven't found a workable approach yet...

I'm not sure what to think about the addition of a third flavor of site config validation in addition to "site" and "domain", but I guess this can't be avoided completely. Instead of "custom", I'd prefer the term "local" (this is what OpenEmbedded/Yocto uses for config local to a single build).

We should also discuss whether we really want to expose the additional config as JSON. JSON is a good format for data that is primarily read and written by software, but it's cumbersome to write for humans - something like TOML or our existing Lua format are better suited for that purpose.

CodeFetch:
> We should also discuss whether we really want to expose the additional config as JSON. JSON is a good format for data that is primarily read and written by software, but it's cumbersome to write for humans - something like TOML or our existing Lua format are better suited for that purpose.

My experience with the site.conf format was that every third build or so failed because of a typo or something. JSON is a standard format and there is good support for it in text editors and there are editors specifically for that purpose. With the custom Lua site.conf format there isn't even syntax highlighting without explicitly selecting Lua as highlighting language. Thus I think quite the opposite is the case and the site.conf should be changed to JSON format and renamed to site.json.

lemoer:
@NeoRaider 

> The hacked IB approach is something Gluon did a few years ago, and I was extremely happy to get rid of it. Let's not go there again - the current build code is already a maintenance nightmare.

Please elaborate: What were (and are) the problems with the IB?

> That said, I like the idea of allowing a simplified preconfig of images. It's something I've wanted for a long time, but I haven't found a workable approach yet...

As long as you want to create (or modify) an binary image, you need all the tools of OpenWrt that create valid images. For example, it is impossible to modify only a specific block in the image binary, as this will most likely invalidate the checksum of the image. This is something we cannot work around, as the checksums are usually validated by the vendor firmware as well.

To me, it seemed quite obvious, that the most efficient way to make use of the existing image creation tools of OpenWrt is the IB. As you did not explain the issues of the IB yet, I still have this opinion.

Maybe I will change my mind, if you elaborate on the issues of IB a bit. Especially I am wondering, if we can't just fix the issues, that we have with the IB upstream? Or if the problems are target dependent, we could e.g. limit the targets that allow customized firmwares to the ones that have proper IB support?

NeoRaider:
One thing we need to verify is that the content of the images generated by the image builder are identical to the regular Gluon images with the exception of the custom config file. As mentioned, I find it unlikely that this is currently the case, as you have to pass `DEVICE_TYPE=other`, which reminds me of hacks that were used in earlier Gluon versions.

Another thing I'm wondering about is if this feature will be used at all if it requires building the image builder in addition to the regular Gluon build. It is already doubtful whether the opkg repo generation feature sees much use - I assume many communities will just put the firmware images on their webserver and nothing else.

Incremental builds with modified site.conf are already very fast, so the whole thing would only be useful when someone who is not the regular firmware builder of a community wants to build modified images... But this use case would have to be anticipated by the firmware builder, as just dropping the IB archives on the webserver is just a huge waste of space when it it not needed.

CodeFetch:
@NeoRaider The basic idea is to have something like http://meshkit.freifunk.net/. The approach could be extended to allow communities which use "plain Gluon" to not even have to build Gluon on their own. So we could make a community-independent Gluon image builder which saves a lot of processing and storage ressources if that is a serious concern.

lemoer:
Indeed, I am aiming at something similar to the Meshkit webapp. In my vision, this app should be usable for end users with little to no experience. But the second part, that @CodeFetch mentioned is not currently one of my aims. I do not want to focus on building images with custom *site.conf* or *domain.conf* yet. While I agree, that this would be nice, this is not part of my current efforts. We can pick this up later, if we already support building images with something like a *custom.conf*.

The [asu project](https://github.com/aparcar/asu) allows to build images with the imagebuilder via Web API. From there, it is merely a SPA what we need, that generates a *custom.conf* and triggers builds in the asu server. Each community could be set up their own asu server+SPA instance. That way, users of gluon wouldn't even have to get the imagebuilder running on their own computers. They could simply use the imagebuilder instance hosted by the community.

@NeoRaider I understand your idea, that the images should be identical. I think, ensuring this is a good idea. And of course, we should try to get rid of the "hacks" and try to implement our requirements into the imagebuilder upstream. But I am not sure about the way, how we get there yet.

A few questions:
- What is the desired level of equality of the images?
    - Do we want to have identical *.bin files? Or is it enough, if the file system contents are identical?
    - Do have suggestions/experience how to "diff" the results?
- @blocktrron mentioned, there may be other issues. He mentioned something with "non-shared packages that are built in phase 2 and are not kernel-dependent" and that packages "that are not non-shared" are thrown out in phase 1. But I do not really understand this. Maybe someone could explain this a little more in detail?
- Are there any other problems that you see coming?
- Are there any resources or code, which you suggest to understand the build process and the imagebuilder in enough detail to finish this project?

CodeFetch:
> I understand your idea, that the images should be identical.

From what I know OpenWrt's reproducible builds are broken atm and nobody cares. Thus I don't understand why that is a requirement. What is the aim? When I add a modified custom.json to that image, it will have a different checksum.

NeoRaider:
I don't care about binary reproducibility. The aim is to have the same packages installed in the images. `DEVICE_TYPE` modifies the package selection, so if you need to set that, something is broken.

package/gluon.mk
```diff
@@ -20,8 +20,24 @@ shell-unescape = $(subst @1,@,$(subst @2,$(newline),$(1)))
...
+# Note that all variables in GluonCheckSite except for $(1) are expanded when
+# opkg calls the post-install scripts, not during the package build.
 define GluonCheckSite
-[ -z "$$IPKG_INSTROOT" ] || "${TOPDIR}/staging_dir/hostpkg/bin/lua" "${TOPDIR}/../scripts/check_site.lua" <<'END__GLUON__CHECK__SITE'
+if [ -d "$${TOPDIR}/gluon" ]; then
+	GLUON_DIR="$${TOPDIR}/gluon"
+else
+	GLUON_DIR="$${TOPDIR}/.."
+fi
+[ -z "$$IPKG_INSTROOT" ] || LUA_CPATH="$${TOPDIR}/staging_dir/hostpkg/lib/lua/5.1/?.so" "$${TOPDIR}/staging_dir/hostpkg/bin/lua" "$${GLUON_DIR}/scripts/check_site.lua" <<'END__GLUON__CHECK__SITE'
```
lemoer, line 40: The `"$${GLUON_DIR}/scripts/check_site.lua"` part has to be rebased onto master.

patches/openwrt/0018-imagebuilder-hacky-way-to-use-ib-for-gluon.patch
```diff
@@ -0,0 +1,46 @@
...
+--- a/target/imagebuilder/Makefile
++++ b/target/imagebuilder/Makefile
+@@ -35,6 +35,11 @@ $(BIN_DIR)/$(IB_NAME).tar.xz: clean
+ 		$(TMP_DIR)/.packageinfo \
+ 		$(PKG_BUILD_DIR)/
+ 
++	mkdir -p $(PKG_BUILD_DIR)/gluon
++	$(CP) -L \
++		$(TOPDIR)/../scripts \
++		$(PKG_BUILD_DIR)/gluon/
```
lemoer, line 20: I suppose, this will become unnecessary, since only `check_site.lua` was used and now (on the master branch) it is built into `gluon-core`. But this should be cross checked.

package/gluon.mk
lemoer: Done.

patches/openwrt/0018-imagebuilder-hacky-way-to-use-ib-for-gluon.patch
lemoer: Done.

lemoer:
I rebased this PR and fixed the `DEVICE_TYPE=other` issue. So `DEVICE_TYPE=other` is no longer used or required.

I am currently building to test this. I will report if this is successful.

package/gluon.mk
```diff
@@ -20,6 +20,14 @@ shell-unescape = $(subst @1,@,$(subst @2,$(newline),$(1)))
...
 
 
+# As of now, we need to set the LUA_CPATH variable. This would not be necessary
+# for the normal gluon build but is necessary for the imagebuilder.
+# Note that all variables in GluonCheckSite except for $(1) are expanded when
+# opkg calls the post-install scripts, not during the package build.
+#
+# [ -z "$$STAGING_DIR_HOSTPKG" ] || export LUA_CPATH="$${STAGING_DIR_HOSTPKG}/lib/lua/5.1/?.so"
+#
+# TODO: But maybe we do not need it anymore, since PATH is now altered?
```
lemoer, line 30: TODO: Check this.

lemoer:
I just created https://github.com/openwrt/openwrt/pull/4554 to upstream some of the logic in *build_gluon.sh*.

aparcar:
> From what I know OpenWrt's reproducible builds are broken atm and nobody cares. 

What gives you this impressions?

lemoer:
@NeoRaider I have fixed the problems you mentioned. Can we move on? What is missing from your perspective?

blocktrron:
Discussed with @lemoer at RC3 meetup

lemoer:
I will summarize:
- We currently do not want to maintain two different ways to build images.  This would create too much test efforts.
- A "local.conf" can introduced that allows to override everything that is allowed in site.conf and domain.conf.

PR #2177, opened 2021-01-24
gluon-web-private-wifi: define ifname for WAN radio

blocktrron:
Set the ifname for the WAN radio (Private WLAN) to wanX, X being the
radio index.

All other radios created by Gluon already have their ifname defined
following this pattern.


NeoRaider:
Makes sense. I wonder if we should also fix this up in an upgrade script.

blocktrron:
Added an upgrade script. Not tested yet, will report back.

PR #2178, opened 2021-01-25
kernel: bridge: Fix a deadlock when enabling multicast snooping

T-X:
[ Upstream commit 851d0a73c90e6c8c63fef106c6c1e73df7e05d9d ]

From: Joseph Huang \<Joseph.Huang@garmin.com\>

When enabling multicast snooping, bridge module deadlocks on multicast_lock
if 1) IPv6 is enabled, and 2) there is an existing querier on the same L2
network.

The deadlock was caused by the following sequence: While holding the lock,
br_multicast_open calls br_multicast_join_snoopers, which eventually causes
IP stack to (attempt to) send out a Listener Report (in igmp6_join_group).
Since the destination Ethernet address is a multicast address, br_dev_xmit
feeds the packet back to the bridge via br_multicast_rcv, which in turn
calls br_multicast_add_group, which then deadlocks on multicast_lock.

The fix is to move the call br_multicast_join_snoopers outside of the
critical section. This works since br_multicast_join_snoopers only deals
with IP and does not modify any multicast data structures of the bridge,
so there's no need to hold the lock.

Steps to reproduce:
1. sysctl net.ipv6.conf.all.force_mld_version=1
2. have another querier
3. ip link set dev bridge type bridge mcast_snooping 0 && \
   ip link set dev bridge type bridge mcast_snooping 1 < deadlock >

A typical call trace looks like the following:

[  936.251495]  _raw_spin_lock+0x5c/0x68
[  936.255221]  br_multicast_add_group+0x40/0x170 [bridge]
[  936.260491]  br_multicast_rcv+0x7ac/0xe30 [bridge]
[  936.265322]  br_dev_xmit+0x140/0x368 [bridge]
[  936.269689]  dev_hard_start_xmit+0x94/0x158
[  936.273876]  __dev_queue_xmit+0x5ac/0x7f8
[  936.277890]  dev_queue_xmit+0x10/0x18
[  936.281563]  neigh_resolve_output+0xec/0x198
[  936.285845]  ip6_finish_output2+0x240/0x710
[  936.290039]  __ip6_finish_output+0x130/0x170
[  936.294318]  ip6_output+0x6c/0x1c8
[  936.297731]  NF_HOOK.constprop.0+0xd8/0xe8
[  936.301834]  igmp6_send+0x358/0x558
[  936.305326]  igmp6_join_group.part.0+0x30/0xf0
[  936.309774]  igmp6_group_added+0xfc/0x110
[  936.313787]  __ipv6_dev_mc_inc+0x1a4/0x290
[  936.317885]  ipv6_dev_mc_inc+0x10/0x18
[  936.321677]  br_multicast_open+0xbc/0x110 [bridge]
[  936.326506]  br_multicast_toggle+0xec/0x140 [bridge]

Fixes: 4effd28c1245 ("bridge: join all-snoopers multicast address")
Signed-off-by: Joseph Huang \<Joseph.Huang@garmin.com\>
Acked-by: Nikolay Aleksandrov \<nikolay@nvidia.com\>
Link: https://lore.kernel.org/r/20201204235628.50653-1-Joseph.Huang@garmin.com
Signed-off-by: Jakub Kicinski \<kuba@kernel.org\>
Signed-off-by: Greg Kroah-Hartman \<gregkh@linuxfoundation.org\>
[linus.luessing@c0d3.blue: backported to 4.4]

### DIFF
--- a/patches/openwrt/0024-kernel-bridge-Fix-a-deadlock-when-enabling-multicast-snooping.patch
+++ b/patches/openwrt/0024-kernel-bridge-Fix-a-deadlock-when-enabling-multicast-snooping.patch
@@ -1,0 +1,275 @@
+From: Linus Lüssing <linus.luessing@c0d3.blue>
+Date: Mon, 25 Jan 2021 01:38:19 +0100
+Subject: kernel: bridge: Fix a deadlock when enabling multicast snooping
+
+[ Upstream commit 851d0a73c90e6c8c63fef106c6c1e73df7e05d9d ]
+
+From: Joseph Huang <Joseph.Huang@garmin.com>
+
+When enabling multicast snooping, bridge module deadlocks on multicast_lock
+if 1) IPv6 is enabled, and 2) there is an existing querier on the same L2
+network.
+
+The deadlock was caused by the following sequence: While holding the lock,
+br_multicast_open calls br_multicast_join_snoopers, which eventually causes
+IP stack to (attempt to) send out a Listener Report (in igmp6_join_group).
+Since the destination Ethernet address is a multicast address, br_dev_xmit
+feeds the packet back to the bridge via br_multicast_rcv, which in turn
+calls br_multicast_add_group, which then deadlocks on multicast_lock.
+
+The fix is to move the call br_multicast_join_snoopers outside of the
+critical section. This works since br_multicast_join_snoopers only deals
+with IP and does not modify any multicast data structures of the bridge,
+so there's no need to hold the lock.
+
+Steps to reproduce:
+1. sysctl net.ipv6.conf.all.force_mld_version=1
+2. have another querier
+3. ip link set dev bridge type bridge mcast_snooping 0 && \
+   ip link set dev bridge type bridge mcast_snooping 1 < deadlock >
+
+A typical call trace looks like the following:
+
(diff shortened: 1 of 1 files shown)


blocktrron:
Are you aware if this is staged for being backported to stable?

T-X:
Yep. This patch was initially added with Linux v5.10. It was then backported in v5.9.16 and in v5.4.85.

The issue was introduced in v5.1, 4effd28c1245 ("bridge: join all-snoopers multicast address") ). Which I had backported to Linux v4.4 for Gluon in b2add48d ("kernel: bridge: Implement MLD Querier wake-up calls / Android bug workaround"). Therefore I backported this patch from v5.4.85 down to v4.4, too.

blocktrron:
Thanks for the explanation. I assume you mean 4.14 instead of 4.4?

blocktrron (approved):

mweinelt:
I assume this needs to be ported to v2020.2.x, since both master and v2020.2.x rely on OpenWrt 19.07?

T-X:
@blocktrron yes, sorry, I ment Linux 4.14, correct.

@mweinelt: nope, Gluon v2020.2.x does not have the commit which introduced the issue
("bridge: join all-snoopers multicast address"). It's only in master right now. So no backport to v2020.2.x needed.

PR #2179, opened 2021-02-03
build: add support for multiple mmodule sources

blocktrron:
This PR adds support for multiple module sources to Gluon. This allows a module to be obtained from a different git source in case it cannot be obtained from the primary source. It preserves full backwards compatibility for existing module files contained in site repositories.

### Module availability

As it became [apparent](https://github.com/openwrt/openwrt/pull/3701#issuecomment-770882601) that the availability of the OpenWrt GitHub mirror is not guaranteed and not under full control of the project, a way to ensure Gluon can be built even if such a situation occurs should be provided. As git.openwrt.org is under full control of the project, this is a suitable source.

On the other hand Gluon makes extensive usage of the GitHub actions service. As it was discovered multiple times in the past, git.openwrt.org has a rate limit which we frequently hit when using the Actions CI, forcing us to switch to the GitHub mirror in the first place. With this PR, this is not an issue, as the clone happens primarily from GitHub and only falls back to git.openwrt.org in case the GitHub mirror is not available.

### Building Gluon in IPv6 only environments

The git.openwrt.org server is IPv6 capable and all modules apart from Gluons packages feed can be obtained from there. Add a GitHub IPv6 mirror service as a fallback source for the Gluon packages feed in order to support building Gluon in IPv6 only environments

I yet have to verify the last part, as Gluon downloads it's sources from a lot of different sources. In theory, the IPv6 sources mirror is capable and should serve as a fallback in case the primary source for a package does not support IPv6. I'll do a test build on an IPv6 only machine in the coming days.

### DIFF
--- a/modules
+++ b/modules
@@ -1,14 +1,14 @@
 GLUON_FEEDS='packages routing gluon'
 
-OPENWRT_REPO=https://github.com/openwrt/openwrt.git
+OPENWRT_REPO='https://github.com/openwrt/openwrt.git https://git.openwrt.org/openwrt/openwrt.git'
 OPENWRT_BRANCH=openwrt-19.07
 OPENWRT_COMMIT=a7a207e18bf7fa04f265bb95cbe6fa91561fbfe8
 
-PACKAGES_PACKAGES_REPO=https://github.com/openwrt/packages.git
+PACKAGES_PACKAGES_REPO='https://github.com/openwrt/packages.git https://git.openwrt.org/feed/packages.git'
 PACKAGES_PACKAGES_BRANCH=openwrt-19.07
 PACKAGES_PACKAGES_COMMIT=2974079d3db786fe5da00c10f1d80e79b0112093
 
-PACKAGES_ROUTING_REPO=https://github.com/openwrt-routing/packages.git
+PACKAGES_ROUTING_REPO='https://github.com/openwrt-routing/packages.git https://git.openwrt.org/feed/routing.git'
 PACKAGES_ROUTING_BRANCH=openwrt-19.07
 PACKAGES_ROUTING_COMMIT=02b4dbfcb7b8f8b566940847d22d5a6f229d2e66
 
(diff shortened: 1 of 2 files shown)


NeoRaider:
I don't really agree with the motivation:
- Module availablilty: `make update` is usually run after `git pull`, which requires access to Github anyways. Network access is already only needed when the `modules` file has changed since the last `make update` and the referenced commits aren't available locally yet.
- For IPv6-only environments, a simple `insteadOf` in ~/.gitconfig or /etc/gitconfig is much more effective, and avoids multiple tries until a working entry in the mirror list is found.

mweinelt:
> - For IPv6-only environments, a simple insteadOf in ~/.gitconfig or /etc/gitconfig is much more effective, and avoids multiple tries until a working entry in the mirror list is found.

Requiring everyone who wants to benefit from this to configure their Git accordingly is not very user-friendly. It requires that users discover such features in Git exist in the first place and then they have to know they can use git.openwrt.org as a stand-in. All this instead of things just working.

NeoRaider:
> > * For IPv6-only environments, a simple insteadOf in ~/.gitconfig or /etc/gitconfig is much more effective, and avoids multiple tries until a working entry in the mirror list is found.
> 
> Requiring everyone who wants to benefit from this to configure their Git accordingly is not very user-friendly. It requires that users discover such features in Git exist in the first place and then they have to know they can use git.openwrt.org as a stand-in. All this instead of things just working.

`insteadOf` will be needed on IPv6-only systems anyways, for example for https://github.com/freifunk-gluon/packages/tree/master/utils/libbabelhelper, which isn't in standard OpenWrt and as such will never end up on the official source mirrors. To me, having an early failure during `make update` instead of somewhere down the line sounds actually *more* user-friendly. That would also be an argument for keeping `github.com` as our module source: IPv4 internet access may be required for later source downloads anyways.

PR #2183, opened 2021-02-20
gluon-mesh-batman-adv: Use source routing for br-client

### LINKED ISSUE #2180: Bug: Gluon Routes its Packets to Wrong IPv6 Default Routers
### Bug report

**What is the problem?**

Router advertisement servers advertise *prefixes* together with a *default router* in layer 2 networks. There may be multiple router advertisement servers in one layer 2 network. Each of those router advertisement servers can announce different *prefixes* and a different *default router*.

A client can accept multiple of the advertised *prefixes* at the same time. If it does that, it installs an address in each of the advertised prefixes. It also has to install default routes from the router advertisements. Since default routers are not expected to be able to forward ip packets from arbitrary source addresses, the client should install routes to all *default routers* and use the correct one depending on which of the installed addresses it is sending from.

A gluon node receives all router advertisements on the `br-client`. It accepts all *prefixes* and therefore installs one address in each received *prefix*. But it installs just one *default route*. Therefore it sends all packets to the same *default router*. As this *default router* is not the designated *default router* for all prefixes, it may happen that the packet is dropped because of its source ip. This means that the gluon node will not be publicly reachable on some of its addresses.

Gluon nodes should either:
- accept only one router advertisement at a time,
- or should make sure, that when they are sending packets, they use the correct default route based on the source address.

**How do you recognize, that a gluon node is affected?** 
 
- The gluon node will not be publicly reachable on some of its addresses.

**What are the conditions, this bug is triggered:**
- Multiple router advertisement servers are in the mesh network.
- These servers advertise different *default routers* and different *prefixes*.
- The *default routers* are not able to route packets with source ips of other *prefixes*.

**Not affected:**
- Clients of the gluon node are not affected. The *radv-filterd* is filtering here.
- Forwarding is not affected. This is only about the packets, whose source is the gluon node itself. 
- IPv4.

**Possible workarounds:**
- Implement source based routing between the *default routers*:
    - Linux has supported this without policy-based routing for quite some time, but documentation on the Internet is very sparse. (It's almost not findable via google.)
    - Using the ip-tools: `ip route add ::/0 from PREFIX via OTHER_ROUTER dev IFACE`.
    - In systemd-networkd, the `Source=...` can be used inside the `[Route]` section.

**Gluon Version:**
- master

**Site Configuration:**
- n/a

**Custom patches:**
- n/a

lemoer:
Router advertisement servers advertise prefixes together with a default
router in layer 2 networks. There may be multiple router advertisement
servers in one layer 2 network. Each of those router advertisement
servers can advertise different prefixes and a different default router.

A client can accept multiple of the advertised prefixes at the same
time. If it does that, it installs an address in each of the advertised
prefixes. It also has to install default routes from the router
advertisements. Since default routers are sometimes not able to
forward ip packets from arbitrary source addresses, the client should
install routes to all default routers and use the correct one depending
on which of the installed addresses it is sending from.

This can be addressed using source based routing. The odhcpv6 script
already attempts to implement this. The setting is named
"sourcefilter". In this commit, we enable it.

	Note, that the following examples show output of ip-full.
	I am not sure, whether the "from" part is acutally shown by
	ip-tiny.

	Before this commit:

	root@platzhalter-525400123457:/# ip -6 route show default
	default via fe80::8ae6:40ff:feba:a099 dev br-client proto static metric 512 pref medium

	After this commit:

	root@platzhalter-525400123457:/# ip -6 route show default
	default from 2001:678:978:199::/64 via fe80::8ae6:40ff:feba:1099 dev br-client proto static metric 512 pref medium
	default from 2a02:790:ff:799::/64 via fe80::8ae6:40ff:feba:1099 dev br-client proto static metric 512 pref medium
	default from 2a02:790:ff:899::/64 via fe80::8ae6:40ff:feba:1099 dev br-client proto static metric 512 pref medium
	default from 2a02:790:ff:999::/64 via fe80::8ae6:40ff:feba:1099 dev br-client proto static metric 512 pref medium
	default from 2a02:790:ff:1099::/64 via fe80::8ae6:40ff:feba:1099 dev br-client proto static metric 512 pref medium
	default from fdca:ffee:8:99::/64 via fe80::8ae6:40ff:feba:1099 dev br-client proto static metric 512 pref medium

Fixes #2180
Tested-By: Leonardo Mörlein <git@irrelefant.net>

### DIFF
--- a/package/gluon-mesh-batman-adv/luasrc/lib/gluon/upgrade/320-gluon-mesh-batman-adv-client-bridge
+++ b/package/gluon-mesh-batman-adv/luasrc/lib/gluon/upgrade/320-gluon-mesh-batman-adv-client-bridge
@@ -14,7 +14,7 @@ uci:section('network', 'interface', 'client', {
 	proto = 'dhcpv6',
 	reqprefix = 'no',
 	peerdns = not site.dns.servers(),
-	sourcefilter = false,
+	sourcefilter = true,
 	keep_ra_dnslifetime = true,
 	robustness = 9,
 	query_interval = 2000,


lemoer:
Actually I just noticed, that the destination ip of all routes is the same `fe80::8ae6:40ff:feba:1099`. This means, that this does not fix anything. Everything is still sent to only one supernode.

lemoer:
Unfortunately the mechanism is completely broken in OpenWrt. Or is not meant to fulfill what we want. You can interpret this as you like.

The odhcp6c passes two variables to the dhcpv6.scropt:
- `$ADDRESSES` contains the information about the announced prefixes from **all** received RA packets.
- `$RA_ROUTES` contains the announced routes from **all** received RA packets.

However, there is no association which routes and prefixes came from the same RA server. The code calls `proto_add_ipv6_route` for each route with every prefix (for loop in for loop). This means in our case, that the route is overwritten again and again. In the end, the last entry of `$ADDRESSES` wins and all routes have the same `from`.

NeoRaider:
Source Routing without a generic default route breaks Source Address Selection; the node will need some way to decide which address to use for outgoing packets.

We don't have a good concept how to deal with multiple prefixes in the same network, especially in the presence of roaming clients etc. For now the rule for Gluon is "all IPv6 routers must announce the same set of prefixes", and any attempt to change this rule will require some serious design work.

lemoer:
Thanks for the explanations. I wasn't aware of that "rule".

From the explanations of @awlx  and @NeoRaider in #2180, it sounds like the "rule" could be changed to:
- "all IPv6 routers must ~~announce~~ be able to route the same set of prefixes".

Or am I missing a point?

NeoRaider:
> From the explanations of @awlx and @NeoRaider in #2180, it sounds like the "rule" could be changed to:
> 
>     * "all IPv6 routers must ~announce~ be able to route the same set of prefixes".

Yes, this is correct and more precise than my previous wording.

lemoer:
I'll close this, as the suggested changes do not make sense. Discussion can continue in #2180.

PR #2185, opened 2021-03-08
gluon-neighbour-info: end before timeout

### LINKED ISSUE #2184: gluon-neighbour-info returns late
### Bug report

**What is the problem?**
The tool `gluon-neighbour-info` does return late (until timeout), given the right parameters.

I think it's always been this way, and has just not bothered anyone yet.
It does behave similiary on all my devices spread across several gluon versions. 

`gluon-neighbour-info -d ::1 -p 1001 -r nodeinfo` should return as fast as the explicit `gluon-neighbour-info -c 1 -d ::1 -p 1001 -r nodeinfo`.

e.g.
`gluon-neighbour-info -d ::1 -p 1001 -r nodeinfo` is expected to return one value, as there is no `-l` parmeter given.
It does print the expected value rather fast, but does not return until the default timeout of `-t`=`3` passed.

Wouldn't it be sensible to let -c default to `1` in case `-l` is not given?

Or are there scenarios, where two structs are returned at once?
I expected `-r neighbours`, but even that does just return one dictionary.

**Gluon Version:**

e.g. v2020.2-187-g57b40f5d
and v2020.1.3-3-g83b67555
but very likely any.

**Site Configuration:**

https://github.com/freifunkh/site/commit/7b398092ab2e504fca9ba6f62cb0063e2545af03
and
https://github.com/freifunkh/site/commit/7b398092ab2e504fca9ba6f62cb0063e2545af03
though it's likely not related to a specific site.

**Custom patches:**

I have not patched anything yet, but testing for the existence of `-l` and letting `-c` default appropriately may be a good start.

AiyionPrime:
This is has been compiled and tested.

Results of `time ./gluon-neighbour-info -s neighbour -d ff02::2:1001 -p 1001 -r nodeinfo`
before and after:

`real	0m3,013s`
and
`real	0m0,168s`

> End the process after one result in case  is not given.
> Reduces singleshot execution time from timeout seconds to around 150ms.
> 
> resolves #2184

### DIFF
--- a/package/gluon-neighbour-info/src/gluon-neighbour-info.c
+++ b/package/gluon-neighbour-info/src/gluon-neighbour-info.c
@@ -44,7 +44,8 @@ void usage() {
 	puts("  -t <sec>         timeout in seconds (default: 3)");
 	puts("  -s <event>       output as server-sent events of type <event>");
 	puts("                   or without type if <event> is the empty string");
-	puts("  -c <count>       only wait for at most <count> replies");
+	puts("  -c <count>       only wait for at most <count> replies (default: 1");
+	puts("                   if -l is not given for unicast destination addresses)");
 	puts("  -l               after timeout (or <count> replies if -c is given),");
 	puts("                   send another request and loop forever");
 	puts("  -h               this help\n");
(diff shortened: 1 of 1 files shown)


NeoRaider:
This breaks the use in `package/gluon-status-page/files/lib/gluon/status-page/www/cgi-bin/dyn/neighbours-nodeinfo`.

PR #2190, opened 2021-04-05
[v2020.2.x] Backport module updates since 02/2021

mweinelt:
Backports module updates since the beginning of february.


rotanid:
build tested with ar71xx-generic - run-tested on TP-Link TL-WR1043N/ND v2

PR #2191, opened 2021-04-05
Relax CI Runner version pin

mweinelt:
There should be no reason why our CI cannot run on the latest Ubuntu LTS.


SvenRoederer:
he only Thing I realized: on the Ubuntu-20.04 runners the OS takes up some gigabytes more, which made the Freifunk-Berlin firmware running out of disc-space.

mweinelt:
Interesting, forunately no such issue over here.

SvenRoederer:
Was only happening for ath79-generic, where we build images for all boards to include into Imagebuilder.

PR #2196, opened 2021-04-07
statuspage: fix tq width

### LINKED ISSUE #2193: status-page TQ cut off
### Bug report

**What is the problem?**

On gluon master, as well as recent gluon releases the neighbors transmission quality is displayed on the right side of the frame.
When the value reaches `100 %`, the text gets cropped to `100...`.

The problem was introduced sometime after gluon v2020.1.3 (792a93b4dd8841daf6e8ff57be055b12384b2ce2).

The problem occurs on several devices with latest firefox or chromium.

The problem is target independent.

The cropping appears to be unintended and does not improve readability.
In case this is a problem only for some clients, I can provide screenshots.

**Gluon Version:**

Looking at the commit history of the file commit 016b323a1710cb205ce4862e9f4a8f9cb4db6320 would be my first guess to have introduced this behavior,
containing layout changes of @nalxnet, which would be in v2020.2.2 (90d0e33c619cef9e0af928ef4d6477f6c1bdc0de)

**Site Configuration:**
The problem appears to be site independent, as I found it in testing-releases in ffh 
http://[2001:678:978:116:a2f3:c1ff:fe5e:c8e6]/cgi-bin/status
as well as in ffrn.
http://[2a01:4f8:171:fcff:1ad6:c7ff:fe90:dec5]/cgi-bin/status

AiyionPrime:
...overflow for 'DejaVu Sans'.

This resolves #2193.

This needs testing - I don't usually do web development.


AiyionPrime:
Compiled a new version containing the patch - works fine for me.
Waiting for @bschelms Feedback on it.

AiyionPrime:
@bschelm just verified, it works on his pc as well.
I think this PR is fine and ready to merge.

PR #2197, opened 2021-04-07
 contrib/Dockerfile: use apt-get instead of apt

txt-file:
* apt is meant for user interactive usage. apt does not guarantee a stable CLI.
* set DEBIAN_FRONTEND=noninteractive to tell apt-get that no user interaction is wanted

### DIFF
--- a/contrib/Dockerfile
+++ b/contrib/Dockerfile
@@ -1,6 +1,7 @@
 FROM debian:buster-slim
 
-RUN apt update && apt install -y --no-install-recommends \
+ARG DEBIAN_FRONTEND=noninteractive
+RUN apt-get update && apt-get install -y --no-install-recommends \
     ca-certificates \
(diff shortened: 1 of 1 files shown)


contrib/Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM debian:buster-slim
 
-RUN apt update && apt install -y --no-install-recommends \
+ARG DEBIAN_FRONTEND=noninteractive
```
mweinelt, line 3:
```suggestion
ENV DEBIAN_FRONTEND=noninteractive
```

txt-file: According to https://docs.docker.com/engine/reference/builder/#env ENV is persisted in the resulting image. ARG is not persisted in the final image. That's why I have chosen ARG.

PR #2199, opened 2021-04-12
ath79: add NBG6616 and NBG6716 in gluon/next

belzebub40k:
The two devices were part of the ar71xx target in previous releases and are not yet included in ath79. I did test builds and flashed both devices without problems. But for the NBG6716 only the 5.4 kernel worked as the build process was complaining about a to big kernel when TESTING_KERNEL was set to true and the 5.10 kernel was used.

`WARNING: Image file /gluon/gluon/openwrt/build_dir/target-mips_24kc_musl/linux-ath79_nand/zyxel_nbg6716-kernel.bin is too big:  > 4194304
`


rotanid:
thanks, but typically we don't take pull requests against the next branch, its only a playground for the maintainers.
regarding the kernel version, please note that the next openwrt release uses kernel 5.4, not 5.10, therefore 5.10 won't be relevant anytime soon for Gluon.

belzebub40k:
Okay, I already thought so. The problem with the two devices is that they seem to be broken in recent Gluon releases based on OpenWrt 19.07 as the kernel is to big to boot. With OpenWrt 21.02 on the horizon which already fixes those issue I thought it wouldn't be worth to backport the fixes from ath79 to the already dropped ar71xx target and instead get them added to the next branch.

rotanid:
> The problem with the two devices is that they seem to be broken in recent Gluon releases based on OpenWrt 19.07 as the kernel is to big to boot. 

this hasn't been reported, otherwise we could have put that information in the latest release notes or even disable these or mark them as broken...

blocktrron:
I'm going to pick the first (It can always fall off the tree - but like in the other PRs you know the gist.

I don't pick the one who requires disabling of the testing kernel. This is a dead end and the WDR4900 also went to the bin because of that.

blocktrron:
Picked the non-NAND one. Please note, adding a device to next will not cause the device to eventually land in a regular Gluon branch / release.

PR #2200, opened 2021-04-12
gluon-status-page: add role to status-page

AiyionPrime:
Implements the next open crumb of #1415.

### DIFF
--- a/package/gluon-status-page/i18n/de.po
+++ b/package/gluon-status-page/i18n/de.po
@@ -97,6 +97,9 @@ msgstr "RAM"
 msgid "Received"
 msgstr "Empfangen"
 
+msgid "Role"
+msgstr "Rolle"
+
 msgid "Site"
 msgstr "Site"
 
--- a/package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
+++ b/package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
@@ -126,6 +126,9 @@
 							<%- end %>
 						</dd>
 					<%- end %>
+					<% if nodeinfo.system.role then -%>
+						<dt><%:Role%></dt><dd><%| nodeinfo.system.role %></dd>
+					<%- end %>
 					<% if nodeinfo.software.autoupdater then -%>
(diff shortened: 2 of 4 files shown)


package/gluon-status-page/i18n/de.po
```diff
@@ -97,6 +97,9 @@ msgstr "RAM"
 msgid "Received"
 msgstr "Empfangen"
 
+msgid "Received"
```
lemoer, line 100:
```suggestion
msgid "Received"
```

This should be 'Role'.

AiyionPrime: indeed thanks.

mweinelt (approved):
LGTM.

lemoer:
LGTM.

lemoer (approved):

lemoer:
Let's merge this?

NeoRaider:
I think this should resolve the human-readable role name via i18n instead of using the internal ID.

lemoer:
For reference: Neoraider is talking about this: https://github.com/freifunk-gluon/gluon/blame/a2be901803469d378d6469551a65c20f744aefff/docs/releases/v2015.1.rst#L189-L200

AiyionPrime:
Thanks @lemoer, though the referenced code is outdated;
the translations are now in `gluon-web-node-role:role:<rolename>`

AiyionPrime:
> I think this should resolve the human-readable role name via i18n instead of using the internal ID.

@NeoRaider The last commit adds your requested translation, though I do not know, whether there's a better syntax inside of views.

NeoRaider:
> > I think this should resolve the human-readable role name via i18n instead of using the internal ID.
> 
> @NeoRaider The last commit adds your requested translation, though I do not know, whether there's a better syntax inside of views.

I think it's fine, the special i18n view syntax isn't usable here because the site is a foreign i18n package.

AiyionPrime:
Is there anything left to do on my end, then?

mweinelt (approved):

PR #2203, opened 2021-04-14
ath79: add TP-Link EAP225 Outdoor v1

herbetom:
I'm absolutly fine with this only beeing for future reference.

- [ ] must be flashable from vendor firmware
  - [ ] webinterface
  - [ ] tftp
  - [ ] other: 
    - ssh into target device with recent (>= v1.6.0) firmware
    - run `cliclientd stopcs` on target device
    - upload factory image via web interface.
- wired network
- wifi (if applicable)
- led mapping
  - radio leds (The device has no such LEDs)
    - [ ] should show activity
  - switchport leds (The device has no such LEDs)
- outdoor devices only

I did not test the gluon factory image, but I did test the then current openwrt factory snaphot, which worked flawlessly.


herbetom:
Thanks, i didn't expect anything else :+1:

Issue #131, opened 2014-07-09
configmode: prevent loss of state when entering advanced settings


lost-tty:
When navigating from configmode to expertmode any state will be lost. This is especially annoying when geo-coordinates have been entered.

As we are planning to add modularity to configmode anyway we could add a special expertmode module to the configmode page that can be expanded by clicking on some button.

Example:

```
   Configmode
     Hostname
     ...

-> Advanced Configuration

[Save]
```

after clicking on `Advanced Configuration` the section would be expanded like this:

```
   Configmode
     Hostname
     ...

v  Advanced Configuration
     SSH Key
     WAN Port Mode
     ...

[Save]
```

rubo77:
I can imagine two solutions:
Either all settings are moved into one html page and the switch between expert and not is done by js or css.

Or you modify the expert switch-link into a save-and-switch to expert

tcatm:
What about using a luci compound widget?

kgbvax:
Just got this as feedback from a communal IT dept. ;-)

kpanic23:
Workaround:
Right-click on "Expert Mode" and click "Open in new Tab". Then, after saving the changes in Expert Mode, you can just close the tab and hit the "Save and Restart" button on the already filled-out Basic Config form.

rubo77:
This problem is not really very important, cause you have a workaround, but couldn't we just simply set the "advanced settings" link with `target="_blank"` until this is solved with a better solution?

Or maybe change the target of the wizzard link `onchange` of any  of the input fields with javascript.

Adorfer:
I presume this behavior of losing the setting when clicking on the "advanced" tab is just meant to scholar users for being new to gluon or not being familiar how to operate a web browser decently.

rubo77:
We could simply add a Java script check on click, that creates a warning

'you have unsaved changes, do you want to leave this page? You can use right click in a new tab'

rotanid:
this would be annoying for everyone who knows the consequences...

rubo77:
This would not Amoy anyone if the message is only shown if you changed something already.

Surely the message should not show if you just open 192.168.1.1 and directly click on the advanced settings.

belzebub40k:
What about adding a "save" button next to "save & reboot"?

rotanid:
well, let's add the "patch welcome" label.
see https://github.com/freifunk-gluon/gluon/labels/patch%20welcome for a list of such issues :)

[...]

PR #2205, opened 2021-04-22
gluon-{config,web}-*: show "save & apply" button during runtime

lemoer:
If the user has left the setup mode and starts a uhttpd instance to
serve the config mode during normal runtime of the gluon node, this
change is helpful. Instead of the "Save" or "Save & reboot" button,
now a "Save & apply" button is presented. If this button is pressed,
"gluon-reload" including a "gluon-reconfigure" is invoked internally
causing the node to regenerate and reload its configuration
immediately.

Currently such an uhttpd instance is not running by default, but a
user can start such an instance manually and access it via ssh port
forwarding.

---
Before:
Screenshot from 2021-04-23 00-52-35

After:
Screenshot from 2021-04-23 00-53-46

---

Tested:
- Changing hostname works.
- Changing domain works.
- Changing password works.
- Disabling/enabling vpn works.
- Enabling/disabling mesh on wan works.

### DIFF
--- a/package/gluon-config-mode-core/i18n/de.po
+++ b/package/gluon-config-mode-core/i18n/de.po
@@ -13,6 +13,9 @@ msgstr ""
 msgid "Save & restart"
 msgstr "Speichern & Neustarten"
 
+msgid "Save & apply"
+msgstr "Speichern & Anwenden"
+
 msgid "Welcome!"
 msgstr "Willkommen!"
 
--- a/package/gluon-config-mode-core/i18n/fr.po
+++ b/package/gluon-config-mode-core/i18n/fr.po
@@ -13,6 +13,9 @@ msgstr ""
 msgid "Save & restart"
 msgstr "Enregistrer & Redémarer"
 
+msgid "Save & apply"
+msgstr "Enregistrer & Appliquer"
+
 msgid "Welcome!"
 msgstr "Bienvenue!"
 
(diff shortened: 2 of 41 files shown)


NeoRaider:
Hmm. I understand that we already have some experimental support for reconfiguration without reboot for the hoodselector, but I'm generally not a fan of this. Also, my stance on running the config mode web UI outside of config mode is "please don't ever do this".

lemoer:
Can you please elaborate the reasons for these gut feelings? Do you think this will eventually complicate the code of the config mofe too much?

Background: My goal is to work towards integrating a "controller solution", were you can log into a webgui (with ssl of course), see all your devices and reconfigure them as you like. Currently I implement this, by tunneling the config mode webgui via ssh. To provide reasonable user experience in the "controller", this is done without entering the config mode first. Nobody needs a controller, where nodes have to be moved to config mode manually first.

If you think, the controller solution goes in the wrong direction or have better ideas, please let me know soon. The code for an initial version of the "gluon-controller" is almost finished.

AiyionPrime:
This might be a resolution for #131.

> What about adding a "save" button next to "save & reboot"?

_Originally posted by @belzebub40k in https://github.com/freifunk-gluon/gluon/issues/131#issuecomment-411040722_

blocktrron:
Closed as #2171 will not happen.

Issue #2212, opened 2021-05-05
let "make all" call "make config" and "make update" automatically


SvenRoederer:
Is there a specific reason why the make target "all" only depends on target "config"? Why not having "config" depending on "update"?
This way a simple "make all" call will be all that needs to be done for a correctly setup build.

SvenRoederer:
For the Freifunk-Berlin firmware I can checkout, do my "site config", call make and take a coffee or so.

blocktrron:
AFAIR the reason was that invoking `update` requires internet access while this is not necessarily the case when having cloned all modules and cached all downloads.

SvenRoederer:
ok, seems a valid argument.

NeoRaider:
`make update` doesn't require internet access when all referenced modules are up-to-date, but there lots of situations where an implicit `make update` is not desirable:
- `make update` takes some time, and it needs to be run only once and not for each target
- `make update` will revert modules to their original commits after manual checkouts, which is undesirable during development

SvenRoederer:
But is this really working? I assume in most cases when the modules files changes the upstream repo has some changes to it's package version. This in turn will again trigger downloads at OpenWrt buildtime. So it might end up in a 50-50 chance that you don't need internet to build.

PR #2215, opened 2021-05-06
docs: update development basics regarding 'make update'

SvenRoederer:
based on #2212  I added some sentences why ```make update``` has to be called explicitly and not by an dependency of ```make all```.

### DIFF
--- a/docs/dev/basics.rst
+++ b/docs/dev/basics.rst
@@ -52,6 +52,13 @@ and you can try rebasing it onto the new `base` branch yourself and after that c
 Always call `make update-patches` after making changes to a module repository as `make update` will overwrite your
 commits, making `git reflog` the only way to recover them!
 
+Then all is set start the build by calling (probably add -jn):
+
+::
+
+	make all
+
+
 Development Guidelines
 ----------------------
 Lua should be used instead of sh whenever sensible. The following criteria
--- a/docs/dev/basics.rst
+++ b/docs/dev/basics.rst
@@ -37,6 +37,10 @@ rerun
 `make update` also applies the patches that can be found in the directories found in
 `patches`; the resulting branch will be called `patched`, while the commit specified in `modules`
 can be referred to by the branch `base`.
+To separately call `make update` was choosen, to let you decide when to update and reset the external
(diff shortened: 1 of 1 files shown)


docs/dev/basics.rst
```diff
@@ -37,6 +37,9 @@ rerun
 `make update` also applies the patches that can be found in the directories found in
 `patches`; the resulting branch will be called `patched`, while the commit specified in `modules`
 can be referred to by the branch `base`.
+To separately call `make update` was choosen, to let you decide when to update and reset the external
+repos. This will prevent potential loss of commits during local development amd allow building Gluon
+without the need of having an internet connection.
```
mweinelt, line 42: Not really, given that `make download` is a thing.

SvenRoederer: I assume you are referencing to OpenWrts make download. But that's out of Gluons Makefile ...

Anyway I updated the sentence, as I found a typo.

docs/dev/basics.rst
```diff
@@ -37,6 +37,10 @@ rerun
 `make update` also applies the patches that can be found in the directories found in
 `patches`; the resulting branch will be called `patched`, while the commit specified in `modules`
 can be referred to by the branch `base`.
+To separately call `make update` was choosen, to let you decide when to update and reset the external
```
lemoer, line 40:
```suggestion
The separately call `make update` was choosen to let you decide when to update and reset the external
```

docs/dev/basics.rst
```diff
@@ -37,6 +37,10 @@ rerun
...
+repos. This will prevent potential loss of commits during local development and allow building Gluon
```
lemoer, line 41:
```suggestion
repos. This avoids potential loss of commits during local development and allows building Gluon
```

docs/dev/basics.rst
mweinelt: This sounds like you misread the sentence?

NeoRaider:
The explanation is not very accurate, and I don't think "working with repositories" is the right section to add this information. I've opened #2356 instead.

Issue #644, opened 2016-02-07
respondd: transmit additional node data


Adorfer:
i'd like to have transmitted via the respondd (gluon collector) more node static data, in order to monitor remotely the setup of more complex mesh networks without the need to ssh to nodes to check for data.
- radioX.wifichannel selected.
- radioX.HT-mode operating
- radioX.txpower running
- radioX.DFSevent(format)?
- radioX.ClientAP disabled
- radioX.wifimesh disabled 
- current systemtime in node (perhaps timestamp the datagram)

perhaps 
- meshvpn type (fastd, wireguard, l2tptunneldigger...) 
- meshvpn status /disabled/disconnected/connected
- meshvpn simple-tc: limit_enabled, limit_ingress, limit_egress
- MoW enabled
- MoL enabled

nice to have
- either name of IFs per bridge/vswitch, 
  in order to detect setup-foo like (MoL&(MeshVPn|Client)) akticvated on same phy)

ultimativly even, if info available from switch
- ethernetports up

plumpudding:
we have meshvpn already
MoW and MoL can probably be extracted from the current data too (only if one of the two is active, not which)

NeoRaider:
I think you mean respondd, not collectd, right? We don't support collectd.

NeoRaider:
We already provide a list of mesh interfaces with type (wireless/tunnel/other, where other is usually wired).

I guess we could add:
* Channel of each wireless mesh interface
* List of wireless client interfaces (with channel)

I think we shouldn't add anything hardware-specific, like refering to specific radio devices or specific ethernet ports, or differentiating between LAN and WAN ports.

plumpudding:
> I think you mean respondd, not collectd, right? We don't support collectd.

He wants to collect this data with gluon-collector from ffdo/node-informant
So it has to be provided by respondd, yes.

Adorfer:
@NeoRaider 
> I think you mean respondd, not collectd, right? We don't support collectd.
Correct, my fault. I took the "gluon collector" for collectd... but i off course it's (c)-respondd

tcatm:
> fastd is up, but not batman originators on the other side

The neighbour information provides just that. Please keep in mind that it is possible to have batman originators on an interface that you aren't made visible by the kernel module (due to lower TQ mostly).

viisauksena:
you maybe should have a look at https://github.com/viisauksena/gluon-airtime 
[...]

PR #2214, opened 2021-05-06
gluon-status-page: add mesh protocol to status-page

AiyionPrime:
This is a draft to show the deployed mesh protocol as well as its version on the routers statuspage.

Its addresses another open point on the agenda of #1415.

An example router can be found at http://lilienstra-e.n.ffh.zone/, or in case our dns misbehaves http://[2001:678:978:114:b64b:d6ff:fe20:2778]/cgi-bin/status .

PR #2216, opened 2021-05-07
gluon-mesh-vpn-core: respondd bandwidth_limit

AiyionPrime:
Implements a respondd provider for bandwidth limits using simple-tc.
Due to their static nature, I added the limits to nodeinfo.
This is what the network tree looks like after merging this:

**this was edited**
```
"network": {
  "addresses": [
    "2a02:790:ff:714:b64b:d6ff:fe22:4ebc",
    "2a02:790:ff:914:b64b:d6ff:fe22:4ebc",
    "2a02:790:ff:1014:b64b:d6ff:fe22:4ebc",
    "2001:678:978:114:b64b:d6ff:fe22:4ebc",
    "2a02:790:ff:514:b64b:d6ff:fe22:4ebc",
    "fdca:ffee:8:14:b64b:d6ff:fe22:4ebc",
    "fe80::b64b:d6ff:fe22:4ebc"
  ],
  "mesh": {
    "bat0": {
      "interfaces": {
        "wireless": [
          "b6:4b:d6:22:4e:c0",
          "9a:62:78:2a:dd:59"
        ],
        "tunnel": [
          "9a:62:78:2a:dd:5f"
        ],
        "other": [
          "9a:62:78:2a:dd:5b"
        ]
      }
    }
  },
  "mesh_vpn": {
    "bandwith_limit": {
      "enabled": true,
      "ingress": 5000,
      "egress": 800
    }
  },
  "mac": "b4:4b:d6:22:4e:bc"
},
```

Resolves a part of  #644 .

### DIFF
--- a/package/gluon-mesh-vpn-core/src/respondd.c
+++ b/package/gluon-mesh-vpn-core/src/respondd.c
@@ -23,12 +23,62 @@
 
 #include <respondd.h>
 
+#include <stdbool.h>
 #include <stdio.h>
 #include <string.h>
+#include <unistd.h>
 
 #include <json-c/json.h>
 #include <libgluonutil.h>
+#include <uci.h>
 
+
+static struct json_object * get_bandwidth_limit(void) {
+	bool enabled = false;
+	int egress = -1;
+	int ingress = -1;
+
+	struct json_object *ret = json_object_new_object();
+
+	struct uci_context *ctx = uci_alloc_context();
+	if (!ctx)
+		goto disabled;
+	ctx->flags &= ~UCI_FLAG_STRICT;
+
+	struct uci_package *p;
+	if (uci_load(ctx, "gluon", &p))
+		goto disabled;
+
+	struct uci_section *s = uci_lookup_section(ctx, p, "mesh_vpn");
+	if (!s)
+		goto disabled;
+
+	const char *enabled_str = uci_lookup_option_string(ctx, s, "limit_enabled");
+	if (enabled_str && strcmp(enabled_str, "1"))
+		goto disabled;
+
+	enabled = true;
+
+	const char *egress_str = uci_lookup_option_string(ctx, s, "limit_egress");
+	if (strcmp(egress_str, "-"))
+		egress = atoi(egress_str);
+
+	const char *ingress_str = uci_lookup_option_string(ctx, s, "limit_ingress");
+	if (strcmp(ingress_str, "-"))
+		ingress = atoi(ingress_str);
+
+	if (egress >= 0)
+		json_object_object_add(ret, "egress", json_object_new_int(egress));
+	if (ingress >= 0)
+		json_object_object_add(ret, "ingress", json_object_new_int(ingress));
+
+disabled:
+	if (ctx)
+		uci_free_context(ctx);
+
+	json_object_object_add(ret, "enabled", json_object_new_boolean(enabled));
+	return ret;
+}
 
 char * read_stdout(const char *command) {
 	FILE *f = popen(command, "r");
@@ -89,6 +139,7 @@ static struct json_object * respondd_provider_nodeinfo(void) {
 	struct json_object *network = json_object_new_object();
 	struct json_object *mesh_vpn = json_object_new_object();
 
+	json_object_object_add(mesh_vpn, "bandwidth_limit", get_bandwidth_limit());
 	json_object_object_add(mesh_vpn, "provider", get_active_vpn_provider());
 	json_object_object_add(mesh_vpn, "enabled", get_mesh_vpn_enabled());
 	json_object_object_add(network, "mesh_vpn", mesh_vpn);
--- a/package/gluon-mesh-vpn-core/src/Makefile
+++ b/package/gluon-mesh-vpn-core/src/Makefile
@@ -3,4 +3,4 @@ all: respondd.so
 CFLAGS += -Wall -Werror-implicit-function-declaration
 
 respondd.so: respondd.c
-	$(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) -shared -fPIC -D_GNU_SOURCE -o $@ $^ $(LDLIBS) -lgluonutil
+	$(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) -shared -fPIC -D_GNU_SOURCE -o $@ $^ $(LDLIBS) -lgluonutil -luci


package/gluon-mesh-vpn-core/src/respondd.c
```diff
@@ -0,0 +1,94 @@
...
+}
+
+
+static struct json_object * respondd_provider_nodeinfo(void) {
+	struct json_object *ret = json_object_new_object();
+	struct json_object *network = json_object_new_object();
+	struct json_object *mesh_vpn = json_object_new_object();
+
+	json_object_object_add(mesh_vpn, "simple_tc", get_simple_tc());
+	json_object_object_add(network, "mesh_vpn", mesh_vpn);
```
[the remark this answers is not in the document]
AiyionPrime: `mesh_vpn`was intended to match `statistics.mesh_vpn` not the interface `mesh-vpn`, that (as unrelated sidenote) might be renamed to `vpn` in #2168.

I think it's unlikely we are going to use simple-tc on other interfaces than the uplink, don't you think?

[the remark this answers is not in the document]
AiyionPrime: > Or maybe call it "simple-tc"?

I think using dashes instead of underscores defies the usage of dot-syntax in lua like here with `batman-adv`

> `<% if nodeinfo.software['batman-adv'] then -%>`

_Originally posted by @AiyionPrime in [#2214/.../status-page.html#L144](https://github.com/freifunk-gluon/gluon/pull/2214/files#diff-32907ffa1a5b6379d134e03669b9df3de2aab7d16fd543b22147208388c9a67eR144)

which is why I wouldn't change it lightheartedly.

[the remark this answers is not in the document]
CodeFetch: I agree it makes sense if it matches. There was a reason to not rename mesh-vpn to vpn, but I don't remember why...

lemoer (changes requested):
Generally this looks good. Two (or more) remarks:
1. Please use uci values from `gluon.mesh_vpn.limit_*` instead of `simple-tc.*` to make things protocol-agnostic. (See [here](https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-mesh-vpn-core/luasrc/lib/gluon/mesh-vpn/update-config#L13-L18))
2. Rebasing on #2102 is not necessary.
3. I would suggest to use `mesh_vpn.generic.*` as json path for the respondd responses.

AiyionPrime:
after talking to @lemoer we'd be going for
```
"mesh_vpn": {
    "bandwidth_limit": {
      "enabled": true,
      "ingress": 5000,
      "gress": 800
    }
  },
```

AiyionPrime:
rebased on master

AiyionPrime:
> Generally this looks good. Two (or more) remarks:

done

lemoer:
LGTM

lemoer (approved):

AiyionPrime:
simple-tc supports infinity as bandwidth limit.

There are three options.

1. Gluon does not support this officially,
2. Gluon does support this and we adapt simple-tcs syntax: a dash, which would change the bandwidth limit field from int to string
3. Gluon does support this, but uses any negative number as infinity

This does not change any of the internal workings of gluon, it's just about what output respondd should produce.

CodeFetch:
> Not sure if it is, what you wanted; but communicating an unlimited bandwidth limit by the absence of the ingress or egress in question is an option, yes.

That's what I meant. I don't like the idea with the dash.

AiyionPrime:
Maybe I was not clear enough.
simple-tc **does** take a dash as mark for infinity. It does so even when the dash comes from the simple-tc entry in uci.
The question is, how is infinity marked in the **other** uci section `gluon.mesh_vpn.limit_*` that leo mentioned earlier?

CodeFetch:
@AiyionPrime I think the dash is just a really bad idea as I would expect tonumber("-") to be zero...
https://github.com/freifunk-gluon/gluon/blob/0ac3061225b255ecf81224e96ee3b80862f2d853/package/gluon-mesh-vpn-core/luasrc/lib/gluon/upgrade/500-mesh-vpn#L77

CodeFetch:
@AiyionPrime tonumber("-") is actually nil... Now I understand... I think "-1" doesn't work, too...
From what I see negative numbers are not allowed:
https://github.com/freifunk-gluon/packages/blob/8d53ff54e562ddb2ed8397781dd78edc76f6ff38/net/simple-tc/src/simple-tc.c#L266

Another thing that makes me curious is this line:
https://github.com/freifunk-gluon/packages/blob/8d53ff54e562ddb2ed8397781dd78edc76f6ff38/net/simple-tc/src/simple-tc.c#L269

The maximum throughput I've measured on a FritzBox 4040 with WireGuard was about 120 Mbit/s while I've expected about 200... Is it possible that's the reason?

My two cents: It needs a fix. Maybe just use "0" for no limit...

AiyionPrime:
> From what I see negative numbers are not allowed:
> https://github.com/freifunk-gluon/packages/blob/8d53ff54e562ddb2ed8397781dd78edc76f6ff38/net/simple-tc/src/simple-tc.c#L266

That's a good point.

> Another thing that makes me curious is this line:
> https://github.com/freifunk-gluon/packages/blob/8d53ff54e562ddb2ed8397781dd78edc76f6ff38/net/simple-tc/src/simple-tc.c#L269
> 
> The maximum throughput I've measured on a FritzBox 4040 with WireGuard was about 120 Mbit/s while I've expected about 200... Is it possible that's the reason?

The line you found appears to be a conversion from kilobit to byte. I do not think, that's related.
But you can test, if you just disable bandwidth limits on the device in question.
This has likely nothing todo with this PR though.

> My two cents: It needs a fix. Maybe just use "0" for no limit...

Yes maybe, or a dash, that are the two remaining options :D

CodeFetch:
@AiyionPrime Ah... Now I see. https://github.com/freifunk-gluon/packages/blob/48cddbbb55691825596f7f640848dd24d4d4edff/net/simple-tc/files/etc/hotplug.d/net/50-simple-tc#L20

Okay... If there is no entry, it will use "-" as a parameter for tc... So I'd stick with not setting the entry if there is no limit in UCI as this is what happens when you do tonumber("-"). Even if you use "-" it will become a nil which will be handled as a "-" again by the hotplug-script... Totally confusing...

AiyionPrime:
I implemented the idea of @CodeFetch, if a dash is read as e- or ingress, the limit is not published via respondd.

The code is compiled and runs on http://lilienstra-e.n.ffh.zone/.

This is the above node with egress set to `-` which produces a nodeinfo containing this excerpt:

```json
"mesh_vpn": {
  "bandwidth_limit": {
    "enabled": true,
    "ingress": 4000
  }
},
```

I think it's ready for review again.

CodeFetch (approved):

CodeFetch:
@AiyionPrime Hm... Why are the changes missing here you have made in #2217? Or better say why don't you move them from #2217 into this PR? I just thought there was a glitch again :grin:...

AiyionPrime:
Because I overlooked they were reported in the wrong PR. I'm on it, thanks.

AiyionPrime:
Thats it.

package/gluon-mesh-vpn-core/src/respondd.c
```diff
@@ -23,13 +23,64 @@
...
+		goto disabled;
+
+	const char *enabled_str = uci_lookup_option_string(ctx, s, "limit_enabled");
+	if (enabled_str && strcmp(enabled_str, "1"))
+		goto disabled;
+
+	enabled = true;
+
+	const char *egress_str = uci_lookup_option_string(ctx, s, "limit_egress");
+	if (strcmp(egress_str, "-") != 0)
```
blocktrron, line 63: `if (strcmp(egress_str, "-"))`

package/gluon-mesh-vpn-core/src/respondd.c
```diff
@@ -23,13 +23,64 @@
...
+		egress = atoi(egress_str);
+
+	const char *ingress_str = uci_lookup_option_string(ctx, s, "limit_ingress");
+	if (strcmp(ingress_str, "-") != 0)
```
blocktrron, line 67: `if (strcmp(ingress_str, "-"))`

package/gluon-mesh-vpn-core/src/respondd.c
```diff
@@ -23,13 +23,64 @@
...
+
+	if (egress >= 0)
+		json_object_object_add(ret, "egress", json_object_new_int(egress));
+	if (ingress >= 0)
+		json_object_object_add(ret, "ingress", json_object_new_int(ingress));
+
+disabled:
+	uci_free_context(ctx);
+
+disabled_nofree:
```
blocktrron, line 78: drop this label

package/gluon-mesh-vpn-core/src/respondd.c
```diff
@@ -23,13 +23,64 @@
...
+	if (strcmp(ingress_str, "-") != 0)
+		ingress = atoi(ingress_str);
+
+	if (egress >= 0)
+		json_object_object_add(ret, "egress", json_object_new_int(egress));
+	if (ingress >= 0)
+		json_object_object_add(ret, "ingress", json_object_new_int(ingress));
+
+disabled:
+	uci_free_context(ctx);
```
blocktrron, line 76:
```
if (ctx)
	uci_free_context(ctx);
```

package/gluon-mesh-vpn-core/src/respondd.c
```diff
@@ -23,13 +23,64 @@
...
+static struct json_object * get_bandwidth_limit(void) {
+	bool enabled = false;
+	int egress = -1;
+	int ingress = -1;
+
+	struct json_object *ret = json_object_new_object();
+
+	struct uci_context *ctx = uci_alloc_context();
+	if (!ctx)
+		goto disabled_nofree;
```
blocktrron, line 45: `goto disabled;`

blocktrron (commented):
Few minor remarks. Otherwise LGTM.

Please don't forget to squash your commits.

AiyionPrime:
compiles.

package/gluon-mesh-vpn-core/src/respondd.c
```diff
@@ -42,7 +42,7 @@ static struct json_object * get_bandwidth_limit(void) {
 
 	struct uci_context *ctx = uci_alloc_context();
 	if (!ctx)
-		goto disabled_nofree;
+		goto disabled;
```
CodeFetch, line 45: This was previously more efficient. Now you have two ifs...
@blocktrron Why should he have changed this? What's the problem with having two labels?
Do you mean it is more efficient, because of branch prediction?

blocktrron: In it's current state it doesn't really matter. However, in case this code is extended and more memory is allocated on the heap, freeing everything on completion is more readable instead of having multiple locations to jump to.

Performance-wise, these additional checks don't matter that much, given requests for responss from the running node are multiplexed anyhow.

CodeFetch:
@blocktrron @lemoer Can you merge this now?

AiyionPrime (commented):
Rebased on master for #2217, compiled and deployed on http://meshvpnabstracttest.n.ffh.zone/cgi-bin/status.

I think this is ready to merge.

blocktrron:
Let's go

PR #2217, opened 2021-05-07
gluon-status-page: add bandwidth limit

### LINKED ISSUE #1415: gluon-status-page: show additional info

AiyionPrime:
**blocked** by #2216

There will be a bandwidth limit on the statuspage, eventually.

And likewise this might finally resolve #1415 .

This router has the necessary commits applied:
http://meshvpnabstracttest.n.ffh.zone/cgi-bin/status or

img

I implemented a bit-formatting function that looks like this, whether this should go in statuspage or a util file is up to decide.

```lua
local function formatBits(v, factor)
    --to format a value given in kilobits use a factor of 1000
    factor = factor or 1
    units = {'k', 'M', 'G'}
    units[0] = ''

    bits = math.max(0, v*factor)
    pow = math.floor(math.log(bits)/math.log(1000))
    known_pow = math.min(pow, #units)

    bits = bits/1000^known_pow
    return string.format('%s%sbit', bits, units[known_pow])
end
```

This is what it works like:

```console
$ lua demo.lua 
(5000, 1000) -> 	5.0Mbit
(800, 1000) -> 	        800.0kbit
(133742, 1000) -> 	133.742Mbit
(13371337, 1000) -> 	13.371337Gbit
```

### DIFF
--- a/package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
+++ b/package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
@@ -64,6 +64,16 @@
 		return v and translate('enabled') or translate('disabled')
 	end
 
+	local function formatBits(bits)
+		local units = {[0]='', 'k', 'M', 'G'}
+
+		local pow = math.floor(math.log(math.max(math.abs(bits), 1)) / math.log(1000))
+		local known_pow = math.min(pow, #units)
+
+		local significand = bits/(1000^known_pow)
+		return string.format('%g %sbit', significand, units[known_pow])
+	end
+
 	local function statistics(key, format)
 		return string.format('<span data-statistics="%s" data-format="%s"></span>', pcdata(key), pcdata(format or 'id'))
 	end
@@ -122,6 +132,17 @@
 							<br /><%| nodeinfo.network.mesh_vpn.provider %>
 							<%- end %>
 						</dd>
+						<% if nodeinfo.network.mesh_vpn.bandwidth_limit.enabled then -%>
+						<dt><%:Bandwidth limit%></dt>
+						<dd>
+							<% if nodeinfo.network.mesh_vpn.bandwidth_limit.egress then -%>
+								▲ <%| formatBits(nodeinfo.network.mesh_vpn.bandwidth_limit.egress*1000) %>/s <%:upstream%><br />
+							<%- end %>
+							<% if nodeinfo.network.mesh_vpn.bandwidth_limit.ingress then -%>
+								▼ <%| formatBits(nodeinfo.network.mesh_vpn.bandwidth_limit.ingress*1000) %>/s <%:downstream%>
+							<%- end %>
+						</dd>
+						<%- end %>
 					<%- end %>
 					<dt><%:Site%></dt><dd><%| site.site_name() %></dd>
 					<% if nodeinfo.system.domain_code then -%>
--- a/package/gluon-status-page/i18n/de.po
+++ b/package/gluon-status-page/i18n/de.po
@@ -28,6 +28,9 @@ msgstr "1 Tag"
 msgid "Automatic updates"
 msgstr "Automatische Updates"
 
+msgid "Bandwidth limit"
+msgstr "Bandbreitenlimit"
+
 msgid "Channel"
 msgstr "Kanal"
 
@@ -42,6 +45,9 @@ msgstr "Entfernung"
 
 msgid "Domain"
 msgstr "Domäne"
+
+msgid "downstream"
+msgstr "Downstream"
 
 msgid "Error"
 msgstr "Fehler"
@@ -123,6 +129,9 @@ msgstr ""
 
 msgid "Transmitted"
 msgstr "Gesendet"
+
+msgid "upstream"
+msgstr "Upstream"
 
 msgid "Uptime"
 msgstr "Laufzeit"
--- a/package/gluon-status-page/i18n/fr.README
+++ b/package/gluon-status-page/i18n/fr.README
@@ -1,0 +1,10 @@
+Draft for french translation, taken from gluon-config-mode-mesh-vpn
+
+msgid ""
+msgstr ""
+
+msgid "downstream"
+msgstr "débit déscendant"
+
+msgid "upstream"
+msgstr "débit ascendant"
--- a/package/gluon-status-page/i18n/gluon-status-page.pot
+++ b/package/gluon-status-page/i18n/gluon-status-page.pot
@@ -19,6 +19,9 @@ msgstr ""
 msgid "Automatic updates"
 msgstr ""
 
+msgid "Bandwidth limit"
+msgstr ""
+
 msgid "Channel"
 msgstr ""
 
@@ -32,6 +35,9 @@ msgid "Distance"
 msgstr ""
 
 msgid "Domain"
+msgstr ""
+
+msgid "downstream"
 msgstr ""
 
 msgid "Error"
@@ -113,6 +119,9 @@ msgid "Traffic"
 msgstr ""
 
 msgid "Transmitted"
+msgstr ""
+
+msgid "upstream"
 msgstr ""
 
 msgid "Uptime"
--- a/package/gluon-status-page/i18n/ru.README
+++ b/package/gluon-status-page/i18n/ru.README
@@ -10,6 +10,7 @@ if we ever add Russion to gluon-web, the following strings can be reused:
 "Primary MAC": "Основной MAC",
 "IP Address": "IP Адрес",
 "Automatic updates": "Автоматические обновления",
+"Bandwidth limit": "Ограничение пропускной способности",
 "Overview": "Обзор",
 "used": "используется",
 "Uptime": "Время работы",


AiyionPrime:
Now with correct SI-unit shorthand `k` for `kilo`...

img

herbetom:
I think adding a space between the number and the unit would be better. Also i'm uncertain if "up" and "down" are the best words to describe those values. I suppose it's okay, but if there are other suggestions, I guess they should be considered.

AiyionPrime:
> I think adding a space between the number and the unit would be better.
Inded, if only to match the traffic section in the statistics column.

> Also i'm uncertain if "up" and "down" are the best words to describe those values. I suppose it's okay, but if there are other suggestions, I guess they should be considered.

Other names used in this project are `upstream & downstream` or more technical `egress & ingress`.
Currently not used is `up- and  download`, i think.

I just realized, whatever terms we use, they need translations.

Is using Unicode okay? Maybe we could arrows (up and down) to emphasize and support the (_too?_ technical term).

herbetom:
I thought about Unicode too, but I think that would probably be hard to decipher too.

In general, I probably favor the terms used in config mode.

mweinelt:
Unicode sounds worth a try.

edeso:
will this survive the value for unlimited '-' ? i tend to limit egress only. like

```
config interface 'mesh_vpn'
        option ifname 'mesh-vpn'
        option enabled '1'
        option limit_ingress '-'
        option limit_egress '10000'

**Comment by AiyionPrime** [comment]
limit egress and ingress are parsed to int anything that cannot be parsed as such becomes 0, more details in `man atoi`;
is using a dash something documented I overlooked, or something you do on your router?

**Comment by AiyionPrime** [comment]
No, you're right; at least for fastd a minus is at least supported by the underlying simple-tc.
https://github.com/freifunk-gluon/packages/blob/8d53ff54e562ddb2ed8397781dd78edc76f6ff38/net/simple-tc/src/simple-tc.c#L264

I am not sure, if tunneldigger does support the syntax; hence gluon.

Keep in mind though, this PR does not fiddle with the bandwidth limits, it only presents them.

Setting a dash would result in a `0` being presented as limit.
@edeso If that's not enough, I'd ask you to either find the documentation of this behaviour in gluon, or at least raise an issue to document it.

**Comment by edeso** [comment]
> No, you're right; at least for fastd a minus is at least supported by the underlying simple-tc.
> https://github.com/freifunk-gluon/packages/blob/8d53ff54e562ddb2ed8397781dd78edc76f6ff38/net/simple-tc/src/simple-tc.c#L264

yupp, it's badly documented, but there. for wireguard i actually need to simply do not set ingress.
 
> I am not sure, if tunneldigger does support the syntax; hence gluon.
> 
> Keep in mind though, this PR does not fiddle with the bandwidth limits, it only presents them.

sure. just wanted to make sure you are aware and it won't choke on the value
 
> Setting a dash would result in a `0` being presented as limit.
> @edeso If that's not enough, I'd ask you to either find the documentation of this behaviour in gluon, or at least raise an issue to document it.

sounds reasonable, albeit fixing up the underlying implementations to take (value<=0) as unlimited would enable the status page to represent it more clearly. but as you wrote, wrong pull request for that :)

**Comment by mweinelt** [comment]
> ingress and one of these?
> ⇩⤓▼arrow_down↓
> or even ingress ⇵ egress?
> 
> The above appear properly, this one is not part of my charset. ⭳

IMHO ↓ (U+2193) and ↑ (U+2191) sound like they're worth a try.

The case where the value is zero should be special cased, because it would lead to a misleading statement otherwise, though I'm not sure whether to hide the relevant direction in that case, or make the unlimited nature explicit.

**Comment by AiyionPrime** [comment]
I'd be fine with both; but if we go for the explicit way, I'd want it in the docs as well.

**Comment by AiyionPrime** [comment]
Is it a concern, that if a user set the bandwidth to zero the same statuspage output would be generated as if a user set the limit to `-`?

Is a user setting it to zero a concern at all?

**Comment by AiyionPrime** [comment]
> [...] though I'm not sure whether to hide the relevant direction in that case, or make the unlimited nature explicit.

I'm not sure either. Would it be bad, if the section was empty, when someone enabled the limit but set both directions to unlimited?

I think the more important question is, can we infer `-` as unlimited? Or is it just not properly defined yet, as maybe just simple-tc supports it? Setting a dash while using e.g. tunneldigger would do nothing, iirc; but would be stated as unlimited on the statuspage.

**Comment by AiyionPrime** [comment]
I'll do a try with vanishing entries and some arrows, then we can discuss this better I think.

**Comment by AiyionPrime** [comment]
Done.
The nodes egress is currently set to `-` and does therefore not show the open limit.

http://lilienstra-e.n.ffh.zone/cgi-bin/status

Any thoughts on the arrow?

**Comment by AiyionPrime** [comment]
Seeing the page, I think `↓ 4 Mbit/s ingress` would look better than the current `4 Mbit/s ↓ ingress`.

**Comment by mweinelt** [comment]
The arrow is much to subtle on a 4K display sadly. :thinking:

**Comment by AiyionPrime** [comment]
One of these then? ⇩▼

**Comment by belzebub40k** [comment]
> The arrow is much to subtle on a 4K display sadly. thinking

Probably using an other arrows helps. e.g. ▼▲

https://unicode-table.com/en/sets/arrow-symbols/#up-arrows
https://unicode-table.com/en/sets/arrow-symbols/#down-arrows

**Comment by AiyionPrime** [comment]
build is compiling, result in about 16 minutes.

**Comment by AiyionPrime** [comment]
At least on 2k this looks better.

http://lilienstra-e.n.ffh.zone/cgi-bin/status

File: package/gluon-mesh-vpn-core/src/respondd.c [hunk 1]
```diff
@@ -0,0 +1,98 @@
...
+	if (uci_load(ctx, "gluon", &p))
+		goto disabled;
+
+	struct uci_section *s = uci_lookup_section(ctx, p, "mesh_vpn");
+	if (!s)
+		goto disabled;
+
+	const char *enabled_str = uci_lookup_option_string(ctx, s, "limit_enabled");
+	if (!enabled_str || !strcmp(enabled_str, "1"))
+		enabled = true;
```
CodeFetch, line 57:
```
if (enabled_str || strcmp(enabled_str, "1"))
		goto disabled;

enabled = true;
```

package/gluon-mesh-vpn-core/src/respondd.c
```diff
@@ -0,0 +1,98 @@
...
+
+	const char *ingress_str = uci_lookup_option_string(ctx, s, "limit_ingress");
+	if (strcmp(ingress_str, "-") != 0)
+		ingress = atoi(ingress_str);
+
+disabled:
+	uci_free_context(ctx);
+
+disabled_nofree:
+	json_object_object_add(ret, "enabled", json_object_new_boolean(enabled));
```
CodeFetch, line 71:
```
disabled:
	uci_free_context(ctx);

disabled_nofree:
	json_object_object_add(ret, "enabled", json_object_new_boolean(enabled));
```
Move below...

package/gluon-mesh-vpn-core/src/respondd.c
```diff
@@ -0,0 +1,98 @@
...
+	if (enabled) {
```
CodeFetch, line 72: Remove if...

package/gluon-mesh-vpn-core/src/respondd.c
```diff
@@ -0,0 +1,98 @@
...
+		if (egress >= 0)
+			json_object_object_add(ret, "egress", json_object_new_int(egress));
+		if (ingress >= 0)
+			json_object_object_add(ret, "ingress", json_object_new_int(ingress));
+	}
```
CodeFetch, line 77: Add here:
```
disabled:
	uci_free_context(ctx);

disabled_nofree:
	json_object_object_add(ret, "enabled", json_object_new_boolean(enabled));
```

CodeFetch (changes requested):
Code can be simplified.

package/gluon-mesh-vpn-core/src/respondd.c
AiyionPrime: you did mean `&&` right?

package/gluon-mesh-vpn-core/src/respondd.c
AiyionPrime: Thought about that too, but decided against it in favor of the improved readability. Whatever...

AiyionPrime:
@CodeFetch your remarks have been taken care of. Anything else?

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
```diff
@@ -116,6 +130,17 @@ <h2><%:Overview%></h2>
 					<dt><%:Firmware%></dt><dd><%| nodeinfo.software.firmware.release %></dd>
 					<% if nodeinfo.software.fastd then -%>
 						<dt><%:Mesh VPN%></dt><dd><%| enabled(nodeinfo.software.fastd.enabled) %></dd>
+						<% if nodeinfo.network.mesh_vpn and nodeinfo.network.mesh_vpn.bandwidth_limit.enabled then -%>
+						<dt><%:Bandwidth limit%></dt>
+						<dd>
+							<% if nodeinfo.network.mesh_vpn.bandwidth_limit.egress then -%>
```
CodeFetch, line 136: @AiyionPrime Can't you do `(nodeinfo.network.mesh_vpn.bandwidth_limit.egress or -1) >= 0` here?

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
```diff
@@ -64,6 +64,20 @@
 		return v and translate('enabled') or translate('disabled')
 	end
 
+	local function formatBits(v, factor)
+		--to format a value given in kilobits use a factor of 1000
```
CodeFetch, line 68: Isn't the value always in kilobits? Why not simplify the whole thing and call the function formatkilobits and drop the second parameter?

CodeFetch: I think could be simplified to (not tested):
```
	local function formatKbits(v)
		units = {'k', 'M', 'G'}
		pow = math.floor(math.log(v)/math.log(1000))
		return string.format('%s %sbit', v, units[pow])
	end
```

AiyionPrime: This value is always kilobits, yes. I just didn't like to write the function less generic for an arbitrary factor.

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
AiyionPrime: Not sure what that does in lua. But I may read up on it.
In order to achieve what?

AiyionPrime:
I'll address the last two proposed changes once someone seconds them...
I think I like the code better the way it's now.

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
CodeFetch: An unneeded feature in embedded development is called dead code :grin:...

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
CodeFetch: `a=b or c` means `if b then a=b else a=c`
You use it here yourself: `factor = factor or 1`

My suggestion leads to not showing ingress or egress if it is not set...

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
CodeFetch:
```
	local function formatKbits(v, factor)
		units = {'', 'k', 'M', 'G'}
                bits=v*(factor or 1)
		pow = math.floor(math.log(bits)/math.log(1000))
		return string.format('%s %sbit', bits/1000^pow, units[pow+1])
	end
```

That's the simplified version if you insist on keeping the feature I guess.

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
CodeFetch: What was 0 meaning again? If it was "unlimited", too then there should be a `> 0 `...

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
CodeFetch: I just see no point in showing 32342424242443232 Gbit/s

CodeFetch: Maybe then change the code to show the mantissa or something.

CodeFetch: Ah... I've found a solution for the display problem... Use "%g" instead of "%s". This forces the shortest notation.

CodeFetch: So it becomes:
```
	local function formatKbits(v, factor)
		units = {'', 'k', 'M', 'G'}
                bits=v*(factor or 1)
		pow = math.min(math.floor(math.log(bits)/math.log(1000)), #units-2)
		return string.format('%g %sbit', bits/1000^pow, units[pow+1])
	end
```

CodeFetch: I think you shouldn't handle negative values here, but with my proposed change of the if-clause in status-page.html directly... Maybe add an "else *UNLIMITED*"...

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
AiyionPrime: This node runs the current codebase.
http://lilienstra-e.n.ffh.zone/cgi-bin/status

Values are not shown, if the bandwidth limit is 0.
There should not need to be a need to change the code, I think it already does what you intend?

CodeFetch: Isn't -1 unlimited? That's the case I cover...

AiyionPrime: in case egress or ingress is less than zero, the value is not added to the json struct.
And if the values are not there, they're sections are not shown.
There really is nothing to fix here, I think.
Please mark it as resolved when you agree.

https://github.com/freifunk-gluon/gluon/pull/2217/files#diff-c39d8b60296397950f7acc8f7b9015c5a7a37f4fb2a5ee7b09224477be5af678R69

CodeFetch: I can't resolve things.

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
```diff
@@ -64,6 +64,20 @@
...
+		factor = factor or 1
+		units = {'k', 'M', 'G'}
+		units[0] = ''
```
CodeFetch, line 71: I don't like what you are doing here... By making 0 an index it is no longer a sequence as tables in Lua start with index 1. It is a bad example for other people even though it works for your code.

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
```diff
@@ -64,6 +64,20 @@
...
+
+		bits = math.max(0, v*factor)
+		pow = math.floor(math.log(bits)/math.log(1000))
+		known_pow = math.min(pow, #units)
+
+		bits = bits/1000^known_pow
+		return string.format('%s %sbit', bits, units[known_pow])
```
CodeFetch, line 78: `return string.format('%**g** %sbit', bits, units[known_pow])` to make the numbers better readable.

CodeFetch (changes requested):

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
```diff
@@ -116,6 +130,17 @@ <h2><%:Overview%></h2>
 					<dt><%:Firmware%></dt><dd><%| nodeinfo.software.firmware.release %></dd>
 					<% if nodeinfo.software.fastd then -%>
 						<dt><%:Mesh VPN%></dt><dd><%| enabled(nodeinfo.software.fastd.enabled) %></dd>
+						<% if nodeinfo.network.mesh_vpn and nodeinfo.network.mesh_vpn.bandwidth_limit.enabled then -%>
+						<dt><%:Bandwidth limit%></dt>
+						<dd>
+							<% if nodeinfo.network.mesh_vpn.bandwidth_limit.egress then -%>
+							▲ <%| formatBits(nodeinfo.network.mesh_vpn.bandwidth_limit.egress, 1000) %>/s egress<br />
```
CodeFetch, line 137: Ingress/egress is incoherent with the config mode. There it is called Up/Downstream.

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
AiyionPrime: I learned the other day this would be the proper one-line solution for my intention:
```lua
units = {[0]='','k', 'M', 'G'}
```

I haven't tested it on gluons lua version, yet.

However this is perfectly fine lua, which makes the following calculation lighter by one offset, right?

CodeFetch: @AiyionPrime It's not fine Lua although it works. It's not a sequence anymore then. In your case there is no good reason for using this "hack" and you are giving a bad example.

CodeFetch: You could just write:
`string.format('%s %sbit', bits, units[known_pow+1])`

CodeFetch: You would make people believe they can just do this trick in any case which will lead to erroneous behavior. Therefore it is not fine, but **rude** Lua...

CodeFetch: By the way. I tried to make a joke if that was not obvious...

AiyionPrime: Which part?

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
AiyionPrime: Yep, both terms are used in gluon. E- and ingress are uci-terms.
You two outvote my opinion on this though; I'll implement it.

The line will then become `▼ 4 Downstream (Mbit/s)`.

> In general, I probably favor the terms used in config mode.

_Originally posted by @herbetom in https://github.com/freifunk-gluon/gluon/issues/2217#issuecomment-835493203_

CodeFetch:
@AiyionPrime Why don't you remove the respondd files here?

AiyionPrime:
Not sure if I understand your question correctly.
If I remove the commits, that modify those files the rest of this PR would become untestable by the github bot.

Actually I thought it was enough to mark this PR as **blocked** by 2216 in the very first post?
Despite that I think I have not seen a blocked PR yet, that does not contain the commits of the PR it is blocked by.
Is that really what you are suggesting?

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
AiyionPrime: > I don't like what you are doing here... By making 0 an index it is no longer a sequence as tables in Lua start with index 1. It is a bad example for other people even though it works for your code.

Not that I aimed for it to be one, in fact this is unrelated, but I think it actually is a sequence.
Whatever the joke is supposed to be I don't get it. Is this resolved?

_Originally documented in section 3.4.7 https://www.lua.org/manual/5.3/manual.html_

CodeFetch: No, the joke was "**rude** Lua".

Excerpt from https://www.lua.org/pil/3.6.html
```
For those that really want their arrays starting at 0, it is not difficult to write the following:

    days = {[0]="Sunday", "Monday", "Tuesday", "Wednesday",
            "Thursday", "Friday", "Saturday"}

Now, the first value, "Sunday", is at index 0. That zero does not affect the other fields, but "Monday" naturally goes to index 1, because it is the first list value in the constructor; the other values follow it. Despite this facility, I do not recommend the use of arrays starting at 0 in Lua. Remember that most functions assume that arrays start at index 1, and therefore will not handle such arrays correctly. 
```

There is no strong reason for using for doing this **hack** in your code as you could just write `string.format('%s %sbit', bits, units[known_pow+1])` which is even shorter. Therefore you give a bad example.

CodeFetch:
@AiyionPrime No, if you do two PRs the code of the PR that another PR relies on should not be included in the other PR.

That the build bot does not support dependent PRs is a different problem: https://github.com/isaacs/github/issues/959

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
```diff
@@ -64,6 +64,19 @@
 		return v and translate('enabled') or translate('disabled')
 	end
 
+	local function formatBits(v, factor)
+		--to format a value given in kilobits use a factor of 1000
+		factor = factor or 1
+		units = {[0]='', 'k', 'M', 'G'}
+
+		bits = math.max(0, v*factor)
```
CodeFetch, line 72: The math.max is unneeded here as you said that the data comes from respondd and can't be negative... It is unexpected behavior to get 0 when you input a negative value.

Furthermore I'm now even more convinced that you should omit the "factor" parameter as you could just call the function as
`formatBits(bits * factor)` instead of `formatBits(bits, factor)`. Having factor as a parameter just adds to complexity and confusion...

AiyionPrime: I feel like I'd reduce code quality if I removed such checks, just because I know what one data provider acts like.
Lets call it a weekend. Maybe another pair of eyes should decide on those enhancements you proposed?

CodeFetch: I don't think you would reduce code quality, but the opposite. There may be use-cases where negative values might be useful. Handling negative values as 0 is not a good solution.

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
```diff
@@ -64,6 +64,19 @@
...
+		pow = math.floor(math.log(bits)/math.log(1000))
```
lemoer, line 73:
```suggestion
		pow = math.floor(math.log(math.max(math.abs(bits),1))/math.log(1000))
```

The log() function will become negative for input values smaller than 1. This change should fix this.

Furthemore, this change should enable us to handle negative input values correctly.

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
CodeFetch: Why do you think having "factor" as a second parameter is wishful if it just does a multiplication while it is possible to do the multiplication beforehand with less code and better readability? I don't understand why you think anyone experienced in programming would have a different opinion on that...

lemoer:
```suggestion
		bits = v*factor
```

Together with the other suggestion, this should be fine now.

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
```diff
@@ -64,6 +64,19 @@
 		return v and translate('enabled') or translate('disabled')
 	end
 
+	local function formatBits(v, factor)
+		--to format a value given in kilobits use a factor of 1000
+		factor = factor or 1
+		units = {[0]='', 'k', 'M', 'G'}
```
lemoer, line 70: Please make all variables local.
```suggestion
		local units = {[0]='', 'k', 'M', 'G'}
```

lemoer (changes requested):

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
AiyionPrime: 7618dcb07a3aa0c780d5cf975f4f3d26c8aa1b96

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
AiyionPrime: 7618dcb07a3aa0c780d5cf975f4f3d26c8aa1b96

AiyionPrime:
Compiled and runs on http://lilienstra-e.n.ffh.zone/

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
AiyionPrime: http://lilienstra-e.n.ffh.zone/

AiyionPrime: negative values work, there's only one paramter, zero values work now.

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
AiyionPrime: a73c007d7f0cd123de994b0d58c9382fa1fc4157

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
AiyionPrime: @CodeFetch this is outdated, don't you agree?

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
```diff
@@ -116,6 +126,17 @@ <h2><%:Overview%></h2>
 					<dt><%:Firmware%></dt><dd><%| nodeinfo.software.firmware.release %></dd>
 					<% if nodeinfo.software.fastd then -%>
 						<dt><%:Mesh VPN%></dt><dd><%| enabled(nodeinfo.software.fastd.enabled) %></dd>
+						<% if nodeinfo.network.mesh_vpn and nodeinfo.network.mesh_vpn.bandwidth_limit.enabled then -%>
+						<dt><%:Bandwidth limit%></dt>
+						<dd>
+							<% if nodeinfo.network.mesh_vpn.bandwidth_limit.egress then -%>
+							▲ <%| formatBits(nodeinfo.network.mesh_vpn.bandwidth_limit.egress*1000) %>/s egress<br />
+							<%- end %>
```
CodeFetch, line 134: Still should be Upstream/Downstream for consistency. Other things look good to me (Just philosophical issues , but if it works, it works).

CodeFetch (changes requested):

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
AiyionPrime: Lets use the thread you already opened for this.
https://github.com/freifunk-gluon/gluon/pull/2217#discussion_r645780228

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
CodeFetch: @AiyionPrime I don't understand... Why do you want to introduce a term which is potentially confusing for the users? Why do you want to break with consistency? Or better say why do we have this discussion if two people already said it should be consistent?

AiyionPrime: 0a64f04a28a04a356e9033e69bee007c24ef6004

AiyionPrime: @CodeFetch You really need to start reading comments above.
> I'll implement it.

_Originally posted by @AiyionPrime in https://github.com/freifunk-gluon/gluon/pull/2217#discussion_r645821640_

CodeFetch (approved):
Thanks! Please remove the respondd provider commit once the tests have passed and I think it is ready to merge.

@blocktrron @lemoer Can you have a last look, so we can get this merged?

AiyionPrime:
The current build runs on http://meshvpnabstracttest.n.ffh.zone/cgi-bin/status.

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
```diff
@@ -64,6 +64,16 @@
 		return v and translate('enabled') or translate('disabled')
 	end
 
+	local function formatBits(bits)
+		local units = {[0]='', 'k', 'M', 'G'}
+
+		local pow = math.floor(math.log(math.max(math.abs(bits), 1))/math.log(1000))
+		local known_pow = math.min(pow, #units)
+
+		local significand = bits/1000^known_pow
```
blocktrron, line 73: `local significand = bits / (1000 ^ known_pow)`

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
```diff
@@ -64,6 +64,16 @@
 		return v and translate('enabled') or translate('disabled')
 	end
 
+	local function formatBits(bits)
+		local units = {[0]='', 'k', 'M', 'G'}
+
+		local pow = math.floor(math.log(math.max(math.abs(bits), 1))/math.log(1000))
```
blocktrron, line 70: `local pow = math.floor(math.log(math.max(math.abs(bits), 1)) / math.log(1000))`

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
AiyionPrime: 2401f5ca7902064926b2c059dc28e49163747fac

package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
AiyionPrime: 2401f5ca7902064926b2c059dc28e49163747fac

AiyionPrime:
Compiled and deployed on http://meshvpnabstracttest.n.ffh.zone/

herbetom:
Now it's to late. But swapping Upstream and Downstream to get a ⌛ could have been a nice touch ;)

AiyionPrime:
@herbetom Would have been fine with me. But thtas nothing I open another PR for, sorry.
But feel free to go ahead. If you're fast, you might get it in in time :)

Thanks everyone for reviewing this!

PR #2219, opened 2021-05-13
gluon-mesh-vpn-core: respondd active, enabled

AiyionPrime:
Adds another provider to the same nodeinfo branch as #2216,
adds the fields `enabled` (bool) and `active` (string).

The contents of are taken from @blocktrron s lua api implemented in #2102 and therefore depends on it.

The code compiled and runs on http://meshvpnabstracttest.n.ffh.zone/

This is an excerpt from its nodeinfo:

```json
 "mesh_vpn": {
  "active": "fastd",
  "enabled": false
},
```

This resolves another part of #644.

### DIFF
--- a/package/gluon-mesh-vpn-core/src/respondd.c
+++ b/package/gluon-mesh-vpn-core/src/respondd.c
@@ -1,0 +1,104 @@
+/*
+  Copyright (c) 2021, Aiyion <gluon@aiyionpri.me>
+  Copyright (c) 2016, Matthias Schiffer <mschiffer@universe-factory.net>
+  All rights reserved.
+  Redistribution and use in source and binary forms, with or without
+  modification, are permitted provided that the following conditions are met:
+    1. Redistributions of source code must retain the above copyright notice,
+       this list of conditions and the following disclaimer.
+    2. Redistributions in binary form must reproduce the above copyright notice,
+       this list of conditions and the following disclaimer in the documentation
+       and/or other materials provided with the distribution.
+  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+  IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+  DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+  FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+  DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+  SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+  OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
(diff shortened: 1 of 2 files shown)


lemoer:
In harmony with #2102, I would suggest:

```json
"mesh_vpn": {
  "proto": "fastd",
  "enabled": false
},
```

AiyionPrime:
`function M.get_active_proto()`

I'd be fine with proto.

AiyionPrime:
Altered `proto` to `provider`. Compiles and runs on http://meshvpnabstracttest.n.ffh.zone/cgi-bin/status

AiyionPrime:
@mweinelt, @blocktrron this one works and is waiting for review.

package/gluon-mesh-vpn-core/src/respondd.c
```diff
@@ -0,0 +1,102 @@
...
+		line = NULL;
+	}
+	return line;
+}
+
+static struct json_object * get_mesh_vpn_enabled() {
+	int enabled = -1;
+	char *line = read_stdout("exec lua -e 'print(require(\"gluon.mesh-vpn\").enabled())'");
+
+	if (strncmp(line, "true", 4) == 0)
```
blocktrron, line 61: No need to use strncmp when comparing with literals. They are always proper terminated.

Also 
```
if (!strcmp(line, "true"))
```

package/gluon-mesh-vpn-core/src/respondd.c
```diff
@@ -0,0 +1,102 @@
...
+		enabled=1;
+	if (strncmp(line, "false", 5) == 0)
```
blocktrron, line 63: same

package/gluon-mesh-vpn-core/src/respondd.c
```diff
@@ -0,0 +1,102 @@
...
+	}
+	return line;
+}
+
+static struct json_object * get_mesh_vpn_enabled() {
+	int enabled = -1;
+	char *line = read_stdout("exec lua -e 'print(require(\"gluon.mesh-vpn\").enabled())'");
+
+	if (strncmp(line, "true", 4) == 0)
+		enabled=1;
```
blocktrron, line 62: spacing around `=` when performing assignments.

package/gluon-mesh-vpn-core/src/respondd.c
```diff
@@ -0,0 +1,102 @@
...
+	if (strncmp(line, "false", 5) == 0)
+		enabled=0;
```
blocktrron, line 64: spacing around `=` when performing assignments.

blocktrron (changes requested):

package/gluon-mesh-vpn-core/src/respondd.c
```diff
@@ -0,0 +1,103 @@
...
+	if (r >= 0) {
+		// The len given by getline is the buffer size, not the string length
+		len = strlen(line);
+
+		if (len && line[len-1] == '\n')
+			line[len-1] = 0;
+	} else {
+		free(line);
+		return NULL;
+	}
```
blocktrron, line 54: I was a bit confused of the desired outcome for 1-char strings only containing `\n`, however as the else block is only for error handling, how about this:

```      
ssize_t r = getline(&line, &len, f);

pclose(f);
if (r < 0) {
	free(line);
	return NULL;
}

/* The len given by getline is the buffer size, not the string length */
len = strlen(line);

if (len && line[len-1] == '\n')
	line[len-1] = 0;

return line;
```

Note also the change from CPP comment style to C.

blocktrron (changes requested):

package/gluon-mesh-vpn-core/src/respondd.c
AiyionPrime: implemented

blocktrron:
@AiyionPrime Can you squash your comments so I can merge this?

AiyionPrime:
Sure, I hoped github had a button to perform the autosquash?
But I'll get to it in a few minutes.

blocktrron:
Sadly, it doesn't have one.

AiyionPrime:
@blocktrron that's it

blocktrron:
Let's do this.

PR #2222, opened 2021-05-13
gluon-web-private-wifi: add support for 802.11r

mweinelt:
Add support for Fast BSSID transition for the private WLAN (can we rename the package, please?) feature.

The config mode gets extended by a check box for 802.11r, and an input for the mobility domain, which needs to match between vifs, so that clients will try to roam between them. Additionally `ft_psk_generate_local` is set, which only works for PSK protected networks.

- [ ] Translations
- [ ] Defaults
  - I have no clue where to properly set the defaults...
- [ ] Testing (written for @azrdev, hoping for his feedback).
- [ ] Open questions
  - 802.11k (Radio Resource Management)
    - probably useless without a tool like usteer
  - 802.11v (Wireless Network Management, e.g. power savings and assisted roaming)

### DIFF
--- a/package/gluon-web-private-wifi/luasrc/lib/gluon/config-mode/model/admin/privatewifi.lua
+++ b/package/gluon-web-private-wifi/luasrc/lib/gluon/config-mode/model/admin/privatewifi.lua
@@ -1,4 +1,6 @@
 local uci = require("simple-uci").cursor()
+local unistd = require 'posix.unistd'
+
 local platform = require 'gluon.platform'
 local wireless = require 'gluon.wireless'
 
@@ -45,6 +47,12 @@ if platform.device_supports_mfp(uci) then
 end
 mfp.default = uci:get('wireless', primary_iface, 'ieee80211w') or "0"
 
+local ieee80211r = nil
+if unistd.access('/lib/gluon/features/wpa3') then
+	ieee80211r = s:option(Flag, "80211r", translate("Fast BSSID transition"))
+	ieee80211r:depends(encryption, "psk2")
+	ieee80211r.default = uci:get('wireless', primary_iface, 'ieee80211r') or false
+end
 
 function f:write()
 	wireless.foreach_radio(uci, function(radio, index)
@@ -56,16 +64,20 @@ function f:write()
 			local macaddr = wireless.get_wlan_mac(uci, radio, index, 4)
 
 			uci:section('wireless', 'wifi-iface', name, {
-				device     = radio_name,
-				network    = 'wan',
-				mode       = 'ap',
-				encryption = encryption.data,
-				ssid       = ssid.data,
-				key        = key.data,
-				macaddr    = macaddr,
-				ifname     = suffix and 'wan' .. suffix,
-				disabled   = false,
+				device                = radio_name,
+				network               = 'wan',
+				mode                  = 'ap',
+				encryption            = encryption.data,
+				ssid                  = ssid.data,
+				key                   = key.data,
+				macaddr               = macaddr,
+				ifname                = suffix and 'wan' .. suffix,
+				disabled              = false,
 			})
+
+			if ieee80211r ~= nil then
+				uci:set('wireless', name, 'ieee80211r', ieee80211r.data)
+			end
 
 			-- hostapd-mini won't start in case 802.11w is configured
 			if platform.device_supports_mfp(uci) then


rotanid:
> for the private WLAN (can we rename the package, please?) feature.

sure, but in fact we haven't tried renaming packages before, or have we?

azrdev:
I edited `/lib/gluon/config-mode/model/admin/privatewifi.lua` in-place, and had to do some adjustments due to the variable name mangling. With the following code I end up with an HTTP 500 shown in configmode/privatewifi webinterface: 
~~~
Failed to execute dispatcher target for entry '/admin/privatewifi'.
/usr/lib/lua/simple-uci.lua:22: bad argument #3 to '?' (string expected, got nil)
~~~

file `/lib/gluon/config-mode/model/admin/privatewifi.lua`:
~~~lua
local e=require("simple-uci").cursor()
local r=require'gluon.platform'
local d=require'gluon.wireless'
local o='wan_radio0'
local h=Form(translate("Private WLAN"))
local i=h:section(Section,nil,translate(
'Your node can additionally extend your private network by bridging the WAN interface '
..'with a separate WLAN. This feature is completely independent of the mesh functionality. '
..'Please note that the private WLAN and meshing on the WAN interface should not be enabled '
..'at the same time.'
))
local t=i:option(Flag,"enabled",translate("Enabled"))
t.default=e:get('wireless',o)and not e:get_bool('wireless',o,"disabled")
local n=i:option(Value,"ssid",translate("Name (SSID)"))
n:depends(t,true)
n.datatype="maxlength(32)"
n.default=e:get('wireless',o,"ssid")
local s=i:option(Value,"key",translate("Key"),translate("8-63 characters"))
s:depends(t,true)
s.datatype="wpakey"
s.default=e:get('wireless',o,"key")
local a=i:option(ListValue,"encryption",translate("Encryption"))
a:depends(t,true)
a:value("psk2",translate("WPA2"))
if r.device_supports_wpa3()then
a:value("psk3-mixed",translate("WPA2 / WPA3"))
a:value("psk3",translate("WPA3"))
end
a.default=e:get('wireless',o,'encryption')or"psk2"

local ieee80211r = i:option(Flag, "80211r", translate("Fast BSSID transition"))
ieee80211r.default = e:get('wireless', primary_iface, 'ieee80211r') or false

local mobility_domain = i:option(Value, "mobility_domain", "Mobility Domain")
mobility_domain:depends(ieee80211r, true)
mobility_domain = e:get('wireless', primary_iface, 'mobility_domain') or '4f57'

local i=i:option(ListValue,"mfp",translate("Management Frame Protection"))
i:depends(t,true)
i:value("0",translate("Disabled"))
if r.device_supports_mfp(e)then
i:value("1",translate("Optional"))
i:value("2",translate("Required"))
end
i.default=e:get('wireless',o,'ieee80211w')or"0"

function h:write()
d.foreach_radio(e,function(l,u)
local h=l['.name']
local o="wan_"..h
if t.data then
local t=d.get_wlan_mac(e,l,u,4)
e:section('wireless','wifi-iface',o,{
device=h,
network='wan',
mode='ap',
encryption=a.data,
ssid=n.data,
key=s.data,
ieee80211r=ieee80211r.data,
mobility_domain=mobility_domain.data,
ft_psk_generate_local=true,
macaddr=t,
disabled=false,
})
if r.device_supports_mfp(e)then
e:set('wireless',o,'ieee80211w',i.data)
else
e:delete('wireless',o,'ieee80211w')
end
else
e:set('wireless',o,"disabled",true)
end
end)
e:commit('wireless')
end
return h
~~~

mweinelt:
Can you use the completely unmangeled code, please? Line 22 would be the encryption select list.

blocktrron:
FT is not supported with hostapd-mini (see https://github.com/openwrt/openwrt/blob/master/package/network/services/hostapd/files/hostapd-mini.config#L151 )

Furthermore, `ft_psk_generate_local` does not work with SAE, so in the current state, this would introduce broken 11r on the devices when used with SAE

azrdev:
indeed with the complete unmangled file I get this UI:
image

and after enabling the "fast bssid transition" checkbox another row "mobility domain". I left the latter empty and the checkbox enabled, resulting in 

~~~sh
root@64347-arya:~# uci show  | grep wan_radio
...
wireless.wan_radio0.ieee80211r='1'
wireless.wan_radio0.ft_psk_generate_local='1'
...
wireless.wan_radio1.ieee80211r='1'
wireless.wan_radio1.ft_psk_generate_local='1'
~~~

as expected, only `mobility_domain = ` is missing.

blocktrron:
I think we can safely drop the mobility-domain, as OpenWrt sets a default by creating a hash over the SSID:

https://github.com/openwrt/openwrt/blob/ec6293febc244d187e71a6e54f44920be679cde4/package/network/services/hostapd/files/hostapd.sh#L747

mweinelt:
> I think we can safely drop the mobility-domain, as OpenWrt sets a default by creating a hash over the SSID:
> 
> [openwrt/openwrt@`ec6293f`/package/network/services/hostapd/files/hostapd.sh#L747](https://github.com/openwrt/openwrt/blob/ec6293febc244d187e71a6e54f44920be679cde4/package/network/services/hostapd/files/hostapd.sh#L747)

We just agreed to drop the mobility domain as introducing compatbility with 3rd party (non OpenWrt) APs is an advanced feature and if required `mobility_domain` could be set within uci manually for now.

package/gluon-web-private-wifi/luasrc/lib/gluon/config-mode/model/admin/privatewifi.lua
```diff
@@ -45,6 +45,13 @@ if platform.device_supports_mfp(uci) then
 end
 mfp.default = uci:get('wireless', primary_iface, 'ieee80211w') or "0"
 
+local ieee80211r = s:option(Flag, "80211r", translate("Fast BSSID transition"))
```
mweinelt, line 48:
```suggestion
local ieee80211r = s:option(Flag, "80211r", translate("Fast Roaming"))
```

package/gluon-web-private-wifi/luasrc/lib/gluon/config-mode/model/admin/privatewifi.lua
```diff
@@ -45,6 +45,13 @@ if platform.device_supports_mfp(uci) then
...
+ieee80211r.default = uci:get('wireless', primary_iface, 'ieee80211r') or false
+
+local mobility_domain = s:option(Value, "mobility_domain", "Mobility Domain")
+mobility_domain:depends(ieee80211r, true)
+mobility_domain = uci:get('wireless', primary_iface, 'mobility_domain') or '4f57'
```
mweinelt, line 53: Will be dropped, since it is an advanced feature and we are compatible with OpenWrt anyhow based on the matching ESSID.

CodeFetch:
@mweinelt Lemoer and I looked at 802.11k/v support a while ago. There is a lot more work to put into that. I think it is out-of-scope for this PR. The best approach so far is DAWN https://github.com/berlin-open-wireless-lab/DAWN. But for using that for Gluon it needs something like [meshrealmd](https://pad.freifunk.net/p/meshrealmd) (unfinished concept).

mweinelt:
We don't want to extend support for private WLANs with additional daemons, that is honestly too much.

If you are intereste in these kinds of things, please look into https://github.com/blogic/usteer.

CodeFetch:
@mweinelt usteer isn't going to happen either. So go for it...

mweinelt:
> @mweinelt usteer isn't going to happen either. So go for it...

Uh-huh. https://github.com/freifunk-gluon/gluon/tree/next-usteer

blocktrron:
This get's a bit more complicated with SAE. @mweinelt and me discussed and we currently not have time to progress this.

Anyone feel free to pick this up.

PR #2223, opened 2021-05-14
Simplify save/commit handling of config wizard sections

NeoRaider:
- Add a write() to form sections, so the wizard sections can call `save()` on the UCI cursor themselves
- Always run reconfigure after the wizard. This commits all packages.

This also avoids the `uci_sections()` handling in #2102.

### DIFF
--- a/package/gluon-config-mode-core/luasrc/lib/gluon/config-mode/model/gluon-config-mode/wizard.lua
+++ b/package/gluon-config-mode-core/luasrc/lib/gluon/config-mode/model/gluon-config-mode/wizard.lua
@@ -1,15 +1,6 @@
 local util = require "gluon.util"
 local uci = require("simple-uci").cursor()
 
-
-local wizard = {}
-for _, entry in ipairs(util.glob('/lib/gluon/config-mode/wizard/*')) do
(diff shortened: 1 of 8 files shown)


blocktrron (approved):
Tested on x86-generic.

package/gluon-config-mode-core/luasrc/lib/gluon/config-mode/model/gluon-config-mode/wizard.lua
```diff
@@ -18,21 +9,10 @@ local s = f:section(Section)
 s.template = "wizard/welcome"
 s.package = "gluon-config-mode-core"
 
-local commit = {'gluon-setup-mode'}
```
blocktrron, line 21: We still need to save `gluon-setup-mode`. Currently, everything else is saved w/o problems but we're stuck in setup mode.

blocktrron (changes requested):
Missed the face node was booting back so setup mode and only checked if after a reboot all settings in the wizard have been saved.

blocktrron (approved):
Works now on x86-generic as well as ar71xx-generic.

PR #2229, opened 2021-06-08
gluon-check-connection: initial commit

CodeFetch:
This commit adds a new package which can be used for scheduled connectivity checks using
ping. Performing ping tests to global targets was previously decided to be the best fit for
the offline SSID changer. It has been tested and seems to work properly.
I'll do further tests to be sure.


CodeFetch:
Will reopen it in community repository...

PR #2234, opened 2021-06-09
treewide: use lua-jsonc instead of lua-cjson for JSON handling during build

NeoRaider:
As proposed in https://github.com/freifunk-gluon/gluon/pull/2023#discussion_r648459621

Depends on https://github.com/freifunk-gluon/packages/pull/245

I have checked that the generated `site.json` is identical with lua-cjson and lua-jsonc.


AiyionPrime:
http://testbench-01.n.ffh.zone/cgi-bin/status contains both this and [freifunk-gluon/packages#245](https://github.com/freifunk-gluon/packages/pull/245).

Compiled without issues @neoraider you should have ssh-access on the device.

PR #2235, opened 2021-06-10
Update master to OpenWrt 21.02

mweinelt:
As OpenWrt 19.07 has a projected EOL within this year we need to update to OpenWrt 21.02.

TODO:
- Remove untested device support for non-broken devices


targets/x86-64
```diff
@@ -1,6 +1,9 @@
...
-factory_image('x86-64', 'combined-squashfs', '.vdi')
-factory_image('x86-64', 'combined-squashfs', '.vmdk')
-sysupgrade_image('x86-64', 'combined-squashfs', '.img.gz')
+packages {
+        'kmod-gpio-nct5104d',
+        'kmod-leds-gpio',
+        'kmod-pcengines-apuv2',
+}
+
+device('x86-64', 'generic')
```
NeoRaider, line 9: By switching to the proper device handlers, we can also change the
```
try_config('TARGET_MULTI_PROFILE', true)
try_config('TARGET_PER_DEVICE_ROOTFS', true)
```
to `config(...)` in `targets/generic`.

targets/x86-geode
```diff
@@ -18,5 +18,9 @@ packages {
 	'kmod-via-velocity',
 }
 
-factory_image('x86-geode', 'combined-squashfs', '.img.gz')
-sysupgrade_image('x86-geode', 'combined-squashfs', '.img.gz')
+device('x86-geode', 'generic', {
+	factory = '-squashfs-combined',
+	factory_ext = '.img.gz',
```
NeoRaider, line 23: What about VMDK/VDI images?

mweinelt: For geode?

NeoRaider:
We can also finally drop
```
config('BUSYBOX_CUSTOM', true)
config('BUSYBOX_CONFIG_FEATURE_PREFER_IPV4_ADDRESS', false)
```
from `targets/generic`, as this is the default now.

targets/x86-geode
NeoRaider: Oops, I meant to add this to `targets/x64.inc`.

targets/x86.inc
```diff
@@ -42,3 +39,10 @@ packages {
 	'kmod-mt7603',
 	'kmod-mt7615e',
 }
+
+defaults {
+	factory = '-squashfs-combined',
+	factory_ext = '.img.gz',
```
NeoRaider, line 45: What about VMDK/VDI images?

NeoRaider:
```suggestion
	factory_ext = {'.img.gz', '.vmdk', '.vdi'},
```

PR #2237, opened 2021-06-13
Run site check for manual package installations

NeoRaider:
Install the check_site main script into the Gluon image to allow using it for later package installations, as needed for #2023.

As an additional benefit, with this change the value of TOPDIR doesn't leak into ipk packages anymore, improving reproducibility across different build hosts.


NeoRaider:
An alternative to installing `gluon-check-site.lua` to `/usr/lib/opkg/info/` would be to copy the whole content of the script into each package's postinst section (as the postinst script is deleted by opkg after installation, this wouldn't waste space either). The current solution feels a bit more hacky, but it should result in better line number output in backtraces when debugging the site checker itself.

lemoer (approved):
The file `check_site.lua` is about 8KB, which is not too much space. In my opinion it is worth spending this to have better backtraces.

I did some tests with this PR and a modified version of #2023. I could not find any problems.

So this is an ACK from my side.

NeoRaider:
Converted to draft, blocked by #2271 - with the removal of the `no_opkg` target flag, the location of `gluon-check-site.lua` doesn't make sense anymore.

lemoer:
Can we please merge this PR ahead of #2271 to reduce the level of unnecessaryily blocked PR's? IMHO #2271 could then contain a change of the location of *gluon-check-site.lua*. If this PR would have already been merged, this would have to be done in #2271 anways.

NeoRaider:
Updated - while this is influenced by #2271, both PRs can be merged independently, so they don't need to block each other.

lemoer (approved):
LGTM.

PR #2240, opened 2021-06-14
gluon-mesh-vpn-core: fix potential null pointer dereferences

blocktrron:
In case the limit_ingress or limit_egress options are not present in
gluon's mesh_vpn section the respondd provider compares a string literal
with a NULL pointer, crashing respondd.

Check both pointers prior to comparing them in order to mitigate this
issue.

Suggested-by: Matthias Schiffer <mschiffer@universe-factory.net>

-----

read_stdout can return NULL and thus the return value need to be checked
prior to accessing it.

### DIFF
--- a/package/gluon-mesh-vpn-core/src/respondd.c
+++ b/package/gluon-mesh-vpn-core/src/respondd.c
@@ -110,6 +110,9 @@ static struct json_object * get_mesh_vpn_enabled() {
 	int enabled = -1;
 	char *line = read_stdout("exec lua -e 'print(require(\"gluon.mesh-vpn\").enabled())'");
 
+	if (!line)
+		return NULL;
+
 	if (!strcmp(line, "true"))
 		enabled = 1;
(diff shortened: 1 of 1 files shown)


package/gluon-mesh-vpn-core/src/respondd.c
```diff
@@ -110,6 +110,9 @@ static struct json_object * get_mesh_vpn_enabled() {
 	int enabled = -1;
 	char *line = read_stdout("exec lua -e 'print(require(\"gluon.mesh-vpn\").enabled())'");
 
+	if (!line)
+		return NULL
```
AiyionPrime, line 114:
```suggestion
		return NULL;
```

AiyionPrime (changes requested):
The rest looks good. Thanks you two -.-'

AiyionPrime:
With the suggested fix the code compiles and runs on http://lilienstra-e.n.ffh.zone/ .
You two should have ssh access.

PR #2244, opened 2021-06-17
gluon-core: implement popen3() in gluon/util.lua

### LINKED ISSUE #2243: popen2/3 in gluon
Recently I wrote some code, that will both need both the contents of the standard I/O streams, as well as the return value of an executed process.

in #gluon I got some references using fork() and pexec(), like we already do in
https://github.com/freifunk-gluon/gluon/blob/a2be901803469d378d6469551a65c20f744aefff/package/gluon-web-admin/luasrc/lib/gluon/config-mode/model/admin/remote.lua#L80
and https://github.com/freifunk-gluon/gluon/blob/069c8b2035c53b2235c716ec7c640e892351f5b2/package/gluon-web-admin/luasrc/lib/gluon/config-mode/controller/admin/upgrade.lua#L39

Furthermore:
> Might make sense to add a helper to gluon.util

_Originally posted by @NeoRaider  in hackint#gluon on 09.06.21 at 18:47:51_

**edit**
I'd therefore like to add a helper function `popen2(path, ...)`,
that returns a newly forked pid, as well as the two streams on success,
or nil and and possibly an error message, if wanted.

AiyionPrime:
This intends to resolve #2243.

I'd go forward and add another commit, that uses this code in https://github.com/freifunk-gluon/gluon/blob/a2be901803469d378d6469551a65c20f744aefff/package/gluon-web-admin/luasrc/lib/gluon/config-mode/model/admin/remote.lua#L78

### DIFF
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
@@ -1,6 +1,8 @@
 local bit = require 'bit'
+local posix_fcntl = require 'posix.fcntl'
 local posix_glob = require 'posix.glob'
 local posix_syslog = require 'posix.syslog'
+local posix_unistd = require 'posix.unistd'
 local hash = require 'hash'
 local sysconfig = require 'gluon.sysconfig'
 local site = require 'gluon.site'
@@ -188,4 +190,73 @@ function M.log(message, verbose)
 	posix_syslog.syslog(posix_syslog.LOG_INFO, message)
 end
 
+M.PipePolicies = {
+	DISCARD=-1,
+	INHERIT=0,
+	CREATE=1
+}
+
+-- Execute a program found using command PATH search, like the shell.
+-- Return the pid, as well as the I/O streams as pipes or nil on error.
+function M.popen3(policies, path, ...)
+	local pipes = {}
+	local intern = {}
+	local extern = {}
+
+	for fd, policy in pairs(policies) do
+		if M.PipePolicies.CREATE==policy then
+			pipes[fd]={posix_unistd.pipe()}
+			if posix_unistd.STDIN_FILENO==fd then
+				intern[fd]=pipes[fd][1]
+				extern[fd]=pipes[fd][2]
+			else
+				intern[fd]=pipes[fd][2]
+				extern[fd]=pipes[fd][1]
+			end
+		end
+	end
+
+	-- intern: r0, w1, w2
+	-- extern: w0, r1, r2
+
+	local pid = posix_unistd.fork()
+
+	if pid < 0 then
+		return nil
+	elseif pid == 0 then
+		local null=-1;
+		if M.contains(policies, M.PipePolicies.DISCARD) then
+			-- only open, if there's anything to discard
+			null = posix_fcntl.open('/dev/null', posix_fcntl.O_WRONLY)
+		end
+
+		for fd, policy in pairs(policies) do
+			if M.PipePolicies.DISCARD==policy then
+				if posix_unistd.STDIN_FILENO~=fd then
+					posix_unistd.dup2(null, fd)
+				end
+			elseif M.PipePolicies.CREATE==policy then
+				-- only close these, if they exist
+				posix_unistd.close(extern[fd])
+				posix_unistd.dup2(intern[fd], fd)
+				-- close all the dups
+				posix_unistd.close(intern[fd])
+			end
+		end
+
+		-- close potential null
+		if null > 2 then
+			posix_unistd.close(null)
+		end
+
+		posix_unistd.execp(path, ...)
+		posix_unistd._exit(127)
+	end
+
+	for _, v in pairs(intern) do
+		posix_unistd.close(v)
+	end
+
+	return pid, extern
+end
 return M
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
@@ -193,7 +193,6 @@ end
 M.subprocess = {}
 
 M.subprocess.DEVNULL = -1
-M.subprocess.INHERIT = nil
 M.subprocess.PIPE = 1
 
 -- Execute a program found using command PATH search, like the shell.
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
@@ -202,15 +202,15 @@ function M.subprocess.popen(path, argt, options)
 	local childfds = {}
 	local parentfds = {}
 
-	for fd, option in pairs(options) do
+	for iostream, option in pairs(options) do
 		if option == M.subprocess.PIPE then
 			local piper, pipew = posix_unistd.pipe()
-			if fd == posix_unistd.STDIN_FILENO then
-				childfds[fd] = piper
-				parentfds[fd] = pipew
+			if iostream == "stdin" then
+				childfds[iostream] = piper
+				parentfds[iostream] = pipew
 			else
-				childfds[fd] = pipew
-				parentfds[fd] = piper
+				childfds[iostream] = pipew
+				parentfds[iostream] = piper
 			end
 		end
 	end
@@ -224,20 +224,21 @@ function M.subprocess.popen(path, argt, options)
 		return nil, errmsg, errnum
 	elseif pid == 0 then
 		local null = -1
+		local stdiostreams = {["stdin"] = 0, ["stdout"] = 1, ["stderr"] = 2}
 		if M.contains(options, M.subprocess.DEVNULL) then
 			-- only open, if there's anything to discard
 			null = posix_fcntl.open('/dev/null', posix_fcntl.O_RDWR)
 		end
 
-		for fd, option in pairs(options) do
+		for iostream, option in pairs(options) do
 			if option == M.subprocess.DEVNULL then
-				posix_unistd.dup2(null, fd)
+				posix_unistd.dup2(null, stdiostreams[iostream])
 			elseif option == M.subprocess.PIPE then
 				-- only close these, if they exist
-				posix_unistd.close(parentfds[fd])
-				posix_unistd.dup2(childfds[fd], fd)
+				posix_unistd.close(parentfds[iostream])
+				posix_unistd.dup2(childfds[iostream], stdiostreams[iostream])
 				-- close all the dups
-				posix_unistd.close(childfds[fd])
+				posix_unistd.close(childfds[iostream])
 			end
 		end
 
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
@@ -190,10 +190,13 @@ function M.log(message, verbose)
 	posix_syslog.syslog(posix_syslog.LOG_INFO, message)
 end
 
--- Blindly close the file descriptors in a given table.
-local function close_fds(descriptors)
-	for _, fd in pairs(descriptors) do
-		posix_unistd.close(fd)
+-- Blindly close the file descriptors in an arbitrary amount of given tables.
+local function close_fds(...)
+	for i=1, select("#", ...) do
+		local descriptor_table = select(i, ...)
+		for _, fd in pairs(descriptor_table) do
+			posix_unistd.close(fd)
+		end
 	end
 end
 
@@ -228,8 +231,7 @@ function M.subprocess.popen(path, argt, options)
 	local pid, errmsg, errnum = posix_unistd.fork()
 
 	if pid == nil then
-		close_fds(childfds)
-		close_fds(parentfds)
+		close_fds(childfds, parentfds)
 		return nil, errmsg, errnum
 	elseif pid == 0 then
 		local null = -1
@@ -243,12 +245,9 @@ function M.subprocess.popen(path, argt, options)
 			if option == M.subprocess.DEVNULL then
 				posix_unistd.dup2(null, stdiostreams[iostream])
 			elseif option == M.subprocess.PIPE then
-				-- only close these, if they exist
-				posix_unistd.close(parentfds[iostream])
 				posix_unistd.dup2(childfds[iostream], stdiostreams[iostream])
-				-- close all the dups
-				posix_unistd.close(childfds[iostream])
 			end
+			close_fds(childfds, parentfds)
 		end
 
 		-- close potential null
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
@@ -203,14 +203,14 @@ function M.subprocess.popen(policies, path, ...)
 	local parentfds = {}
 
 	for fd, policy in pairs(policies) do
-		if policy==M.subprocess.PIPE then
+		if policy == M.subprocess.PIPE then
 			local piper, pipew = posix_unistd.pipe()
-			if fd==posix_unistd.STDIN_FILENO then
-				childfds[fd]=piper
-				parentfds[fd]=pipew
+			if fd == posix_unistd.STDIN_FILENO then
+				childfds[fd] = piper
+				parentfds[fd] = pipew
 			else
-				childfds[fd]=pipew
-				parentfds[fd]=piper
+				childfds[fd] = pipew
+				parentfds[fd] = piper
 			end
 		end
 	end
@@ -223,18 +223,18 @@ function M.subprocess.popen(policies, path, ...)
 	if pid == nil then
 		return nil, errmsg, errnum
 	elseif pid == 0 then
-		local null=-1
+		local null = -1
 		if M.contains(policies, M.subprocess.DEVNULL) then
 			-- only open, if there's anything to discard
 			null = posix_fcntl.open('/dev/null', posix_fcntl.O_WRONLY)
 		end
 
 		for fd, policy in pairs(policies) do
-			if policy==M.subprocess.DEVNULL then
-				if fd~=posix_unistd.STDIN_FILENO then
+			if policy == M.subprocess.DEVNULL then
+				if fd ~= posix_unistd.STDIN_FILENO then
 					posix_unistd.dup2(null, fd)
 				end
-			elseif policy==M.subprocess.PIPE then
+			elseif policy == M.subprocess.PIPE then
 				-- only close these, if they exist
 				posix_unistd.close(parentfds[fd])
 				posix_unistd.dup2(childfds[fd], fd)
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
@@ -190,20 +190,20 @@ function M.log(message, verbose)
 	posix_syslog.syslog(posix_syslog.LOG_INFO, message)
 end
 
-M.PipePolicies = {
-	DISCARD=-1,
-	INHERIT=0,
-	CREATE=1
-}
+M.subprocess = {}
+
+M.subprocess.DEVNULL = -1
+M.subprocess.INHERIT = nil
+M.subprocess.PIPE = 1
 
 -- Execute a program found using command PATH search, like the shell.
 -- Return the pid, as well as the I/O streams as pipes or nil on error.
-function M.popen3(policies, path, ...)
+function M.subprocess.popen(policies, path, ...)
 	local childfds = {}
 	local parentfds = {}
 
 	for fd, policy in pairs(policies) do
-		if policy==M.PipePolicies.CREATE then
+		if policy==M.subprocess.PIPE then
 			local piper, pipew = posix_unistd.pipe()
 			if fd==posix_unistd.STDIN_FILENO then
 				childfds[fd]=piper
@@ -224,17 +224,17 @@ function M.popen3(policies, path, ...)
 		return nil, errmsg, errnum
 	elseif pid == 0 then
 		local null=-1
-		if M.contains(policies, M.PipePolicies.DISCARD) then
+		if M.contains(policies, M.subprocess.DEVNULL) then
 			-- only open, if there's anything to discard
 			null = posix_fcntl.open('/dev/null', posix_fcntl.O_WRONLY)
 		end
 
 		for fd, policy in pairs(policies) do
-			if policy==M.PipePolicies.DISCARD then
+			if policy==M.subprocess.DEVNULL then
 				if fd~=posix_unistd.STDIN_FILENO then
 					posix_unistd.dup2(null, fd)
 				end
-			elseif policy==M.PipePolicies.CREATE then
+			elseif policy==M.subprocess.PIPE then
 				-- only close these, if they exist
 				posix_unistd.close(parentfds[fd])
 				posix_unistd.dup2(childfds[fd], fd)
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
@@ -224,7 +224,7 @@ function M.popen3(policies, path, ...)
 	if pid == nil then
 		return nil
 	elseif pid == 0 then
-		local null=-1;
+		local null=-1
 		if M.contains(policies, M.PipePolicies.DISCARD) then
 			-- only open, if there's anything to discard
 			null = posix_fcntl.open('/dev/null', posix_fcntl.O_WRONLY)
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
@@ -203,9 +203,9 @@ function M.popen3(policies, path, ...)
 	local parentfds = {}
 
 	for fd, policy in pairs(policies) do
-		if M.PipePolicies.CREATE==policy then
+		if policy==M.PipePolicies.CREATE then
 			local piper, pipew = posix_unistd.pipe()
-			if posix_unistd.STDIN_FILENO==fd then
+			if fd==posix_unistd.STDIN_FILENO then
 				childfds[fd]=piper
 				parentfds[fd]=pipew
 			else
@@ -230,11 +230,11 @@ function M.popen3(policies, path, ...)
 		end
 
 		for fd, policy in pairs(policies) do
-			if M.PipePolicies.DISCARD==policy then
-				if posix_unistd.STDIN_FILENO~=fd then
+			if policy==M.PipePolicies.DISCARD then
+				if fd~=posix_unistd.STDIN_FILENO then
 					posix_unistd.dup2(null, fd)
 				end
-			elseif M.PipePolicies.CREATE==policy then
+			elseif policy==M.PipePolicies.CREATE then
 				-- only close these, if they exist
 				posix_unistd.close(parentfds[fd])
 				posix_unistd.dup2(childfds[fd], fd)
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
@@ -190,6 +190,13 @@ function M.log(message, verbose)
 	posix_syslog.syslog(posix_syslog.LOG_INFO, message)
 end
 
+-- Blindly close the file descriptors in a given table.
+local function close_fds(descriptors)
+	for _, fd in pairs(descriptors) do
+		posix_unistd.close(fd)
+	end
+end
+
 M.subprocess = {}
 
 M.subprocess.DEVNULL = -1
@@ -221,6 +228,8 @@ function M.subprocess.popen(path, argt, options)
 	local pid, errmsg, errnum = posix_unistd.fork()
 
 	if pid == nil then
+		close_fds(childfds)
+		close_fds(parentfds)
 		return nil, errmsg, errnum
 	elseif pid == 0 then
 		local null = -1
@@ -251,9 +260,7 @@ function M.subprocess.popen(path, argt, options)
 		posix_unistd._exit(127)
 	end
 
-	for _, v in pairs(childfds) do
-		posix_unistd.close(v)
-	end
+	close_fds(childfds)
 
 	return pid, parentfds
 end
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
@@ -224,7 +224,7 @@ function M.subprocess.popen(path, argt, options)
 		return nil, errmsg, errnum
 	elseif pid == 0 then
 		local null = -1
-		local stdiostreams = {["stdin"] = 0, ["stdout"] = 1, ["stderr"] = 2}
+		local stdiostreams = {stdin = 0, stdout = 1, stderr = 2}
 		if M.contains(options, M.subprocess.DEVNULL) then
 			-- only open, if there's anything to discard
 			null = posix_fcntl.open('/dev/null', posix_fcntl.O_RDWR)
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
@@ -200,9 +200,10 @@ M.subprocess.PIPE = 1
 function M.subprocess.popen(path, argt, options)
 	local childfds = {}
 	local parentfds = {}
+	local stdiostreams = {stdin = 0, stdout = 1, stderr = 2}
 
-	for iostream, option in pairs(options) do
-		if option == M.subprocess.PIPE then
+	for iostream in pairs(stdiostreams) do
+		if options[iostream] == M.subprocess.PIPE then
 			local piper, pipew = posix_unistd.pipe()
 			if iostream == "stdin" then
 				childfds[iostream] = piper
@@ -223,13 +224,13 @@ function M.subprocess.popen(path, argt, options)
 		return nil, errmsg, errnum
 	elseif pid == 0 then
 		local null = -1
-		local stdiostreams = {stdin = 0, stdout = 1, stderr = 2}
 		if M.contains(options, M.subprocess.DEVNULL) then
 			-- only open, if there's anything to discard
 			null = posix_fcntl.open('/dev/null', posix_fcntl.O_RDWR)
 		end
 
-		for iostream, option in pairs(options) do
+		for iostream in pairs(stdiostreams) do
+			local option = options[iostream]
 			if option == M.subprocess.DEVNULL then
 				posix_unistd.dup2(null, stdiostreams[iostream])
 			elseif option == M.subprocess.PIPE then
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
@@ -241,12 +241,12 @@ function M.subprocess.popen(path, argt, options)
 			null = posix_fcntl.open('/dev/null', posix_fcntl.O_RDWR)
 		end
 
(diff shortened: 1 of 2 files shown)


NeoRaider:
It doesn't make much sense to open PRs just for helpers, especially when it's just a single commit. We need to see the code that uses the helpers to understand whether the helper makes sense like this.

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -188,4 +189,34 @@ function M.log(message, verbose)
...
+		posix_unistd.close(w0)
+		posix_unistd.close(r1)
+		posix_unistd.close(r2)
+
+		posix_unistd.dup2(r0, posix_unistd.STDIN_FILENO)
+		posix_unistd.dup2(w1, posix_unistd.STDOUT_FILENO)
+		posix_unistd.dup2(w2, posix_unistd.STDERR_FILENO)
+
+		posix_unistd.execp(path, ...)
+		return nil
```
NeoRaider, line 213: You are in a forked process here. When the exec fails, the child process must be terminated without running any cleanup (as that would confuse the parent process). `unistd._exit(127)` takes care of this (with 127 being a common exit code that appears in other similar code)

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -188,4 +189,34 @@ function M.log(message, verbose)
...
+	if pid < 0 then
+		return nil
+	elseif pid == 0 then
+		posix_unistd.close(w0)
+		posix_unistd.close(r1)
+		posix_unistd.close(r2)
+
+		posix_unistd.dup2(r0, posix_unistd.STDIN_FILENO)
+		posix_unistd.dup2(w1, posix_unistd.STDOUT_FILENO)
+		posix_unistd.dup2(w2, posix_unistd.STDERR_FILENO)
```
NeoRaider, line 210: The child process shouldn't get any file descriptors except for 0, 1 and 2. While there is no nice way to ensure this for *all* files that might be opened by the parent process except for looping over all possible FDs, at least `r0`, `w1` and `w2` should be closed after dup2ing them.

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
AiyionPrime: reoslved in 246ed9760184da60ff3b3908afbcce2532cb4578

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
AiyionPrime: resolved in 246ed9760184da60ff3b3908afbcce2532cb4578

blocktrron:
@AiyionPrime I second Neoraider here. Can you incorporate this helper to the PR requiring this feature? Otherwise we just add dead code here.

AiyionPrime:
Though I'll mostly intend to use this in #2245, I'd refactor the remote.lua, if no one objects?

NeoRaider:
> Though I'll mostly intend to use this in #2245, I'd refactor the remote.lua, if no one objects?
> The two PR are not blocking each other; I'll refactor whatever gets merged last.

For `remote.lua` we'd need a more configurable popen2/3 that allows to select `/dev/null` or pipe FD for each of the standard streams (or at least stdin and stdout). For stderr `/dev/null` or an inherited FD often make more sense than a pipe, so maybe all 3 streams should support all 3 options?

AiyionPrime:
like, popen takes three arguments, which each might be false, true or a pipe? (In case typeof is a thing in our lua).
False to discard them as fast as possible, true to return them as newly spawned and a pipe as "inherited"?

@NeoRaider  If you mean something thing else with _inherited_, I need a bigger hint.

NeoRaider:
> like, popen takes three arguments, which each might be false, true or a pipe? (In case typeof is a thing in our lua).
> False to discard them as fast as possible, true to return them as newly spawned and a pipe as "inherited"?
> 
> @NeoRaider If you mean something thing else with _inherited_, I need a bigger hint.

Inherit means not to touch the fd of the parent process at all.

AiyionPrime:
> so maybe all 3 streams should support all 3 options

I like the idea, though the resulting function might become quite a blob, as there'd be three loops with conditions in them, one before the fork that initiates the pipes, and then one in each branch of the fork.

I'd like to start with how the functions signature will look like.
Maybe something like this:

```lua
PipePolicies = {
        DISCARD=-1,
        INHERIT=0,
        CREATE=1
}

function M.popen3(policies, path, ...)
    [...]
return pid, pipes
```

A call could look like this:
```lua
local myorder = {posix_unistd.STDIN_FILENO=PipePolicies.CREATE,
                 posix_unistd.STDOUT_FILENO=PipePolicies.INHERIT,
                 posix_unistd.STDERR_FILENO=PipePolicies.DISCARD}
local mypid, mypipes = popen3(myorder, 'nc', {'nc', '::1', '33123'})
```

I think it would be intuitive to build mypipes the same way mywishes is constructed.
In this example one would access the returned `stdin`-pipe like this:
```lua
local stdin = mypipes[posix_unistd.STDIN_FILENO]
```

AiyionPrime:
The current implementation reflects all of @NeoRaider s enhancements.
It compiles, runs on http://lilienstra-e.n.ffh.zone/ and I just changed the password of the router using the new function.

Please have a thorough look at this though. Thanks for more feedback!

AiyionPrime:
I searched for other forks but only found some where even the current implementation might not be sufficient yet.

This one for example would need an optional parameter `delay`.
https://github.com/freifunk-gluon/gluon/blob/069c8b2035c53b2235c716ec7c640e892351f5b2/package/gluon-web-admin/luasrc/lib/gluon/config-mode/controller/admin/upgrade.lua#L38-L64

AiyionPrime:
Whatever, I'm fine with this, as it is. Delay should be added later, if someone wants it.

NeoRaider:
> Whatever, I'm fine with this, as it is. Delay should be added later, if someone wants it.

Both the chdir and the sleep as too specialized to implement it as part of a popen helper. While slightly less efficient, just running `sh -c 'cd /; sleep 1; exec ...'` as the command passed to popen should work fine in this case.

AiyionPrime:
What's left for me to do here?

NeoRaider:
I think it would be nice to use the naming from the Python `subprocess` module:
- Call the function `util.subprocess.popen`
- Use `util.subprocess.PIPE`, `util.subprocess.DEVNULL` and `nil` instead of `CREATE`, `DISCARD` and `INHERIT`

I'm having another look at the code now.

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -188,4 +190,73 @@ function M.log(message, verbose)
...
 
+M.PipePolicies = {
+	DISCARD=-1,
+	INHERIT=0,
+	CREATE=1
+}
+
+-- Execute a program found using command PATH search, like the shell.
+-- Return the pid, as well as the I/O streams as pipes or nil on error.
+function M.popen3(policies, path, ...)
```
NeoRaider, line 201: Having the policies first feels a bit weird, especially as a user might want to the policies (if all FDs are to be inherited). I suggest making the command a table instead, so the signature would become
```lua
function popen(cmd, fds)
```

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -188,4 +190,73 @@ function M.log(message, verbose)
...
+	local pipes = {}
+	local intern = {}
+	local extern = {}
+
+	for fd, policy in pairs(policies) do
+		if M.PipePolicies.CREATE==policy then
```
NeoRaider, line 207: Please don't use Yoda conditions, they are generally confusing, and in Lua they don't even prevent mistakes, as there aren't constants.

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -188,4 +190,73 @@ function M.log(message, verbose)
...
+			pipes[fd]={posix_unistd.pipe()}
```
NeoRaider, line 208: `pipes` is not used after this loop, and you can avoid the use of a temporary table for the return values:
```lua
local piper, pipew = posix_unistd.pipe()
```
The variable `pipes` can be removed completely.

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -188,4 +190,73 @@ function M.log(message, verbose)
...
+
+	local pid = posix_unistd.fork()
+
+	if pid < 0 then
+		return nil
+	elseif pid == 0 then
+		local null=-1;
+		if M.contains(policies, M.PipePolicies.DISCARD) then
+			-- only open, if there's anything to discard
+			null = posix_fcntl.open('/dev/null', posix_fcntl.O_WRONLY)
```
NeoRaider, line 230: stdin should be supported as well; `O_WRONLY` will be incorrect in this case.

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -188,4 +190,73 @@ function M.log(message, verbose)
...
+	INHERIT=0,
+	CREATE=1
+}
+
+-- Execute a program found using command PATH search, like the shell.
+-- Return the pid, as well as the I/O streams as pipes or nil on error.
+function M.popen3(policies, path, ...)
+	local pipes = {}
+	local intern = {}
+	local extern = {}
```
NeoRaider, line 204: How about calling these `parentfds` and `childfds`?

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -188,4 +190,73 @@ function M.log(message, verbose)
...
+
+	-- intern: r0, w1, w2
+	-- extern: w0, r1, r2
+
+	local pid = posix_unistd.fork()
+
+	if pid < 0 then
+		return nil
+	elseif pid == 0 then
+		local null=-1;
```
NeoRaider, line 227: Unnecessary semicolon

package/gluon-web-admin/luasrc/lib/gluon/config-mode/model/admin/remote.lua
```diff
@@ -76,29 +75,12 @@ function pw2.cfgvalue()
...
-
-		unistd.dup2(inr, unistd.STDIN_FILENO)
-		unistd.close(inr)
-
-		unistd.execp('passwd', {[0] = 'passwd'})
-		os.exit(127)
-	end
+	local policies = {[unistd.STDIN_FILENO]=util.PipePolicies.CREATE,
+			  [unistd.STDOUT_FILENO]=util.PipePolicies.DISCARD,
+			  [unistd.STDERR_FILENO]=util.PipePolicies.DISCARD}
```
NeoRaider, line 80: `[unistd.STDOUT_FILENO]=util.PipePolicies.DISCARD` etc. is fairly verbose. How about changing to string keys `stdout = ...`?

Together with the changes suggested about, it would even be possible to merge the `command` and `policies`/`fds` arguments and make the function even more similar to Python's `subprocess`:

```lua
local pid, pipes = util.subprocess.popen {'passwd', stdin=util.subprocess.PIPE, stdout=util.subprocess.DEVNULL, stderr=util.subprocess.DEVNULL}
local inw = pipes.stdin
```

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -188,4 +190,73 @@ function M.log(message, verbose)
...
+			end
+		end
+	end
+
+	-- intern: r0, w1, w2
+	-- extern: w0, r1, r2
+
+	local pid = posix_unistd.fork()
+
+	if pid < 0 then
```
NeoRaider, line 224: `fork()` returns `nil` on error according to the docs, this was wrong in the old code.

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -188,4 +190,73 @@ function M.log(message, verbose)
...
+		return nil
```
NeoRaider, line 225: Also return the errmsg, errnum on errors.

NeoRaider (changes requested):
Thanks for working on this, this is looking fairly good. I mostly have suggestions regarding the naming of things and the API.

AiyionPrime:
> I think it would be nice to use the naming from the Python `subprocess` module:
> 
>     * Call the function `util.subprocess.popen`
> 
>     * Use `util.subprocess.PIPE`, `util.subprocess.DEVNULL` and `nil` instead of `CREATE`, `DISCARD` and `INHERIT`

The way I implemented policies
allows the filedescriptors to be anything, not limited to 0, 1 and 2.

Using `nil` instead of discard does break quite a bit of my intended logic including iterating policies on a key value basis.

> [unistd.STDOUT_FILENO]=util.PipePolicies.DISCARD etc. is fairly verbose. How about changing to string keys stdout = ...?

This would allow to at least check if anything is nil, as the keys are then known;
but which filedescriptor each of the I/O-streams has could than not be taken from policies anymore.

@NeoRaider Is alternative filedescriptors just a feature we don't want the popen function to have?

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
AiyionPrime: Where `cmd` contains both `path` and the command table (`argt`)?

This change would drop the similarity to `execp`s signature.
https://luaposix.github.io/luaposix/modules/posix.unistd.html#execp

NeoRaider:
> > [...]
> > * Use `util.subprocess.PIPE`, `util.subprocess.DEVNULL` and `nil` instead of `CREATE`, `DISCARD` and `INHERIT`
> 
> The way I implemented policies
> allows the filedescriptors to be anything, not limited to 0, 1 and 2.
> 

`nil` would correspond to `INHERIT`, which doesn't need any handling. But instead of iterating, it would also be possible to pull out the logic into helper functions - the iteration is already somewhat awkward because stdin needs different handling than stdout and stderr.

> 
> > [unistd.STDOUT_FILENO]=util.PipePolicies.DISCARD etc. is fairly verbose. How about changing to string keys stdout = ...?
> 
> This would allow to at least check if anything is nil, as the keys are then known;
> but which filedescriptor each of the I/O-streams has could than not be taken from policies anymore.
> 
> @NeoRaider Is alternative filedescriptors just a feature we don't want the popen function to have?

To support pipes for additional FDs, you would also need a way to specify the direction, further complicating the implementation. I don't think it's worth the trouble.

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
NeoRaider: I like `execp`'s signature, let's do it like that.

AiyionPrime:
Alright, then I'm good to go, I don't think I'll finish this today;
I'll be done resolving your remarks before noon tomorrow, though.
Thanks for the input and have a nice evening!

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
AiyionPrime: Not just there. https://github.com/luaposix/luaposix/issues/340

AiyionPrime: resolved in daa95f1

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
AiyionPrime: resolved in fd6c77d

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
AiyionPrime: resolved in 12d2f3a

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
AiyionPrime: resolved in ba2886e

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
AiyionPrime: The problem is then the trailing vararg; which leaves us with the options of
`path, policies, ...`
and the current
`policies, path, ...`.

Which do you favor? Or did you intend something else with 
> let's do it like that

?

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
AiyionPrime: resolved in e2ec7f6

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
AiyionPrime: resolved in c1df156

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
AiyionPrime: Reading `/dev/null` to feed the result to the process?

AiyionPrime:
> I think it would be nice to use the naming from the Python `subprocess` module:
> 
>     * Call the function `util.subprocess.popen`
> 
>     * Use `util.subprocess.PIPE`, `util.subprocess.DEVNULL` and `nil` instead of `CREATE`, `DISCARD` and `INHERIT`

resolved in 0b3f0e9b4ee0cedbbe3282f881c515bd32a28215 and 50ef57f7f51249f159120f5cb625d71c1886fff8

Let me know if you intended to drop the line `M.subprocess.INHERIT = nil`

package/gluon-web-admin/luasrc/lib/gluon/config-mode/model/admin/remote.lua
AiyionPrime: You mean like this?

```suggestion
	local policies = {["stdin"]=util.subprocess.PIPE,
			  ["stdout"]=util.subprocess.DEVNULL,
			  ["stderr"]=util.subprocess.DEVNULL}
```

AiyionPrime: Well that didn't handle the last commits well. Ill add the proposal below.

package/gluon-web-admin/luasrc/lib/gluon/config-mode/model/admin/remote.lua
```diff
@@ -76,29 +75,12 @@ function pw2.cfgvalue()
...
-
-		unistd.dup2(inr, unistd.STDIN_FILENO)
-		unistd.close(inr)
-
-		unistd.execp('passwd', {[0] = 'passwd'})
-		os.exit(127)
-	end
+	local policies = {[unistd.STDIN_FILENO]=util.subprocess.PIPE,
+			  [unistd.STDOUT_FILENO]=util.subprocess.DEVNULL,
+			  [unistd.STDERR_FILENO]=util.subprocess.DEVNULL}
```
AiyionPrime, line 80: @NeoRaider this would be less verbose for the calling function.

```suggestion
	local policies = {["stdin"]=util.subprocess.PIPE,
			  ["stdout"]=util.subprocess.DEVNULL,
			  ["stderr"]=util.subprocess.DEVNULL}
```

AiyionPrime: I don't think its much cleaner; but creates the need for some kind of lookup in `popen()` to translate the stream names to file descriptors. Your choice.

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -188,4 +190,72 @@ function M.log(message, verbose)
...
+M.subprocess.PIPE = 1
+
+-- Execute a program found using command PATH search, like the shell.
+-- Return the pid, as well as the I/O streams as pipes or nil on error.
+function M.subprocess.popen(policies, path, ...)
+	local childfds = {}
+	local parentfds = {}
+
+	for fd, policy in pairs(policies) do
+		if policy==M.subprocess.PIPE then
```
2tata, line 206: some inconsistency in spacing.

2tata (changes requested):
all in all looks good so far.

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
AiyionPrime: Indeed, thanks.
I think i got them all in 2502b2c, right?

package/gluon-web-admin/luasrc/lib/gluon/config-mode/model/admin/remote.lua
2tata, line 80: here are some spacing issues left.

2tata (changes requested):

package/gluon-web-admin/luasrc/lib/gluon/config-mode/model/admin/remote.lua
AiyionPrime: Will fix them, when @neoraider has decided, what the keys are going to be.
https://github.com/freifunk-gluon/gluon/pull/2244#discussion_r664472959

AiyionPrime: Sorry @2tata, thought this would happen faster. No need for you to wait as well ^^
resolved in 939720d

AiyionPrime: I just repasted the draft on how this change could reflect in the function: https://bpa.st/XHUQ (now with proper spacing)

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
AiyionPrime: resolved in 63d44093162078b04492e0645179f667dfd86159

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
NeoRaider: I don't think the function should be vararg, `execp` isn't vararg either. `popen(path, argt, options)` would make sense to me.

package/gluon-web-admin/luasrc/lib/gluon/config-mode/model/admin/remote.lua
NeoRaider: I'd expect the user to assign the `subprocess` module to a local variable:
```lua
sp.popen('ls', {'-l', '/'}, {
  stdin = sp.PIPE,
  stdout = sp.DEVNULL,
  stderr = sp.DEVNULL,
})
```

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
AiyionPrime: resolved in a66e8b815dc1ba40e5e99efe07d7e2bb264001d2 and e9e42fa261a9089bf67a1d085517189a62eebc3e

package/gluon-web-admin/luasrc/lib/gluon/config-mode/model/admin/remote.lua
AiyionPrime: resolved in 436b06d91094c3dbe49b8c40ec2c88e881cd87aa and e07c2abba3c51aef9af7e91055e04ff4516e9fd3

package/gluon-web-admin/luasrc/lib/gluon/config-mode/model/admin/remote.lua
AiyionPrime: https://github.com/freifunk-gluon/gluon/pull/2244#discussion_r667534113

AiyionPrime:
Compiles and luacheck is fine with it as well.
I'll test this tomorrow morning.

@NeoRaider, @2tata, @blocktrron I think this ready to review one more time.
Thanks for your patience!

AiyionPrime:
I deployed this on lilienstraße.n.ffh.zone earlier this morning.
It's reachable via [http://[2a02:790:ff:1014:b64b:d6ff:fe20:2778]/cgi-bin/status](http://[2a02:790:ff:1014:b64b:d6ff:fe20:2778]/cgi-bin/status) due to an unrelated ongoing rdns migration.

I successfully changed the password via configmode (which uses the modified version of remote.lua)

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -188,4 +190,71 @@ function M.log(message, verbose)
 	posix_syslog.syslog(posix_syslog.LOG_INFO, message)
 end
 
+M.subprocess = {}
+
+M.subprocess.DEVNULL = -1
+M.subprocess.INHERIT = nil
```
NeoRaider, line 196: Assigning `nil` to an unset table key is meaningless. Let's just scrap the name "INHERIT" and just always use `nil`.

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -188,4 +190,71 @@ function M.log(message, verbose)
...
+	-- childfds: r0, w1, w2
+	-- parentfds: w0, r1, r2
+
+	local pid, errmsg, errnum = posix_unistd.fork()
+
+	if pid == nil then
+		return nil, errmsg, errnum
+	elseif pid == 0 then
+		local null = -1
+		local stdiostreams = {["stdin"] = 0, ["stdout"] = 1, ["stderr"] = 2}
```
NeoRaider, line 227:
```suggestion
		local stdiostreams = {stdin = 0, stdout = 1, stderr = 2}
```

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -188,4 +190,71 @@ function M.log(message, verbose)
...
+M.subprocess.INHERIT = nil
+M.subprocess.PIPE = 1
+
+-- Execute a program found using command PATH search, like the shell.
+-- Return the pid, as well as the I/O streams as pipes or nil on error.
+function M.subprocess.popen(path, argt, options)
+	local childfds = {}
+	local parentfds = {}
+
+	for iostream, option in pairs(options) do
```
NeoRaider, line 205: I think we should iterate over {'stdin', 'stdout', 'stderr'} explicitly here - `options` may be used for other options later.

package/gluon-web-admin/luasrc/lib/gluon/config-mode/model/admin/remote.lua
```diff
@@ -76,29 +76,12 @@ function pw2.cfgvalue()
...
-
-		unistd.execp('passwd', {[0] = 'passwd'})
-		os.exit(127)
-	end
+	local options = {stdin = sp.PIPE,
+			 stdout = sp.DEVNULL,
+			 stderr = sp.DEVNULL}
 
-	unistd.close(inr)
+	local pid, pipe = sp.popen('passwd', {[0] = 'passwd'}, options)
```
NeoRaider, line 83: Is element 0 necessary when it is the same as the command?

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -188,4 +190,71 @@ function M.log(message, verbose)
...
+		end
+	end
+
+	-- childfds: r0, w1, w2
+	-- parentfds: w0, r1, r2
+
+	local pid, errmsg, errnum = posix_unistd.fork()
+
+	if pid == nil then
+		return nil, errmsg, errnum
```
NeoRaider, line 224: We need to close all childfds and parentfds on errors.

How about introducing a helper to close a list of FDs? This can then also be used to close all parentfds in the child and all childfds in the parent in the successful case.

package/gluon-web-admin/luasrc/lib/gluon/config-mode/model/admin/remote.lua
AiyionPrime: I thought so, but it is not:

https://github.com/luaposix/luaposix/blob/master/ext/posix/unistd.c#L302-L307

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
AiyionPrime: resolved in b5655410e5fb463795b42c8e6372ff840cad7406

package/gluon-web-admin/luasrc/lib/gluon/config-mode/model/admin/remote.lua
AiyionPrime: resolved in 47f835e254f2046fe2e57c0dfff21a7737a3f011

AiyionPrime:
> I deployed this on lilienstra-e.n.ffh.zone earlier this morning.
> It's reachable via [http://[2a02:790:ff:1014:b64b:d6ff:fe20:2778]/cgi-bin/status](http://%5B2a02:790:ff:1014:b64b:d6ff:fe20:2778%5D/cgi-bin/status) due to an unrelated ongoing rdns migration.
> 
> I successfully changed the password via configmode (which uses the modified version of remote.lua)

Same today.

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -188,4 +190,78 @@ function M.log(message, verbose)
...
+			if option == M.subprocess.DEVNULL then
+				posix_unistd.dup2(null, stdiostreams[iostream])
+			elseif option == M.subprocess.PIPE then
+				-- only close these, if they exist
+				posix_unistd.close(parentfds[iostream])
+				posix_unistd.dup2(childfds[iostream], stdiostreams[iostream])
+				-- close all the dups
+				posix_unistd.close(childfds[iostream])
+			end
+		end
```
NeoRaider, line 252: I think the following is be a bit easier to understand:
```suggestion
		for iostream, fd in pairs(stdiostreams) do
			local option = options[iostream]
			if option == M.subprocess.DEVNULL then
				posix_unistd.dup2(null, fd)
			elseif option == M.subprocess.PIPE then
				posix_unistd.dup2(childfds[iostream], fd)
			end
		end
		close_fds(childfds)
		close_fds(parentfds)
```

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -188,4 +190,78 @@ function M.log(message, verbose)
...
+end
+
+M.subprocess = {}
+
+M.subprocess.DEVNULL = -1
+M.subprocess.PIPE = 1
+
+-- Execute a program found using command PATH search, like the shell.
+-- Return the pid, as well as the I/O streams as pipes or nil on error.
+function M.subprocess.popen(path, argt, options)
```
NeoRaider, line 207: How about adding `argt = argt or {}` somewhere? This would allow to call the function with `argt == nil` - in particular, the `argt` and `options` arguments can be omitted completely when they are not needed.

package/gluon-web-admin/luasrc/lib/gluon/config-mode/model/admin/remote.lua
```diff
@@ -76,29 +76,12 @@ function pw2.cfgvalue()
...
-		unistd.execp('passwd', {[0] = 'passwd'})
-		os.exit(127)
-	end
+	local options = {stdin = sp.PIPE,
+			 stdout = sp.DEVNULL,
+			 stderr = sp.DEVNULL}
 
-	unistd.close(inr)
+	local pid, pipe = sp.popen('passwd', {}, options)
+	local inw = pipe["stdin"]
```
NeoRaider, line 84:
```suggestion
	local inw = pipe.stdin
```

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
AiyionPrime: I think the closing should stay in the PIPE-branch, as it's not necessary in the others, right?
Everything else looks good.

AiyionPrime: Well, as long as it does no harm, I'm fine with both.

package/gluon-web-admin/luasrc/lib/gluon/config-mode/model/admin/remote.lua
AiyionPrime: resolved in 6aa7d85c12b5452cd995407f20f1e984fd6c4b46

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
AiyionPrime: resolved in d513d001f4ea9614a8dd36bec4f572148583ae71

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
AiyionPrime: resolved in 1acc783 and 98efb82

AiyionPrime:
No need to review right now, the last four commits broke something.

AiyionPrime:
I do not see the error yet, but I just bisected it to be introduced in 1acc783a0990bcb9286c61c79b6b4757a5c5b62f.

Trying to change the password after it results in
> Unable to change the password.

That means `set_password(pw)` does not return true anymore.

After further debugging; it appears, calling closing file descriptors unconditionally is a bad idea and results in an error.
> Bad file descriptor

I'd go for not using the helper in the PIPE-branch like mentioned earlier in https://github.com/freifunk-gluon/gluon/pull/2244#discussion_r670172466.

The other option would be to make the closing helper more error resistant and ignore them, when they happen.
I don't think the cleaner call in the main function would justify such means.

AiyionPrime:
> > I deployed this on lilienstra-e.n.ffh.zone earlier this morning.
> > It's reachable via [http://[2a02:790:ff:1014:b64b:d6ff:fe20:2778]/cgi-bin/status](http://%5B2a02:790:ff:1014:b64b:d6ff:fe20:2778%5D/cgi-bin/status) due to an unrelated ongoing rdns migration.
> > I successfully changed the password via configmode (which uses the modified version of remote.lua)
> 
> Same today.

Works again.

NeoRaider:
Hmm, no idea what issue you were seeing, it's working fine here with my proposed change. I'll make a few more tests.

NeoRaider:
Thanks, merged!

AiyionPrime:
Sure, no problem; but with out the last commit I was not able to change passwords using the remote.lua (in advanced configmode).

NeoRaider:
> Sure, no problem; but without the last commit I was not able to change passwords using the remote.lua (in advanced configmode).

You accidentally closed the FDs in the loop instead of after the loop.

AiyionPrime:
> You accidentally closed the FDs in the loop instead of after the loop.

I don't think that's true. Have a look at the last three changes on the page
https://github.com/freifunk-gluon/gluon/pull/2244/commits/1acc783a0990bcb9286c61c79b6b4757a5c5b62f
close fds is outside of the loop.

I'll test, whether the problem does not exist on master, as soon as my build finishes.

NeoRaider:
> > You accidentally closed the FDs in the loop instead of after the loop.
> 
> I don't think that's true. Have a look at the last three changes on the page
> [1acc783](https://github.com/freifunk-gluon/gluon/commit/1acc783a0990bcb9286c61c79b6b4757a5c5b62f)
> close fds is outside of the loop.
> 
> I'll test, whether the problem does not exist on master, as soon as my build finishes.

I was looking at your last commit, and you were closing the FDs in the loop.

AiyionPrime:
Indeed, my bad. Not sure, what i saw there earlier. Thanks.

PR #2245, opened 2021-06-19
gluon-state: implement ipv6 default route check

### LINKED ISSUE #2228: gluon-online-status concept
As we closed #1930 and #1684 today with a reference to an IRC discussion that was ongoing I want to present the conclusion of that discussion.

We think that having all nodes ping into the world in regular intervals is not something we would like to see in a first implementation of this feature. Instead we would like to focus on information that we can cheaply derive from the local node and offer a multitude of flags to various packages, that they can easily test for to see if the node is or isn't in a required state.

For that we define a directory `/var/gluon/online/` that carries empty marker files. With our initial proposal we think of two simple markers that we would like to see in the first version:

- `neighbors` or `mesh` to reflect that the node has neighbors that it meshes with
- `route_default4/6` or `default_gw4/6`, to reflect that the respective network stack has a default route

In later versions these can be extended by a multitude of things, I could imagine exposing whether we have an active NTP sync for example. This would also be open to contributions of markers through community packages.

A script should run in a regular fashion, that calls the canonical checks for the mesh protocols provider and based on these touches certain files, should they not exist, or deletes them should they no longer be valid. I think it would be helpful not to touch already existing markers so that we can look at the mtime of these markers to know when they entered a certain state.

For batman-adv we think `batctl n` and `batctl gwl` are strong contenders, and for babeld something can be grepped from the `dump` command of its control socket. 

*N.B: I'm not sure what state the babel setup is in, and I wouldn't want to make it a mandatory part of the implementation if, as I have recently heard, it doesn't build for multiple releases now and nobody noticed.*

@NeoRaider @blocktrron @AiyionPrime @T-X I hope this summarises what we talked about, if not feel free to add your larification below.

AiyionPrime:
~This intends to eventually resolve 2228.~
This will no longer complete #2228.

It does it's job, but is still up for refactoring;
I just wanted to give a chance for early feedback and a forming discussion for this,
in case it's not what it's supposed to become.

- [x] code works
- [x] Makefile exists
- [x] micrond is taken care of

### DIFF
--- a/package/gluon-state-check/Makefile
+++ b/package/gluon-state-check/Makefile
@@ -1,0 +1,28 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-state-check
+PKG_VERSION:=1
+
+include ../gluon.mk
+
+define Package/gluon-state-check
+  TITLE:=Provides info about the routers state
+  DEPENDS:=+gluon-core +micrond
+endef
+
+define Package/gluon-state-check/description
+  gluon-state-check executes checks in `/lib/gluon/state/check.d/` and provides
+  a flag file for each check in `/var/gluon/state` depending on the return code
+  of the check. A flag file is created (or "touched") if the corresponding check
+  exits cleanly and gets removed otherwise. If the flags are "touched", they
+  are only accessed, but not modified. In this way, the atime of a flag file
+  reflects when the last check was performed and the mtime reflects when
+  when the state was last changed.
+
+  This package provides the following checks:
+  - `has_default_gw6` - check whether the router has a default IPv6-route on br-client.
+
+  The checks are executed once every minute (by micron.d).
+endef
+
+$(eval $(call BuildPackageGluon,gluon-state-check))
--- a/package/gluon-state-check/luasrc/usr/sbin/gluon-state-check
+++ b/package/gluon-state-check/luasrc/usr/sbin/gluon-state-check
@@ -1,0 +1,41 @@
+#!/usr/bin/lua
+
+local util = require 'gluon.util'
+local unistd = require 'posix.unistd'
+
+local state_dir = "/var/gluon/state/"
+local check_dir = "/lib/gluon/state/check.d/"
+
+
+local function set_flag(stateflag, state)
+	if state then
+		-- this does not modify atime
+		local flaghandle = io.open(stateflag, "w")
+		flaghandle:close()
+	else
+		os.remove(stateflag)
+	end
+end
+
+local function exec_check(checkpath)
+	local checkname = string.sub(checkpath, #check_dir+1)
+	local ret = os.execute(checkpath)
+	local flagfile = state_dir..checkname
+	set_flag(flagfile, 0==ret)
+end
+
+local function run_executable_checks()
+	for _, v in ipairs(util.glob(check_dir..'*')) do
+		if unistd.access(v, 'x') then
+			exec_check(v)
+		end
+	end
+end
+
+
+-- ensure state path exists
+if not unistd.access(state_dir) then
+	os.execute("mkdir -p "..state_dir)
+end
+
+run_executable_checks()
--- a/package/gluon-state-check/files/lib/gluon/state/check.d/has_default_gw6
+++ b/package/gluon-state-check/files/lib/gluon/state/check.d/has_default_gw6
@@ -1,0 +1,2 @@
+#!/bin/sh
+out=$(ip -6 route show default dev br-client 2>/dev/null) && [ -n "$out" ]
--- a/package/gluon-state-check/files/usr/lib/micron.d/gluon-state-check
+++ b/package/gluon-state-check/files/usr/lib/micron.d/gluon-state-check
@@ -1,0 +1,1 @@
+* * * * * /usr/sbin/gluon-state-check


package/gluon-state-online/luasrc/usr/sbin/gluon-state-online
```diff
@@ -0,0 +1,171 @@
...
+		return false
+	end
+	for line in handle:lines() do
+		if line:find("default") then
+			handle:close()
+			return true
+		end
+	end
+	handle:close()
+	return false
```
lemoer, line 127: I would suggest:

```suggestion
	return check_command({[0] = 'ip', '-6', 'route', 'show', 'default'}, 'default')
```

AiyionPrime:
I think the path `package/gluon-state-online/files/usr/sbin/gluon-state-online` should not contain `files`, but likely `luasrc`?

Other than that, this is ready for some review.

mweinelt:
> I think the path package/gluon-state-online/files/usr/sbin/gluon-state-online should not contain files, but likely luasrc?

The only difference between `files` and `luasrc is, that the latter minifies lua files. So if what you are deploying is lua, put it below `luasrc`.

AiyionPrime:
Done.

AiyionPrime:
let's make this smaller and **blocked** until #2244 is done.

AiyionPrime:
#2244 is closed, this is open for review.

package/gluon-state-online/luasrc/usr/sbin/gluon-state-online
```diff
@@ -0,0 +1,143 @@
...
+
+local function has_batman_neighbours()
+	local neighbourcmd = {[0] = 'batctl', 'n', '-H'}
+	return check_command(neighbourcmd, "")
+end
+
+local function has_babel_neighbours()
+	local socket_command = {[0] = 'nc', '::1', '33123'}
+	return check_command(socket_command, "add neighbour", "dump\n")
+end
```
NeoRaider, line 85: Let's leave out the protocol-specific parts for now, I'd like to move those to the `gluon-mesh...` packages.

AiyionPrime: resolved in 15d389a.

AiyionPrime:
As of 15d389a this PR wont resolve #2228 anymore.

> Let's leave out the protocol-specific parts

I think it would have been nicer to have heard this in early june, when hexa summarized what you discussed in #gluon and [asked for clarifications](https://github.com/freifunk-gluon/gluon/issues/2228#issue-914132522) on the matter.

AiyionPrime:
@NeoRaider you left a thumbsup at lemoer's proposed usage of check command.
Looking at what's left in this PR, let me know if you still want it.

NeoRaider:
> @NeoRaider you left a thumbsup at lemoer's proposed usage of check command.
> Looking at what's left in this PR, let me know if you still want it.

The `check_command` is not relevant, but the call to `ip` is: You definitely want `ip` only to print the right route instead of printing the whole table. This way, you only need to check whether there is output at all (and the whole thing is more efficient).

At this point I'm wondering if Lua is even the right choice for this, as the main part is calling another command, and the actual logic is trivial...

AiyionPrime:
> At this point I'm wondering if Lua is even the right choice for this

That's a little too vague for me, I'm afraid. You'd rather want this packet to be a shell script?

I really think it would have been and still would be be best, if you'd write down in #2228 what you want it's solution to look like.

AiyionPrime:
> You definitely want ip only to print the right route instead of printing the whole table.

resolved in 4b742dca23bb6d9d251233aa0905aafd45496647

lemoer:
I would suggest to stay with lua due to multiple reasons:
- The code has been written. It's discouraging for PR authors to drop the code only due to minor concerns.
- The essence of #2228 is to start simple. But later we want to add more things. This stuff can get more complex, of course. Switching to Shell now only to switch back to Lua later makes no sense.

package/gluon-state-online/luasrc/usr/sbin/gluon-state-online
```diff
@@ -0,0 +1,59 @@
...
+local unistd = require 'posix.unistd'
+local stat = require 'posix.sys.stat'
+
+local state_dir = "/var/gluon/state/"
+
+local route4_flag    = state_dir .. "default_route_4"
+local route6_flag    = state_dir .. "default_route_6"
+
+local function has_v6_default()
+	local handle = io.popen("ip -6 route show default")
```
lemoer, line 12:
```suggestion
local function has_default(ip_version, iface)
	local handle = io.popen("ip -" .. tostring(ip_version) .. " route show default table all dev " .. iface)
```

package/gluon-state-online/luasrc/usr/sbin/gluon-state-online
```diff
@@ -0,0 +1,59 @@
...
+		return false
+	end
+	for line in handle:lines() do
+		handle:close()
+		return true
+	end
+	handle:close()
+	return false
+end
+
```
lemoer, line 36:
```suggestion
```

package/gluon-state-online/luasrc/usr/sbin/gluon-state-online
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/lua
+
+local unistd = require 'posix.unistd'
+local stat = require 'posix.sys.stat'
+
+local state_dir = "/var/gluon/state/"
+
+local route4_flag    = state_dir .. "default_route_4"
+local route6_flag    = state_dir .. "default_route_6"
```
lemoer, line 9:
```suggestion
local wan_route4_flag    = state_dir .. "wan_default_route_4"
local wan_route6_flag    = state_dir .. "wan_default_route_6"
local mesh_route6_flag    = state_dir .. "mesh_default_route_6"
```

package/gluon-state-online/luasrc/usr/sbin/gluon-state-online
```diff
@@ -0,0 +1,59 @@
...
+end
+
+local v4_default = has_v4_default()
+local v6_default = has_v6_default()
+
+print("\thas_default_gw4: " .. tostring(v4_default))
+print("\thas_default_gw6: " .. tostring(v6_default))
+
+set_flag(route4_flag, v4_default)
+set_flag(route6_flag, v6_default)
```
lemoer, line 59:
```suggestion
local wan_v4_default = has_default(4, "br-wan")
local wan_v6_default = has_default(6, "br-wan")
local mesh_v6_default = has_default(6, "br-client")

print("\twan_has_default_gw4: " .. tostring(wan_v4_default))
print("\twan_has_default_gw6: " .. tostring(wan_v6_default))
print("\tmesh_has_default_gw6: " .. tostring(mesh_v6_default))

set_flag(wan_route4_flag, wan_v4_default)
set_flag(wan_route6_flag, wan_v6_default)
set_flag(mesh_route6_flag, mesh_v6_default)
```

lemoer:
I proposed some changes, because I think before the naming of the flags was misleading. The old flags had the following meaning:
- `default_route_4` - Check whether we have a **wan** default route for IPv4.
- `default_route_6` - Check whether we have a default route for IPv6 via the **mesh**.

I suspect it was not the intention to check one flag for wan and the other for mesh. The reason that this happens is that the routing tables for ipv6 and ipv4 are used differently.

So in my suggestions, I also added `wan_default_route_6` and changed the naming scheme to:
- `wan_default_route_4` - Check whether we have a wan default route for IPv4. *(before: `default_route_4`)*
- `wan_default_route_6` - Check whether we have a wan default route for IPv6. *(new)*
- `mesh_default_route_6` - Check whether we have a default route for IPv6 via the mesh. *(before:  `default_route_6`)*

NeoRaider:
> I have suggested some changes because I think the previous naming of the flags was misleading. The old flags had the following meaning:
> 
>     * `default_route_4` - Check whether we have a **wan** default route for IPv4.
> 
>     * `default_route_6` - Check whether we have a default route for IPv6 via the **mesh**.
> 
> 
> I suspect it was not the intention to check one flag for wan and the other for mesh. The reason that this happens is that the routing tables for ipv6 and ipv4 are used differently.
> 
> So in my suggestions, I also added `wan_default_route_6` and changed the naming scheme to:
> 
>     * `wan_default_route_4` - Check whether we have a wan default route for IPv4. _(before: `default_route_4`)_
> 
>     * `wan_default_route_6` - Check whether we have a wan default route for IPv6. _(new)_
> 
>     * `mesh_default_route_6` - Check whether we have a default route for IPv6 via the mesh. _(before:  `default_route_6`)_

I don't think WAN routes are in the scope of the proposal, I don't see a usecase at the moment. What we discussed was:

- `has_default_gw6`: Set flag when there is a route in the table
- `has_default_gw4: Set flag when there is a gateway in `batctl gwl`
- `has_neighbours`: Set based on `batctl o` or `batctl n`

Each of these can be determined using 1-2 lines of shell (+ some common code to manage flag files). Each flag can be handled by a separate script, so I don't see why adding more flags would be a reason to use Lua for these checks - flags that need more complex logic should be implemented in Lua though.

I'm thinking about the following:
- Have a directory of checker scripts that packages can install their checks into (`/lib/gluon/state/check.d`) - the package would also be called `gluon-state`
- `gluon-state` contains a Lua script and cron job to run each of the scripts regularly and create or delete the flag files based on their exit code
- `gluon-state` would also provide the `has_default_gw6` checker
- The `has_default_gw4` and `has_neighbours` scripts would be provided by `gluon-mesh-batman-adv` and `gluon-mesh-babel`

I think an advantage of this approach is that the different checkers can be provided by different packages, but they are only run if the `gluon-state` package is installed, which is not (easily) possible when the checker scripts come with their own cron jobs.

CodeFetch:
> * Have a directory of checker scripts that packages can install their checks into (`/lib/gluon/state/check.d`) - the package would also be called `gluon-state`

That was one of the ideas lemoer and I had at the beginning. As a note:
It would make sense to have numeric "runlevel" prefixes in that case so that the order of execution can be used to make dependent checks.

AiyionPrime:
@NeoRaider This should be roughly what you described.
Other packages can add executable files in `check.d`; which will then be called as well.
This leaves room to put non-executable lua-files in the same directory, that then can reflect more complex checks.

> [...]It would make sense to have numeric "runlevel" prefixes[...]

@CodeFetch I think that's a really good idea; I'd like to keep it in mind though;
as we currently have no checks depending on each other.
As the flagfiles (contrary to the checkfiles) should not contain the runlevels, adding that feature later does not inflict compatibility issues.

AiyionPrime:
Compiled and tested; works as expected.

http://[2a02:790:ff:714:b64b:d6ff:fe20:2778]/cgi-bin/status

You four should have ssh-access.

package/gluon-state/luasrc/usr/sbin/gluon-state
```diff
@@ -0,0 +1,45 @@
...
+			exec_check(v)
+		end
+	end
+end
+
+
+-- ensure state path exists
+if not unistd.access(state_dir) then
+	stat.mkdir(state_dir)
+end
```
lemoer, line 43:
```suggestion
-- ensure state path exists
if not unistd.access(state_dir) then
	mkdir_recursive(state_dir)
end
```

While you ensure that `/var/gluon/state` is created, you do not ensure that `/var/gluon/` is already created. `/var/gluon` is only created [here](https://github.com/freifunk-gluon/gluon/blob/d3d22ba677f3547c105c05fd5e66e80453e4b824/package/gluon-wan-dnsmasq/luasrc/lib/gluon/wan-dnsmasq/update.lua). Let's not create an implicit dependency here.

As we ignore the return values of *stat.mkdir()* anyways, checking for the existence with *unistd.access()* is unnecessary.

package/gluon-state/luasrc/usr/sbin/gluon-state
```diff
@@ -0,0 +1,45 @@
...
+local function run_executable_checks()
+	for _, v in ipairs(util.glob(check_dir..'*')) do
+		local statmode = stat.stat(v).st_mode
+		local is_file = statmode ~= nil and stat.S_ISREG(statmode)
+		if is_file and bit.band(statmode, stat.S_IXUSR) ~= 0 then
+			exec_check(v)
+		end
+	end
+end
+
```
lemoer, line 38:
```suggestion

local function mkdir_recursive(dir)
	if dir ~= '/' then
		mkdir_recursive(libgen.dirname(dir))
	end
	stat.mkdir(dir)  -- errors (especially existing directories) are ignored.
end

```

package/gluon-state/luasrc/usr/sbin/gluon-state
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/lua
+
+local bit = require 'bit'
+local util = require 'gluon.util'
+local unistd = require 'posix.unistd'
+local stat = require 'posix.sys.stat'
+
```
lemoer, line 7:
```suggestion
local libgen = require 'posix.libgen'

```

package/gluon-state/luasrc/usr/sbin/gluon-state
AiyionPrime: ooor we just `mkdir var/gluon` above that line :D

package/gluon-state/Makefile
```diff
@@ -0,0 +1,23 @@
...
+endef
+
+define Package/$(PKG_NAME)/description
+  This script runs every test file in `/lib/gluon/state/check.d` and provides the results as flags in `/var/gluon/state`.
+  It is called by micron.d once every minute.
+  A flag exists in case of its description being true and gets removed otherwise.
+  The flags get opened regardless them being changed.
+
+  This package provides the check `has_default_gw6`, that depends on
+  whether the router has a default IPv6-route on br-client.
```
lemoer, line 20:
```suggestion
  gluon-state executes checks in `/lib/gluon/state/check.d/` and provides
  a flag file for each check in `/var/gluon/state` depending on the return code
  of the check. A flag file is created (or "touched") if the corresponding check
  exits cleanly and gets removed otherwise. If the flags are "touched", they
  are only accessed, but not modified. This way, the atime of the flag files
  reflect when the last check was performed and the mtime reflects when
  when the state was last changed.

  This package provides the following checks:
  - `has_default_gw6` - check whether the router has a default IPv6-route on br-client.
  
  The checks are executed once every minute (by micron.d).
```

package/gluon-state/luasrc/usr/sbin/gluon-state
lemoer: Yep. This would also be possible. But then you would use the variable `state_dir` in one line and use the hardcoded value `/var/gluon` in the line above that line.

lemoer (commented):
I have noted some minor things. The rest looks good to me. :) Once they are addressed, this is an ACK from my side.

package/gluon-state/Makefile
AiyionPrime: Thanks, this sounds way better.

package/gluon-state/luasrc/usr/sbin/gluon-state
AiyionPrime: How do you feel about `os.execute("mkdir -p  "..state_dir)`, similar to this script?
https://github.com/freifunk-gluon/gluon/blob/14236ed8f6c77f93e75a7c25007ff1bd358f2ef9/scripts/copy_output.lua#L24

lemoer: That is also a valid option. As this call is encapsulated in the `unistd.access()` if clause, the additional fork+exec does not really matter in 99.99% of cases anyway. So let's go with your `lib.exec {'mkdir', '-p', dir}` suggestion. It adds the least lines of code.

AiyionPrime:
Compiled, still works.

package/gluon-state/luasrc/usr/sbin/gluon-state
AiyionPrime: resolved since 46b32d8a5a74562de174c907fcc53eac75de1d0e

package/gluon-state/luasrc/usr/sbin/gluon-state
AiyionPrime: outdated since 46b32d8a5a74562de174c907fcc53eac75de1d0e

package/gluon-state/luasrc/usr/sbin/gluon-state
AiyionPrime: outdated since 46b32d8a5a74562de174c907fcc53eac75de1d0e

lemoer (approved):
Thanks for the adjustments. ACK from my side.

package/gluon-state/files/lib/gluon/state/check.d/has_default_gw6
```diff
@@ -0,0 +1,2 @@
+#!/bin/ash
```
NeoRaider, line 1: Just use `/bin/sh`, OpenWrt's scripts never specify ash explicitly.

package/gluon-state/Makefile
```diff
@@ -0,0 +1,28 @@
...
+  reflects when the last check was performed and the mtime reflects when
+  when the state was last changed.
+
+  This package provides the following checks:
+  - `has_default_gw6` - check whether the router has a default IPv6-route on br-client.
+  
+  The checks are executed once every minute (by micron.d).
+endef
+
+$(eval $(call BuildPackageGluon,$(PKG_NAME)))
```
NeoRaider, line 28: The common style of other Gluon packages is not to use `$(PKG_NAME)`.

package/gluon-state/files/lib/gluon/state/check.d/has_default_gw6
```diff
@@ -0,0 +1,2 @@
...
+out=$(ip -6 route show default table all dev br-client 2>/dev/null) && [ -n "$out" ]
```
NeoRaider, line 2: Hmm, why the `table all`? The default table should be sufficient I think (and unless something has changed recently, the `table X` argument of busybox's `ip` implementation is extremely broken and shouldn't be used).

package/gluon-state/Makefile
```diff
@@ -0,0 +1,28 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-state
```
NeoRaider, line 3: Something more descriptive like `gluon-state-check` would be preferable.

NeoRaider (changes requested):

package/gluon-state/Makefile
CodeFetch: On first thought that repeats what it does unnecessarily. Either gluon-state or gluon-check.

package/gluon-state/files/lib/gluon/state/check.d/has_default_gw6
CodeFetch: It should be fixed if it is broken :smile:.

package/gluon-state/Makefile
CodeFetch: Why is that and if there is no strong reason why not change all to $(PKG_NAME) for convenience?

package/gluon-state/files/lib/gluon/state/check.d/has_default_gw6
AiyionPrime: Resolved in 270f1bb5e06377eabccc62aa6400dde3d537e9c4

package/gluon-state/Makefile
AiyionPrime: The latter question looks valid to me, but is out of this PRs scope. It can be taken care of over there (#2286).
Resolved in 8a813a7cf25cf9078209323b8a0b7d663cc56f1c in order to match the current codebase.

package/gluon-state/files/lib/gluon/state/check.d/has_default_gw6
AiyionPrime: Because @lemoer suggested it and I really favor explicit over implicit behaviour in programs.
It's less error prone and even while `ip` is not "extremely broken", using the explicit variant we would not have had to rely on it not changing its default output later.
Resolved in 270f1bb5e06377eabccc62aa6400dde3d537e9c4

package/gluon-state/Makefile
AiyionPrime: Like gluon-state-online? Just kidding...
Renamed once again in 8a813a7cf25cf9078209323b8a0b7d663cc56f1c

AiyionPrime:
I think this is done and ready to be squashed. Let me know if you think otherwise.
I'll compile it once again.

package/gluon-state-check/files/usr/lib/micron.d/gluon-state
```diff
@@ -0,0 +1 @@
+* * * * * /usr/sbin/gluon-state
```
lemoer, line 1: I would suggest to rename the binary to `gluon-state-check` as well.

package/gluon-state-check/Makefile
```diff
@@ -0,0 +1,28 @@
...
+PKG_VERSION:=1
+
+include ../gluon.mk
+
+define Package/$(PKG_NAME)
+  TITLE:=Provides info about the routers state
+  DEPENDS:=+gluon-core +micrond
+endef
+
+define Package/$(PKG_NAME)/description
```
lemoer, line 13: Still a $(PKG_NAME) left here.

package/gluon-state-check/Makefile
```diff
@@ -0,0 +1,28 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-state-check
+PKG_VERSION:=1
+
+include ../gluon.mk
+
+define Package/$(PKG_NAME)
```
lemoer, line 8: Another $(PKG_NAME) here.

package/gluon-state-check/files/usr/lib/micron.d/gluon-state
AiyionPrime: resolved in ce76c7b69d1d10ad39583ffb2dcfc5defb036ce4

package/gluon-state-check/Makefile
AiyionPrime: all done in ce76c7b69d1d10ad39583ffb2dcfc5defb036ce4

package/gluon-state-check/Makefile
AiyionPrime: even that. ce76c7b69d1d10ad39583ffb2dcfc5defb036ce4

package/gluon-state/Makefile
CodeFetch: @AiyionPrime 

> Like gluon-state-online? Just kidding...

That makes much more sense to me... There could also be gluon-state-meshlink or something.

CodeFetch: It's also much more descriptive... gluon-state-check - what state?

CodeFetch: I'd call it gluon-check-online

lemoer: I'd call it `gluon-state-check`. As it is now.

package/gluon-state-check/luasrc/usr/sbin/gluon-state-check
```diff
@@ -0,0 +1,45 @@
...
+	local ret = os.execute(checkpath)
+	local flagfile = state_dir..checkname
+	set_flag(flagfile, 0==ret)
+end
+
+local function run_executable_checks()
+	for _, v in ipairs(util.glob(check_dir..'*')) do
+		local statmode = stat.stat(v).st_mode
+		local is_file = statmode ~= nil and stat.S_ISREG(statmode)
+		if is_file and bit.band(statmode, stat.S_IXUSR) ~= 0 then
```
NeoRaider, line 33: Weird, I thought I had commented on this in my last review, but I can't find the comment now... here's the suggestion again:
```suggestion
		if unistd.access(v, 'x') then
```
(and the `bit` and `stat` requires can be removed then)

AiyionPrime: That's a really nice suggestion, I didn't know `access()` yet.
Though technically I checked for _file_ and _executable_; `access(v, 'x')` tests only for the latter, right?

I'd go with this, following the usage in the [lua-posix](http://luaposix.github.io/luaposix/modules/posix.unistd.html#access) section.
```suggestion
		if unistd.access(v, 'fx') then
```

AiyionPrime: implemented in fa523d6052224188dc6fd2967af1ef0f8283c98f

AiyionPrime:
Compiling and testing again.

AiyionPrime:
Compiles without errors and does its job: https://[2a02:790:ff:1014:b64b:d6ff:fe20:2778]

package/gluon-state-check/luasrc/usr/sbin/gluon-state-check
NeoRaider: The luaposix docs are a bit inaccurate here, the `access` manpage reveals:

> The value of amode is either the bitwise-inclusive OR of the access permissions to be checked (R_OK, W_OK, X_OK) or the existence test (F_OK).

Meaning that for the Lua version, either a string containing any combination of 'r', 'w' and 'x' can be used, or the value 'f'. Combining 'f' with any other flag doesn't make sense, as 'f' only checks for existence, not for the file type.

So either we keep the `stat`, or we use `unistd.access(v, 'x')`. I don't think we need to care about non-regular files here, so using `access` would be the simplest (and most common) solution.

AiyionPrime: The idea was to give us room to expand later; but I don't mind dropping regular file check if it gets this merged.
resolved in 7f991bf610fc62cb5f5ea9c28e04e36a2c5fa71c

package/gluon-state/Makefile
AiyionPrime: Fine with me. I consider this resolved. Objections?

NeoRaider (approved):

AiyionPrime:
Just squashed it; compiles right now; will report back once it's running.

AiyionPrime:
Works as expected; this is ready to merge.
`-rw-r--r--    1 root     root             0 Aug 10 09:46 /var/gluon/state/has_default_gw6`

PR #2246, opened 2021-06-21
site: disallow usage of outdoor mode in case outdoor_chanlist is missing

blocktrron:
This is currently not yet tested on the repsective hardware, so RFC for now.

------

The documentation states the outdoor mode is only enabled in case
outdoor_chanlist is defined in the sites wifi5 configurationsection:

 > If enabled, by setting ``wifi5.outdoor_chanlist``, a number of
 > devices that are commonly installed outdoors will have outdoor
 > mode automatically enabled during their initial setup

This currently does not hold true, as Outdoor mode is always activated
for these devices on first setup, even in case the site does not specify
an outdoor chanlist.

Furthermore, it also does not make sense to show the option neither in
setup-modes basic nor advences configuration view in case the site is
lacking an out door chanlist.

This patch will disable outdoor mode on new installs in case the outdoor
chanlist is missing. It will also not show the respective configuration
options in this case.


package/gluon-core/check_site.lua
```diff
@@ -44,8 +44,10 @@ for _, config in ipairs({'wifi24', 'wifi5'}) do
 				120, 122, 124, 126, 128, 132, 134, 136, 138, 140, 142, 144,
 				149, 151, 153, 155, 157, 159, 161, 165, 169, 173 }
 			need_one_of({config, 'channel'}, channels)
-			need_chanlist({config, 'outdoor_chanlist'}, channels, false)
-			need_one_of({config, 'outdoors'}, {true, false, 'preset'}, false)
+
+			local outdoors = need_one_of({config, 'outdoors'}, {true, false, 'preset'}, false)
+			local outdoor_chanlist_required = (outdoors == "preset" or outdoors == true)
```
NeoRaider, line 49:
```suggestion
			local outdoor_chanlist_required = (outdoors ~= false)
```
`outdoors` defaults to 'preset', which makes sense - we want communities to support outdoor mode to respect regulatory rules.

package/gluon-web-wifi-config/luasrc/lib/gluon/config-mode/model/admin/wifi-config.lua
```diff
@@ -142,7 +142,7 @@ uci:foreach('wireless', 'wifi-device', function(config)
 end)
 
 
-if has_5ghz_radio() and not wireless.preserve_channels(uci) then
+if has_5ghz_radio() and not wireless.preserve_channels(uci) and wireless.site_supports_outdoor_mode() then
```
NeoRaider, line 145: I think we should also display the form when outdoor mode is already enabled.

PR #2258, opened 2021-06-30
actions: add mediatek-mt7622 target

blocktrron:
Add the newly added mediatek-mt7622 to the firmware build-test matrix.


mweinelt:
Target name mismatches, also we have no mt7622 target yet.

PR #2259, opened 2021-07-01
gluon-core: util: check if file exists prior to reading

blocktrron:
The file_contains_line helper function was not testing whether a file
exists or not prior attempting to read from it.

Add this check to circumvent errors on the private WiFi config in
case the hwflags file is missing.

Reported-by: Tom Herbers <freifunk@tomherbers.de>
Tested-by: Tom Herbers <freifunk@tomherbers.de>


herbetom (approved):
Solved my issue below: 

Screenshot_20210701_023938

Issue #2256, opened 2021-06-30
Statuspage-error on some TP-Link WR1043v1 with gluon-v2021.1


tackin:
With TP-Lnik WR1043v1-nodes on gluon-v2021.1-1-g0f9a6334+  the call for the statuspage randomly ends with a:
```
XML-Verarbeitungsfehler: nicht wohlgeformt
Adresse: http://[2001:bf7:fc0f:1:76ea:3aff:fee4:d1da]/cgi-bin/status
Zeile Nr. 7, Spalte 12:
 		<title><!DOCTYPE html>
```
curl tells:
```
The called action terminated with an exception:
/usr/lib/lua/gluon/web/template.lua:43: Failed to execute template &#39;status-page&#39;.
A runtime error occurred: [string &#34;/lib/gluon/status-page/view/status-page.htm...&#34;]:90: attempt to index global &#39;nodeinfo&#39; (a nil value)</pre>

```
Seems as if nodeinfo sometimes is not defined because of a timeout of  a "gluon-neighbour-info"-call.

Usually we see this only on small and older nodes-hardware with a higher load. But this one was deeply relaxed.

mweinelt:
Question is if we want to relax the timeout a bit, or if that has downsides.

https://github.com/freifunk-gluon/gluon/blob/ab4c99842a18a88d9752a30b41ba9611d8d52622/package/gluon-status-page/luasrc/lib/gluon/status-page/controller/status-page.lua#L64

grische:
I was about to report a similar bug on my wr841n v13, but those might be related.

Whenever the router boots without WiFi (disabled via the physical button), I am unable to find any status-page/provider processes:
```
root@test:~# ps  | grep status-page/provider
30986 root      1216 S    grep status-page/provider
```

That, in turn, causes the gluon-neighbour-info to return error code 1 and print no output
```
root@test:~# gluon-neighbour-info -d ::1 -p 1001 -t 60 -r nodeinfo
root@test:~# echo $?
1
```

In comparison to a working machine, I can see plenty of status-page/provider processes:
```
root@itworkshere:~# ps  | grep "status-page/provider"
10408 root       704 S    sse-multiplex exec /lib/gluon/status-page/providers/stations 'mesh0'
10414 root      1328 S    /lib/gluon/status-page/providers/stations mesh0
10415 root       704 S    sse-multiplex exec /lib/gluon/status-page/providers/stations 'mesh1'
[...]

PR #2262, opened 2021-07-03
gluon-status-page: relax nodeinfo query timeout

mweinelt:
It was found that a one second timeout for nodeinfo data may be too low,
so that when a node is otherwise occupied that timeout may be reached
too often.

The nodeinfo query response is also vital to the status-page base
template, so that when it times out, the site will be turned in a broken
state, that it cannot recover from.

Fixes: #2256


blocktrron (approved):

NeoRaider:
Maybe we could also improve error handling and at least present an error message to the user when the nodeinfo query fails.

PR #2267, opened 2021-07-09
gluon-core: wireless: support PHY lookup for multi-PHY devices

blocktrron:
The PHY lookup helper "find_phy_by_path" could not lookup the PHY name
for paths from multi-phy devices.

An example for such a path would be:
'1e140000.pcie/pci0000:00/0000:00:01.0/0000:02:00.0+1'

The integer after the plus (+) character determines the PHY index of the
specific device in relation to the PHY with the lowest index of the
device.

For example, if the device provides phy2 and phy3, the above path would
describe phy3. In case the device provides phy0 and phy1, it would
describe phy1.

Rewrite the "find_phy_by_path" function to support those paths as well
as regular device paths in a universal manner.

### DIFF
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/wireless.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/wireless.lua
@@ -2,16 +2,42 @@ local sysconfig = require 'gluon.sysconfig'
 local site = require 'gluon.site'
 local util = require 'gluon.util'
 
+local unistd = require 'posix.unistd'
+local dirent = require 'posix.dirent'
+
 
 local M = {}
 
 local function find_phy_by_path(path)
-	local phy = util.glob('/sys/devices/' .. path .. '/ieee80211/phy*')[1]
-		or util.glob('/sys/devices/platform/' .. path .. '/ieee80211/phy*')[1]
+	local device_path, phy_offset = string.match(path, "^(.+)%+(%d+)$")
 
-	if phy then
-		return phy:match('([^/]+)$')
+	-- Special handling required for multi-phy devices
+	if device_path == nil then
+		device_path = path
+		phy_offset = '0'
 	end
+
+	-- Find the device path. Either it's located at /sys/devices or /sys/devices/platform
+	local path_prefix = ''
+	if not unistd.access('/sys/devices/' .. device_path .. '/ieee80211') then
+		path_prefix = 'platform/'
+	end
+
+	-- Get all available PHYs of the device and dertermine the one with the lowest index
+	local phy_names = dirent.dir('/sys/devices/' .. path_prefix .. device_path .. '/ieee80211')
+	local device_phy_idxs = {}
+	for _, v in ipairs(phy_names) do
+		local phy_idx = v:match('^phy(%d+)$')
+
+		if phy_idx ~= nil then
+			table.insert(device_phy_idxs, tonumber(phy_idx))
+		end
+	end
+
+	table.sort(device_phy_idxs)
+
+	-- Index starts at 1
+	return 'phy' .. device_phy_idxs[tonumber(phy_offset) + 1]
 end
 
 local function find_phy_by_macaddr(macaddr)


package/gluon-core/luasrc/usr/lib/lua/gluon/wireless.lua
```diff
@@ -2,16 +2,49 @@ local sysconfig = require 'gluon.sysconfig'
...
 
 local M = {}
 
 local function find_phy_by_path(path)
-	local phy = util.glob('/sys/devices/' .. path .. '/ieee80211/phy*')[1]
-		or util.glob('/sys/devices/platform/' .. path .. '/ieee80211/phy*')[1]
+	local device_path, phy_offset
+	-- Special handling required for multi-phy devices
+	if string.match(path, '+%d$') then
+		device_path, phy_offset = string.match(path, "([^,]+)+(%d)$")
```
NeoRaider, line 15: The literal `+` characters in both match strings should be escaped using `%`.

package/gluon-core/luasrc/usr/lib/lua/gluon/wireless.lua
```diff
@@ -2,16 +2,49 @@ local sysconfig = require 'gluon.sysconfig'
...
+		phy_idx = tonumber(phy_idx)
+
+		if phy_idx ~= nil and (lowest_phy == nil or lowest_phy > phy_idx) then
+			lowest_phy = phy_idx
+		end
+	end
+
+	if lowest_phy == nil then
+		return nil
+	end
```
NeoRaider, line 45: I think it would be better to create a sorted list of all PHYs at the device and then return the entry at the correct index. This would also work when the PHY indices are not consecutive (which might happen in theory when devices appear and disappear, for example with USB WLAN adapters)

package/gluon-core/luasrc/usr/lib/lua/gluon/wireless.lua
NeoRaider: Also, why the `[^,]` pattern? I would use `^(.+)%+(%d+)$`, which also anchors the beginning of the string and handles PHY indices with multiple digits correctly.

blocktrron: No specific reason, this regex was tanken from iwinfo source.

package/gluon-core/luasrc/usr/lib/lua/gluon/wireless.lua
blocktrron: Okay, I was misunderstanding the iwinfo code a bit. I was under the impression, the PHY index is always absolute to the lowest index, but it is not.

```
	while ((e = readdir(d)) != NULL) {
		int cur_id;

		if (strncmp(e->d_name, "phy", 3) != 0)
			continue;

		cur_id = atoi(e->d_name + 3);
		if (cur_id >= phy_id)
			continue;

		seq++;
	}
```

blocktrron:
Pushed two commits, addressing reviews. Not tested yet.

package/gluon-core/luasrc/usr/lib/lua/gluon/wireless.lua
```diff
@@ -2,16 +2,45 @@ local sysconfig = require 'gluon.sysconfig'
...
+	-- Special handling required for multi-phy devices
+	if string.match(path, '+%d$') then
+		device_path, phy_offset = string.match(path, "^(.+)%+(%d+)$")
+
+		-- The UCI radio path is split into the path and a PHY offset,
+		-- which is incremented for each additional PHY available on
+		-- the device.
+	else
+		device_path = path
+		phy_offset = 0
```
NeoRaider, line 22: There is something wrong here - in the `if` branch, `phy_offset` is a string, while it is a number in the `else` branch.

Also, calling `string.match` twice is unnecessary - just try the split and check if `device_path` is nil afterwards (and the first check is still limited to single-digit indices).

package/gluon-core/luasrc/usr/lib/lua/gluon/wireless.lua
```diff
@@ -2,16 +2,45 @@ local sysconfig = require 'gluon.sysconfig'
...
+	if not unistd.access('/sys/devices/' .. device_path .. '/ieee80211') then
+		path_prefix = 'platform/'
+	end
+
+	-- Get all available PHYs of the device and dertermine the one with the lowest index
+	local phy_names = dirent.dir('/sys/devices/' .. path_prefix .. device_path .. '/ieee80211')
+	local device_phy_idxs = {}
+	for _, v in ipairs(phy_names) do
+		local phy_idx = v:match('(%d)$')
+		phy_idx = tonumber(phy_idx)
```
NeoRaider, line 36: I think this should match `^phy(%d+)$` and check for nil - the iwinfo code also ignores entries that don't start with `phy`.

It is unclear to me how reliable the iwinfo code is, as it doesn't sort the entries returned by readdir. Maybe that's fine for sysfs?

blocktrron: I can't commit on that, but it might have been built with less edge-cases in mind.

NeoRaider (approved):

package/gluon-core/luasrc/usr/lib/lua/gluon/wireless.lua
NeoRaider: Ah, I misunderstood the code of iwinfo, not sorting the entries is fine there.

NeoRaider (approved):

blocktrron:
Works great on my Tenbay MB5EU as well as U6-Lite.

PR #2271, opened 2021-07-11
Remove ramips-rt305x and now unused build features

NeoRaider:


lemoer:
From my side (who added the VoCore long time ago), I'm fine with removing it.

rotanid:
the reason for the removal of this target could have been the reason to remove it years ago, so why now? does not really make sense to me.

lemoer:
@rotanid I think, this comment from IRC explains the "why now":

```
06:24:34       neoraider | blocktrron: since the switch to OpenWrt 21.02, ramips-rt305x is the only remaining target that builds the modules into the
                         | kernel. And I think we could just drop that target as well and get rid of a lot of code...
```

blocktrron:
@NeoRaider

Why not leave the VoCore but drop the size optimizations? VoCore has 8MB of flash, unless the partitioning is very inefficient, we can continue to support it.

NeoRaider:
> the reason for the removal of this target could have been the reason to remove it years ago, so why now? does not really make sense to me.

Before the switch to OpenWrt 21.02, the ramips-rt305x target had other devices. Now that no useful devices are left in the target, we should get rid of it.

rotanid:
> > the reason for the removal of this target could have been the reason to remove it years ago, so why now? does not really make sense to me.
> 
> Before the switch to OpenWrt 21.02, the ramips-rt305x target had other devices. Now that no useful devices are left in the target, we should get rid of it.

i missed that one.
what about the comment by @blocktrron  ?

NeoRaider:
> > > the reason for the removal of this target could have been the reason to remove it years ago, so why now? does not really make sense to me.
> > 
> > 
> > Before the switch to OpenWrt 21.02, the ramips-rt305x target had other devices. Now that no useful devices are left in the target, we should get rid of it.
> 
> i missed that one.
> what about the comment by @blocktrron ?

Keeping the target without the size optimizations would be an option, but I don't see why we should. The rt2x00 WLAN driver is badly maintained (seeing that it currently crashes when trying to use WPA3/OWE). In addition, each target wastes resources for communities that just build all targets, so a target should at least have *some* useful devices - if we want to keep it, I'd at least make it BROKEN.

blocktrron:
@NeoRaider Alright, I was just wondering whether or not you've taken this option into consideration.

I agree that the target itself is not really useful with the current device set (which nobody uses anyway). Let's drop it.

blocktrron (approved):

PR #2285, opened 2021-08-01
docs: add Gluon v2021.1.1 release notes

blocktrron:
Currently soft-blocked by #2284

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ the future development of Gluon.
 
 Please refrain from using the `master` branch for anything else but development purposes!
 Use the most recent release instead. You can list all releases by running `git tag`
-and switch to one by running `git checkout v2021.1 && make update`.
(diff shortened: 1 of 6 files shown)


rotanid (approved):

rotanid:
@blocktrron maybe we want to also rebase to 19.07.8 before releasing?

blocktrron:
I'd like to avoid bumping the OpenWrt feed that close to a release. We are not tracking OpenWrt releases for our releases anyways, so I'm not that concerned about it.

NeoRaider:
#2289 (plus the subsequent revert of the PHY index field in `respondd-airtime`) is too late for v2021.1.1 I guess, but I'd like to get that fixed as soon as possible (to avoid having anyone depend on the PHY index for some reason).

The release notes LGTM.

rotanid:
@blocktrron was it intentional to not send a mail about the release like described in the release process?

PR #2297, opened 2021-08-13
gluon-mesh-babel: implement neighbour check

### LINKED ISSUE #2228: gluon-online-status concept

AiyionPrime:
This lacks a check for `has_default_gw4` or the decision, that babel can't give suitable response.

Resolves #2228.

### DIFF
--- a/package/gluon-mesh-babel/files/lib/gluon/state/check.d/has_neighbours
+++ b/package/gluon-mesh-babel/files/lib/gluon/state/check.d/has_neighbours
@@ -1,0 +1,2 @@
+#!/bin/sh
+out=$(echo dump | nc ::1 33123 | grep "add neighbour" 2>/dev/null) && [ -n "$out" ]


package/gluon-mesh-babel/files/lib/gluon/state/check.d/has_neighbours
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+out=$(echo dump | nc ::1 33123 | grep "add neighbour" 2>/dev/null) && [ -n "$out" ]
```
NeoRaider, line 2:
```suggestion
echo dump | nc ::1 33123 | grep -qF 'add neighbour'
```
As we're only interested whether there is matching output or not, `grep -q` should be used. Adding `-F` makes it more efficient, as it doesn't interpret the pattern as a regex.

AiyionPrime: I still lack a node/network to test this on;
One possible issue I see are possible error-messages of nc.
I'll apply the commit in a moment. Afterwords this can be marked as blocked by #2353 .

christf:
this works as is. However babeld dump is CPU bound on large meshes and can have adverse effects on the performance of route distribution. It is ready for merge. And  after https://github.com/jech/babeld/pull/57 is merged we should rework every component using dump on the babeld socked to write out less data.

AiyionPrime:
@christf Thank you very much.

PR #2304, opened 2021-08-30
gluon-status-page: swap bandwidth limits

herbetom:
I fell like showing the Downstream Limit first is what most people would expect. 

Additionally, the ⌛ that would be created is more in line with what these settings are doing. ;)

cc @AiyionPrime (I think I had this comment before, but decided not to pursue it. Well, I guess that has changed now.)


AiyionPrime (approved):
That's still a nice idea, looks good to me.

PR #2309, opened 2021-09-11
treewide: remove duplicate i18n translations

lemoer:
Before this commit, some *.po files contained the same translation
twice within the same file. While this did not led to errors in
gluon yet, it is still invalid. This commit fixes that and removes
the duplicates.


NeoRaider:
We should probably add something like
```sh
for dir in package/*/i18n; do (cd "$dir"; for po in *.po; do msgmerge -U "$po" *.pot; done); done
```
to CI to catch invalid PO files...

If you run this command on top of your commit, you'll see that you removed the "wrong" translation in some cases - please update your PR with the changes generated by msgmerge.

lemoer:
@NeoRaider Done.

PR #2310, opened 2021-09-11
ramips-mt7621: add support for Ubiquiti UniFi 6 Lite

Dark4MD:
  - [ ] webinterface
  - [ ] tftp
  - [x] other: ssh
as described in openwrt commit https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=fb4d7a9680117a00721936c98ce41eeb2dea95c9
```
Installation
------------

1. Connect to the booted device at 192.168.1.20 using username/password
   "ubnt".

2. Update the bootloader environment.

   $ fw_setenv devmode TRUE
   $ fw_setenv boot_openwrt "fdt addr \$(fdtcontroladdr);
     fdt rm /signature; bootubnt"
   $ fw_setenv bootcmd "run boot_openwrt"

3. Transfer the OpenWrt sysupgrade image to the device using SCP.

4. Check the mtd partition number for bs / kernel0 / kernel1

   $ cat /proc/mtd

5. Set the bootselect flag to boot from kernel0

   $ dd if=/dev/zero bs=1 count=1 of=/dev/mtdblock4

6. Write the OpenWrt sysupgrade image to both kernel0 as well as kernel1

   $ dd if=openwrt.bin of=/dev/mtdblock6
   $ dd if=openwrt.bin of=/dev/mtdblock7

7. Reboot the device. It should boot into OpenWrt.
```

- wired network
- wifi (if applicable)
- led mapping
  - radio leds
    - [n/a] should map to their respective radio
    - [n/a] should show activity
  - switchport leds
    - [n/a] should map to their respective port (or switch, if only one led present) 
    - [n/a] should show link state and activity
- outdoor devices only

Device is running here:
https://hannover.freifunk.net/karte/#/de/map/245a4c1354d4

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -223,6 +223,7 @@ ramips-mt7621
 
   - EdgeRouter X
   - EdgeRouter X-SFP
+  - UniFi 6 Lite
 
 * ZBT
 
--- a/targets/ramips-mt7621
+++ b/targets/ramips-mt7621
@@ -29,6 +29,13 @@ device('netgear-wndr3700-v5', 'netgear_wndr3700-v5', {
 })
 
 
+-- Ubiquiti
+
+device('ubiquiti-unifi-6-lite', 'ubnt_unifi-6-lite', {
+	factory = false,
+})
+
+
 -- Xiaomi
 
 device('xiaomi-mi-router-4a-gigabit-edition', 'xiaomi_mi-router-4a-gigabit', {


blocktrron:
mt7620 --> mt7621

Have you checked whether the Config-Mode WiFi page works correctly? I can't remember entirely whether this was already completely fixed in openwrt-21.02 or we are missing iwinfo backports for handling multi-phy devices.

Dark4MD:
> mt7620 --> mt7621

And i checked everything twice... Corrected now.

> Have you checked whether the Config-Mode WiFi page works correctly? I can't remember entirely whether this was already completely fixed in openwrt-21.02 or we are missing iwinfo backports for handling multi-phy devices.

Didn't had an error on first setup where i mostly go through every page.
Just checked again to be sure and still no errors. But it also is just the normal dual band hardware config.

blocktrron:
Sorry, my claim wasn't valid. The U6-lite does use a separate age-old mt7603 for the 2.4 GHz. SO there shouldn't be an issue there.

PR #2313, opened 2021-09-15
x86: don't build ext4 images

lemoer:
Since we're discarding the ext4 images anyways, we now stop building 
them as well and save a few seconds of build time.

### DIFF
--- a/contrib/actions/generate-target-filters.py
+++ b/contrib/actions/generate-target-filters.py
@@ -3,6 +3,8 @@
 # Update target filters using
 #   make update-ci
 
+import re
+import os
 import sys
 import json
 
@@ -23,6 +25,13 @@ extra = [
 
 _filter = dict()
 
+# INCLUDE_PATTERN matches:
+# include '...'
+# include "..."
+# include("...")
+# include('...')
+INCLUDE_PATTERN = "^\\s*include *\\(? *[\"']([^\"']+)[\"']"
+
 # construct filters map from stdin
 for target in sys.stdin:
     target = target.strip()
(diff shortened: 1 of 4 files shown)


NeoRaider:
`filters.yml` is generated by `scripts/update-ci.sh`, so your changes will get lost with the next update...

lemoer:
@NeoRaider Thanks. Updated.

contrib/actions/generate-target-filters.py
```diff
@@ -31,6 +31,12 @@
         f"targets/{target}"
     ] + common
 
+    if target.startswith('x86-'):
+        _filter[target].append('targets/x86.inc')
+
+    if target.startswith('bcm27xx-'):
+        _filter[target].append('targets/bcm27xx.inc')
```
NeoRaider, line 38: Hmm... on the one hand I'd like to have a more generic solution like the following (untested):
```py
if '-' in target:
    _filter[target].append('targets/{}.inc'.format(target.split('-')[0]))
```

On the other hand, we are building more targets than necessary now: `x86-geode` does not include `x86.inc`.

lemoer:
@NeoRaider Ok. Updated.

contrib/actions/generate-target-filters.py
```diff
@@ -23,6 +25,13 @@
 
 _filter = dict()
 
+# INCLUDE_PATTERN matches:
+# include '...'
+# include "..."
+# include("...")
+# include('...')
+INCLUDE_PATTERN = "^ *include[\(\ ][\"'](.*)[\"']\)? *$"
```
NeoRaider, line 33: I propose the following pattern:
```suggestion
INCLUDE_PATTERN = "^\\s*include *\\(? *[\"']([^\"']+)[\"']"
```
- Allow any leading whitespace (in Gluon we usually use tabs)
- Allow any trailing string after the closing quote (the code might have a comment)
- Make whitespace matching between `include` and the opening quote more flexible
- Fix number of backslashes in a few places
- Only allow non-quote characters between the quotes, require at least one character
I have only tested this on a Python shell and not with the Gluon code, but it seems to work as expected.

lemoer:
@NeoRaider Ok. Updated again.

PR #2314, opened 2021-09-17
v2021.1.x: ipq40xx: Fix automatic ipq-wifi selection for Plasma Cloud devices

ecsv:
It was noticed that various devices had not the correct board-2.bin installed. This was caused by a typo in the package name. The ath10k driver (unfortunately) is then loading a completely unrelated BDF from the ath10k-board-qca4019 board-2.bin. It is
usually a rather bad idea to use calibration data from a different board - but the effects depend on the actual device.

For the PA1200, it was mostly noticed by the bad 2.4GHz performance.

Upstream PR is https://github.com/openwrt/openwrt/pull/4538


NeoRaider:
Huh. Shouldn't the build fail when a DEVICE_PACKAGE is missing?

ecsv:
Not sure about what was expected from the OpenWrt DEVICE_PACKAGES implementation. But it doesn't fail when a package is not available.

PR #2316, opened 2021-09-21
Clean up Gluon version number handling, allow .scmversion override

NeoRaider:
Alternative implementation for #2315 (@lemoer), and some more or less related cleanup.


scripts/getversion.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+if [ $# -ne 1 ]; then
+	echo >&2 "Usage: getversion.sh <directory>"
+	exit 1
+fi
+
+cd "$1" || exit 1
+
+cat .scmversion 2>/dev/null && exit 0
```
lemoer, line 10: I would suggest something like this:

```suggestion
if [ "$2" -ne "--ignore-scmversion" ]; then
  cat .scmversion 2>/dev/null && exit 0
fi
```

So you can do `./scripts/getversion.sh . --ignore-scmversion > .scmversion` in your meta build script.

lemoer (changes requested):
I requested a little change above.

Apart from that, please add a description like this to the commit message (or even better to the docs):

---

To override the version from your meta build system, you can do something like this in your meta build system:

```
cd gluon
./scripts/getversion.sh . --ignore-scmversion > .scmversion
# apply your own patches to gluon
git am < ...
# make gluon
make GLUON_TARGET=...
```

scripts/getversion.sh
NeoRaider: `rm -f .scmversion;` is shorter than `--ignore-scmversion` :)

NeoRaider:
I'll add a doc section to `dev/build.rst`.

PR #2319, opened 2021-09-23
rockchip: fix broken squashfs sysupgrade

blocktrron:
The rockchip platform supports squashfs SD card images. However, the
resulting image is not padded to completely fill the rootfs partition.

Because of that, the f2fs overlay might not be erased, resulting in
uci-defaults not bing executed or the configuration not being erased,
even though drop config was selected.

Modify the image generation process so the image is padded to cover the
entire root filesystem partition.


blocktrron:
Closed with https://github.com/freifunk-gluon/gluon/commit/4684f42173bea54c6626ce16fe14397692047775

PR #2322, opened 2021-10-03
gluon-neighbour-info: fix broken output with large results

T-X:
Currently a buffer with a fixed size of 8192 bytes is used. However the
result can potentially be larger, which leads to a truncated JSON
output on stdout. UDP packets, without compression and with IP
fragmentation, can be up to 64KiB large.

Instead of using a fixed size buffer on the stack ask the kernel first
about the size of the UDP data and allocate a buffer of appropriate size
on the heap before receiving the UDP data.

The issue was observed with a custom respondd provider.

### DIFF
--- a/package/gluon-neighbour-info/src/gluon-neighbour-info.c
+++ b/package/gluon-neighbour-info/src/gluon-neighbour-info.c
@@ -69,8 +69,23 @@ void tv_subtract (struct timeval *r, const struct timeval *a, const struct timev
 	}
 }
 
-ssize_t recvtimeout(int socket, void *buffer, size_t length, int flags, const struct timeval *timeout) {
+void resize_recvbuffer(char **recvbuffer, size_t *recvbuffer_len, size_t recvlen)
+{
+	free(*recvbuffer);
+	*recvbuffer = malloc(recvlen);
+
+	if (!(*recvbuffer)) {
+		perror("Could not resize recvbuffer");
+		exit(EXIT_FAILURE);
+	}
+
+	*recvbuffer_len = recvlen;
+}
+
+ssize_t recvtimeout(int socket, char **recvbuffer, size_t *recvbuffer_len,
+		    const struct timeval *timeout) {
 	struct timeval now, timeout_left;
+	ssize_t recvlen;
 
 	getclock(&now);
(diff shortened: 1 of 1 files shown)


T-X:
Changelog v2:
* added missing error check (e.g. on timeout) after recv msg-peek

package/gluon-neighbour-info/src/gluon-neighbour-info.c
```diff
@@ -69,8 +72,22 @@ void tv_subtract (struct timeval *r, const struct timeval *a, const struct timev
 	}
 }
 
-ssize_t recvtimeout(int socket, void *buffer, size_t length, int flags, const struct timeval *timeout) {
+void resize_recvbuffer(size_t recvlen)
+{
+	free(recvbuffer);
+	recvbuffer = malloc(recvlen);
```
blocktrron, line 78: Any reason not to use `realloc` here?

NeoRaider: If there is no data to be retained in the buffer, `realloc` would just caused an unnecessary memcpy if the buffer needs to be relocated.

package/gluon-neighbour-info/src/gluon-neighbour-info.c
```diff
@@ -35,6 +35,9 @@
 #include <string.h>
 #include <time.h>
 
+char *recvbuffer;
+size_t recvbuffer_len;
```
NeoRaider, line 39: I don't see why this needs to be global - just pass pointers to the functions that need the buffer.

And if you need global variables, always mark them as 'static' to avoid adding them to the dynamic symbol table.

T-X:
Changelog v2:
* moved recvbuffer and recvbuffer_len from data segment to stack

package/gluon-neighbour-info/src/gluon-neighbour-info.c
T-X: done

AiyionPrime:
huh ^^
I stumbled across the fixed buffer on friday and was curious why its limited size has not ever led to problems, yet.
@T-X Thanks for picking this up!

PR #2323, opened 2021-10-05
gluon-neighbour-info: avoid recv() with NULL buffer

NeoRaider:
Calling functions like recv() with a NULL buffer is not explicitly
allowed by the POSIX standard, so it must be avoided to be portable
across different libc implementations. Allocate an initial buffer before
handling requests, and also pass this buffer to the peek recv() call.

Fixes: 531937cf6f3c ("gluon-neighbour-info: fix broken output with large results")

### DIFF
--- a/package/gluon-neighbour-info/src/gluon-neighbour-info.c
+++ b/package/gluon-neighbour-info/src/gluon-neighbour-info.c
@@ -95,7 +95,7 @@ ssize_t recvtimeout(int socket, char **recvbuffer, size_t *recvbuffer_len,
 
 	setsockopt(socket, SOL_SOCKET, SO_RCVTIMEO, &timeout_left, sizeof(timeout_left));
 
-	recvlen = recv(socket, NULL, 0, MSG_PEEK | MSG_TRUNC);
+	recvlen = recv(socket, *recvbuffer, 0, MSG_PEEK | MSG_TRUNC);
 	if (recvlen < 0)
 		return recvlen;
 
@@ -268,6 +268,8 @@ int main(int argc, char **argv) {
 		fputs("Content-Type: text/event-stream\n\n", stdout);
 		fflush(stdout);
 	}
+
+	resize_recvbuffer(&recvbuffer, &recvbuffer_len, 8192);
 
 	do {
 		ret = request(sock, &recvbuffer, &recvbuffer_len, &client_addr,


T-X:
Ok, let's check.

Userspace, musl: In musl musl-1.1.24, it seems to be just a syscall, I can't see anything in userspace that would complain about NULL [here](https://git.musl-libc.org/cgit/musl/tree/src/network/recv.c?h=v1.1.24) and [here](https://git.musl-libc.org/cgit/musl/tree/src/network/recvfrom.c?h=v1.1.24).

Userspace, glibc master: On Linux, it generally also seems to be just a plain syscall [here](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/recv.c;h=a26ee5e98381b93ae90bb2e876fba111e86722df;hb=HEAD). For hurd, it seems glibc added [support for NULL in 1999](https://sourceware.org/git/?p=glibc.git;a=commitdiff;h=d012636f1f27cfd;hp=9cfba5dc11db58e94c1ef828aa9274b58bdfc513) // [Changelog](https://sourceware.org/git/?p=glibc.git;a=blob;f=ChangeLog.old/ChangeLog.10;h=f2eb42b300ff80e015efeabfd4b5e8fcef3c19b4;hb=HEAD#l12453) (not that we would care about Hurd, I guess). So it sounds to me that glibc is safe, too.

On the Linux kernel side:

Did not seem to get rejected in my tests. And Linux should not silently write to the NULL pointer as the length is 0. In the code, [__sys_recvfrom()](https://elixir.bootlin.com/linux/latest/source/net/socket.c#L2051) doesn't do anything fancy with a NULL buffer. It calls [import_single_range](https://elixir.bootlin.com/linux/latest/source/lib/iov_iter.c#L1957) which calls [access_ok(buf, len)](https://elixir.bootlin.com/linux/latest/source/arch/mips/include/asm/uaccess.h#L79), but that again will just return (0 == 0), so true/ok.

On [stackoverflow](https://stackoverflow.com/questions/16433311/what-happens-when-recv-called-with-null-buffer), I see some people saying "never use NULL", but without much further explanation or links to sources.

So to me it seems that using NULL is fine when you use a length of 0.

T-X:
Also this change doesn't make much sense to me, on the first run *recvbuffer is NULL anyway.

PR #2330, opened 2021-10-26
Status page cleanup

### LINKED ISSUE #2329: Minor statuspage issues
### Bug report

**What is the problem?**
In JavaScript `return` does behave like `continue` in a `forEach()`-iteration.
The function `address_to_groups()` is intended to return nothing on a failed parse, but does indeed a shorter array than expected.

https://github.com/freifunk-gluon/gluon/blob/7b2ab5e6c649f7c6c3f115773cdde82f0d883a85/package/gluon-status-page/javascript/status-page.js#L556-L566

(length of 8 expected, returns less on error)

This hasn't caused any trouble yet, I just stumbled over it, while reading the statuspages codebase.
The introducing commit was about four years ago.

### Bug report
The other minor issue was introduced in the same migration and was acked by neoraider yesterday:
`lastImage` is never set and therefore evaluated uselessly afterwords.

https://github.com/freifunk-gluon/gluon/blob/7b2ab5e6c649f7c6c3f115773cdde82f0d883a85/package/gluon-status-page/javascript/status-page.js#L285-L294

**Gluon Version:**
latest master as well as several past releases

I'll send a PR in a few minutes.

AiyionPrime:
This intends to resolve #2329 and fixes a few missing (explicit) semicolons while at it.

Code is not tested yet and should be, as
though I am sure it does now, what is was expected to do in the first place, four years are a long time and maybe a now working resize-feature breaks stuff today.


AiyionPrime:
> Please add the explanation for the forEach issue from #2329 to the commit message, otherwise it is hard to understand what you are fixing without looking at Github.

Sorry, been otherwise busy in the last weeks, it's done.

PR #2331, opened 2021-10-29
Add archer c25 v1

GoliathLabs:
Hello,
I've added the archer c25 v1 again, because we got a firmware request for the device over at Freifunk München. We built an image and sent that to the requestor. He seems to be able to use it now without OOM problems. Maybe the OOM problem was fixed by OpenWrt 21.02?


GoliathLabs:
You're right. The memory usage has gone up. Therefore I've added the broken flag again

blocktrron:
Please add the device support checklist from the wiki.

PR #2334, opened 2021-11-26
Added: Stale GitHub Action

GoliathLabs:
Hello,
I have added the stale GitHub action that now closes issues and PRs that have no activity within 30 days. If the PR or issue is still relevant, one can comment below or remove the "stale" tag.

This will sort out some old stuff

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,15 @@
+name: 'Close stale issues and PRs'
+on:
+  schedule:
+    - cron: '30 1 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v4
+        with:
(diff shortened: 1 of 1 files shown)


blocktrron:
In my opinion, closing stuff should be an active decision, which does not always apply to stuff because it is "old". Or was the intention behind this change something different?

GoliathLabs:
Hmm, I think you are right. I just took a look at the "issues" section and some date back to 2014. Maybe something like the Stale action would point out the things that need to be done, since you have to actively remove the "Stale" tag or post a comment below.

blocktrron:
Issues are around mostly for lack of decision. Closing them after 30 days won't solve the underlying problem (which is limited time from maintainers). So a auto-close action would just hide them from visibility.

With the current volume Gluon has, closing issues manually is still doable.

In addition, Closing a PR / Ticket should be a well decided and well explained step, so it is clear for someone stumbling on such a contribution, how and why this decision was done.

PR #2338, opened 2021-12-15
Adjust for OpenWrt hwmode -> band wireless config change

NeoRaider:
OpenWrt recently backported a number of changes to the wireless setup scripts to 21.02, breaking our wireless configuration.

- Change our code to use the new option (by @blocktrron, added commit message)
- Add migration code

PR #2339, opened 2021-12-15
gluon-status-page: use UCI + iwinfo to get channel information

### LINKED ISSUE #2336: Status page channel display broken with outdoor mode enabled
### Bug report

**What is the problem?**

Status page radio channel display fails with outdoor mode.

```html

				<h3>Radios</h3>
					<table>
						<tr>
								<th>radio0</th>
								<td><!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml" lang="" xml:lang="">
	<head>
		<meta charset="utf-8" />
		<meta name="viewport" content="width=device-width, user-scalable=no" />

		<title>Error</title>

		<link rel="stylesheet" href="/static/status-page.css" type="text/css" />
	</head>
	<body>
		<header>
			<h1>Error</h1>
		</header>
		<div class="container">
			<div class="frame">
				<h2 name="content">500 Internal Server Error</h2>
<p>Sorry, the server encountered an unexpected error.</p>
<pre class="error500">Failed to execute dispatcher target for entry &#39;/&#39;.
/usr/lib/lua/gluon/web/template.lua:43: Failed to execute template &#39;status-page&#39;.
A runtime error occurred: /usr/lib/lua/gluon/web/i18n.lua:29: bad argument #1 to &#39;format&#39; (integer expected, got string)</pre>

			</div>
```

This is the result of querying `ubus call networking.wireless status`, which only holds the configuration, but no live channel data. In this special case the config dictionary does not contain `channel`, but `channels`, so channel is `nil` in L46:

https://github.com/freifunk-gluon/gluon/blob/e153ba6bd70ed8bd1d78b095f173bb75c95fb082/package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html#L38-L49

This is how the status response looks like with outdoor mode enabled:

```json
{
	"radio0": {
		"up": true,
		"pending": false,
		"autostart": true,
		"disabled": false,
		"retry_setup_failed": false,
		"config": {
			"hwmode": "11a",
			"path": "pci0000:00/0000:00:00.0",
			"channel": "auto",
			"htmode": "VHT20",
			"country": "DE",
			"channels": [
				"96-116",
				"132-140"
			],
			"hostapd_options": [
				"country3=0x4f"
			]
		},
```

To get the actual channel we would need to query `ubus call hostapd.client0 get_status` and make sure index for the client or owe interfacecs , which are managed by hostapd, match the relevant phy index.

```
# ubus call hostapd.client0 get_status
{
	"status": "ENABLED",
	"freq": 5520,
	"phy": "client0",
	"airtime": {
		"time": 1340796,
		"time_busy": 8998,
		"utilization": 1
	},
	"dfs": {
		"cac_seconds": 60,
		"cac_active": false,
		"cac_seconds_left": 0
	}
}
```

**Gluon Version:**

v2021.1-169-ge153ba6b

**Site Configuration:**

n/a

**Custom patches:**

n/a

NeoRaider:
The network.wireless status ubus call only returns the configured
channel from UCI, breaking the status page in outdoor mode, where the
configuration contains 'auto' instead of a number.

Fixes: 0d3fa6b59be7 ("gluon-status-page: use ubus to get radio channels")
Closes #2336

(apply on top of #2338 to get a usable configuration for testing)

### DIFF
--- a/package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
+++ b/package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
@@ -1,7 +1,11 @@
 <%-
+	local iwinfo = require 'iwinfo'
 	local ubus = require 'ubus'
 	local unistd = require 'posix.unistd'
 	local util = require 'gluon.util'
+	local wireless = require 'gluon.wireless'
+
+	local uci = require('simple-uci').cursor()
 
 	local translations = {}
 	local site_i18n = i18n 'gluon-site'
@@ -35,22 +39,22 @@
 		return interfaces
 	end
 
-	local function get_radios(uconn)
-		local radios = uconn:call("network.wireless", "status", {})
+	local function get_radios()
 		local ret = {}
 
-		for radio, info in pairs(radios) do
-			if info.up then
+		wireless.foreach_radio(uci, function(radio)
+			local channel = iwinfo.nl80211.channel(wireless.find_phy(radio))
+			if channel then
 				table.insert(ret, {
-					name = radio,
-					channel = info.config.channel,
+					name = radio['.name'],
+					channel = channel,
 				})
 			end
-		end
-
+		end)
 		table.sort(ret, function(a, b)
 			return a.name < b.name
 		end)
+
 		return ret
 	end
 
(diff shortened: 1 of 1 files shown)


herbetom:
@NeoRaider Since #2338 was merged, this can now be rebased.

I've build an image and for me #2336 seems to be solved with this :+1:

blocktrron (approved):

PR #2347, opened 2021-12-23
targets/ramips-mt76x8: add device TP-Link RE200 V4

misanthropos:
Signed-off-by: Richard Fröhning <misanthropos@gmx.de>

  - [x] webinterface
  - [x] tftp
  - [ ] other: <specify>
- wired network
- wifi (if applicable)
  - [ ] association with AP must be possible on all radios
- led mapping
  - radio leds
    - [x] should show activity
  - switchport leds
- outdoor devices only

### DIFF
--- a/targets/ramips-mt76x8
+++ b/targets/ramips-mt76x8
@@ -60,6 +60,10 @@ device('tp-link-archer-c50-v4', 'tplink_archer-c50-v4', {
 
 device('tp-link-re200-v2', 'tplink_re200-v2')
 
+device('tp-link-re200-v4', 'tplink_re200-v4', {
+	broken = true, -- no 5GHz connectivity (scanning works)
+})
+
 device('tp-link-re305', 'tplink_re305-v1', {
 	class = 'tiny', -- Only 6M of usable Firmware space
 })


mweinelt:
Please fill out the device integration checklist and copy it into the top post.

https://github.com/freifunk-gluon/gluon/wiki/Device-Integration-checklist.

blocktrron:
> 16:51 <Misanthropos> hexa-, thank you for the input. Added the Device Integration checklist. There is something odd however. The Power/Sys LED is off by default. In config-mode the LAN interface blinks. Changing trigger of power:green affects all leds.


Which revision is this referring to?

misanthropos:
> Which revision is this referring to?

Both, V3 and V4 show the same behaviour.

blocktrron:
In this case, this sounds like the GPIO assignment of the LEDs is off. Check the GPL tarball, maybe they are defined in there.

AiyionPrime:
For the next one passing by:
@misanthropos is the author of https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=33b76d66d1afa9d888f32587b18d78a930593ca5.

misanthropos:
I have found correct GPIO assignments which will find its way into OpenWRT with this PR:
https://github.com/openwrt/openwrt/pull/4954

AiyionPrime:
@misanthropos in order to get this going again, might you start resolvin the conflicts?
maybe @adschm could then do another review.

AiyionPrime:
tested backport patch is open for review
https://patchwork.ozlabs.org/project/openwrt/patch/20221108123102.666033-1-openwrt@aiyionpri.me/

AiyionPrime:
currently blocked by #2698

AiyionPrime:
@misanthropos This needs a rebase, as well as testing.
As your previous tests suggest, this revision might have trouble with at least one of its wifi interfaces and as such should be marked as broken, until it 's resolved.

docs/user/supported_devices.rst
```diff
@@ -407,6 +407,7 @@ ramips-mt76x8
   - Archer C50 (v3)
   - Archer C50 (v4)
   - RE200 (v2)
+  - RE200 (v4)
```
AiyionPrime, line 410:
```suggestion
```

broken devices are not supported

AiyionPrime (changes requested):

docs/user/supported_devices.rst
misanthropos: I can close this PR and reopen it by the time its fixed?

AiyionPrime: You don't need to close this. You just need to remove it from the docs.
Once the issue is resolved you can open a PR that removes the `broken` tag and adds it as supported to the docs.

blocktrron:
As @AiyionPrime told me, the 5 GHz interface does not work on channels above 64, right?

Can you provide a image of the EEPROM / caldata partition? The MT7612 can have channel allowance for controlling permitted channels with a single SKU programmed into it's EEPROM, i have never seen those but given it does seem to outright "not work", i assume this is the case here.

This might be fixable by extending mt76 to override this setting (which should be done anyways, given we use the wireless-regdb for this case).

blocktrron:
@misanthropos 

Check which partition is labled eeprom with `cat /proc/mtd`, dump the content using `dd if=/dev/mtdX of=/tmp/eeprom.bin` and transfer this file to GitHub.

misanthropos:
This is the result of cat /proc/mtd
```
root@re200v4:~# cat /proc/mtd 
dev:    size   erasesize  name
mtd0: 00020000 00010000 "u-boot"
mtd1: 007a0000 00010000 "firmware"
mtd2: 0020f81b 00010000 "kernel"
mtd3: 005907e5 00010000 "rootfs"
mtd4: 00250000 00010000 "rootfs_data"
mtd5: 00030000 00010000 "config"
mtd6: 00010000 00010000 "radio"
```

AiyionPrime:
@misanthropos dumps of `/dev/mtd/radio` (and `/dev/mtd/config` to be sure) should be sufficient.

blocktrron:
@misanthropos can you try lowering the TX power? Start with txpower of 1 and work your way up until it does not work.

Looking at the code, "MAC error detected" seems to happen when beacons are stuck. This either indicates a bug in SW / HW or possibly noise on the channel, which might also be fedd back to the chip from itself.

My initial theory was that maybe the  MAC prohibits sending frames due to channels being excluded in EEPROM data. This is not the case, as EEPROM does not specify channel limits / regdom.

misanthropos:
@blocktrron so here are my observations:

- Only txpower mode set to "auto" possible

´´´
iw client1 set txpower 1
Invalid parameter: 1
iw client1 set txpower fixed 10
command failed: Not supported (-122)
´´´
Same issue on re200v3.

Those are the defaults:
                  re200v3                                    re200v4
channel     44                                             36
power        11 dBm                                     10 dBm

AiyionPrime:
@blocktrron I think this might be the commit, that introduced the behavior described above.
https://github.com/openwrt/mt76/commit/20d1fed838b958701ed5d030b663ba9227477ab1

As the issue was reportedly resolved:

> "MAC error detected" problem was fixed. Although channel 149-161 still doesn't work, it might be a hardware bug, so maybe we can close this issue.

_Originally posted by @DragonBluep in https://github.com/openwrt/mt76/issues/666#issuecomment-1166278390_

I would have expected @misanthropos reported problems are gone on a branch that contains the commit @nbd did two days earlier:

https://github.com/openwrt/mt76/commit/ba39ed3b44f1a0b3bd87c72452a261908afba754

That is the case for both OpenWrt's `master` and `openwrt-22.03` though.

**22.03** https://github.com/openwrt/mt76/commit/d70546462b7b51ebc2bcdd5c534fdf3465be62a4 referenced in https://github.com/openwrt/openwrt/blob/openwrt-22.03/package/kernel/mt76/Makefile#L12

That's from early september.
Looking at @misanthropos branch history (rebased 15 days ago) I'd assume they are included.

@blocktrron Any idea what I'm missing?

misanthropos:
Needs reviewing of 5GHz - will reopen if it gets fixed in upstream.

kpoman:
Hey guys, this is still failing on a fresh 23.05.0 installed on my RE200 V4... I just get mac errors. I tried different txpower levels with no luck.
`root@OpenWrt:~# logread | grep mt76x0e
Tue Nov 14 01:17:29 2023 kern.info kernel: [   12.597389] mt76x0e 0000:01:00.0: card - bus=0x1, slot = 0x0 irq=4
Tue Nov 14 01:17:29 2023 kern.info kernel: [   12.610061] mt76x0e 0000:01:00.0: ASIC revision: 76100002
Tue Nov 14 01:17:29 2023 kern.info kernel: [   12.700810] mt76x0e 0000:01:00.0: Firmware Version: 0.1.00
Tue Nov 14 01:17:29 2023 kern.info kernel: [   12.985556] mt76x0e 0000:01:00.0: EEPROM ver:01 fae:00
Tue Nov 14 01:17:58 2023 kern.err kernel: [   46.895026] mt76x0e 0000:01:00.0: MAC error detected
...
Tue Nov 14 01:30:50 2023 kern.err kernel: [  118.665109] mt76x0e 0000:01:00.0: MAC error detected
Tue Nov 14 01:30:51 2023 kern.err kernel: [  119.895105] mt76x0e 0000:01:00.0: MAC error detected
root@OpenWrt:~#`

AiyionPrime:
Welcome @kpoman. This is the gluon repository.
We are currently not tracking this device as supported as we found the same flaws you just encountered.

Once the Issues are addressed in OpenWRT (if possible at all) we can continue this thread.
Until then I'd recommend OpenWRT's issue tracker for this bugreport.

PR #2351, opened 2021-12-28
docs: add docs for gluon-mesh-vpn-wireguard

lemoer:

PR #2352, opened 2021-12-28
RFC: gluon-mesh-vpn-*: make MTU of VPN device provider specific

lemoer:
If a community uses different vpn providers, they typically
assume the same MTU for the wan device underneath the VPN. As
different VPN providers however have different overhead, the MTU
of the VPN device differs for each provider. Therefore this
commit makes the MTU of the VPN device provider specific.

This has two advantages:
1. The same site.conf can used to bake firmwares for different
   VPN providers (only by selecting a diferent vpn feature in the
   site.mk).
2. We are coming closer to the option of integrating multiple VPN
   providers into one firmware.

### DIFF
--- a/contrib/ci/minimal-site/site.conf
+++ b/contrib/ci/minimal-site/site.conf
@@ -93,7 +93,6 @@
 
   mesh_vpn = {
     -- enabled = true,
-    mtu = 1312,
 
     fastd = {
       -- Refer to https://fastd.readthedocs.io/en/latest/ to better understand
@@ -101,6 +100,7 @@
 
       -- List of crypto-methods to use.
       methods = {'salsa2012+umac'},
+      mtu = 1312,
       -- configurable = true,
       -- syslog_level = 'warn',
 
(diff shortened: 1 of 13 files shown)


lemoer:
This is currently RFC. It's neither properly tested yet, neither the docs are adjusted. Feel free to comment.

If the idea is generally well accepted, I'll do the rest.

lemoer:
The changes are now tested successfully and docs have been added.

lemoer:
To adjust docs/user/site.rst, I rebased this PR on #2351 (which itself is rebased on #2349 and #2186).

lemoer:
We just discussed on our RC3 meeting.

 @NeoRaider is against the move of the MTU. We need a third opinion here.

blocktrron (approved):

blocktrron:
Discussed with @AiyionPrime and @lemoer , I'm in favor of merging this proposal, as the MTU is a property of the link which depends on the VPN protocol. Therefor, this change is logical to me.

PR #2363, opened 2021-12-31
scripts/container.sh: allow for empty Branch Names

### LINKED ISSUE #2362: scripts/container.sh: script doesn't work if there's currently no branch
The new (and quite helpfull) `container.sh` script introduced in #2292 doesn't work if you're not on a branch:

```
$ bash -x scripts/container.sh 
+ set -euo pipefail
++ dirname scripts/container.sh
+ cd scripts/..
++ git branch --show-current
++ sed 's/[^a-z0-9-]/_/ig'
+ BRANCH=
+ TAG=gluon:
++ command -v podman
+ '[' '' ']'
++ command -v docker
+ '[' /usr/bin/docker ']'
+ docker build -t gluon: contrib/docker
invalid argument "gluon:" for "-t, --tag" flag: invalid reference format
See 'docker build --help'.
```

This situation occurs, for example, when you check out a release tag. So there should be probably some kind of fallback.

cc @mweinelt

herbetom:
I'm sure there are better solutions. I just want to put the simplest and fastest one I can think of with the PR up for discussion. 

An interesting solution I could think of would be to get a hash of the Dockerfile and tag the container with it. But that would probably be a bit over engineered and in itself error prone.

resolves #2362

### DIFF
--- a/scripts/container.sh
+++ b/scripts/container.sh
@@ -6,8 +6,8 @@ set -euo pipefail
 cd "$(dirname "$0")/.."
 
 # normalize branch name to reflect a valid image name
-BRANCH=$(git branch --show-current | sed 's/[^a-z0-9-]/_/ig')
-TAG=gluon:${BRANCH}
+BRANCH=$(git branch --show-current 2>/dev/null | sed 's/[^a-z0-9-]/_/ig')
+TAG="gluon:${BRANCH:-latest}"
 
 if [ "$(command -v podman)" ]
 then


scripts/container.sh
```diff
@@ -7,6 +7,7 @@ cd "$(dirname "$0")/.."
 
 # normalize branch name to reflect a valid image name
 BRANCH=$(git branch --show-current | sed 's/[^a-z0-9-]/_/ig')
```
NeoRaider, line 9:
```suggestion
BRANCH=$(git branch --show-current 2>/dev/null | sed 's/[^a-z0-9-]/_/ig')
```
Let's also hide error messages.

scripts/container.sh
```diff
@@ -7,6 +7,7 @@ cd "$(dirname "$0")/.."
...
+[ -z "${BRANCH}" ] && BRANCH=default
 TAG=gluon:${BRANCH}
```
NeoRaider, line 11: I think using `latest` as the fallback makes more sense. This can also be written more concisely:
```suggestion
TAG="gluon:${BRANCH:-latest}"
```

NeoRaider (changes requested):

scripts/container.sh
herbetom: nice, I googled for a short oneliner like this but hadn't found anything useful.

NeoRaider (approved):

PR #2364, opened 2021-12-31
Clean up obsolete upgrade migrations

NeoRaider:
As preparation for the regeneration of /etc/config/network on upgrades, clean up the current upgrade scripts.


rotanid:
with v2021.1 we wrote in the release notes, after a cleanup:
**Upgrades to v2021.1 and later releases are only supported from releases v2018.2 and later. This is due to migrations that have been removed to simplify maintenance.**

do your changes need a similar sentence in the v2022.1 release notes? if yes, we need to track this somehow.

NeoRaider:
@rotanid All migrations removed by this PR are much older than v2018.2, with the exception of the `batman-adv-legacy` migration, which was already useless, since it modified the file `/etc/config/batman-adv-legacy`, which is not used by anything anymore.

It is possible that we will have to drop further migration support with the new network configuration scripts I'm working on, but I'm still figuring out the finer details of that.

PR #2366, opened 2022-01-01
libpcap: Add support for B.A.T.M.A.N. Advanced

T-X:
This adds support for the layer 2 mesh routing protocol
B.A.T.M.A.N. Advanced. "batadv" can be used to filter on batman-adv
packets. It also allows later filters to look at frames inside the
tunnel when both "version" and "type" are specified.

Documentation for the batman-adv protocol can be found at the following
locations:

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/networking/batman-adv.rst
https://www.open-mesh.org/

---

This is a requirement for the gluon-statistics-mcast package and eases batman-adv network debugging in general. A pull-request to upstream libpcap is pending and has gone through some first rounds of reviewing here:

https://github.com/the-tcpdump-group/libpcap/pull/980

### DIFF
--- a/patches/openwrt/0007-libpcap-Add-support-for-B.A.T.M.A.N.-Advanced.patch
+++ b/patches/openwrt/0007-libpcap-Add-support-for-B.A.T.M.A.N.-Advanced.patch
@@ -1,0 +1,678 @@
+From: Linus Lüssing <linus.luessing@c0d3.blue>
+Date: Fri, 1 Jan 2021 02:15:44 +0100
+Subject: libpcap: Add support for B.A.T.M.A.N. Advanced
+
+This adds support for the layer 2 mesh routing protocol
+B.A.T.M.A.N. Advanced. "batadv" can be used to filter on batman-adv
+packets. It also allows later filters to look at frames inside the
+tunnel when both "version" and "type" are specified.
+
+Documentation for the batman-adv protocol can be found at the following
+locations:
+
+https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/networking/batman-adv.rst
+https://www.open-mesh.org/
+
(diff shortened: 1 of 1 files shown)


NeoRaider:
It seems that the upstream PR needs to be rebased again...

T-X:
Changelog v2:
* updated to OpenWrt 22.03 / libpcap 0.10.1


> It seems that the upstream PR needs to be rebased again...

Thanks for the remark. Had updated the upstream PR on the 26th January right after your remark. (Still no one at libpcap/tcpdump has merged it yet though, even after another half a year gone...)

T-X:
Changelog v3:
* rebased to current Gluon master: moved libpcap patch from 0004-... to 0006-...

T-X:
Changelog v4:
* rebased to current Gluon master: moved libpcap patch from 0006-... to 0007-...

Djfe:
it's been 2 years since the upstream pr had any comment, what should we do?

AiyionPrime:
We wait. @T-X resolved the issues partially by getting it merged into OpenWrt;

and "two years no comment" is not true; I commented in late april and @fxlb rebases the upstream PR on a regular basis and requested reviews; the last one was only a week ago.

Maybe he has more insights, why the PR is so troublesome, but I'd say they are at least partially working on it and there is no need for us to nag about it.

Djfe:
I read it as April 2021 🙈

herbetom:
The Patch seems to have been upstreamed and is available with openwrt-23.05 and therefore Gluon master: https://github.com/openwrt/openwrt/blob/openwrt-23.05/package/libs/libpcap/patches/300-Add-support-for-B.A.T.M.A.N.-Advanced.patch

PR #2367, opened 2022-01-01
gluon-statistics-mcast: adding initial package

T-X:
This package collects disected statistics on multicast
and management traffic and provides them via respondd in
"statistics-extended". Such information can be helpful
for further development and comparison of mesh routing
protocols and troubleshooting network issues.

The statistics gathered disect into batman-adv
packets as well. Statistics are collected both on client
bridge ports and br-client as well as mesh interfaces.

This introduces a performance penalty but tries to
keep the impact on unicast traffic small.

Note that respondd replies are rather large
(several UDP datagrams, even when LZMA compressed).
Therefore you might want to use slower intervals for 
querying respondd "statistics-extended".

---

This PR requires / depends on the libpcap batman-adv support patch provided in a separate PR here: https://github.com/freifunk-gluon/gluon/pull/2366


T-X:
Changelog v2:
* adding some more double quotes to setup and teardown shell scripts to make the linter happy

Note: The linter still complains about the following:

> package/gluon-statistics-mcast/files/lib/gluon/core/mesh/teardown.d/40-bpfcountd:3:7: warning: Possible misspelling: IFNAME may not be assigned, but ifname is. [SC2153]

Which is bogus, that's exactly what this line there is supposed to check: https://github.com/freifunk-gluon/gluon/pull/2367/files#diff-9c405f8600368ddf185bf7fee9dfef42fd93d0ad9009c1919c1a0bf898308c3aR3. And the linter seems to get confused about these two distinct variables, the global "IFNAME" vs. the function local "ifname".

T-X:
Changelog v3:
* renamed a shell script variable from "ifname" to "uci_ifname" to work around the following, wrong linter warning:

> package/gluon-statistics-mcast/files/lib/gluon/core/mesh/teardown.d/40-bpfcountd:3:7: warning: Possible misspelling: IFNAME may not be assigned, but ifname is. [SC2153]

T-X:
Changelog v4:

* Added more protocols to track for their multicast packets:
  * [Syncthing Local Peer Discovery](https://docs.syncthing.net/specs/localdisco-v4.html#localdisco-v4)
  * OSPF for IPv6 ([RFC5340](https://datatracker.ietf.org/doc/html/rfc5340#section-4.2)) - before only IPv4 was tracked 
  * [usteer](https://git.openwrt.org/?p=project/usteer.git;a=shortlog)
  * l3roamd
  * mmfd
  * MANET ([RFC5498](https://datatracker.ietf.org/doc/html/rfc5498#section-3), as used by OLSRv2 [RFC7181](https://datatracker.ietf.org/doc/html/rfc7181#section-5.1) )
* Fixed gluon-bpfcountd cleanup for a wifi AP interfaces
  * before tried to find bridge ports of br-client via /etc/config/network/<client>.ifname, which would leave out the wifi AP interface(s)) -> fixing this by checking current bridge ports and storing it for later cleanup on ifdown (on ifdown, the bridge is already gone in hotplug script, so we need to store that information in advance)
* Added mmfd0 interface tracking support
* Added ip(6) protocols MANET and OSPF to allowed unicast pre-filter (PREFILTER_IP{4,6}_UC)

Tested in BABEL VMs that babel/l3roamd/mmfd multicast packets are tracked fine (note that for instance mmfd seems to forward via individual unicast transmissions? - so for overall protocol overhead evaluation the multicast packet counts might not be complete for BABEL networks).

NeoRaider:
The bpffilter package could be added to https://github.com/freifunk-gluon/packages instead. It might also make sense to submit it for https://github.com/openwrt/packages, it looks quite useful. Please also add a link to the tool's README to the Makefile's description field.

package/gluon-statistics-mcast/files/lib/gluon/core/mesh/setup.d/40-bpfcountd
```diff
@@ -0,0 +1,63 @@
...
+	set bpfcountd.${SECTION_IN}.prefilter="inbound and (${PREFILTER})"
+	set bpfcountd.${SECTION_IN}.filterfile='/lib/gluon/bpfcountd/mesh.filters'
+	set bpfcountd.${SECTION_IN}.disabled='0'
+
+	set bpfcountd.${SECTION_OUT}.ifname=${IFNAME}
+	set bpfcountd.${SECTION_OUT}.prefilter="outbound and (${PREFILTER})"
+	set bpfcountd.${SECTION_OUT}.filterfile='/lib/gluon/bpfcountd/mesh.filters'
+	set bpfcountd.${SECTION_OUT}.disabled='0'
+
+	commit bpfcountd
```
NeoRaider, line 54: Writing files to flash from `setup.d` scripts should be avoided. Better options would be to either run bpfcountd directly without procd, or to create a separate initscript that uses configuration stored in `/var` (with or without UCI).

package/gluon-statistics-mcast/src/respondd.c
```diff
@@ -0,0 +1,478 @@
...
+#include <sys/types.h>
+#include <sys/un.h>
+#include <unistd.h>
+
+#define BPFCOUNTD_SOCKS_DIR "/var/run/bpfcountd/gluon-sockets"
+
+enum capdir {
+	CAP_IN,
+	CAP_OUT,
+	__CAP_COUNT,
```
NeoRaider, line 46: All identifiers starting with `__` or `_`+uppercase letter are reserved for the compiler and libc and should never be defined in program code.

package/gluon-statistics-mcast/src/respondd.c
```diff
@@ -0,0 +1,478 @@
...
+	ret = parse_socket_iface(socketname, *dir, iface);
+	if (ret < 0)
+		return NULL;
+
+	return get_json_new(obj, iface);
+}
+
+static int parse_num(const char *start, int64_t *num)
+{
+	unsigned long long __num;
```
NeoRaider, line 176: Another reserved identifier.

package/gluon-statistics-mcast/src/respondd.c
```diff
@@ -0,0 +1,478 @@
...
+	/* limit to int64 max for json-c */
+	if (__num > INT64_MAX)
+		return -EINVAL;
+
+	*num = (int64_t)__num;
+	return 0;
+}
+
+static int add_num_to_json(struct json_object *dir_obj, const char *key,
+			  int64_t *num)
```
NeoRaider, line 196: Why is `num` passed as a pointer here?

package/gluon-statistics-mcast/src/respondd.c
```diff
@@ -0,0 +1,478 @@
...
+	CAP_OUT,
+	__CAP_COUNT,
+};
+
+static const char *dirstr[__CAP_COUNT] = {
+	[CAP_IN] = "in",
+	[CAP_OUT] = "out",
+};
+
+static int check_socket_suffix(char *socketname)
```
NeoRaider, line 54: This should use `const char *`. There are lots more arguments that should be `const` in the code.

package/gluon-statistics-mcast/src/respondd.c
```diff
@@ -0,0 +1,478 @@
...
+
+	if (chdir(BPFCOUNTD_SOCKS_DIR) < 0)
+		return obj;
+
+	dir = opendir("./");
+	if (!dir)
+		return obj;
+
+	while ((dp = readdir(dir))) {
+		if (!strcmp(dp->d_name, ".") || !strcmp(dp->d_name, ".."))
```
NeoRaider, line 407: Maybe just skip all files starting with `.`?

package/gluon-statistics-mcast/src/respondd.c
```diff
@@ -0,0 +1,478 @@
...
+	/* skip UCI section name */
+	iface_start = strchr(socketname, '.');
+	if (!iface_start || iface_start[1] == '\0')
+		return -EINVAL;
+
+	iface_start += 1;
+	iface_end = socketname + strlen(socketname);
+
+	if (dir == CAP_IN)
+		iface_end -= strlen(".in.sock");
```
NeoRaider, line 112: The way this code deals with the filename suffixes feels very fragile... it would be great it we could split there strings in a way that didn't require pointer arithmetic.

NeoRaider (changes requested):
This is only a very superficial review - I'd like to have a closer look soon.

NeoRaider:
> Note that respondd replies are rather large
> (several UDP datagrams, even when LZMA compressed).
> Therefore you might want to use slower intervals for
> querying respondd "statistics-extended".

Is respondd really the right tool to provide this data? UDP is lossy, even more so when many nodes send long, fragmented packets more or less at the same time... How long does the data generation take? I'm wondering if the data might be provided through uhttpd instead...

package/gluon-statistics-mcast/files/lib/gluon/core/mesh/setup.d/40-bpfcountd
T-X: It is actually stored in /var, more precisely in /var/run/bpfcountd/gluon-config :-). See UCI_CONFIG_DIR.

package/gluon-statistics-mcast/src/respondd.c
T-X: Oh, didn't know about that. Are you sure? Because I see it in so many places for enums, like in netifd (__DEV_ATTR_MAX,  __BOND_ATTR_MAX, __BRIDGE_ATTR_MAX, ...). Or even in the Linux kernel (__NL80211_ATTR_AFTER_LAST, __IFLA_BRIDGE_VLAN_TUNNEL_MAX, ...).

T-X:
> Is respondd really the right tool to provide this data? UDP is lossy, even more so when many nodes send long, fragmented packets more or less at the same time... How long does the data generation take? I'm wondering if the data might be provided through uhttpd instead...

Yes, had been wondering about this, too, relying on fragments is not really nice... At least we have support for delayed responses since https://github.com/freifunk-gluon/packages/commit/e7fbcf75a95da3c39a3a7a85a2da481c3bf7d93d.

For an HTTP / uhttpd provider I was wondering whether we should put more thought on a unified format before adding that. As it might be interesting to add other detailed statistics later for all kinds of data. Or maybe we might want to move the existing statistics provided in respondd to HTTP at some point, too?

An alternative I was thinking about:

Adding TCP support to respondd. For all the respondd providers we have so far a respondd querier should of course keep using UDP. But then for more verbose providers, like statistics-extended, a respondd querier could then just cherry-pick it via a unicast+TCP query from the few, specific nodes which offer this provider. We could add a list of available respondd providers to the "nodeinfo" provider, so that a respondd statistics-extended querier knows in advance which few nodes it can query for it via unicast/TCP.

This should basically be a similar packet overhead as HTTP in the end?

package/gluon-statistics-mcast/src/respondd.c
T-X: I like that idea! Will change that.

package/gluon-statistics-mcast/src/respondd.c
NeoRaider: In the Linux kernel it's not strictly incorrect, as the kernel is basically its own libc and can therefore use the reserved namespace. For userspace code, it is definitely incorrect according to the C standard (see https://en.cppreference.com/w/c/language/identifier for a summary of the rules).

package/gluon-statistics-mcast/src/respondd.c
T-X: Urgh, I just realized how evil strchr() is. Why the hell can it return a "char *" from a "const char *"??

For socketname it seems okay to change it to const (as long as I'm not missing something). But with parse_line() I probably shouldn't change "char *line" to "const char *line", as there is a `*start = '\0';` in there, which modifies `*line`.

T-X:
Or another, easy way would be to add a small, generic wrapper to http://.../cgi-bin/respondd. Which could receive a respondd query encoded as a HTTP POST. And cgi-bin-respondd could proxy it as UDP to "[::1]:1001", which could take advantage of the higher, 65536 bytes MTU on the "lo" interface. And the UDP reply from [::1]:1001 would then be sent back via HTTP in one or more TCP segments.

T-X:
Changelog v5:

* rebased to current master (no changes)
* adding Reticulum default multicast peer discovery port 29716 (https://github.com/markqvist/Reticulum, https://github.com/markqvist/Reticulum/blob/master/RNS/Interfaces/AutoInterface.py#L12)
* removing "bpfcountd." prefix from socket files for simplified parsing
* changing __CAP_COUNT to CAP_MAX, adding CAP_UNDEF for error return code
* changing "unsigned long long __num" to "unsigned long long value"
* removing some pointers for int64_t bytes/packets to function passing
* adding some "const" for socketname to function passing
* adding/updating some code comments
* instead of skipping "." and "..", skip all files/directories starting with "."
* moved "check_socket_suffix()" check
* simplified parse_socket_dir() by using regex.h
* simplified parse_socket_iface() by removing iface_{start,end} pointers

Still ToDo:

* more reliable transfer way than UDP via IP fragments
 * also remeasured statistics-extended message sizes on an actual 2.4+5ghz node, which has more interfaces than my previous VM tests:

| compression algorithm | size |
| ----------- | ------------- |
| uncompressed | 416 kB |
| deflate compressed | 21 kB |

With xz instead of deflate the following results would in theory be possible:

| compression algorithm | size |
| ----------- | ------------- |
| xz (-6 / default) compressed | 5.7 kB |
| xz -6e compressed | **5.2 kB** |
| xz -9 compressed | 5.7 kB |
| xz -9e compressed | **5.2 kB** |
| xz -3 compressed | 6.2 kB |
| xz -3e compressed | **5.2 kB** |
| xz -0 compressed | 9.2 kB |
| xz -0e compressed | **5.2 kB** |

T-X:
Changelog v6:
* adding commit "gluon-status-page: add generic respondd cgi-bin"

This allows to query bulky respondd providers like the "statistics-extended" introduced in this PR over HTTP instead of UDP, to avoid retrieving the UDP data in multiple IP fragments.

Together with https://github.com/freifunk-gluon/packages/pull/256 this would allow a respondd querier to retrieve the multicast statistics as follows:

```
Every n minutes:
  send UDP respondd query "GET nodeinfo respondd-providers" -> [ff05::2:1001]:1001

  for each reply:
    if reply:respondd-providers->statistics-extended:
      HTTP GET "http://<nodeip>/cgi-bin/respondd?statistics-extended"
```

T-X:
Changelog v7:
* rebased to master (no changes)
* fixed a typo in the label for RETICULUM (vs. RETICULIM) protocol
* added OLSRv1 UDP port (698) for counting (OLSRv2 already present under the MANET labels)

T-X:
Changelog v8:
* use colon separator for BLA, too (BLA-CLAIM, BLA-UNCLAIM, etc. to BLA:UNCLAIM, BLA:UNCLAIM, etc.)

T-X:
Changelog v9:
* rebased to current master (no changes/conflicts)
* bpfcountd: updating bpfcountd to add support for libpcap buffer size configuration option, adding buffersize option to init script
* gluon-statistics-mcast: reducing libpcap buffer size from 2MB (default) to 128KB -> reduces overall memory usage of gluon-statistics-mcast to ~13MB, now all bpfcountd instances start fine on a 64MB device like the the Gl.inet GL-USB150 without causing an instant out-of-memory

package/bpfcountd/Makefile
```diff
@@ -0,0 +1,34 @@
+include $(TOPDIR)/rules.mk
```
blocktrron, line 1: Move this package to OpenWrt packages (I See that LICENCE identifies is missing).

T-X:
Changelog v11:
* fix gluon-statistics-mcast socket symlink in setup script (regression introduced in v10)
* explicitly use /bin/ash in gluon-statistics-mcast setup and teardown scripts to make linter happy about string replacements

T-X:
Changelog v12:
* workaround for linter did not work, so reverting back to "!/bin/sh" and instead adding a new commit with an exception to silence linter

AiyionPrime:
```console
package/gluon-statistics-mcast/files/etc/hotplug.d/iface/40-bpfcountd:
	2: Wrong indentation type(spaces instead of tabs)
	3: Wrong indentation type(spaces instead of tabs)
	5: Wrong indentation type(spaces instead of tabs)
	58: Wrong indentation type(spaces instead of tabs)
	61: Wrong indentation type(spaces instead of tabs)
	64: Wrong indentation type(spaces instead of tabs)
package/gluon-statistics-mcast/src/respondd.c:
	2: Wrong indentation type(spaces instead of tabs)
	3: Wrong indentation type(spaces instead of tabs)
	5: Wrong indentation type(spaces instead of tabs)
	6: Wrong indentation type(spaces instead of tabs)
	8: Wrong indentation type(spaces instead of tabs)
	9: Wrong indentation type(spaces instead of tabs)
	10: Wrong indentation type(spaces instead of tabs)
	11: Wrong indentation type(spaces instead of tabs)
	12: Wrong indentation type(spaces instead of tabs)
	14: Wrong indentation type(spaces instead of tabs)
	15: Wrong indentation type(spaces instead of tabs)
	16: Wrong indentation type(spaces instead of tabs)
	17: Wrong indentation type(spaces instead of tabs)
	18: Wrong indentation type(spaces instead of tabs)
	19: Wrong indentation type(spaces instead of tabs)
	20: Wrong indentation type(spaces instead of tabs)
	21: Wrong indentation type(spaces instead of tabs)
	22: Wrong indentation type(spaces instead of tabs)
	23: Wrong indentation type(spaces instead of tabs)
	100: Wrong indentation type(spaces instead of tabs)
	146: Wrong indentation type(spaces instead of tabs)
	147: Wrong indentation type(spaces instead of tabs)
	173: Wrong indentation type(spaces instead of tabs)
	185: Wrong indentation type(spaces instead of tabs)
	229: Wrong indentation type(spaces instead of tabs)
	261: Wrong indentation type(spaces instead of tabs)
	319: Wrong indentation type(spaces instead of tabs)
	377: Wrong indentation type(spaces instead of tabs)
package/gluon-status-page/files/lib/gluon/status-page/www/cgi-bin/respondd:
	10: Wrong indentation type(spaces instead of tabs)
	12: Wrong indentation type(spaces instead of tabs)
	15: Wrong indentation type(spaces instead of tabs)

37 errors found

```

package/gluon-statistics-mcast/files/etc/hotplug.d/iface/40-bpfcountd
```diff
@@ -0,0 +1,68 @@
+case "$INTERFACE" in
+  "client"|\
+  "mmfd")
+	;;
+  *)
+	exit 0
+	;;
```
AiyionPrime, line 7:
```suggestion
	"client"|\
	"mmfd")
		;;
	*)
		exit 0
		;;
```

package/gluon-statistics-mcast/files/etc/hotplug.d/iface/40-bpfcountd
```diff
@@ -0,0 +1,68 @@
...
+
+case "$ACTION" in
+  "ifup")
+	bpfcountd_start
+	;;
+  "ifdown")
+	bpfcountd_stop
+	;;
+  *)
+	;;
```
AiyionPrime, line 65:
```suggestion
	"ifup")
		bpfcountd_start
		;;
	"ifdown")
		bpfcountd_stop
		;;
	*)
		;;
```

package/gluon-status-page/files/lib/gluon/status-page/www/cgi-bin/respondd
```diff
@@ -0,0 +1,35 @@
...
+
+case "${REQUEST_METHOD}" in
+  GET)
+	;;
+  POST)
+	QUERY_STRING="$(cat)"
+	;;
+  *)
+	badrequest
+	;;
```
AiyionPrime, line 17:
```suggestion
	GET)
		;;
	POST)
		QUERY_STRING="$(cat)"
		;;
	*)
		badrequest
		;;
```

package/gluon-statistics-mcast/src/respondd.c
```diff
@@ -0,0 +1,491 @@
...
+  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+  IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+  DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+  FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+  DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+  SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+  OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
```
AiyionPrime, line 24:
```suggestion
/* SPDX-FileCopyrightText: 2021, Linus Lüssing <linus.luessing@c0d3.blue> */
/* SPDX-License-Identifier: BSD-2-Clause */
```

package/gluon-statistics-mcast/src/respondd.c
```diff
@@ -0,0 +1,491 @@
...
+	if (!regexecmp("\\.in\\.sock$", socketname))
+		return CAP_IN;
+	else if (!regexecmp("\\.out\\.sock$", socketname))
+		return CAP_OUT;
+	else
+		return CAP_UNDEF;
+}
+
+static int parse_socket_iface(const char *socketname, enum capdir dir,
+			      char *iface)
```
AiyionPrime, line 100:
```suggestion
		char *iface)	      
```

package/gluon-statistics-mcast/src/respondd.c
```diff
@@ -0,0 +1,491 @@
...
+		return -EINVAL;
+
+	memset(iface, 0, IFNAMSIZ);
+	strncpy(iface, socketname, iface_len);
+
+	return 0;
+}
+
+static struct json_object *get_json_new(struct json_object *parent,
+					const char *key)
```
AiyionPrime, line 121:
```suggestion
		const char *key)
```

package/gluon-statistics-mcast/src/respondd.c
```diff
@@ -0,0 +1,491 @@
...
+
+	/* json-c 0.12, should check return codes for >= 0.13 */
+	json_object_object_add(parent, key, child);
+
+	return child;
+}
+
+static struct json_object *parse_socket_name(const char *socketname,
+					     struct json_object *obj,
+					     enum capdir *dir)
```
AiyionPrime, line 147:
```suggestion
		struct json_object *obj,
		enum capdir *dir)
```

package/gluon-statistics-mcast/src/respondd.c
```diff
@@ -0,0 +1,491 @@
...
+{
+	unsigned long long value;
+	char *endptr;
+
+	if (*start == '-')
+		return -EINVAL;
+
+	value = strtoull(start, &endptr, 10);
+	if ((value == ULLONG_MAX && errno == ERANGE) ||
+	    start == endptr)
```
AiyionPrime, line 173:
```suggestion
			start == endptr)
```

package/gluon-statistics-mcast/src/respondd.c
```diff
@@ -0,0 +1,491 @@
...
+	/* limit to int64 max for json-c */
+	if (value > INT64_MAX)
+		return -EINVAL;
+
+	*num = (int64_t)value;
+	return 0;
+}
+
+static int add_num_to_json(struct json_object *dir_obj, const char *key,
+			   int64_t num)
```
AiyionPrime, line 185:
```suggestion
		int64_t num)
```

package/gluon-statistics-mcast/src/respondd.c
```diff
@@ -0,0 +1,491 @@
...
+		goto err;
+
+	return dir_obj;
+err:
+	json_object_put(dir_obj);
+	return NULL;
+}
+
+static int add_line_to_json(struct json_object *iface_obj, enum capdir dir,
+			    const char *rule, int64_t bytes, int64_t packets)
```
AiyionPrime, line 229:
```suggestion
		const char *rule, int64_t bytes, int64_t packets)
```

package/gluon-statistics-mcast/src/respondd.c
```diff
@@ -0,0 +1,491 @@
...
+	return -ENOMEM;
+}
+
+/* Parses a line of the following format:
+ * <rulename>:<bytes>:<packets>
+ *
+ * And adds it to the provided json object.
+ */
+static int parse_line(struct json_object *obj, enum capdir dir,
+		      char *line)
```
AiyionPrime, line 261:
```suggestion
		char *line)
```

package/gluon-statistics-mcast/src/respondd.c
```diff
@@ -0,0 +1,491 @@
...
+	sd = socket(AF_UNIX, SOCK_STREAM, 0);
+	if (sd < 0)
+		return sd;
+
+	memset(&addr, 0, sizeof(addr));
+	addr.sun_family = AF_UNIX;
+	strncpy(addr.sun_path, socketname, sizeof(addr.sun_path) - 1);
+
+	ret = connect(sd, (const struct sockaddr *)&addr,
+		      sizeof(addr));
```
AiyionPrime, line 319:
```suggestion

			sizeof(addr));
```

package/gluon-statistics-mcast/src/respondd.c
```diff
@@ -0,0 +1,491 @@
...
+		struct json_object *iface_obj;
+		enum capdir capdir;
+
+		/* skip emtpy names, hidden files, "." and ".." */
+		if (dp->d_name[0] == '\0' || dp->d_name[0] == '.')
+			continue;
+
+		/* skip files not ending in ".sock" */
+		if (check_socket_suffix(dp->d_name) < 0)
+		       continue;
```
AiyionPrime, line 377:
```suggestion
			continue;
```

AiyionPrime (changes requested):
Not a complete review, but only the results of me checking for affected PRs in #2765.

rotanid:
@T-X no activity in over two years, feel free to re-open once you continue.

T-X:
@rotanid thanks for the reminder! And thanks for the partial review @AiyionPrime, had forgotten about it.

Generally I'm still interested in this. I think something like this, some detailed management and layer 2 specific traffic monitoring, would be important to have in Gluon to better and objectively compare different approaches, like @SmithChart's gluon-parker or @mkg20001's gluon-olsr (or babel). Or to evaluate future enhancements (or regressions) in batman-adv.

But I won't continue working on this immediately, went for manual tcpdump's and gnuplot'ing for now for my batman-adv enhancements.

Also another point was that I think @neocturne would prefer an alternative approach to the HTTP wrapper I had around respondd. Instead wanted a respondd native HTTP (or at least TCP) support, if i recall correctly.

PR #2372, opened 2022-01-09
Regenerate network and system UCI configs on every reconfigure, switch to role-based interface configuration

NeoRaider:
~~This PR is now based on #2382 (therefore blocked)~~

- lan_ifname, wan_ifname and setup_ifname are reset on every reconfigure
  - This makes it impossible to change these manually; instead, the roles of these logical interfaces can be modified on a higher level (see below). Changing the config mode interface (setup_ifname) is not possible.
  - When only one interface exists, it is called single_ifname, and lan_ifname/wan_ifname are unset to allow separate configuration via site.conf for this case
- Generate fresh network config
  - Settings that are configurable in the Web UI are preserved, + minor additions
- Generate fresh system config
  - Only the "system" section (hostname, remote syslog), NTP servers and GPIO (PoE passthrough) settings are preserved
- Interface configuration is now role-based. A list of roles ("client", "uplink", "mesh") can be assigned to each of the logical interfaces (LAN/WAN or single). See the individual commit messages, in particular "gluon-core: initialize interfaces role configuration" and "treewide: use interface roles as basis for network configuration", for details
- Modifying port assignment on swconfig-based devices is not possible at the moment (preserve flag only allows adding new custom sections, not modifying existing ones)

TODO:
- ~~Preserve manually added sections if they have a 'gluon_preserve' marker option~~
- ~~Migrate old configuration to role-based settings~~
- ~~Much more testing~~
- ~~Documentation~~
- ~~Blocked: Rebase after #2395~~

To test:
- Migrations from 2021.1.x
  - [x] Correct migration of Mesh-on-LAN/WAN to mesh role on 2-interface devices
  - [x] Correct migration of Mesh-on-WAN on 1-interface devices
  - [x] Correct migration of Mesh-on-LAN on 1-interface devices that were originally set up with `single_as_lan = true`
- [x] Network config works for DSA-based devices
- [x] Network config works for swconfig-based devices
- Other things?


mkg20001:
This kills the loopback interfacce

NeoRaider:
> This kills the loopback interfacce

Oops, fixed.

NeoRaider:
Proposal for new site.conf settings for default interface assignment, superseding `mesh_on_wan`, `mesh_on_lan` and `single_as_wan`:
```lua
	interfaces = {
		lan = {
			default_roles = { 'client' },
		},
		wan = {
			default_roles = { 'uplink', 'mesh' },
		},
		single = {
			default_roles = { 'uplink', 'mesh' },
		},
	},
```
`single` will default to the same setting used for `wan`, but can be configured independently.

NeoRaider:
Migration of Mesh-on-WAN/LAN setting from the old configuration is implemented now.

I've also added a simple migration for single -> WAN and vice-versa to account for added and removed interfaces. This is currently not fully working without an upgrade, as `/etc/board.json` is not regenerated by default. We could account for this unusual case by regenerating `board.json` on every reconfigure.

AiyionPrime:
http://lilienstra-e.n.ffh.zone/ runs the current version of the PR.
mesh-on-lan does apparently still work, gluon-reconfigure does not break the device and dmesg does not throw obvious errors.

Wireless mesh is disabled as it has been before the update.

https://hannover.freifunk.net/karte/#/en/map/b44bd6202778

This appears not to break a device that has nothing to migrate, which is nice.
What else should be tested?

NeoRaider:
The PR description now contains a list of things to test. I have two devices for tests that actually need migration of interface names (a TP-LINK CPE510 v1 and a D-Link DIR-860L), but I don't think I have a single-interface device like that. Might be sufficient to test with x86 VMs, as `/etc/config/network` is reset anyways, so device-specific migration issues are unlikely (and if there are device-specific issues, they should also happen without migration).

I have also pushed a small change that should handle migration of the Mesh-on-LAN/WAN flag more correctly for single-interface devices.

AiyionPrime:
Thanks for the list.
I'll give it a try, but not before monday.
If I'd like to test a device, do I need further device specific migration files?

NeoRaider:
> If I'd like to test a device, do I need further device specific migration files?

No, as the migration works by throwing away most of the config and then repeating the initial setup, no additional device-specific code is necessary for the migration.

AiyionPrime:
Given the limited amount of unused devices I can offer tests on a CPE210 v1.1.

After upgrading from vH21 (v2020.1.3) it appears lan and wan are swapped.

Other than that I do not see errors yet;
maybe there's an option for swapping I overlooked, that has to be added to the device section?

NeoRaider:
> Other than that I do not see errors yet; maybe there's an option for swapping I overlooked, that has to be added to the device section?

CPE210 and 510 v1 need to be added to `package/gluon-core/luasrc/lib/gluon/upgrade/020-interfaces` again. I forgot about that - it doesn't only affect upgrades from older versions, but also clean installations of the new version.

AiyionPrime:
Thanks, for the hint; I forgot that file, but sure, you're right.

AiyionPrime:
> This PR is now based on #2382 (therefore blocked)

Might want to get rid of the notice by now.

AiyionPrime:
> CPE210 and 510 v1 need to be added to `package/gluon-core/luasrc/lib/gluon/upgrade/020-interfaces` again.

I did this in https://github.com/freifunk-gluon/gluon/pull/2389/commits/b7c5870b46dd921610d1d2992b2f81100fccac00.
The running result is reachable under http://[2a02:790:ff:514:1aa6:f7ff:feb0:4662]/cgi-bin/status .

The migration appears to work mostly as intended, though the configmode mode port was the "wrong" port after flashing the new firmware using `sysupgrade -n`.

AiyionPrime:
Fuck, got it. Will mark it as outdoor device...

AiyionPrime:
I created a PR-draft for demonstration purposes, the running result is reachable under http://[2a02:790:ff:514:1aa6:f7ff:feb0:4662]/cgi-bin/status .

The migration appears to work as intended.
`dmesg` and `lastlog` are inconspicuous.

package/gluon-mesh-vpn-wireguard/luasrc/lib/gluon/upgrade/400-mesh-vpn-wireguard
```diff
@@ -3,7 +3,7 @@
 local uci = require('simple-uci').cursor()
 local site = require 'gluon.site'
 
-local private_key = uci:get("network", 'wg_mesh', "private_key")
+local private_key = uci:get("network_gluon-old", 'wg_mesh', "private_key")
 
 if not private_key or not private_key:match("^" .. ("[%a%d+/]"):rep(42) .. "[AEIMQUYcgkosw480]=$") then
```
lemoer, line 8: The wireguard private key is lost if the firmware is upgraded to a firmware without the package `gluon-mesh-vpn-wireguard`. I am not yet sure, if this is a problem for us, but I thought, it might be worth noting.

lemoer:
I just did a successful migration of my TP-Link TL-WR1043ND v3.

package/gluon-mesh-vpn-wireguard/luasrc/lib/gluon/upgrade/400-mesh-vpn-wireguard
NeoRaider: Interesting. I don't think it's critical to fix this right now - after all, an upgrade on a node that never had `gluon-mesh-vpn-wireguard` in the first place would lead to a similar situation. A possible solution would be to move the private key to `/etc/config/gluon`.

lemoer: > an upgrade on a node that never had gluon-mesh-vpn-wireguard in the first place would lead to a similar situation.

I do not understand? Of course, a node that never had `gluon-mesh-vpn-wireguard` does not have a wireguard key. So there is nothing to loose, but in which regard is this similar to a node that looses it's key?

NeoRaider: Well, a site going from firmwares (wireguard) -> (no wireguard) -> (wireguard) would be very unusual (as Gluon generally doesn't support downgrades), but if such a thing actually happened, the site would already need to have some solution in place for nodes that were first installed with the (no wireguard) firmware - either by requiring node owners to return to the config mode to register their key, or by having some kind of auto registration.

So this would just mean that every node going through the (no wireguard) -> (wireguard) migration will have to go through the same steps, regardless of whether it had a firmware with wireguard in the past or not. I think that's good enough for this unusual case.

lemoer: Ok.

lemoer: In general, I would like to see this moved to /etc/config/gluon.

As an upgrade with `sysupgrade` happens silently (without a config mode that shows you your new vpn key) this may remain unnoticed by the user and the user might start to wonder, why his vpn doesn't work anymore.

I agree, that this will not happen often most likely, but it might happen. As the fix is easy, I would suggest to go for it.

NeoRaider: Well, cleaning up behind packages that aren't installed anymore could also be seen as a feature - actually one that I'd like to have in all packages eventually.

NeoRaider:
@AiyionPrime @lemoer Thanks for your testing. I also finished my tests with qemu (x86-64), CPE510v1 and DIR-860L. DIR-860L will need some more work (#2391), but that doesn't need to block this PR.

lemoer:
Some notes here:
- On swconfig systems, the switch configuration can no longer be modified manually by users.
    - This is due to the fact that /etc/config/network will be regenerated.
    - The preserve flag cannot be applied to sections that OpenWrt created initially. 
    - We do not really care, since we hope that swconfig will phase out in favor of DSA.
- After this, we now have a stable way to support adding roles to vlan interfaces.
    - This is a feature long requested by some people.
    - Webinterface for that feature could be added later.

belzebub40k:
I've did some basic testing with the Virtualbox image and noticed two problems in the Config-Mode.

Mesh-Interfaces can not be activated individually. Judging by the ids in the checkboxes it looks like the setup is detected a device with only one port.



When saving the config only a page with this message is shown instead of the usual message showing the info text and VPN-key.



Site: https://github.com/freifunk-mwu/site-ffmwu/tree/uci-regen
Images: https://firmware.freifunk-mwu.de/_library/2022.1%2Bmwu~exp20220202~uci1/

NeoRaider:
> I've did some basic testing with the Virtualbox image and noticed two problems in the Config-Mode.
> 
> Mesh-Interfaces can not be activated individually. Judging by the ids in the checkboxes it looks like the setup is detected as device with only one port.
> 
> 
> 
> When saving the config only a page with this message is shown instead of the usual message showing the info text and VPN-key.
> 
> 
> 
> Site: https://github.com/freifunk-mwu/site-ffmwu/tree/uci-regen Images: https://firmware.freifunk-mwu.de/_library/2022.1%2Bmwu~exp20220202~uci1/

Should be fixed now.

belzebub40k:
> Should be fixed now.

The problem with the selection of the mesh interfaces is fixed but the "Bad Gateway" message still shows up when clicking save & reboot.

NeoRaider:
> > Should be fixed now.
> 
> The problem with the selection of the mesh interfaces is fixed but the "Bad Gateway" message still shows up when clicking save & reboot.

Fixed as well now. And now I'm confused why this worked at all in Gluon 2021.1.x...

package/gluon-core/files/lib/gluon/upgrade/001-reset-uci
```diff
@@ -0,0 +1,20 @@
...
+if [ -s "$NETWORK_CFG" ] && ! [ -s "$NETWORK_SAVED" ]; then
+	mv -f "$NETWORK_CFG" "$NETWORK_SAVED"
+fi
+if [ -s "$SYSTEM_CFG" ] && ! [ -s "$SYSTEM_SAVED" ]; then
+	mv -f "$SYSTEM_CFG" "$SYSTEM_SAVED"
+fi
+
+# Generate a new network config
+rm -f /etc/board.json "$NETWORK_CFG" "$SYSTEM_CFG"
+config_generate
```
lemoer, line 20: This command doesn't seem to be silent... 

```
root@wgtest-1043-lemoer:~# gluon-reconfigure 
Configuring: 001-reset-uci
cfg030f15
cfg073777
cfg081ec7
cfg091ec7
cfg01e48a
Configuring: 002-migrate-system
Configuring: 005-set-domain
Configuring: 005-wireless-migration
...
```

Can we silence it?

belzebub40k:
> > > Should be fixed now.
>>
> > The problem with the selection of the mesh interfaces is fixed but the "Bad Gateway" message still shows up when clicking save & reboot.
>
> Fixed as well now. And now I'm confused why this worked at all in Gluon 2021.1.x...

Strange but I can confirm it is fixed now.

package/gluon-core/files/lib/gluon/upgrade/001-reset-uci
NeoRaider: We can redirect the output to `/dev/null`. I'm not sure if the script ever prints something interesting that might help with debugging though.

belzebub40k:
I noticed a new problem. When using this branch I'm not able to change the hostname via the config-mode but pretty-hostname on the shell works. I did not notice any other options which are not saved. I tried to reproduce the problem with the master branch but there it is working fine.

AiyionPrime:
Starting a build for the CPE210 now, @belzebub40k sorry I effectively put the hostname-discussion in the wrong PR.

NeoRaider:
Rebased and unblocked again.

lemoer:
What is the current state here? How are we going to proceed?

NeoRaider:
From my side this is ready to merge.

blocktrron (approved):

AiyionPrime:
Thanks!

PR #2373, opened 2022-01-09
gluon-core: add gluon info binary

mkg20001:
This copies the code from web-admin and uses it to create a neat 
cli-accessible summary about a node

This could also be extended or possibly have all the data the status 
page has (the one when the node is no longer in config mode9

Example

```
Hostname				gluon-b0a7b902343c
MAC address				b0:a7:b9:02:34:3c
Hardware model				TP-Link CPE210 v3
Gluon version / Site version		v2021.1-309-g9085237 / 8ed7e6a
Firmware release			0.0+exp20220109
Site					Funkfeuer Graz
Public VPN key				nil
```
(tabs display properly in terminal - in retrospect not sure if spaces would've been better)

### DIFF
--- a/package/gluon-core/luasrc/usr/bin/gluon-info
+++ b/package/gluon-core/luasrc/usr/bin/gluon-info
@@ -22,8 +22,8 @@ local values = {
 	{ 'Hostname', pretty_hostname.get(uci) },
 	{ 'MAC address', sysconfig.primary_mac },
 	{ 'Hardware model', platform.get_model() },
-	{ 'Gluon version' .. " / " .. 'Site version', util.trim(util.readfile('/lib/gluon/gluon-version'))
-		.. " / " .. util.trim(util.readfile('/lib/gluon/site-version')) },
+	{ 'Gluon version / Site version', util.trim(util.readfile('/lib/gluon/gluon-version'))
+		.. ' / ' .. util.trim(util.readfile('/lib/gluon/site-version')) },
 	{ 'Firmware release', util.trim(util.readfile('/lib/gluon/release')) },
 	{ 'Site', site.site_name() },
 	{ 'Public VPN key', pubkey or 'n/a' },
--- a/package/gluon-core/luasrc/usr/bin/gluon-info
+++ b/package/gluon-core/luasrc/usr/bin/gluon-info
@@ -1,0 +1,38 @@
+#!/usr/bin/lua
+
+local uci = require('simple-uci').cursor()
+local pretty_hostname = require 'pretty_hostname'
+
+local site = require 'gluon.site'
+local sysconfig = require 'gluon.sysconfig'
+local platform = require 'gluon.platform'
+local util = require 'gluon.util'
+local has_vpn, vpn = pcall(require, 'gluon.mesh-vpn')
+
+local pubkey
+if has_vpn and vpn.enabled() then
+	local _, active_vpn = vpn.get_active_provider()
+
+	if active_vpn ~= nil then
+		pubkey = active_vpn.public_key()
+	end
+end
+
+local values = {
+	{ 'Hostname', pretty_hostname.get(uci) },
+	{ 'MAC address', sysconfig.primary_mac },
+	{ 'Hardware model', platform.get_model() },
+	{ 'Gluon version' .. " / " .. 'Site version', util.trim(util.readfile('/lib/gluon/gluon-version'))
+		.. " / " .. util.trim(util.readfile('/lib/gluon/site-version')) },
+	{ 'Firmware release', util.trim(util.readfile('/lib/gluon/release')) },
+	{ 'Site', site.site_name() },
+	{ 'Public VPN key', pubkey or 'n/a' },
+}
+
+local padTo = 24
+
+for _, info in ipairs(values) do
+	local labelLen = string.len(info[1]) + 1
+
+	print(info[1] .. ':' .. string.rep(' ', padTo - labelLen), info[2])
+end
--- a/package/gluon-mesh-vpn-fastd/luasrc/usr/lib/lua/gluon/mesh-vpn/provider/fastd.lua
+++ b/package/gluon-mesh-vpn-fastd/luasrc/usr/lib/lua/gluon/mesh-vpn/provider/fastd.lua
@@ -7,7 +7,13 @@ local vpn_core = require 'gluon.mesh-vpn'
 local M = {}
 
 function M.public_key()
-	return util.trim(util.exec('/etc/init.d/fastd show_key mesh_vpn'))
+	local key = util.trim(util.exec('/etc/init.d/fastd show_key mesh_vpn'))
+
+	if key == '' then
+		key = nil
+	end
+
+	return key
 end
 
 function M.enable(val)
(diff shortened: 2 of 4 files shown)


blocktrron:
Hi, this looks like it's predestined for a community-package. Please File a PR in this Repository: https://github.com/freifunk-gluon/community-packages/

mkg20001:
Why tho? This is a mostly simple binary that helps with diagnostics and web-admin already displays the same data. I don't see how having it in core hurts. Or am I missing something?

blocktrron:
The script is not an extension of Gluons core functionalities and can be packages independently of Gluons Core. So it does not belong to `gluon-core`.

To help iterate such packages quicker (and reduce the burden of long feedback cycles), they should be developed in the community-packages. Open a PR there, after merge you will be granted with the necessary permissions to develop your packages within the community-packages repository.

mkg20001:
Does it make sense to put it into core at some point?

gluon-show-site has exactly the same criteria applying

It's use is also mostly diagnostic

Is the justification there because that was once done so and has stayed so?

NeoRaider:
I'd like to have the script in gluon-core.

package/gluon-core/luasrc/usr/bin/gluon-info
```diff
@@ -0,0 +1,52 @@
...
+    pubkey = active_vpn.public_key()
+  end
+
+  if pubkey == '' then
+    pubkey = nil
+  end
+end
+
+local values = {
+  { _('Hostname'), pretty_hostname.get(uci) },
```
NeoRaider, line 30: Let's remove the unnecessary `_` function.

package/gluon-core/luasrc/usr/bin/gluon-info
```diff
@@ -0,0 +1,52 @@
...
+  { _('MAC address'), sysconfig.primary_mac },
+  { _('Hardware model'), platform.get_model() },
+  { _('Gluon version') .. " / " .. _('Site version'), util.trim(util.readfile('/lib/gluon/gluon-version'))
+    .. " / " .. util.trim(util.readfile('/lib/gluon/site-version')) },
+  { _('Firmware release'), util.trim(util.readfile('/lib/gluon/release')) },
+  { _('Site'), site.site_name() },
+  { _('Public VPN key'), pubkey },
```
NeoRaider, line 37:
```suggestion
  { _('Public VPN key'), pubkey or 'n/a' },
```

package/gluon-core/luasrc/usr/bin/gluon-info
```diff
@@ -0,0 +1,52 @@
...
+local pubkey
+if has_vpn and vpn.enabled() then
+  local _, active_vpn = vpn.get_active_provider()
+
+  if active_vpn ~= nil then
+    pubkey = active_vpn.public_key()
+  end
+
+  if pubkey == '' then
+    pubkey = nil
```
NeoRaider, line 25: Hmm, it would be great if we could fix the VPN providers never to return an empty string. Maybe they even do that already?

package/gluon-core/luasrc/usr/bin/gluon-info
```diff
@@ -0,0 +1,52 @@
...
+  { _('Hardware model'), platform.get_model() },
+  { _('Gluon version') .. " / " .. _('Site version'), util.trim(util.readfile('/lib/gluon/gluon-version'))
+    .. " / " .. util.trim(util.readfile('/lib/gluon/site-version')) },
+  { _('Firmware release'), util.trim(util.readfile('/lib/gluon/release')) },
+  { _('Site'), site.site_name() },
+  { _('Public VPN key'), pubkey },
+}
+
+local expectedTabCount = 3
+local tabSize = 8
```
NeoRaider, line 41: Let's use spaces instead of tabs for the output alignment.

package/gluon-core/luasrc/usr/bin/gluon-info
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/lua
+
+local uci = require('simple-uci').cursor()
+local pretty_hostname = require 'pretty_hostname'
+
+local site = require 'gluon.site'
+local sysconfig = require 'gluon.sysconfig'
+local platform = require 'gluon.platform'
+local util = require "gluon.util"
```
NeoRaider, line 9:
```suggestion
local util = require 'gluon.util'
```

NeoRaider (changes requested):
Please use tabs instead of spaces for indentation of Lua code.

package/gluon-core/luasrc/usr/bin/gluon-info
mkg20001: This is copypasta from web-admin package, so chances are it's a leftover

mkg20001: I'm thinking whether it makes sense to have an exact copy of this and if it's better to move it to some lib instead, that both the UI and CLI source the function from. Perhaps the translate function could be an optional argument then in order to get translations.

mkg20001:
@NeoRaider how do I lint the script? My local lua linter doesn't show anything.

NeoRaider:
> @NeoRaider how do I lint the script? My IDE's lua linter doesn't show anything.

If you have luacheck installed, you can just `make lint-lua` in the root of the git repository.

.editorconfig
```diff
@@ -34,3 +34,6 @@ indent_size = 2
 [CMakeLists.txt]
 indent_style = space
 indent_size = 2
+
+[*.lua]
+indent_style = tab
```
NeoRaider, line 39: This should already be covered by the `[*]` entry at the top (also, the majority of our Lua scripts don't have the `.lua` extension).

package/gluon-core/luasrc/usr/bin/gluon-info
NeoRaider: Hmm, both the fastd and wireguard VPN providers (`gluon-mesh-vpn-fastd/luasrc/usr/lib/lua/gluon/mesh-vpn/provider/fastd.lua`, `gluon-mesh-vpn-wireguard/luasrc/usr/lib/lua/gluon/mesh-vpn/provider/wireguard.lua`) currently return an empty string rather than `nil` when no key exists. I consider this a bug that should be fixed there, so we don't have to handle the empty string in web-admin or in gluon-info.

package/gluon-core/luasrc/usr/bin/gluon-info
```diff
@@ -0,0 +1,38 @@
...
+		pubkey = active_vpn.public_key()
+	end
+end
+
+local values = {
+	{ 'Hostname', pretty_hostname.get(uci) },
+	{ 'MAC address', sysconfig.primary_mac },
+	{ 'Hardware model', platform.get_model() },
+	{ 'Gluon version' .. " / " .. 'Site version', util.trim(util.readfile('/lib/gluon/gluon-version'))
+		.. " / " .. util.trim(util.readfile('/lib/gluon/site-version')) },
```
NeoRaider, line 26: Consistently use `'` instead of `"` and merge the label to a single string (not sure of LuaSrcDiet can do this for us)

NeoRaider (changes requested):
Thanks, this is shaping up nicely, I have just two minor comments left. Also, the gluon-web-admin commit title has two typos.

NeoRaider (approved):

PR #2374, opened 2022-01-12
ath79-generic: add support for TP-Link Archer A7 v5

Luflosi:
This is my first contribution, so please tell me if I did something incorrectly.

  - [ ] webinterface
  - [x] tftp
  - [ ] other: <specify>
- wired network
- wifi (if applicable)
- led mapping
  - radio leds
    - [x] should show activity
  - switchport leds
- outdoor devices only

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -44,6 +44,7 @@ ath79-generic
 
 * TP-Link
 
+  - Archer A7 (v5)
   - Archer C6 (v2)
   - CPE220 (v3.0)
   - CPE510 (v2.0)
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac
@@ -74,6 +74,7 @@ local primary_addrs = {
 		{'ath79', 'generic', {
 			'glinet,gl-ar750s-nor',
 			'ocedo,raccoon',
+			'tplink,archer-a7-v5',
 			'tplink,archer-c2-v3',
 			'tplink,archer-d50-v1',
 		}},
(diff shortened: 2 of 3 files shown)


Luflosi:
I think it did not work but I'm not 100% sure. I got a little confused because initially I thought I had an Archer C7 v5. The web interface rejected the OpenWrt factory image for the C7 and I after I realised my mistake I think it also rejected the A7 image. I can try to flash the stock firmware back and test again if you want.

NeoRaider:
Testing this once more would be great. If it doesn't work, it is likely that it can be made to work again by applying the same fix that was done for the Archer C7v5 to the A7 as well: https://git.openwrt.org/?p=project/firmware-utils.git;a=commitdiff;h=c5b10f60e21ab85989fa19324de901dceef69450

Luflosi:
I flashed the stock firmware again and updated to the latest version (1.1.4 Build 20211022 rel.58866). When trying to flash the gluon factory image via the web interface, it rejects the image with the message "Invalid file type.". Flashing the same image over TFTP works.

NeoRaider:
Can you try with the attached patch applied to the OpenWrt repo? You can just put the file into `patches/openwrt` as `0007-firmware-utils-tplink-safeloader-fix-Archer-A7v5-factory-flashing-from-vendor-fw-v1.1.x.patch` and run `make update` again.

[0007-firmware-utils-tplink-safeloader-fix-Archer-A7v5-factory-flashing-from-vendor-fw-v1.1.x.patch.txt](https://github.com/freifunk-gluon/gluon/files/7872378/0007-firmware-utils-tplink-safeloader-fix-Archer-A7v5-factory-flashing-from-vendor-fw-v1.1.x.patch.txt)

Luflosi:
Yep, with your patch I was able to flash the factory image using the web ui.
Should I include it as a commit in this PR?

NeoRaider:
No need, I've pushed the patch to OpenWrt master and 21.02, so we'll get it with the next update.

Luflosi:
Thank you!

Issue #2375, opened 2022-01-18
babel: firewall issues with mesh-on-wan without vxlan


T-X:
#### Setup Description

* Local, virtual test setup on a laptop with 2 nodes and 2 clients:

```
client0 <-> [node0] <= mesh-on-wan => [node1] <-> client1
```
* node0, node1: x86 KVM Gluon nodes (current master, on 7427ba2280ac1)
* client0, client1: hosts in separate network namespaces
* mesh-on-WAN, no VXLAN

#### Issue

Nodes and clients can't ping each other on their "node_prefix6" and "node_client_prefix6" addresses when using mesh-on-wan without VXLAN. Nodes don't recognize each other in BABEL.

Babeld's status dump only shows a node's own routes and own client routes, but not from other nodes. tcpdump shows BABEL's hello messages, but no "I-heard-you" responses.

Calling `ip6tables -I INPUT -p udp --dport 6696 -j ACCEPT` and `ip6tables -I INPUT -p udp --dport 5523 -j ACCEPT` makes BABEL IHU messages appear and allows BABEL's routing table to fill up correctly. And allows the two nodes to ping each other and allows a node to ping both clients. While pinging between clients still is rejected by the direct node ("Destination unreachable: Port unreachable"). Enabling VXLAN makes things work as well, including pinging between the two clients.

#### Additional Information

**Gluon Version:**
* a recent Gluon master: 7427ba2280ac1

**Site Configuration:**
* https://gist.github.com/T-X/9dbab63be36782aa73c7694fced3643d
  * site-default-novxlan.conf -> not working
  * site-vxlan.conf -> working

Concerns: #2353

T-X:
What actually surprises me a bit is that mesh-on-wan with VXLAN works. When doing a `$ iptables -S | grep 6696` in my KVM setup with no wifi interface I get no results. So I'm wondering if it works by accident, if VXLAN by accident is less strict than mesh-on-wan without VXLAN (zone "wired_mesh" rejects INPUT and FORWARD, but I don't see a zone for "wired_mesh_vxlan").

There are rules "mesh_babel" and "mesh_l3roamd" for zone "mesh" (which I believe is wireless only?). But I don't see such babel/l3roamd rules for zone "wired_mesh".

NeoRaider:
With Babel, there is no way to distinguish mesh and non-mesh traffic on the same interface by the packet type, so this simply cannot work - after all, mesh traffic must never leak into the WAN network for security reasons!

[...]

PR #2376, opened 2022-01-18
gluon-mesh-babel: do not allow disabling VXLAN

NeoRaider:
With Babel, wired meshing cannot run on the same logical interface as
non-mesh traffic, so using VXLAN is mandatory.

See https://github.com/freifunk-gluon/gluon/issues/2375#issuecomment-1015683959


NeoRaider:
Another option would be to use `obsolete()` to disallow setting `vxlan` altogether when `mesh-babel` is installed, but that currently conflicts with the requirement to set `vxlan` (to `true` or `false`) when the multidomain feature is not used. This could be solved by making the `vxlan` setting optional for single-domain setups as well (defaulting to `true`).

PR #2377, opened 2022-01-19
ath79-mikrotik: add support for Mikrotik RB951Ui-2nD

mkg20001:
  - [ ] webinterface
  - [x] tftp
  - [ ] other: <specify>
- wired network
- wifi (if applicable)
- led mapping
  - radio leds
    - n/a should map to their respective radio
    - n/a should show activity
  - switchport leds
- outdoor devices only

Note that mikrotik was a target back in the days,
but that target was mostly a generic catch-all image

Now that the images are per-device the sysupgrade
should just work :tm:

That means there's basically no migration path,
but this is a very special case as previously it was
a generic image and now it's a custom one,
so forgive that.

Installation is done by netbooting the image (initramfs),
then in config-mode doing a manual upgrade

Perhaps that second sysupgrade step could be automated somehow

### DIFF
--- a/.github/filters.yml
+++ b/.github/filters.yml
@@ -17,6 +17,16 @@
     "targets/generic",
     "targets/targets.mk"
   ],
+  "ath79-mikrotik": [
+    "targets/ath79-mikrotik",
+    "modules",
+    "Makefile",
+    "patches/**",
+    "scripts/**",
+    "targets/generic",
+    "targets/targets.mk",
+    "targets/mikrotik.inc"
+  ],
   "bcm27xx-bcm2708": [
     "targets/bcm27xx-bcm2708",
     "modules",
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -118,6 +118,13 @@ ath79-nand
   - GL-AR300M
   - GL-AR750S
 
+ath79-mikrotik
+--------------
+
+* Mikrotik
+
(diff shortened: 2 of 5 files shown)


mkg20001:
I'm still testing this one and 802.11s doesn't like to work. Not sure if it's the driver or the device.

mkg20001:
The label has a mac for eth1 and wlan, currently 010-primary-mac picks up the label mac.

Might make sense to switch to wifi mac as that one's more relevant for meshing? Or should I just leave it as is?

NeoRaider:
It seems that this device is not in OpenWrt master yet. We generally follow an "upstream-first" development model, so that we don't end up with downstream patches that we need to maintain forever.

This usual way for new hardware support to be added to Gluon is the following:

1. Add device to OpenWrt master
2. Wait for new major OpenWrt release, or have device backported to last stable OpenWrt (the latter is not always possible)
3. Add new device entry to Gluon

txt-file:
Adding to NeoRaiders raised issues: patches/openwrt/0015-ath79-add-support-for-MikroTik-hAP-ac-RB951Ui-2nD.patch should be renamed. filename says "hap ac" but file content is for "hap".

mkg20001:
This needs https://github.com/openwrt/openwrt/pull/9825 and https://github.com/openwrt/openwrt/pull/4967 merged upstream for the patches to be mainline.

PR #2379, opened 2022-01-23
kernel: drop obsolete ebtables patches

NeoRaider:
We are on Linux 5.4, so these patches for Linux 4.14 don't do anything.


NeoRaider:
Closing, will be combined with another update.

PR #2383, opened 2022-01-25
gluon-web: prohibit cross-origin POST requests

NeoRaider:
It is currently possible for malicious sites to inject POST requests, allowing to change settings on nodes in config mode that are reachable from the client's host. A simple PoC can be found under https://home.universe-factory.net/neoraider/gluon-cors.html (note that you will need to allow mixed content to make the HTTPS -> HTTP request work - an actual attacker would simply host the site on HTTP). My example simply changes the fastd mode from "security" to "performance", but uploading an SSH key would be just as simple.

I did not actually test the HTTPS code path, but I used curl to simulate various different Origin headers (including different ports and schemes) and tested using the config mode both without and with redirects TCP from Firefox and Chrome, including manual firmware updates.

### DIFF
--- a/package/gluon-web/luasrc/usr/lib/lua/gluon/web/dispatcher.lua
+++ b/package/gluon-web/luasrc/usr/lib/lua/gluon/web/dispatcher.lua
@@ -184,9 +184,15 @@ local function dispatch(config, http, request)
 		return
 	end
 
-	http:parse_input(node.filehandler)
+	local ok, err = pcall(http.parse_input, http, node.filehandler)
+	if not ok then
+		http:status(400, "Bad request")
+		http:prepare_content("text/plain")
+		http:write(err .. "\r\n")
+		return
+	end
 
-	local ok, err = pcall(node.target)
+	ok, err = pcall(node.target)
 	if not ok then
 		http:status(500, "Internal Server Error")
 		renderer.render_layout("error/500", {
--- a/package/gluon-web/luasrc/usr/lib/lua/gluon/web/dispatcher.lua
+++ b/package/gluon-web/luasrc/usr/lib/lua/gluon/web/dispatcher.lua
@@ -208,6 +208,6 @@ return function(config, http)
 	if not ok then
 		http:status(500, "Internal Server Error")
 		http:prepare_content("text/plain")
-		http:write(err)
+		http:write(err .. "\r\n")
 	end
 end
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -10,7 +10,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Install Dependencies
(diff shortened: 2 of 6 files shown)


NeoRaider:
I have no idea what is wrong with our Lua CI jobs today, but they pass locally O_o

NeoRaider:
I just pushed another update; now an Origin header is required for all POST requests. This works fine with all reasonably recent browsers; however it is only required by the Fetch spec since 2016, as stated in the footnote of https://stackoverflow.com/questions/42239643/when-do-browsers-send-the-origin-header-when-do-browsers-set-the-origin-to-null/42242802#42242802 .

We can't really do anything about such old browsers, as they don't necessarily send an Origin header for POST requests from `<form>` elements - so allowing requests without Origin would break a part of the attack mitigation.

NeoRaider:
Ugh, found another bug in my changes, the check breaks when the host is a literal IPv6 address.

NeoRaider:
> Ugh, found another bug in my changes, the check breaks when the host is a literal IPv6 address.

Fixed as well now.

lemoer (approved):
I like that we use pcall() and error() here now.

Besides looking at the code, I did some tests and found that:
- Cross origin requests are now prohibited, if the HTTP-Host and the Origin-Header do not match.
- Invalid Origin-Headers are properly handled.
- The error "Invalid Content-Type found" is thrown, when I expect it.

Overall, this looks good to me. From my side, this is ready for merge.

package/gluon-web/luasrc/usr/lib/lua/gluon/web/dispatcher.lua
```diff
@@ -184,9 +184,15 @@ local function dispatch(config, http, request)
 		return
 	end
 
-	http:parse_input(node.filehandler)
+	local ok, err = pcall(http.parse_input, http, node.filehandler)
+	if not ok then
+		http:status(400, "Bad request")
+		http:prepare_content("text/plain")
+		http:write(err)
```
lemoer, line 191:
``` suggestion
		http:write(err .. '\n')
```

A minor thing. We could add a newline here. This looks nicer, e.g. if you are using curl. Otherwise my terminal looks like this:

```
/usr/lib/lua/gluon/web/http/protocol.lua:189: POST request with invalid Origin header[lemoer@orange ~]$ 
```

NeoRaider: Good idea. I used `\r\n` (which may or may not be better than `\n` - FWIW, it's what's used in the HTTP header) and also added it to the other `text/plain` error output.

lemoer (approved):

NeoRaider:
@mweinelt I think you had reservations regarding this PR? I'm open to other solutions, if you have any suggestions...

PR #2387, opened 2022-01-31
workflows: add backports action

mweinelt:
By applying a label `backport <branch>` the action will automatically
try to cherry-pick the change to the target branch after the pull
request was successfully merged.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,0 +1,21 @@
+name: Backport
+on:
+  pull_request_target:
+    types: [closed, labeled]
+jobs:
+  backport:
+    name: Backport Pull Request
+    if: github.repository_owner == 'freifunk-gluon' && github.event.pull_request.merged == true && (github.event_name != 'labeled' || startsWith('backport', github.event.label.name))
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: Create backport PRs
+        uses: zeebe-io/backport-action@v0.0.7
+        with:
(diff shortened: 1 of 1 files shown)


NeoRaider:
It looks like the current version of the action is v0.0.7, and it doesn't require specifying the version twice (anymore?): https://github.com/zeebe-io/backport-action#installation

NeoRaider:
Also, how are the backported commits formatted? Is it possible to get a commit ID reference to the original commits like `git cherry-pick -x`?

mweinelt:
Yes, it cherry-picks with `-x`. Looking into the update in a bit.

mweinelt:
And updated to 0.0.7.

.github/workflows/backport.yml
```diff
@@ -0,0 +1,22 @@
...
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: Create backport PRs
+        uses: zeebe-io/backport-action@v0.0.7
+        with:
+          # Config README: https://github.com/zeebe-io/backport-action#backport-action
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          github_workspace: ${{ github.workspace }}
+          pull_description: |-
+            Automatic backport to `${target_branch}`, triggered by a label in #${pull_number}.
+
```
NeoRaider, line 22: Extra trailing newline

NeoRaider:
LGTM with the extra trailing newline removed.

.github/workflows/backport.yml
mweinelt: Removed.

NeoRaider:
I tried out the action in #2383, and found 2 more issues:
- Action runs twice when a PR is labelled right after merging
- Action failed because the checkout action only fetched the v2021.1.x branch, not the master branch, so the commits for cherry-picking where not found.

PR #2388, opened 2022-02-01
modules: update openwrt

mweinelt:
Just a quick modules update to test the backport action.


NeoRaider:
Needs a patch update, which I already did in #2381 (currently waiting for review)

NeoRaider:
(also, not useful for any backport testing, wrong OpenWrt version :grimacing:)

mweinelt:
Yep, that's why I dropped the label again. Forgot that wouldn't make sense. Failed to edit the op.

PR #2399, opened 2022-02-20
treewide: Replace xhtml content type with html, remove xhtml workaround, replace self closing tags

MyIgel:
As a follow up to #2398 / bbc00017
* Replace xhtml content type with html -> pages are interpreted as HTML5 which is the current standard and much more flexible
* Remove old xhtml workaround which is not needed anymore
* Updated `prepare_content` docs accordingly


skorpy2009 (approved):

docs/releases/v2016.1.rst
```diff
@@ -236,7 +236,7 @@ Site changes
     "</p>"
     "<div class=\"the-key\">"
     " # <%= hostname %>"
-    " <br/>"
+    " <br>"
```
NeoRaider, line 239: Please don't change old release notes.

NeoRaider (changes requested):

docs/releases/v2016.1.rst
MyIgel: Ärm jeah, thats a great idea & fixed :D

NeoRaider (approved):

rubo77:
Nice.

Why was the XHTML tag there in the first place?

MyIgel:
I would guess legacy reasons? Meaning: It was required for some browsers to work back then and is now totally obsolete

NeoRaider:
Yeah, the code originated from LuCI, which was written before HTML5 existed. Back then XHTML was considered the future of HTML I guess (and personally I still prefer XML syntax even with HTML5 for many usecases - but for gluon-web it causes more problems than it solves).

PR #2401, opened 2022-02-23
v2021.1.x: update modules

NeoRaider:
Lots of updates, in particular to kernel, batman-adv, SSL libraries.

~~Not tested yet.~~


NeoRaider:
Updated to latest module versions. Boot tested (config + regular mode) on x86-64 QEMU, D-link DIR-860L-B1 and TP-Link TL-WDR3600 v1.

rotanid:
currently running without an issue on a (only one) test device

PR #2404, opened 2022-02-26
x86: switch to EFI-compatible images

### LINKED ISSUE #2403: x86-64: Add target for UEFI boot compatible images
A lot of newer PCs don't have CSM any more and are unable to boot the regular x86 images.
Hence it would be great if the build scripts could create EFI-compatible images as well.

(see https://openwrt.org/docs/guide-developer/uefi-bootable-image)

NeoRaider:
The main difference between the non-EFI and EFI images generated by
OpenWrt is that the former uses an MS-DOS partition table, while the
latter uses GPT. The EFI images still have a BIOS-compatible MBR, so
they work fine on non-EFI systems.

Closes #2403

Draft PR for testing. Boot test in qemu was successful for x86-64 both with default BIOS and UEFI (OVMF). In particular it needs to be tested whether upgrading from a non-EFI image to an EFI image works smoothly - ideally replacing the partition table with GPT, so we don't need to think about upgrades from old systems using MS-DOS partitions forever...

### DIFF
--- a/targets/x86.inc
+++ b/targets/x86.inc
@@ -44,8 +44,8 @@ packages {
 config('TARGET_ROOTFS_EXT4FS', false)
 
 defaults {
-	factory = '-squashfs-combined',
+	factory = '-squashfs-combined-efi',
 	factory_ext = {'.img.gz', '.vmdk', '.vdi'},
-	sysupgrade = '-squashfs-combined',
+	sysupgrade = '-squashfs-combined-efi',
 	sysupgrade_ext = '.img.gz',
 }


AiyionPrime:
@kpanic23 would you mind testing this?
Would be nice to know if it successfully resolves your issue #2403.

AiyionPrime:
@mkg20001 Did you see @NeoRaider s suggestions, about what should be tested? And if, what results did you get?

mkg20001:
I did that, got "wrong efi loader signature", but it boots despite that

image

mkg20001:
mbr -> uefi with mbr-only: works
uefi -> mbr with uefi-only: works ( = breaks as supposed to)

mkg20001:
mbr -> uefi with mbr and then switching to uefi works (previously uefi failed to boot, now both qemu with mbr and qemu with uefi boots)

AiyionPrime (approved):
Looks good, @mkg20001's tests look promising and the action ran successfully. I'd opt for merging.

AiyionPrime:
@blocktrron you worked on these files before, maybe you can think of a setup previous tests failed to evaluate.
If not, feel free to merge.

mkg20001:
can this be merged?

christf:
LGTM

PR #2405, opened 2022-03-05
gluon-blink: add command to make the current device's status led blink

mkg20001:
This allows a user to find out which device he's connected to by running 
gluon-blink

The command will trigger a fast-blinnk on the status led


NeoRaider:
I like the idea, but I think it would make sense to remove the Gluon-specific parts and add this to OpenWrt's base-files instead.

mkg20001:
Is now a community-package, will PR to openwrt eventually, somewhere after olsr2, maybe

PR #2407, opened 2022-03-16
docs: fix invalid fritztools link

AiyionPrime:
As @ambassador86 spotted earlier, the domain https://fritzfla.sh has changed its target audience.
The appropriate substitute would be the one referenced on the projects page https://github.com/freifunk-darmstadt/fritz-tools.

I think this could be merged as soon as somebody confirmed the link works as intended ;)


rotanid:
> As @ambassador86 spotted earlier, the old domain has changed its target audience. 

nice way to put it ;-)
i removed the actual domain to not support the grabber.

PR #2408, opened 2022-03-19
docs: move release notes TOC to sub page

NeoRaider:
This allows us to organize the TOC a bit better by adding sections per
major version. We can even increase the maxdepth to 2 now, which looks
great in my opinion.

In addition, the full list of releases is not shown in the sidebar
anymore when viewing a completely different part of the documentation,
which took up more than half of the total sidebar entries.

Screenshot 2022-03-19 at 15-52-41 Release Notes — Gluon 2021 1 documentation


mweinelt:
And the releases are collapsed in the navbar. Looks great.

image

PR #2389, opened 2022-02-02
ath79: (re)add support for CPE210 v1.1

AiyionPrime:
This is obviously **blocked**.

I just wanted to take neoraiders PR #2372 for a testride and thought I'd document what I've done here.
This is just a draft, I'll convert this to a proper device PR once the migration-PR becomes stable.

PR #2410, opened 2022-03-20
ath79: (re)add support for CPE210 v1.1

AiyionPrime:
These tests have been done for #2389, but my master just rebased about 17 commits, so I'll do them again for completeness.
The Device has ben in ar71xx and migrated without problems.

- [ ] must be flashable from vendor firmware
  - [ ] webinterface
  - [ ] tftp
  - [ ] other: <specify>
  - [ ] must have working sysupgrade
    - [ ] must keep/forget configuration (if applicable)
- [ ] reset/wps/phone button must return device into config mode
- [ ] primary mac should match address on device label (or packaging) (https://gluon.readthedocs.io/en/latest/dev/hardware.html#notes)
- wired network
- wifi (if applicable)
  - [ ] association with AP must be possible on all radios
- led mapping
    - [ ] lit while the device is on
  - radio leds
    - [ ] should show activity
  - switchport leds
- outdoor devices only


AiyionPrime:
This works as expected and is ready for review.

targets/ath79-generic
```diff
@@ -123,6 +123,11 @@ device('tp-link-archer-d50-v1', 'tplink_archer-d50-v1', {
 	broken = true, -- 64M ath9k + ath10k & power LED not working
 })
 
+device('tp-link-cpe210-v1', 'tplink_cpe210-v1', {
+	manifest_aliases = {
+		'tp-link-cpe210-v1.1',
```
NeoRaider, line 128: Please also add back the `tp-link-cpe210-v1.0` manifest alias, so upgrades from that revision work. It is the same hardware as the v1.1 and has always used the same images.

AiyionPrime: I have no such device at hand and cannot verify it does work. Still?

NeoRaider: Yes, people will install the image on the device anyways, as it is called just "tp-link-cpe210-v1". So it makes sense to allow upgrades as well.

AiyionPrime: resolved in 3c34a01c64e5c2c3f818e757f8657c945bcd4f58

AiyionPrime:
That should've been everything.

targets/ath79-generic
```diff
@@ -123,6 +123,12 @@ device('tp-link-archer-d50-v1', 'tplink_archer-d50-v1', {
 	broken = true, -- 64M ath9k + ath10k & power LED not working
 })
 
+device('tp-link-cpe210-v1', 'tplink_cpe210-v1', {
+	manifest_aliases = {
+		'tp-link-cpe210-v1.0',
```
NeoRaider, line 128: Please add a comment to these manifest aliases that they are only used for upgrades from OpenWrt 19.07, so we know that we can get rid of them in the future.

AiyionPrime: Resolved in b2dd511785b611e612d626f14a2379d36cc77d3a, I think.

NeoRaider: The Gluon version in the comment is wrong, the last version for which the manifest alias is relevant is 2021.1, not 2020.1. Or just remove it and only mention the OpenWrt version.

Also, I'd prefer a short comment for each manifest alias that can be removed in the future, maybe like https://github.com/freifunk-gluon/gluon/pull/2417#issuecomment-1076863818, so we don't need to guess which aliases the comment refers to when we look at it in a few years.

AiyionPrime: You are right, I looked for the first and not the last gluon version.

Resolved in 440fb7032256317889942e09341b3790a5596180.

PR #2414, opened 2022-03-21
ath79-generic: add support for D-Link DAP-2660 A1

s-2:
```
Specifications:

 * QCA9557, 16 MiB Flash, 128 MiB RAM, 802.11n 2T2R
 * QCA9882, 802.11ac 2T2R
 * Gigabit LAN Port (AR8035), 802.3af PoE
 * 3,5 x 1,35 mm DC plug (12V) alternative supply
```
 
  - [x] webinterface
  - [ ] tftp
  - [x] other: recovery webinterface (keep reset button pressed during power-on, until upload started)
- wired network
- wifi (if applicable)
- led mapping
  - radio leds
    - [ ] should show activity
  - switchport leds
- outdoor devices only

Signed-off-by: Sebastian Schaper <openwrt@sebastianschaper.net>

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -13,6 +13,10 @@ ath79-generic
   - WiFi pro 1750i
   - WiFi pro 1750x
 
+* D-Link
+
+  - DAP-2660 A1 [#lan_as_wan]_
+
 * Enterasys
 
   - WS-AP3705i
--- a/targets/ath79-generic
+++ b/targets/ath79-generic
@@ -55,6 +55,14 @@ device('devolo-wifi-pro-1750x', 'devolo_dvl1750x', {
 })
 
 
+-- D-Link
+
+device('d-link-dap-2660-a1', 'dlink_dap-2660-a1', {
+	factory_ext = '.img',
+	packages = ATH10K_PACKAGES_QCA9880,
+})
+
+
 -- Enterasys
 
 device('enterasys-ws-ap3705', 'enterasys_ws-ap3705i', {


s-2:
There are more devices from the series that have been added to OpenWrt in late 2020, thus included in the 21.02 release:
* DAP-2230 (2.4G only variant)
* DAP-2680 (3-stream Wave 2 variant)
* DAP-3320 (outdoor variant of 2230)
* DAP-3662 (outdoor variant of 2660).

Also, DAP-1330 could be merged from `ar71xx` to `ath79`.

I'm not sure though whether there are any policies regarding the relevance of devices for inclusion to gluon (similar to the wikipedia "inclusionism" debate), or if all devices available in OpenWrt should be added as soon as they were tested.

The most relevant to add would probably be DAP-3662 and DAP-2680, but I could fully test the others as well (https://github.com/s-2/gluon/commit/0a3ac9746c9ca8a62dce2abfc94c7c7f71338f64) if there is interest.

NeoRaider:
> * radio leds
>   
>   * [ ]  should map to their respective radio
>   * [ ]  should show activity
> 
> * switchport leds
>   
>   * [ ]  should map to their respective port (or switch, if only one led present)
>   * [ ]  should show link state and activity

Do the unchecked boxes mean that no such LEDs exist?

> I'm not sure though whether there are any policies regarding the relevance of devices for inclusion to gluon (similar to the wikipedia "inclusionism" debate), or if all devices available in OpenWrt should be added as soon as they were tested.

As long as someone is interested in using a device with Gluon, and it is working without additional OpenWrt patches, adding it to Gluon is fine. When nobody actually wants to use it, it should not be added just because it exists in OpenWrt.

s-2:
> Do the unchecked boxes mean that no such LEDs exist?

Yes, there is just a single red and green power/status LED that is not used to show wireless or ethernet activity in the OEM firmware, red is only used during boot or firmware upgrade. To be consistent with both OpenWrt and OEM firmware, I would leave it this way.

PR #2415, opened 2022-03-21
ath79: (re)add support for tl-wr1043nd-v3

AiyionPrime:
Has been tested by @lemoer, will verify the checklist tomorrow.

  - [x] webinterface
  - [x] tftp
  - ~~other: <specify>~~
- wired network
- wifi (if applicable)
- led mapping
  - radio leds
    - [x] should show activity
  - switchport leds
- ~~outdoor devices only~~
  - added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`

### DIFF
--- a/targets/ath79-generic
+++ b/targets/ath79-generic
@@ -134,4 +134,10 @@ device('tp-link-eap225-outdoor-v1', 'tplink_eap225-outdoor-v1', {
 device('tp-link-tl-wdr3600-v1', 'tplink_tl-wdr3600-v1')
 device('tp-link-tl-wdr4300-v1', 'tplink_tl-wdr4300-v1')
 
+device('tp-link-tl-wr1043nd-v3', 'tplink_tl-wr1043nd-v3', {
+	manifest_aliases = {
+		'tp-link-tl-wr1043n-nd-v3', -- upgrade from OpenWrt 19.07
+	},
+})
+
 device('tp-link-wbs210-v2', 'tplink_wbs210-v2')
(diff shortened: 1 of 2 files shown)


targets/ath79-generic
```diff
@@ -131,6 +131,8 @@ device('tp-link-eap225-outdoor-v1', 'tplink_eap225-outdoor-v1', {
        packages = ATH10K_PACKAGES_QCA9888,
 })
 
+device('tp-link-tl-wr1043n-nd-v3', 'tplink_tl-wr1043nd-v3')
```
NeoRaider, line 134: Is `tp-link-tl-wr1043n-nd-v3` what is reported as the image name by libplatforminfo? I'd expect this to use the model name from the device tree, in which case it would be "tp-link-tl-wr1043nd-v3", but maybe I'm looking at the wrong string.

AiyionPrime: platforminfo confirmed it, you were right.
resolved in 952f818a0992679e0f602773386aa0323c51e08c

targets/ath79-generic
```diff
@@ -131,6 +131,12 @@ device('tp-link-eap225-outdoor-v1', 'tplink_eap225-outdoor-v1', {
        packages = ATH10K_PACKAGES_QCA9888,
 })
 
+device('tp-link-tl-wr1043nd-v3', 'tplink_tl-wr1043nd-v3', {
+	manifest_aliases = {
+		'tp-link-tl-wr1043n-nd-v3',
```
NeoRaider, line 136: Missing "upgrade from 19.07" comment.

AiyionPrime: resolved in eaea64c905bb7c8c56925847d3f86fb8d71d751b

AiyionPrime:
The target was wrongly put above the 3600 and 4300 in targets. Other than that nothing changed.

PR #2416, opened 2022-03-22
ath79-generic: (re)add support for UniFi AP PRO

AiyionPrime:
The device appears to be working as expected.
This is ready for review.

Lastly the LED (the circle in the middle is off during normal operation).

  - ~~webinterface~~
  - [x] tftp
  - [x] other: some versions are flashable from the unifi firmware. not recommended.
- wired network
- wifi (if applicable)
- led mapping
    - [x] lit while the device is on (most of these devices have a broken blue led these days, but when its perfectly dark its working)
  - ~~radio leds~~ (there are none)
  - ~~switchport leds~~ (there are none)
- ~~outdoor devices only~~

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -86,6 +86,7 @@ ath79-generic
 
   - UniFi AP
   - UniFi AP LR
+  - UniFi AP PRO
 
 ath79-nand
 ----------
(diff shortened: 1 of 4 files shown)


AiyionPrime:
> Either it's using the wrong interface for its label MAC, or the interface MACs don't match what the stock firmware does. Regardless it's a bug.

_Originally posted by @neoraider in hackint#gluon_

PR #2417, opened 2022-03-22
ath79: migrate D-Link DAP-1330 and DIR-550 from ar71xx

s-2:
also add DAP-1365, which is a different case variant of DAP-1330

The checklist actually applies to all 3 devices; is it acceptable to have these within a single PR and/or should these be split into several PRs / commits?

  - [x] webinterface
  - [ ] tftp
  - [x] other: web recovery (works best with curl); 192.168.0.50 (DAP-1330/65), 192.168.0.1 (DIR-505)
- wired network
- wifi (if applicable)
- led mapping
  - radio leds
    - [ ] should show activity
  - switchport leds
- outdoor devices only

Signed-off-by: Sebastian Schaper <openwrt@sebastianschaper.net>

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -19,7 +19,11 @@ ath79-generic
 
 * D-Link
 
+  - DAP-1330 A1 [#lan_as_wan]_
+  - DAP-1365 A1 [#lan_as_wan]_
   - DAP-2660 A1 [#lan_as_wan]_
+  - DIR-505 A1 [#lan_as_wan]_
+  - DIR-505 A2 [#lan_as_wan]_
 
 * Enterasys
 
--- a/targets/ath79-generic
+++ b/targets/ath79-generic
@@ -63,9 +63,20 @@ device('devolo-wifi-pro-1750x', 'devolo_dvl1750x', {
 
 -- D-Link
 
+device('d-link-dap-1330-a1', 'dlink_dap-1330-a1')
+device('d-link-dap-1365-a1', 'dlink_dap-1365-a1')
+
 device('d-link-dap-2660-a1', 'dlink_dap-2660-a1', {
 	factory_ext = '.img',
 	packages = ATH10K_PACKAGES_QCA9880,
+})
+
+device('d-link-dir-505', 'dlink_dir-505', {
+        factory = false,
+        manifest_aliases = {
+                'd-link-dir-505-rev-a1', -- Upgrade from OpenWrt 19.07
+                'd-link-dir-505-rev-a2', -- Upgrade from OpenWrt 19.07
+        },
 })
 
 


s-2:
Tested DAP-1330 and DAP-1365 successfully,
https://hannover.freifunk.net/karte/#/6c722005f39c
https://hannover.freifunk.net/karte/#/409bcd00b325

however DIR-505 is still confusing me. The device works,
https://hannover.freifunk.net/karte/#/c8be196d14e1

but it seems both the profile name and filenames seem to have changed in OpenWrt from ar71xx -> ath79, also there is no more factory image in ath79.
After multiple attempts, I cannot seem to find a combination of profile names that would neither cause a build error very early (`could not enable...`) or very late in the process (`cannot stat 'openwrt/bin/...'`), nor give the wrong output of `platform_info` image name.

I'll spend some more time on this tomorrow, maybe I'll just leave the device.

By the way, since I only have the A1 version of that device, how exactly does `alias` work, will it just place a copy of the file to the output dir with a different name? Thus during autoupdate, all A2 devices will actually request the A1 URL from the update server (which doesn't matter, as the files are identical)?

NeoRaider:
> but it seems both the profile name and filenames seem to have changed in OpenWrt from ar71xx -> ath79, also there is no more factory image in ath79. After multiple attempts, I cannot seem to find a combination of profile names that would neither cause a build error very early (`could not enable...`) or very late in the process (`cannot stat 'openwrt/bin/...'`), nor give the wrong output of `platform_info` image name.

I can have a look. The first argument to `device()` must be the image name returned by libplatforminfo, the second argument is the OpenWrt profile name. If an OpenWrt profile can be enabled, but the image is not found, this usually means that the image type (sysupgrade/factory) does not exist, or the file extension is wrong.

> By the way, since I only have the A1 version of that device, how exactly does `alias` work, will it just place a ~copy of~ symlink to the file to the output dir with a different name? Thus during autoupdate, all A2 devices will actually request the A1 URL from the update server (which doesn't matter, as the files are identical)?

`alias` creates a symlink, so the image will be available under alternative names both for the autoupdater and in directory listings etc. There is also `manifest_alias`, which will only create an additional entry in the autoupdater manifest, which is useful when the second name doesn't need to be visible to users.

s-2:
> The first argument to `device()` must be the image name returned by libplatforminfo, the second argument is the OpenWrt profile name. If an OpenWrt profile can be enabled, but the image is not found, this usually means that the image type (sysupgrade/factory) does not exist, or the file extension is wrong.

The last variant I tried was
```
device('d-link-dir-505-rev-a1', 'dlink_dir-505-a1', {
       profile = 'dlink_dir-505',
       aliases = {'d-link-dir-505-rev-a2'},
       factory = false,
})
```
which generates `openwrt-ath79-generic-dlink_dir-505-squashfs-sysupgrade.bin`
but `scripts/target_lib.lua:117` expects `-a1` to be included.

`profile = 'dlink_dir-505-a1',` will fail to enable the device. I'll try a few more combinations tomorrow :slightly_smiling_face:

NeoRaider:
I think the following should work:

```lua
device('d-link-dir-505', 'dlink_dir-505', {
        factory = false,
        manifest_aliases = {
                'd-link-dir-505-rev-a1', -- Upgrade from OpenWrt 19.07
                'd-link-dir-505-rev-a2', -- Upgrade from OpenWrt 19.07
        },
})
```

s-2:
great, there is obviously no way to quickly resolve merge conflicts using the github web editor and rebase rather than create a new commit, nor to undo this... will rebase later in the evening.

PR #2420, opened 2022-03-23
ath79-generic: (re)add AVM Fritz!WLAN Repeater 450e

AiyionPrime:
  - ~~webinterface~~
  - ~~tftp~~
  - [x] other: avmflash
- wired network
  - [x] must have correct port assignment (WAN)
- wifi (if applicable)
- led mapping
  - radio leds
    - [x] should show activity
  - switchport leds
- ~~outdoor devices only~~
  - added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -4,6 +4,10 @@ Supported Devices & Architectures
 ath79-generic
 --------------
 
+* AVM
+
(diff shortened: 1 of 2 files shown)


AiyionPrime:
This looks fine as well, does work as expected. Thanks @bschelm for lending it.

PR #2421, opened 2022-03-23
ath79-generic: (re)add support for UniFi AP

AiyionPrime:
This device had an alias for the UniFi AP LR device.
I could not find code in gluon that would have treated them differently.
When someone comes to the same conclusion, we could add that variant to this PR, as I've only got the device of a foreigner here, which I'd rather not flash for the purpose.

Other than that, @neoraider highlighted his (temporarily) closed PR #2066.
If I understand it correctly, I should be fine adding the alias for now.

  - ~~webinterface~~
  - [x] tftp
  - ~~other: console was possible in some versions not recommended~~
- wired network
- wifi (if applicable)
- led mapping
  - ~~radio leds~~
    - should map to their respective radio
    - should show activity
  - ~~switchport leds~~
    - should map to their respective port (or switch, if only one led present) 
    - should show link state and activity
- ~~outdoor devices only~~
  - ~~added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`~~


targets/ath79-generic
```diff
@@ -135,3 +135,12 @@ device('tp-link-tl-wdr3600-v1', 'tplink_tl-wdr3600-v1')
 device('tp-link-tl-wdr4300-v1', 'tplink_tl-wdr4300-v1')
 
 device('tp-link-wbs210-v2', 'tplink_wbs210-v2')
+
+-- Ubiquiti
+
+device('ubiquiti-unifi', 'ubnt_unifi', {
+	aliases = {
+		'ubnt_unifi-ap',
```
NeoRaider, line 143: Aliases should "look like" the Gluon device name, and they should usually refer to alternative devices that use the same image (so only the LR gets an alias, the non-LR will be the primary name). There should never be an underscore, and the naming of vendors etc. should be consistent ("ubiquiti" vs. "ubnt" should be avoided).

If an alias is only used for upgrades (possibly from older Gluon versions), a manifest_alias should be used instead.

AiyionPrime: I tried to recreate the ar71xx structure, but fill fix it after I got rid of the mac issue.

https://github.com/freifunk-gluon/gluon/commit/45c84a117bf820bf13e72187b46321b959e23aed#diff-ce406397f31c274de9f8a5907eaf20b0194cccbb6995fa59490e0c56dbd3dea5L472-L477

NeoRaider: In ar71xx the vendor name was consistent, but it did have the issue that separate files "ubiquiti-unifi" and "ubiquiti-unifi-ap" were created, even though both are the same device (the underlying issue being that "ubiquiti-unifi" is technically the wrong name...

One option would be to make "ubiquiti-unifi-ap" the primary name, and "ubiquiti-unifi" a manifest alias (as that is the expected image name).

AiyionPrime: libplatforminfo returns `ubiquiti-unifi` as imagename with a recent OpenWrt.

The current OpenWrt profile name appears to be `ubnt_unifi` which is an inconsistency that should be fixed in OpenWrt, if at all?

Following you recommendation that would leave us with:
```
device('ubiquiti-unifi-ap', 'ubnt_unifi', {
```
the LR alias,
```
	aliases = {
		'ubiquiti-unifi-ap-lr',
	},
```
the manifest alias for the actual image name from libplatforminfo:
```
manifest_aliases = {
		'ubiquiti-unifi-ap-',
	},
```
And no more manifest_aliases since the names of these images match the ones dropped in ar71xx.

AiyionPrime: The produced imagenames do lok properly now:
```console
gluon-ffh-exp-20220324-ubiquiti-unifi-ap-lr-sysupgrade.bin (a symlink to the latter)
gluon-ffh-exp-20220324-ubiquiti-unifi-ap-sysupgrade.bin
```

PR #2435, opened 2022-03-28
 ath79-generic: (re)add support for avm_fritz4020

Dark4MD:
  - n/a webinterface
  - n/a tftp
  - [x] other: fritzflash tools
- [ ] ! primary mac should match address on device label (or packaging) (https://gluon.readthedocs.io/en/latest/dev/hardware.html#notes)
`My EOLO Edition of the 4020 has a Mac Address written on it that is different from the one that openwrt and gluon uses even before migration. In the bootloader it is labeled as macdsl and the one gluon uses is labeled as macwlan. The Mac Address that gluon uses is also found on the label but in the CWMP account (ID?) on my device.`
- wired network
- wifi (if applicable)
- led mapping
  - radio leds
    - [x] should show activity
  - switchport leds
- outdoor devices only

made it a draft until i got the ok that the mac address issue ist not really an issue of openwrt but my "speciel edition" device.

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -7,6 +7,7 @@ ath79-generic
 * AVM
 
   - Fritz!WLAN Repeater 450E [#avmflash]_
+  - Fritz!Box 4020 [#avmflash]_
 
 * devolo
 
--- a/targets/ath79-generic
+++ b/targets/ath79-generic
@@ -24,6 +24,10 @@ local ATH10K_PACKAGES_QCA9888 = {
 
 -- AVM
 
+device('avm-fritz-box-4020', 'avm_fritz4020', {
+	factory = false,
+})
+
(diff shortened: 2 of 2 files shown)


Dark4MD:
I need an opinion on this:
My EOLO Edition of the 4020 has a Mac Address written on it that is different from the one that openwrt and gluon uses even before migration. In the bootloader it is labeled as macdsl and the one gluon uses is labeled as macwlan. The Mac Address that gluon uses is also found on the label but in the CWMP account (ID?) on my device.
Should i see this as an issue or, because the EU Retail devices don't have the Mac Address on the label, "ignore" this as not an issue?

rotanid:
i'd say ignore.
i have the retail edition, if someone provides me with a ar71xx and a ath79 image i could also test it, if necessary.

NeoRaider:
I think the current behavior is the best we can do, given the existence of different brandings of the same device, with different choices whether a label MAC address exists, and which it is.

PR #2441, opened 2022-03-30
ath79-generic: (re)add support for GL-AR150

AiyionPrime:
  - [x] Web interface
  - [x] TFTP
  - ~~Other: <specify>~~
  - [x] Must have working autoupdate
- Wired network
- LED mapping
  - Power/sys LED
  - Radio LEDs
    - [x] Should show activity
  - Switch port LEDs
- Outdoor devices only:

### DIFF
--- a/targets/ath79-generic
+++ b/targets/ath79-generic
@@ -89,6 +89,10 @@ device('enterasys-ws-ap3705', 'enterasys_ws-ap3705i', {
 
 -- GL.iNet
 
+device('gl.inet-gl-ar150', 'glinet_gl-ar150', {
+	factory = false,
+})
+
 device('gl.inet-gl-ar300m-lite', 'glinet_gl-ar300m-lite', {
 	factory = false,
 })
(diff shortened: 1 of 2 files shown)


AiyionPrime:
This migration works as expected.
Thanks to the 'Freifunk Hannover'-user, that provided us the test device.

targets/ath79-generic
```diff
@@ -89,6 +89,13 @@ device('enterasys-ws-ap3705', 'enterasys_ws-ap3705i', {
 
 -- GL.iNet
 
+device('gl.inet-gl-ar150', 'glinet_gl-ar150', {
+	factory = false,
+	manifest_aliases = {
+		'gl-ar150', -- upgrade from OpenWrt 19.07
```
NeoRaider, line 95: This can be dropped. The alias was only used for updates from LEDE 17.01, OpenWrt 18.06 already had the new name (and that is the oldest version that we still support upgrades from).

AiyionPrime: Thanks, resolved in 7a47873e06467233caeb059dbffcd86143d8241e.

PR #2443, opened 2022-03-31
ath79-generic: (re)add NanoStation M (XW)

AiyionPrime:
I'll get the test device on saturday afternoon.

  - [x] Web interface
  - [x] TFTP
  - ~~Other: <specify>~~
  - [x] Must have working autoupdate
- Wired network
- LED mapping
  - Power/sys LED
  - Radio LEDs
    - [x] Should show activity
  - Switch port LEDs
- Outdoor devices only:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -89,6 +89,7 @@ ath79-generic
 
 * Ubiquiti
 
+  - NanoStation M2/M5 (XW)
   - UniFi AC Lite
   - UniFi AC LR
   - UniFi AC Mesh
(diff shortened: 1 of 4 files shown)


AiyionPrime:
Rebased on mater to resolve conflicts. Migration worked without problems.

docs/user/supported_devices.rst
```diff
@@ -89,6 +89,7 @@ ath79-generic
 
 * Ubiquiti
 
+  - NanoStation M2/M5 XW
```
NeoRaider, line 92:
```suggestion
  - NanoStation M2/M5 (XW)
```
"XW" is basically a revision identifier, so let's put it in parentheses. The same is done in OpenWrt's Device Tree.

AiyionPrime: resolved in e97f0801073a40289e605505b8d86b312d41b5bc

PR #2448, opened 2022-04-02
ath79-generic: (re)add unifi-ac-{lite,lr}

AiyionPrime:
This is a followup of #2428.

> I think the test of the AC Mesh is sufficient

_@neoraider in hackint#gluon_

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -89,6 +89,7 @@ ath79-generic
 
 * Ubiquiti
 
+  - UniFi AC Lite
   - UniFi AC LR
   - UniFi AC Mesh
   - UniFi AP
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -89,6 +89,7 @@ ath79-generic
 
 * Ubiquiti
 
+  - UniFi AC LR
   - UniFi AC Mesh
   - UniFi AP
   - UniFi AP LR
--- a/targets/ath79-generic
+++ b/targets/ath79-generic
@@ -333,6 +333,11 @@ device('tp-link-wbs210-v2', 'tplink_wbs210-v2')
 
 -- Ubiquiti
 
+device('ubiquiti-unifi-ac-lite', 'ubnt_unifiac-lite', {
+	factory = false,
+	packages = ATH10K_PACKAGES_QCA9880,
+})
+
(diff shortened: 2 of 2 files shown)


package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua
```diff
@@ -35,6 +35,7 @@ function M.is_outdoor_device()
 		'tplink,cpe510-v3',
 		'tplink,eap225-outdoor-v1',
 		'tplink,wbs210-v2',
+		'ubnt,unifiac-lite',
```
NeoRaider, line 38: The AC Lite is an indoor device.

docs/user/supported_devices.rst
```diff
@@ -91,6 +91,7 @@ ath79-generic
 
   - UniFi AC Mesh
   - UniFi AP
+  - UniFi AP AC Lite
```
NeoRaider, line 94: For consistency with the other UniFi AC variants, this should be "UniFi AC Lite" without the "AP".

package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua
AiyionPrime: resolved in d3a2c657f912b1d7354c962fc91bcc5ba437f739

docs/user/supported_devices.rst
AiyionPrime: resolved in d3a2c657f912b1d7354c962fc91bcc5ba437f739

rotanid:
@NeoRaider why did you suggest to change the name of the device in the supported hardware list?
it was called the other way in our ar71xx docs and also in openwrt docs it is called like we did before...
https://github.com/freifunk-gluon/gluon/blob/v2021.1.x/docs/user/supported_devices.rst

NeoRaider:
@rotanid In OpenWrt it's consistently "UniFi AC" now, in both Device Trees and image definitions.

rotanid:
> @rotanid In OpenWrt it's consistently "UniFi AC" now, in both Device Trees and image definitions.

hm, ok, have only looked in the openwrt wiki and TOH.

we need to adjust this PR then, too : https://github.com/freifunk-gluon/gluon/pull/2439

PR #2449, opened 2022-04-02
ath79-generic: (re)add unifi-ap-ac-lr

AiyionPrime:
the second follw-up on #2428.

> AC LR can also be added without another test

_@neoraider in hackint#gluon_

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -91,6 +91,7 @@ ath79-generic
 
   - UniFi AC Mesh
   - UniFi AP
+  - UniFi AP AC LR
   - UniFi AP LR
   - UniFi AP PRO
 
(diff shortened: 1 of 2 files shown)


docs/user/supported_devices.rst
```diff
@@ -91,6 +91,7 @@ ath79-generic
 
   - UniFi AC Mesh
   - UniFi AP
+  - UniFi AP AC LR
```
NeoRaider, line 94: For consistency with the other UniFi AC variants, this should be "UniFi AC LR" without the "AP".

AiyionPrime: In ar71xx they were called this way and unifi sells them as such, i think; but maybe its just resellers.
https://github.com/freifunk-gluon/gluon/commit/003c90f0b20af7dc0f51893a3931ad8bada725d2#diff-569125591014883ee4f7ab0a04f63c29743d7eacdabf8cf031c4d9e73bf03e3dL117-L118

AiyionPrime:
closed in favor of #2448

PR #2453, opened 2022-04-02
ath79-generic: (re)add support for glinet_6416

AiyionPrime:
@rotanid received the images for testing

  - [ ] Web interface
  - [ ] TFTP
  - [ ] Other: <specify>
  - [ ] Must have working autoupdate
- Wired network
- LED mapping
  - Power/sys LED
  - Radio LEDs
    - [X] Should show activity
  - Switch port LEDs
- Outdoor devices only:

### DIFF
--- a/targets/ath79-generic
+++ b/targets/ath79-generic
@@ -93,6 +93,13 @@ device('enterasys-ws-ap3705', 'enterasys_ws-ap3705i', {
 
 -- GL.iNet
 
+device('gl.inet-6416', 'glinet_6416', {
+	factory = false,
+	manifest_aliases = {
+		'gl-inet-6416a-v1', -- Upgrade from OpenWrt 19.07
+	},
+})
+
 device('gl.inet-gl-ar150', 'glinet_gl-ar150', {
 	factory = false,
 })
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -32,6 +32,7 @@ ath79-generic
 
 * GL.iNet
 
+  - 6416A
   - GL-AR150
   - GL-AR300M-Lite
   - GL-AR750


rotanid:
flashed Freifunk Hannover ar71xx image vH21, sysupgrade to the test-image based on this PR, tested and checked the list in this PR.
didn't test autoupdate or flash from factory.

seems to be merge-ready

AiyionPrime:
Thank you both!

targets/ath79-generic
```diff
@@ -89,6 +89,13 @@ device('enterasys-ws-ap3705', 'enterasys_ws-ap3705i', {
 
 -- GL.iNet
 
+device('gl.inet-6416a-v1', 'glinet_6416', {
```
NeoRaider, line 92: According to the DTS, the image name should be "gl.inet-6416".

AiyionPrime: I really need to read up on the first field. Until now I'm just blindly guessing.
Is the rule just `lowercase of the dts' modelname and replace spaces with dashes`?

AiyionPrime: resolved and rebased to master in edc8d0a433571b84d026320be6a5263ba9b18b45

```diff
-device('gl.inet-6416a-v1', 'glinet_6416', {
+device('gl.inet-6416', 'glinet_6416', {
```

NeoRaider: > I really need to read up on the first field. Until now I'm just blindly guessing. Is the rule just `lowercase of the dts' modelname and replace spaces with dashes`?

The current algorithm is: String is lowercased. All characters that aren't alphanumeric, `.` or `+` are replaced with dashes. Then double dashes and dashes at the end are removed. "-." is replaced with "-". This is implemented here: https://github.com/freifunk-gluon/packages/blob/master/libs/libplatforminfo/src/common.h#L63

AiyionPrime: Thanks, that helped a lot.

PR #2455, opened 2022-04-02
ath79-nand: (re)add support for glinet_gl-ar300m

AiyionPrime:
@rotanid received the image for testing.

  - [ ] Web interface
  - [ ] TFTP
  - [ ] Other: <specify>
  - [ ] Must have working autoupdate
- Wired network
- LED mapping
  - Power/sys LED
  - Radio LEDs
    - [X] Should show activity
  - Switch port LEDs
- Outdoor devices only:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -99,6 +99,7 @@ ath79-nand
 
 * GL.iNet
 
+  - GL-AR300M
   - GL-AR750S
 
 brcm2708-bcm2708
--- a/targets/ath79-nand
+++ b/targets/ath79-nand
@@ -7,6 +7,10 @@ local ATH10K_PACKAGES_QCA9887 = {
 }
 
 
+device('gl.inet-gl-ar300m-nor', 'glinet_gl-ar300m-nor', {
+	factory = false,
+})
+
 device('gl.inet-gl-ar750s-nor', 'glinet_gl-ar750s-nor', {
 	factory = false,
 	packages = ATH10K_PACKAGES_QCA9887,


rotanid:
flashed Freifunk Hannover ar71xx image vH21, sysupgrade to the test-image based on this PR, tested and checked the list in this PR.
can't test autoupdate, won't test flash from factory.

seems to be merge-ready

AiyionPrime:
The image as part of the generic subtarget in ar71xx.
In order for it to become a "real" nand subtarget its necessary to have this sort of intermediate nor image. There is a similar device that awaits this migration as well, if I'm not mistaken:

> Changed subtarget in between, similar change in
commit https://github.com/freifunk-gluon/gluon/commit/bed66fa24856057eadd252eef8144df1c894112f ("ath79: move GL.iNet GL-AR750S to NAND subtarget")

NeoRaider:
> The image as part of the generic subtarget in ar71xx. In order for it to become a "real" nand subtarget its necessary to have this sort of intermediate nor image. There is a similar device that awaits this migration as well, if I'm not mistaken:
> 
> > Changed subtarget in between, similar change in
> > commit [bed66fa](https://github.com/freifunk-gluon/gluon/commit/bed66fa24856057eadd252eef8144df1c894112f) ("ath79: move GL.iNet GL-AR750S to NAND subtarget")

@AiyionPrime  It is also in the nand subtarget to include the NAND drivers in the kernel - I think this may even make sense when the NAND is not used for the firmware image at all.

Do you know if migration to the NAND image will actually be necessary at some point, or will the NOR image just continue to work? Do upgrades from NOR to NAND preserve configuration?

AiyionPrime:
@NeoRaider I have not seen movement regarding nor-nand migration of that device, yet.

Looking at growing image sizes, not using the 128MB NAND but only the 16MB NOR would be a waste.

From what I read, updates where definitely not possible from ath79 prior November 15th, 2019.
Upgrading would work most of the time but might result in booting the old system [1](https://openwrt.org/toh/gl.inet/gl-ar300m). Under which circumstances it does, I do not know yet.
Neither whether the config is kept while migrating from NOR to NAND.

That is not yet the scope of this PR; this does 'only' account for the ath79-migration.
I thought one would be enough per PR :D
I could extend the PR and build the NAND-image too, if @rotanid was interested in testing that as well.


As there are variants of the gl-ar300m, that do not come with NAND-storage, but only NOR, it's unlikely the NOR image will be gone soon for this particular device.

NeoRaider:
As long as Gluon fits in the NOR nicely, I'd like to avoid migration to NAND, as NAND is less reliable.

AiyionPrime:
fine with me.

PR #2457, opened 2022-04-03
gluon-status-page: Use <tbody> as browsers can add it automatically which broke the page

### LINKED ISSUE #2419: statuspage: duplicate meshneighbors
### Bug report

**What is the problem?**

The statuspage shows (local) mesh neighbors of both bands twice or thrice.

img

The js code that updates them throws an index error, which is only a symptom.

I first encountered the problem yesterday, when I built a PR on master.

I bisected the master branch, starting in with a commit in january I think:

> d728f17b8fc848b35e93eebb8e948798a1e4440a is the first bad commit

It appears the error is target independent, I encountered it in `ath79` and reproduced it in `ramips-mt76x8`.

The meshneighbors adresses (which exist once) should be shown once.

**Gluon Version:**

The error is present in the current master.
Our local nightly version does contain it, if somebody wants to see it in action:

Thanks to snuggels for providing a test device:
http://[2001:678:978:222:8ede:f9ff:fec3:7cde]/cgi-bin/status

**Site Configuration:**

The site config does not have an effect, tested with our current and one of a year ago.
Our nightly builds with https://github.com/freifunkh/site/commit/d6ce7f414b117f2b5a1b173b66cc3ccb1a73c741

**Custom patches:**

There are multiple patches in our site repo, whih are not related to the bug ither as stated above:

https://github.com/freifunkh/site/tree/master/patches

MyIgel:
This fixes #2419 by adding explicit `<tbody>` tags which browsers will add anyhow. An even better option would imho be to use selectors to select elements instead of hard references but that would be a bigger change which should be considered in the future.


AiyionPrime:
My currently available testnet appears to be unstable, but the PR does its job:

http://[2001:678:978:222:b64b:d6ff:fe20:2778]/cgi-bin/status

Don't mind "⬤ fa:1f:4d:92:2e:79" that's a cpe that was my first test device. Not sure what's wrong with it yet, but its likely unrelated to this PR.

AiyionPrime (approved):
Does resolve the issue.

package/gluon-status-page/javascript/status-page.js
```diff
@@ -192,8 +192,8 @@
 
 		div.style.display = '';
 		var table = document.getElementById('mesh-vpn-peers');
```
NeoRaider, line 194: We can avoid repeating `firstElementChild` a lot in the JS code by moving the `"mesh-vpn-peers"` ID from the `<table>` to the `<tbody>`.

package/gluon-status-page/javascript/status-page.js
```diff
@@ -444,8 +444,8 @@
 	}
 
 	function Neighbour(iface, addr, color, destroy) {
-		var th = iface.table.firstElementChild;
-		var el = iface.table.insertRow();
+		var th = iface.table.firstElementChild.firstElementChild;
+		var el = iface.table.firstElementChild.insertRow();
```
NeoRaider, line 448: We can avoid changing this code by assigning `el.firstElementChild.firstElementChild` to `info.table` in `Interface()`.

NeoRaider (changes requested):
Thanks for looking into this.

I have a few suggestions to reference the `tbody` instead of `table` in the JS to keep the code shorter. Might also make sense to rename the variables to `tbody` if we go with that.

package/gluon-status-page/javascript/status-page.js
MyIgel: That would then imho change the meaning of the code, i would prefer some way of
```
var tb = iface.table.firstElementChild;
var th = tb.firstElementChild;
var el = tb.insertRow();
```
which would make it clearer.

The better way would even be here to select the elements via css selectors.

package/gluon-status-page/javascript/status-page.js
MyIgel: same here, i think that the following could be clearer:

```javascript
var table = document.getElementById('mesh-vpn-peers');
var tbody = table.firstElementChild;
while (tbody.lastChild)
	tbody.removeChild(tbody.lastChild);
```

package/gluon-status-page/javascript/status-page.js
NeoRaider: Well, or
```js
var th = iface.tbody.firstElementChild;
var el = iface.tbody.insertRow();
```
removing the field `table` from `iface`, and assigning `tbody` instead. As far as I can tell, nothing else uses the field `table`.

package/gluon-status-page/javascript/status-page.js
NeoRaider: I don't see how that would be clearer than
```js
var tbody = document.getElementById('mesh-vpn-peers');
while (tbody.lastChild)
	tbody.removeChild(tbody.lastChild);
```
(with the ID moved in the HTML)

package/gluon-status-page/javascript/status-page.js
NeoRaider: I'm also fine with using more CSS selectors instead of all that `firstElementChild`, but that's mostly orthogonal to the question which elements to reference in the JS data structures.

package/gluon-status-page/javascript/status-page.js
MyIgel: Updated it to use id on the element so its basically a naming change

package/gluon-status-page/javascript/status-page.js
MyIgel: I added css selectors in the last commit, feel free to remove it before merging :D

MyIgel:
That should be it, @NeoRaider could you have a look at the changes?

herbetom:
I just build a firmware with those patches and i'm still seeing those two additional macs @MyIgel?

image

NeoRaider (approved):

PR #2459, opened 2022-04-05
gluon-setup-mode: add gluon-enter-setup-mode

### LINKED ISSUE #2412: Feature request: Script for easily reentering config mode on x86 devices
It would be nice if we could put the
```
uci set gluon-setup-mode.@setup_mode[0].enabled='1'
uci commit gluon-setup-mode
reboot
```
commands into a small script, e.g. /bin/cm.
That way on x86-based offloaders like the Futro you can easily re-enter config mode blindly by just connecting a USB keyboard.

mkg20001:
Script to enable config mode and reboot

Fixes #2412

### DIFF
--- a/package/gluon-setup-mode/files/etc/hotplug.d/button/50-gluon-setup-mode
+++ b/package/gluon-setup-mode/files/etc/hotplug.d/button/50-gluon-setup-mode
@@ -6,9 +6,7 @@ wait=3
 
 wait_setup_mode() {
 	sleep $wait
-	uci set 'gluon-setup-mode.@setup_mode[0].enabled=1'
-	uci commit gluon-setup-mode
-	reboot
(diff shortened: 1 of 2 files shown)


NeoRaider:
How about changing `/etc/hotplug.d/button/50-gluon-setup-mode` to use the new helper?

NeoRaider:
Thanks, the changes are looking good.

The "Fixes" tag in the commit message needs to be on a separate line, otherwise Github will not notice it (and it's also nicer for consistency with other commits).

NeoRaider:
Thanks, the changes are looking good.

The "Fixes" tag in the commit message needs to be on a separate line, otherwise Github will not notice it (and it's also nicer for consistency with other commits).

PR #2464, opened 2022-04-11
ipq40xx-mikrotik: add mikrotik-hap-ac2

mkg20001:
  - [ ] Web interface
  - [x] TFTP
  - ~~Other: <specify>~~
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [x] Should show activity
  - Switch port LEDs
- Outdoor devices only:

### DIFF
--- a/targets/ipq40xx-mikrotik
+++ b/targets/ipq40xx-mikrotik
@@ -1,6 +1,12 @@
 include 'mikrotik.inc'
 
+local ATH10K_PACKAGES_IPQ40XX = {}
+
 device('mikrotik-sxtsq-5-ac-rbsxtsqg-5acd', 'mikrotik_sxtsq-5-ac', {
 	factory = false,
 	aliases = {'mikrotik-discg-5acd'},
+})
+
+device('mikrotik-hap-ac2', 'mikrotik_hap-ac2', {
+	packages = ATH10K_PACKAGES_IPQ40XX,
 })
--- a/.github/filters.yml
+++ b/.github/filters.yml
@@ -56,6 +56,16 @@
     "targets/generic",
     "targets/targets.mk"
   ],
+  "ipq40xx-mikrotik": [
+    "targets/ipq40xx-mikrotik",
+    "modules",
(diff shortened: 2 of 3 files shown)


mkg20001:
Device is stuck in hungary. I'll finish up the todo list once the device comes into my hands again. :P

mkg20001:
Wifi was not working with switched ath10k firmware, is now working with whatever vanilla ships.

mkg20001:
Will do checklist once it's back from hungary (again).

// cc @5gbr can you remind me then? (or just do it yourself, there's a 802.11s mesh in r3 now)

mkg20001:
primary mac is the one from wan port (both are on the label). should this be the mac of wlan or is this ok?

targets/ipq40xx-mikrotik
```diff
@@ -1,6 +1,12 @@
...
 
+local ATH10K_PACKAGES_IPQ40XX = {}
+
 device('mikrotik-sxtsq-5-ac-rbsxtsqg-5acd', 'mikrotik_sxtsq-5-ac', {
 	factory = false,
 	aliases = {'mikrotik-discg-5acd'},
 })
+
+device('mikrotik-hap-ac2', 'mikrotik_hap-ac2', {
+  packages = ATH10K_PACKAGES_IPQ40XX,
```
AiyionPrime, line 11: indentation does not match `mikrotik-sxtsq-5-ac-rbsxtsqg-5acd` above
```suggestion
	packages = ATH10K_PACKAGES_IPQ40XX,
```

AiyionPrime (changes requested):
`mikrotik-hap-ac2` has already been migrated to DSA, so it will have OpenWrt support for a while.

@blocktrron I think his mac-issue is matching our default correctly.
Keep the mac on update, but use the 'correct' label mac on settings reset.

I'd say we can merge this after he fixes the indentation. Any objections?

AiyionPrime:
@mkg20001 merged in #2712, thanks!

PR #2465, opened 2022-04-12
gluon-mesh-vpn-fastd: Expose negotiated method

mweinelt:
Broad idea was to also bring this information to the status page.

Untested.


package/gluon-mesh-vpn-fastd/src/respondd.c
```diff
@@ -229,6 +229,11 @@ static bool get_peer_connection(struct json_object **ret, struct json_object *co
 		struct json_object *jso = json_object_new_double(established_time/1000.0);
 		json_object_set_serializer(jso, json_object_double_to_json_string, "%.3f", NULL);
 		json_object_object_add(*ret, "established", jso);
+
+		if (json_object_object_get_ex(connection, "method", &method)) {
+			const char* method_value = json_object_get_string(method);
+			json_object_object_add(*ret, "method", method_value);
```
NeoRaider, line 235: This doesn't look right - I guess we should add some `-Werror` flags to make the build fail for implicit pointer casts (from `const char*` to `struct json_object*` in this case).  You need a `json_object_get()` however to adjust the reference count.

```suggestion
			json_object_object_add(*ret, "method", json_object_get(method));
```

neocturne (dismissed):
One of your commit messages lost a character.

package/gluon-mesh-vpn-fastd/src/respondd.c
mweinelt: Applied, thanks.

package/gluon-status-page/javascript/status-page.js
```diff
@@ -210,7 +210,8 @@
 
 			var td = document.createElement('td');
 			if (peer[1] && peer[1].established != null)
-				td.textContent = _['connected'] + ' (' + formats.time(peer[1].established) + ')';
+				const method = peer[1].method ? ', ' + peer[1].method : '';
```
NeoRaider, line 213: So far we avoid ES6+ features in our code. I think it would be fine to allow these features by now, but it should be a deliberate decision to change this policy.

NeoRaider:
The PR description states that this is untested, is that right?

package/gluon-status-page/javascript/status-page.js
mweinelt: Haven't even thought about that, given that const is so widely supported and it is the ring thing to use in this scenario iirc.

https://caniuse.com/const

blocktrron:
Updated the PR and run-tested the result:

image

PR #2475, opened 2022-04-14
ath79-generic: (re)add support for tl-wr1043nd-v1

AiyionPrime:
@rotanid intends to test this device and has been provided access to the images.

  - [ ] Web interface
  - [ ] TFTP
  - [ ] Other: <specify>
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [X] Should show activity
  - Switch port LEDs
- Outdoor devices only:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -88,7 +88,7 @@ ath79-generic
   - TL-WDR4300 (v1)
   - TL-WR810N (v1)
   - TL-WR842N/ND (v3)
-  - TL-WR1043N/ND (v2, v3, v4)
+  - TL-WR1043N/ND (v1, v2, v3, v4)
   - WBS210 (v1.20)
   - WBS210 (v2.0)
 
--- a/targets/ath79-generic
+++ b/targets/ath79-generic
@@ -340,6 +340,13 @@ device('tp-link-tl-wr842n-v3', 'tplink_tl-wr842n-v3', {
 	},
 })
 
+device('tp-link-tl-wr1043nd-v1', 'tplink_tl-wr1043nd-v1', {
+	class = 'tiny', -- 32M ath9k
+	manifest_aliases = {
+		'tp-link-tl-wr1043n-nd-v1', -- upgrade from OpenWrt 19.07
+	},
+	packages = { 'zram-swap' },
+})
 device('tp-link-tl-wr1043nd-v2', 'tplink_tl-wr1043nd-v2', {
 	manifest_aliases = {
 		'tp-link-tl-wr1043n-nd-v2', -- upgrade from OpenWrt 19.07


rotanid:
flashed Freifunk Hannover ar71xx image vH22, sysupgrade to the test-image based on this PR, tested and checked the list in this PR.

BUT: i'm not sure about stability, at first glance it seems the device is really overloaded with the new firmware.
i think i will test a build based on Freifunk Altdorf site-ffa instead of Freifunk Hannover, in order to rule out issues related to the specific site.

rotanid:
built an image based on site-ffa and still not running stable, i'll stop the tests at this point, maybe someone else wants to try

rotanid:
@AiyionPrime i would suggest to merge this as BROKEN with a comment stating the instability, if nothing changes until our next Gluon release

AiyionPrime:
I thought about that last week, but was hesitating as devices would update to the broken firmware regardless of the flag.

So if this device was usable in ar71xx, but vastly decreased in stability in ath79, we might render a vast number of devices useless.

I think the device was unstable in ar71xx as well, but could you elaborate how the instability looks like now and if you experienced it in ar71xx as well?

I could verify your findings this week, as I got my hands on such a device as well.

NeoRaider:
RAM usage on 32MiB devices was critical in 19.07 already, so it's expected that it is worse in 21.02, and even worse in 22.03. So it might simply be the end of the line for these devices...

rotanid:
i will try to optimize further, like removing status page or something.
for me, a working up-to-date(!) image with stripped down features would be better than no image (and therefore devices running with old firmware)

AiyionPrime:
Careful, with leaving out the statuspage in particular; there were pretty strong opinions on that in the past.

But yeah, stripping down the image might be worthwhile.

AiyionPrime:
Speaking of, @bschelm has started to work on a mesh-only mode for our non-modded 841 devices.
Heated topic - I know, not my point:
The devices he worked with got ripped of like every unneccessary package.
Maybe he still has the list of things he dropped successfully.

bschelm:
@AiyionPrime: It's been a long time but I left away everything web-related (gluon-web*) and fastd. Would be good, to have a ssh-key on the device for config changes and a script that looks automatically for a mesh-network to join.

rotanid:
> Careful, with leaving out the statuspage in particular; there were pretty strong opinions on that in the past.

to save memory, we need to remove stuff which uses memory in idle or even worse when used/accessed. and to my knowledge, the status page is the only thing left before the node is completely useless.



> Speaking of, @bschelm has started to work on a mesh-only mode for our non-modded 841 devices. Heated topic - I know, not my point: The devices he worked with got ripped of like every unneccessary package

that doesn't help, as the optimizations for WR841 were mostly for saving flash space, which is not an issue with 8MB flash.
i already disabled fastd so it does not use memory.

rotanid:
i suggest to merge this only with BROKEN flag, as the memory usage has really gone up and the device tends to reboot due to OOM often.
this may even get worse when using openwrt 22.03

rotanid:
closing.
as discussed during our Gluon meetup today, we don't see a future for these kind of devices (8mb flash, 32mb memory)

PR #2476, opened 2022-04-14
ath79-generic: (re)add support for gl-usb150

AiyionPrime:
@rotanid intends to test this device.
He got the image a moment ago.

  - [ ] Web interface
  - [ ] TFTP
  - [ ] Other: <specify>
- Wired network
  - [ ] should support all network ports on the device - **device has no LAN ports**
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [X] Should show activity
  - Switch port LEDs - **device has no LAN ports**
- Outdoor devices only:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -36,6 +36,7 @@ ath79-generic
   - GL-AR150
   - GL-AR300M-Lite
   - GL-AR750
+  - GL-USB150 (Microuter)
 
 * Joy-IT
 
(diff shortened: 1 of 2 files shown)


rotanid:
flashed Freifunk Altdorf ar71xx image based on Gluon v2021.1.x , then sysupgrade to a test-image based on this PR, tested and checked the list in this PR.

seems to be merge-ready

AiyionPrime:
Thanks!

PR #2478, opened 2022-04-14
gluon-core: use copy of "WAN" role list for "single" default

NeoRaider:
Instead of using roles.wan directly as the default for roles.single,
create a copy of the table, so subsequent modifications of roles.single
don't affect roles.wan as well.

Fixes migration of Mesh-on-WAN status when no default for "single"
interfaces is set in site.conf.


herbetom:
Works for me. With this Patch i was able to upgrade a tl-wr1043n-v5 from a v2021.1.1 based firmware and Mesh on WAN stayed enabled. :+1:

PR #2479, opened 2022-04-14
push_pkg.sh: use scp protocol for openssh scp

lemoer:
Starting from OpenSSH 9.0p, scp started to use sftp in favor of the
scp protocol by default. As dropbear in OpenWrt currently does not
support sftp by default, we now use the fallback cli switch "-O"
to use the scp protocol for scp.

### DIFF
--- a/contrib/push_pkg.sh
+++ b/contrib/push_pkg.sh
@@ -127,7 +127,7 @@ while [ $# -gt 0 ]; do
 
 		# shellcheck disable=SC2029
 		if [ -n "$filename" ]; then
-			scp -P "${ssh_port}" "$feed/$filename" "root@${BL}${ssh_host}${BR}:/tmp/${filename}"
+			scp -O -P "${ssh_port}" "$feed/$filename" "root@${BL}${ssh_host}${BR}:/tmp/${filename}"
 			ssh -p "${ssh_port}" "root@${ssh_host}" "
 				set -e
 				echo Running opkg:


AiyionPrime:
tested with:
- OpenSSH_8.7p1
- OpenSSH_8.8p1
- OpenSSH_8.9p1
- OpenSSH_9.0p1

Maybe someone could test this with another distribution, just to make sure there are no debian/ubuntu/centos/... packaging-patches that mess with the version string you rely on.

NeoRaider:
I don't like the autodetection, seems too fragile. How about adding a flag to pass to `push_pkg.sh` instead?

lemoer:
> I don't like the autodetection, seems too fragile.

Well, I see your point. However, it's just a development tool. I would argue, it's not such a big deal if it would brick one day. At that point, we could fix it or decide that we have a better approach.

> How about adding a flag to pass to push_pkg.sh instead?

I am not really a fan of this. Things will crash with a more or less crappy error and people will start to debug things, they will find out that they have to add a flag to scp and finally they figure out that this can also be done by adding a flag to push_pkg.sh.

And also probably I would know about the flag, but would forget to specify it half of the runs. Then after few seconds the command will fail and and I have to call it again with an additional flag.

AiyionPrime:
> I don't like the autodetection, seems too fragile. How about adding a flag to pass to `push_pkg.sh` instead?

I share @lemoer's concerns about the manual flag I think it would hinder the workflow.
It's likely the `-O`-flag will stay for the foreseeable future.

I think it would suffice to verify it currently works on other distributions.
Any further ideas which should be tested?

lemoer:
Now, it is used by default.

mkg20001 (approved):

Issue #2393, opened 2022-02-11
Whishlist: Config-Mode UI for Interface Roles Feature 


lemoer:
*(At the 2022-02 Gluon meeting it was decided to open this Issue)*.

With #2372 gluon will officially start supporting the assignment of *roles* to network interfaces. This means that there is no longer a fixed mapping between network interface and the functionality assigned to that interface. Instead, each network interface is now assigned a user-modifiable list of *roles*. Since this list is not only user-modifiable, but also allows for additional user-defined interfaces, Gluon's network configuration is much more flexible than before.

However, only the backend part is implemented in #2372. The user interface still needs to be discussed. Since there have been numerous discussions (in many issues) over a long period of time, this issue is intended to work towards a coordinated wish list (that can be implemented).

So far, I have collected links to related issues:
- https://github.com/freifunk-gluon/gluon/issues/284
- https://github.com/freifunk-gluon/gluon/issues/1104
- https://github.com/freifunk-gluon/gluon/issues/1174
- https://github.com/freifunk-gluon/gluon/issues/1208
- https://github.com/freifunk-gluon/gluon/issues/1564
- https://github.com/freifunk-gluon/gluon/issues/1670
- https://github.com/freifunk-gluon/gluon/issues/1987
- https://github.com/freifunk-gluon/gluon/issues/2295

### How you can play around with role support from #2372:

The interface to role assignment is defined in /etc/config/gluon:

```
root@router:~# cat /etc/config/gluon
[...]
config interface 'iface_wan'
	option name '/wan'
	list role 'uplink'

config interface 'iface_lan_vlan200'
	option name 'eth0.43'
	list role 'client'
```

There are two ways of specifying the `name` option an interface here:
- The `/` in at the beginning of `/wan` indicates that the actual interface name should be read from `/lib/gluon/core/sysconfig/wan_ifname`.
- `eth0.43` directly refers to the interface name.

VLAN devices like `eth0.43` will be working directly (without explicit configuration of a vlan interface), since netifd creates vlan interfaces implicitly on demand.

Currently the roles `uplink`, `client` and `mesh` exist.

After /etc/config/gluon has been modified, `gluon-reconfigure` has to be called to regenerate `/etc/config/network`. After that, the network has to be restarted using `/etc/init.d/network restart`.

belzebub40k:
[...]

PR #2480, opened 2022-04-15
Config-Mode UI for Interface Role Assignment

lemoer:
First steps towards https://github.com/freifunk-gluon/gluon/issues/2393.

Next steps:
- [x] Allow creation of vlan interfaces.
- [x] Allow deletion of vlan interfaces.
- [ ] Allow expanding /lan, /wan, /single interfaces.
- [ ] Create sections for additional ethernet interfaces in /etc/config/gluon if they do not exist.
- [x] Make *Uplink* and *Client* role mutually exclusive.
- [ ] Test this with a DSA device.
- [ ] Test what happens if we create vlans on swconfig interfaces.
- [ ] Maybe forbidd creating vlans on swconfig interfaces.
- [ ] Readd i18n.


belzebub40k:
@lemoer I'm working on implementing the bridge isolation feature for mesh links (https://github.com/freifunk-gluon/gluon/pull/2484) and @AiyionPrime suggested that I mention it here as those features are related to the same issue.

lemoer:
Currently this PR looks like this: 
image

lemoer:
Currently the workflow to create a VLAN looks like this:

lemoer:
Roles "Uplink" and "Client" are now mutually exclusive:

lemoer:
I will continue working on this, if #2526 is merged/discussed/decided.

lemoer:
Discussion from the gluon dev meeting 2022-03:

- Maybe build the UI without adding a possibility to expand /lan, /wan and /single interfaces in the first place.
- In the second instance:
    - By default, /lan refers to lan1, lan2, lan3, lan4.
    - If an interface for lan1 is defined, then:
        - lan1 is removed from /lan.
        - So this means that /lan only refers to lan2, lan3, lan4.

T-X:
Looks nice! I'm wondering if there should be some warning + explanation from the UI if the user selects uplink + mesh on an interface. I assume that users are usually not aware why/when this configuration could make their WiFi experience incredibly slow.

lemoer:
In order to simplify things, this is closed in favor of https://github.com/freifunk-gluon/gluon/pull/2688.

lemoer:
> I assume that users are usually not aware why/when this configuration could make their WiFi experience incredibly slow.

@T-X : I do not really understand? This PR is only about the configuration of wired interfaces. How does this influence the WiFi experience?

T-X:
@lemoer it'll influence the WiFi experience when someone uses a standard home router on the uplink side which also has WiFi enabled. Which I guess is a typical scenario? If then using the uplink+mesh roles at the same time then the home wifi router will usually send all the layer 2 broadcast and routing protocol management packets at a slow 1MBit/s wifi bitrate. Depending on the overall mesh network size this could lead to high airtime usage on whatever channel the home router uses.

But anyway, we had the "mesh-on-wan" checkbox before, ready to use in parallel with mesh-vpn. So wouldn't be a "regression", was just wondering if there could be an easy way to give a hint about that potential issue while the UI is updated.

PR #2481, opened 2022-04-15
ramips-mt76x8: add support for TP-Link Archer C20 v5

eberhab:
With this PR I would like to propose TP-Link Archer C20 v5 support. Owning a C20 v5, based on @herbetom's commit, I did the device-Integration-checklist[1], please find it here:

  - [X] Web interface
  - [ ] TFTP (possible[2] but not tested)
  - [ ] Other: TP-Link recovery mode (possible[2] but not tested)
  - [X] Must have working sysupgrade (`sysupgrade sysupgrade.bin` works)
        -> Gluon profile name: tp-link-archer-c20-v5 (device name in ramips-mt76x8 target file)
        -> Autoupdater image name: tp-link-archer-c20-v5 (via lua get_image_name())
      -> /lib/gluon/core/sysconfig/primary_mac contains primary mac on device label
- Wired network
        -> Wan: eth0.2 (SW Port 0), Lan: eth0.1 (SW Ports 1-4)
- LED mapping
  - Power/system LED
          - The device has no "sys" LED, so instead the power LED is flashing
  - Radio LEDs
    - [X] Should show activity
  - Switch port LEDs
          - only one LED present

One other thing to mention with this device, is that OpenWRT cannot build factory firmware[2], only sysupgrade. In order to create a first-install/ factory image one needs to download the official tp-link firmware[3] and prepend the first 131584 bytes to the sysupgrade image. This is one extra step and very easy to do. It is explained in more detail in [2].

Links:
[1] https://github.com/freifunk-gluon/gluon/wiki/Device-Integration-checklist
[2] https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=86e7353bff2a5de257de8ec62e782f016eed143c
[3] https://static.tp-link.com/2021/202101/20210127/Archer%20C20(US)_V5_201214.zip

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -262,6 +262,7 @@ ramips-mt76x8
 
 * TP-Link
 
+  - Archer C20 (v5)
   - Archer C50 (v3)
   - Archer C50 (v4)
   - RE200 (v2)
--- a/targets/ramips-mt76x8
+++ b/targets/ramips-mt76x8
@@ -32,6 +32,10 @@ device('ravpower-rp-wd009', 'ravpower_rp-wd009')
 
 -- TP-Link
 
+device('tp-link-archer-c20-v5', 'tplink_archer-c20-v5', {
+	factory = false,
+})
+
 device('tp-link-archer-c50-v3', 'tplink_archer-c50-v3', {
 	factory = false,
 	extra_images = {


eberhab:
Yes, 5ghz mesh is working on the v5. According to the Openwrt wiki the v5
and v4 seem to have the same 5GHz radio (MT7610E). So I would say give it a
try with the v4.

What I can do is bit longer term 5GHz mesh stability test on my test device
if that's helpful?

PR #2483, opened 2022-04-17
ath79-generic: (re)add support for tl-wr1043n-v5

herbetom:
Gone due to commit 45c84a117bf8 ("ar71xx: drop target")

  - [x] Web interface
  - [x] TFTP
  - Other: <specify>
  - [x] When re-adding a device that was supported by an earlier version of Gluon, a
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [x] Should show activity
  - Switch port LEDs
- Outdoor devices only:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -88,7 +88,7 @@ ath79-generic
   - TL-WDR4300 (v1)
   - TL-WR810N (v1)
   - TL-WR842N/ND (v3)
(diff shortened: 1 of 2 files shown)


rotanid:
@herbetom test checklist looks finished, is it ready for merge from your side?

herbetom:
I'm currently not aware of any problems which are specific to this device. So yes, i think it's ready for merge from my side.

Issue #1104, opened 2017-04-24
[enhancement] Seperate mesh-on-LAN links


MPW1412:
Hello,

in the very typical setup of three routers connected over mesh-on-lan like this

A <--LAN--> B <--LAN-->C

Batman sees only one hop between A and C instead of two, because the packages a switched directly in the build in switch of router B and don't go through the CPU and Batman there.

I want to suggest to change this behavior. Although there might be advantages in this configuration in special situations, it generally distorts the topology of the network and leads to ”all routers see all routers over one hop“ in bigger setups, that are connected over lan. 

The behavior of Linux software bridges can be altered by this ebtables command (typically used in L2TP Gateway setups, where all vpn interfaces are connected through one bridge):

`ebtables -A FORWARD --logical-in $IFACE -j DROP`

This forces all packages to cycle through the CPU and by so through batman. So the topology of cable-connected networks would be correctly represented in Batman. 

I would like to start a discussion about weather this is possible for the hardware switches aswell and weather we might want to make this the standard behavior in Gluon?

Regards,
Matthias

NeoRaider:
Another argument in favour of this proposal:
* The hardware switches of cheap nodes don't speak STP, so a switch loop would currently break the network.

Another argument for keeping the current behaviour:
* Gluon doesn't touch the switch configuration at all at the moment. Separating the LAN ports introduces hardware-specific configuration I'd like to avoid for now.

We have plans to encapsulate wired mesh traffic in VXLANs (and I'm working on patches for that); I'd like to defer this discussion at least until we have sorted that out.

Adorfer:
it would be a very usefule enhancement for situations like

![grafik](https://cloud.githubusercontent.com/assets/10302493/25363548/a809c82c-295b-11e7-8da2-484816f7a1f6.png)

jplitza:
[We experimented with gathering neighbor information from LLDP](https://github.com/FreifunkBremen/ffhb-packages/tree/master/gluon-respondd-lldp) instead of batman-adv, which would combine the "real" network topology with keeping the switching behavior (which I would strongly favor, as it is far more performant and resilient). In the end, you are apparently arguing about the ways of displaying information, not about network topology. So gather the right information and display it adequately instead of changing the topology.

Adorfer:
and this is the reason why I do not want to connect large number of nodes via layers switches. 
If you look at it just from visual perspective (or consider others to do that): that's wrong. 
this is about OGM on 100TX.

[...]

PR #2484, opened 2022-04-19
gluon-core: add isolated option to gluon_wired proto

belzebub40k:
As discussed in https://github.com/freifunk-gluon/gluon/issues/2393 and https://github.com/freifunk-gluon/gluon/issues/1104 it would be nice if the individual switch ports could be isolated and connected nodes would only see nodes they are directly attached to. With devices already migrated to DSA this is fairly easy as the isolated flag can be set on the bridged interfaces.

As I could not find any option in OpenWrt to set the isolated flag on bridge ports I added it to the gluon_wired proto which takes care of setting all ports to isolated in interfaces with type=bridge and proto=gluon_wired.

**TODO**
- [x] test on DSA device
- [x] test on swconfig device (check that nothing breaks)
- [ ] ~add config mode option (maybe wait for #2480?)~

### DIFF
--- a/package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
+++ b/package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
@@ -56,6 +56,12 @@ proto_gluon_wired_setup() {
         proto_init_update "$ifname" 1
         proto_send_update "$config"
 
+        if [ -d "/sys/devices/virtual/net/${ifname}/bridge" ]; then
+                for brif in "/sys/class/net/${ifname}/brif/"* ; do
+                        echo 1 > "${brif}/isolated"
+                done
+        fi
+
         if [ "$vxlan" -eq 1 ]; then
                 meshif="vx_$config"
 


belzebub40k:
@AiyionPrime at the moment the option is not really configurable as the upgrade scripts will overwrite the isolated value in mesh_other on every execution. I initially thought about adding the option to the interface entries in the gluon uci section but at the moment I don't see a reason why one would like to not have the ports separated.

If it should be configurable by the user I see two options:
1. Make the option a per interface setting and add it in the interface entries in the gluon uci section.
2. Add a new entry in the gluon uci section for global wired mesh options.

Option 1. sounds more complicated as I don't see an easy way to access the gluon config section within the gluon_wired.sh script. Option 2. should be easy as we can handle the setting of the isolated option in the upgrade script.

T-X:
> @AiyionPrime at the moment the option is not really configurable as the upgrade scripts will overwrite the isolated value in mesh_other on every execution. I initially thought about adding the option to the interface entries in the gluon uci section but at the moment I don't see a reason why one would like to not have the ports separated.

One potential advantage of no bridge port isolation over setting it is speed. Frames being forwarded through the switch in hardware, with no forwarding/routing through userspace.

However I think there are very good reasons to have it and have it enabled by default (some of them were already mentioned in the referenced tickets):

* safer - less risk of loops through miscabling, similar argument as for vxlan being enabled by default
* smaller wired broadcast domain, less protocol overhead: batman-adv overhead increases squared to the number of neighbours on an interface, so this helps on larger wired setups (and I think this was similar for BABEL and OLSR, too)
* potentially better routing decisions due to every wired forwarding handled by routing protocol
* potentially better topology view on map

In my opinion it could be just enabled by default with no user configuration option for now. At least until someone comes along with a real world setup where s/he would really need to disable it again.

Maybe add some of the pro and con arguments to the git commit message?

belzebub40k:
@T-X Thanks for the feedback after a night of sleep I also thought that somebody might use a node as dumb switch to connect several nodes. I put some more work into the PR to allow setting the option per mesh interface entry in /etc/config/gluon. I also reverted the default behavior to not setting the isolated flag as setups like the describe might already exist in the wild. Maybe we can only set it as default for new nodes.

lemoer:
I agree with @T-X, that the behaviour should **not** be configurable. Let us just enable port isolation on all ports with *Mesh* role unconditionally.

In addition to all points mentioned by @T-X, I would like to add:
- It doesn't brick existing setups.
- It's a low level network configuration option. Unless it is really necessary, we usually avoid making low-level network configuration options available to users. It keeps UI simple, it keeps code simple and it also avoids misconfigurations.

@belzebub40k: Regarding the "dump switch" example you mentioned. I am not really sure what is meant, but I understand your "dump switch" the following way: You want to have certain ports (that are assigned to none of the existing roles *Uplink*, *Mesh*, *Client*) bridged together. Such a setup was never supported in gluon so far, there is no existing setup to be bricked, is there? Or am I getting you wrong?

AiyionPrime:
Can someone estimate how much this is gonna impact performance?
Not that I don't like the feature, but

> potentially better routing decisions due to every wired forwarding handled by routing protocol

sounds like we'd be losing a lot cycles to the cpu doing the switches job?

How are small devices (not the erx you mnetioned before) handling that change?

belzebub40k:
@lemoer I agree on the part that it should not be necessary to make it configurable via Config-Mode but I still think it should be configurable from the command line as it might (partially) break certain existing setups where a full mesh is desired. This is also what I meant with dum**b** switch where all connected nodes should be able to talk to each other. As the the amount of devices using DSA is very low I don't expect to break any setups with the next release but when more devices are migrated the behavior of nodes will change from they can see all connected nodes to only seeing directly attached nodes.

@AiyionPrime We configure the nodes in our wireless backbone like for a long time now by creating individual VLANs per mesh-link to prevent full meshes. For example this node with 10 mesh-links is a TL-WR1043N v2 with a 720MHz MIPS CPU and 64MB of RAM and the load looks normal. https://map.ffmwu.org/#!/de/map/e894f6696abc

NeoRaider:
I don't see a reason to allow disabling isolation.

The proposed implementation might be okay as a temporary solution, but this should be handled in netifd in the long term instead of writing to sysfs directly.

belzebub40k:
Sorry, didn't had time to look into netifd until now. I checked it and this is beyond my very basic C skills. I someone wants to implement it upstream in netifd I'm fine with it and the PR can be closed.

Regarding this PR as the consensus seems to be to not have it configurable I dropped the isolated option completely and every bridge with gluon_wired will set all ports to isolated but I see one potential downside with this. If someone sets roles uplink and mesh the clients communicating in br-wan will also be isolated. This is not an problem if it is only the single WAN port every device has but if you also set role uplink on the LAN ports those ports will all be isolated in br-wan.

belzebub40k:
Just noticed something has broken. For some reason the isolated flag no longer has an effect on my test device.

AiyionPrime:
@belzebub40k which is you test device, and what exactly does not work anymore?

belzebub40k:
@AiyionPrime It's an ERX-SFP. With an OpenWrt 22.03 base the isolation flag did not have an effect anymore. I tried to downgrade to an image bases on 21.02 but since then it does not keep its config. Did not have time to look into it.

herbetom:
At least with an NanoPi R2S and an Extreme Networks WS-AP3825i this PR also works with Gluon based on OpenWRT 22.03.

We're already using this to configure multiple Mesh VLANs on a WS-AP3825i so that not every device has to mesh with all others:
image

I think isolation should be the enabled for every interface (`mesh`, `uplink`) except for `client` interfaces.

I'm not completely sure if there should be an option to disable isolation. The only kinda valid case i'm seeing is using the gluon node as a switch for the Uplink Network. But that's nothing i would really recommend.

AiyionPrime:
I built yesterdays master with this patch on top, applies fine.
After adding a few debug echos, logread contains the hint, that br-wan's interface was marked as isolated.

The option did not hold, but after setting it after installing the bridge command using that it does. Not sure if it does have the intended effect though.
I'll setup a smaller testsetup.
But can confirm, something is indeed not working right now.
As @herbetom observed earlier, my test devices are ath79, his are not.

Issue #2413, opened 2022-03-21
ar71xx - ath79 migration progress


AiyionPrime:
@NeoRaider's recent contribution of the network regeneration got merged (thanks to all involved!)
and if I'm not mistaken, we've agreed on testing the devices that were gone again, as if we'd be adding new support for them.

That's arguably an effort, but hopefully one divided on several shoulders.

Below is a list of devices that got dropped with the target ar71xx, which hadn't been marked as tiny back then and was not within the mikrotik subtarget.
I think that would be a solid start for the migration.

I'll order them in ongoing, todo and done. Please sue `ctrl+f` to find the device you are interested in.

Testing the devices is not only flashing from stock and via sysupgrade, but sysupgrading from an old ar71xx image as well.
You do not need to build an ar71xx image for this, as hanover has them available [here](http://firmware.ffh.zone/vH21/) and will provide you with access to its VPN infrastructure. If possible let us know when you're done with the key.

Testing will result in failures for some devices due to unhandled migrations for now.

Thanks for your help,
Aiyion.

**in progress**

| subtarget | tester         | PR      | device                              |
|-----------|----------------|---------|-------------------------------------|

**todo**

@rotanid just showed me a (shorter?) [list of devices to migrate](https://md.darmstadt.ccc.de/gluon-ath79-migration?view), which includes tiny targets as well.
I extracted the parts, where someone announced he would be willing to test the device.
For now I wont update from that list on a regular basis, but just this once.
If somebody wants to test a device, submit a PR draft to gluon and I'll update this accordingly.
Thanks for the hint.

| subtarget | wanted to test | devices | device                              |
|-----------|----------------|---------|-------------------------------------|

**aborted**

These devices were too hard to get a hold on in the last three years.
Migration is not feasible anymore.

Who ever lacks a device below, feel free to just re-add it; migration phase is over.

| subtarget | wanted to test | devices | device                              |
|-----------|----------------|---------|-------------------------------------|
| generic   |                |         | 8devices-carambola2-board           |
| generic   |                |         | _allnet-all0315_                    |
| generic   | [Nordmann](https://forum.freifunk.net/u/nordmann) |         | _buffalo-wzr-hp-g300nh2_              |
| generic   |                |         | buffalo-wzr-hp-g450h                |
| generic   |                |         | gl-inet-6408a-v1                    |
| generic   |                |         | meraki-mr12 (was broken)            |
| generic   |                |         | meraki-mr16 (was broken)            |
[...]

PR #2485, opened 2022-04-19
ath79-generic: (re)add support for wndr3700v2

AiyionPrime:
[rabbit83](https://forum.freifunk.net/u/rabbit83/) intends to test this device.
He built and ran his own build based on 2021.1.1 a while ago, building on master failed for his local site, so he was provided an image.

  - [x] Web interface
  - [x] TFTP
  - ~~Other: <specify>~~
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [x] Should show activity
  - Switch port LEDs
- Outdoor devices only:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -41,6 +41,10 @@ ath79-generic
 
   - JT-OR750i
 
+* Netgear
+
+  - WNDR3700 (v2)
+
 * OCEDO
 
   - Raccoon
--- a/targets/ath79-generic
+++ b/targets/ath79-generic
@@ -121,6 +121,15 @@ device('joy-it-jt-or750i', 'joyit_jt-or750i', {
 })
 
 
+-- Netgear
+device('netgear-wndr3700-v2', 'netgear_wndr3700-v2', {
+	manifest_aliases = {
+		'netgear-wndr3700v2', -- Upgrade from OpenWrt 19.07
+	},
+	factory_ext = '.img',
+})
+
+
 -- OCEDO
 
 device('ocedo-raccoon', 'ocedo_raccoon', {


rabbit83ka:
Thanks @AiyionPrime, I'll test the images provided by you and give feedback here.

rabbit83ka:
Hi, I conducted the above tests:

- Flash from stock works via Webinterface (TFTP also possible)
- Sysupgrade from ar71 works, configuration is kept, "firstboot" resets config
- WPS leads to config mode
- Primary MAC matches the label on the bottom of the device
- WAN/LAN-ports assigned correctly
- Wireless works on both 2.4/5 GHz
- LEDs are working

AiyionPrime:
@rabbit83ka could look into the `get_image_name()` output missing above;
But I'm pretty confident `Netgear WNDR3700 v2` from [1]
translates to `netgear-wndr3700-v2` in [2]

[1] https://github.com/openwrt/openwrt/blob/0d2d52df69e91fe2f6e3e65750c0728fc45f5ab2/target/linux/ath79/dts/ar7161_netgear_wndr3700-v2.dts#L7
[2] https://github.com/freifunk-gluon/gluon/pull/2485/files#diff-8bd1ca94227cccf7c95011ebe4f9d1b7ae82d4c1495d27c12532af82df0d1312R125

AiyionPrime:
@rabbit83ka Thanks for your effort, somebody will merge/review this soon enough.
Do you have any other devices you want an image for listed as todo in #2413?

rabbit83ka:
@AiyionPrime I don't have another device from #2413. However, I have an [RT-AC51U](https://openwrt.org/toh/asus/rt-ac51u) which is currently "broken" due to 5 GHz and not-working LEDs. Both are fixed in OpenWRT meanwhile. I'm just testing an own gluon build and open a PR if everything works.

AiyionPrime:
Then thanks again for the testing.

PR #2487, opened 2022-04-21
ath79-generic: (re)add support for nanostation-loco-m

AiyionPrime:
@rotanid intends to test this device.

  - [ ] Web interface
  - [ ] TFTP
  - [ ] Other: <specify>
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [ ] Should show activity
  - Switch port LEDs
- Outdoor devices only (wasn't in ar71xx, but is an outdoor device - supposedly not in ar71xx because it has only one port):


rotanid:
flashed Freifunk Altdorf ar71xx image based on Gluon v2021.1.x , then sysupgrade to a test-image based on this PR, tested and checked the list in this PR.

several issues appeared:

1. the lua command outputs: ubiquiti-nanostation-loco-m-xm
so there's in issue, right? @NeoRaider @blocktrron 
2. the devices case has no mac address on it, so primary mac cant be verified?
3. the device LEDs don't show wifi activity, but this may also be the case with ar71xx, we haven't checked
4. stability. as with the TP-Link WR1043ND v1 , the device tends to reboot under some load (like accessing the status page), it didn't reboot though when doing SSH-stuff or doing Wifi-Traffic , maybe removing the status-page would be a nice compromise (also for the 1043v1) , i will test further

AiyionPrime:
Point one is my fault, not sure what dts file I looked at.
Translating from `Ubiquiti Nanostation Loco M (XM)` it should be `ubiquiti-nanostation-loco-m-xm`. I'lö fix that in a minute, sorry.

The mac adress could be verified using the stock firmware; or at least using a newly installed ar71xx vs a newly installed ath79 image and check whether they at least match.

I have not verified it yet, but at first glance it appears the device has board leds defined, but lacks the rssi package in openwrt.
I'm in a hurry right now, but thats something I can patch upstream, I think.
At least I saw someone doing a similar patch for other? ubiqity devices a while ago.

with number four I cannot help; looking at it, it might even be worthwile not to attemt to fix three :D

AiyionPrime:
Just resolved the first point I hope.

Regarding the rssi package I think I am wrong, Its added in the Makefile:
https://github.com/openwrt/openwrt/blame/46980294f66ab07b79f9037758a7ea94072340df/target/linux/ath79/image/generic-ubnt.mk#L285

Not sure why they would not work.

mbaumga:
> Regarding the rssi package I think I am wrong, Its added in the Makefile: https://github.com/openwrt/openwrt/blame/46980294f66ab07b79f9037758a7ea94072340df/target/linux/ath79/image/generic-ubnt.mk#L285
> 
> Not sure why they would not work.

I'll clarify a bit: I can see the 4 LED indicators (Red, Amber, Green, Green) Lighting up without issues, so overall connection is being indicated, but there is no kind of blinking to show traffic being transfered, so if only indicating connectivity is sufficient, I think we can check the Wifi LED stuff.

What I just saw while flashing the second Loco was that the "config mode blinking sequence" isn't working. My fault for overseeing this. We should uncheck this one again.

rotanid:
i just provided @mbaumga with an updated image using the fixed naming and some additional strip-down.
i suggest to test the correct naming first and afterwards compare three things with an ar71xx flashed device: 

1. the MAC address
2. wifi activity LED
3. config mode LED blinking

AiyionPrime:
Do the leds in the bar of a cpe blink (in normal mode)? I always thought of them as rssi?

rotanid:
@AiyionPrime @NeoRaider @blocktrron do you have an idea why the config mode LED blink sequence does not work for this device on ath79?

AiyionPrime:
I think I have a guess, that neoraider or blocktrron could confirm or deny:

This is the dtsi file, thats included for the nanostation loco xm, which contains led assignments, that are not overwritten further down the dts tree.

https://github.com/openwrt/openwrt/blob/master/target/linux/ath79/dts/ar724x_ubnt_xm_outdoor.dtsi

In setup-mode gluon assigns the variable status_led through calling get_status_led() for targets that provide the function in their respectiv diag.sh or  eg for ath79 which does not contain such a diag.sh file in https://github.com/openwrt/openwrt/tree/master/target/linux/ath79/base-files/etc fallback to the assignment of the led referenced by `$running`.

https://github.com/freifunk-gluon/gluon/blob/099d2bd693a3606f0262c54f79c4f35024fef464/package/gluon-setup-mode/files/lib/gluon/setup-mode/rc.d/S96led#L10-L14

That variable in turn is defined in https://github.com/openwrt/openwrt/blob/ec6293febc244d187e71a6e54f44920be679cde4/package/base-files/files/etc/diag.sh#L8

`running="$(get_dt_led running)"` which calls a funcrion meant to extract the running led alias from the respective devices devicetree.

The (faulty) assumption in our S96led is that ath79 devices have an alias called running in their dts.

That simply not true for the given device and likely not true for similar devices like some bullet variants as well, but that's just a wild guess.

Looking through other targets diag.sh override I found it common to assign the status led another one, likely boot.

My suggestion would be to add a second fallback to S96led to set the LED for this device, as the dts is rightfully not stating running leds.

A second option would be to add a diag.sh for the ath79 target upstream and do this fallback magic there like done in other targets. Maybe an overkill.

The second option would be to extend duggestion one, and not only add a second fallback seeting status_led to boot, but to provide even more aliases in an order, ike if not even boot was defined, take the `failsafe` led.

For now I'd add a commit to this PR that goes with option one, so on of you (rotanid, mbaumga) can verify with the compiled image that I'm not missing something big time.

mbaumga:
Config mode LED sequence now works. Thanks for the fix/workaround.

rotanid:
we suggest to merge this only with BROKEN flag, as the memory usage has really gone up and the device tends to reboot due to OOM often.
this may even get worse when using openwrt 22.03

AiyionPrime:
No strong opinions on this; BROKEN would be the least imho.
Not sure if we do users a favor upgrading these to ath79 at all.

PR #2493, opened 2022-04-26

blocktrron:


NeoRaider:
Why another 21.02 update?

blocktrron:
This fixes an issue where iwinfo would not return a correct modelist, which breaks distribution of multi-band radios.

PR #2495, opened 2022-04-30
mpc85xx-p1020: add Extreme Networks WS-AP3825i

herbetom:
  - Web interface
  - [x] TFTP
  - Other: <specify>
- Wired network
         LAN 1 is WAN (eth1)
         LAN 2 is LAN (eth0)
         It is possible to supply the AP with power via either of the two ports. However, there is no seamless failover. Additionally there is a DC-IN.
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [x] Should show activity
  - Switch port LEDs
- Outdoor devices only:

The instructions from the initial OpenWRT commit didn't work for me. I've written down some that should work: https://forum.darmstadt.freifunk.net/t/flashing-of-the-extreme-networks-ws-ap3825i/923

cc @blocktrron

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -233,6 +233,10 @@ mpc85xx-p1020
 
   - WS-AP3710i
 
+* Extreme Networks
+
+  - WS-AP3825i
+
 * OCEDO
 
   - Panda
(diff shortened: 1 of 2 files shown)


rotanid:
i read the flashing instructions an have a question:
is the serial reachable without opening the case ?

herbetom:
Yes. There are three RJ45 Ports on that AP. One Serial Console and two Ethernet Ports.

I thought that was kinda implied with "Cisco Style RJ45 Serial Cable to the AP".

PR #2500, opened 2022-05-04
ramips-mt76x8: add support for TP-Link Archer C20 v4

AiyionPrime:
@gerhardt tracked this device for a while now and lent me one to add it in gluon as well.

The device is expected to works good as revision five, which was added a few weeks ago.

Flashing the device must only be done via tftp from stock, if flashed via their UI the router is soft-bricked.

Test is scheduled for tomorrow 09:00.

  - ~~Web interface~~ Does softbrick the device
  - [x] TFTP
  - ~~Other: <specify>~~
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [x] Should show activity
  - Switch port LEDs
- Outdoor devices only:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -323,7 +323,7 @@ ramips-mt76x8
 
 * TP-Link
 
-  - Archer C20 (v5)
+  - Archer C20 (v4, v5)
   - Archer C50 (v3)
   - Archer C50 (v4)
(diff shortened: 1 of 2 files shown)


AiyionPrime:
Just like revision 5, this one does work without obvious problems.

@blocktrron any objections?

AiyionPrime:
That was fast, thank you very much.

PR #2502, opened 2022-05-05
gluon-status-page: fix mesh-vpn section for wg

### LINKED ISSUE #2498: Statuspage WireGuard connected NaN
### Bug report

**What is the problem?**

The Statuspage of WireGuard-nodes in Hanover shows all available supernodes to be connected since "NaN".

error_image

Current master is affected. I'm betting on that xhtml commit, but will verify the faulty commit this afternoon.

The device was purged using firstboot, the misbehaving page is not due to migration artifacts or something. I'm expecting the error on all devices, but only tested/seen it on the archer60 and an old CPE210v2 yet.

Not sure if that changed, but while mesh-vpn is disabled, that paragraph is not expected to show up at all.

It worked in Hanovers vH24 so gluon-v2021.1-148-g9e3cf00e

**Gluon Version:**
current master

**Site Configuration:**
The used siteconf is [master-wireguard](https://github.com/freifunkh/site/tree/master-wireguard)

**Custom patches:**
There are [four custom patches](https://github.com/freifunkh/site/tree/master-wireguard/patches)

Of which only the last has something todo with WireGuard, and is adding an update-script to rename the netork interface `vpn` to `wg_mesh`, so that we don't have to do this manually for our testing devices from prior the merge of WireGuard into gluon.

AiyionPrime:
> Since freifunk-gluon/packages#250 mesh-vpn-peers
> can be empty arrays if they're not connected
> and the node is in a WireGuard site.

This is tested on a WireGuard node and works, but has not been tested on a fastd node yet.

@genofire no need to fix this anymore.

resolves #2498


lemoer:
I tested this on a fastd node. It seems to work well for me.

lemoer:
I tested this on a fastd node. It seems to work well for me.

lemoer (approved):
LGTM.

package/gluon-status-page/javascript/status-page.js
```diff
@@ -206,7 +206,7 @@
 			tr.appendChild(th);
 
 			var td = document.createElement('td');
-			if (peer[1])
+			if (peer[1] && Object.keys(peer[1]).length !== 0)
```
NeoRaider, line 209: Why not check for the established field, as that is what we're interested in?
```suggestion
			if (peer[1] && peer[1].established != null)
```
(this will actually be `undefined` when the field is not set, however `!= null` seems to be a common idiom to check for both `null` and `undefined`)

AiyionPrime: Because I found the undefined part looked uglier. But I've got no strong opinions regarding any js code. If you want that instead, I'll update the PR tomorrow or so.

AiyionPrime: What I feared was the undefined coming up in the developer console, like it does for updateNodeinfo and the location part. (another issue I'd like to get to in the next days...)

AiyionPrime: resolved in b5242a530e0893e591f0f09cadefac5dc93cdf17

AiyionPrime:
The alternative @NeoRaider is now tested as well and appears to be working on WireGuard nodes.

lemoer:
LGTM.

lemoer (approved):

PR #2508, opened 2022-05-06
generic: reduce flash consumption

blocktrron:
Reduce flash consumption by removing dependencies pulling in nftables as well as stripping parts of the kernel which are not required for Gluon.

```
OCEDO Raccoon
-------------
Before: 6226728
After: 6030120
```


NeoRaider:
There's also package `procd-seccomp` in DEFAULT_PACKAGES.

NeoRaider:
Ah, I didn't see that it is also removed with `CONFIG_SECCOMP`.

NeoRaider (approved):

PR #2512, opened 2022-05-07
ath79: add patch to fix serial console and sysupgrade on OR750

nrbffs:
The serial console on the Joy-IT OR750i doesn't work, because the UART
node is disabled in device tree. This also breaks sysupgrade.

The issue is already fixed upstream in OpenWRT by
7054721cf94f4aa8fe71ac4c28c922b19a69ba1d ("ath79: enable UART in SoC
DTSI files"), but since 19.07 doesn't support the OR750, we need our own
patch.

Signed-off-by: Nico Boehr <freifunk@nicoboehr.de>


blocktrron:
ath71 --> ath79

nrbffs:
Ooops, thanks, fixed.

PR #2515, opened 2022-05-08
ath79-generic: (re)add wzr-hp-ag-300h / wzr-600dhp

AiyionPrime:
@ce-4 intends to test this device.
 
> Both are effectively the same hardware, the latter being Bufalos
> replacement model.

> Gone due to

- ~~Must be flashable from vendor firmware~~ wonttest
  - ~~Web interface~~
  - ~~TFTP~~
  - ~~Other: <specify>~~
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [x] Should show activity **not sure whether they blink for beacons, but they do until the respective device is down**
  - Switch port LEDs **none present**
- Outdoor devices only:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -9,6 +9,10 @@ ath79-generic
   - Fritz!WLAN Repeater 450E [#avmflash]_
   - Fritz!Box 4020 [#avmflash]_
 
+* Buffalo
+
+  - WZR-HP-AG300H / WZR-600DHP
+
 * devolo
 
   - WiFi pro 1200e [#lan_as_wan]_
--- a/targets/ath79-generic
+++ b/targets/ath79-generic
@@ -26,6 +26,11 @@ device('avm-fritz-wlan-repeater-450e', 'avm_fritz450e', {
 	factory = false,
 })
 
+-- Buffalo
+
+device('buffalo-wzr-hp-ag300h', 'buffalo_wzr-hp-ag300h')
+device('buffalo-wzr-600dhp', 'buffalo_wzr-600dhp')
+
 -- devolo
 
 device('devolo-wifi-pro-1200e', 'devolo_dvl1200e', {


ce-4:
I set up and tested a Buffalo WZR-HP-AG300H with name "ath79-test-freifunk-gluon-PR-2515"

Findings:
It mostly works (sysupgrade, vpn, wifi, network port assignments, firstboot, primary mac)

some things don't work:
- Gluon profile name matches autoupdater image name (lua -e 'print(require("platform_info").get_image_name())')):
   NOPE(?): stable fw string is "buffalo-wzr-hp-ag300h-wzr-600dhp" while on experimental fw string is "buffalo-wzr-hp-ag300h"
- LED: the switch-port-LED doesnt work, it's a single LED with label "ROUTER"
   path for led entry: /sys/devices/platform/ath9k-leds/leds/green:router
   path for devicetree entry: /sys/firmware/devicetree/base/ath9k-leds/router
things I havent tested yet/wont test:
- Flash-from-vendor-firmware: won't test, stock-fw is ancient and i found no method to revert to the stock fw
- Mesh-Wifi 2.4/5ghz: not yet tested, need to set up a second ffh device for that

rotanid:
> Maybe also of note, to revert to my standard ffmuc-fw, i had to use the -F flag with sysupgrade: $ sysupgrade -n -F gluon-ffmuc-v2022.5.2-buffalo-wzr-hp-ag300h-wzr-600dhp-sysupgrade.bin Tue Sep 27 04:39:11 CEST 2016 upgrade: Image metadata not present Image check failed but --force given - will update anyway! Tue Sep 27 04:39:11 CEST 2016 upgrade: Commencing upgrade. Closing all shell sessions.

this is absolutely expected, downgrades are never supported (although they usually work with forgetting settings)

AiyionPrime:
> Gluon profile name matches autoupdater image name (lua -e 'print(require("platform_info").get_image_name())')):
NOPE(?): stable fw string is "buffalo-wzr-hp-ag300h-wzr-600dhp" while on experimental fw string is "buffalo-wzr-hp-ag300h"

Not sure, what you mean: The name of the image changed, it was a single image in ar71xx, for what are (technically) two devices.
What the firmware image was called in ar71xx does not matter for you though:
What's important ist, that, when you call `lua -e 'print(require("platform_info").get_image_name())'` on the new ath79 firmware, it outputs `buffalo-wzr-hp-ag300h` as that is what I worte in the PR in [line 37](https://github.com/freifunk-gluon/gluon/pull/2515/files#diff-8bd1ca94227cccf7c95011ebe4f9d1b7ae82d4c1495d27c12532af82df0d1312R37).

> LED: the switch-port-LED doesnt work, it's a single LED with label "ROUTER"

Possible, that OpenWrt cleaned that up and dropped the 'misuse' of that LED for the ports, will look into that later.

> wifi activity leds:

I'm happy the assignment works, I do not know, what was forgotten in OpenWrt, that the LEDs blink in sync but not with traffic, maybe @rotanid has an idea?

The reset button works? Anything else I used check for now?

Thanks for testing this!

ce-4:
Ah ok, i was mistaken about the device name, if the images were split, "buffalo-wzr-hp-ag300h" is of course fine then. I thought it should be the same as with ar71xx

As for the rest:
[ok] Reset/WPS/... button must return device into config mode: works as with ar71xx
[ok] Power/system LED / Lit while the device is on: works as with ar71xx
[ok] Power/system LED / Should display config mode blink sequence: works like on stable
[ok] Wireless network (if applicable) / Association with AP must be possible on all radios: 2.4+5ghz non-mesh both work

ce-4:
Update:
2.4GHz meshing works (testet with a spare GL.inet MT300N-v2).

As for the LED issues: I have no idea how to test https://openwrt.org/docs/techref/hardware/port.gpio#script_blink
the LEDs are already named and working, it's just that the new firmware doesn't know that there's a single "ROUTER"-LED instead of separate switch-port LEDs, also the DIAG-led doesnt show config-mode yet. The router led was used for switch-port activity (i.e. LAN-ports 1/2/3/4 in the past):
experimental fw, path for led entry: /sys/devices/platform/ath9k-leds/leds/green:router
experimental fw, path for devicetree entry: /sys/firmware/devicetree/base/ath9k-leds/router.

The router has six LEDs:
ON (probably USB-port activity, not used with gluon, but LED handle exists also in experimental/new image)
DIAG (used to discern config-mode on stable gluon)
ROUTER (activity led for the switch port)
5GHz
2.4GHz
POWER

AiyionPrime:
So. The things left are the LAN LED and the WAN LED, which do technically not exist.
The "ROUTER" LED does. In my opinion either of the following would be fine:
- leave as is, as OpenWrt is technically right about there not being LAN or WAN LEDs
- set the "ROUTER" LED  as either LAN or WAN identifier matching ar71xx if possible at all?
- or possibly even better use both so it does blink if either is active

Looking at [other Buffalo devices](https://github.com/openwrt/openwrt/blob/f5a87a0a7b8d5aa83435868eb324601df6c4f4dc/target/linux/ath79/generic/base-files/etc/board.d/01_leds#L107) we could go with option two either upstream in OpenWrt of patch it downstream

Any opinions, @neoraider, @blocktrron?

Anyway @ce-4 and me are satisfied as is and would even merge this as is and go with option one.

AiyionPrime:
> I'd go for "whatever OpenWrt does"

_originally posted by @neoraider in hackint#gluon 17.06.22 12:59:17_

So this is ready to review/merge.

AiyionPrime:
Just a typo in the commitmessage.

AiyionPrime:
Rebased onto master in order to avoid the docs warning about language being none.

PR #1986, opened 2020-04-21
targets/ath79-generic: add archer-D50 v1

2tata:
- [x] must be flashable from vendor firmware
  - [x] tftp
- [x] must support upgrade mechanism
  - [x] must have working sysupgrade
    - [x] must keep/forget configuration (if applicable)
      *think `sysupgrade [-n]` or `firstboot`*
  - [x] must have working autoupdate
    *usually means: gluon profile name must match image name*
- [x] reset/wps/phone button must return device into config mode
- [x] **primary mac should match address on device label (or packaging) (https://gluon.readthedocs.io/en/latest/dev/hardware.html#notes)**
- wired network
  - [x] should support all network ports on the device
  - [x] must have correct port assignment (WAN/LAN)
- wifi (if applicable)
  - [x] association with AP must be possible on all radios
  - [x] association with 802.11s mesh must be working on all radios 
  - [x] ap/mesh mode must work in parallel on all radios
- led mapping
  - power/sys led (_critical, because led definitions are setup on firstboot only_)
    - [x] lit while the device is on
   - [ ]  **should display config mode blink sequence** <-- NOT Working
(https://gluon.readthedocs.io/en/latest/features/configmode.html)
  - radio leds
    - [x] should map to their respective radio
    - [x] should show activity
  - switchport leds
    - [x] should map to their respective port (or switch, if only one led present) 
    - [x] should show link state and activity

Issue #1687, opened 2019-03-27
raspberry pi 3 broken status


ghost:
I'm running gluon on a rpi3 for a while and everything seems to be good so far:
http://mgmt.saar.freifunk.net/hopglass/#!v:m;n:b827eb0b08a9

It does only support one wifi interface and according to `iw` no 11s. Client AP works fine, though, and so does meshing over Ethernet. VLANs are also supported.

What's required to mark this as unbroken?

kpanic23:
All WiFi interfaces working, with 11s. Sorry...

bobcanthelpyou:
@miyuuli 

A pull request with the following device checklist is required.
https://github.com/freifunk-gluon/gluon/wiki/Device-Integration-checklist

Raspberry pi 3 is already supported as `BROKEN` by Gluon. 
https://github.com/freifunk-gluon/gluon/blob/master/targets/brcm2708-bcm2710

But marked as `# BROKEN: Untested` in `targets.mk`
https://github.com/freifunk-gluon/gluon/blob/master/targets/targets.mk

Test the device by filling out the checklist and move the target to a different place in `targets.mk`.

rotanid:
@bobcanthelpyou you pointed out the process correctly, but as @kpanic23 already wrote, this won't happen here because 11s is not working.
therefore, the only possible change could be to change the reason for it being broken.

bobcanthelpyou:
@rotanid sorry, i was confused, cause @miyuuli wrote "no 11s" and @kpanic23 wrote "All WiFi interfaces working, with 11s" which i might misinterpreted as everything is fine. Maybe a "need" / "must" / "should" / ., is missing to work with my build-in interpreter. Sorry again.

Could someone at least update the target file with "11s not working / unsupported / .." or something like that for the raspberry pi 3? It seems to be tested somehow.

Adding something like `ifneq ($(GLUON_WLAN_MESH_ibss)$(BROKEN),)` for the target doesn't make sense?

blocktrron:
@bobcanthelpyou I highly doubt the WiFi driver of the Raspberry Pi 3 supports anything else than the simple AP or Client case (It's Broadcom after all). It uses the brcmfmac driver and the WiFi chip itself is attached via SDIO. The fact it is attached via SDIO alone makes meshing a big nono in terms of practicability.

mweinelt:
For reference

```
Wiphy phy0
	max # scan SSIDs: 10
[...]

PR #2516, opened 2022-05-09
sunxi-cortexa7: add support for FriendlyElec NanoPi R1

goligo:
**blocked by** https://github.com/openwrt/openwrt/issues/10080

  - [ ] Web interface
  - [ ] TFTP
  - [x] Other: SD card image / no vendor firmware
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs - there is no radio LED
    - [ ] Should show activity
  - Switch port LEDs
- Outdoor devices only:

### DIFF
--- a/targets/sunxi-cortexa7
+++ b/targets/sunxi-cortexa7
@@ -15,3 +15,7 @@ device('lemaker-banana-pro', 'lemaker_bananapro', {
 device('lamobo-r1', 'lamobo_lamobo-r1', {
 	broken = true, -- AP+11s not working
 })
+
+device('friendlyarm-nanopi-r1', 'friendlyarm_nanopi-r1', {
+	broken = true, -- 11s not working
+})


goligo:
- Mesh does not work - BCM43430 does not support 802.11s
- Reset button does not have any effect - neither resets nor starts config mode

goligo:
@AiyionPrime

rotanid:
please read the requirements, especially that devices without mesh support won't be merged without the BROKEN flag.
https://github.com/freifunk-gluon/gluon/wiki/Device-Integration

AiyionPrime:
Did not know about the broken mesh; but having this device added as `BROKEN` would still be worthwile, as its a pretty decent offloader.

goligo:
> please read the requirements, especially that devices without mesh support won't be merged without the BROKEN flag. https://github.com/freifunk-gluon/gluon/wiki/Device-Integration

I tend do disagree in this case, BROKEN implies there is something which can and should be fixed, which is not the case here. Like the Raspberry Pi 3, which can´t do mesh as well, but is not marked as BROKEN

belzebub40k:
@goligo the whole target of the RPi3 is marked as broken

https://github.com/freifunk-gluon/gluon/blob/master/targets/targets.mk#L24

AiyionPrime:
And here is the raspberrys current discussion: #1687

I think this would then be one take form the referenced PR:

> At the very least we would need to ensure that Gluon doesn't even attempt to use the built-in WLAN for meshing.

_Originally posted by @NeoRaider in https://github.com/freifunk-gluon/gluon/issues/1687#issuecomment-485495687_

goligo:
Added BROKEN flag and removed it from supported devices list.

blocktrron:
> Primary MAC address should match address on device label

At least the R2S does not have a unique MAC address stored anywhere on the board. I was also not able to find the MAC address extraction logic in U-Boot / OpenWrt. Any chance you have two of said boards? I suspect the MAC address is generated randomly on boot, which can lead to two devices ending up with the same MAC address (No-Go in a batman-adv network).

The R2S uses the unique MMC serial number to derive it's MAC address.

goligo:
> which can lead to two devices ending up with the same MAC address

Yes, we experienced that painfully with early firmware on the R2S in FFMUC. I know we have one other instance of the R1 in the community, that did get a different MAC address than mine, without any manual intervention.

rotanid:
> I tend do disagree in this case, BROKEN implies there is something which can and should be fixed, which is not the case here. Like the Raspberry Pi 3, that can´t do mesh as well, but is not marked as BROKEN

the raspberry pi 3 was integrated before we had agreed on these rules.

i wonder why we set up rules and now we tend to discuss the applicability on each occasion? (see #1986) might as well drop the rules...
The rules were set up by the maintainers, not by the public, to be able to stop discussing about each issue again and again - in the end everyone profits if there are clear rules.
everyone is able to have a Gluon fork (which even my community has, by the way)

goligo:
> i wonder why we set up rules and now we tend to discuss the applicability on each occasion?

Rules are great and have a purpose, but as the world usually is more complex than expected, there can also be exception to rules. Rules shouldn't be followed just for the sake of following rules. However, as you might have noticed, I already have changed the PR, so there is no need to start a basic discussion about it.

targets/sunxi-cortexa7
```diff
@@ -15,3 +15,7 @@ device('lemaker-banana-pro', 'lemaker_bananapro', {
 device('lamobo-r1', 'lamobo_lamobo-r1', {
 	broken = true, -- AP+11s not working
 })
+
+device('friendlyarm-nanopi-r1', 'friendlyarm_nanopi-r1', {
+        broken = true, -- 11s not working
```
rotanid, line 20: sounds picky i know, but please use the same indentation as elsewhere, like 4 lines above

goligo: Done.

blocktrron:
Okay, I've had a more in-depth look into U-Boot and the OpenWrt userland. Can you check if using the device with OpenWrt generates random MAC addresses on subsequent boots? In case the auto-generation is done within Linux, this is usually also echoed to the kernel ringbuffer.

In this case, a workaround similar to the one used on the R2S is necessary, as a) this platforms usually do not have a reliable entropy source and b) the WAN mac-address in Gluon would not be stable.

goligo:
> Can you check if using the device with OpenWrt generates random MAC addresses on subsequent boots?

Yes exactly this is the case - WAN MAC address changes on every restart of the device.

blocktrron:
@goligo Did you asses my comment regarding reading the CID of the SD card to generate a stable MAC address? I see this as the biggest blocker for this board.

goligo:
@blocktrron That would be a patch for OpenWrt, right? Like you recently did for the R4S?

blocktrron:
Correct, you can take it basically as a blueprint.

goligo:
As discovered by @AiyionPrime , when trying to make progress with the stable MAC, there are more severe issues with NanoPi R1 introduced with OpenWrt 22.03 / Kernel 5.10. MMC detection is unstable and randomly assigns different device names on each boot (mmcblk0 or mmcblk2). There also seems to be an issue with networking, even when it boots, in Gluon v2022.1.1 network connection on br-wan is not functional.

AiyionPrime:
I'll update this comment with an upstream issue reference once I opened it either today or tomorrow. This is expected to be resolvable. Hauke already suggested smth. On it.

AiyionPrime:
First patch is online: https://patchwork.ozlabs.org/project/openwrt/patch/20221211130446.23492-1-git@aiyionpri.me/

Might have found a tester for it as well.
MMC detection is still not solved, I think.

AiyionPrime:
Second Patch is online as well: https://patchwork.ozlabs.org/project/openwrt/patch/20221215141833.100172-1-git@aiyionpri.me/

Third is done as well, but rests until the other two patches got any answers.
Merging it without them is worthless.

AiyionPrime:
First two patches have been merged into master.
The last one is now submitted as well.

https://patchwork.ozlabs.org/project/openwrt/patch/20221222201201.283913-1-git@aiyionpri.me/

@stintel suggested waiting for the second patches fallout and do the backport if it does not occur.

AiyionPrime:
The third patch has been merged as well.
While doing the patches, I learnt about the `Secure ID` every H2+/H3/H5 Soc is supposed to have. Basically a 16 byte serial number.

U-Boot uses that as source for the mac-addresses and I intend to implement it the same way for the NanoPi R1.

Other than that, on OpenWrts Master the NanoPi is now fully functional, does have a MAC based on the (e)MMC and should therefore only have MAC collisions, if cheap micro SD cards are used.
(A risk, thats shared with the R2 and R4 rockchip variants as well)

So once the patches two and three have been backported and we updated our modules afterwords, this can finally continue.

AiyionPrime:
Currently waiting on the merge of this upstream backport series:
https://patchwork.ozlabs.org/project/openwrt/list/?series=335395

blocktrron:
Series has been backported, whats the status here?

Djfe:
the patch series fixed the flash detection but the wifi path also changes randomly between mmc0 and mmc1. That would probably throw Gluon off as well, right?
https://github.com/openwrt/openwrt/issues/12237

AiyionPrime:
I asked @goligo in early april whether he wants to put more effor tinto this right now, but did not receive an answer yet.
For me the priority of this device falls behind nearly every other thing sitting on my desk rn.

The current next steps would be:
- [ ] verify my upstream patches not only allow the device to boot in 22.03 and master
- [ ] but also have a mac-address depending on the hardware

The pitfall is that this device comes in two variants: with eMMC and without.

@goligo and me have access to a variant each, but apparently lack the time to dig into this further right now.

@goligo I hope you don't mind me closing this for the moment.
Let me know if you intend to pursue this and I'll reopen it at once.

PR #2520, opened 2022-05-15
gluon-autoupdater: check default branch name, check number of pubkeys, revert invalid branch on update/reconfigure

### LINKED ISSUE #2519: Autoupdate Not working if branch is not in site
We accidentally offered a firmware in Freifunk Nord that had the Autoupdate branch defined to "2020.1.x" instead of "stable".

Since there's no firmware mirror defined for that, the au is not working.

It would have been helpful, if the au would fall back to stable if the branch is not defined

### LINKED ISSUE #2532: autoupdater: Add check, if number needed sigs>=keys
You could create a firmware where the autoupdater would never work, if you add only one key in site.conf and needed keys=2 ;)

maybe add a small check

NeoRaider:
Closes #2519
Closes #2532

Checked build with:
- Valid branch: OK
- Unset branch: OK
- Invalid branch: Build failed as expected

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -25,6 +25,7 @@ files["package/**/check_site.lua"] = {
 		"extend",
 		"in_domain",
 		"in_site",
+		"value",
 		"need",
 		"need_alphanumeric_key",
 		"need_array",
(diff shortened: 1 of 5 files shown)


rubo77:
please also backport this to the release branches

NeoRaider:
Hmm, this solution is still not ideal: When setting the branch using `GLUON_AUTOUPDATER_BRANCH`, the site check is ineffective. I think this should be easy to fix as well though.

NeoRaider:
`GLUON_AUTOUPDATER_BRANCH` is now checked as well.

lemoer (approved):
LGTM.

PR #2521, opened 2022-05-15
ath79-generic: (re)add Archer C58 v1

AiyionPrime:
@moridius intends to test this device.

The device is expected to not work properly due to 64M ath9k + ath10k.
Its marked as `BROKEN`, like it was in ar71xx.

@moridius please take special care regarding testing the primary mac.

- ~~Must be flashable from vendor firmware~~ **not tested, expected to work**
  - ~~Web interface~~
  - ~~TFTP~~
  - ~~Other: <specify>~~
    - [x] Must forget configuration (`sysupgrade [-n]`, `firstboot`)
    - [x] Must keep configuration (`sysupgrade [-n]`, `firstboot`) ~~keep was not tested, forget works~~ **both tested, both work***
- Wired network
- Wireless network (if applicable) **wont work reliably**
  - ~~Association with AP must be possible on all radios~~
  - ~~Association with 802.11s mesh must work on all radios~~
  - ~~AP+mesh mode must work in parallel on all radios~~
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [x] Should show activity
  - Switch port LEDs
- Outdoor devices only:

### DIFF
--- a/targets/ath79-generic
+++ b/targets/ath79-generic
@@ -312,6 +312,12 @@ device('tp-link-archer-c5-v1', 'tplink_archer-c5-v1', {
 	packages = ATH10K_PACKAGES_QCA9880,
 })
 
+device('tp-link-archer-c58-v1', 'tplink_archer-c58-v1', {
+	packages = ATH10K_PACKAGES_QCA9888,
+	broken = true, -- OOM with 5GHz enabled in most environments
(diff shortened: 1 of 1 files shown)


AiyionPrime:
@moridius concluded his tests. Thanks for the work!

Anything left to fix?

PR #2524, opened 2022-05-17
ath79-generic: migrate D-Link DIR-825 B1 from ar71xx

s-2:
also fix whitespace (tabs vs. spaces) my text editor messed up while migrating dir-505

  - [x] Web interface
  - [-] TFTP
  - [x] Other: _recovery web interface, but very cumbersome with latest operating systems / browsers_
    - [x Must keep/forget configuration (`sysupgrade [-n]`, `firstboot`)
- [-] Primary MAC address should match address on device label (or packaging)
      _-> MAC address on map equals label mac - 1, consistent with ar71xx (will not spawn a new node on the map)_
    _-> label mac -1 prevails even after reset_
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [x] Should show activity
  - Switch port LEDs
- Outdoor devices only:
  - [-] Added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`

### DIFF
--- a/targets/ath79-generic
+++ b/targets/ath79-generic
@@ -83,6 +83,14 @@ device('d-link-dir-505', 'dlink_dir-505', {
 	},
 })
 
+device('d-link-dir825b1', 'dlink_dir-825-b1', {
+	factory = false,
+	class = 'tiny', -- Only 6M of usable Firmware space
+	manifest_aliases = {
+		'd-link-dir-825-rev-b1', -- Upgrade from OpenWrt 19.07
+	},
+})
+
 
 -- Enterasys
 
--- a/targets/ath79-generic
+++ b/targets/ath79-generic
@@ -76,11 +76,11 @@ device('d-link-dap-2660-a1', 'dlink_dap-2660-a1', {
 })
 
 device('d-link-dir-505', 'dlink_dir-505', {
-        factory = false,
-        manifest_aliases = {
-                'd-link-dir-505-rev-a1', -- Upgrade from OpenWrt 19.07
-                'd-link-dir-505-rev-a2', -- Upgrade from OpenWrt 19.07
-        },
+	factory = false,
+	manifest_aliases = {
+		'd-link-dir-505-rev-a1', -- Upgrade from OpenWrt 19.07
+		'd-link-dir-505-rev-a2', -- Upgrade from OpenWrt 19.07
+	},
 })
 
 
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -25,6 +25,7 @@ ath79-generic
   - DAP-2660 A1 [#lan_as_wan]_
   - DIR-505 A1 [#lan_as_wan]_
   - DIR-505 A2 [#lan_as_wan]_
+  - DIR-825 B1
 
 * Enterasys
 


AiyionPrime (changes requested):
Deviceclass tiny is missing, other than that both commits look fine to me.

AiyionPrime (approved):
Looks fine

NeoRaider:
The MAC address not matching the label is a bug that should be fixed.

s-2:
> The MAC address not matching the label is a bug that should be fixed.

I guess the ideal solution here would be fixing this upstream and then backporting an OpenWrt patch to gluon?

Digging out the device again to check MAC assignment (this would soon need to be touched upstream anyways, regarding DSA migration).

s-2:
Works as expected, new map entry: https://hannover.freifunk.net/karte/#/de/map/002401e7f251
old node: https://hannover.freifunk.net/karte/#/de/map/002401e7f250

How do we deal with patches, I guess this would need refreshing with every new version, until it would be included upstream; or should an alternative way of setting label MAC be used here in the meantime?

s-2:
> please add a description to the OpenWrt patch and open a PR for OpenWrt / send patch to list.

https://github.com/openwrt/openwrt/pull/9947

So the actual patch can be dropped from the patches folder for this PR here, since it will be backported to the release branch?

targets/ath79-generic
```diff
@@ -76,11 +76,19 @@ device('d-link-dap-2660-a1', 'dlink_dap-2660-a1', {
...
-                'd-link-dir-505-rev-a2', -- Upgrade from OpenWrt 19.07
-        },
+	factory = false,
+	manifest_aliases = {
+		'd-link-dir-505-rev-a1', -- Upgrade from OpenWrt 19.07
+		'd-link-dir-505-rev-a2', -- Upgrade from OpenWrt 19.07
+	},
+})
+
+device('d-link-dir825b1', 'dlink_dir-825-b1', {
```
rotanid, line 86: @s-2 @blocktrron why is it referenced as "d-link-dir825b1" and not as "d-link-dir-825-b1" following the scheme of all other devices?

s-2: When the device was ported to ath79 in OpenWrt (https://github.com/openwrt/openwrt/commit/97de133368762824ba3d7e9d95f97011a597ec1a), the author of the patch apparently chose to leave out any spaces or dashes within the `model` property in the .dts file, unlike other D-Link devices, which mostly use a scheme like "DIR-825 A1".
This string is the source of the model name for gluon, so we need the `manifest-aliases` property here, to allow ar71xx sysupgrade to even recognize this as a compatible image.

rotanid: always thought the first string does not directly depend on OpenWrt-stuff and only the 2nd string is the openwrt-reference. so i was wrong?

s-2: The second string is indeed the OpenWrt model name, and the first could be arbitrary, but I think we would need a third one then (probably `profile`?) to make it match the output of `lua -e 'print(require("platform_info").get_image_name())'` as mentioned in the integration checklist;
I've only just learned about most of this during #2417 myself, but maybe we could even have the model string fixed directly in the .dts file, since I'm not sure whether OpenWrt folks have any strong opinions on the model name in the first place, besides it should not break anything in OpenWrt itself, as sysupgrade would rely on the model name only.

herbetom: Just stumpled accross this while trying to adapt this to https://github.com/freifunk-darmstadt/gluon-firmware-selector

I think a consistent naming would be better.

```suggestion
device('d-link-dir825-b1', 'dlink_dir-825-b1', {
```

AiyionPrime: @rotanid The first string depends on OpenWrt as well.
At least in ath79 it is an escaped version of either [the model](https://github.com/openwrt/openwrt/blob/master/target/linux/ath79/dts/ar7161_dlink_dir-825-b1.dts#L10)
iirc.

Take the string, lowercase it,
replace every space and every non alphanumeric to dashes,
remove leading, trailing, as well as double dashes until they are all gone.
 
In this case `D-Link DIR825B1` --> `d-link dir825b1` --> `d-link-dir825b1`, done.

So in order to appropriately change the parameter as @herbetom suggested, one would need to first address the corresponding upstream OpenWrt line.
Once that's patched we can match it here.

AiyionPrime: > The second string is indeed the OpenWrt model name, and the first could be arbitrary, [...]

_Originally posted by @s-2 in https://github.com/freifunk-gluon/gluon/pull/2524#discussion_r898543164_

@s-2 likely you meant the right thing, but the second is not matching the model name, but [the definition](https://github.com/openwrt/openwrt/blob/master/target/linux/ath79/image/generic.mk#L992), which is added to `TARGET_DEVICES`.

s-2: > but the second is not matching the model name, but [the definition](https://github.com/openwrt/openwrt/blob/master/target/linux/ath79/image/generic.mk#L992)

I probably meant to say 'board name' or something (i.e. what's used within `base-files` scripts, where the variable is called `$board`), but that turns out to be yet another different string used for a device... (if only just underscore vs. comma)

It seems nothing within OpenWrt itself depends on the `model` attribute from the dts, so someone might as well attempt to upstream this change. @herbetom Are there other devices affected due to weird model strings, e.g. what about DIR-505, which used an alias to provide `a1` and `a2` images in ar71xx, but in ath79 there's just one image now (and we can expect there never to be any new version of that device being introduced).

NeoRaider: I think we're going to stop using the model name in Gluon for the image name eventually. Using it was a necessary hack for ar71xx, where multiple models used the same board name, but for ath79 we can clean this up at some point.

herbetom: It's most probably not worth the effort changing it now since it's not a problem. 

https://github.com/freifunk-darmstadt/gluon-firmware-selector/pull/145/files#diff-5520dd49755e0a353d3857b8963a91cb4b8d87a01b6b8eb7bb6cce1c180eac83R66

The naming is a bit weird and not really consistent, but to me it seems manageable.

PR #2529, opened 2022-05-22
generic: add openssh-sftp-server

txt-file:
* adds ~800 KB binary size to ath79-generic because it pulls in openssl
* since OpenSSH 9.0 the old scp protocol is finally deprecated. To use
  that ancient protocol OpenSSH scp client needs a special parameter or
  a sftp server on the server side.

EDIT: kind of an alternative to https://github.com/freifunk-gluon/gluon/pull/2479 Sooner or later we should get rid of the old scp protocol anyway.

### DIFF
--- a/targets/generic
+++ b/targets/generic
@@ -95,4 +95,5 @@ packages {
 	'-firewall4',
 	'gluon-core',
 	'ip6tables-legacy',
+	'openssh-sftp-server',
 }


NeoRaider:
Pulling in OpenSSL by default is unacceptable. If desired, the package should be added to the site configuration.

I hope that there will be another SFTP server implementation to use with dropbear eventually. For now, relying on the legacy SCP protocol seems preferable to me.

NeoRaider:
Hmm, also the only reason for openssh-sftp-server's dependency on OpenSSL is `RAND_bytes`, so it might be easy to patch it to use another source of randomness. But even without OpenSSL, OpenSSH's SFTP implementation is a bit larger than I'd like for Gluon.

txt-file:
I just found the upstream issue at https://github.com/openwrt/openwrt/issues/9889

I'll close this PR as this is an upstream problem.

PR #2530, opened 2022-05-24
gluon-web-model: add Warning

mkg20001:
Adds a new Warning Node that can be used to display warnings to the user

Also adds a new Warning on web-private-wifi if mesh-on-wan and private wifi is enabled


package/gluon-web-private-wifi/luasrc/lib/gluon/config-mode/model/admin/privatewifi.lua
```diff
@@ -13,9 +14,27 @@ local s = f:section(Section, nil, translate(
...
 local enabled = s:option(Flag, "enabled", translate("Enabled"))
 enabled.default = uci:get('wireless', primary_iface) and not uci:get_bool('wireless', primary_iface, "disabled")
 
+if mesh_on_wan and enabled.data then
+	local w = Warning()
+	w:setcontent(translate(
+		'Meshing on WAN interface is enabled.' ..
+		'This can lead to problems.'
+	))
+	s:append(w)
```
NeoRaider, line 35: Form nodes must never be added conditionally, as it will break the ID numbering. Instead, always add the node and make the template show the div conditionally (for example only when a content has been set).

NeoRaider: Instead of adding more classes, the `Template` class can be used. Adding trivial setter methods is also unusual for `gluon-web-model`.

Something like the following should work:

```lua
let warning = s:option(Template, 'model/warning')
if mesh_on_wan and enabled.data then
	w.content = translate(
		'Meshing on WAN interface is enabled.' ..
		'This can lead to problems.'
	)
end
```

NeoRaider (changes requested):

package/gluon-web-private-wifi/luasrc/lib/gluon/config-mode/model/admin/privatewifi.lua
mkg20001: I'm getting this error

```
Failed to execute dispatcher target for entry '/admin/privatewifi'.
/usr/lib/lua/gluon/web/model/classes.lua:423: class must be a descendant of AbstractValue
```

NeoRaider: Hmm, it seems that `Template` is currently meant to be used as a section rather than an option. So using
```lua
local warning = f:section(Template, 'model/warning')
```
before creating the section `s` might work, if that position of the warning looks okay.

If we really want it inside the existing section, allowing `Template` in addition to `AbstractValue` for `s:option()` would be acceptable.

mkg20001: My other usecase involves having it next to the field that's causing the warning, so having it inside the section would be preferable.

mkg20001: A custom method `s:template('model/warning', {content="..."})` perhaps?

mkg20001:
<video src="https://user-images.githubusercontent.com/7735145/170814258-908342f5-0fc3-4065-88f2-25df4f598a35.mp4" autoplay loop muted></video>

mkg20001:
Now warning is displayed only when private wifi is enabled

mkg20001:
@NeoRaider do you have any comments to add? Can I fixup history and then have this merged?

AiyionPrime:
@mkg20001 Would you squash and rebase this into a mergeable state?
No promises, but that way we can talk about it on tuesday and merge it, if/once it's tested.

PR #2535, opened 2022-05-28
OLSRDv2 Support: Initial Meshing

mkg20001:
This just includes basic olsr2 meshing to make it easier to review


package/gluon-mesh-olsrd/luasrc/lib/gluon/upgrade/360-gluon-mesh-olsrd-setup-intf
```diff
@@ -0,0 +1,212 @@
...
+	uci:section('firewall', 'rule', 'allow_olsr2_mesh', {
+		-- src = 'mesh',
+		-- HACK: this is a temporary hack because firewall doesn't want to work
+		src = '*',
+		dest_port = '269',
+		proto = 'udp',
+		target = 'ACCEPT',
+	})
+
+	uci:section('firewall', 'rule', 'allow_olsr2_wired_mesh', {
```
mkg20001, line 198: is this even needed?

package/gluon-mesh-olsrd/luasrc/lib/gluon/upgrade/360-gluon-mesh-olsrd-setup-intf
```diff
@@ -0,0 +1,212 @@
...
+		})
+	end
+
+	uci:section('olsrd2', 'interface', 'loopback', {
+		ifname = { 'loopback' },
+		bindto = addrs,
+	})
+
+	uci:section('firewall', 'rule', 'allow_olsr2_mesh', {
+		-- src = 'mesh',
```
mkg20001, line 190: todo

package/gluon-mesh-olsrd/luasrc/lib/gluon/upgrade/370-gluon-mesh-olsrd-setup-fw
```diff
@@ -0,0 +1,79 @@
...
+local uci = require('simple-uci').cursor()
+local site = require 'gluon.site'
+local util = require 'gluon.util'
+local wireless = require 'gluon.wireless'
+
+networks = uci:get_list('firewall', 'drop', 'network')
+util.remove_from_set(networks, 'client')
+uci:set_list('firewall', 'drop', 'network', networks)
+
+uci:section('firewall', 'rule', 'allow_mesh_ping', {
```
mkg20001, line 12: I'm not sure about those rules

package/gluon-mesh-olsrd/luasrc/lib/gluon/upgrade/360-gluon-mesh-olsrd-setup-intf
```diff
@@ -0,0 +1,212 @@
...
+for _,i in pairs(mesh_interfaces) do
+	if util.contains(uplink_interfaces, i) then
+		has_uplink_mesh = true
+	else
+		has_other_mesh = true
+	end
+end
+
+if has_uplink_mesh then
+	-- use uplink with macvlan so we do not have traffic leaks
```
mkg20001, line 95: this fixes the problem that happens when meshing on wan and having mesh vpn enabled, for ex. this might make sense for all of gluon. thoughts?

targets/generic
```diff
@@ -59,6 +59,8 @@ config('KERNEL_SECCOMP', false)
 -- use try_config, so enabling the package is still possible
 try_config('PACKAGE_kmod-mt7915e', false)
 
+config('OONF_GENERIC_HTTP', true)
```
mkg20001, line 62: this needs to be enabled when mesh olsrd is used so http module in olsr2 gets enabled (used for nodeinfo, which I'll pr later). is there any way I can enable it via the package? it breaks gluon build otherwise

mkg20001: perhaps try_config?

mkg20001:
MMFD firewall rule split is in #2553 for seperate review

T-X:
> MMFD firewall rule split is in #2553 for seperate review

Thanks! This goes in the right direction. But maybe we could avoid even more duplicate definitions?

Maybe we could get it in a shape where the *gluon-mesh-babel package*, via *310-gluon-mesh-babel-firewall*, only needs to add the following firewall rule:

```
uci:section('firewall', 'rule', 'mesh_babel', {
	src = 'mesh',
	src_ip = 'fe80::/64',
	dest_port = '6696',
	proto = 'udp',
	target = 'ACCEPT',
})
```

And *gluon-mesh-olsrd*, via *310-gluon-mesh-olsrd-firewall* only adds this?

```
uci:section('firewall', 'rule', 'mesh_olsrd', {
	src = 'mesh',
	src_ip = 'fe80::/64',
	dest_port = '269',
	proto = 'udp',
	target = 'ACCEPT',
})
```

And move all the other rules like this maybe:

* fcc, fcm, fmc, fmm
  * to a common package, like gluon-mesh-layer3-common, make gluon-mesh-{olsrd,babeld} depend on it
* flc, fcl
  * -> gluon-l3roamd package
* mesh_respondd_mcast_ll, mesh_respondd_mcast2, mmfd_respondd_ll, mmfd_respondd_mesh
  * -> to gluon-mmfd

(@NeoRaider would that make sense to you, too, or any alternative suggestions?)

Btw. do you have some more insights on the firewall "HACK" comment in 360-gluon-mesh-olsrd-setup-intf, why would you need to add it there instead of in 310-gluon-mesh-babel-firewall? And would olsrd2 be fine with src_ip = fe80::/64 instead of "*"? (Or is there a bug, a regression that was added at some point, to BABEL, too?)

Also gluon-mesh-babel/luasrc/lib/gluon/upgrade/430-gluon-mesh-babel-add-mmfd-interface looks like it should be moved to the gluon-mmfd package, to be able to reuse it for olsrd, too?

T-X:
Also, I don't see where you configure the site.node_prefix6() in /etc/config/network at the moment. Am I missing something? Or should *package/gluon-mesh-babel/luasrc/lib/gluon/upgrade/300-gluon-mesh-babel-ip6* be moved to a gluon-mesh-layer3-common package, too?

mkg20001:
> Btw. do you have some more insights on the firewall "HACK" comment in 360-gluon-mesh-olsrd-setup-intf, why would you need to add it there instead of in 310-gluon-mesh-babel-firewall? And would olsrd2 be fine with src_ip = fe80::/64 instead of "*"?

That was from back when gluon_mesh wasn't used and we were hacking ip4 in with static. Now there's a patch on graz branch to have ip4 with gluon_mesh.

Maybe it'll just work.

As for fe80:: : OLSR just uses the first IP it finds. Currently the rule is to simply allow anything from mesh zone, as that just works. Plus this makes patching in ip4 less of a PITA for me.

Another thing: Do I need any rules for wired_mesh? Or is the actual mesh interface always in mesh zone?

package/gluon-mesh-olsrd/luasrc/lib/gluon/upgrade/360-gluon-mesh-olsrd-setup-intf
mkg20001: apparently not

mkg20001:
I've rebased the pull request against master, it should be much smaller now. This is now ready for review. I will fixup the commit messages after review.

T-X:
> I've rebased the pull request against master, it should be much smaller now. This is now ready for review. I will fixup the commit messages after review.

Shouldn't [310-gluon-mesh-olsrd-firewall](https://github.com/freifunk-gluon/gluon/pull/2535/files#diff-faee97edd2c925ba72ff05ac3d5938d0ee887088525cadb254ed8448d65d71d3) be removed, too?

mkg20001:
@T-X fixed

package/gluon-mesh-olsrd/luasrc/lib/gluon/upgrade/360-gluon-mesh-olsrd-setup-intf
T-X: The macvlan part is interesting, but I think it should be removed for now, too. If there is a need to use mesh-vpn + wan without vxlan, then this should be a separate pull-request. And should ideally behave the same and work both for babel and olsr2. With emphasizes on "if", I think @NeoRaider was not very keen on supporting mesh-on-wan/lan without vxlan.

contrib/ci/olsr-site/site.conf
```diff
@@ -0,0 +1,161 @@
...
+-- https://gluon.readthedocs.io/ for details.
+--
+-- This configuration will not work as is. You're required to make
+-- community specific changes to it!
+{
+  -- Used for generated hostnames, e.g. freifunk-abcdef123456. (optional)
+  -- hostname_prefix = 'freifunk-',
+
+  -- Name of the community.
+  site_name = 'Continious Integration',
```
T-X, line 13: typo: Continious -> Continuous

package/gluon-mesh-olsrd/luasrc/lib/gluon/upgrade/360-gluon-mesh-olsrd-setup-intf
mkg20001: it has been moved to the graz branch, but I'd like to still know the opinion of @NeoRaider on this, since it seems useful

package/gluon-mesh-olsrd/luasrc/lib/gluon/upgrade/360-gluon-mesh-olsrd-setup-intf
```diff
@@ -0,0 +1,166 @@
...
+	end
+
+	if site.prefix6() and not cfg.ip4_exclusive_mode(false) then
+		table.insert(lan, site.prefix6())
+	end
+
+	table.insert(addrs, 'default_accept')
+
+	uci:set("olsrd2", config, "originator", addrs)
+	uci:set("olsrd2", config, "lan", lan)
```
mkg20001, line 125: I'm not sure if it makes sense to specify the prefix6 as lan, since that is equal to HNA in olsr1 and this is prob not what we want

@5gbr ideas?

mkg20001: While we're at it: Not sure if the IPs given to olsr2 are supposed to be /128 or /64

mkg20001: it's /128 and the lan stuff is bs, case closed

mkg20001:
Meshing with role uplink+mesh set on a single interface and vxlan not set fails. The workarround for this would be macvlan, but that was rejected. But it's better that it's broken, otherwise mesh traffic could end up leaking for real and no one wants that mess.

mkg20001:
This PR is now done from my side and everything's tested. Note that clientap is not included, neither are any respondd modules or anything, that'll come later. Depends #2404 #2739

mkg20001:
#2404 is now merged and https://github.com/freifunk-gluon/gluon/pull/2739 is not relevant as a dependency without graz modules installed

christf (approved):
as discussed via IRC I like the changeset.

For some time in the future it would be interesting to remove the dependency on one single mesh-vpn interface. We should be able to support more than one.

package/gluon-mesh-olsrd/usr/lib/autoupdater/abort.d/10olsrd
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+. /lib/gluon/autoupdater/lib.sh
```
NeoRaider, line 3: These files under `package/gluon-mesh-olsrd/usr/` are unused, and there are similar files under `package/gluon-mesh-olsrd/files/`.

mkg20001:
tested with the change, node runs and meshes over ethernet. ssh to/from other nodes works.

AiyionPrime:
Merging this, as we agreed yesterday in the meetup.

PR #2541, opened 2022-06-02
ipq40xx: switch QCA4019 firmware to -ct

blocktrron:
Use the candelatech firmware for the QCA4019 WiSoC.

The Qualcomm firmware used for this specific chip in OpenWrt in 22.03
is experiencing heavily degraded performance due to excessive
retransmits when using A-MSDU. Disabling VHT modes or switching to the
candelatech firmware circumvents this issue.

### DIFF
--- a/targets/ipq40xx-generic
+++ b/targets/ipq40xx-generic
@@ -1,19 +1,5 @@
-local ATH10K_PACKAGES_IPQ40XX = {
-	'kmod-ath10k',
-	'-kmod-ath10k-ct',
-	'-kmod-ath10k-ct-smallbuffers',
-	'ath10k-firmware-qca4019',
-	'-ath10k-firmware-qca4019-ct',
-}
-local ATH10K_PACKAGES_IPQ40XX_QCA9888 = {
-	'kmod-ath10k',
-	'-kmod-ath10k-ct',
-	'-kmod-ath10k-ct-smallbuffers',
-	'ath10k-firmware-qca4019',
-	'-ath10k-firmware-qca4019-ct',
-	'ath10k-firmware-qca9888',
-	'-ath10k-firmware-qca9888-ct',
-}
+local ATH10K_PACKAGES_IPQ40XX = {}
+local ATH10K_PACKAGES_IPQ40XX_QCA9888 = {}
 
 
 defaults {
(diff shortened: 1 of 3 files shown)


NeoRaider:
Should we also switch to the ct kernel module? Maybe it would be a good idea to keep the default driver and module for all ath10k variants where ct supports 11s?

blocktrron:
@NeoRaider Verified mesh working for QCA9884 and QCA9886.

Regarding switching to -ct kernel-module: For sake of upstream consistency, I'd say this might be a good idea. The rationale for keeping the non-ct flavor was the lacking mesh-feature back in the day (which is only the case on Wave-1 nowadays).

T-X:
For ath9k/minstrel there are known issues that they become exceedingly aggressive when a neighbor vanishes without dissociation. Especially when there is data without congestion control from the network stack which keeps on pushing. In that case Minstrel goes down to 1 MBit/s, uses many retries and even uses CTS/RTS.

@blocktrron just out ouf curiousity, if ath10k firmware's internal rate control might do something similar to Minstrel-HT, did you have lingering neighbors in "iw dev wlanX station dump" with a high inactive timer? Or did you also see these many retries with only active neighbors/clients, too?

targets/ipq40xx-generic
```diff
@@ -1,18 +1,18 @@
...
+	'kmod-ath10k-ct',
 	'-kmod-ath10k-ct-smallbuffers',
-	'ath10k-firmware-qca4019',
-	'-ath10k-firmware-qca4019-ct',
-	'ath10k-firmware-qca9888',
-	'-ath10k-firmware-qca9888-ct',
+	'-ath10k-firmware-qca4019',
+	'ath10k-firmware-qca4019-ct',
+	'-ath10k-firmware-qca9888',
+	'ath10k-firmware-qca9888-ct',
```
NeoRaider, line 15: I think we should just drop these variables altogether where they don't change anything.

blocktrron: Thought the same, but i won't rule out we need to change it in the future, so I would keep them.

NeoRaider: Hmm, or keep the variable, but make it an empty list? That would also give us the smallbuffers configuration where OpenWrt defaults to it.

blocktrron: Empty list sounds like a good compromise.

rotanid:
there are issues with that change, see #2692

PR #2542, opened 2022-06-03
ath79-generic: (re)add support for TL-WR710N v1

AiyionPrime:
> Gone due to

@moridius tested the device yesterday and I confirmed his conclusion using a fresh build based on todays master:
in ath79 this device is currently a waste of energy;
powercycles every few minutes.

Even though the checklist is technically fulfilled this should be marked as **blocked**, until we decide if it can be saved or is done for good. (same decision as for 1043v1 I suppose)

- ~~Must be flashable from vendor firmware~~ **won't test**
  - ~~Web interface~~
  - ~~TFTP~~
  - ~~Other: <specify>~~
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs **not available**
    - ~~Should show activity~~
  - Switch port LEDs **not available**
    - ~~Should map to their respective port (or switch, if only one led present) ~~
- Outdoor devices only:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -94,6 +94,7 @@ ath79-generic
   - TL-WDR3500 (v1)
   - TL-WDR3600 (v1)
   - TL-WDR4300 (v1)
+  - TL-WR710N (v1)
   - TL-WR810N (v1)
   - TL-WR842N/ND (v3)
   - TL-WR1043N/ND (v2, v3, v4, v5)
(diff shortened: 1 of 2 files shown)


AiyionPrime:
Moridius answered yesterday: The device appears to be unusable in ath79 rn.

rotanid:
closing.
as discussed during our Gluon meetup today, we don't see a future for these kind of devices (8mb flash, 32mb memory)

PR #2543, opened 2022-06-03
ath79-generic: (re)add WBS510 v1

AiyionPrime:
> Gone due to

@dk8hc intends to test this device within the next two weeks.

- ~~Must be flashable from vendor firmware~~ **won't test**
  - ~~Web interface~~
  - ~~TFTP~~
  - ~~Other: <specify>~~
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [x] Should show activity
  - Switch port LEDs
- Outdoor devices only:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -99,6 +99,7 @@ ath79-generic
   - TL-WR1043N/ND (v2, v3, v4, v5)
   - WBS210 (v1.20)
   - WBS210 (v2.0)
+  - WBS510 (v1.20)
 
 * Ubiquiti
 
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/020-interfaces
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/020-interfaces
@@ -48,6 +48,7 @@ if platform.match('ath79', 'generic', {
 	'tplink,cpe510-v1',
 	'tplink,wbs210-v1',
 	'tplink,wbs210-v2',
+	'tplink,wbs510-v1',
 	'ubnt,nanostation-m-xw',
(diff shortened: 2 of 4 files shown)


AiyionPrime:
LAN was on the wrong port and apparently needs to be swapped.

AiyionPrime:
Wrongly rebased, sorry.

AiyionPrime:
Should be fine now, @dk8hc announced to do the (hopefully) final test this weekend.

AiyionPrime:
@blocktrron, @NeoRaider: left unanswered in #gluon for now:

@dk8hc tested the current patch and came to the conclusion tha my addition here:

https://github.com/freifunk-gluon/gluon/pull/2543/files#diff-9683bc89cf803c758bfcc7b5f963bfbaf22d650a191a00743b697286d426bb4dR51

does not affect the lan wan assignment at all.

Any ideas why that would be?

NeoRaider:
> @blocktrron, @NeoRaider: left unanswered in #gluon for now:
> 
> @dk8hc tested the current patch and came to the conclusion that my addition here:
> 
> https://github.com/freifunk-gluon/gluon/pull/2543/files#diff-9683bc89cf803c758bfcc7b5f963bfbaf22d650a191a00743b697286d426bb4dR51
> 
> does not affect the lan wan assignment at all.
> 
> Any ideas why that would be?

Hmm, no idea, the code looks correct.

AiyionPrime:
Maybe something's up with the testsetup. I'll see if I can get access to the device.

AiyionPrime:
Testsetup was fine all along.
Turns out, the buildroot contained a backup of an early (unconditional) version of `115-swconfig`...

My bad. This works very well now and is ready for review.

PR #1889, opened 2019-12-09
lantiq-xrx200: add support for AVM FRITZ!Box 7360 (v1, v2) and 7360 SL

Dark4MD:
* [x]  must be flashable from vendor firmware
      
      * [ ]  webinterface
      * [ ]  tftp
      * [x]  other: fritzfla.sh script or adam2 ftp

    * [x]  must support upgrade mechanism
      
      * [x]  must have working sysupgrade
        
        * [x]  must keep/forget configuration (if applicable)
          _think `sysupgrade [-n]` or `firstboot`_
      * [?] must have working autoupdate
        _usually means: gluon profile name must match image name_

    * [x]  wps button must return device into config mode

    * [n/a]  primary mac should match address on device label (or packaging) (https://gluon.readthedocs.io/en/latest/dev/hardware.html#notes)

    * wired network
      
      * [x]  should support all network ports on the device
      * [x]  must have correct port assignment (WAN/LAN)

    * wifi (if applicable)
      
      * [x]  association with AP must be possible on all radios
      * [x]  association with 802.11s mesh must be working on all radios
      * [x]  ap/mesh mode must work in parallel on all radios

    * led mapping
      
      * power/sys led (_critical, because led definitions are setup on firstboot only_)
        
        * [x]  lit while the device is on
        * [x]  should display config mode blink sequence
          (https://gluon.readthedocs.io/en/latest/features/configmode.html)
      * radio leds
        
        * [x]  should map to their respective radio
        * [x]  should show activity
      * switchport leds
        
        * [n/a]  should map to their respective port (or switch, if only one led present)
        * [n/a]  should show link state and activity

    * outdoor devices only
      
      * [n/a]  added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`
        ~
All three versions share the same Hardware except the Flash (32MB for v2 else 16MB) and no POTS for SL. Tested on v2 so the bigger Flash is definitely no Issue.
Flashing with modified fritzfla.sh script successfully tested.

PR #2544, opened 2022-06-11
lantiq-xrx200: split 7360v2 into own device

AiyionPrime:
Talked with @Dark4MD about the device on Thursday and we came to the conclusion, that it could not be using all of it's available memory if it had the same image as `v1`/`sl`.

@grische from @freifunkMUC encountered an issue with the current implementation as well, as it's currently not flashable on their stable or next branch:

```console
root@OpenWrt:/tmp# sysupgrade -n gluon-ffmuc-v2022.5.2-next3-avm-fritz-box-7360-v2-sysupgrade.bin
Sat Apr 16 13:21:46 UTC 2022 upgrade: Device avm,fritz7360-v2 not supported by this image
Sat Apr 16 13:21:46 UTC 2022 upgrade: Supported devices: avm,fritz7360sl FRITZ7360SL
```

I think this would be sufficient, let me know if I forgot something.
I'll ask @grische to test the PR before opening it for merging.

### DIFF
--- a/targets/lantiq-xrx200
+++ b/targets/lantiq-xrx200
@@ -32,7 +32,11 @@ device('avm-fritz-box-3370-rev-2-micron-nand', 'avm_fritz3370-rev2-micron', {
 
 device('avm-fritz-box-7360-sl', 'avm_fritz7360sl', {
 	factory = false,
-	aliases = {'avm-fritz-box-7360-v1', 'avm-fritz-box-7360-v2'},
+	aliases = {'avm-fritz-box-7360-v1'},
+})
+
+device('avm-fritz-box-7360-v2', 'avm_fritz7360-v2', {
+	factory = false,
 })
 
 device('avm-fritz-box-7362-sl', 'avm_fritz7362sl', {


targets/lantiq-xrx200
```diff
@@ -32,7 +32,14 @@ device('avm-fritz-box-3370-rev-2-micron-nand', 'avm_fritz3370-rev2-micron', {
...
 device('avm-fritz-box-7360-sl', 'avm_fritz7360sl', {
 	factory = false,
-	aliases = {'avm-fritz-box-7360-v1', 'avm-fritz-box-7360-v2'},
+	aliases = {'avm-fritz-box-7360-v1'},
+})
+
+device('avm-fritz-box-7360-v2', 'avm_fritz7360-v2', {
+	factory = false,
+	manifest_aliases = {
+		'avm,fritz7360sl' -- upgrade from OpenWrt 22.03
```
NeoRaider, line 41: The alias is definitely incorrect, `image_name()` will never return a string with a comma in it.

AiyionPrime: Thanks for spotting

targets/lantiq-xrx200
```diff
@@ -32,7 +32,14 @@ device('avm-fritz-box-3370-rev-2-micron-nand', 'avm_fritz3370-rev2-micron', {
...
 device('avm-fritz-box-7360-sl', 'avm_fritz7360sl', {
 	factory = false,
-	aliases = {'avm-fritz-box-7360-v1', 'avm-fritz-box-7360-v2'},
+	aliases = {'avm-fritz-box-7360-v1'},
+})
+
+device('avm-fritz-box-7360-v2', 'avm_fritz7360-v2', {
+	factory = false,
+	manifest_aliases = {
+		'avm-fritz-box-7360-sl' -- upgrade from OpenWrt 22.03
```
grische, line 41: The comment seems wrong. OpenWRT 22.03 correctly identifies the device as `avm,fritz7360-v2`.

AiyionPrime: you can read the marker as:
this alias can be dropped once we drop support for Gluon-Releases based on OpenWrt 22.03

AiyionPrime:
@grische and I agreed it would be enough as a first step to split this up.
We can still figure out the best way to reclaim the unused memory on wrongly flashed `SL`s and `v1`s.

But as he put it in the PR:

> Tested on v2 so the bigger Flash is definitely no Issue.

_Originally posted by @Dark4MD  in https://github.com/freifunk-gluon/gluon/pull/1889#issue-535000834_

grische (approved):

PR #2546, opened 2022-06-13
ath79-generic: (re)add TL-WR902AC v1

Dark4MD:
  - [x] Web interface
  - [x] TFTP
  - [n/a] Other: <specify>
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [x] Should show activity
  - Switch port LEDs
- Outdoor devices only:
  - [n/a] Added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`
Tested device: https://hannover.freifunk.net/karte/#/de/map/7c8bca3f7b66

### DIFF
--- a/targets/ath79-generic
+++ b/targets/ath79-generic
@@ -397,6 +397,12 @@ device('tp-link-tl-wr842n-v3', 'tplink_tl-wr842n-v3', {
 	},
 })
 
+device('tp-link-tl-wr902ac-v1', 'tplink_tl-wr902ac-v1', {
+	packages = ATH10K_PACKAGES_QCA9887,
+	broken = true, -- OOM with 5GHz enabled in most environments
+	class = 'tiny', -- 64M ath9k + ath10k
+})
+
 device('tp-link-tl-wr1043nd-v2', 'tplink_tl-wr1043nd-v2', {
 	manifest_aliases = {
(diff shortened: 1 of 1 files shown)


AiyionPrime:
The load of the device is likely to be lower on other devices, as his test device has WireGuard enabled but no connection to a gateway.
wgpeerselector will work overtime in order to change that as fast as possible.

@Dark4MD might wanna connect a gateway or disable meshvpn on the device :)

Dark4MD:
> @Dark4MD might wanna connect a gateway or disable meshvpn on the device :)

Added the Key to our rep, VPN should be enabled any minute now.

Will test another build with "packages = ATH10K_PACKAGES_QCA9887," as @AiyionPrime mentioned that it might be missing and actually the 5ghz wireless radio is missing at the moment without it.

AiyionPrime:
@Dark4MD just flashed a clean vH25 (ar71xx), upgraded to a fresh build of your PR and can confirm, starting from
> Must support upgrade mechanism

everything including 5GHz mesh works.

Please open the PR for merging.

AiyionPrime (approved):
Does work as expected, I just tested everything but stock-flashing.

Dark4MD:
> 
> everything including 5GHz mesh works.

guess it was simply my build then that messed that up.

PR #2547, opened 2022-06-14
ath79-generic: (re)add support for NanoStation Loco M2 (XW)

AiyionPrime:
> Gone due to

"nordm4nn" from @freifunkMUC intends to test this device.
Will likely be done by tomorrow evening.

- ~~Must be flashable from vendor firmware~~ **wont test**
  - ~~Web interface~~
  - ~~TFTP~~
  - ~~Other: <specify>~~
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [x] Should show activity
  - Switch port LEDs
- Outdoor devices only:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -103,6 +103,7 @@ ath79-generic
 * Ubiquiti
 
   - NanoBeam M5 (XW)
+  - NanoStation Loco M2 (XW)
   - NanoStation M2/M5 (XW)
   - UniFi AC Lite
   - UniFi AC LR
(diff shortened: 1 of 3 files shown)


AiyionPrime:
"nordm4nn" concluded his tests, this is ready to review.

docs/user/supported_devices.rst
```diff
@@ -102,6 +102,7 @@ ath79-generic
 
 * Ubiquiti
 
+  - Loco M2 XW
```
rotanid, line 105: the device should be called like the others are:
NanoStation Loco M2 (XW)

this also applies to the commit text

AiyionPrime: Indeed, thanks

AiyionPrime: resolved in 2d481765ee1732ec4f8066bad284800fd642b630

rotanid (approved):

PR #2552, opened 2022-06-16
ath79-nand: (re)add hiveap-121

AiyionPrime:
> ath79-nand: (re)add hiveap-121
>     
>     Gone due to
>     commit 45c84a117bf8 ("ar71xx: drop target")

> ath79-nand: add vendor to GL.iNet devices

@kpanic23 intends to test this device starting monday.

  - ~~Web interface~~
  - ~~TFTP~~
  - ~~Other: install via serial, and load via tftp following the commit message https://github.com/openwrt/openwrt/commit/62db255543c8fb5d1787caf36c6c7c9c51db71cd~~
  - [x] Other other: In case the bootloader does not allow regular flashing, it does allow flashing these ancient initramfs files: https://github.com/riptidewave93/LEDE-HiveAP-121 (in Releases)
- Wired network
- LED mapping
  - Power/system LED
  - ~~Radio LEDs~~ none
    - ~~Should show activity~~
  - ~~Switch port LEDs~~ has none
- Outdoor devices only:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -139,6 +139,10 @@ ath79-generic
 ath79-nand
 ----------
 
+* Aerohive
+
+  - HiveAP 121
+
 * GL.iNet
 
   - GL-AR300M
--- a/targets/ath79-nand
+++ b/targets/ath79-nand
@@ -9,6 +9,11 @@ local ATH10K_PACKAGES_QCA9887 = {
 local ATH10K_PACKAGES_QCA9888 = {}
 
 
+-- Aerohive
+
+device('aerohive-hiveap-121', 'aerohive_hiveap-121')
+
+
 -- GL.iNet
 
 device('gl.inet-gl-ar300m-nor', 'glinet_gl-ar300m-nor', {


AiyionPrime:
The patch appears to work well. Checklist follows, when @kpanic23 gets to it.

@blocktrron The patch I'd like to hand in at OpenWrt would use symlinks instead of copies, so would not increase the sysupgrade-imagesize this drastically.
It will contain some logic to handle symlinks in the updater, so it will not help in this migration.

I'd recommend to keep this patch as long as the migrations from ar71xx.

Djfe:
How is this coming along? Do you need an active tester?
There are 5 of these installed and used in Freifunk Aachen. I could try getting into contact with one of the owners to avoid this from being dropped.

AiyionPrime:
There are few options for this device; one is to re-add it without adding a migration.
Owners of the device will need to manually update their router.
@kpanic23 received an image for testing this approach last week.
He'll get to it.

The other approach would have been a proper migration, involving upstream patches of tar based sysupgrades, backporting them and then a two image based update; one to have the migration and one two recognize it.
This would have been a little to much effort for the few devices deployed.

A compromise would have included keeping a downstream patch. And at least one update would have been double the size of a regular update.
@blocktrron didn't like that approach to much.
So unless someone is willing to refactor and upstream the sysupgrade-tar-patch, this will stay without migration. Sorry :/

Djfe:
how does the manual user interaction look like? do they have to return to factory and then flash?

AiyionPrime:
> how does the manual user interaction look like? do they have to return to factory and then flash?

`sysupgrade -nF <image>`

> edit: is the two image approach used for other ath79 devices as well?

Upstream there might be more, anything that used `sysupgrade-tar` and changed its `board_name`.

> what do we need two updates for?

While you need the update to contain a reference to the old `board_name`, you also need the update process in the router to be able to handle such references correctly.
Neither is done yet, so if you currently implemented such references by e.g. symlinks, the currently running images won't be able to unpack the update properly.
So you'll need an update of the stable branch which contains code that handles such references/symlinks properly.

> (is there maybe a post on an issue that explains this further?

I don't think so.
You can look at past commits of this PR to see the compromise I described above.
https://github.com/freifunk-gluon/gluon/commit/82fd03556a9c71167aca7be682dbfd5599568ee3

Maybe some past protocol of a developer meetup contains an explanation; I explained it and our options at some point, when blocktrron dismissed the approach with the bloated image and the staying patch in the old-stable branch.

> I get what you just explained to me but this is the first I read about needing two updates to complete the dsa migration and I wonder why.

> It's not the DSA migration, that requires a second update, but the implicit modification of `sysupgrade-tar`, which is necessary due to changes in the `board_name` of the device, which happened while cleaning up devices in the migration.

> Are there sysupgrade paths that were never modified upstream to support dsa migration and tar is one of those rarer paths?)

Not many anymore, I'd say. And this is not necessary for the DSA-migration, but just happened during it.

blocktrron:
CLosing as discussed at last meeting. This can be re-opened once someone is interested to pursue this further.

AiyionPrime:
@kpanic23 Just finished testing the proposed changes. Results are above, this is ready for review.

AiyionPrime:
I just ran `make lint-editorconfig` locally, no issues.
GitHubs networking ftw.

Djfe:
update-path from v2021 is not working. see #3181

PR #2554, opened 2022-06-16
ath79-nand: (re)add WNDR4300v1

AiyionPrime:
> Gone due to
> commit 45c84a1 ("ar71xx: drop target")

~~"alexp" from hackint#ffda intends to test this device.~~

- Must be flashable from vendor firmware
  - ~~Web interface~~
  - ~~TFTP~~
  - Other: https://github.com/jclehner/nmrpflash
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [x] Should show activity
  - Switch port LEDs
- Outdoor devices only:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -146,6 +146,7 @@ ath79-nand
 * Netgear
 
   - WNDR3700 (v4)
+  - WNDR4300 (v1)
 
 ath79-mikrotik
 --------------
--- a/targets/ath79-nand
+++ b/targets/ath79-nand
@@ -33,6 +33,10 @@ device('netgear-wndr3700-v4', 'netgear_wndr3700-v4', {
 	},
 })
 
+device('netgear-wndr4300', 'netgear_wndr4300', {
+	factory_ext = '.img',
+})
+
 
 -- ZTE
 


AiyionPrime:
A misunderstanding, he thought we'd need a TP-Link device with 4300 in its name.

AiyionPrime:
The kernel partition size was increased in https://github.com/openwrt/openwrt/commit/0d28e5d6440d2a37841a207f943e6e5a23172883 which broke ar71xx compatibility.

I've got the device laying here, I could add it as a new device for the eight users of similar devices. Any recommendations on how to proceed?

AiyionPrime:
I'd suggest we add it as device again, and mark the missing migration in the releasenotes.
OpenWrt removed `SUPPORTED_DEVICES` entry, so that should be fine.
Any objections?

blocktrron:
@AiyionPrime No, we don't add the device without completing the integration verification please.

AiyionPrime:
I did not intend to, sorry.
Wanted to express that I'll intend to leave the code as is, not that I'll be merging this untested in a moment.

blocktrron:
AH, I've misread that (And also the checkboxes were not there for me with your last comment?).

Missing migration should be okay. There is nothing reasonable to do for us anyways.

PR #2561, opened 2022-06-18
openwrt: revert mesh-breaking commit

### LINKED ISSUE #2559: 802.11s meshing broken in master branch due to a change in OpenWrt
802.11s meshing doesn't work anymore, sometimes it works but it's messy.

the problem starts with commit 52d4ae5ad71fb8d9a96acb0f48486d9a57f8facd and more specifically with OpenWrt upstream commit https://github.com/openwrt/openwrt/commit/33df033b73365487c5bb5a58b77aed04d4ca6ac1 titled `mac80211: add a bug fix for a rare crash`

the issue was observed on multiple different device in at least two targets: ath79-generic, ramips-mt7621
due to the nature of the bug, i have reason to believe all devices and targets are affected

blocktrron:
This commit reportedly breaks 11s meshing, rendering mesh-only nodes
useless.

Fixes #2559

### DIFF
--- a/patches/openwrt/0004-Revert-mac80211-add-a-bug-fix-for-a-rare-crash.patch
+++ b/patches/openwrt/0004-Revert-mac80211-add-a-bug-fix-for-a-rare-crash.patch
@@ -1,0 +1,53 @@
+From: David Bauer <mail@david-bauer.net>
+Date: Sat, 18 Jun 2022 03:01:39 +0200
+Subject: Revert "mac80211: add a bug fix for a rare crash"
+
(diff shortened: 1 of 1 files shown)


rotanid:
thanks, unfortunately this does not do the trick for me, i did a make clean and built an image with #2556 and meshing still doesn't work, although it did work when manually reverting in modules to openwrt 22.03 d04a947b6b2fee6359aa6126068f70efaa3e48bc
any suggestion?

blocktrron:
@rotanid Tested this myself on a ASUS RT-AC57U and mesh works again with this PR

rotanid:
superseeded by #2563

PR #2563, opened 2022-06-18
modules: update to latest upstream HEAD

### LINKED ISSUE #2559: 802.11s meshing broken in master branch due to a change in OpenWrt

blocktrron:
Fixes #2559


rotanid (approved):
works for me on ramips-mt7621

PR #2565, opened 2022-06-21
mediatek-mt7622: add support for Linksys E8450

mbaumga:
  - [ ] Web interface
  - [ ] TFTP
  - [x] Other: https://github.com/dangowrt/owrt-ubi-installer
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs (None available)
    - [ ] Should show activity
  - Switch port LEDs
    - [X] Should map to their respective port (or switch, if only one led present) (Only WAN Port)
- Outdoor devices only:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -226,6 +226,10 @@ lantiq-xway
 mediatek-mt7622
 ---------------
 
+* Linksys
+
+  - E8450
+
 * Ubiquiti
 
   - UniFi 6 LR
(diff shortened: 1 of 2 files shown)


mbaumga:
Since according to the OpenWrt wiki the only way to flash the Linksys e8450 with the "non-UBI" way, I took the liberty to only support the "UBI" way for now.

https://openwrt.org/toh/linksys/e8450#installation

This is of course up for discussion, but I don't really see a reason to do this otherwise at the moment.

PR #2566, opened 2022-06-21
mediatek-mt7622: add support for Xiaomi AX3200

mbaumga:
  - [ ] Web interface
  - [ ] TFTP
  - [ ] Other: 
                   https://openwrt.org/toh/xiaomi/ax3200#installation
                   https://forum.openwrt.org/t/adding-openwrt-support-for-xiaomi-redmi-router-ax6s-xiaomi-router-ax3200/111085/513
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs (None available)
    - [ ] Should show activity
  - Switch port LEDs
    - [X] Should map to their respective port (or switch, if only one led present) (Only WAN Port)
- Outdoor devices only:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -230,6 +230,10 @@ mediatek-mt7622
 
   - UniFi 6 LR
 
+* Xiaomi
+
+  - AX3200 (RB03)
+
 mpc85xx-p1010
 -------------
 
--- a/targets/mediatek-mt7622
+++ b/targets/mediatek-mt7622
@@ -1,3 +1,10 @@
 device('ubiquiti-unifi-6-lr', 'ubnt_unifi-6-lr', {
 	factory = false,
 })
+
+-- Xiaomi
+
+device('xiaomi-redmi-router-ax6s', 'xiaomi_redmi-router-ax6s', {
+	factory = false,
+})
+


mbaumga:
According to the installation instructions of the OpenWrt wiki, this device should be flashable via telnet, if it is enabled by default from factory. This is the case in the RB03 model (chinese version).
https://openwrt.org/toh/xiaomi/ax3200#installation

If the device is a RB01 model (international version), the telnet is disabled by default and there is no official way on how to flash a firmware.
An unofficial way is described here:
https://forum.openwrt.org/t/adding-openwrt-support-for-xiaomi-redmi-router-ax6s-xiaomi-router-ax3200/111085/513
That's how I flashed my device initially to OpenWrt.

I marked this pull request as draft because of this reason, because there is no official way to flash the RB01 model.

GoliathLabs:
@mbaumga I have always used OpenWRTInvasion for my Xiaomi router. I do not have a Xiaomi AX3200 myself. However, is it possible that its firmware still has the vulnerability? Or is it a completely different firmware than the one of e.g. 4A Gigabit Edition?

mbaumga:
@GoliathLabs According to the OpenWrt Forum it seems unlikely that OpenWrtInvasion works. 
https://forum.openwrt.org/t/adding-openwrt-support-for-xiaomi-redmi-router-ax6s-xiaomi-router-ax3200/111085/90

In addition the Author of the thread himself crossed out pretty much everything regarding exploits and recommends the UART method.
https://forum.openwrt.org/t/adding-openwrt-support-for-xiaomi-redmi-router-ax6s-xiaomi-router-ax3200/111085

blocktrron:
Please add  `(RB03)` to the documentation entry for this device. Apart from that, the PR looks good and i see we can move this forward.

mbaumga:
> Please add `(RB03)` to the documentation entry for this device. Apart from that, the PR looks good and i see we can move this forward.

Should be done.

PR #2567, opened 2022-06-22
Fix make lint

AiyionPrime:
`make lint` throws three errors, this should fix it.

I think `gluon-enter-setup-mode` never had `pipefail`-support, if we want to change that, I could implement that as well, but I suppose it's not necessary int this case, as its output appears to be not evaluated.

### DIFF
--- a/package/gluon-setup-mode/files/usr/bin/gluon-enter-setup-mode
+++ b/package/gluon-setup-mode/files/usr/bin/gluon-enter-setup-mode
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-set -euo pipefail
+set -eu
 
 echo "Entering setup mode and rebooting..."
 uci set gluon-setup-mode.@setup_mode[0].enabled='1'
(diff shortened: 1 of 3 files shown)


package/gluon-setup-mode/files/usr/bin/gluon-enter-setup-mode
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-set -euo pipefail
+set -eu
```
rotanid, line 3: this change does a bit more than cosmetics, could you argue why we need to do this and how we make sure this doesn't break?

AiyionPrime: posix-sh does not support the option pipefail.
bash does, dash does, sh does not.

I think it currently does nothing but make you believe the script would respect pipefail, while it does not.

If we want pipefail-like behavior I could implement that as well.

NeoRaider:
We should document which shellcheck version was used to test this in the commit messages - current scripts obviously passed our CI before, so it seems like this is about some other version?

Also note that the pipefail option only has an effect on pipe constructs in the same script, so it is fine to remove it in this case. However, ash does support pipefail, and it is often considered good style to simply include it in all scripts. If shellcheck now reports it for Gluon's scripts, we need to fix our lint command and not our scripts.

AiyionPrime:
I ran version: `0.8.0`.

Using shellcheck for ash scripts in my experience only gives mediocre results due to this:
https://www.shellcheck.net/wiki/SC2187

PR #2569, opened 2022-06-25
gluon-l3roamd: site: make node_client_prefix6 optional+deprecated

T-X:
The site.node_client_prefix6() is only used internally by the l3roamd
protocol. Therefore it is unnecessary to expose it to an administrator.
    
Instead, if node_client_prefix6 is unspecified in the site, generate an
IPv6 Unique Local Address prefix from the site domain_seed.
    
This updates the site documentation as well and marks this setting as
both optional and deprecated.
    
Note: If you had the node_client_prefix6 specified before and want to
use the new autogeneration from the domain_seed instead then this will
break compatibility and will need a gluon-scheduled-domain switch.
    
### DIFF
--- a/package/gluon-mesh-babel/check_site.lua
+++ b/package/gluon-mesh-babel/check_site.lua
@@ -1,5 +1,4 @@
 need_string_match(in_domain({'node_prefix6'}), '^[%x:]+/64$')
-need_string_match(in_domain({'node_client_prefix6'}), '^[%x:]+/64$')
 
 need_string_match(in_domain({'next_node', 'ip6'}), '^[%x:]+$', false)
 need_string_match(in_domain({'next_node', 'ip4'}), '^%d+.%d+.%d+.%d+$', false)
--- a/package/gluon-mesh-babel/luasrc/lib/gluon/upgrade/300-gluon-mesh-babel-mkconfig
+++ b/package/gluon-mesh-babel/luasrc/lib/gluon/upgrade/300-gluon-mesh-babel-mkconfig
@@ -1,6 +1,7 @@
 #!/usr/bin/lua
 
 local site = require 'gluon.site'
+local l3 = require 'gluon.l3'
 local uci = require('simple-uci').cursor()
 local nodeip = uci:get('network', 'loopback', 'ip6addr'):match('^[^/]+')
 local babelconf='/etc/gluon-babeld.conf'
@@ -14,7 +15,7 @@ file:write("import-table 254\n")
 file:write("out ip " .. site.next_node.ip6() .. "/128 deny\n")
 file:write("redistribute ip " .. site.next_node.ip6() .. "/128 deny\n")
 file:write("redistribute ip " .. site.prefix6() .. " eq 128  allow\n")
-file:write("redistribute ip " .. site.node_client_prefix6() .. " eq 128  allow\n")
+file:write("redistribute ip " .. l3.node_client_prefix6() .. " eq 128  allow\n")
 file:write("redistribute ip " .. site.node_prefix6() .. " eq 128  allow\n")
 file:write("redistribute ip 2000::/3 allow\n")
 file:write("redistribute local if br-wan deny\n")
--- a/package/gluon-mesh-layer3-common/luasrc/usr/lib/lua/gluon/l3.lua
+++ b/package/gluon-mesh-layer3-common/luasrc/usr/lib/lua/gluon/l3.lua
@@ -1,0 +1,20 @@
+local site = require("gluon.site")
+local util = require("gluon.util")
+
+local M = {}
+
+-- returns a prefix generated from the domain-seed
+-- for l3roamd -P <node-client-prefix>
+function M.node_client_prefix6()
+	local key = "gluon-l3roamd.node_client_prefix6"
+	local prefix = site.node_client_prefix6()
+
+	if not prefix then
+		local prefix_seed = util.domain_seed_bytes(key, 7)
+		prefix = ("fd" .. prefix_seed):gsub(("(%x%x%x%x)"):rep(4), "%1:%2:%3:%4" .. "::/64")
+	end
+
+	return prefix
+end
+
+return M
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -49,10 +49,13 @@ node_prefix6
 
     node_prefix6 = 'fdca::ffee:babe:2::/64'
 
-node_client_prefix6
-  The ipv6 prefix from which the client-specific IP-address is calculated that
-  is assigned to each node by l3roamd to allow efficient communication when 
-  roaming. This is exclusively useful when running a routing mesh protocol
+node_client_prefix6 \: optional, deprecated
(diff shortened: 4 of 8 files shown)


T-X:
Commit is rebased on top of #2553.

Tested that the node_client_prefix6() is both correct in the l3roamd command line and in the babel config file. Still ToDo: Test roaming of a client device with the autogenerated prefix.

@mkg20001 @christf

package/gluon-mesh-babel/check_site.lua
```diff
@@ -1,4 +1,5 @@
-need_string_match(in_domain({'node_client_prefix6'}), '^[%x:]+/64$')
+need_string_match(in_domain({'prefix6'}), '^[%x:]+/64$')
+need_string_match(in_domain({'node_prefix6'}), '^[%x:]+/64$')
```
mkg20001, line 2: this is already required by layer3-common

package/gluon-mesh-babel/check_site.lua
```diff
@@ -1,4 +1,5 @@
-need_string_match(in_domain({'node_client_prefix6'}), '^[%x:]+/64$')
+need_string_match(in_domain({'prefix6'}), '^[%x:]+/64$')
```
mkg20001, line 1: this is already required by gluon-core

mkg20001 (commented):
not sure if we really need to have a unique l3roamd prefix. it might make sense to have it domain-specific aswell though then, which maybe would make handling of accidental cross-domain meshing better.

package/gluon-mesh-babel/luasrc/lib/gluon/upgrade/300-gluon-mesh-babel-mkconfig
```diff
@@ -14,7 +15,7 @@ file:write("import-table 254\n")
 file:write("out ip " .. site.next_node.ip6() .. "/128 deny\n")
 file:write("redistribute ip " .. site.next_node.ip6() .. "/128 deny\n")
 file:write("redistribute ip " .. site.prefix6() .. " eq 128  allow\n")
-file:write("redistribute ip " .. site.node_client_prefix6() .. " eq 128  allow\n")
+file:write("redistribute ip " .. l3.node_client_prefix6() .. " eq 128  allow\n")
```
mkg20001, line 18: I don't know if this is correct at all, since the prefix per doc doesn't seem to be used for routing packets? Or is it being used for that in some kindof crazy way?

T-X: As far as I understand l3roamd uses this prefix to announce the roaming of a client to another node. The receiving node listens on an address in this range with IP_FREEBIND. To be able to receive such an announcement, nodes between the old and the new node for the client that roamed need to forward these l3roamd packets with this unicast node-client-IP destination address.

package/gluon-mesh-babel/check_site.lua
T-X: I think I'd add that anyways, even if a dependency pulls in the same check, too. Tracking if a check was added by another dependency seems tedious. Also if someone were to remove the usage of the site setting for gluon-layer3-common later, there is a high chance that the check would be removed in gluon-layer3-common, too and that the needed check in gluon-mesh-babel would then be missed.

Therefore it seems the most simple and less error prone approach to me, to stick to this simple rule: If this package requires a site/domain setting, then add it to this package's check_site.lua. Even if another dependcy package has it in its check_site.lua as well.

package/gluon-mesh-babel/check_site.lua
T-X: same reasoning as above

package/gluon-layer3-common/luasrc/usr/lib/lua/gluon/l3.lua
```diff
@@ -0,0 +1,21 @@
+local util = require("gluon.util")
+local site = require("gluon.site")
+
+local M = {}
+
+-- returns a prefix generated from the domain-seed
+-- for l3roamd -P <node-client-prefix>
+function M.node_client_prefix6()
+	local key = "gluon-l3roamd.node_client_prefix6"
```
mkg20001, line 9: This key could include the current domain, for the client address to be unique per domain. Would that be useful?

T-X: I think that's what should already be happening a few lines later, with the call to `util.domain_seed_bytes(key, 7)`. The key + the site's/domain's domain_seed will be the seeds to generate a prefix unique for this domain.

Check the other usages of domain_seed_bytes() in Gluon. We are using it like `...domain_seed_bytes("gluon-mesh-vxlan", 3)' to generate a VXLAN ID which is unique per domain, as well.

mkg20001: Ah, perfect

mkg20001:
@T-X the firewall pr is now merged, could you rebase this then? Other than that LGTM

T-X:
Changelog v2:

* rebased to current master (renamed gluon-layer3-common to gluon-mesh-layer3-common)
* added node_client_prefix6 to gluon-mesh-layer3-common's check_site.lua, with required=false
* removed additions of prefix6 and node_prefix6 in check_site.lua again

package/gluon-mesh-babel/luasrc/lib/gluon/upgrade/300-gluon-mesh-babel-mkconfig
christf: this is exactly how it works.

mkg20001:
I think this is good for merge.

mkg20001 (approved):
LGTM

package/gluon-mesh-layer3-common/luasrc/usr/lib/lua/gluon/l3.lua
```diff
@@ -0,0 +1,21 @@
...
+local M = {}
+
+-- returns a prefix generated from the domain-seed
+-- for l3roamd -P <node-client-prefix>
+function M.node_client_prefix6()
+	local key = "gluon-l3roamd.node_client_prefix6"
+	local prefix = site.node_client_prefix6()
+
+	if not prefix
+	then
```
blocktrron, line 13: we usually put this in a single line.

blocktrron (approved):
LGTM apart from the minor style improvement.

AiyionPrime (approved):
Will merge this when the actions ran through.

package/gluon-mesh-layer3-common/luasrc/usr/lib/lua/gluon/l3.lua
AiyionPrime: resolved in [b3a9221](https://github.com/freifunk-gluon/gluon/commit/b3a9221b07f083d95aff1b3057ccf5dafa7efd42)

PR #2573, opened 2022-07-04
ipq40xx-mikrotik: add support for mikrotik-sxtsq-5-ac

timniklas:
https://openwrt.org/toh/hwdata/mikrotik/mikrotik_sxtsq_5_ac_rbsxtsqg-5acd


AiyionPrime:
Good morning @timniklas and thanks for your proposed contribution.
Please copy the [Device-Integration-checklist](https://github.com/freifunk-gluon/gluon/wiki/Device-Integration-checklist) and mark your results according to your tests, as untested devices are usually not merged into master.
The device looks interesting though, looking forward to your build and test results.

timniklas:
  - [ ] Web interface
  - [X] TFTP
  - [ ] Other: <specify>
        (`lua -e 'print(require("platform_info").get_image_name())'`) -> mikrotik-sxtsq-5-ac-rbsxtsqg-5acd
- Wired network
- LED mapping
  - Power/system LED -> there are none
  - Radio LEDs
    - [X] Should show activity
  - Switch port LEDs  -> there are none
- Outdoor devices only:

timniklas:
image

targets/ipq40xx-mikrotik
```diff
@@ -0,0 +1,6 @@
+include 'mikrotik.inc'
+
+device('mikrotik-sxtsq-5-ac', 'mikrotik_sxtsq-5-ac', {
```
AiyionPrime, line 3:
```suggestion
device('mikrotik-sxtsq-5-ac-rbsxtsqg-5acd', 'mikrotik_sxtsq-5-ac', {
```

AiyionPrime: This should result in a matching string referencing the [modelname](https://github.com/openwrt/openwrt/blob/ebfe66e494e57f4b421f1190d6bff1d361db1b3d/target/linux/ipq40xx/image/mikrotik.mk#L60) from OpenWrt.

AiyionPrime (changes requested):
While you're looking at the suggestions, you might look at your commit message as well. Should become something like `ipq40xx-mikrotik:  add support for mikrotik-sxtsq-5-ac` or similar.

I'm fairly new too this compared to other actives, so if I left something out it was not ill intended, but I did just not see it.

targets/ipq40xx-mikrotik
```diff
@@ -0,0 +1,6 @@
...
+	factory = false,
+	aliases = {'mikrotik-discg-5acd'},
```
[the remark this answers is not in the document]
timniklas: Im actualy using a MikroTik RBDiscG-5acD not a RBSXTsqG-5acD but in openwrt there is no seperate build for that device since the hardware is identical.
image

AiyionPrime (approved):

AiyionPrime:
Looks promising, I think. You could squash the five commits down into the first one, so the CI can do it's job.

mbaumga:
Don't forget to add the device in docs/user/supported_devices.rst !

docs/user/supported_devices.rst
```diff
@@ -196,6 +196,14 @@ ipq40xx-generic
   - NBG6617
   - WRE6606  [#device-class-tiny]_
 
+ipq40xx-mikrotik
+----------------
+
+* Mikrotik
+
+  - RBSXTsqG-5acD (SXTsq 5 ac)
+  - RBDiscG-5acD (DISC Lite5 ac)
```
AiyionPrime, line 205:
```suggestion
  - DISC Lite5 ac (RBDiscG-5acD)
  - SXTsq 5 ac (RBSXTsqG-5acD)
```

AiyionPrime: Matching the model name again.

https://github.com/openwrt/openwrt/blob/master/target/linux/ipq40xx/image/mikrotik.mk#L60

AiyionPrime (changes requested):

docs/user/supported_devices.rst
AiyionPrime: https://mikrotik.com/product/disc_lite5_ac

AiyionPrime (approved):

PR #2574, opened 2022-07-05
lantiq-xrx200: all updates from swconfig to DSA

blocktrron:
This PR adds the same workaround to allow older swconfig based Gluon versions to update to OpenWrt 22.03+ based DSA images for the lantiq-xrx200 target.

### DIFF
--- a/patches/openwrt/0006-lantiq-xrx200-make-DSA-images-swconfig-upgradable.patch
+++ b/patches/openwrt/0006-lantiq-xrx200-make-DSA-images-swconfig-upgradable.patch
@@ -1,0 +1,30 @@
+From: David Bauer <mail@david-bauer.net>
+Date: Tue, 5 Jul 2022 23:49:31 +0200
+Subject: lantiq-xrx200: make DSA images swconfig upgradable
+
+diff --git a/target/linux/lantiq/image/tp-link.mk b/target/linux/lantiq/image/tp-link.mk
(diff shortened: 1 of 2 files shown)


patches/openwrt/0006-lantiq-xrx200-make-DSA-images-swconfig-upgradable.patch
```diff
@@ -0,0 +1,30 @@
...
+index 756105b94c7f844439132ab57837c9ac87d231ca..5563968de9ab3f491fa89d755fcc6e6fff79948f 100644
+--- a/target/linux/lantiq/image/tp-link.mk
++++ b/target/linux/lantiq/image/tp-link.mk
+@@ -1,7 +1,7 @@
+ DEVICE_VARS += TPLINK_FLASHLAYOUT TPLINK_HWID TPLINK_HWREV TPLINK_HWREVADD TPLINK_HVERSION
+ 
+ define Device/dsa-migration
+-  DEVICE_COMPAT_VERSION := 1.1
++  DEVICE_COMPAT_VERSION := 1.0
+   DEVICE_COMPAT_MESSAGE := Config cannot be migrated from swconfig to DSA
```
rotanid, line 15: aren't these "messages" obsolete then for both mk files?

blocktrron: They are not included to the metadata in case the Compat-version is set to 1.0.

PR #2582, opened 2022-07-10
ath79-generic: migrate wnr2200 and add similar variant

AiyionPrime:
> ath79-generic: add WNR2200-16M
>     
> The CN/RU variant of the device has the same hardware
> as the EU already supported, but contains a 16M flash chip.

> ath79-generic: (re)add WNR2200-8M
>     
> Gone due to
>     
> ar71xx only supported the 8M variant

@ce-4 intends to test the 8M device, the 16M is only different regarding the bigger flash and should not be a problem.

  - [ ] Web interface
  - [ ] TFTP
  - [ ] Other: <specify>
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [ ] Should show activity
  - Switch port LEDs
- Outdoor devices only:


ce-4:
First of all, flashing ath79-factory (ffh) over stock firmware works as expected.
As for the rest:
- OK: Must be flashable from vendor firmware: flash over stock from web interface works
- OK: Must support upgrade mechanism (firstboot erases config properly, sysupgrade -n from vH25 works also)
- OK: Autoupdater name: the lua command returns "netgear-wnr2200-8m"
- OK: Reset/WPS/... button must return device into config mode, yep works + blink mode as well
- OK: Primary MAC address should match address on device label: yep, works (tested on ath79 after factory reset)
- NOT OK; on vH25 the primary mac  is wrong (labeled value $mac+2)
- OK: when upgrading from vH25, $mac+2 is inherited, only after a firstboot, it's $mac again
- OK: Wired network: all 4 lan ports work, wan port works also (vpn)
- OK: Wired network: lan/wan port assignment is correct
- OK: Wireless network: association with ap works
- OK: Wireless network: association with 802.11s mesh works, tested with glinet-mt300n-v2
- OK: Wireless network: ap + mesh work in parallel
- OK: LED mapping: Power/system LED, yep works (lit), including blink sequence in config mode
- OK: LED mapping: Radio LED works (it's only one LED, device has 2.4Ghz but no 5Ghz radio)
- OK: switch port LEDs work, are correctly assigned + blink on activity

AiyionPrime (commented):
looks fine on our end

AiyionPrime:
@ce-4 is done, thanks!

PR #2585, opened 2022-07-14
ath79-generic: (re)add support for TP-Link re355

AiyionPrime:
> Gone due to

@heini66 intends to test this device this weekend

it is remarkable similar to re455

  - [ ] Web interface
  - [ ] TFTP
  - [ ] Other: <specify>
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [ ] Should show activity
  - Switch port LEDs
- Outdoor devices only:


AiyionPrime:
Wrong device, maybe it would be sensible to @maurerle's PR coming in this afternoon, as they are effectively the same device.

maurerle:
Commenting to reference my pull request for later analysis: #2586 
I added the TP-Link RE355 to my pull request.

PR #2594, opened 2022-08-06
gluon-mesh-vpn-fastd: fix respondd segfault under load

NeoRaider:
When running "fastd -v" fails, line may be NULL, causing a segfault in
strncmp.


lemoer (approved):
LGTM.

rotanid:
@NeoRaider can we backport this?
our affected device (rocket m2) won't be part of the initial v2022.1 release unfortunately

PR #2596, opened 2022-08-06
gluon-mesh-batman-adv: add "gateway_tq" field to respondd statistics

lemoer:
This new field reflects the TQ to the selected gateway.

Before this commit, if you had connectivity issues in a larger mesh,
it was a tedious task to understand which nodes are affected and which
are not. By providing this new value for each node, it becomes easier
to see which nodes are affected by the connectivity issues and which
are not.

The new field "gateway_tq" is located at the toplevel of the
statistics resource (next to "gateway" and "gateway_nexthop"):

    gluon-neighbour-info -d ::1 -r statistics
    {
        ...
        "gateway": "02:a1:71:04:09:10",
        "gateway_nexthop": "88:e6:40:20:90:10",
        "gateway_tq": 193,
        ...
    }


AiyionPrime (changes requested):
Commit message contains a mistake:
> The new field "gateway_nexthop"

Should be `The new field "gateway_tq"`

Rest looks correct.

lemoer:
This is tested.

PR #2597, opened 2022-08-06
gluon-mesh-batman-adv: add "routes" to respondd neighbours

lemoer:
This commit adds counters about the number of imported and selected
routes for each neighbour. The following stanza

    "routes": {
        "imported": 171,
        "selected": 120
    },

means, that 171 routes have been learned from this neighbour, but only 120
of them are selected. (For the other 51 routes, the node prefers other
neighbours that also provide routes to the same destination).

These fields are added to the gluon to identify route flaps. This should
make it easier to troubleshoot connectivity problems that only occur when
the route swaps to a weak link.

After the new field "routes" is added, the neighbours response looks like
this:

    {
      "wifi": [
      ],
      "batadv": {
        "72:4c:e2:db:6f:37": {
          "neighbours": {
            "88:e6:40:20:90:10": {
              "routes": {
                "imported": 171,
                "selected": 171
              },
              "best": true,
              "lastseen": 0.95599999999999996,
              "tq": 247
            }
          }
        }
      },
      "node_id": "525400123457"
    }

### DIFF
--- a/package/gluon-mesh-batman-adv/src/respondd-neighbours.c
+++ b/package/gluon-mesh-batman-adv/src/respondd-neighbours.c
@@ -74,6 +74,18 @@ static const enum batadv_nl_attrs parse_orig_list_mandatory[] = {
 	BATADV_ATTR_LAST_SEEN_MSECS,
 };
 
+void increment_json_int_by_key(struct json_object *obj, const char *key) {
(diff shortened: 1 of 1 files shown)


AiyionPrime:
> Ich brauchs nicht.

_@lemoer42_

PR #2607, opened 2022-08-10
gluon-core: increase ath10k peer limit

### LINKED ISSUE #2604: ath10k-ct: STA limit of 32 concurrent connections
The ath10k-ct firmware imposes a lower STA limit than the QCA based flavor, namely 32 vs. 128 connected STAs.

This might be an issue in environments with a lot of concurrently connected clients. The ath10k-ct driver allows configuration of these limits using its fwcfg configuration framework.

The STA limit comes from memory constraints within the wireless system. Thus, we need to reduce the number of concurrent vdevs to increase the number of concurrent stations. This is not much of a problem, as Gluon uses a maximum of 4 VAPs per radio.

See https://www.candelatech.com/ath10k-10.4.php#config

blocktrron:
Increase the peer limit for ath10k-ct from 32 to 96 STAs like it is set
for the non-ct firmware / driver. In order to make this work with the
memory constraints of the wireless platform, reduce the number of
concurrent vdevs to the maximum Gluon uses (4).

Closes #2604

### DIFF
--- a/package/gluon-core/files/etc/hotplug.d/firmware/20-gluon-ath10k-fwcfg
+++ b/package/gluon-core/files/etc/hotplug.d/firmware/20-gluon-ath10k-fwcfg
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+[ -e "/lib/firmware/$FIRMWARE" ] && exit 0
+
+case "$FIRMWARE" in
+ath10k/fwcfg*)
+	cp "/lib/gluon/ath10k-fwcfg.txt" "/lib/firmware/$FIRMWARE"
+	;;
+esac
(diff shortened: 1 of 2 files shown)


mweinelt:
Will this be counted per radio? Because in dual-ath10k instances 4 vdevs is often not enough. Otherwise we should be fine with 3 vdevs (client, owe, mesh) per radio.

blocktrron:
@mweinelt This is per-radio, so on dual-band it is 192 total.

4 vdevs is the minimum, you are missing private-wifi :smile:

mweinelt:
No, I'm not missing that feature at all. You must be mistaken :hankey:

PR #2621, opened 2022-08-23
Include OpenWrt's target default package lists in config generation

### LINKED ISSUE #2619: Missing kmod-ath10k on Netgear R7800
Problem introduced in https://github.com/freifunk-gluon/gluon/commit/15ef885836907b27b2c0481f7ea83d1dd504c5d5.

After install kmod-ath10k and modprobing `ath10k-core` and `ath10k-pci` the phys appear. Running `gluon-reconfigure` followed by a reboot of the device  makes it then work again normally.

NeoRaider:
See the commit message of patch 3 ("scripts: target_config_lib: prepend target default package list from openwrt/tmp/.targetinfo") for an explanation of the issue.

Patch 1 ("scripts: target_config_lib: remove handling for targets without devices") looks much more complex that it is due to unfortunate diffing - it only removes the obsolete `if #lib.devices > 0` check and the corresponding else branch.

I have diffed the generated `.config` for all targets with BROKEN=1 before and after the change (filtering out `CONFIG_GLUON_VERSION lines) with ~~a minimal site~~ the ffda site to ensure that these patches don't break any devices. The resulting diff looks good: [configs.diff.txt](https://github.com/freifunk-gluon/gluon/files/9410756/configs.diff.txt)

Closes #2619

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -181,6 +181,7 @@ config: $(LUA) FORCE
 		$(call CheckSite,$(conf)); \
 	)
 
+	$(OPENWRTMAKE) prepare-tmpinfo
 	$(GLUON_ENV) $(LUA) scripts/target_config.lua > openwrt/.config
 	$(OPENWRTMAKE) defconfig
 	$(GLUON_ENV) $(LUA) scripts/target_config_check.lua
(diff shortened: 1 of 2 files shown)


blocktrron (approved):
Works with ffda site. R7800 does not miss ath10k-ct with this patchset.

PR #2618, opened 2022-08-21
modules: update modules

blocktrron:
Update mainly because of 

97213c7335 mac80211: parse the correct set of HE capabilities for AP mode

PR #2622, opened 2022-08-27
modules: update modules

herbetom:
alternative to #2618 which should pass the CI


rotanid:
this should wait until after the release i think - @blocktrron ?

blocktrron:
```
97213c7335 mac80211: parse the correct set of HE capabilities for AP mode
```

This commit is quite important, as it degrades performance for ax clients.

PR #2626, opened 2022-08-29
ramips-mt76x8: add support for TP-Link RE305 v1

ambassador86:
  - [X] Web interface
        (`lua -e 'print(require("platform_info").get_image_name())'`) = tp-link-re305-v1
- Wired network
  - [X] must have correct port assignment (WAN)
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [X] Should show activity

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -391,6 +391,7 @@ ramips-mt76x8
   - Archer C50 (v3)
   - Archer C50 (v4)
   - RE200 (v2)
+  - RE305 (v1) [#device-class-tiny]
   - TL-MR3020 (v3)
   - TL-MR3420 (v5)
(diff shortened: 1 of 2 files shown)


docs/user/supported_devices.rst
```diff
@@ -451,7 +452,7 @@ Footnotes
 
 .. [#device-class-tiny]
   These devices only support a subset of Gluons capabilities due to flash or memory
-  size constraints. Devices are classified as tiny in they provide less than 7M of usable
+  size constraints. Devices are classified as tiny if they provide less than 7M of usable
```
blocktrron, line 455: Separate commit please. Can be withing this PR.

ambassador86:
@blocktrron here you go. you can amend the commit comment if the naming does not fit

PR #2627, opened 2022-09-01
modules: update to v22.03.0+1

herbetom:
As it happens, v22.03.0 has now been tagged. Even if Gluon releases do not actually follow OpenWRT tags but rather the relase branches, the opportunity would be good to bump the modules to be then on the same level as v22.03.0 for v2022.1.

As far as I have read in our IRC channel, there is a strong tendency towards a Friday release. 
The next Friday is still a few days away. So there would be some time left to validate that everything runs.

I've also opened #2628 which bumps the openwrt modules commit to the currently latest HEAD.


blocktrron:
Closed in favor of #2628

PR #2631, opened 2022-09-07
gluon-core / web: add cellular WWAN support

blocktrron:
Add support for cellular WWAN to Gluon. This allows users to utilize integrated WWAN cards of FWA or travel routers (e.g. Teltonika / GL.iNet).

This works by creating an abstraction layer into Gluon which takes common options (SIM PIN / APN) and translates it to modem-specific settings based on the specific device using. Doing so limits the use-case onto specific models (no LTE sticks possible) but provides a common interface.

The WWAN is assigned the firewall WAN zone and wires WAN can still be used, however without prioritization, so traffic path is not configured to prefer one uplink source or the other.

Screenshot from 2022-09-07 00-19-07

### DIFF
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/250-cellular
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/250-cellular
@@ -1,0 +1,55 @@
+#!/usr/bin/lua
+
+local uci = require('simple-uci').cursor()
+local platform = require 'gluon.platform'
+
+if not platform.is_cellular_device() then
+	return
+end
+
+local function set_or_delete(package, section, option, val)
+	if val ~= nil and string.len(val) ~= 0 then
+		uci:set(package, section, option, val)
+	else
+		uci:delete(package, section, option)
+	end
+end
+
+local function setup_ncm_qmi(devpath, control_type, delay)
+	local pdptype = uci:get('gluon', 'cellular', 'pdptype')
+
+	local pincode = uci:get('gluon', 'cellular', 'pin')
+	local username = uci:get('gluon', 'cellular', 'username')
+	local password = uci:get('gluon', 'cellular', 'password')
+
+	uci:section('network', 'interface', 'cellular', {
+		proto = control_type,
+		device = devpath,
+		disabled = not uci:get_bool('gluon', 'cellular', 'enabled'),
+		pdptype = pdptype,
+		apn = uci:get('gluon', 'cellular', 'apn'),
+	})
+
+	if pdptype ~= 'IP' then
+		uci:set('network', 'cellular', 'ipv6', 'auto')
+	else
+		uci:delete('network', 'cellular', 'ipv6')
+	end
+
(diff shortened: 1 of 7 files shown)


AiyionPrime (changes requested):
I find the code is easy to read and am looking forward to be able to test this in the next days.
For now the first testdevice is still with my roommate,
the second is not yet allowing me to run OpenWrt.

Looks good though, I'd like to have this in gluon.

blocktrron:
@AiyionPrime Can you have a second look here?

AiyionPrime:
The patch for my router is still open, but passed iteration four.
I rebased the OpenWrt part on 22.03, added it as patches in my local tree and added the regular gluon definitions as well as your new is_cellular` references.
After adding gluon-web-cellular` to the site.mk I was able to setup the cellular connection in configmode.

@blocktrron Not sure for how long, as the connected data volume is fairly limited; but the device works flawlessly with your code.
http://aiyion-ws-wn572hp3-4g.n.ffh.zone/cgi-bin/status or if our dns does not work:
http://[2001:678:978:221:823f:5dff:fe3b:7905]/cgi-bin/status

The device is conected via mesh-vpn, without regular wan-uplink.


I think the configmode UI should aim for sane defaults,
maybe apn: `auto`?
and if possible ipv4 or mixed could be pre-selected. Not sure what would work there best.

Last remark for tonight is, the necessity to add `gluon-web-cellular` to the site.mk manually.

The other extensions are usually just feature-decisions for the community to make, (likedo wen want node-roles), but this extension of gluon-web is device specific;
there should be a better way to ensure its on for cellular devices per default, other than adding it to communities site.mk. But that's just my two pence.


I like the code, I like the (basic) functionality, I have not tested username, password or sim-pin.

AiyionPrime:
Just stumbled across an OpenWrt patch from september.
It allows for devices to have certain metric.

Not meant to be reflected in this PR, but a follow-up.
Might be nice to have such a metric option in the UI, to allow fallback to WWAN.

https://patchwork.ozlabs.org/project/openwrt/patch/20220903123129.5713-1-zajec5@gmail.com/

blocktrron:
@AiyionPrime As mentioned yesterday, this is only metadata and does not implicit a route preference.

AiyionPrime:
@blocktrron General ack for this PR from todays gluon meetup.

lemoer (changes requested):
Looks good to me. Good work!

Only one thing: You are trying to add an empty file `package/gluon-web-cellular/gluon-web-mesh-vpn-fastd.pot`. This doesn't seem to make sense to me.

heini66:
thxs 4 the nice work.
for some provider, the auth feature is required to get a connection.
could you be so kind and add an dropdown menu for the auth with:
none, pap, chap or both
txs!!!

lemoer (approved):

AiyionPrime (approved):
Looks good, @lemoer and I looked over it.
Docs are below average though and Heinis suggestion might be a nice follow up.

PR #2645, opened 2022-09-15
[next] enable outdoor mode for ZyXEL NWA55AXE

herbetom:
@blocktrron


herbetom:
has been applied

Djfe:
Why was the ZyXEL NWA55AXE never added to [docs/user/supported_devices.rst](https://github.com/freifunk-gluon/gluon/blob/master/docs/user/supported_devices.rst) on master branch?

herbetom:
I don't quite get why you're asking in this PR. #gluon on hackint.org might be a better place.

Devices marked as broken don't get added to the docs. The issue with the NWA55AXE is that it has no reset Button. #2778 is supposed to help remove the Broken flag.

Djfe:
thank you for your answer and for referring me to to IRC

PR #2648, opened 2022-09-18
xrx200: ensure the correct version for fritz7360

grische:
This is a follow-up of https://github.com/freifunk-gluon/gluon/pull/2544 .

With the new Gluon release, there are dedicated images for 7360v1 and 7360v2 that each has their correct flash layouts. The main difference between v1 and v2 (w.r.t. Gluon) is the size of the flash (and not the flash vendor/model, see Details below).

In order to "fix" existing, and wrongly flashed, devices, I created a script that checks AVM's urlader partition (i.e. bootloader partition) and reads the size of the flash. Depending on the size of the flash, either sets the `model` and `board_name` to that of a [7360v1/sl](https://github.com/openwrt/openwrt/blob/v22.03.0/target/linux/lantiq/files/arch/mips/boot/dts/lantiq/vr9_avm_fritz7360sl.dts#L6-L7) or to that of a [7360v2](https://github.com/openwrt/openwrt/blob/v22.03.0/target/linux/lantiq/files/arch/mips/boot/dts/lantiq/vr9_avm_fritz7360-v2.dts#L6-L7).

This will PR allow smooth migration of wrongly flashed 7360v2 devices to use the correct image.

### Details
During the manufacturing process of a specific AVM router model, vendors and model numbers of elements of the board (like the flash chip in this case) are not identical across identical models with different serial numbers. Only the "metadata", like the exact size of the flash, are identical.
For more details on the different versions, I highly recommend Boxmatrix: https://boxmatrix.info/wiki/FRITZ!Box_Fon_WLAN_7360_Versions

An AVM device has usually 2 places for metadata that persists across firmware upgrades, the urlader partition (i.e. bootloader metadata), and the TFFS partitions. When using a "firmware restore" tool from AVM, the TFFS partition is re-generated from the data in the urlader partition. Hence, the "source of truth" is the urlader partition.

For that, I did some reverse engineering to identify the format and data of the urlader partition: https://github.com/grische/urlader-parser-py

Luckily for the migration, we do not need the full logic and only need to read the first few bytes:
* the urlader version (0x580) to make sure we are in the correct position
* the memsize (0x584) to make sure we were not lucky before and confirm we do have a 7360
* the flashsize (0x588) to read the actual size of the flash

### Background
I have been working with @AiyionPrime on trying to fix the flash layout on some 7360 devices that were unable to read the TFFS partitions. It turned out that 7360v2 devices did not have the correct flash layout with Gluon, as the v2 download was actually a v1 image that used the wrong flash layout.

It turns out that the same problem is with OpenWRT as a lot of 7360 devices were incorrectly flashed with the 7360v1/sl image, instead the 7360v2 image. **The label on the back rarely indicates the correct version.** 
This was "fixed" with OpenWRT by [extending the flashing instructions](https://openwrt.org/toh/avm/fritz.box.wlan.7360?do=diff&rev2%5B0%5D=1652087491&rev2%5B1%5D=1656151163&difftype=sidebyside).

### Tests
I tested this migration on a local 7360v2 device.

EDIT:
For reference (and easier readability), that's the actual OpenWRT commit: https://github.com/grische/openwrt/commit/fa3af8c6583b5338499f10aec926f49d399e84ee

### DIFF
--- a/patches/openwrt/0009-xrx200-migrate-fritz7360-v2-using-incorrect-image.patch
+++ b/patches/openwrt/0009-xrx200-migrate-fritz7360-v2-using-incorrect-image.patch
@@ -1,0 +1,81 @@
+From: Grische <github@grische.xyz>
+Date: Sun, 18 Sep 2022 14:03:16 +0200
+Subject: xrx200: migrate fritz7360-v2 using incorrect image
+
+Migrate AVM FRITZ!Box 7360 v2 boards flashed with the incorrect v1 image to use
+the newly added v2 target image during the next upgrade.
+Using the v2 target image allows the boards to read the TFFS partition, which
+is misaligned when using the v1 image.
+
+Ref: https://github.com/freifunk-gluon/gluon/pull/2648
+
+Co-authored-by: Jan-Niklas Burfeind <git@aiyionpri.me>
+
+diff --git a/target/linux/lantiq/xrx200/base-files/lib/preinit/01_sysinfo.sh b/target/linux/lantiq/xrx200/base-files/lib/preinit/01_sysinfo.sh
+new file mode 100644
+index 0000000000000000000000000000000000000000..fab50d708e872f819c643cea79327e4f438de524
+--- /dev/null
++++ b/target/linux/lantiq/xrx200/base-files/lib/preinit/01_sysinfo.sh
+@@ -0,0 +1,62 @@
++set_sysinfo_xrx200_for_fritz7360_model() {
++    local board_name=$1
++    local model
++    local urlader_version urlader_memsize urlader_flashsize
++    local hexdump_format='4/1 "%02x""\n"'
++
++    # Values are based on urlader-parser-py
++    # https://github.com/grische/urlader-parser-py/blob/42970bf8dec7962317df4ff734c57ebf36df8905/parser.py#L77-L84
++    urlader_version="$(dd if=/dev/mtd0ro bs=1 skip=$((0x580+0x0)) count=4 | hexdump -e "${hexdump_format}")"
++    if [ "${urlader_version}" != "00000003" ]; then
++        logger -s -p warn -t sysinfo-xrx200 "unexpected urlader version found: ${urlader_version}"
++        return
++    fi
++
++    urlader_memsize="$(dd if=/dev/mtd0ro bs=1 skip=$((0x580+0x4)) count=4 | hexdump -e "${hexdump_format}")"
++    if [ "${urlader_memsize}" != "08000000" ]; then
++        logger -s -p warn -t sysinfo-xrx200 "unexpected memsize found: ${urlader_memsize}"
++        return
++    fi
++
++    urlader_flashsize="$(dd if=/dev/mtd0ro bs=1 skip=$((0x580+0x8)) count=4 | hexdump -e "${hexdump_format}")"
++    case "${urlader_flashsize}" in
++        "02000000")    # 32MB
++            # see vr9_avm_fritz7360-v2.dts
++            board_name="avm,fritz7360-v2"
++            model="AVM FRITZ!Box 7360 V2"
++            ;;
++        "01000000")    # 16MB
++            return
++            ;;
++        *)
++            logger -s -p warn -t sysinfo-xrx200 "unexpected flashsize found: ${urlader_flashsize}"
++            return
++            ;;
++    esac
++
++    logger -s -p notice -t sysinfo-xrx200 "detected ${board_name} from urlader partition /dev/mtd0ro. Enforcing model ${model}."
++    echo "${board_name}" > /tmp/sysinfo/board_name
++    echo "${model}" > /tmp/sysinfo/model
++}
++
++do_sysinfo_xrx200() {
++    local reported_board board_name model
++
++    [ -d /proc/device-tree ] || return
++    reported_board="$(strings /proc/device-tree/compatible | head -1)"
++
++    mkdir -p /tmp/sysinfo
++    # 7360 is notoriously known for not writing "v2" on their labels and many
++    # routers have flashed the wrong firmware with the wrong flash layout.
++    # We ensure that the underlying hardware is reported correctly, so that
++    # future upgrades will use the correct flash layout.
++    # Using 7360v2 hardware, an upgrade from a 7360v1/sl firmware to a 7360v2
++    # is working.
++    case "${reported_board}" in
++        avm,fritz7360sl)
++            set_sysinfo_xrx200_for_fritz7360_model "${reported_board}"
++            ;;
++    esac
++}
++
++boot_hook_add preinit_main do_sysinfo_xrx200


grische:
If there is no interest in fixing old Gluon devices, please let me know.

cc @blocktrron

patches/openwrt/0007-xrx200-ensure-the-correct-version-for-fritz7360.patch
```diff
@@ -0,0 +1,94 @@
...
++    # We ensure that the underlying hardware is reported correctly, so that
++    # future upgrades will use the correct flash layout.
++    # Using 7360v2 hardware, an upgrade from a 7360v1/sl firmware to a 7360v2
++    # is working.
++    case "${reported_board}" in
++        *7360*)
++            do_sysinfo_xrx200_for_fritz7360_model "${reported_board}" "${reported_model}"
++            ;;
++        *)
++            set_sysinfo_xrx200 "${reported_board}" "${reported_model}"
```
blocktrron, line 86: Is this necessary? Read, wouldn't a simple `return` achieve the same result?

patches/openwrt/0007-xrx200-ensure-the-correct-version-for-fritz7360.patch
```diff
@@ -0,0 +1,94 @@
...
++    local board_name=$1
++    local model=$2
++    local urlader_version urlader_memsize urlader_flashsize
++    local hexdump_format='4/1 "%02x""\n"'
++
++    # Values are based on urlader-parser-py
++    # https://github.com/grische/urlader-parser-py/blob/42970bf8dec7962317df4ff734c57ebf36df8905/parser.py#L77-L84
++    urlader_version="$(dd if=/dev/mtd0ro bs=1 skip=$((0x580+0x0)) count=4 | hexdump -e "${hexdump_format}")"
++    if [ "${urlader_version}" != "00000003" ]; then
++        >&2 echo "Warning: unexpected urlader version found: ${urlader_version}"
```
blocktrron, line 35: Use the OpenWrt logging framework: https://openwrt.org/docs/guide-user/base-system/log.essentials

patches/openwrt/0007-xrx200-ensure-the-correct-version-for-fritz7360.patch
```diff
@@ -0,0 +1,94 @@
...
++        set_sysinfo_xrx200 "${board_name}" "${model}"
++        return
++    fi
++
++    urlader_memsize="$(dd if=/dev/mtd0ro bs=1 skip=$((0x580+0x4)) count=4 | hexdump -e "${hexdump_format}")"
++    if [ "${urlader_memsize}" != "08000000" ]; then
++        >&2 echo "Warning: unexpected memsize found: ${urlader_memsize}"
```
blocktrron, line 42: Same here

patches/openwrt/0007-xrx200-ensure-the-correct-version-for-fritz7360.patch
```diff
@@ -0,0 +1,94 @@
...
++            board_name="avm,fritz7360-v2"
++            model="AVM FRITZ!Box 7360 V2"
++            ;;
++        "01000000")    # 16MB
++            # see vr9_avm_fritz7360sl.dts
++            board_name="avm,fritz7360sl"
++            model="AVM FRITZ!Box 7360 SL"
++            ;;
++        *)
++            >&2 echo "Warning: unexpected flashsize found: ${urlader_flashsize}"
```
blocktrron, line 60: Same here

patches/openwrt/0007-xrx200-ensure-the-correct-version-for-fritz7360.patch
```diff
@@ -0,0 +1,94 @@
+From fa3af8c6583b5338499f10aec926f49d399e84ee Mon Sep 17 00:00:00 2001
+From: Grische <github@grische.xyz>
+Date: Sun, 18 Sep 2022 14:03:16 +0200
+Subject: [PATCH] xrx200: ensure the correct version for fritz7360
+
```
blocktrron, line 5: Please add a Link to this PR in the patch description, so it is possible to correlate it later down the road.

blocktrron:
Do i understand correctly that the goal is to migrate existing devices using the autoupdater? In this case this patch would only be applied temporarily and dropped in 1 - 2 Gluon releases?

AiyionPrime:
> Do i understand correctly that the goal is to migrate existing devices using the autoupdater? In this case this patch would only be applied temporarily and dropped in 1 - 2 Gluon releases?

That's correct. Currently there are a few hundred devices which only make use of half their memory, as they got flashed with v1 firmware; as it was falsely assumed all revisions had 8M.

This is basically a migration script that can be gone in a year or two, once the devices flipped over to their correct image introduced in #2544.

blocktrron:
Are there any consequences by not reading the TFFS partitions currently? If i read the code for these devices correctly, they do not depend on the TFFS presence for reading MAC addresses.

grische:
Depends on the community. The TFFS does contain metadata for the device (much more easily accessible than the urlader partition).

There is an OpenWRT package that can read that data:
 https://openwrt.org/packages/pkgdata/fritz-tffs

blocktrron:
> Depends on the community.

Can you elaborate on that? MAC addresses are usually set-up in OpenWrt, and i see no usage of the fritz-tffs package here.

grische:
Sure.

If a community wants to use fritz-tffs to export some metadata (e.g. via a custom script) this will only work on properly flashed 7360v1 and will fail (or create an obscure error) when used on a 7360v2 with the V1 Image installed.

In my case, I tried reading ProductID to potentially improve statistics of actually hardware that is in use (7520 and 7530 devices cannot be distinguished otherwise, similar for others) and ran it on AVM devices available to me. That's how I encountered the issue with the 7360v2 initially.

With the exact ProductIDs, I would be able to deduce more detailed hardware capabilities (assuming devices are unmodified - which most seem to be).

This metadata is in general of no concern to us right now, except for the flash and memory size. So this is more of a theoretical scenario.

patches/openwrt/0007-xrx200-ensure-the-correct-version-for-fritz7360.patch
grische: I wanted to explicitly set it ourselves, to be independent of potential OpenWRT changes.

If you think that falling back to the [upstream OpenWRT detection](https://github.com/openwrt/openwrt/blob/openwrt-22.03/package/base-files/files/lib/preinit/02_sysinfo) is a better, fine with me.

Let me know what you prefer.

patches/openwrt/0007-xrx200-ensure-the-correct-version-for-fritz7360.patch
grische: Thanks, I did not know this one. Should be fixed now.

patches/openwrt/0007-xrx200-ensure-the-correct-version-for-fritz7360.patch
grische: Fixed

patches/openwrt/0007-xrx200-ensure-the-correct-version-for-fritz7360.patch
grische: Fixed

patches/openwrt/0007-xrx200-ensure-the-correct-version-for-fritz7360.patch
grische: Added

patches/openwrt/0007-xrx200-ensure-the-correct-version-for-fritz7360.patch
blocktrron: Please use the upstream one

grische: Done

grische:
By falling back to OpenWRT's default naming detection, the code got a lot simpler. @blocktrron that should address all your comments.

It is worth mentioning that I cannot see the log messages, using `logread`:
```
# logread -e ".*xrx200.*"
Tue Dec 22 04:38:38 2020 kern.info kernel: [    0.276860] dcdc-xrx200 1f106a00.dcdc: Core Voltage : 1016 mV
Tue Dec 22 04:38:38 2020 kern.err kernel: [    0.304358] pcie-xrx200 1d900000.pcie: failed to get the PCIe PHY
Tue Dec 22 04:38:38 2020 kern.err kernel: [    1.156991] pcie-xrx200 1d900000.pcie: failed to get the PCIe PHY
```

blocktrron:
Now I got it. I was still under the impression the misaligned TFFS impedes Gluons core functionality, but if it does not we don't need to backport this.

grische:
> we don't need to backport this.

@blocktrron Do you mean, you do not want to backport the patch to the v2022.1.x branch? Would you want to have a PR against master instead? Or abandon it completely?

Could you also tell me why the "Check patches" action fails? I do not think I understand the goal of it. Should `git status; git diff` be empty in the step before?
https://github.com/freifunk-gluon/gluon/actions/runs/3189230111/jobs/5205800618

herbetom:
@grische all PRs should target master expect it's an backport or some really special circumstance.

grische:
@herbetom I assumed the circumstances here are special

blocktrron:
Yes, please target master, not a release branch.

You can execute `make refresh-patches` this should automatically fix your patch.

grische:
@blocktrron I changed the base to master and ran `make refresh-patches GLUON_SITEDIR="contrib/ci/minimal-site"` to clean up the patch.

patches/openwrt/0008-xrx200-ensure-the-correct-version-for-fritz7360.patch
```diff
@@ -0,0 +1,80 @@
...
++    # routers have flashed the wrong firmware with the wrong flash layout.
++    # We ensure that the underlying hardware is reported correctly, so that
++    # future upgrades will use the correct flash layout.
++    # Using 7360v2 hardware, an upgrade from a 7360v1/sl firmware to a 7360v2
++    # is working.
++    case "${reported_board}" in
++        *7360*)
++            set_sysinfo_xrx200_for_fritz7360_model "${reported_board}"
++            ;;
++        *)
```
blocktrron, line 73: Drop the whole case.

patches/openwrt/0008-xrx200-ensure-the-correct-version-for-fritz7360.patch
```diff
@@ -0,0 +1,80 @@
...
++
++    mkdir -p /tmp/sysinfo
++    # 7360 is notoriously known for not writing "v2" on their labels and many
++    # routers have flashed the wrong firmware with the wrong flash layout.
++    # We ensure that the underlying hardware is reported correctly, so that
++    # future upgrades will use the correct flash layout.
++    # Using 7360v2 hardware, an upgrade from a 7360v1/sl firmware to a 7360v2
++    # is working.
++    case "${reported_board}" in
++        *7360*)
```
blocktrron, line 70: As this suggests it is a one-way migration, we can limit the migration source by using the full compatible:

```
avm,fritz7360sl
```

patches/openwrt/0008-xrx200-ensure-the-correct-version-for-fritz7360.patch
```diff
@@ -0,0 +1,80 @@
...
++    fi
++
++    urlader_flashsize="$(dd if=/dev/mtd0ro bs=1 skip=$((0x580+0x8)) count=4 | hexdump -e "${hexdump_format}")"
++     case "${urlader_flashsize}" in
++        "02000000")    # 32MB
++            # see vr9_avm_fritz7360-v2.dts
++            board_name="avm,fritz7360-v2"
++            model="AVM FRITZ!Box 7360 V2"
++            ;;
++        "01000000")    # 16MB
```
blocktrron, line 40: We can also drop this block in this case

blocktrron (commented):
Looks good overall, but as we should drop this downstream patch in the future, I'd keep the character of a one-way one-time migration. See the comments for that.

Also, the commit message can be improved. My proposal:

```
lantiq-xrx200: migrate incorrectly flashed FRITZ!Box 7360 v2

Migrate AVM FRITZ!Box 7360 v2 boards  flashed with the incorrect v1 image to use the v2
target image. This allows the boards to read the TFFS partition, which is misaligned
with the v1 image.

Ref: #2544
```

grische:
That's a surprisingly long list of changes suddenly. I will try to implement, build and test them as soon as I have time available.

patches/openwrt/0008-xrx200-ensure-the-correct-version-for-fritz7360.patch
blocktrron: For clarification: We only want to migrate v2 with v1 image to v2 with v2 image. The other way around is not desired. So detecting a v2 and potentially overwriting the upstream model string is not in our interest.

blocktrron:
I've explained the intentions more if that helps.

patches/openwrt/0008-xrx200-ensure-the-correct-version-for-fritz7360.patch
AiyionPrime: resolved in aa5065c

patches/openwrt/0008-xrx200-ensure-the-correct-version-for-fritz7360.patch
AiyionPrime: resolved in aa5065c

patches/openwrt/0008-xrx200-ensure-the-correct-version-for-fritz7360.patch
AiyionPrime: resolved in aa5065c

AiyionPrime:
The test device @Dark4MD lent me is a few hundred km away though, so this fixup is untested.

Upon squashing we can fix the commit message.

@blocktrron: mind to take another look though?

AiyionPrime:
Just tested this on a v2 and an older device, v2 was marked as such, the other was left untouched.

https://hannover.freifunk.net/karte/#/en/map/e0286d100812
https://hannover.freifunk.net/karte/#/en/map/2465119311ab

@blocktrron anything left from your side?

Djfe:
@AiyionPrime backported needed

maurerle:
This patch is backported in ffmuc and ffac site config in the current v2022.1.x branch, it would be cool if it gets backported officially to remove the patch from the site config

rotanid:
feel free to open a pull request against the gluon v2022.1.x branch.
this can't be done automatically as it contains a patch which has to match exactly.

PR #2651, opened 2022-09-20
ramips-mt7621: add Ubiquiti UniFi Switch Flex

herbetom:
Currently [`poemgr`](https://github.com/blocktrron/poemgr) is included but not enabled. I kinda  think it would be a sensible default to enable it as well as poe on all ports. Or if there would be a Config Mode Page to enable those ports.

  - [ ] Web interface
  - [ ] TFTP
  - [ ] Other: <specify>
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [ ] Should show activity
  - Switch port LEDs
- Outdoor devices only:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -355,6 +355,7 @@ ramips-mt7621
   - EdgeRouter X
   - EdgeRouter X-SFP
   - UniFi 6 Lite
+  - UniFi Switch Flex
 
 * ZBT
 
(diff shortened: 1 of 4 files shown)


blocktrron:
I think we can reopen that as soon as the checklist has been finished.

Djfe:
If anyone ever decides to implement this switch officially for Gluon, they might want to look into replacing uboot with a variant that supports both cpu cores. If it actually uses an MT7621AT, then Ubiquiti (accidentally?) limited it to one core (Like the MT7621ST) by building uboot without dual core support (`RALINK_DUAL_CORE_FUN=y`):
https://twitter.com/FreifunkDA/status/1436778889297207299/photo/1
https://forum.openwrt.org/t/how-to-enable-hyperthreading-on-mediatek-mt7621at/78419
This is relevant for VPN Performance.

Although the current master branch patch is what really makes the MT7621 shine:
https://forum.openwrt.org/t/2-gbps-wan-lan-nat-routing-on-ramips-mt7621-devices/131478

PR #2653, opened 2022-09-21
draft: ath79-generic: add support for TP-Link EAP225-Outdoor v3

yksflip:
This backports the TP-Link EAP225-Outdoor v3. Does it need to be tested again (merged to next in #2643)? If anyone is faster than me, feel free to help out!

  - [ ] Web interface
  - [ ] TFTP
  - [ ] Other: <specify>
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [ ] Should show activity
  - Switch port LEDs
- Outdoor devices only:

Signed-off-by: Philipp Rothmann <philipprothmann@posteo.de>

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -104,7 +104,7 @@ ath79-generic
   - CPE510 (v2.0)
   - CPE510 (v3.0)
   - CPE710 (v1.0)
-  - EAP225-Outdoor (v1)
+  - EAP225-Outdoor (v1, v3)
   - RE450 (v1)
   - TL-WDR3500 (v1)
   - TL-WDR3600 (v1)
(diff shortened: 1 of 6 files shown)


herbetom:
As far as i remember the device support commits for the v3 weren't the only ones needed for supporting the v3 and there was a reason why the v3 wasn't alreafy backported to openwrt-22.03.

I think it had something to do with bumping the [firmware-utils in OpenWrt](https://github.com/openwrt/openwrt/blob/openwrt-22.03/tools/firmware-utils/Makefile) which would also affect other devices etc.

Sorry that I do not have better news. I had looked this up at some point but then discarded the idea of getting support in 22.03.

What would be required to get support for the eap225-outdoor v3 is in my eyes:
- create a backport PR for OpenWrt which includes everything to get it working
- once/if that is merged, check if everything in Gluon works and fill out the checklist
- update this PR. It should point to master, not v2022.1.x, and if it's merged you can request a backport to v2022.1.x

ambassador86:
I also took a look into the changes needed to get this device working in openwrt 22.03 and came to the same conclousion as @herbetom.

rotanid:
please also see the rules for device integration: https://github.com/freifunk-gluon/gluon/wiki/Device-Integration

blocktrron:
No activity and targets the wrong branch (Device support has to first land in master). Feel free to reopen with a tested PR targeting the master branch.

PR #2664, opened 2022-10-11
ath79-generic: Fix autoupdater for Loco M XW from v2021.1.x to v2022.1.x

### LINKED ISSUE #2663: v2022.1: Autoupdater from old release does not work for Loco M XW
We are currently rolling out our first release based on v2022.1, and I noticed that Loco M XW is not updated, although a firmware image for the device exists. Reason seems to be a rename without proper alias in v2022.1:

In v2021.1.x, targets/ar71xx-generic:
```
device('ubiquiti-loco-m-xw', 'ubnt-loco-m-xw', {
	aliases = {
		'ubiquiti-nanostation-loco-m2-xw',
		'ubiquiti-nanostation-loco-m5-xw',
		-- 'ubiquiti-nanobeam-m5', -- untested
	},
})
```

In v2022.1.x, targets/ath79-generic
```
device('ubiquiti-nanostation-loco-m-xw', 'ubnt_nanostation-loco-m-xw', {
	manifest_aliases = {
		'ubiquiti-nanostation-loco-m2-xw', -- upgrade from OpenWrt 19.07
		'ubiquiti-nanostation-loco-m5-xw', -- upgrade from OpenWrt 19.07
	},
})
```

As far I understand the old name from v2021.1.x must be included as an alias that the firmware can be found by the autoupdater, like this:
```
device('ubiquiti-nanostation-loco-m-xw', 'ubnt_nanostation-loco-m-xw', {
	manifest_aliases = {
		'ubiquiti-loco-m-xw', -- upgrade from OpenWrt 19.07
		'ubiquiti-nanostation-loco-m2-xw', -- firmware selector
		'ubiquiti-nanostation-loco-m5-xw', -- firmware selector
	},
})
```

goligo:
Ubiquiti Nanostation Loco M XW was renamed in v2022.1.x, but the alias to the old name was missing, so devices running the old release did not update.

Fixes #2663

### DIFF
--- a/targets/ath79-generic
+++ b/targets/ath79-generic
@@ -511,6 +511,7 @@ device('tp-link-wbs510-v1', 'tplink_wbs510-v1', {
 
 device('ubiquiti-nanostation-loco-m-xw', 'ubnt_nanostation-loco-m-xw', {
 	manifest_aliases = {
+		'ubiquiti-loco-m-xw', -- upgrade from OpenWrt 19.07
 		'ubiquiti-nanostation-loco-m2-xw', -- upgrade from OpenWrt 19.07
 		'ubiquiti-nanostation-loco-m5-xw', -- upgrade from OpenWrt 19.07
 	},


herbetom:
I think convention would be to prefix your commit message with `ath79-generic: ` like you did in the title.

rotanid:
i doesn't look right to me to have *three* aliases to update from OpenWrt 19.07
my guess(!) is, that some of them are not necessary.

goligo:
As I understand, the two more explicit ones have been introduced for the firmware selector.

herbetom:
As discussed at the meetup the alias are only for the manifest and not symlinks for the firmware selector.
Instead it seems to be so that there were in the past two diffrent firmwares which ran on those devices and it's possible that devices will have two diffrent identifiers depending on which firmware was originally used.

PR #2671, opened 2022-10-14
mpc85xx-p1010: re-enable TP-Link WDR4900

### LINKED ISSUE #2491: TP-Link WDR4900 does not boot with OpenWrt 22.03
The TP-Link WDR4900 does not boot with a 22.03 based Gluon due to the the kernel exceeding the max bootable size.

@NeoRaider is working on a SPI loader to allow larger kernels to boot. This issue is for tracking said process after the merge of #2426

blocktrron:
Upstream added a standalone SPI kernel-loader which fixes the unbootable
image for the WDR4900. Thus, we can re-introduce this device to Gluon.

closes #2491


herbetom:
closes #2491

(Won't work this way, should be added to the initial comment.)

PR #2681, opened 2022-10-18
ipq40xx-generic: add AVM FRITZBox 7520 explicitely

grische:
AVM Fritz!Box 7520 and Fritz!Box 7530 use the same hardware platform and can only be distinguished by using the urlader partition or the fritz-tffs tools and read the ProductID (Fritz_Box_HW247).

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -180,6 +180,7 @@ ipq40xx-generic
 * AVM
 
   - FRITZ!Box 4040 [#avmflash]_
+  - FRITZ!Box 7520 (v1) [#eva_ramboot]_
   - FRITZ!Box 7530 [#eva_ramboot]_
   - FRITZ!Repeater 1200 [#eva_ramboot]_
 
--- a/targets/ipq40xx-generic
+++ b/targets/ipq40xx-generic
@@ -36,6 +36,7 @@ device('avm-fritz-box-4040', 'avm_fritzbox-4040', {
 
 device('avm-fritz-box-7530', 'avm_fritzbox-7530', {
 	factory = false,
+	aliases = {'avm-fritz-box-7520'},
 })
 
 device('avm-fritz-repeater-1200', 'avm_fritzrepeater-1200', {


blocktrron:
Add the device to the docs. Have you checked if it also works with the v2 version of the 7520? Otherwise we should clarify this in the docs.

grische:
@blocktrron Added to [docs/user/supported_devices.rst](https://github.com/freifunk-gluon/gluon/compare/bf94e85f056ee6bd30e3e85f94b693d9c0201e0c..053ac0d1f246428c3176f924ffa65b4c6e501cc6#diff-569125591014883ee4f7ab0a04f63c29743d7eacdabf8cf031c4d9e73bf03e3d).

> Have you checked if it also works with the v2 version of the 7520? Otherwise we should clarify this in the docs.

I do not have access to a 7520v2, but it should be compatible with Gluon as Gluon does not use the VDSL modem:

> The 7520v2 is based on the same quad core [Dakota](https://boxmatrix.info/wiki/Property:Dakota) as the [7520v1](https://boxmatrix.info/wiki/7520v1), so it shares the same [WLAN](https://boxmatrix.info/w/index.php?title=WLAN&action=edit&redlink=1), networking and [USB](https://boxmatrix.info/w/index.php?title=USB&action=edit&redlink=1) capabilities, but this is the first box using a [Realtek](https://boxmatrix.info/w/index.php?title=Realtek&action=edit&redlink=1) [RTL86](https://boxmatrix.info/wiki/Property:RTL86) [VDSL](https://boxmatrix.info/w/index.php?title=VDSL&action=edit&redlink=1) modem frontend.

Src: https://boxmatrix.info/wiki/FRITZ!Box_7520_v2

blocktrron:
@grische I suspect they've had to do at least some modifications to the base board, which in turn might have changed the GPIO assignment in a way that is incopatbile with the v1. In case you did not verify with one or diffed the vendor DTS, i would refrain from assuming implicit compatibility.

grische:
>  In case you did not verify with one or diffed the vendor DTS, i would refrain from assuming implicit compatibility.

I did not mean to say that I think it is compatible.

@blocktrron Are there any blockers from getting this PR approved and merged?

docs/user/supported_devices.rst
```diff
@@ -180,6 +180,7 @@ ipq40xx-generic
 * AVM
 
   - FRITZ!Box 4040 [#avmflash]_
+  - FRITZ!Box 7520 [#eva_ramboot]_
```
blocktrron, line 183: See my first Comment in this PR - add a `v1 / VRX518` suffix here.

grische: added

Issue #2672, opened 2022-10-14
bridge MLD wakeup call feature missing from stable branch


rotanid:
the "bridge MLD wakeup call feature" which was introduced by @T-X with PR #2076 and commit b2add48d2a7fa86c05eff3a2e6935168d60b2bf8 was removed when switching the OpenWrt Base to 21.02 in commit aab2b914b8687fa91849998e27ed6b9cf27688db .
It was reintroduced in Gluon's master branch with #2365 commit 0530b06557ca2a7141e76d9f3c490f97383ae83c
Therefore, this feature was part of a Gluon stable release but is not part of the current stable release branch 2022.2.x as of now, which we also state in the "known issues" part of the release notes.
I suggest we backport it to fix this regression.

blocktrron:
It was removed because the patch did not apply and was not re-applied to v2022.1 because it was not applied to master back when we prepared the release.

rotanid:
> It was removed because the patch did not apply and was not re-applied to v2022.1 because it was not applied to master back when we prepared the release.

sure, i rephrased my issue to better suit the circumstances.

rotanid:
#2685

rotanid:
fixed by #2685 with next minor release

PR #2682, opened 2022-10-18
docs: update v2022.1.x release notes known issues section

rotanid:
add link to issue #2672 in known issues section of the release notes


AiyionPrime (approved):
I don't think adding the link to the description would hurt.
The rst format appears to be fine as well.

Is there a reason why we should not backport this to `2022.1.x`?

PR #2684, opened 2022-10-22
[RFC] docs: update multidomain-site example

grische:
I tried expanding and and updating the multidomain-site documentation, based on docs/site-example/site.conf

Notable changes:
* moved `mesh.batman_adv` from the domain-specific configuration to site.conf
* moved the wifi channels from the domain-specific configuration to site.conf (opinions?)
* moved `mesh_vpn.fastd` from site.conf to the domain-specific configuration (opinions?)
* added new wifi owe examples to the domain-specific configuration
* replaced dummy parameter values with the ones used in [docs/site-example/site.conf](https://github.com/freifunk-gluon/gluon/blob/v2022.1.1/docs/site-example/site.conf)

### DIFF
--- a/docs/multidomain-site-example/domains/alpha_centauri.conf
+++ b/docs/multidomain-site-example/domains/alpha_centauri.conf
@@ -1,3 +1,10 @@
+-- This is an example domain configuration for Gluon v2022.1
+--
+-- Take a look at the documentation located at
+-- https://gluon.readthedocs.io/ for details.
+--
+-- This configuration will not work as is. You're required to make
+-- community specific changes to it!
 {
   -- multiple codes/names can be defined, the first one is the primary name
   -- additional aliases can be defined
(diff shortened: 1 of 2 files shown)


lemoer:
The improvements to the multidomain-site are well appreciated! :) 

Please split this into multiple commits and explain for each commit, why you propose the changes. Unfortunately, like this, it's kind of hard/impossible to review.

grische:
@lemoer is this really necessary? It is just a cleanup or sync between both files to illustrate how to actually use it.

AiyionPrime:
If you want comments on your changes, it is I'm afraid :)

grische:
Feel free to reopen if this changes.

AiyionPrime:
In our experience reviewing a few smaller descriptive commits leads to faster and better results in terms of finishing the PR for both the project and you.
Furthermore it makes bisecting problems or git-blaming easier - for others coming along later - as well.

You already identified those five core-changes in the list in the top-post.
As soon as they are separate commits, we can go on reviewing this.

If you've got trouble with git we can help you doing so; but committing "docs: update $file" and changing 50% of it won't happen, sorry :/

We'd appreciate if you'd continue this PR instead of throwing the work you put in it away.

grische:
I don't have the time to split up this commit and I don't see any benefit of splitting up a commit that just updates documentation. How often do you need to bisect documentation files?

These file are touched only every few years, so I don't see an urgency to put much effort into this right now. If you think it's important, feel free to take the work and improve upon it.

PR #2688, opened 2022-10-30
Config-Mode UI for Interface Role Assignment (v2)

lemoer:
This PR is a second attempt to bring flexible role assignment for network interfaces into the config mode UI. 

Implemented Features:
- The roles "Uplink", "Mesh" and "Client" can be assigned to interfaces.
- Roles can also be assigned to additional interfaces (beyond lan/wan/single).
- The roles "Uplink"/"Mesh" and "Client" are mutually exclusive.

Current Limitations:
- As of now, additional interfaces (eth, vlan, ...) have to be created manually from CLI.
- Assigning different roles to individual LAN/WAN ports is also only possible from CLI.

Once this PR is merged, follow-up PRs will remove the limitations and extend the functionality.

This is an example how the UI look with this PR:

image

Additional interfaces (beyond lan/wan/single) are handled correctly:

image

Some references:
- This PR builds upon the flexible role assignment from (#2372).
- This PR picks up some the wishes from (#2393).
- The is a sucessor of the PR (#2480), which was abandonned since it introduced to many features at once.
- The functionality will be expanded in future PRs.


lemoer:
(This PR is ready for review.)

mkg20001:
This is being used successfully in the grazer mesh

lemoer:
So far, the mesh on lan checkbox adds the ``  mesh``   role additionally to the ` client`   role.

This means that when mesh on lan was previously enabled, the config looks like this:
````
config interface 'iface_lan'
	option name '/lan'
	list role 'client'
	list role 'mesh'
```

When you open this config in the UI, it looks like this:

![image](https://user-images.githubusercontent.com/601153/207439256-eb41eb8c-028e-4320-97e0-49aef142bee0.png)

After pressing save, the config looks like this:
` ``` `` 
config interface 'iface_lan'
	option name '/lan'
	list role 'client'
````

**Comment by lemoer** [comment]
TODO: Add an upgrade script.

**Comment by lemoer** [comment]
Rebased.

**Comment by AiyionPrime** [comment]
Are we not still using the client interface to determine on which port to deliver the config-mode?

This allows users to get rid of all client interfaces, not the way, where its still there and just overridden by another role, but gone completely, no?

That way this PR would enable users to lock themselves out of their router for good?

**Comment by AiyionPrime** [comment]
```console
package/gluon-web-model/files/lib/gluon/web/view/model/mlvalue.html:
	Wrong line endings or no final newline
	7: Wrong indentation type(spaces instead of tabs)
	8: Wrong indentation type(spaces instead of tabs)
	9: Wrong indentation type(spaces instead of tabs)
	10: Wrong indentation type(spaces instead of tabs)
	11: Wrong indentation type(spaces instead of tabs)
	12: Wrong indentation type(spaces instead of tabs)
	13: Wrong indentation type(spaces instead of tabs)
	14: Wrong indentation type(spaces instead of tabs)
	15: Wrong indentation type(spaces instead of tabs)
	16: Wrong indentation type(spaces instead of tabs)
	17: Wrong indentation type(spaces instead of tabs)
	18: Wrong indentation type(spaces instead of tabs)
	19: Wrong indentation type(spaces instead of tabs)
	20: Wrong indentation type(spaces instead of tabs)
	21: Wrong indentation type(spaces instead of tabs)
	22: Wrong indentation type(spaces instead of tabs)
	23: Wrong indentation type(spaces instead of tabs)
	24: Wrong indentation type(spaces instead of tabs)

19 errors found

```

package/gluon-web-model/files/lib/gluon/web/view/model/mlvalue.html
```diff
@@ -0,0 +1,25 @@
...
+                attr("value", entry.key) ..
+                attr("checked", (util.contains(self:cfgvalue(), entry.key)) and "checked") ..
+                attr("data-exclusive-with", self.exclusions[entry.key]) ..
+                attr("data-update", "change")
+            %>>
+            <label<%= attr("for", id.."."..entry.key)%>></label>
+            <span class="gluon-multi-list-option-descr"><%|entry.value%></span>
+        </label>
+        <% if i ~= #entries then write(br) end %>
+    <% end %>
```
AiyionPrime, line 24:
```suggestion
	<% for i, entry in pairs(entries) do %>
		<label<%=
			attr("data-index", i) ..
			attr("data-depends", self:deplist(entry.deps))
		%>>
			<input data-update="click change" type="checkbox"<%=
				attr("id", id.."."..entry.key) ..
				attr("name", id) ..
				attr("value", entry.key) ..
				attr("checked", (util.contains(self:cfgvalue(), entry.key)) and "checked") ..
				attr("data-exclusive-with", self.exclusions[entry.key]) ..
				attr("data-update", "change")
			%>>
			<label<%= attr("for", id.."."..entry.key)%>></label>
			<span class="gluon-multi-list-option-descr"><%|entry.value%></span>
		</label>
		<% if i ~= #entries then write(br) end %>
	<% end %>
```

AiyionPrime: html files should be indented by tabs and not spaces.

AiyionPrime (changes requested):
Not a complete review, but only the results of me checking for affected PRs in #2765.

package/gluon-web-network/luasrc/lib/gluon/config-mode/model/admin/network.lua
```diff
@@ -76,36 +75,32 @@ end
...
-		else
-			util.remove_from_set(roles, 'mesh')
-		end
-		uci:set_list('gluon', 'iface_' .. iface, 'role', roles)
+	ifaces.default = config.role
 
-		-- Reconfigure on next reboot
-		uci:set('gluon', 'core', 'reconfigure', true)
+	function ifaces:write(data)
+		-- TODO: create section (and assign name) if not existing
```
AiyionPrime, line 99: is this done already?

lemoer: Yep. Thanks.

package/gluon-web-model/luasrc/usr/lib/lua/gluon/web/model/classes.lua
```diff
@@ -361,6 +361,66 @@ function ListValue:validate()
...
+end
+
+function MultiListValue:defaultvalue()
+	local value = self.default
+
+	if type(value) == "table" then
+		return value
+	else
+		return { value }
+	end
```
NeoRaider, line 440: It would be better to disallow non-table default values instead of fixing them up (same for `DynamicList`, where this was copied from)

package/gluon-web-network/luasrc/lib/gluon/config-mode/model/admin/network.lua
```diff
@@ -76,36 +75,23 @@ end
...
-			util.add_to_set(roles, 'mesh')
-		else
-			util.remove_from_set(roles, 'mesh')
-		end
-		uci:set_list('gluon', 'iface_' .. iface, 'role', roles)
+uci:foreach('gluon', 'interface', function(config)
+	local section_name = config['.name']
+	local ifaces = s:option(MultiListValue, section_name, config.name)
+
+	ifaces.widget = 'radio'
```
NeoRaider, line 88: `widget` seems to be unused.

package/gluon-web-network/luasrc/lib/gluon/config-mode/model/admin/network.lua
```diff
@@ -76,36 +75,31 @@ end
...
-local function wired_mesh(iface)
-	if not sysconfig[iface .. '_ifname'] then return end
-	local iface_roles = uci:get_list('gluon', 'iface_' .. iface, 'role')
+uci:foreach('gluon', 'interface', function(config)
+	local section_name = config['.name']
+	local ifaces = s:option(MultiListValue, section_name, pretty_ifnames[config.name] or config.name)
 
-	local option = s:option(Flag, 'mesh_' .. iface, wired_mesh_help[iface])
-	option.default = util.contains(iface_roles, 'mesh') ~= false
+	ifaces.widget = 'radio'
```
lemoer, line 88: Seems to be unused.

package/gluon-web-model/luasrc/usr/lib/lua/gluon/web/model/classes.lua
AiyionPrime: https://github.com/freifunk-gluon/gluon/blob/670da836c7c75769dd7abbf87fde66b713cdf2a1/package/gluon-web-model/luasrc/usr/lib/lua/gluon/web/model/classes.lua#L372-L380

package/gluon-web-network/luasrc/lib/gluon/config-mode/model/admin/network.lua
```diff
@@ -76,36 +75,23 @@ end
...
+	local section_name = config['.name']
+	local ifaces = s:option(MultiListValue, section_name, config.name)
+
+	ifaces.widget = 'radio'
+	ifaces.orientation = 'horizontal'
+	ifaces:value('uplink', 'Uplink') -- TODO: Uplink and Client should be mutually exclusive.
+	ifaces:value('mesh', 'Mesh')
+	ifaces:value('client', 'Client')
+
+	ifaces.default = config.role
```
NeoRaider, line 96: If the type fixup is removed from `classes.lua`, this should be `config.role or {}`

package/gluon-web-model/luasrc/usr/lib/lua/gluon/web/model/classes.lua
NeoRaider: We can still allow `nil` for convenience, making this something like `return self.default or {}`.

package/gluon-web-model/luasrc/usr/lib/lua/gluon/web/model/classes.lua
```diff
@@ -3,6 +3,7 @@
 -- Licensed to the public under the Apache License 2.0.
```
AiyionPrime, line 3: spdx

package/gluon-web-model/javascript/gluon-web-model.js
```diff
@@ -533,6 +547,7 @@
 			init_dynlist(node, attr);
 		}
 
+
```
NeoRaider, line 550: Stray new newline

package/gluon-web-model/javascript/gluon-web-model.js
```diff
@@ -13,7 +13,7 @@
 /*
```
AiyionPrime, line 13: spdx

package/gluon-web-network/luasrc/lib/gluon/config-mode/model/admin/network.lua
```diff
@@ -9,13 +9,12 @@ http://www.apache.org/licenses/LICENSE-2.0
 ]]--
```
AiyionPrime, line 9: spdx

AiyionPrime (commented):
Would be a good opportunity to fix these headers.

package/gluon-web-model/javascript/gluon-web-model.js
```diff
@@ -219,6 +219,20 @@
 				parent.parentNode.style.display = (parent.options.length <= 1) ? 'none' : '';
 		}
 
+		var nodes = document.querySelectorAll('[data-exclusive-with]');
+		for (var i = 0, node; (node = nodes[i]) !== undefined; i++) {
+			var excusive_with = JSON.parse(node.getAttribute('data-exclusive-with'));
+
+			node.disabled = false;
+			for (let list_item of excusive_with) {
```
NeoRaider, line 227: Let's use `var` for now, as the rest of the file also contains old JS only.

package/gluon-core/luasrc/lib/gluon/upgrade/021-interface-roles
```diff
@@ -63,4 +63,19 @@ for iface in pairs(interfaces) do
...
+	local function has_role(role)
+		return util.contains(interface.role, role)
+	end
+
+	if has_role('client') and (has_role('mesh') or has_role('uplink')) then
+		-- remove 'client' role
+		util.remove_from_set(interface.role, 'client')
+		uci:set('gluon', interface['.name'], 'role', interface.role)
+	end
+end)
```
NeoRaider, line 79: The exclusivity should also be checked in `check_site.lua` now.

lemoer: Now
````
	interfaces = {
		wan = {
			default_roles = { 'uplink', 'client' }
		}
	}
````
results in
```
*** site.conf error: expected interfaces.wan.default_roles to contain either "client" or "uplink", but not both.
```

**Comment by lemoer** [comment]
I addressed all requested changes. I pushed some fixups and some additional commits. Once this is approved, the fixup commits can be squashed.

This is ready for review again.

**Comment by Djfe** [comment]
Anything that speaks against offering client on a one lan port powerplug access point?
currently it's greyed out.
It obviously only works if the plug is connected via wifi mesh but could be used to connect wired clients to a Freifunk mesh.

**Comment by Djfe** [comment]
Oh, these are actually checkboxes not radio buttons.
I still feel like it's maybe not the most obvious, that you have to deselect uplink to be able to enable client, hmm
(atleast the first time you want to do that)

How do uplink and mesh work at the same time? VLAN?

Just an idea for a followup PR:
Would it make sense to disable vpn as long as there is no interface that is set to the role uplink? (even if the checkbox for vpn stays set)

File: package/gluon-web-network/luasrc/lib/gluon/config-mode/model/admin/network.lua [hunk 13]
```diff
@@ -76,36 +69,30 @@ end
 
 s = f:section(Section)
 
-local wired_mesh_help = {
-	single = translate('Enable meshing on the Ethernet interface'),
-	wan = translate('Enable meshing on the WAN interface'),
-	lan = translate('Enable meshing on the LAN interface'),
+local pretty_ifnames = {
+	["/wan"] = translate("WAN-Interfaces"),
```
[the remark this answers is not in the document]
NeoRaider: Please also update the POT file using `i18n-scan.pl`.

NeoRaider (dismissed):

Djfe:
Just a suggestion:
We could make the disabled checkboxes clickable. Clicking a disabled checkbox will uncheck and disable other checkbox(es) (in the same row) which are (mutually) exclusive to the one that was clicked.

lemoer:
@Djfe I would like to get this PR merged as it is. It's been around for long enough. I will not delay it even further by adding additional features which need to be reviewed again.

Feel free to open a PR including your suggestions once this is megred.

Djfe:
I agree. I don't want to delay this PR with my suggestion.

package/gluon-web-model/javascript/gluon-web-model.js
```diff
@@ -219,6 +215,20 @@
...
 		}
 
+		var nodes = document.querySelectorAll('[data-exclusive-with]');
+		for (var i = 0, node; (node = nodes[i]) !== undefined; i++) {
+			var excusive_with = JSON.parse(node.getAttribute('data-exclusive-with'));
+
+			node.disabled = false;
+			for (var list_item of excusive_with) {
+				var el = document.getElementById(node.name + '.' + list_item);
+				node.disabled |= el.checked;
```
NeoRaider, line 225:
```suggestion
				node.disabled ||= el.checked;
```

Also, this section of code could be implemented a bit more efficiently by having `node` disable `el` instead of vice-versa - this way, the inner loop could be skipped completely for unchecked `node`. We would need to iterate over `nodes` twice though (first to enable everything, then to disable what conflicts), so it would only make a difference for much longer exclusion lists (and might even make performance slightly worse in simple cases). So maybe it's not worth changing?

package/gluon-core/luasrc/lib/gluon/check-site.lua
```diff
@@ -370,6 +378,19 @@ function M.need_array_of(path, array, required)
...
 end
 
+function M.need_array_elements_exclusive(path, a, b, required)
+	local val = need_type(path, 'table', required, 'be an array')
+	if not val then
+		return nil
+	end
+
+	if contains(val, a) and contains(val, b) then
+		config_error(conf_src(path), 'expected %s to contain either "%s" or "%s", but not both.', path_to_string(path), a, b)
```
NeoRaider, line 388: This message is not quite accurate - it sounds like one of the values is required, but actually not having either is fine, too. Also, it should use the `format()` helper to achieve correct formatting for non-string array elements.

How about the following?
```suggestion
		config_error(conf_src(path), 'expected %s to contain only one of the elements %s and %s, but not both.', path_to_string(path), format(a), format(b))
```

(side node: `need_value()` also doesn't handle this correctly, it uses `tostring()` where `format()` would be preferable)

NeoRaider (changes requested):

package/gluon-web-model/javascript/gluon-web-model.js
```diff
@@ -219,6 +215,20 @@
 				parent.parentNode.style.display = (parent.options.length <= 1) ? 'none' : '';
 		}
 
+		var nodes = document.querySelectorAll('[data-exclusive-with]');
+		for (var i = 0, node; (node = nodes[i]) !== undefined; i++) {
+			var excusive_with = JSON.parse(node.getAttribute('data-exclusive-with'));
```
Djfe, line 220: I just noticed a typo
`exclusive_with` would be correct

lemoer: Thanks.

lemoer:
Thanks for the reviews I addressed all suggestions.

Djfe:
awesome! Thanks for your work @lemoer ❤️

PR #2690, opened 2022-10-31
gluon-status-page: show gateway_tq

lemoer:
The gateway_tq was added to respondd in #2596. Now, we also show it on the status page.

It looks like this:
image


package/gluon-status-page/javascript/status-page.js
```diff
@@ -133,6 +133,9 @@
 			}
 
 			return 'via ' + addr + ' (unknown iface)';
+		},
+		'tq': function(value) {
+			return formatNumber(100/255 * value, 1) + ' %';
```
AiyionPrime, line 138:
```suggestion
			return formatNumber(100/255 * value, 1) + '%';
```

removing the space would align it with RAM and fileysystem usage above.

AiyionPrime (changes requested):
Usually I'm a fan of the "one info per line"-concept the statuspage seems to follow,
but looking at the `(unknown iface)` below your addition this should be fine.

AiyionPrime (approved):
Looks fine, thanks.
Will merge once the action is done.

PR #2695, opened 2022-11-05
gluon-core: preserve channel list for 5G radios

Kistelini:
Currently the 'channels' setting of a 5G radio gets overwritten or deleted even if the 'preserve_channels' option is enabled. Only set the list of channels in outdoor mode if 'preserve_channels' is disabled and never delete the channel list (like for 2G radios).

### DIFF
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
@@ -53,16 +53,12 @@ local function is_outdoor()
 end
 
 local function get_channel(radio, config)
-	local channel
-	if wireless.preserve_channels(uci) then
-		-- preserved channel always wins
-		channel = radio.channel
-	elseif radio.band == '5g' and is_outdoor() then
+	if radio.band == '5g' and is_outdoor() then
 		-- actual channel will be picked and probed from chanlist
-		channel = 'auto'
+		return 'auto'
 	end
 
-	return channel or config.channel()
+	return config.channel()
 end
 
 local function get_htmode(radio)
@@ -179,6 +175,20 @@ local function configure_mesh_wireless(radio, index, config, disabled)
 	)
 end
 
+local function set_channels(radio, radio_name, config)
+	if wireless.preserve_channels(uci) then
+		return
+	end
+	local channel = get_channel(radio, config)
+	uci:set('wireless', radio_name, 'channel', channel)
+
+	local chanlist
+	if radio.band == '5g' and is_outdoor() then
+		chanlist = config.outdor_chanlist()
+	end
+	uci:set('wireless', radio_name, 'channels', chanlist)
+end
+
 wireless.foreach_radio(uci, function(radio, index, config)
 	local radio_name = radio['.name']
 
@@ -194,13 +204,13 @@ wireless.foreach_radio(uci, function(radio, index, config)
 		return
 	end
 
-	local channel = get_channel(radio, config)
 	local htmode = get_htmode(radio)
 	local beacon_interval = config.beacon_interval()
 
 	uci:delete('wireless', radio_name, 'disabled')
 
-	uci:set('wireless', radio_name, 'channel', channel)
+	set_channels(radio, radio_name, config)
+
 	uci:set('wireless', radio_name, 'htmode', htmode)
 	uci:set('wireless', radio_name, 'country', site.regdom())
 
@@ -218,13 +228,10 @@ wireless.foreach_radio(uci, function(radio, index, config)
 		uci:set_list('wireless', radio_name, 'hostapd_options', hostapd_options)
 
 		if is_outdoor() then
-			uci:set('wireless', radio_name, 'channels', config.outdoor_chanlist())
-
 			-- enforce outdoor channels by filtering the regdom for outdoor channels
 			uci:set('wireless', radio_name, 'country3', '0x4f')
 			configure_mesh_wireless(radio, index, config, true)
 		else
-			uci:delete('wireless', radio_name, 'channels')
 			uci:delete('wireless', radio_name, 'country3')
 			configure_mesh_wireless(radio, index, config)
 		end


package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -214,13 +214,14 @@ wireless.foreach_radio(uci, function(radio, index, config)
...
-			uci:set('wireless', radio_name, 'channels', config.outdoor_chanlist())
+			if not wireless.preserve_channels(uci) then
+				uci:set('wireless', radio_name, 'channels', config.outdoor_chanlist())
+			end
 
 			-- enforce outdoor channels by filtering the regdom for outdoor channels
 			uci:set('wireless', radio_name, 'country3', '0x4f')
 			configure_mesh_wireless(radio, index, config, true)
 		else
-			uci:delete('wireless', radio_name, 'channels')
```
blocktrron, line 223: This still needs to be wrapped in a condition to only apply in case the channel should not be preserved. Otherwise, without channel preservation the configuration could become invalid.

Kistelini: In which case?
```
channel '1'
channels '2-3'
```
is a valid configuration and will choose channel 1. This could lead to confusion if outdoor mode is turned on, then off and the `channels` configuration still exists, but I think it's even more confusing if the `channels` option is removed for 5G devices when it's not removed for 2G devices.
Another possibility would be to remove the `channels` option for 2G devices, too (without `preserve_channels`).

blocktrron: I think I have to wrap my head around that a little more.

The issue is, we only support `channels` in outdoor-mode, as the mesh-interface is only configured using netifd scripts which only utilize the `channel` setting.

blocktrron: On the other hand - one can also configure channel 99999 and choose to preserve it's settings, which will also result in broken configuration.

Given we don't expose this key in the UI - I think we should just prevent all altering of channel settings from upgrade-scripts in case this option is enabled.

Kistelini: So should I leave this as-is or should I add the if-condition?

blocktrron: Looking at it, i think we need a different approach

1. Refactoring the code flow, moving the the "channel" / "channels" configuration to a different dedicated function
2. Only calling this function in case preserve-channels is not activated.

Kistelini:
I'm not sure a function makes this more readable, but maybe like this?

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -179,6 +175,19 @@ local function configure_mesh_wireless(radio, index, config, disabled)
 	)
 end
 
+local function set_channels(radio, radio_name, config)
+	if not wireless.preserve_channels(uci) then
```
NeoRaider, line 179: Can return early if `wireless.preserve_channels(uci)` is true.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -179,6 +175,19 @@ local function configure_mesh_wireless(radio, index, config, disabled)
...
+		local channel = get_channel(radio, config)
+		uci:set('wireless', radio_name, 'channel', channel)
+
+		if is_outdoor() then
+			uci:set('wireless', radio_name, 'channels', config.outdoor_chanlist())
+		else
+			uci:delete('wireless', radio_name, 'channels')
+		end
```
NeoRaider, line 187: Set and delete can be combined:
```suggestion
		local chanlist
		if is_outdoor() then
			chanlist = config.outdoor_chanlist()
		end
		uci:set('wireless', radio_name, 'channels', chanlist)
```

NeoRaider (changes requested):

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -194,13 +204,11 @@ wireless.foreach_radio(uci, function(radio, index, config)
 		return
 	end
 
-	local channel = get_channel(radio, config)
 	local htmode = get_htmode(radio)
 	local beacon_interval = config.beacon_interval()
 
 	uci:delete('wireless', radio_name, 'disabled')
 
-	uci:set('wireless', radio_name, 'channel', channel)
```
NeoRaider, line 203: We're now missing channel configuration for `band == '2g'`. I think we can just call `set_channels()` before the band-specific part, and add a band check inside `set_channels`, so the outdoor_chanlist is only applied for 5g.

NeoRaider (changes requested):

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
Kistelini: Whoops, good catch!

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -179,6 +175,22 @@ local function configure_mesh_wireless(radio, index, config, disabled)
...
 end
 
+local function set_channels(radio, radio_name, config)
+	if wireless.preserve_channels(uci) then
+		return
+	end
+	local channel = get_channel(radio, config)
+	uci:set('wireless', radio_name, 'channel', channel)
+
+	if radio.band == '5g' then
```
NeoRaider, line 185: I think for consistency it would make sense to use the same condition `radio.band == '5g' and is_outdoor()` as in `get_channel()` (where the `is_outdoor()` check is right now, so the `uci:set(..., 'channels', chanlist)` call is also done for 2g, always setting it to nil)

NeoRaider (changes requested):

NeoRaider (approved):

PR #2696, opened 2022-11-05
gluon-core: preserve wireless channel width during updates

Kistelini:
This adds the uci option 'gluon.wireless.preserve_htmode'. When set to true the channel width of wireless radios (htmode) is preserved during updates.


blocktrron:
Hmm, I think it is a better idea to implicitly preserve the HT-mode when channel preservation is enabled. Operating mode in the end is defined by channel, band and HT-mode.

@mweinelt @NeoRaider any takes?

blocktrron:
Apart from the minor docs misspell, this looks very good.

Issue #2687, opened 2022-10-26
Re-enable ath10k-ct-smallbuffers for 64M ath10k boards


edeso:
i am using gluon on an TP-Link Archer C25v1 since years and 5GHz Wifi worked fine with  `ath10k-ct-smallbuffers` up until this was forced to use the mainline ath10k, 
https://github.com/freifunk-gluon/gluon/blob/master/targets/ath79-generic#L359-L363
which does not have a smallbuffers version. for now i resort to manually removing the enforcement.

would it be possible to remove the mainline `ath10k` enforcement from `tiny` devices in general in favor of the openwrt default, which is usually `ath10k-ct-smallbuffers`?  thanks!

blocktrron:
I've adjusted the title a bit.

> would it be possible to remove the mainline ath10k enforcement from tiny devices in general in favor of the openwrt default, which is usually ath10k-ct-smallbuffers?

The driver yes, the firmware no as far as I am mistaken, as ath10k-ct firmware does not support 11s on first generation chips.

Apart from that - go ahead. I don't necessarily think we should remove the broken flag, but it is primarily a task of doing research on a devices specifications and altering in the respective target file. A PR for that is very welcome!

edeso:
> I've adjusted the title a bit.

sure np 

> > would it be possible to remove the mainline ath10k enforcement from tiny devices in general in favor of the openwrt default, which is usually ath10k-ct-smallbuffers?
> 
> The driver yes, the firmware no as far as I am mistaken, as ath10k-ct firmware does not support 11s on first generation chips.

does ath10k-ct work with the mainline firmware? if so agreed, if not, considering it is broken completely with mainline ath10k i'd prefer the little bit more unbroken state, even not supporting 11s

> Apart from that - go ahead. I don't necessarily think we should remove the broken flag, but it is primarily a task of doing research on a devices specifications and altering in the respective target file. A PR for that is very welcome!

not sure how to proceed. i suggested to remove the `packages = ATH10K_PACKAGES_****,`  line. 
enforcing mainline non-ct firmware (assuming ct-driver works with it) will probably need a new variable definition, or?

blocktrron:
> does ath10k-ct work with the mainline firmware?

yes, firmware and driver are completely interchangable, ath10k-ct provides exclusive interfaces to -ct firmware

> if so agreed, if not, considering it is broken completely with mainline ath10k i'd prefer the little bit more unbroken state, even not supporting 11s

[...]

PR #2699, opened 2022-11-09
Enable ath10k-smallbuffers for low memory ath10k boards

edeso:
fixes OOM reboots due too limited ram with ath10k 5Ghz enabled add some comments to describe the need for ath10k-ct replacement tested stable on an TP-Link Archer C25v1
more details
https://github.com/openwrt/openwrt/commit/694757a08f620a9f24b70003542d9dcd0abeac46

fix issue [Re-enable ath10k-ct-smallbuffers for 64M ath10k boards #2687](https://github.com/freifunk-gluon/gluon/issues/2687)

### DIFF
--- a/targets/ath79-generic
+++ b/targets/ath79-generic
@@ -1,3 +1,6 @@
+-- enforce mainline ath10k kmod/firmware over openwrt default ath10k-ct usage
+-- ath10k-ct is unstable/broken with 11s meshing, works only wave2 chipsets
+
 local ATH10K_PACKAGES_QCA9880 = {
 	'kmod-ath10k',
 	'-kmod-ath10k-ct',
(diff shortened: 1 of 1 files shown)


blocktrron:
This is missing out on QCA9888 / 9886 based boards with 64M of RAM. They can also benefit from a similar adjustment.

edeso:
> This is missing out on QCA9888 / 9886 based boards with 64M of RAM. They can also benefit from a similar adjustment.

it does not. as you are not overriding ath10k packages for these boards the openwrt defaults apply. or what am i missing?

blocktrron:
You are correct, this is a side-effect of us removing the override we've had in the past.

blocktrron:
I've fixed this up when merging - Next time please put a proper prefix (ath79-generic in this case) in the commit title.

PR #2710, opened 2022-12-04
Wavlink WS-WN572HP3 (4G)

AiyionPrime:
  - ~~Web interface~~
  - [x] TFTP
  - [x] Other: Mediatek Webinterface details in OpenWrt commit
  - https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=dce66899bf243d78689afcc693340b891bbf4c2d
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs (only 4G LEDs available, none for wifi)
    - ~~Should show activity~~
  - Switch port LEDs
- Outdoor devices only:
- Docs:

------------------
  
This PR was intended to show @blocktrron's PR works, and does. 

> Not sure for how long, as the connected data volume is fairly limited; but the device works flawlessly with your code.  
  
_Originally posted by @AiyionPrime in https://github.com/freifunk-gluon/gluon/issues/2631#issuecomment-1336534774_

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -368,6 +368,10 @@ ramips-mt7621
   - EdgeRouter X-SFP
   - UniFi 6 Lite
 
+* Wavlink
+
+  - WS-WN572HP3 (4G)
+
 * ZBT
 
   - WG3526-16M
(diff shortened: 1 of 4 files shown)


blocktrron:
> Mediatek Webinterface detial sin OpenWrt commit

I can't read that

AiyionPrime:
> > Mediatek Webinterface detial sin OpenWrt commit
> 
> I can't read that

Jup, fixed it and added the commit id I meant.

Djfe:
oh cellular isn't supported in 22, yet.

PR #2716, opened 2022-12-07
ramips-mt7621: add TP-Link Archer C6 v3

Kistelini:
  - [x] Web interface
  - [x] TFTP
  - [ ] Other: <specify>
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [X] Should show activity
  - Switch port LEDs
- Outdoor devices only:
- Docs:

### DIFF
--- a/targets/ramips-mt7621
+++ b/targets/ramips-mt7621
@@ -49,6 +49,10 @@ device('netgear-wndr3700-v5', 'netgear_wndr3700-v5', {
 
 -- TP-Link
 
+device('tp-link-archer-c6-v3', 'tplink_archer-c6-v3', {
+	broken = true, -- LAN LED not working - review after resolving #2756
+})
+
 device('tp-link-re500-v1', 'tplink_re500-v1')
 
 device('tp-link-re650-v1', 'tplink_re650-v1')


Kistelini:
Testing isn't finished. There's a problem with the LAN LED: `gluon-reconfigure` generates this piece for `/etc/config/system`
```
config led 'led_lan'
	option name 'LAN'
	option sysfs 'green:lan'
	option trigger 'netdev'
	option mode 'link tx rx'
	option dev 'br-lan'
```
because of [these defaults](https://github.com/openwrt/openwrt/blob/master/target/linux/ramips/mt7621/base-files/etc/board.d/01_leds#L150), but Gluon doesn't have `br-lan`  so the LED doesn't work. How do I fix this?

Kistelini:
Those `mt7530-*` triggers don't do anything. Everything else works!

blocktrron:
I think we have to take a look at software-controlled LEDs spanning multiple LAN ports in conjunction with DSA, as this is currently not handled nicely upstream. I assume some other devices have lost sane handling with migrating to DSA also.

targets/ramips-mt7621
```diff
@@ -49,6 +49,10 @@ device('netgear-wndr3700-v5', 'netgear_wndr3700-v5', {
 
 -- TP-Link
 
+device('tp-link-archer-c6-v3', 'tplink_archer-c6-v3', {
+	broken = true, -- LAN LED not working
```
AiyionPrime, line 53:
```suggestion
	broken = true, -- LAN LED not working - review after resolving #2756
```

AiyionPrime (changes requested):
@Kistelini Sorry this took so long, took a moment to understand how swconfig did it and why this is not trivial to fix upstream.
I'd like to add the expiration note for when the broken status is expected to be fixed, but other than that the PR looks fine.

Djfe:
Should we add the Archer A6 v3 as well? They share the exact same dtsi/dts, the same led config etc. They just have a different name, which might influence the tp-link safeloader. (so it probably has to be a different entry instead of an alias, right?)

AiyionPrime:
As soon as someone wants to use the device, we've got someone to test it.
We usually do not merge untested devices anymore. Even if we think the difference is minor and it should just work.
Whoever wants to use and test it, will have a smooth run integrating it :)

Djfe:
@AiyionPrime backport needed

AiyionPrime:
> @AiyionPrime backport needed

Won't happen, it's broken.

We'll publish the list of upcoming backports in the tracking issue tomorrow morning again. (Hoping GitHub does not eat it twice -.-')

PR #2719, opened 2022-12-11
docs: add broken devices and architectures

Kistelini:
I think it would be useful to have the broken devices in the documentation, too.


AiyionPrime:
I'm not sure I'd like this, as it might encourage users to add more broken devices.
It feels like this draft gives them some sort of official character.
But that's only my opinion, others might differ.

AiyionPrime:
@Kistelini I'm sorry; we spoke about this PR in hackint#gluon and the others tend to agree on the matter.
I'll close this for now, but feel free discuss this with us in #gluon.

PR #2723, opened 2022-12-12
gluon-web-model: resolve hang when submitting form with disabled element

mkg20001:
I was made aware of a bug when submitting the form while the element is
disabled based on it's dependencies

(disable "enabled" in "private wifi", then save, while having mesh on wan enabled)

The fix was to inherit from AbstractValue instead of just node

AbstractValue's AbstractValue:resolve_node_depends() in particular
solves the issue, but it made more sense to just use the full base class

Also includes translations

### DIFF
--- a/package/gluon-web-private-wifi/i18n/de.po
+++ b/package/gluon-web-private-wifi/i18n/de.po
@@ -49,6 +49,9 @@ msgstr ""
 msgid "WPA3"
 msgstr ""
 
+msgid "Meshing on WAN interface is enabled. This can lead to problems."
+msgstr "Mesh auf WAN ist aktiv. Dies kann zu Problemen führen."
+
 msgid ""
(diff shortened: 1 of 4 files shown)


package/gluon-web-private-wifi/i18n/de.po
```diff
@@ -49,6 +49,9 @@ msgstr ""
 msgid "WPA3"
 msgstr ""
 
+msgid "Meshing on WAN interface is enabled. This can lead to problems."
+msgstr "Mesh auf WAN ist aktiv. Das kann zu Problemen führen."
```
NeoRaider, line 53: Vorschlag:
```suggestion
msgstr "Mesh auf WAN ist aktiv. Dies kann zu Problemen führen."
```

AiyionPrime:
This looks promising. Thanks for the fix!

PR #2726, opened 2022-12-16
docs: consistently indent .rst files with 2 spaces

AiyionPrime:
This PR updates to match our editorconfig like @neoraider did before in https://github.com/freifunk-gluon/gluon/commit/854fef4e12c07751c4da78081ac534e87ade9afd.

The rendered docs do not change.

Not sure about the fixup though.
It appears the index files were not touched in @neoraiders commit.
Do we want the toc to stay three spaced?

If so, we'd need to add an exception to our editorconfig like this:

```
[docs/**/index.rst]
indent_size = 3
```

~~Please let me know, whether to include the exception or to squash the fixup.~~

**edit:** we don't bother much, as long as it's consistent either way.
We might change the indetation to three spaces one day to match sphinx' defaults, once we've got no other problems left.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -6,80 +6,80 @@ Several Freifunk communities in Germany use Gluon as the foundation of their Fre
 
 
 .. toctree::
-   :caption: User Documentation
-   :maxdepth: 2
+  :caption: User Documentation
+  :maxdepth: 2
 
-   user/getting_started
-   user/site
-   user/supported_devices
-   user/x86
-   user/faq
-   user/mtu
-
-.. toctree::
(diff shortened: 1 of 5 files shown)


AiyionPrime:
> 11:14 <neoraider> aiyion: hmm, it's the default created by sphinx-quickstart. I guess it should be considered alignment with the ".. " lines, rather than indentation, and thus be unaffected by indentation rules?      
11:15 <neoraider> I don't care too much though - if it's easier for the tooling, we can just say it's indentation and use 2 spaces

_Originally posted by @NeoRaider in gluon#hackint_

PR #2728, opened 2022-12-18
editorconfig: add dia files

AiyionPrime:
> 2 spaces

the hoodselector contains such in his docs


AiyionPrime:
curly brackets were an artifact; does work now.

PR #2730, opened 2022-12-18
Minimized js naming

### LINKED ISSUE #2729: .min.js are not always using their appropriate suffix
minified js files should end in `.min.js`.
We mostly adhere to this, but overlooked three occurrences:

- [ ] [package/gluon-status-page/files/lib/gluon/status-page/www/static/status-page.js](https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-status-page/files/lib/gluon/status-page/www/static/status-page.js)
- [ ] [package/gluon-web-model/files/lib/gluon/web/www/static/gluon-web-model.js](https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-web-model/files/lib/gluon/web/www/static/gluon-web-model.js)
- [ ] [package/gluon-web-osm/files/lib/gluon/web/www/static/gluon-web-osm.js](https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-web-osm/files/lib/gluon/web/www/static/gluon-web-osm.js)

Renaming them and their usages to end in `min.js` would help IDEs to not reformat them.

AiyionPrime:
~This won't compile. I've made at least one mistake in the statuspage I think.~

Compiles just fine, testing is needed though.

The goal is to keep the `js`-suffix for deployed files, regardless of whether they are minimized or not.

Eventually, this resolves #2729


package/gluon-config-mode-core/Makefile
```diff
@@ -18,7 +18,7 @@ define Package/gluon-config-mode-core/install
 	$(INSTALL_DIR) $(1)/lib/gluon/config-mode/view/ $(1)/lib/gluon/config-mode/www/static/
 	$(LN) /lib/gluon/web/i18n $(1)/lib/gluon/config-mode/
 	$(LN) /lib/gluon/web/view/{error,model} $(1)/lib/gluon/config-mode/view/
-	$(LN) /lib/gluon/web/www/static/gluon-web-model.js $(1)/lib/gluon/config-mode/www/static/
+	$(LN) /lib/gluon/web/www/static/gluon-web-model.min.js $(1)/lib/gluon/config-mode/www/static/gluon-web-model.js
```
AiyionPrime, line 21: This...

package/gluon-config-mode-geo-location-osm/Makefile
```diff
@@ -13,7 +13,7 @@ define Package/gluon-config-mode-geo-location-osm/install
 	$(Gluon/Build/Install)
 
 	$(INSTALL_DIR) $(1)/lib/gluon/config-mode/www/static/
-	$(LN) /lib/gluon/web/www/static/gluon-web-osm.js $(1)/lib/gluon/config-mode/www/static/
+	$(LN) /lib/gluon/web/www/static/gluon-web-osm.min.js $(1)/lib/gluon/config-mode/www/static/gluon-web-osm.js
```
AiyionPrime, line 16: ...and this are hardly right.
I do not see yet why this file lies in `/lib/gluon/web/www/static/`, yet.

The config-mode uses it from `/lib/gluon/config-mode/www/static`, so why put it here as well?

AiyionPrime:
@NeoRaider I added minify support for `gluon-config-mode-core` and `gluon-config-mode-geo-location-osm`, as requested. Anything else?

package/gluon-config-mode-geo-location-osm/Makefile
```diff
@@ -13,7 +13,11 @@ define Package/gluon-config-mode-geo-location-osm/install
 	$(Gluon/Build/Install)
 
 	$(INSTALL_DIR) $(1)/lib/gluon/config-mode/www/static/
-	$(LN) /lib/gluon/web/www/static/gluon-web-osm.js $(1)/lib/gluon/config-mode/www/static/
```
NeoRaider, line 16: This is incorrect and won't build without `CONFIG_GLUON_MINIFY` - the file is in `gluon-web-osm`, `gluon-config-mode-geo-location-osm` only creates a link. I don't think `gluon-config-mode-geo-location-osm` needs to be changed in this PR. Same thing for `gluon-config-mode-core` - just have `gluon-web-model` install one or the other file, without changing the creation of the config-mode link.

AiyionPrime: First part I missed, thanks.
Second was by choice, so that we'd keep the `.min.js` in `/lib/gluon/web/www/static/`.
But I'm indifferent, will change that there as well.

AiyionPrime:
Built locally, saving the image, doing another compile with the minified flag.

package/gluon-config-mode-geo-location-osm/Makefile
NeoRaider: I don't think we should leave the `.min.js` for any files in `/lib/gluon/web/www/static/`, so other packages' symlinks don't need to depend on `CONFIG_GLUON_MINIFY`.

package/gluon-web-osm/Makefile
```diff
@@ -8,4 +8,14 @@ define Package/gluon-web-osm
   TITLE:=base src for OSM inlay
 endef
 
+define Package/gluon-config-mode-core/install
+       $(Gluon/Build/Install)
+
+ifdef CONFIG_GLUON_MINIFY
```
NeoRaider, line 14: I think it would be a good idea to add `PKG_CONFIG_DEPENDS += CONFIG_GLUON_MINIFY` to `gluon-web-osm` and `gluon-web-model`. While it's added implicitly by `gluon.mk` due to the existence of a `luasrc` directory, it seems clearer to add it explicitly, as it isn't only used for Lua in these packages now.

package/gluon-web-osm/Makefile
```diff
@@ -8,4 +8,14 @@ define Package/gluon-web-osm
...
+       $(CP) ./javascript/gluon-web-osm.min.js $(1)/lib/gluon/web/www/static/
+else
+       $(CP) ./javascript/gluon-web-osm.js $(1)/lib/gluon/web/www/static/
```
NeoRaider, line 17: We could go with something like this:
```suggestion
       $(CP) ./javascript/gluon-web-osm$(if $(CONFIG_GLUON_MINIFY),.min).js $(1)/lib/gluon/web/www/static/gluon-web-osm.js
```
Not entirely sure if I like that better or not though.

AiyionPrime: Thought about it, but didn't like it enough. If you want it I can change it.

AiyionPrime:
Second build succeeded as well, will flash the two and then merge the remaining recommendations.

package/gluon-web-osm/Makefile
AiyionPrime: resolved in ed3551a5ca734556f861f4c62ffb199f2830ad44 and 1ef6daafdff3133314dcb6a38fed2a0f9e93c3f6

package/gluon-config-mode-geo-location-osm/Makefile
AiyionPrime: this is hopefully resolved and outdated since ed3551a5ca734556f861f4c62ffb199f2830ad44

package/gluon-config-mode-core/Makefile
```diff
@@ -18,7 +18,6 @@ define Package/gluon-config-mode-core/install
 	$(INSTALL_DIR) $(1)/lib/gluon/config-mode/view/ $(1)/lib/gluon/config-mode/www/static/
 	$(LN) /lib/gluon/web/i18n $(1)/lib/gluon/config-mode/
 	$(LN) /lib/gluon/web/view/{error,model} $(1)/lib/gluon/config-mode/view/
-	$(LN) /lib/gluon/web/www/static/gluon-web-model.js $(1)/lib/gluon/config-mode/www/static/
```
AiyionPrime, line 21: this line oght not be gone, will fix this in three hours

package/gluon-config-mode-geo-location-osm/Makefile
```diff
@@ -13,7 +13,6 @@ define Package/gluon-config-mode-geo-location-osm/install
...
```
AiyionPrime, line 16: and neither should this.

AiyionPrime (commented):
needs changes.

package/gluon-config-mode-geo-location-osm/Makefile
AiyionPrime: resolved in 325e8733f73c9526bad2f483f8e712c015bef0ba

package/gluon-config-mode-core/Makefile
AiyionPrime: 90b47378b0c99bd6f9e62ff9da1d44f4d5a2e449

AiyionPrime:
Compiled locally with `GLUON_MINIFIED=0` works like a charm and both the config mode as wellas the statuspage files are not minified and served properly.

AiyionPrime:
Compiled it again without minified, web-osm is there, web-model is not.
Looks like I'm on the right track but doing yet another round.

AiyionPrime:
Ugh. neither is working, just hit a cache entry. Time to call it a day -.-'
Good night.

AiyionPrime:
This is tested with both `MINIFIED` enabled and omitted and all three files work jsut fine.

@NeoRaider another review round or merge would be welcome.

package/gluon-web-model/Makefile
```diff
@@ -9,4 +9,17 @@ define Package/gluon-web-model
...
 endef
 
+PKG_CONFIG_DEPENDS += CONFIG_GLUON_MINIFY
+
+define Package/gluon-web-model/install
+	$(Gluon/Build/Install)
+	$(INSTALL_DIR) $(1)/lib/gluon/web/www/static/
+
+ifdef CONFIG_GLUON_MINIFY
+	$(CP) ./javascript/gluon-web-model.min.js $(1)/lib/gluon/web/www/static/gluon-web-model.js
```
NeoRaider, line 19: As these aren't copying recursively, they should use `$(INSTALL_DATA)` instead of `$(CP)` (including the one you didn't change in `gluon-status-page`). The rest LGTM.

AiyionPrime: resolved in 6c0f22445651647a5cd2e7be08e4c7ef6c02e42b, 389314f58a202b36194eec78b084070399025da6, 0ec32078303cb3843408754502529422b9ce5408

AiyionPrime:
Built and verified an image with `GLUON_MINIFY=0`

PR #2738, opened 2022-12-28
ramips-mt7621: Add Xiaomi Mi Router 3G

GoliathLabs:
  - [ ] Web interface
  - [ ] TFTP
  - [X] Other: [https://openwrt.org/toh/xiaomi/mir3g#installation](https://openwrt.org/toh/xiaomi/mir3g#installation)
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [X] Should map to their respective radio (system LED turns from orange into blue)
    - [ ] Should show activity (as no LED for radio is present)
  - Switch port LEDs
- Outdoor devices only:
- Docs:

### DIFF
--- a/targets/ramips-mt7621
+++ b/targets/ramips-mt7621
@@ -66,10 +66,13 @@ device('xiaomi-mi-router-4a-gigabit-edition', 'xiaomi_mi-router-4a-gigabit', {
 	factory = false,
 })
 
-device('xiaomi-mi-router-3g-v2', 'xiaomi_mi-router-3g-v2', {
+device('xiaomi-mi-router-3g', 'xiaomi_mi-router-3g', {
 	factory = false,
 })
 
+device('xiaomi-mi-router-3g-v2', 'xiaomi_mi-router-3g-v2', {
+	factory = false,
+})
 
 -- ZBT
 
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -376,7 +376,7 @@ ramips-mt7621
 * Xiaomi
 
   - Xiaomi Mi Router 4A (Gigabit Edition)
-  - Xiaomi Mi Router 3G v2
+  - Xiaomi Mi Router 3G (v1, v2)
 
 ramips-mt76x8
 -------------


GoliathLabs:
Device has been tested successfully with a FFMUC build

AiyionPrime (changes requested):
Thanks. Looks good in general.
The PR for the v2 variant is not a good example for adding device support though.

Please squash the three commits into one with the commit message
`ramips-mt7621: add Xiaomi Mi Router 3G`.

That should be all.
@rotanid ack?

GoliathLabs:
> Thanks. Looks good in general. The PR for the v2 variant is not a good example for adding device support though.
> 
> Please squash the three commits into one with the commit message `ramips-mt7621: add Xiaomi Mi Router 3G`.
> 
> That should be all. @rotanid ack?

Done :)

AiyionPrime (approved):

AiyionPrime:
Ugh. Just saw the commit message is still borked. I'll fix it.

AiyionPrime:
changed nothing but the commit message, aborting build and merging.

AiyionPrime:
@GoliathLabs Thanks for your contribution.

rotanid:
@GoliathLabs how did you flash the router? the link you provided has multiple instructions and multiple devices...

GoliathLabs:
I got the device bricked from a friend. I went with the UART method

GoliathLabs:
@rotanid Flashed the device via the OpenWRTInvasion method without the need for opening the device.

AiyionPrime:
Thank you very much!

targets/ramips-mt7621
```diff
@@ -66,10 +66,13 @@ device('xiaomi-mi-router-4a-gigabit-edition', 'xiaomi_mi-router-4a-gigabit', {
 	factory = false,
 })
 
-device('xiaomi-mi-router-3g-v2', 'xiaomi_mi-router-3g-v2', {
+device('xiaomi-mi-router-3g', 'xiaomi_mi-router-3g', {
 	factory = false,
 })
```
blocktrron, line 71: FWIW, the filename should probably contain the v1 designator and the string the autoupdater is looking for should be added as a manifest-alias.

AiyionPrime: I thought about patching it in OpenWrt, but would be fine with a downstream solution as well.

Djfe:
@AiyionPrime backport needed

PR #2739, opened 2022-12-29
gluon-respondd: restart instead of reload - fixes #2733

### LINKED ISSUE #2733: respondd reloads: are they working?
### Bug report

**What is the problem?**
reloading respondd should work and respondd should re-listen on interfaces

the init.d trigger uses reload

```
	procd_open_trigger
	procd_add_raw_trigger "interface.*" 0 "/etc/init.d/$name" reload
	procd_close_trigger

```

there's no reload method so it's propably sighup

https://github.com/freifunk-gluon/packages/blob/master/net/respondd/src/respondd.c doesn't show any signs of checking for SIGHUP or other reload mechanisms

note: just patching it so it is restarted instead is working and might be preferred because otherwise an interface similar to mmfd's socket would be needed (to remove/add interfaces) but respondd doesn't really carry that much state or cause trouble when it's down for a sec that I think restart would just do the job

it reloads and picks up the interfaces, then continues doing the job of picking up packets and responding to them

**Gluon Version:**
85f3e40a14c5f346fb3ad320b2efe6d3031459b4

**Site Configuration:**
https://github.com/ffgraz/funkfeuer-graz-gluon-site

**Custom patches:**
whatever graz has

mkg20001:
respondd has to handle interface changes for re-listening

respondd was being sent sighup by procd previously, but didn't handle it

The easiest way is to just restart respondd to make it re-listen, since it does not carry any important state

### DIFF
--- a/package/gluon-respondd/files/etc/init.d/gluon-respondd
+++ b/package/gluon-respondd/files/etc/init.d/gluon-respondd
@@ -24,6 +24,6 @@ service_triggers() {
 	local name=$(basename ${script:-$initscript})
 
 	procd_open_trigger
-	procd_add_raw_trigger "interface.*" 0 "/etc/init.d/$name" reload
+	procd_add_raw_trigger "interface.*" 0 "/etc/init.d/$name" restart
 	procd_close_trigger
 }


mkg20001:
I'm now getting a bunch of

`Sat Dec 31 17:21:49 2022 daemon.err respondd[8718]: bind failed: Address in use
`
not sure if socket is closed correctly or some other error, perhaps allowing multiple processes to bind the address might fix it, if it can't be unlistened properly

T-X:
@mkg20001 that sounds like respondd is doing the lazy approach and is not closing sockets on its own on/before exit (haven't checked). The Linux kernel usually needs a few seconds after the application exited to recognize this to allow reusing them then. If that's the case I think explicit socket closing should be added to respondd to avoid this issue.

(Another workaround was to use SO_REUSEADDR or SO_REUSEPORT, if I remember correctly, not sure. See "man 7 socket".)

NeoRaider:
@mkg20001 This change should not have any effect at all. As the initscript doesn't specify an explicit reload handler, `reload` will already restart the service. (If a SIGHUP was sent - which is not the case AFAICT - this would still cause a restart, as the default action for SIGHUP is to terminate the process)

@T-X I don't think there is any delay in closing sockets on process cleanup for UDP.

mkg20001:
so I've seen this problem occur on lower power devices (ath79, etc) but my x86-64 devices seem unaffected. this could either be because openvpn is more stable there (it's a vm on hetzner) or because the lower power devices don't handle the restarts quick enough for there not to be any races.

does procd handle multiple restarts gracefully? as in queue them or just do one? or does it do them all at once?

the issue persists as in even later when I restart gluon-respondd the address is still in use

also note: in graz we're using a bunch of openvpn that can go up and down, while in prob most other gluon setups interfaces rarely go up and down, so this might add to why that race doesn't occur in your case. maybe just try violently up/downing links with that patch applied on some low-end device.

mkg20001:
@NeoRaider sighup was what I guessed procd would do, but I could be wrong. regardless I haven't seen that issue before, it seems new, so maybe sighup was a wrong guess.

PR #2740, opened 2022-12-30
[v2022.1.x] modules: update to latest in master

rotanid:


herbetom:
I think it was said some time ago that, going forward, modules updates should also be cherry picked from master.

rotanid:
hm, sorry, if that's the case i didn't remember it.
i don't really see the point though, because i change might work in master but not in stable branch because there are other changes done in master which aren't backported to stable branch...

PR #2743, opened 2023-01-03
treewide: fix indentation in .sh files

AiyionPrime:
> to match the editorconfig
> 
> awk code in `contrib/sign.sh` and `contrib/sigtest.sh` reformatted as gawk does

The awk-code is now formatted the same way for both awk snippets,
I tested `contrib/sigtest.sh` on a manifest generated before this PR, does work as expected.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -37,6 +37,8 @@ indent_size = 4
 indent_style = space
 indent_size = 2
 
+[*.sh]
+
 [*.yml]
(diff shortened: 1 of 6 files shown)


AiyionPrime:
I used `gawk -o-` to output the awk snippet formatted.

rotanid (approved):

lemoer (approved):
LGTM.

PR #2746, opened 2023-01-03
html indentation

AiyionPrime:
I'm not sure whether this breaks the layout or not; will test it later this afternoon.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -21,6 +21,8 @@ indent_size = unset
 indent_style = space
 indent_size = 2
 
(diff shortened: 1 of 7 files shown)


rotanid (approved):
i haven't built and checked the result, but i can't see any obvious problem with this code

AiyionPrime:
I have built the image and it does not appear to disturb the frontend.

AiyionPrime:
rebased onto master after merging #2730 introduced a conflict in .editorconfig

PR #2748, opened 2023-01-03
not meant to be merged,

AiyionPrime:
poc that tabs do indeed break makefiles, when used in `define`s


AiyionPrime:
nevermind? Weird, but maybe way easier than I thought.

AiyionPrime:
I should have used a package, that actually gets build.

PR #2752, opened 2023-01-06
gluon-wireless-encryption{,-wpa3}: rename to match actual package name

mkg20001:


NeoRaider:
@blocktrron Do you remember why the source and binary package names differ?

AiyionPrime:
I think one of the earlier drafts was more about holding the package generic, wpa2 and wpa3 were both "just" features in the Makefile which could have been chosen from.

You didn't like the approach and asked him to change it.

> Having a `wpa2` feature at all does not make sense to me; making it a part of the `wpa3` as well package even less.
> 
> Let's drop the flag and the package, and directly reference `hostapd-mini` in `package/features`.

_Originally posted by @NeoRaider in https://github.com/freifunk-gluon/gluon/pull/1876#discussion_r389364546_

mkg20001:
the current wpa3 package adds the wpa3 file which the webui uses to check which encryptions are available. if there's a way to directly get this information from perhaps the hostapd library via a command/binding/whatever then that could also be used and the package could be dropped.

blocktrron:
This is available using `hostapd -vsae`. However, we can not drop this package, as it is required for adding the appropriate hostapd version.

NeoRaider:
> note that currently the old PKG_NAME is invalid but the -wpa3 one works despite that because (presumably) the call to BuildPackageGluon, which uses the right package name, still makes it a valid package

I wouldn't call it "invalid" - there is nothing wrong with source and binary package names not matching. Still, in this case there isn't a good reason to have them differ, so I merged this.

PR #2759, opened 2023-01-15
Restore local DNS caching

mweinelt:
Bringing cached DNS responses closer to the client was always a great idea,

Missing DNSSEC support and space constraints was were the reasons why we
disabled this feature last time round. We now live in an 8 MB-ish world, so it
should be feasible to restore this functionality.

I've taken `dnsmasq-full` and stripped it from most feature, so we arrive at the
previous default package plus DNSSEC support. 

The feature was easy to restore, as I just went for a revert of the revert.

I've verified that I can retrieve DNSSEC records from the cache


rotanid:
thanks :)
did you already have a look at the additionally used space by this change?

mweinelt:
The cost is roughly 60 kB.

192352
133337

docs/features/dns-cache.rst
```diff
@@ -0,0 +1,52 @@
...
+the content of next_node.name may be resolved even without upstream connectivity.
+It is suggested to use the same name as the DNS server provides:
+e.g. nextnode.location.community.example.org (This way the name also works if a
+client uses static DNS Servers). Hint: If next_node.name does not contain a dot
+some browsers would open the searchpage instead.
+
+::
+
+  dns = {
+    cacheentries = 5000,
```
blocktrron, line 40: Does it make sense to make this configurable or should we preset this based on the amount of system memory the running system offers?

mweinelt: 90 Bytes per Cache Entry means we allocate just under 450 kB for 5000 entries.

We should likely default to a reasonable value, like 4k, but I'm not opposed to letting communities tune this value.

docs/features/dns-cache.rst
```diff
@@ -0,0 +1,52 @@
...
+  },
+
+  next_node = {
+    name = { 'nextnode.location.community.example.org', 'nextnode', 'nn' },
+    ip6 = '2001:db8:8::1',
+    ip4 = '198.51.100.1',
+  }
+
+
+The cache will be initialized during startup.
```
blocktrron, line 51: This adds no value

mweinelt: Agreed, let's clean up the docs a bit.

PR #2768, opened 2023-01-23
ath79-nand: add support for GL.iNet GL-XE300

herbetom:
  - [x] Web interface
  - [ ] TFTP
  - [x] Other: Bootloader recovery interface
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [x] Should show activity
  - Switch port LEDs
  - 3G/4G LED
- Outdoor devices only:
- Docs:

Related Upstream PRs:
- https://github.com/openwrt/openwrt/pull/11845
- https://github.com/openwrt/openwrt/pull/11870
- https://github.com/openwrt/openwrt/pull/11908

### DIFF
--- a/package/gluon-setup-mode/files/lib/gluon/setup-mode/rc.d/S96led
+++ b/package/gluon-setup-mode/files/lib/gluon/setup-mode/rc.d/S96led
@@ -3,17 +3,25 @@
 START=96
 
 start() {
+	local custom_led
+
 	/etc/init.d/led start
 
 	. /etc/diag.sh
 
 	get_status_led 2> /dev/null
 
-	if [ -z $status_led ]; then
+	if [ -z "$status_led" ]; then
 		status_led="$running"
 	fi
-	if [ -z $status_led ]; then
+
+	if [ -z "$status_led" ]; then
 		status_led="$boot"
+	fi
+
+	custom_led="$(lua -e 'print(require("gluon.setup-mode").get_status_led() or "")')"
+	if [ -z "$status_led" ]; then
+		status_led="$custom_led"
 	fi
 
 	status_led_set_timer 1000 300
--- a/package/gluon-setup-mode/luasrc/usr/lib/lua/gluon/setup-mode.lua
+++ b/package/gluon-setup-mode/luasrc/usr/lib/lua/gluon/setup-mode.lua
@@ -1,0 +1,14 @@
+local platform = require 'gluon.platform'
+
+
+local M = {}
+
+function M.get_status_led()
+	if platform.match('ath79', 'nand', {
+		'glinet,gl-xe300',
+	}) then
+		return "green:wlan"
+	end
+end
+
+return M
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -143,6 +143,7 @@ ath79-nand
 
   - GL-AR300M
   - GL-AR750S
+  - GL-XE300
 
 * Netgear
 
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/250-cellular
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/250-cellular
@@ -43,6 +43,10 @@ local function setup_ncm_qmi(devpath, control_type, delay)
 end
 
 if platform.match('ath79', 'nand', {
+	'glinet,gl-xe300',
+}) then
+	setup_ncm_qmi('/dev/cdc-wdm0', 'qmi', 15)
+elseif platform.match('ath79', 'nand', {
 	'zte,mf281',
 }) then
 	setup_ncm_qmi('/dev/ttyACM0', 'ncm', 15)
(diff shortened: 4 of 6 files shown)


blocktrron:
> The main issue i'm currently having with this device is that the primary mac is currently the wifi mac and not the wan mac which is printed on the label.

I think this should be fixed in the upstream DTS. There is no label-mac defined in the DTS. Can you verify which interface (eth0 / eth1) is assigned the label-mac?

herbetom:
For me it appears to be the followig with Gluon:

WAN: eth1 with the mac `:21` which is also on the Label
LAN: eth0 with the mac `:22`
WLAN: has the mac `:23`

From this i would have figured `label-mac-device = &eth1;`, hoewer that resulted in `:22` beeing the the primary mac. With `label-mac-device = &eth0;` the primary mac is correctly set to `:21`. I've opened https://github.com/openwrt/openwrt/pull/11870 as draft.

blocktrron:
> hoewer that resulted in :22 beeing the the primary mac. With label-mac-device = &eth0; the primary mac is correctly set to :21.

This is due to the fact the "ethX" blocks in te DTS refers to the GMAC block idx in the address map. The switch being connected to the MDIO Bus of GMAC1 leads to the being probed in reverse due to the PHY of GMAC0 being connected to the MDIO bus of GMAC1.

herbetom:
Currently the config mode blink sequence isn't beeing displayed. 

The XE300 has the following LEDs:
- 4x Battery Charge Indicator LEDs
- PWR
- WAN
- LAN
- WIFI
- 3G/4G

Hoewer it seems that only the last 4 are controlled by OpenWRT:

https://github.com/openwrt/openwrt/blob/afb5d9e9d5a5ecafceb0e3d4ffe1210077f8a40b/target/linux/ath79/dts/qca9531_glinet_gl-xe300.dts#L42-L63

I've confirmed that with the Vendor Firmware also only those are available:

```
root@GL-XE300:~# ls /sys/class/leds/
ath9k-phy0           gl-xe300:green:lan   gl-xe300:green:lte   gl-xe300:green:wan   gl-xe300:green:wlan
```

So currently there is no `led_status` or `led-boot` and therefore no blink sequence.  

Also there currently is no config for the 3G/4G LED and it stays off. I guess OpenWRT doesn't know which name a user will give the LTE interface when defining it so there isn't a really good way of definig that.
If the LTE LED should be utilized for displaing the LTE activity the LED mapping probably needs to happen in Gluon. On the GL-XE300 the WWAN interface is called wwan0.

herbetom:
Only remaining issue is the missing Config Mode Blink sequence.

herbetom:
All issues are resolved and it's ready for review.

package/gluon-setup-mode/files/lib/gluon/setup-mode/rc.d/S96led
```diff
@@ -16,5 +18,10 @@ start() {
 		status_led="$boot"
 	fi
 
+	custom_led="$(lua -e 'print(require("gluon.setup-mode").get_custom_led() or "")')"
```
NeoRaider, line 21: If we really want to do this in Gluon (which I'm not convinced of - having a failsafe indicator in OpenWrt is always valuable), I'd rather name the function `get_status_led()`.

package/gluon-setup-mode/files/lib/gluon/setup-mode/rc.d/S96led
```diff
@@ -16,5 +18,10 @@ start() {
...
+	if [ -z $status_led ]; then
```
NeoRaider, line 22: Hmm, does our linter not cover this file yet? Shellcheck should complain:
```suggestion
	if [ -z "$status_led" ]; then
```

package/gluon-setup-mode/luasrc/usr/lib/lua/gluon/setup-mode.lua
```diff
@@ -0,0 +1,15 @@
...
+
+local M = {}
+
+function M.get_custom_led()
+	if platform.match('ath79', 'nand', {
+		'glinet,gl-xe300',
+	}) then
+		return "green:wlan"
+	end
+	return nil
```
NeoRaider, line 12: Trailing `return nil` can be omitted.

NeoRaider (dismissed):

package/gluon-setup-mode/files/lib/gluon/setup-mode/rc.d/S96led
AiyionPrime: @blocktrron your argument for not adding them upstream could be valuable here.

herbetom: I'm fine with it either way. I would just appreciate a clear direction before I test it hopefully one last time.

Djfe:
@AiyionPrime backport needed

PR #2770, opened 2023-01-28
[v2022.1.x] update modules

blocktrron:
This fixes mesh on WiFi 6 devices being used with 54 Mbit/s max due to insufficient parsing of HE operation modes.


rotanid:
this also fixes flashing Gluon on some Xiaomi AX3200 / Redmi AX6S

blocktrron:
@rotanid Which commit fixes that or are you referring to an earlier bump?

rotanid:
1f32774ded kernel: mtk-bmt: fix usage of _oob_read

blocktrron:
Good, to know. I assume other devices using BMT will beneift from that too.

PR #2771, opened 2023-01-28

blocktrron:
This fixes mesh on WiFi 6 devices being used with 54 Mbit/s max due to insufficient parsing of HE operation modes.


rotanid:
this also fixes flashing Gluon on some Xiaomi AX3200 / Redmi AX6S

Issue #2679, opened 2022-10-18
Bridge Port Isolation not working on UBNT ER-X


mweinelt:
The current working theory is that this may be caused by offloading of the bridge port forwarding.

We would need someone to test this theory by disabling the relevant offloading features.

See https://github.com/freifunk-gluon/gluon/pull/2600#issuecomment-1245229432

AiyionPrime:
I flashed our current nightly based on gluons master; the problem persists.
This [meshlink](https://hannover.freifunk.net/karte/#/en/map/0418d668f3b4-44d1fa59f87c) should not be there.

On the ERX I performed the following actions:

```bash
root@erx-bridge-port-isolation-debugger:~#  opkg update; opkg install ethtool
```

```bash
root@erx-bridge-port-isolation-debugger:~#  ip a
```

```console
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: dsa: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1504 qdisc fq_codel state UP qlen 1000
    link/ether b4:fb:e4:53:ae:e8 brd ff:ff:ff:ff:ff:ff
    inet6 fe80::b6fb:e4ff:fe53:aee8/64 scope link 
       valid_lft forever preferred_lft forever
3: eth0@dsa: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master br-wan state UP qlen 1000
    link/ether b4:fb:e4:53:ae:e8 brd ff:ff:ff:ff:ff:ff
4: eth1@dsa: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue master br-mesh_other state LOWERLAYERDOWN qlen 1000
    link/ether b4:fb:e4:53:ae:e9 brd ff:ff:ff:ff:ff:ff
5: eth2@dsa: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master br-mesh_other state UP qlen 1000
    link/ether b4:fb:e4:53:ae:ea brd ff:ff:ff:ff:ff:ff
[...]

PR #2773, opened 2023-01-28
Gluon v2022.1.2 Release Notes

blocktrron:
This adds Release-Notes for the Gluon v2022.1.2 bugfix-release.

This release only adds bugfixes to Gluons core functionality as well as build-process.

### DIFF
--- a/docs/releases/v2022.1.2.rst
+++ b/docs/releases/v2022.1.2.rst
@@ -1,0 +1,37 @@
+Gluon 2022.1.2
+==============
+
+Bugfixes
+--------
+
+ * Various build-errors which sporadically occur when building with a large thread-count have been fixed
+ 
+ * Android devices do not lose their IPv6 connectivity after extended idle-time
+ 
+ * The 802.11s mesh network is now using 802.11ax HE-modes when supported by hardware
+
+
+Known issues
+------------
+
+* Upgrading EdgeRouter-X from versions before v2020.1.x may lead to a soft-bricked state due to bad blocks on the NAND flash which the NAND driver before this release does not handle well.
+  (`#1937 <https://github.com/freifunk-gluon/gluon/issues/1937>`_)
+
+* The integration of the BATMAN_V routing algorithm is incomplete.
+
+  - Mesh neighbors don't appear on the status page. (`#1726 <https://github.com/freifunk-gluon/gluon/issues/1726>`_)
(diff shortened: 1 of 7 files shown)


skorpy2009 (approved):
LGTM

docs/site-example/site.conf
```diff
@@ -1,4 +1,4 @@
--- This is an example site configuration for Gluon v2022.1.1
```
AiyionPrime, line 1: If this site.conf gets bumped, the one in https://github.com/freifunk-gluon/gluon/blob/0a2eac6e5d8c1940749b0dc1b53d0c254e6295d8/contrib/ci/minimal-site/site.conf#L1
should be as well, right?
If so, we should add it to the checklist.

AiyionPrime (changes requested):
The suggested changes do look fine, thanks for the effort.

One thing this PR currently overlooks is bridgeport isolation merged in #2680, that does not only not work on the ER-X, but appears to affect the devices networking in some communities, as @herbetom tested.

I think marking it as known issue would not suffice. But maybe others differ.
We tracked part of the issue in #2679.

blocktrron:
@AiyionPrime The Bridge-Port isolation feature still needs documentation and shortcomings as well as limitaations is something i do not see happening for this bugfix release.

And to stress it the last time - Bridge Port isolation is a NOP in the DSA code which incorrectly does not fail on setting the flag, as it is supported in the switchdev layer but unsupported in DSA.

See  

https://elixir.bootlin.com/linux/latest/source/net/dsa/port.c#L914
https://elixir.bootlin.com/linux/latest/source/net/dsa/port.c#L319
https://elixir.bootlin.com/linux/latest/source/net/bridge/br_switchdev.c#L73
https://elixir.bootlin.com/linux/latest/source/drivers/net/dsa/mt7530.c#L1175

AiyionPrime (approved):

docs/releases/v2022.1.2.rst
```diff
@@ -0,0 +1,37 @@
+Gluon 2022.1.2
+==============
+
+Bugfixes
+--------
+
+ * Various build-errors which sporadously occur when building with a large thread-count have been fixed
```
rotanid, line 7: "sporadously" doesn't exist i think, please switch to "sometimes" or "sporadic"

skorpy2009: sporadically would be best fitting

rotanid (approved):

PR #2781, opened 2023-02-05
buildsystem: allow building custom buildtypes

mkg20001:
This adds GLUON_BUILDTYPE, which is an option that allows specifying an alternate
base configuration

Other additions:

- GLUON_PREFIX option to specify the filename prefix, if
  VERSION_DIST is set

- Skip gluon target-definitions if TARGET_ALL_PROFILES is set

- Allow overriding any file in targets/ by adding the file to site/
  - Allow loading the original file using include_gluon() from gluon


mkg20001:
This is a continuation of the work in https://github.com/freifunk-gluon/gluon/pull/2142

mkg20001:
to give some context about mixing gluon and vanilla/luci: this is planned to be used as a way to get respondd, mmfd, etc working on devices where there needs to be a lot of custom config that otherwise can't be done with gluon. but it can also be used in a way similar to the previous pr, where built images simply have a gluon-free custom package set.

mkg20001:
requested changes:
- gluon packages that are usable outside of gluon should be moved to packages repo, instead of hacking gluon to be usable outside of context
- instead of many custom options there should be a way to replace the lua files that are pulled in (for ex GLUON_CUSTOM_BUILDCONFIGURATION)

AiyionPrime:

Thanks for trying!

PR #2782, opened 2023-02-05
gluon-authorized-keys: add unauthorized_keys to remove access

mkg20001:


mkg20001:
revoew discussion summarized:
- the unauthorized_keys feature in it's current state is not wanted
- the goal of authorized_keys was never meant to be a mechanism for rolling out keys

the reworked version:
- groups instead of main authorized keys, which are disabled by default (similar to ffda-ssh-manager)
- /etc/dropbear/authorized_keys -> /var/authorized_keys symlink, generate from /etc/config/authorized_keys? (persistent keys) and the enabled groups
- migration that moves /etc/dropbear/authorized_keys to persistent location and replaces it with symlink to generated version. generated version will also have read-only flag set.
- persistent additions to the persistent (!) authorized_keys file should still be possible via site.conf
- there should be a webui for the groups

blocktrron:
I think this is a compromise in every aspect.

Over at https://github.com/freifunk-gluon/community-packages/pull/24 we have a IMHO better solution to manage site-keys. This allows also to differentiate between site-originated keys as well as user-added SSH keys.

In my opinion, we can merge this and work towards integrating the community package into core-gluon, deprecating the current way of adding SSH keys using site-information.

docs/features/authorized-keys.rst
```diff
@@ -13,3 +13,13 @@ If you select this package, add a list of authorized keys to ``site.conf`` like
     ...
 
 Existing keys in ``/etc/dropbear/authorized_keys`` will be preserved.
+
+If you want to remove specific keys in the future, specify them like this:::
```
blocktrron, line 17: Take note here that this will also remove SSH key which are originally added by the user.

AiyionPrime:
@mkg20001 do you intend to pursue this further? Or can this be closed?

AiyionPrime:

Thanks for trying!

PR #2783, opened 2023-02-05
firewall: switch to nftables

mkg20001:
Let's do this!

_(not only because I have a passionate hate towards iptables, but because nftables is the cool new firewall that merges all the others)_


mkg20001:
not sure how to go about ebtables. I took a quick glance and it seems there's not really anything missing from nftables that is currently being done in ebtables. If it's a good idea, I could do the rewrite of the ebtables rules.

AiyionPrime:
we'll build a firmware together in the next days; @mkg20001, @AiyionPrime

mkg20001:
IPTables migration is done, the goal is to go ahead with migrating ebtables to nftables. input is appreciated.

mkg20001:
how migrations are handled:

ebtables -> nftables:
 - delete remaining ebtables config entierly

nftables in general:
- since the snippets installed might disappear when a mesh chooses to remove a package, which could potentially break fw4 (ok turns it it's just a warning but let's keep it clean)
  - individual includes are prefied with gluon_nftables_ and are removed once no longer needed

for appending the includes I've choosen a similar style to what we already have with the webinterface elements. I hope I've found the best middleground between boilerplate and complexity.

if wanted we could extend the removal/readd to all firewall rules (or extend /lib/gluon/nftables to become /lib/gluon/firewall with nftables includes aswell as regular firewall rules)

mkg20001:
>  I hope I've found the best middleground between boilerplate and complexity.

alternative would be a module similar to what wireless does

rotanid:
well, re450 was already marked as tiny, if it's the only device...
but it's said once again, that something new is using more space rather than "we wrote from scratch and saved space by not having to deal with the old stuff" :(

mkg20001:
@rotanid nftables itself is among the biggest things, we can't really get rid of much

replacing fw4 isn't really an option either as that won't be maintainable

```
usr/lib/libnftables.so.1.1.0: 612kib
usr/lib/libnftnl.so.11.6.0: 132kib
usr/lib/libjansson.so.4.13.0: 40kib
usr/lib/libmnl.so.0.2.0: 16kib
bin/nft: 12kib
````

maybe disabling some features in nftables cli will help (not sure if that's possible)

**Comment by mkg20001** [comment]
fw4 is written in ucode, which also uses a bunch of space

```
$ du -hs ./usr/{bin,lib,share}/ucode
16K	./usr/bin/ucode
76K	./usr/lib/ucode
80K	./usr/share/ucode
```

mkg20001:
Considering the nftables error I'm currently hitting and the fact that openwrt 23 is pretty close and may have some cleanup for legacy iptables, I'd just leave this hanging and see how much things change with the new release.

But the size changes appear considerable and even if we get the tiny images to build with 23 removing leftover iptables, we'll still lose a lot of space.

One thing for consideration would be transpiling firewall4 from ucode to lua if that's automatically doable, which should give us a few hundred kb.

mkg20001:
Also I came accross this: https://github.com/openwrt/openwrt/pull/11895

This might help with space problems in general, but since mips is not supported yet it wouldn't do too much.

mkg20001:
I'd need some help enabling the right nftables modules as they seem to be missing.

mkg20001:
A potential fix for tiny would be including the minimal dnsmasq again, but this time only for tiny only. That way we should have enough space.

neocturne:
Okay, I have finally started looking at this a little bit (and rebased it onto main locally so I can test things). Some very high-level comments - I will need more time to go over the details:

- I'm not sure if there should be a gluon-nftables package at all, or if it should be part of gluon-core
- I would like to rename all the firewall packages (both `gluon-ebtables-*` and `gluon-iptables-*`) to `gluon-firewall-*`. This could be done beforehand, before switching out the actual implementation. I have also started working on this locally, but I have not rebased your implementation on top of that
- I'm unsure if I want Gluon's custom rules (in particular what was ebtables) to be written to `/etc/config/firewall` at all, or if we should have a custom init script like in `gluon-ebtables`. If we keep using UCI, every reconfigure should reset the config to a known base state before adding custom config, like it is already handled for `system` and `network` (but ultimately, I'd prefer to use UCI as little as possible)

@mkg20001 Do you have an opinion on these points?

mkg20001:
I'm currently reworking this on https://github.com/ffgraz/gluon/tree/nft-v2 and in parallel I also added some tests using https://github.com/freifunk-gluon/gluon-qemu-testlab in https://github.com/ffgraz/gluon/tree/babel-revive-next I'll create a new PR once the batman changes are working on my side.

PR #2785, opened 2023-02-10
sunxi: add devices bananapi m2 ultra and berry

LizenzFass78851:
adds support for the banana pi m2 ultra and banana pi m2 berry

PR #2786, opened 2023-02-10
sunxi: add devices bananapi m2 ultra and berry

LizenzFass78851:
adds support for the banana pi m2 ultra and banana pi m2 berry


herbetom:
I'm sorry, but Backports are only possible once device are merged into master. Therefore first #2785 has to be merged.

(Two other thins are that it's always nice to Prefix such backport PRs with something like "[v2022.1.x]" and they should be cherry picked with the reference of the merged master commit (`git cherr-pick -x`).)

AiyionPrime:
@LizenzFass78851 Thanks for your contribution.
We are doing backports via a GitHub action; no need for manually opening separate PRs.
As @herbetom said, backports **may** only happen on successfully tested and `main`-merged devices, I suggest focusing on #2785 instead.

PR #2796, opened 2023-02-24
uclient: patch: "uclient: cancel state change timeout in uclient_disconnect()"

NeoRaider:
Fixes a use-after-free seen in the autoupdater when opening a TCP connection times out (resulting in a segfault in certain conditions).

First half of patches to fix https://github.com/freifunk-gluon/packages/issues/261 (the other half applying to the autoupdater itself).

Patch has also been submitted upstream: https://patchwork.ozlabs.org/project/openwrt/patch/5191625b48c2dd44d5e094d12a12b608d1a6bae5.1677273776.git.mschiffer@universe-factory.net/


AiyionPrime (approved):

NeoRaider:
I have pushed the patch to OpenWrt upstream and cherry-picked it into the stable branches, so we will get it with the next module update.

PR #2797, opened 2023-02-25
bcm63xx: add support for TP Link TD-W8968 V3

LizenzFass78851:
adds support for the tp link td w8968 v3

  - ~[ ] Web interface~
  - [x] TFTP
  - [x] Other: CFE web recovery
    - ##### output is "sagem-f-st-2704n"
- Wired network
  - ~[ ] Association with AP must be possible on all radios~
  - ~[ ] Association with 802.11s mesh must work on all radios~
  - ~[ ] AP+mesh mode must work in parallel on all radios~
    - ##### Wifi chip not supported
- LED mapping
  - Power/system LED
      - ##### was not displayed
  - Radio LEDs
    - [ ] Should show activity
       - ##### Wifi chip not supported
  - Switch port LEDs
       - ##### only one led is displayed which only shows the activity of the wan port but not the activity of the other lan ports
- Outdoor devices only:
  - ~[ ] Added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`~
- Docs:
  - ~[ ] Added Device to `docs/user/supported_devices.rst`~

the internet speed, tested with freifunk chemnitz as an example, comes to 10 mbit/s with the tp link td w8968 v3 and that can be a little higher speed via freifunk chemnitz. it is therefore not a very fast router that unfortunately does not have wifi via gluon/openwrt but since it has a total of 4 lan ports (100 mbit/s) it is at least possible as a lan switch for the lan-based use of gluon.

### DIFF
--- a/targets/bcm63xx-generic
+++ b/targets/bcm63xx-generic
@@ -4,6 +4,6 @@ device('tp-link-td-w8968-v3', 'sagem_fast-2704n', {
 	sysupgrade = '-squashfs-cfe', 
 	sysupgrade_ext = '.bin', -- No sysupgrade image available
 
-
+aliases = {'tp-link-td-w8968-v3'},
 	broken = true, -- WiFi chip not supported
 })
--- a/targets/bcm63xx-generic
+++ b/targets/bcm63xx-generic
@@ -4,9 +4,6 @@ device('tp-link-td-w8968-v3', 'sagem_fast-2704n', {
 	sysupgrade = '-squashfs-cfe', 
 	sysupgrade_ext = '.bin', -- No sysupgrade image available
 
-	manifest_aliases = {
-		'FAST2704N', -- Upgrade from OpenWrt 19.07
-	},
 
 	broken = true, -- WiFi chip not supported
 })
--- a/targets/bcm63xx-generic
+++ b/targets/bcm63xx-generic
@@ -1,0 +1,12 @@
+include 'bcm63xx.inc'
+
+device('tp-link-td-w8968-v3', 'sagem_fast-2704n', {
+	sysupgrade = '-squashfs-cfe', 
+	sysupgrade_ext = '.bin', -- No sysupgrade image available
+
+	manifest_aliases = {
+		'FAST2704N', -- Upgrade from OpenWrt 19.07
+	},
+
+	broken = true, -- WiFi chip not supported
+})
--- a/targets/targets.mk
+++ b/targets/targets.mk
@@ -25,5 +25,6 @@ $(eval $(call GluonTarget,x86,64))
 
 ifneq ($(BROKEN),)
 $(eval $(call GluonTarget,bcm27xx,bcm2710)) # BROKEN: Untested
+$(eval $(call GluonTarget,bcm63xx,generic)) # BROKEN: Untested
 $(eval $(call GluonTarget,mvebu,cortexa9)) # BROKEN: No 11s support
 endif
--- a/targets/targets.mk
+++ b/targets/targets.mk
@@ -25,6 +25,6 @@ $(eval $(call GluonTarget,x86,64))
 
 ifneq ($(BROKEN),)
 $(eval $(call GluonTarget,bcm27xx,bcm2710)) # BROKEN: Untested
-$(eval $(call GluonTarget,bcm63xx,generic)) # BROKEN: Untested
+$(eval $(call GluonTarget,bcm63xx,generic)) # BROKEN: No 11s support
 $(eval $(call GluonTarget,mvebu,cortexa9)) # BROKEN: No 11s support
 endif
--- a/targets/bcm63xx.inc
+++ b/targets/bcm63xx.inc
@@ -1,0 +1,6 @@
+defaults {
+	factory = '-squashfs-cfe',
+	factory_ext = '.bin',
+	sysupgrade = '-squashfs-sysupgrade',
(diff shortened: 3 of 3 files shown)


targets/targets.mk
```diff
@@ -25,5 +25,6 @@ $(eval $(call GluonTarget,x86,64))
 
 ifneq ($(BROKEN),)
 $(eval $(call GluonTarget,bcm27xx,bcm2710)) # BROKEN: Untested
+$(eval $(call GluonTarget,bcm63xx,generic)) # BROKEN: Untested
```
AiyionPrime, line 28:
```suggestion
$(eval $(call GluonTarget,bcm63xx,generic)) # BROKEN: No 11s support
```

Broadcom chips are not capable of mesh. Maybe you can test, the client network is working.

targets/bcm63xx-generic
```diff
@@ -0,0 +1,12 @@
...
+device('tp-link-td-w8968-v3', 'sagem_fast-2704n', {
+	sysupgrade = '-squashfs-cfe', 
+	sysupgrade_ext = '.bin', -- No sysupgrade image available
+
+	manifest_aliases = {
+		'FAST2704N', -- Upgrade from OpenWrt 19.07
+	},
+
+	broken = true, -- WiFi chip not supported
+})
```
AiyionPrime, line 12:
```suggestion
})
```
The file lacks a newline at the end.

targets/bcm63xx-generic
```diff
@@ -0,0 +1,12 @@
+include 'bcm63xx.inc'
+
+device('tp-link-td-w8968-v3', 'sagem_fast-2704n', {
+	sysupgrade = '-squashfs-cfe', 
+	sysupgrade_ext = '.bin', -- No sysupgrade image available
```
AiyionPrime, line 5: Not sure, what you mean by this.
If there wasn't a sysupgrade image available, why bother adding these lines, and how would you have tested sysupgrades were working?

targets/bcm63xx-generic
```diff
@@ -0,0 +1,12 @@
...
+
+	manifest_aliases = {
+		'FAST2704N', -- Upgrade from OpenWrt 19.07
+	},
```
AiyionPrime, line 9:
```suggestion
```

This device, let alone the target? Were never supported in gluon. No need to add an upgrade path.

targets/bcm63xx-generic
```diff
@@ -0,0 +1,12 @@
...
+
```
AiyionPrime, line 10:
```suggestion
aliases = {'tp-link-td-w8968-v3'},
```
And fixing the tp-link reference in line three to the sagem reference you found in your tests should do the trick.

AiyionPrime (changes requested):
Hello @LizenzFass78851, thanks for your contribution.

Just to make this clear; I do strikethrough things in such PRs that do not apply to the device;
Such as radio LEDs or wireless checkboxes for devices that do not have radios at all.
I do so to get and give a better overview, what goals are yet to be met.

I you want to do that too, that's fine by me.
But you should know that you mustn't just strikethrough all the criteria your device technically could meet, but currently doesn't.

Using strikethrough on wifi is debatable at best, given you aim for `broken`.

The blinking configmode-LED is a software feature, every gluon device with software controlled LEDs should be able to meet this.

The autoupdater is mandatory on every device. I added coments how one would fix this.


Those problems I pointed out are not a complete list with what this PR would need to get merged, but a good start as a learning feedback.

That brings us to the biggest problem with this PR:
I'm not convinced it should be merged at all.

Your assessment:

> it is therefore not a very fast router that unfortunately does not have wifi via gluon/openwrt but since it has a total of 4 lan ports (100 mbit/s) it is at least possible as a lan switch for the lan-based use of gluon.
 
is right, the use would be very limited at best.

The major downside of adding this device would be adding a complete gluon target, which would increase build times for such limited use.

I'm not convinced this device is worth any of our times and hope you reconsider intending to support this :D

(If your community would somehow benefit from adding this, because you have like 100pcs of them and are all up for the slowest possible offloader experience we can give it a shot, but other than that maybe just aim for a device that you would actually like to use.)

Maybe you or others differ, so I'll leave this open for now;
but firstly not for you to improve this PR, but to decide on its actual use :)

targets/bcm63xx-generic
LizenzFass78851: if the firmware for the device is built then for some reason there is no sysupgrade image available for the said model (its chipset) because it only creates a factory image and no sysupgrade. I did this there because normally a sysupgrade image is also generated for the target, but not with exactly this chipset

AiyionPrime:
@LizenzFass78851 I really meant it when I wrote about deciding *first* whether adding this target is worthwhile. I'll not answer technical problems until this is decided.

And you shouldn't bother implementing my suggestions either. They'll not bring you closer to getting this merged in this moment.

Making a point about why adding this device is worthwhile for you/your community/gluon is worthwhile might though.

LizenzFass78851:
> And you shouldn't bother implementing my suggestions either. They'll not bring you closer to getting this merged in this moment.

you're right. it seems unsuitable for this said device for the effort of a new goal and its meaning and purpose. I will therefore close this pull request because I understand that it is not really worthwhile for a broken device and therefore add a new target. sorry for the circumstances and the time it takes.

PR #2806, opened 2023-03-01
gluon-core: fix typo in wireless upgrade script

rotanid:


AiyionPrime (approved):
This typo does only exist once in packages.

PR #2815, opened 2023-03-14
docs: fix typos

bobidle:
Noticed some typos and suggest the following changes.


docs/features/vpn.rst
```diff
@@ -125,7 +125,7 @@ Configurable Method
 """""""""""""""""""
 
 From the site configuration, fastd can be allowed to offer
-toggleable encryption in the config mode with the intent to
+togglable encryption in the config mode with the intent to
```
AiyionPrime, line 128: Toggleable/togglable is not a word with formal spelling. Once that changes, we can swap them; for now they are both equally right.

```suggestion
togglable encryption in the config mode with the intent to
```

docs/features/wired-mesh.rst
```diff
@@ -71,7 +71,7 @@ When the client role is assigned to an interface in combination with other roles
 precedence, enabling mesh but not client in the previous example.
 
 The setup/config-mode interface is every interface with the role ``client`` which makes removing
-it from interfaces not only unnecessary, but generally unrecommended.
+it from interfaces not only unnecessary, but generally not recommended.
```
AiyionPrime, line 74: Why would this be wrong?
https://en.wiktionary.org/wiki/unrecommended

```suggestion
it from interfaces not only unnecessary, but generally unrecommended.
```

AiyionPrime (changes requested):
Thanks for the suggestions, I agree with most of them.
Please have a look at the others :)

docs/features/wired-mesh.rst
stweil: Try "unrecommended" with https://www.deepl.com/write. For me "not recommended" sounds better, but I am not a native English speaker.

docs/features/vpn.rst
stweil:
```suggestion
encryption which can be toggled in the config mode with the intent to
```

AiyionPrime: I feel like this already is precise, does not need changing and therefore should not; but maybe others differ.

docs/features/wired-mesh.rst
AiyionPrime: No, thanks.
https://en.wiktionary.org/wiki/unrecommended
https://www.dict.cc/englisch-deutsch/unrecommended.html
https://www.collinsdictionary.com/dictionary/english/unrecommended

bobidle:
Reverted "togglable" and "not recommended"

AiyionPrime (approved):

lemoer (approved):

PR #2817, opened 2023-03-19
ath79-generic: fix WS-AP3705i autoupdater name

herbetom:
The device was introduced in #2332 and merged in 102a4b9350c925cbbe15fdead43a370c445eacf8.

It appears that the autoupdater name wasn't correct and devices therefore don't receive updates.

```
root@64295-ggw3-20b399bb366f-132:~# lua -e 'print(require("platform_info").get_image_name())'
enterasys-ws-ap3705i
```

cc @blocktrron


blocktrron:
Looks correct.

PR #2818, opened 2023-03-19
Add Support for D-Link DAP-X1860 A1

RolandoMagico:
  - [ ] Web interface
  - [ ] TFTP
  - [x] Other: Keep reset button pressed during plug-in. Web Recovery Updater is at 192.168.0.50 (Netmask 255.255.255.0). Upload factory.bin, confirm flashing (seems to work best with Chromium-based browsers)
- Wired network
      - otherwise the first port should be delcared as WAN, all other ports LAN
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [ ] Should show activity
  - Switch port LEDs
- Outdoor devices only:
- Docs:


AiyionPrime:
@RolandoMagico "next" is not the right branch in order to add device support for gluon.

Please open the PR against master.

Issue #2824, opened 2023-03-21
Web status page no longer shows wireless mesh info with newer OpenWrt


stweil:
### Bug report

Firmware which is based on Gluon next or other branches with newer OpenWrt no longer
shows wireless mesh information in the web status page of a node.

**What is the expected behaviour?**

That's a regression, and the previous behaviour is expected.

**Gluon Version:**

Tested with firmware based on branch `next`.

**Site Configuration:**

FFRN

**Custom patches:**

None.

stweil:
See also discussion for issue #2822.

The regression might be caused by OpenWrt commit https://github.com/openwrt/openwrt/commit/6eeb5d45649ffa591b06d2ccd6a43e47939b6b18.

blocktrron:
We do not track bugs with Gluons next branch. `next` is an unsupported tech-demo.

PR #2826, opened 2023-03-23
gluon-status-page: fix status display of wireless mesh for recent openwrt

stweil:
Instead of checking for the deprecated sysfs entry `wireless` which no longer exists when running newer versions of openwrt, testing for `phy80211` can be used and works for old and new versions of openwrt.

### DIFF
--- a/package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
+++ b/package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
@@ -67,7 +67,7 @@
 			iface = lower:sub(pattern:len())
 		end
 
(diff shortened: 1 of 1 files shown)


stweil:
Related issue with more details: #2824. I tested the patch on a router running Gluon next.

rotanid:
have you also tested the fix on a node running Gluon master, as your PR is for master?

stweil:
> have you also tested the fix on a node running Gluon master, as your PR is for master?

`master` uses the same openwrt as `v2022.1.x` (an older revision) which provides `wireless` in the sysfs, but also works with `phy80211`.

The PR can be cherry-picked for any branch of `gluon`.

PR #2828, opened 2023-03-25
docs: use only one example email address in site example

bobidle:
Noticed different email addresses usages in the site example.

This commit changes the addresses to  `keys@alpha-centauri.freifunk.net` as this address is also used in the `en.po` and `fr.po` (and also in the `de.po`).


AiyionPrime (commented):
I don't feel like this would improve the docs, it's an arbitrary example.
It does show: you should put your communities email addresses there.

Maybe another developer feels differently about this.

christf:
Using different addresses is actually on purpose. Most communities I am aware of that do manual key placement, use one address to approach them for first contact and another one with a different mailbox just to keep the keys. I do appreciate and welcome contributions. In this instance, unfortunately I do not think it is an improvement.
This prompts me to wonder though, whether we should put a lantern on this and explain in the text (new communities need to take many decisions, let's make it as easy for them as possible to find a structure that has been proven to work in other communities).

AiyionPrime:
Not looked into it, but if the other translations do use the same address, we could split them up as well.

bobidle:
https://github.com/freifunk-gluon/gluon/blob/master/docs/site-example/i18n/de.po:

AiyionPrime:
I think having `keys@alpha-centauri.freifunk.net` and `kontakt@alpha-centauri.freifunk.net` in all translations would be favorable.

bobidle:
Changed the address to `kontakt@alpha-centauri.freifunk.net`. Found and fixed two minor issues with a missing white space and a missing `<p>` tag.

docs/site-example/i18n/de.po
```diff
@@ -61,7 +61,7 @@ msgstr ""
 "Freifunk-Netz aufbauen, wenn andere Freifunk-Knoten in WLAN-Reichweite sind.</p>"
 "<p>Bitte schicke uns eine E-Mail mit dem Namen deines Knotens "
 "(<em><%= pcdata(hostname) %></em>) und ein paar Informationen an <a href="
-"\"mailto:freifunk-keys@lists.in-kiel.de?"
+"\"mailto:kontakt@alpha-centauri.freifunk.net?"
```
AiyionPrime, line 64:
```suggestion
"\"mailto:keys@alpha-centauri.freifunk.net?"
```

docs/site-example/i18n/en.po
```diff
@@ -55,17 +55,17 @@ msgstr ""
 "<p>You have selected <strong>not</strong> to use the mesh VPN. "
 "Your node will only be able to connect to the Freifunk network if other nodes "
 "in reach already have a connection.</p>"
-"Please send an e-mail with the name of your node "
+"<p>Please send an e-mail with the name of your node "
 "(<em><%=pcdata(hostname)%></em>) and some additional information to "
-"<a href=\"mailto:keys@alpha-centauri.freifunk.net?subject="
+"<a href=\"mailto:kontakt@alpha-centauri.freifunk.net?subject="
```
AiyionPrime, line 60:
```suggestion
"<a href=\"mailto:keys@alpha-centauri.freifunk.net?subject="
```

AiyionPrime (changes requested):
@bobidle Thanks for the update. But @grische is right,

> Using different addresses is actually on purpose.

All three translations should use the keys@-address for key purposes and the kontakt@-address for regular communication in order to show these fields do have nothing in common.

bobidle:
Sorry i am confused by your suggested changes (review).

Please check the translation files:
https://github.com/bobidle/gluon/blob/site-example/docs/site-example/i18n/de.po
https://github.com/bobidle/gluon/blob/site-example/docs/site-example/i18n/en.po

At the moment `keys@` is used for `gluon-config-mode:pubkey` (key propose) and `kontakt@` is used for `gluon-config-mode:novpn` (regular communication).

Your suggestions request to use `keys@` to `gluon-config-mode:novpn` (regular communication).

docs/site-example/i18n/de.po
bobidle: This message block is `gluon-config-mode:novpn` (regular communication), as this is non-key communication, the kontakt-address should be used.

AiyionPrime:
Indeed. Use keys@ in the first occurrence in the block `gluon-config-mode:pubkey` and use kontakt@ for the greeter in both `gluon-config-mode:pubkey` and `gluon-config-mode:novpn`.

AiyionPrime:
Sorry you are right, I was under the impression the `gluon-config-mode:pubkey` section had two mail links, one for key submission and one for regular contact.
And that `gluon-config-mode:novpn` had only the latter. My bad.

AiyionPrime:
@bobidle Thanks for your contribution!

PR #2833, opened 2023-03-30
docs: readd TP-Link TD-W8970 v1

Djfe:
This device is already supported.
It's a lantiq device that was entered as ar71xx
in to the list of supported device and therefore
removed before the release of Gluon 22.

This is a backport of
95e5d382ecc65ed68262b90e2fc5f2d8a13d464c


AiyionPrime:
Not sure a backport of this is necessary, but thanks.
We do have an action though, which generates backports in a proper manner;
so I'll close this in favor of #2834.

Feel free to have a look at it and comment, whether the action missed something.

Djfe:
Ah, I didn't know that. Thanks for enlightening me

PR #2837, opened 2023-04-01
[docs] improve build dependencies and docs

maurerle:
Target x86-64 does not build otherwise and does not show useful messages concerning the missing package either.

This issue was discussed on #gluon on IRC and was found by comparing the install docs to the `contrib/docker/Dockerfile`.

This PR updates the install dependencies and docs
    * add missing libelf-dev dependency (to build x86-64)
    * remove subversion dependency
    * add rsync dependency
    * add ecdsautils to docs (to sign firmware)

### DIFF
--- a/docs/user/getting_started.rst
+++ b/docs/user/getting_started.rst
@@ -25,18 +25,20 @@ An example configuration can be found in the Gluon repository at *docs/site-exam
 Dependencies
 ------------
 To build Gluon, several packages need to be installed on the system. On a
-freshly installed Debian Stretch system the following packages are required:
+freshly installed Debian Bullseye system the following packages are required:
 
 * `git` (to get Gluon and other dependencies)
-* `subversion`
 * `python3`
 * `build-essential`
+* `ecdsautils` (to sign firmware, see `contrib/sign.sh`)
 * `gawk`
 * `unzip`
 * `libncurses-dev` (actually `libncurses5-dev`)
 * `libz-dev` (actually `zlib1g-dev`)
 * `libssl-dev`
+* `libelf-dev` (to build x86-64)
 * `wget`
+* `rsync`
 * `time` (built-in `time` doesn't work)
 * `qemu-utils`
 
--- a/contrib/actions/install-dependencies.sh
+++ b/contrib/actions/install-dependencies.sh
@@ -3,6 +3,6 @@
 set -e
 
 apt-get -y update
-apt-get -y install git subversion build-essential python3 gawk unzip libncurses5-dev zlib1g-dev libssl-dev wget time qemu-utils
+apt-get -y install git build-essential python3 gawk unzip libncurses5-dev zlib1g-dev libssl-dev libelf-dev wget rsync time qemu-utils
 apt-get -y clean
 rm -rf /var/lib/apt/lists/*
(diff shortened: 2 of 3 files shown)


Djfe:
Nice find!

The [wiki](https://openwrt.org/docs/guide-developer/toolchain/install-buildsystem#debianubuntu) also lists these in addition to the ones here:
clang
flex
bison
g++
gcc-multilib
g++-multilib
gettext
python3-distutils
rsync
file

Maybe the docker needs to be adjusted as well to install these, too? (and remove subversion probably)
the [Dockerfile](https://github.com/freifunk-gluon/gluon/blob/master/contrib/docker/Dockerfile) is missing all of the above except rsync and file.
So rsync and file are definitely missing from getting_started.rst still.

docs/user/getting_started.rst
```diff
@@ -25,17 +25,18 @@ An example configuration can be found in the Gluon repository at *docs/site-exam
 Dependencies
 ------------
 To build Gluon, several packages need to be installed on the system. On a
-freshly installed Debian Stretch system the following packages are required:
+freshly installed Debian Bullseye system the following packages are required:
 
 * `git` (to get Gluon and other dependencies)
-* `subversion`
```
NeoRaider, line 31: If subversion is removed here, it should also be dropped in `install-depenencies.sh` and the `Dockerfile`.

Also, these changes aren't mentioned in the commit message.

docs/user/getting_started.rst
```diff
@@ -25,17 +25,18 @@ An example configuration can be found in the Gluon repository at *docs/site-exam
...
 * `python3`
 * `build-essential`
+* `ecdsautils`
```
NeoRaider, line 33: I'm unsure whether ecdsautils should be considered a dependency. Maybe with a note that this is only needed for `contrib/sign.sh`?

NeoRaider (changes requested):

NeoRaider:
> Nice find!
> 
> The [wiki](https://openwrt.org/docs/guide-developer/toolchain/install-buildsystem#debianubuntu) also lists these in addition to the ones here:
> 
> ```
> clang
> flex
> bison
> g++
> gcc-multilib
> g++-multilib
> gettext
> python3-distutils
> rsync
> file
> ```
> 
> Maybe the docker needs to be adjusted as well to install these, too? (and remove subversion probably) the [Dockerfile](https://github.com/freifunk-gluon/gluon/blob/master/contrib/docker/Dockerfile) is missing all of the above except `rsync` and `file`. So `rsync` and `file` are definitely missing from getting_started.rst still.

A lot of these are already pulled in by `build-essential`, and I don't think the multilib toolchain or clang are actually necessary. I'm not sure what `file` is used for, but `rsync` should definitely be added to the docs as well.

It's also unclear to me why `libelf-dev` is only needed on x86-64.

docs/user/getting_started.rst
maurerle: fixed, thanks!

docs/user/getting_started.rst
maurerle: I added a note that this is only needed to sign firmware.
I think it makes it easier to understand the autoupdater procedure if this dependency is already listed here.

maurerle:
What the specific reason for the dependency on libelf-dev is, I don't really know, but I guess it has todo with the UEFI/x86-64 bootloader?

Practically, I know that my gluon build (v2022.1.3) of all targets except for x86-64 worked fine without libelf-dev, and it worked for x86-64 too after I installed the dependency.

AiyionPrime (approved):

NeoRaider (approved):

PR #2838, opened 2023-04-02
contrib: Dockerfile: actually create gluon user directory

skorpy2009:


NeoRaider:
I assume this is already handled by the VOLUME? I don't think it makes much sense to copy the skel files into a directory that is mounted over anyways.

AiyionPrime:
@skorpy2009 is there a problem you intend to solve with this measure?
If not and neoraider is right, we could close this :)

AiyionPrime:
@skorpy2009 is there a problem you intend to solve with this measure?
If not and neoraider is right, we could close this :)

skorpy2009:
@NeoRaider @AiyionPrime currently if you use this container and want to do anything with the user, for example you need git config it breaks because there is no gluon folder

NeoRaider:
The Gluon build directory is supposed to be mounted into the container at `/gluon`.

NeoRaider:
It is unclear to me if the behaviors of Docker and Podman differ here (I only have Podman on my system), but Podman creates an empty volume and mounts it to `/gluon` when starting the container. I would expect the same to happen with Docker, as long as you don't override the mount at `/gluon` on the command line. I would also expect this not to change when another Dockerfile is based on the `gluon/contrib/docker` container.

So an explanation in what situation it is even possible to end up with no `/gluon` when using our Dockerfile would be great.

That being said, I looked at the Dockerfile reference again and found that the contents under the directory specified as a `VOLUME` will end up as initial content for the new volume by default. So maybe this change is not useless after all, as it will install skel contents into the volume, but that is not what the commit message is saying.

neocturne:
Fixed by #2975

PR #2840, opened 2023-04-06
[v2021.1.x] patches: enable uart for Joy-IT OR750i

### LINKED ISSUE #2839: 2021.1.{1,2} Joy-IT OR750i can't sysupgrade
### Bug report

**What is the problem?**

Sysupgrades immediately reboot the device, and silently fail.
The autoupdater therefore does not work either.

We encountered the problem, when we were looking for devices, that have not yet updated to `2022.1.3`, yet.

Sysupgrades actually upgrade the device.

**Gluon Version:**
[v2021.1.2](https://github.com/freifunk-gluon/gluon/tree/v2021.1.2)

The problem is only apparent on `2021.1.x`, not on `master` or newer branches.

**Site Configuration:**
[vH25](https://github.com/freifunkh/site/commit/2538a3f56f2ba1fd6a9c93dfab5c171b68d4ab56)

**Custom patches:**

- [0001-fix-when-vlan-crc-is-00000000.patch](https://github.com/freifunkh/site/blob/2538a3f56f2ba1fd6a9c93dfab5c171b68d4ab56/patches/0001-fix-when-vlan-crc-is-00000000.patch)
- [0002-patches-uradvd-change-ip-lifetimes-to-150-300-secon](https://github.com/freifunkh/site/blob/2538a3f56f2ba1fd6a9c93dfab5c171b68d4ab56/patches/0002-patches-uradvd-change-ip-lifetimes-to-150-300-second.patch)
- [0003-added-TL-WR841ND-N-8M-and-16M-Variants-migration-to-ath79.patch](https://github.com/freifunkh/site/blob/2538a3f56f2ba1fd6a9c93dfab5c171b68d4ab56/patches/0003-added-TL-WR841ND-N-8M-and-16M-Variants-migration-to-ath79.patch)

AiyionPrime:
> resolves silent sysupgrade failure

The missing uart blocks sysupgrade from doing its job.
Why that is, I haven't looked up yet, but @blocktrron confirmed the reason.

master and newer branches are not affected, the issue was a disabled attribute in a backported patch we had in gluon `2021.1.x`.

resolves #2839

### DIFF
--- a/patches/openwrt/0027-ath79-enable-uart-for-Joy-IT-or750i.patch
+++ b/patches/openwrt/0027-ath79-enable-uart-for-Joy-IT-or750i.patch
@@ -1,0 +1,19 @@
+From: Jan-Niklas Burfeind <git@aiyionpri.me>
+Date: Wed, 5 Apr 2023 22:53:59 +0200
+Subject: ath79: enable uart for Joy-IT or750i
+
+diff --git a/target/linux/ath79/dts/qca9531_joyit_jt-or750i.dts b/target/linux/ath79/dts/qca9531_joyit_jt-or750i.dts
+index 042348d82c479c097f534f0bf2bb383ace55e02d..76f863e9d3e92a1edfe4b7351376a7bb0d8e5bf5 100644
+--- a/target/linux/ath79/dts/qca9531_joyit_jt-or750i.dts
++++ b/target/linux/ath79/dts/qca9531_joyit_jt-or750i.dts
+@@ -116,6 +116,10 @@
+ 	status = "okay";
+ };
+ 
++&uart {
++	status = "okay";
++};
++
+ &wmac {
+ 	status = "okay";
+ 


AiyionPrime:
I've built and tested the patched image yesterday, one can now send commands via the serial console and sysupgrading works without issues again.

blocktrron:
Why does the CI break here?

AiyionPrime:
I'd say not related to the PR.

```console
Package python is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  2to3 python2-minimal python2 dh-python python-is-python3

E: Package 'python' has no installation candidate
```

patches/openwrt/0027-ath79-enable-uart-for-Joy-IT-or750i.patch
```diff
@@ -0,0 +1,19 @@
+From: Jan-Niklas Burfeind <git@aiyionpri.me>
+Date: Wed, 5 Apr 2023 22:53:59 +0200
+Subject: ath79: enable uart for Joy-IT or750i
+
```
blocktrron, line 4: Please add a short description of the problem to the patch file to have at hand when re-ordering the patches.

AiyionPrime:
duplicate of #2512

PR #2841, opened 2023-04-06
docs: add known issue on OR750i sysupgrade error

### LINKED ISSUE #2839: 2021.1.{1,2} Joy-IT OR750i can't sysupgrade

AiyionPrime:
> affected versions:
> - v2021.1.1
> - v2021.1.2

I think this PR could use a backport to both `v2021.1.x` and `v2022.1.x`.
If you disagree, just remove the labels prior to merging this.

resolves #2839

### DIFF
--- a/docs/releases/v2021.1.1.rst
+++ b/docs/releases/v2021.1.1.rst
@@ -40,6 +40,9 @@ Bugfixes
 Known issues
 ------------
 
+* Sysupgrade does not work on JOY-IT OR750i, devices running this release need to be recovered with an image newer than v2021.1.2 manually.
+  (`#2839 <https://github.com/freifunk-gluon/gluon/issues/2839>`_)
+
 * Upgrading EdgeRouter-X from versions before v2020.1.x may lead to a soft-bricked state due to bad blocks on the NAND flash which the NAND driver before this release does not handle well.
   (`#1937 <https://github.com/freifunk-gluon/gluon/issues/1937>`_)
 
--- a/docs/releases/v2021.1.2.rst
+++ b/docs/releases/v2021.1.2.rst
@@ -103,6 +103,9 @@ Other improvements
 Known issues
 ------------
 
+* Sysupgrade does not work on JOY-IT OR750i, devices running this release need to be recovered with an image newer than v2021.1.2 manually.
+  (`#2839 <https://github.com/freifunk-gluon/gluon/issues/2839>`_)
+
 * Upgrading EdgeRouter-X from versions before v2020.1.x may lead to a
   soft-bricked state due to bad blocks on the NAND flash which the NAND driver
   before this release does not handle well.


blocktrron:
I don't like altering the content of release-notes after a release is done. NACK from me.

AiyionPrime:
?

We do now know about an Issue for the Release. Where do you think would be the best place to leave that info?
If you meant the backport, then just remove the to labels?

AiyionPrime:
Similarly done in https://github.com/freifunk-gluon/gluon/pull/2108, where we removed a reference to an ath9k issue and added the ER-X notes.

blocktrron:
@AiyionPrime I can not re-construct the context for #2108 anymore. Case in point is that keeping release-notes in sync for all issues that may or may not arise on a release between branches is a lot of work. After all, the reference for the issue should be the issue-tracker.

AiyionPrime:
From a users/community maintainers perspective I can say that I've looked into the known issues before releasing a firmware.
We built firmware with a delta of roughly 2y to gluons master;
I have expected to read about 'all' major known (unfixed) issues for this version when I opened readthedocs, not just the ones that were known when the release was created.
But if we just don't do that, thats fine with me as well.

I think the issue tracker is a fine place to discuss ongoing resolveable issues.
The issue that two of our point releases are effectively dead ends, is not resolveable imho. Users have to reflash their device if they flashed either of those firmwares.
And just having another issue alongside 105 others is just not cutting it for me, I think.

blocktrron:
I stil ldon't agree here. We end up

a) With a notice in the issue-tracker
b) a forward-notice in patched release-notes
c) a backwards-notice in upcoming release-notes

We did not add a forward-notice with the ER-X badblocks issue on upgrade, thus I'd like to continue wiht this approach. Looking at multi.meshviewer.org, the number of affected communities is not very big.

AiyionPrime:
Maybe we can discuss our post-release known-issue tracking on tuesday.

Issue #2827, opened 2023-03-23
ath79-generic: out of memory on EAP225 outdoor v1


lemoer:
### Bug report

**What is the problem?**

The router goes OOM randomly.

<details>
<summary>Click to expand</summary>

```


U-Boot 1.1.4--LSDK-10.2-00082-4 (May 20 2021 - 16:43:27)

board956x - Dragonfly 1.0DRAM:  
sri
ath_ddr_initial_config(287): (ddr2 init)
ath_ddr_initial_config: DDR_EMR2_ADDRESS=0x180000bc, EMR2 value=0x80
ath_sys_frequency: ref_clk 25000000
ath_sys_frequency: cpu 775 ddr 650 ahb 258
Tap values = (0x10, 0x10, 0x10, 0x10)
128 MB
Top of RAM usable for U-Boot at: 88000000
Reserving 165k for U-Boot at: 87fd4000
Reserving 192k for malloc() at: 87fa4000
Reserving 44 Bytes for Board Info at: 87fa3fd4
Reserving 36 Bytes for Global Data at: 87fa3fb0
Reserving 128k for boot params() at: 87f83fb0
Stack Pointer at: 87f83f98
Now running in RAM - U-Boot at: 87fd4000
Flash Manuf Id 0x1c, DeviceId0 0x70, DeviceId1 0x18
flash size 16MB, sector count = 256
Flash: 16 MB
*** Warning - bad CRC, using default environment

In:    serial
Out:   serial
Err:   serial
Setting 0x181162c0 to 0x40802100
Hit Ctrl+B to stop autoboot:  2  1  0 
Loading .text @ 0x80060000 (2223924 bytes)
## Starting application at 0x80060000 ...



OpenWrt kernel loader for AR7XXX/AR9XXX

Copyright (C) 2011 Gabor Juhos <juhosg@openwrt.org>

Decompressing kernel... done!

Starting kernel at 80060000...


[    0.000000] Linux version 5.10.161 (jenkins@ffh-build) (mips-openwrt-linux-musl-gcc (OpenWrt GCC 11.2.0 r19712+7-81388f74b5) 11.2.0, GNU ld (GNU Binutils) 2.37) #0 Sat Jan 1 09:09:13 2022
[    0.000000] printk: bootconsole [early0] enabled
[    0.000000] CPU0 revision is: 00019750 (MIPS 74Kc)
[    0.000000] MIPS: machine is TP-Link EAP225-Outdoor v1
[...]

PR #2848, opened 2023-04-12
ath79-generic: switch Wave2 firmware to -ct

herbetom:
This partially reverts commit 22c47df2423e6149c4b37d5f70741adcf3365f5d.

Devices in ath79-generic like the TP-Link EAP225-Outdoor v1 are really unstable with the non -ct Wave2 firmware and regulary crash with 100% memory consumption when only a handful devices are connected via 5 GHz.

closes freifunk-gluon/gluon#2827

Both Variants aren't without issues, but i feel like this could be a bit less interruptive. The best approach would be probably to just disable the 5 GHz Radios as a fix seems unlikely.

### DIFF
--- a/targets/ath79-generic
+++ b/targets/ath79-generic
@@ -35,13 +35,7 @@ local ATH10K_PACKAGES_SMALLBUFFERS_QCA9887 = {
 	'-ath10k-firmware-qca9887-ct',
 }
 
-local ATH10K_PACKAGES_QCA9888 = {
-	'kmod-ath10k',
-	'-kmod-ath10k-ct',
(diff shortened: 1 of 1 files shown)


rotanid:
@herbetom reverting to the old state seems sane, as the commit was only meant for ipq40xx anyway.

regarding your comment about disabling 5ghz radios: this would mean adding the BROKEN flag in gluon, as other similar devices already have, e.g. Archer C25

AiyionPrime:
Weird. Only received the notification for this after opening the PR.
https://github.com/freifunk-gluon/gluon/pull/2879

PR #2860, opened 2023-04-16
libbatadv: Update UAPI from kernel 6.2

jluebbe:
`batadv_gw_modes` and `BATADV_ATTR_GW_MODE` were missing from the local version, so update it from upstream.


jluebbe:
This is missing the corresponding update of the genl policy table.

PR #2862, opened 2023-04-16
scripts/container.sh: support podman on macOS

stweil:
The default user and group IDs on macOS are different from those used on Linux, so add the expected ones explicitly.

### DIFF
--- a/scripts/container.sh
+++ b/scripts/container.sh
@@ -12,7 +12,7 @@ TAG="gluon:${BRANCH:-latest}"
 if [ "$(command -v podman)" ]
 then
 	podman build -t "${TAG}" contrib/docker
-	podman run -it --rm --userns=keep-id --volume="$(pwd):/gluon" "${TAG}"
+	podman run -it --rm --userns=keep-id:uid=1000,gid=1000 --volume="$(pwd):/gluon" "${TAG}"
 elif [ "$(command -v docker)" ]
 then
 	docker build -t "${TAG}" contrib/docker


AiyionPrime:
@stweil maybe you could start a discussion about your goals and approaches to achieve them with @hafu, who attempted sth. similar in #2868.

stweil:
The goal is very simple: the script should work.

The container creates a user "gluon". That user is the first one in a Linux environment which assigns UID 1000 and GID 1000. "gluon" must have r/w access to the gluon directories from outside. This can be achieved by mapping UID and GID of the user who is running the script to 1000 inside the container. See [documentation for `--userns`](https://docs.podman.io/en/latest/markdown/options/userns.container.html) for more information.

Without my patch a user with a UID != 1000 cannot use the script because then the container's "gluon" user would not have the same rights like the outside user. This is always a problem on macOS, but I expect that there would also be a similar problem on Linux.

NeoRaider:
This change makes sense, but looking at the man page, it seems that the extended keep-id syntax was only introduced recently, breaking support for Podman on Ubuntu 22.04 and Debian Stable.

I managed to get a similar effect by replacing `--userns=...` with `--uidmap=1000:0:1 --uidmap=0:1:999 --gidmap=1000:0:1 --gidmap=0:1:999`. Not sure if there's a better solution...

NeoRaider:
I actually like the podman-specific solution better than #2868, as it doesn't require building the container specifically for a single UID/GID.

NeoRaider:
Hmm, is there any reason why we need the build user to be in `/etc/passwd` at all? The `--userns=keep-id` should work fine if `/gluon` is mounted from outside of the container, even if it doesn't match the original owner of `/gluon` in the container image. So I'm wondering what issue this PR actually tried to solve.

stweil:
That's correct. Removing the `gluon` user from `Dockerfile` and using `--userns=keep-id` works fine and looks like the simplest fix. I updated by commit and reopen the pull request.

PR #2868, opened 2023-04-18
Support non default UID/GID for build container

hafu:
Set the current UID and GID of the executing user as build arguments for the build of the container, so that environments with other UIDs and GIDs than 1000 are supported.

### DIFF
--- a/contrib/docker/Dockerfile
+++ b/contrib/docker/Dockerfile
@@ -1,5 +1,8 @@
 FROM debian:bullseye-slim
 
+ARG DUID=1000
+ARG DGID=1000
+
 ARG DEBIAN_FRONTEND=noninteractive
 RUN apt-get update && apt-get install -y --no-install-recommends \
     ca-certificates \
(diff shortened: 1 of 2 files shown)


AiyionPrime:
Please suggest and discuss your approach in #2868 where @stweil attempted a similar goal.

Closing as duplicate.

stweil:
My pull request is #2862.

stweil (commented):
I suggest to merge this pull request because it not only addresses podman (also for older versions) but also docker.

AiyionPrime:
> @NeoRaider, thank you for pointing out that my change requires a recent podman and won't work with Debian stable or Ubuntu LTS. @AiyionPrime, I suggest to merge PR #2868 which should work with all current podman versions and which also addresses docker. Therefore I close my own pull request.

_Originally posted by @stweil in https://github.com/freifunk-gluon/gluon/issues/2862#issuecomment-1514692702_

neocturne:
Fixed by #2975

PR #2870, opened 2023-04-19
Dockerfile: don't create gluon user (fixes support for UID != 1000)

stweil:
The new user `gluon` gets UID 1000 and GID 1000 and cannot write to the mounted gluon directory unless the calling user has the same UID.

On Linux only the first non admin user gets UID 1000, so docker and podman fail for all other users.

On macOS there is typically no user with UID 1000, so docker and podman fail for all users.

When no `gluon` user is created, a default user with the same UID and GID as the calling user is used and everything works fine.

### DIFF
--- a/scripts/container.sh
+++ b/scripts/container.sh
@@ -16,7 +16,7 @@ then
 elif [ "$(command -v docker)" ]
 then
 	docker build -t "${TAG}" contrib/docker
-	docker run -it --rm --volume="$(pwd):/gluon" "${TAG}"
+	docker run -it --rm --user "$(id -u):$(id -g)" --volume="$(pwd):/gluon" "${TAG}"
 else
 	1>&2 echo "Please install either podman or docker. Exiting" >/dev/null
 	exit 1
--- a/contrib/docker/Dockerfile
+++ b/contrib/docker/Dockerfile
@@ -29,8 +29,5 @@ RUN mkdir /tmp/ec &&\
     mv bin/ec-linux-amd64 /usr/local/bin/editorconfig-checker &&\
     rm -rf /tmp/ec
 
-RUN useradd -d /gluon gluon
-USER gluon
-
 VOLUME /gluon
 WORKDIR /gluon


stweil:
@mweinelt, are you the author of `Dockerfile`? If yes: is the `gluon` user required, or can it be removed?

mweinelt:
The build should simply not run as root, not even in a container.

Instead https://docs.podman.io/en/latest/markdown/options/userns.container.html#userns-mode should be used with podman.

stweil:
@mweinelt, that was my first solution (#2862), but it requires a very recent podman, so does not work with Debian stable for example.

NeoRaider:
Let me check a few things, I also have a branch for improving this, but I still need to verify that my approach works in both Docker and Podman.

Also note that I don't think the commit message of this PR is correct. The Docker image should already work fine for UIDs != 1000. The created passwd entry will be wrong, but that shouldn't matter?

stweil:
> The Docker image should already work fine for UIDs != 1000.

Did you test that? In my test on a Debian host it does not.

stweil:
@mweinelt, I added a commit which changes the UID and GID in docker. It no longer runs as root.

scripts/container.sh
```diff
@@ -16,7 +16,7 @@ then
 elif [ "$(command -v docker)" ]
 then
 	docker build -t "${TAG}" contrib/docker
-	docker run -it --rm --volume="$(pwd):/gluon" "${TAG}"
+	docker run -it --rm --user $(id -u):$(id -g) --volume="$(pwd):/gluon" "${TAG}"
```
hafu:
```suggestion
	docker run -it --rm --user "$(id -u)":"$(id -g)" --volume="$(pwd):/gluon" "${TAG}"
```

At least shellcheck throws a warning in this part. (_Quote this to prevent word splitting_)

hafu (commented):
This seems to work for podman and docker. The only downside is, that commands in the container are run as root. I don't know if it could be a problem.

hafu:
I suggest to add `--userns=host`  to the `docker run` command. If [user namespaces](https://docs.docker.com/engine/security/userns-remap/) are enabled in Docker we run into permission issues with the bind mount. `--userns=host` disables the remapping.

NeoRaider:
> Podman does not work with the current code and UID != 1000. I noticed that on macOS, and that was my initial motivation to look for a fix.

Ah, I wasn't aware of this interaction between `--userid=keep-id` and `USER`.

Removing the `USER` isn't a nice solution though, as it will break other usecases of the container image that don't involve mounting into the container. Instead, we should not touch the Dockerfile at all (except for the bash prompt maybe) and instead add the `--user` option to the `podman run` command as well to override the `USER` setting.

neocturne:
Fixed by #2975

PR #2879, opened 2023-04-24
[Backport v2022.1.x] ath79-generic: switch Wave2 firmware to -ct

AiyionPrime:
> This partially reverts commit 22c47df2423e6149c4b37d5f70741adcf3365f5d.
> 
> Devices in ath79-generic like the TP-Link EAP225-Outdoor v1 are really unstable with the non -ct Wave2 firmware and regulary crash with 100% memory consumption when only a handful devices are connected via 5 GHz.
> 
> closes freifunk-gluon/gluon#2827
> 
> (cherry picked from commit 4badb2702edd396109038da129227413795f7a49)

### DIFF
--- a/targets/ath79-generic
+++ b/targets/ath79-generic
@@ -14,13 +14,7 @@ local ATH10K_PACKAGES_QCA9887 = {
 	'-ath10k-firmware-qca9887-ct',
 }
 
-local ATH10K_PACKAGES_QCA9888 = {
-	'kmod-ath10k',
-	'-kmod-ath10k-ct',
-	'-kmod-ath10k-ct-smallbuffers',
-	'ath10k-firmware-qca9888',
-	'-ath10k-firmware-qca9888-ct',
-}
+local ATH10K_PACKAGES_QCA9888 = {}
 
 -- ALFA NETWORK
 


herbetom:
Since the switch to -ct the EAP225-Outdoor v1 hasn't experienced the reboots previously seen with the no -ct variant.

However, even with the -ct variant, something is not stable and the node doesn't seem to have 5 GHz connections after a while. This hoewer doesn't seem to affect the rest of the node's operation (no reboot and there still are 2,4 GHz connections).

I see three possible approaches (not sorted):
1. leave it as it is, recommend people with frequent reboots to turn off 5 GHz.
2. switch to -ct and see what happens. As long as it fails really "silently" it's not great, but also less annoying than constant OOM events.
3. end 5 GHz support on the affected devices

I haven't looked more closely at the situation when the node no longer has 5 Ghz connections with the -ct firmware. This means I haven't looked at the logs, I haven't looked at what this looks like on a client, and I haven't looked at how to detect and fix this condition. And i can't tell whether i'll find the time to do so either.

AiyionPrime:
@bschelm is currently running [vH31](https://hannover.freifunk.net/wiki/Freifunk/FirmwareReleases#vH31) ([v2022.1.3](https://github.com/freifunk-gluon/gluon/releases/tag/v2022.1.3)) on the EAP225v1 devices, as they are unusable with non-ct and the device run without problems with ct.
We have not experienced problems with failing 5GHz on those devices, yet; but I just pinged him about whether he actually did not. I'll report back in a few minutes.

AiyionPrime:
@bschelm just confirmed, only our vH35 ist unstable o these devices; pre non-ct switch is fine.

PR #2885, opened 2023-05-03
ipq40xx-chromium: add support for Google WiFi (Gale)

AiyionPrime:
> as well as the target itself
>    
>    Tested-by: Andrijan Möcker <amo@ct.de>

My goal is to have this device in the current gluon master, hence in the Gluon-Release we do before/when OpenWrt 23.xx lands.

This PR obviously can not be merged as is.
I started [an effort to backport the ten included patches in OpenWrt](https://patchwork.ozlabs.org/project/openwrt/list/?series=349989), which won't be as easy as I assumed. Though the suggested changes are as isolated as they can be and do not affect other devices let alone targets, adding them in OpenWrt would cost several hours and time of different OpenWrt maintainers.

Adding a new target in OpenWrt's stable is a thing that can be done, but is currently troublesome, due to some signatures that became invalid on OpenWrts side.
Furthermore different maintainers would be involved, whose time would be better spent in upcoming releases than in troublesome backports.

Maintaining the included ten commits downstream should be trivial, as they mostly affect the `chromium`-subtarget introduced in them which does not exist in OpenWrt's stable yet.
If we'd agree to have them downstream, I'd volunteer to maintain them.

Maybe @NeoRaider has a favored approach, @blocktrron was against doing this upstream, as long as he's involved, if not at all.

I'd be fine with either solution, but would like to see this device in gluon, as it has a decent dual-band ac performance and looks nice.

note: once @ctandi commences testing the patches, I'll squash this PR accordingly.


rotanid:
we discussed on IRC that this does not work with the current rules we gave us.
the device is not supported yet in openwrt-22.03 and maybe never will be and the device would also be marked as BROKEN, which excludes it from the upstream-backport-exception...

PR #2898, opened 2023-05-09
[master] [forward-port] docs: Add v2022.1.4 release notes

blocktrron:


AiyionPrime (approved):

herbetom (changes requested):
Just for safety: This needs another force push with the actual commit ID from v2022.1.x

AiyionPrime (changes requested):
@herbetom just pointed out, the referenced commit might be wrong one.

blocktrron:
@AiyionPrime How come? https://github.com/blocktrron/gluon/commit/205e66c74367fdab13a29e03c8db0f1d1bc98831 is the one on v2022.1.x

herbetom:
I had looked at the PR description. But yeah, seems like it's alreeady the correct in the commit.

blocktrron:
GitHub auto-introduces the content from the commit, but it does not make it to the history. This is a nothingburger.

AiyionPrime (approved):
whatever nothingburgers are?

**edit** I see. https://en.wikipedia.org/wiki/Nothingburger

PR #2907, opened 2023-05-21
Add abbility to create custom VLANs to network settings

mkg20001:
This is originally from @lemoer's PR

The patch works as-is but the UI is a bit wonky

DSA is a hard requirement for all switches,
so the existence of swconfig should disable this feature (if there are any non-DSA devices left after 23)

Bugs:
- [x] Creating/deleting a VLAN interface shows change only after reload/save => generate after applying changes
- [x] Interfaces that should not be used can be used (for example eth0 on hap ac2 should be left alone) => disabled on swconfig devices as it's unusuable there

image


mkg20001:
swconfig devices display a warning instead of ui being shown

The current method of only displaying after write is buggy, perhaps there could be a better way to hook that

mkg20001:
the reload fix causes the settings of individual interfaces to not be saved, as they are created too late

maurerle:
Are there still bugs with this PR?
What is the current state of it?

mkg20001:
Sortof. So the issue is that interfaces don't show up / don't vanish correctly aswell as remove not working which is a circular dependency we can't resolve in the old ui system.

There isn't a good solution other than the planned rewrite of the webinterface, which will take a bit longer.

If the patchset works for you despite those bugs you can also unoffically use it until then, but since it's not really working how it's supposed to we won't move forward until I get time to get the new frontend part done.

Djfe:
so if we want to remove an interface we created, we might need to do "firstboot" since it currently doesn't work at all (without the interface rewrite)?

mkg20001:
just removing it from /etc/config/gluon should be enough.

btw, it's just the ui. you can copy-paste those vlans on any device with dsa and they will just work.

```
config interface 'iface_eth0_vlan255'
	option name 'eth0.255'
	list role 'mesh'
````

Djfe:
should or I run `gluon-reconfigure` afterwards?

mkg20001:
yes and then gluon-reload (or better reboot)

AiyionPrime:

Thanks for trying!

PR #2913, opened 2023-06-04
ath79-nand: (re)add ZyXEL NBG6716

Djfe:
> Gone due to
> commit https://github.com/freifunk-gluon/gluon/commit/45c84a117bf820bf13e72187b46321b959e23aed ("ar71xx: drop target")

I want to test two things still:
Upgrade from Gluon v2021.1.x to master
From the check list: All network ports

Readding was possible thanks to @techtimo

  - [x] Web interface
  - [x] TFTP
  - [ ] Other: <specify>
        (`lua -e 'print(require("platform_info").get_image_name())'`) zyxel-nbg6716
      2.4G e8:37:7a:7a:1f:0c
      5G e8:37:7a:7a:1f:0d
      LAN e8:37:7a:7a:1f:0e
      WAN e8:37:7a:7a:1f:0f
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [x] Should show activity
  - Switch port LEDs
- ~~Outdoor devices only:~~
- ~~Cellular devices only:~~
- Docs:


Djfe:
I removed this manifest alias again, because sysupgrades won't work according to the [commit that readded support](https://git.openwrt.org/?p=openwrt/openwrt.git;a=commitdiff;h=99835e09992bb411519fda1f8d16a6d07df7f598). Upgrades/reinstalls need to be done manually.
```
	manifest_aliases = {
		'nbg6716',
	},
```
Notes for sysupgrades from Gluon 2021 and earlier:
Please follow these instructions:
https://git.openwrt.org/?p=openwrt/openwrt.git;a=commitdiff;h=99835e09992bb411519fda1f8d16a6d07df7f598

Autoupdater is incompatible, because the migration can't be done via sysupgrade. It requires flashing the partition using the `mtd` tool.
The device exists in several variants, ar71xx supported all of them. ath79 (Gluon 22) will only support variants with up to 128MB NAND flash.
If your Zyxel has 256MB of NAND flash, then this image should still work but you need to erase your ubi partition first when migrating. (This might require overwriting the mtd partition with zeroes but I don't know. I have no device to test. Maybe the command in the git commit is enough already. In-case of failure you can still try recovering using tftp, see the [OpenWrt wiki for instructions](https://openwrt.org/toh/zyxel/nbg6716#recovery))
Else ubi might cause bootloops when the existing partition on flash doesn't end within the limits set by mtd offset and size.

maurerle (approved):

PR #2905, opened 2023-05-13
openwrt: mask nested A-MSDU support for mesh

blocktrron:
This disables A-MSDU support on mesh-links. This restores throughput on mesh-links established between broken (ath10k) and correct (mt76) implementations of mesh A-MSDU.

This patch can be dropped when switching to the next OpenWrt release, as proper fixes have been applied.

Link: https://www.spinics.net/lists/linux-wireless/msg232410.html
Link: https://github.com/openwrt/mt76/issues/450

### Without this patch

```
root@64367-4040:~# iperf3 -c 2001:67c:2ed8:100e:aa63:7dff:fe2e:9378
Connecting to host 2001:67c:2ed8:100e:aa63:7dff:fe2e:9378, port 5201
[  5] local 2001:67c:2ed8:100e:eadf:70ff:fe79:bbc1 port 43576 connected to 2001:67c:2ed8:100e:aa63:7dff:fe2e:9378 port 5201
[ ID] Interval           Transfer     Bitrate         Retr  Cwnd
[  5]   0.00-1.00   sec  86.1 KBytes   705 Kbits/sec   19   2.36 KBytes       
[  5]   1.00-2.00   sec  0.00 Bytes  0.00 bits/sec    8   2.36 KBytes       
[  5]   2.00-3.00   sec  0.00 Bytes  0.00 bits/sec    8   2.36 KBytes       
[  5]   3.00-4.00   sec  0.00 Bytes  0.00 bits/sec    6   2.36 KBytes       
[  5]   4.00-5.00   sec  41.3 KBytes   338 Kbits/sec   10   2.36 KBytes       
[  5]   5.00-6.00   sec  0.00 Bytes  0.00 bits/sec   13   1.18 KBytes       
[  5]   6.00-7.00   sec  42.5 KBytes   348 Kbits/sec   14   2.36 KBytes       
[  5]   7.00-8.00   sec  41.3 KBytes   338 Kbits/sec   10   2.36 KBytes       
[  5]   8.00-9.00   sec  0.00 Bytes  0.00 bits/sec   10   2.36 KBytes       
[  5]   9.00-10.00  sec  0.00 Bytes  0.00 bits/sec   14   2.36 KBytes       
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.00  sec   211 KBytes   173 Kbits/sec  112             sender
[...]

PR #2793, opened 2023-02-22
ramips: add support for Cudy X6 v1

Djfe:
- [X] Must be flashable from vendor firmware
  - [X] Web interface (explained below)
  - [ ] TFTP
  - [X] Other: Signed openwrt from vendor can be flashed through web interface, next step: sysupgrade to Gluon
- [X] Must support upgrade mechanism
  - [X] Must have working sysupgrade
    - [X] Must keep/forget configuration (`sysupgrade [-n]`, `firstboot`)
  - [X] Gluon profile name matches autoupdater image name
        (`lua -e 'print(require("platform_info").get_image_name())'`) cudy-x6
- [X] Reset/WPS/... button must return device into config mode; works both
- [X] Primary MAC address should match address on device label (or packaging)
      (https://gluon.readthedocs.io/en/latest/dev/hardware.html#hardware-support-in-packages)
  - ~~When re-adding a device that was supported by an earlier version of Gluon, a
    factory reset must be performed before checking the primary MAC address, as
    the setting from the old version is not reset otherwise.~~
- Wired network
  - [X] should support all network ports on the device
  - [X] must have correct port assignment (WAN/LAN)
    - ~~On devices supplied via PoE, there is usually no explicit WAN/LAN labeling on the hardware.
      The PoE input should be the WAN port in this case.~~
- Wireless network (if applicable)
  - [X] Association with AP must be possible on all radios
  - [X] Association with 802.11s mesh must work on all radios 
  - [X] AP+mesh mode must work in parallel on all radios
- LED mapping
  - Power/system LED
    - [X] Lit while the device is on
    - [X] Should display config mode blink sequence 
          (https://gluon.readthedocs.io/en/latest/features/configmode.html)
  - ~~Radio LEDs~~
    - [ ] ~~Should map to their respective radio~~
    - [ ] ~~Should show activity~~
  - Switch port LEDs
    - [X] Should map to their respective port (or switch, if only one led present) 
    - [X] Should show link state and activity
- Outdoor devices only:
  - [ ] ~~Added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`~~
- Docs:
  - [X] Added Device to `docs/user/supported_devices.rst`

PR #2914, opened 2023-06-04
ramips-mt7621: Add support for TP-Link EAP615-Wall

citronalco:
  - [ ] Web interface
  - [ ] TFTP
  - [X] Other: ssh (see https://openwrt.org/toh/tp-link/eap615-wall#oem_easy_installation)
        `tp-link-eap615-wall-v1`
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    N/A, no radio LEDs
    - [ ] Should show activity
  - Switch port LEDs
    N/A, no switch port LEDs
- Outdoor devices only:
    N/A, indoor device
- Cellular devices only:
- Docs:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -387,6 +387,7 @@ ramips-mt7621
 
 * TP-Link
 
+  - EAP615-Wall (v1)
   - RE500 (v1)
   - RE650 (v1)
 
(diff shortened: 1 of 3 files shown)


citronalco:
Added "broken" to device: 802.11s mesh works, but is not stable.

Djfe:
Those are just warnings, that can be ignored according to @blocktrron 
see https://github.com/freifunk-gluon/gluon/pull/2793#issuecomment-1445531195

The instabilities could be due to:
https://github.com/freifunk-gluon/gluon/pull/2905
Please test, if the proposed patch fixes your issues :)
In that case, this device can be added as not broken. (Else we would need whole targets to be declared as broken)

They happen between ath10k wifi radios (5GHz qca988x or ipq40xx for 2.4GHz and 5GHz) and mediatek wifi radios and are caused by a faulty implementation for certain aggregated wifi frames in Linux. Qualcomm offloads them to hardware.

citronalco:
Have this device online for two weeks now. Client connections, mesh, everything works properly and stable.

PR #2915, opened 2023-06-10
Add ethernet module, add get_switch_type, make info shared, display switch type in info

mkg20001:
If wanted I can split the info modularization in it's own PR aswell


lemoer (approved):
LGTM

PR #2916, opened 2023-06-13
ath79-generic: add support for Ubiquiti NanoBeam 5AC 19 (XC)

achterin:
  - [ ] Web interface
  - [ ] TFTP
  - [x] Other: use patched ubntbox binary via ssh
- Wired network
- LED mapping
  - Power/system LED
- Outdoor devices only:
- Docs:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -127,6 +127,7 @@ ath79-generic
 
 * Ubiquiti
 
+  - NanoBeam 5AC 19 (XC)
   - NanoBeam M5 (XW)
   - NanoStation Loco M2/M5 (XW)
   - NanoStation M2/M5 (XW)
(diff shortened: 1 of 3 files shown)


docs/user/supported_devices.rst
```diff
@@ -126,7 +126,7 @@ ath79-generic
   - WBS510 (v1.20)
 
 * Ubiquiti
-
+  - NanoBeam 5AC 19 (XC)
```
AiyionPrime, line 129:
```suggestion

  - NanoBeam 5AC 19 (XC)
```

AiyionPrime (approved):
> The Ubiquiti NanoBeam AC Gen1 XC (NBE-5AC-19) is an outdoor 802.11ac CPE
> with a waterproof casing (ultrasonically welded) and bulb shaped.

lgtm

AiyionPrime (approved):
Nice, thanks.

maurerle (approved):

PR #2918, opened 2023-06-16
ramips-mt76x8: add support for TP-Link TL-MR6400 v5

Djfe:
  - [ ] Web interface
  - [x] TFTP: place renamed bootloader image (tp_recovery.bin) at tftp root, reachable at 192.168.0.225. Put device into tftp recovery by pressing reset/wps button, plugging in power and holding the button for at least 8 seconds. Device pulls firmware, flashes it and restarts. (see https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=b641eb6ecfbfc2c6a488483332d4b5a6ee4c736c)
  - [ ] Other: <specify>
        (`lua -e 'print(require("platform_info").get_image_name())'`) tp-link-tl-mr6400-v5
  - ~~When re-adding a device that was supported by an earlier version of Gluon, a
    the setting from the old version is not reset otherwise.~~
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs (there's one for 2.4GHz)
    - [x] Should show activity
  - Switch port LEDs (one wan and one lan led)
- ~~Outdoor devices only:~~
- Cellular devices only:
  - [x] Added board name with modem setup function `setup_ncm_qmi` to `package/gluon-core/luasrc/lib/gluon/upgrade/250-cellular`
- Docs:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -444,6 +444,7 @@ ramips-mt76x8
   - RE305 (v1) [#device-class-tiny]
   - TL-MR3020 (v3)
   - TL-MR3420 (v5)
+  - TL-MR6400 (v5)
   - TL-WA801ND (v5)
   - TL-WR841N (v13)
   - TL-WR902AC (v3)
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/250-cellular
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/250-cellular
@@ -58,6 +58,10 @@ elseif platform.match('ramips', 'mt7621', {
 	'wavlink,ws-wn572hp3-4g',
 }) then
 	setup_ncm_qmi('/dev/ttyUSB2', 'ncm', 15)
+elseif platform.match('ramips', 'mt76x8', {
+	'tplink,tl-mr6400-v5',
+}) then
+	setup_ncm_qmi('/dev/cdc-wdm0', 'qmi', 15)
 end
 
 uci:save('network')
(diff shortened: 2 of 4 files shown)


Djfe:
### additional notes:
I tested cellular and it's working. (Germany, O2 network)
10mbit/s download and 5.3mbit/s upload (could be due to low signal, my mobile phone contract[1] or our wireguard setup)
For comparison: VPN over my faster VDSL2 uplink achieved 11.8mbit/s DL and 20.7mbit/s UL.

[1] it's 1&1, a sim card with just 100MB with roaming disabled (came with my vdsl2 contract for free), so low speed could also be due to a cheap contract.
I swapped my primary SIM (Telekom flatrate) to an eSIM, so I couldn't test out that one.
_____

There are 3 signal strength LEDs on board for cellular but they aren't used by Gluon currently (similar to rssid but they aren't named/added that way in OpenWrt).
So there's no way to tell, whether cellular is being used at the moment (no activity led, nothing)
The Gluon status page doesn't supply any information either (idea for a future PR)

_____
I'm just wondering: Why are we renaming the tftp-recovery.bin files to bootloader.bin?
I couldn't find an explanation, just the first commit, that did it this way.
8395c08e667ab4eff86c8dfc3c5af6c36a16b2f4
https://github.com/freifunk-gluon/gluon/pull/1487
It was slightly confusing at first since the installation instructions on the OpenWrt wiki talked about a file called tftp-recovery and the firmware selector of OpenWrt also calls it this way.

Djfe:
vlt. nützlich für zukünftige Nutzer:
Der Router unterstützt nur LTE(4G) und UMTS(3G) aber kein GSM(2G), in Deutschland also ausschließlich LTE
Der Router unterstützt ausschließlich Nano-SIMs.
Es werden alle wichtigen deutschen LTE-Frequenzen unterstützt, höchstens Band 28 und 32 werden nicht unterstützt.

Unterstützt werden
FDD-LTE: Bänder 1, 3, 7, 8, 20
TDD-LTE: Bänder 38, 40, 41 (irrelevant für Deutschland)

PR #2920, opened 2023-06-18
modules: switch to openwrt-23.05

herbetom:
just a test for now


targets/mediatek-filogic
```diff
@@ -0,0 +1,20 @@
+-- ASUS
```
blocktrron, line 1: We don't add board-support in base-version updates.

targets/targets.mk
```diff
@@ -8,6 +8,7 @@ $(eval $(call GluonTarget,ipq40xx,mikrotik))
 $(eval $(call GluonTarget,ipq806x,generic))
 $(eval $(call GluonTarget,lantiq,xrx200))
 $(eval $(call GluonTarget,lantiq,xway))
+$(eval $(call GluonTarget,mediatek,filogic))
```
blocktrron, line 11: same

targets/ath79-generic
```diff
@@ -119,13 +119,13 @@ device('d-link-dir-505', 'dlink_dir-505', {
...
 })
 
-device('d-link-dir825b1', 'dlink_dir-825-b1', {
-	factory = false,
-	class = 'tiny', -- Only 6M of usable Firmware space
-	manifest_aliases = {
-		'd-link-dir-825-rev-b1', -- Upgrade from OpenWrt 19.07
-	},
-})
+-- device('d-link-dir825b1', 'dlink_dir-825-b1', {
```
blocktrron, line 122: This only affects factory-installation. Usable space is 7808k when upgrading from older versions.

Two ideas:

1. Backport https://github.com/openwrt/openwrt/commit/aca8bb5cc332f0ffdf4249e76b0a56716f98bef0 to the upcoming Gluon release to provide upgrade-path to newer image
2. Disable factory image generation in Gluon by either patching OpenWrt or disabling factory image build for this device exclusively.

blocktrron (commented):
The devices removed also require to be removed from docs.

targets/mediatek-filogic
herbetom: jup, needs to be dropped obviously (also AFAIK nobody tested those devices following the checklist), but since we're realistically months away from this switch happening i would keep it for now as it means that i don't need to track diffrent branches. (I'll put a "DROP FOR MERGE" into the Commit Message next time i'll force push.)

targets/ath79-generic
herbetom: `factory` is already set to false. Or is that not "strong enough"?

I disabled it because the build failed otherwise. But maybe i've misinterpreted the build error.

targets/mediatek-filogic
blocktrron: This can be weeks if not days away.

herbetom: Seems like your more optimistic in regards to when v2023.1.x will be branched off. I've dropped the commits.

targets/ath79-generic
herbetom: https://github.com/openwrt/openwrt/commit/aca8bb5cc332f0ffdf4249e76b0a56716f98bef0 was already backported to openwrt-22.03, so we should already have the expansion: https://github.com/openwrt/openwrt/commit/be2da47ced2fd0842922fdbd7a428a19138f6fc2

Djfe:
if this one
https://github.com/freifunk-gluon/gluon/pull/2916
gets added soon, then we need to disable it in openwrt-23 until this issues get's fixed:
https://github.com/openwrt/openwrt/issues/12882
flash is write protected bug

blocktrron:
@herbetom The DSA swconfig compat version patch can be dropped, we intended to carry it only for two releases which will conclude with the upcoming 2023.1.x branch.

herbetom:
Okay, that's good. This also simplifies bumping later on. I've dropped `ramips-mt7621-make-DSA-images-swconfig-upgradable.patch` and `lantiq-xrx200-make-DSA-images-swconfig-upgradable.patch`.

herbetom:
Okay, that's good. This also simplifies bumping later on. I've dropped `ramips-mt7621-make-DSA-images-swconfig-upgradable.patch` and `lantiq-xrx200-make-DSA-images-swconfig-upgradable.patch`.

Djfe:
but we need an `ipq40xx-generic-make-DSA-images-swconfig-upgradable.patch` now, right? (maybe also for the mikrotik subtarget)

herbetom:
@Djfe We would only need to do that if out intent was to allow direct updates from earlier versions then v2022.1 to the 23.05 based gluon (probably v2023.2).
With Gluon v2022.1 we started calling the sysupgrade command with the `--ignore-minor-compat-version` parameter:
- https://github.com/freifunk-gluon/gluon/commit/75ceb0d963b53ddde3b0ef2039f8a57ef6116eaf
- https://github.com/freifunk-gluon/packages/commit/b80428166454d184f356e5a14a144911d9ee11aa

Therefore it's not required to patch the `DEVICE_COMPAT_VERSION` in [`target/linux/ipq40xx/image/Makefile`](https://github.com/openwrt/openwrt/blob/894b2086fdf91409980df99ffe6ab8808de0302d/target/linux/ipq40xx/image/Makefile#L13-L14) because that was only required because previously Gluon didn't just ignore minor compat version changes.

Hoewer we probably should look into wheter we have to override some [interface assignments](https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-core/luasrc/lib/gluon/upgrade/020-interfaces).

Djfe:
I forgot how recent this switch was, ty

T-X:
@herbetom 
> Also someone needs to rebase the `kernel-bridge-Implement-MLD-Querier-wake-up-calls-Android-bug-workaround` patch if we want to keep that. Probably @T-X?

I don't want to keep it, I'd love to get rid of it :-). But as long as Android is incapable of fixing their sh*t, I feel like I have to... Will look into updating it, thanks for the ping.

T-X:
Here's a rebase of "kernel-bridge-Implement-MLD-Querier-wake-up-calls-Android-bug-workaround" onto this next-23.05 branch: https://github.com/T-X/gluon/commit/03c36d44c3dc44ec3d15fe755584388713597d76. Relatively straight-forward, some variable passing changes was needed (esp. "br" => "brmctx->br" and "port" => "pmctx->port" in the Linux bridge code) .

So far only compile-time tested. I (or someone else) still need(s) to runtime test it.

mweinelt:
```
2023-07-11T04:01:08.2299995Z 802.1d Ethernet Bridging (BRIDGE) [Y/n/m/?] y
2023-07-11T04:01:08.2300368Z   IGMP/MLD snooping (BRIDGE_IGMP_SNOOPING) [Y/n/?] y
2023-07-11T04:01:08.2301096Z     MLD Querier wake-up calls (BRIDGE_IGMP_SNOOPING_WAKEUPCALLS) [N/y/?] (NEW) make[8]: *** [scripts/kconfig/Makefile:77: syncconfig] Error 1
2023-07-11T04:01:08.2301859Z make[7]: *** [Makefile:628: syncconfig] Error 2
2023-07-11T04:01:08.2304150Z make[6]: *** [Makefile:748: include/config/auto.conf.cmd] Error 2
2023-07-11T04:01:08.2304896Z make[6]: Leaving directory '/home/runner/work/gluon/gluon/openwrt/build_dir/target-arm_cortex-a7+neon-vfpv4_musl_eabi/linux-ipq40xx_mikrotik/linux-5.15.119'
2023-07-11T04:01:08.2306051Z make[5]: *** [Makefile:24: /home/runner/work/gluon/gluon/openwrt/build_dir/target-arm_cortex-a7+neon-vfpv4_musl_eabi/linux-ipq40xx_mikrotik/linux-5.15.119/.modules] Error 2
2023-07-11T04:01:08.2306728Z make[5]: Leaving directory '/home/runner/work/gluon/gluon/openwrt/target/linux/ipq40xx'
2023-07-11T04:01:08.2307152Z make[4]: *** [Makefile:11: compile] Error 2
2023-07-11T04:01:08.2307644Z make[4]: Leaving directory '/home/runner/work/gluon/gluon/openwrt/target/linux'
2023-07-11T04:01:08.2308241Z time: target/linux/compile#23.86#7.19#27.88
2023-07-11T04:01:08.2310375Z     ERROR: target/linux failed to build.
2023-07-11T04:01:08.2310884Z make[3]: *** [target/Makefile:30: target/linux/compile] Error 1

**Comment by T-X** [comment]
@herbetom @mweinelt should be fixed in https://github.com/T-X/gluon/commit/205d92e82bedd5a18fb40516553c2562406b2c89 now.

Changelog v2: added "CONFIG_BRIDGE_IGMP_SNOOPING_WAKEUPCALLS=y" to target/linux/generic/config-5.15.

Also runtime tested now.

**Comment by herbetom** [comment]
@T-X Do you want to handle that in a sepeate PR, or should i add it to my branch or should i add it to my branch and squash it into the `modules: switch to openwrt-23.05` patch?

**Comment by rotanid** [comment]
> @T-X Do you want to handle that in a sepeate PR, or should i add it to my branch or should i add it to my branch and squash it into the `modules: switch to openwrt-23.05` patch?

wouldnt that be something for a separate "update patches" commit under the same PR ? or did you update other patches with the "switch to..." commit?

**Comment by herbetom** [comment]
> wouldnt that be something for a separate "update patches" commit under the same PR ? or did you update other patches with the "switch to..." commit?

I've removed/updated all other patches in that Commit. That's also what we regulary do in the "update to latest HEAD" PRs/commits. And since those changes are really essential (nothing will build without that) and the updated commit IDs are quite a simple change i do think updating the modules and updating the patches is something that fits together. 

I've squashed the updated patch for now. If it should be done diffrently we can change that of course.

**Comment by T-X** [comment]
@herbetom merging it into your [modules: switch to openwrt-23.05](https://github.com/freifunk-gluon/gluon/pull/2920/commits/873705cd20e5bd817892a9ae1fbc21e18484be62) commit, like you already did, looks good to me. Thanks!

Really looking forward to seeing this PR land in Gluon master. Looking forward to directed multicast then also working for not only link-local but also routable multicast addresses with batman-adv (within the mesh), due to the newer Linux (bridge) code.

**Comment by herbetom** [comment]
Currently, this PR disables (i.e. comments out) some devices and drops others (i.e. deletes the lines).
What I chose in each case is based on how likely I think it is that the reasons will persist. Basically: too little flash is a drop and the rest is a disable.
Hoewer, that won't always be true. We could also drop everything. That would interrupt the git history a bit more and it could send the wrong message to some. But we wouldn't have to "permanently" (until someone decides to throw it out) drag it along.
Opinions would be helpful.

**Comment by blocktrron** [comment]
Anything still left before this can be merged?

**Comment by herbetom** [comment]
@blocktrron I think the main thing left is a review. From my testing this seems to work so far.

**Comment by Djfe** [comment]
This patch will be dropped with Gluon v2024, I assume
(whenever upgrades from v2021.x are not supported anylonger)
https://github.com/herbetom/gluon/blob/65defe60822c148a3d703ea37ae9980b0a0fbaf5/patches/openwrt/0004-ramips-add-MT7621-WiFi-devpath-migration.patch

These two are still referring to wolfssl
https://github.com/herbetom/gluon/blob/next-23.05/package/gluon-mesh-wireless-sae/Makefile
https://github.com/herbetom/gluon/blob/next-23.05/package/gluon-wireless-encryption-wpa3/Makefile

File: targets/ath79-generic [hunk 4]
```diff
@@ -542,20 +519,20 @@ device('tp-link-wbs510-v1', 'tplink_wbs510-v1', {
...
+--		'ubiquiti-nanostation-loco-m5-xw', -- upgrade from OpenWrt 19.07
+--	},
+-- })
+
+-- device('ubiquiti-nanostation-m-xw', 'ubnt_nanostation-m-xw', {
+--	manifest_aliases = {
+--		'ubiquiti-nanostation-m2-xw', -- upgrade from OpenWrt 19.07
+--		'ubiquiti-nanostation-m5-xw', -- upgrade from OpenWrt 19.07
+--	},
+--})
```
Djfe, line 535: small thing, I just noticed a missing space here
`-- })`

herbetom:
@Djfe Thanks for taking a look.

I've dropped the `ramips-add-MT7621-WiFi-devpath-migration` patch. It's not planed that, whats probably going to become Gluon v2023.2, can be upgraded from anything earlier then Gluon v2022.1 / openwrt 22.03.
If we wanted to allow that we also would have to keep `ramips-mt7621-make-DSA-images-swconfig-upgradable.patch` arround as well as a similar new patch to allow upgrading ipq40xx to DSA ...

Regarding wolfssl: I'm not that familiar with it but we're not switching to or from wolfssl with this PR. Their's just a commit that's making sure that mbedtls isn't included.

Djfe:
I just found out that wifi is quite broken in rc2
(for some Mediatek devices atleast, like the Asus tuf AX4200)
atleast wifi 5ghz isn't working (no channel selection in luci and it crashes on starting, also wpa2/selecting encryption like wpa2 crashed it too)

rc1 worked fine for the random that I recommended the Asus tuf ax4200 to.

i'm fairly certain most bugs were fixed in master by now but might not have been backported. all in all it could make sense to wait for rc3

what I haven't checked: 23-snapshot (which this is based on)
I just wanted to warn that there were a few Regressions lately with hostapd, mt76 and stuff

Djfe:
I edited my last post. I'm fine with this being merged.

I need to confirm this still, but a test build of this branch failed for two targets:
lantiq-xrx200
ramips-7620

in both cases target/linux failed to build

rotanid:
> I've dropped the `ramips-add-MT7621-WiFi-devpath-migration` patch. It's not planed that, whats probably going to become Gluon v2023.2, can be upgraded from anything earlier then Gluon v2022.1 / openwrt 22.03. 

@herbetom when did we decide to break upgrade compatibility more often and earlier than in the past?
it was changed with v2022.1, v2023.1 (already discussed the problem with this on IRC) and now it should be done again?

herbetom:
@rotanid I've currently got no reference for it but we've said, when they were added, that we would keep those DSA / `DEVICE_COMPAT_VERSION` patches for approximately two major versions (v2022.1.x and v2023.1.x qualifies as that).

Especially the MT7621 patch was always a hassle to rebase. And perfoming a manual `sysupgrade` on those devices (with a patched firmware) results in a DEVICE_COMPAT_VERSION missmatch. All in all not ideal.

rotanid:
my main concern is not that we want to change upgrade compatibility with v2023.2 , but that we did with v2023.1
this way, 2023.1 is NOT the drop-in replacement for 2022.1 it is sometimes called.
that is no concern for myself or my community, to be clear.

herbetom:
@rotanid If you think that's an issue i would suggest opening an seperate Issue/PR.

If you feel stronlgy you could try figuring out why v2023.1 can only be upgraded from v2021.1 and not v2020.1. Maybe those barriers can be removed for v2023.1.x retroactively.

---

For the future it could make sense to have an "upgradeable_from" file/value somewhere in the repo which contains the current version from which upgrades are supported. That would allow quite easy tracking via git and it would also improve the workflow arround creating release notes.

blocktrron:
@rotanid I've had a look into the issue on ipq40xx

While the next release will in fact not be a drop-in replacement (sorry, i misunderstood your point in there), it will be able to be installed starting from Gluon v2022.1.x, as this is the first version to skip the minor compat hack in place.

https://github.com/freifunk-gluon/packages/commit/b80428166454d184f356e5a14a144911d9ee11aa

My above comment was written hastily. If we are fine with this, we do not need additional work for ipq40xx update case. If we want to target older versions, we need to carry hacks for ramips as well as ipq40xx, which i look at as a burden to big to carry.

Thoughts?

herbetom:
Earlier on in the state of this PR i rebased the neccesary patches to keep upgradability similar: https://github.com/herbetom/gluon/commit/a6c0eb7e4ff1d1b2c267c4199ca5bfe6793404b3

Maybe it helps someone. But i don't think those should be upstreamed and stay arround.

That v2023.1.x and v2022.1.x don't share the version from which they can be upgraded is of course not ideal.

PR #2857, opened 2023-04-15
ipq40xx-generic: add support for Meraki MR33

maurerle:
This device was tested by someone in our community.
This looks good so far, we are still investigating the open issues and see if we can get support for 5GHz Wifi out of the box too.

- [x] Must be flashable from vendor firmware
  - [ ] Web interface
  - [ ] TFTP
  - [x] Other: https://openwrt.org/toh/meraki/mr33#installation
- [x] Must support upgrade mechanism
  - [x] Must have working sysupgrade
    - [x] Must keep/forget configuration (`sysupgrade [-n]`, `firstboot`)
  - [x] Gluon profile name matches autoupdater image name
        (`lua -e 'print(require("platform_info").get_image_name())'`)
        #output is "meraki-mr33-access-point" - added to manifest_aliases
- [x] Reset/WPS/... button must return device into config mode
- [x] Primary MAC address should match address on device label (or packaging)
      (https://gluon.readthedocs.io/en/latest/dev/hardware.html#hardware-support-in-packages)
      #eth0 MAC matches - set in 010-primary
  - When re-adding a device that was supported by an earlier version of Gluon, a
    factory reset must be performed before checking the primary MAC address, as
    the setting from the old version is not reset otherwise.
- Wired network
  - [x] should support all network ports on the device
  - [x] must have correct port assignment (WAN/LAN)
    - if there are multiple ports but no WAN port:
      - the PoE input should be WAN, all other ports LAN
      - otherwise the first port should be delcared as WAN, all other ports LAN
- Wireless network (if applicable)
  - [x] Association with AP must be possible on all radios
  - [x] Association with 802.11s mesh must work on all radios 
  - [x] AP+mesh mode must work in parallel on all radios
  # monitor radio0 driver is disabled, radio1 is 2.4GHz and radio2 is 5GHz
- LED mapping
  - Power/system LED
    - [x] Lit while the device is on
    - [x] Should display config mode blink sequence 
          (https://gluon.readthedocs.io/en/latest/features/configmode.html)
  - ~~Radio LEDs~~ no radio LEDs
      - ~~Should map to their respective radio~~
      - ~~Should show activity~~
  - Switch port LEDs
    - [x] Should map to their respective port (or switch, if only one led present) 
    - [x] Should show link state and activity
- ~~Outdoor devices only:~~
  - ~~Added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`~~
- ~~Cellular devices only:~~
[...]

PR #2921, opened 2023-06-20
ipq40xx-generic: add support for ZTE MF289F

Dark4MD:
Also known in Germany as Gigacube CAT20
Added as Broken because you need to open the Device to access the Serial Console for flashing

  - [ ] Web interface
  - [ ] TFTP
  - [x] Other: Serial Console, see Commit in OpenWRT https://github.com/openwrt/openwrt/commit/0de6a3339f1aadc1de2c9371435e3de239a00645
- Wired network
- LED mapping
  - Power/system LED
    Device contains a Status LED inside which blinks in Config Mode. Power LED stays still.
  - Radio LEDs
    - [x] Should show activity
  - Switch port LEDs
- Outdoor devices only:
- Cellular devices only:
  - [x] Added board name with modem setup function `setup_ncm_qmi` to `package/gluon-core/luasrc/lib/gluon/upgrade/250-cellular`
- Docs:

### DIFF
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/250-cellular
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/250-cellular
@@ -53,6 +53,7 @@ elseif platform.match('ath79', 'nand', {
 	setup_ncm_qmi('/dev/ttyACM0', 'ncm', 15)
 elseif platform.match('ipq40xx', 'generic', {
 	'glinet,gl-ap1300',
+	'zte,mf289f',
 }) then
 	setup_ncm_qmi('/dev/cdc-wdm0', 'qmi', 15)
 elseif platform.match('ramips', 'mt7621', {
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua
@@ -80,6 +80,7 @@ function M.is_cellular_device()
 		return true
 	elseif M.match('ipq40xx', 'generic', {
 		'glinet,gl-ap1300',
+		'zte,mf289f',
 	}) then
 		return true
 	elseif M.match('ramips', 'mt7621', {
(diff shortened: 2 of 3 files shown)


herbetom:
Any reason you haven't marked this as ready and it's still a draft? The checklist looks complete?

Djfe:
Some notes:
This is a triband device with 256mb of RAM, if we enable the third radio, this will go OOM (out of memory), unless we use smallbuffers packages.

Gluon currently doesn't support three radios. If there are issues with 5GHz meshing:
I suggest disabling the package "ath10k-firmware-qca9984-ct".
that should fix 5GHz meshing, it will use ipq4019 for meshing then (this was also done for the Meraki MR33 https://github.com/freifunk-gluon/gluon/pull/2857 )
and it also prevents OOMs

I'm not sure whether ipq4019 on this device supports all 5GHz frequencies or only the lower half. (needs to be checked with `iw list`)

The device might still go OOM, if the LTE Modem also requires a substantial amount of system RAM to function (no clue).

__________________
Why was one CI check canceled?
Please rebase this PR and push -f to rerun the CI checks :)

Djfe:
@Dark4MD is your device `HW Version: T2`? (printed on the bottom of the device)
I just bought one used. It's the T2.

According to the commit message:
```
AT1 = QCA9984 5Ghz Radio on PCI-E bus
AT2 = IPQ4019 5Ghz Radio inside IPQ4019 like 2.4Ghz
```

So yeah: I misunderstood the commit message while writing my last post:
This device comes with two phys, it's not a triband device. Sorry for the confusion!

Dark4MD:
Finally got around and found the time to cleanup my build system (Got some weird bugs where it wont mesh over wifi and wont show the assigned ipv6 adresses on our map across all devices i build with it) and build a clean build with which i can finally say its working like i checked in the list.
@Djfe i have the T2 Revision as well (also known as Gigagube 4G in Germany by Vodafone).

Djfe:
awesome
I didn't know OpenWrt supported POTS :O
We could fix the LED thing upstream.

PR #2580, opened 2022-07-08
ath79-generic: (re)add Rocket M2/M5 (XM)

AiyionPrime:
> Gone due to
> commit 45c84a117bf8 ("ar71xx: drop target")

"nrb" intends to test this device "after sleeping".

- [x] Must be flashable from vendor firmware
  - ~~Web interface~~
  - [x] TFTP
  - ~~Other: <specify>~~
- [x] Must support upgrade mechanism
  - [x] Must have working sysupgrade
    - [x] Must keep/forget configuration (`sysupgrade [-n]`, `firstboot`)
  - [x] Gluon profile name matches autoupdater image name
        (`lua -e 'print(require("platform_info").get_image_name())'`)
- [x] Reset/WPS/... button must return device into config mode
- [x] Primary MAC address should match address on device label (or packaging)
      (https://gluon.readthedocs.io/en/latest/dev/hardware.html#notes)
  - When re-adding a device that was supported by an earlier version of Gluon, a
    factory reset must be performed before checking the primary MAC address, as
    the setting from the old version is not reset otherwise.
- Wired network
  - [x] should support all network ports on the device
  - [x] must have correct port assignment (WAN/LAN)
    - On devices supplied via PoE, there is usually no explicit WAN/LAN labeling on the hardware.
      The PoE input should be the WAN port in this case.
- Wireless network (if applicable)
  - [x] Association with AP must be possible on all radios
  - [x] Association with 802.11s mesh must work on all radios 
  - [x] AP+mesh mode must work in parallel on all radios
- LED mapping
  - Power/system LED
    - [x] Lit while the device is on
    - [ ] Should display config mode blink sequence 
          (https://gluon.readthedocs.io/en/latest/features/configmode.html)
  - Radio LEDs
    - [x] Should map to their respective radio
    - [x] Should show activity **rssi**
  - Switch port LEDs
    - [x] Should map to their respective port (or switch, if only one led present) 
    - [x] Should show link state and activity
- Outdoor devices only:
  - [x] Added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`

PR #2925, opened 2023-07-02
Revert "ath79-generic: (re)add Rocket M2/M5 (XM)"

herbetom:
This reverts commit 97efc8abdbd6264a19cbe6c5b95851245f6b798a. (#2580)

Since the commit was drafted the device was moved from ath79-generic to ath79-tiny ([4b5bd15](https://github.com/openwrt/openwrt/commit/4b5bd1509195bc8f19999ebe481b59356b5c3512)).

Therefore the commit breaks builds of the ath79-generic target (`unable to enable device 'ubnt_rocket-m'`).

It might be possible to add support for the ath79-tiny target, and with it the Rocket M2/M5 (XM), in the future.

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -130,7 +130,6 @@ ath79-generic
   - NanoBeam M5 (XW)
   - NanoStation Loco M2/M5 (XW)
   - NanoStation M2/M5 (XW)
-  - Rocket M2/M5 (XM)
   - UniFi AC Lite
   - UniFi AC LR
   - UniFi AC Mesh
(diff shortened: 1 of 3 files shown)


AiyionPrime:
@herbetom guck mal ob die Commitmessage wirklich so rein soll, oder ob du was schöneres findest.

herbetom:
I've changed it up a bit by figuring out the actual commit ID of the commit that was merged and not reverting the Merge commit (which was much easier to find).

AiyionPrime (approved):

Djfe:
Do you expect any testing for the new subtarget or can we resolve this by just adding the target and merging the rocket? (one PR, maybe two commits)
I'm surprised it's the only Gluon device that was moved to tiny so far.

AiyionPrime:
Depending on what functionalities OpenWrt has stripped, testing might be in order.
Given that Murxmaster proposed to test it, I don't see a reason why not to give it a shot.

The integration of a tiny target should get some sort of warning though.
We do not plan on putting much effort into the target and it might be dropped again sooner than expected due to this.

Djfe:
yeah that makes sense, let's see how well this runs and support it as long as possible, if issues arise, we may still add the tiny target/the rocket as broken
(or add a comment to the new target that no new devices will be accepted)

rotanid:
>  I'm surprised it's the only Gluon device that was moved to tiny so far.

see discussion over at #2850

PR #2937, opened 2023-08-17
ath79-generic: add support for Ubiquiti Rocket 5ac lite

ambassador86:
The Ubiquiti Rocket 5AC Lite (R5AC-Lite) is an outdoor router.

Specifications:
 - SoC: Qualcomm Atheros QCA9558
 - RAM: 128 MB
 - Flash: 16 MB SPI
 - Ethernet: 1x 10/100/1000 Mbps
 - WiFi 5 GHz: QCA988x
 - Buttons: 1x (reset)
 - LEDs: 1x power, 1x Ethernet, 4x RSSI

Test Device: https://map.chemnitz.freifunk.net/#!v:m;n:784558acb880

Integration Checklist:
  - [x] Other:  https://openwrt.org/toh/ubiquiti/common
        (`lua -e 'print(require("platform_info").get_image_name())'`) -> ubiquiti-rocket-5ac-lite
- [x] Reset button must return device into config mode
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [x] Should show activity
  - Switch port LEDs
- Outdoor devices only:
- Docs:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -132,6 +132,7 @@ ath79-generic
   - NanoBeam M5 (XW)
   - NanoStation Loco M2/M5 (XW)
   - NanoStation M2/M5 (XW)
+  - Rocket 5AC Lite (XC)
   - UniFi AC Lite
   - UniFi AC LR
   - UniFi AC Mesh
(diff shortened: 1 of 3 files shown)


blocktrron:
This now probably blocked by https://github.com/freifunk-gluon/gluon/issues/2939 ?

ambassador86:
Maybe, but i will not try to check if this device is affected, because it is too far away to debug something if physical access is needed.

blocktrron:
We've discussed this at todays mumble - The Flash locking has to be checked before this PR is merged. As the device seems to be mounted on your side, I'll close this PR until this has been verified.

Sorry, but the worst case happening here is we brick devices due to them leaving the flash locked and thus not able to upgrade until reset via TFTP or serial.

PR #2958, opened 2023-08-30
gluon-status-page: provide empty favicon

### LINKED ISSUE #2951: `gluon-status-page`: provide favicon
### Bug report

**What is the problem?**

No favicon is available.
By default the browser then makes a request to `/favicon.ico` which is not available.

Provide a favicon or disable by `meta`-tag to avoid unnecessary request.

**Gluon Version:**

current master, v2023.1

**Site Configuration:**

Any with package: `gluon-status-page`

**Custom patches:**

None.

Rotzbua:
## Change

Avoids the browser default request to `/favicon.ico` to avoid useless requests.

## Reference

Closes #2951

### DIFF
--- a/package/gluon-status-page/files/lib/gluon/status-page/view/layout.html
+++ b/package/gluon-status-page/files/lib/gluon/status-page/view/layout.html
@@ -9,6 +9,7 @@
 
 		<title><%:Error%></title>
 
+		<link rel="icon" href="data:,">
 		<link rel="stylesheet" href="/static/status-page.css" type="text/css">
 	</head>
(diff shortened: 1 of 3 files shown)


neocturne:
Hmm, when looking for this empty `data:` trick, I only found it mentioned on Stackoverflow. Is this actually valid? I'd rather leave the requests than introduce an unnecessary hack.

(also, you don't need to open issues for each PR)

blocktrron:
@Rotzbua Do you still have plans for addressing the suggestions?

Issue #2953, opened 2023-08-28
`gluon-status-page`: various html improvements


Rotzbua:
### Bug report

**What is the problem?**

* `script` can be move to the `head`
* accessibility issue with `viewport`
* outdated values used for `meta` in redirect
* modernize JS to `es2015`

**What is the expected behaviour?**

Solve above.

**Gluon Version:**

current master, v2023.1

**Site Configuration:**

Any with package: `gluon-status-page`

**Custom patches:**

None.

PR #2959, opened 2023-08-30
gluon-status-page: migrate JS to `es2015`-standard

Rotzbua:
## Change

* Migrate JS to `es2015`-standard
* Fix warning by Chrome:
`Canvas2D: Multiple readback operations using getImageData are faster with the willReadFrequently attribute set to true. See: https://html.spec.whatwg.org/multipage/canvas.html#concept-canvas-will-read-frequently`

## Reference

Mentioned in #2953


package/gluon-status-page/javascript/status-page.js
```diff
@@ -802,19 +805,19 @@
 	}
 
 	document.querySelectorAll('[data-interface]').forEach(function(elem) {
-		var ifname = elem.getAttribute('data-interface');
-		var address = elem.getAttribute('data-interface-address');
-		var wireless = !!elem.getAttribute('data-interface-wireless');
+		const ifname = elem.getAttribute('data-interface');
+		// const address = elem.getAttribute('data-interface-address');
```
neocturne, line 809: You can just remove the unused line altogether instead of commenting it.

package/gluon-status-page/javascript/status-page.js
```diff
@@ -286,42 +286,44 @@
 			'highlight': false,
 
 			'resize': function(w, h) {
-				var lastImage;
+				let lastImage;
 				try {
 					lastImage = ctx.getImageData(0, 0, w, h);
-				} catch (e) {}
+				} catch (e) {
+					console.error(e);
```
neocturne, line 293: I'm not sure whether this should be logged. Unfortunately, I don't remember why I added the try/catch here...

MDN describes two possible exceptions: https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/getImageData#exceptions

I don't think we can hit the SecurityError, so maybe this was just an ugly way to catch `w` or `h` being zero?

neocturne (changes requested):

blocktrron:
@Rotzbua Do you still have plans for addressing the suggestions?

PR #2965, opened 2023-09-01
rampis-mt7621: add support for ZyXEL WSM20

Kistelini:
(soory for the spam, same as https://github.com/freifunk-gluon/gluon/pull/2962, thought I lost access to my account)

Better known as ZyXEL Multy M1

logread is full of `daemon.err hostapd: nl80211: kernel reports: integer out of range`, but that should be gone after [this](https://patchwork.ozlabs.org/project/openwrt/patch/20230830161559.10770-3-nbd@nbd.name/) is merged.

<details><summary>There are kernel warnings, I have no idea what they mean.</summary>

```
[  278.192368] WARNING: CPU: 2 PID: 6161 at backports-6.1.24/net/mac80211/mesh_hwmp.c:353 airtime_link_metric_get+0x110/0x724 [mac80211]
[  278.204640] Modules linked in: iptable_nat batman_adv xt_state xt_nat xt_conntrack xt_REDIRECT xt_MASQUERADE xt_CT wireguard nf_nat nf_conntrack mt7915e mt76_connac_lib mt76 mac80211 libchacha20poly1305 ipt_REJECT ebtable_nat ebtable_filter ebtable_broute cfg80211 xt_time xt_tcpudp xt_quota xt_pkttype xt_owner xt_multiport xt_mark xt_mac xt_limit xt_comment xt_addrtype xt_TCPMSS xt_LOG ts_kmp ts_fsm ts_bm poly1305_mips nf_reject_ipv4 nf_log_syslog nf_defrag_ipv6 nf_defrag_ipv4 libcurve25519_generic iptable_mangle iptable_filter ip_tables hwmon ebtables ebt_vlan ebt_stp ebt_snat ebt_redirect ebt_pkttype ebt_mark_m ebt_mark ebt_limit ebt_ip6 ebt_ip ebt_dnat ebt_arpreply ebt_arp ebt_among ebt_802_3 compat chacha_mips sch_teql sch_sfq sch_multiq sch_gred sch_fq sch_dsmark sch_codel em_text em_nbyte em_meta em_cmp act_simple act_pedit act_csum libcrc32c sch_htb sch_hfsc em_u32 cls_u32 cls_route cls_matchall cls_fw cls_flow act_skbedit act_mirred act_gact ip6table_mangle ip6table_filter
[...]

PR #2966, opened 2023-09-05
ramips-mt7621: ZyXEL WSM20 add alias Multy M1

Djfe:
This improves upon:
982d8ed7cac1353c24772eb6d91b7a92c4dc6f36

These devices are only sold under the name Multy M1.
I added an alias for the official name :)

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -428,7 +428,7 @@ ramips-mt7621
 * ZyXEL
 
   - NWA50AX
-  - WSM20
+  - WSM20 / Multy M1
 
 * Xiaomi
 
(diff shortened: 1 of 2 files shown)


blocktrron:
This was requested to be removed in the original PR: https://github.com/freifunk-gluon/gluon/pull/2965#discussion_r1314240283

Djfe:
Interesting, I didn't know that. Works for me
So the Mikrotik hAP AC² should/could actually be renamed to `RBD52G-5HacD2HnD-TC` then?
https://github.com/freifunk-gluon/gluon/blob/ced0ccaf2bbb26f94df30608b47184125e644862/targets/ipq40xx-mikrotik#L5

Issue #2950, opened 2023-08-28
Broken DNS over WAN / missing dnsmasq after quick link downs+ups


T-X:
### Bug report

**What is the problem?**
- What is not working as expected?
  * fastd is unable to resolve hostnames and therefore unable to connect
  * "gluon-wan ping google.com" is not working, while "ping 141.1.1.1" works fine over WAN
- How is it misbehaving?
  * The dnsmasq instance that is supposed to listen on port 54 is not running:
```
root@nml-pa2200:~# netstat -tulpen | grep :54
udp        0      0 :::546                  :::*                                7382/odhcp6c
udp        0      0 :::546                  :::*                                1971/odhcp6c
```
- When did the problem first start showing up?
  * It seems to have happened after (a few?) quick link ups/downs on the WAN/LAN ports (caused by flaky cabling or power issues on the attached (PoE) switch(es)) - two dmesg examples: https://gist.github.com/T-X/31d9c83e61d533ebc2585db12211be54
- What were you doing when you first noticed the problem?
  * Using the Freifunk network
- On which devices (vendor, model and revision) is it misbehaving?
  * Plasma Cloud PA2200: https://map.luebeck.freifunk.net/#!v:m;n:4c13650016e0, which currently provides the main & only uplink to the local mesh cloud
- Does the issue appear on multiple devices or targets?
  * untested
- Workarounds?
  * the device works fine again after rebooting it, netstat then looks ok:
```
root@nml-pa2200:~# netstat -tulpen | grep :54
tcp        0      0 0.0.0.0:54              0.0.0.0:*               LISTEN      2448/dnsmasq
tcp        0      0 :::54                   :::*                    LISTEN      2448/dnsmasq
udp        0      0 0.0.0.0:54              0.0.0.0:*                           2448/dnsmasq
udp        0      0 :::546                  :::*                                2349/odhcp6c
udp        0      0 :::546                  :::*                                2010/odhcp6c
udp        0      0 :::54                   :::*                                2448/dnsmasq
```

**What is the expected behaviour?**
- How do you think it should work instead?
  * after a link went down and up again then not only routing over WAN but also DNS over WAN should work again
- Did it work like that before?
  * unknown

**Gluon Version:**
- v2022.1

**Site Configuration:**
- https://git.chaotikum.org/freifunk-luebeck/site-ffhl/-/tree/v0.16.0?ref_type=tags

**Custom patches:**
- none

blocktrron:
[...]

PR #2969, opened 2023-09-05
gluon-wan-dnsmasq: rewrite initfile to use procd

blocktrron:
Rewrite the gluon-wan-dnsmasq initfile to use procd. This allows for configuring restart handling by procd in case of a crash.

While at it, disable the caching feature optionally used for the regular dnsmasq on the wan dnsmasq. This daemon is only for redirecting DNS requests to use local network resolvers, not introducing caching.

### DIFF
--- a/package/gluon-wan-dnsmasq/files/etc/init.d/gluon-wan-dnsmasq
+++ b/package/gluon-wan-dnsmasq/files/etc/init.d/gluon-wan-dnsmasq
@@ -2,28 +2,18 @@
 
 START=60
 
-SERVICE_NAME=gluon-wan-dnsmasq
-SERVICE_USE_PID=1
-SERVICE_PID_FILE=/var/run/gluon-wan-dnsmasq.pid
-
-
-PORT=54
-PACKET_MARK=1
+USE_PROCD=1
 
 RESOLV_CONF_DIR=/var/gluon/wan-dnsmasq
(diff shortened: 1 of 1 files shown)


skorpy2009 (approved):

rotanid:
@T-X can you test if this fixes #2950 for you?

package/gluon-wan-dnsmasq/files/etc/init.d/gluon-wan-dnsmasq
```diff
@@ -2,28 +2,18 @@
...
+USE_PROCD=1
 
 RESOLV_CONF_DIR=/var/gluon/wan-dnsmasq
 RESOLV_CONF=$RESOLV_CONF_DIR/resolv.conf
 
-
-start() {
+start_service() {
 	mkdir -p $RESOLV_CONF_DIR
 	/lib/gluon/wan-dnsmasq/update.lua
```
neocturne, line 12: I just noticed that there may be a race condition between the init script and the hotplug.d call of `/lib/gluon/wan-dnsmasq/update.lua` (unrelated to this PR, but we might as well fix it now).

I think it should be sufficient to replace the `update.lua` call here with a `touch "$RESOLV_CONF"`: If the WAN interface is not up at this point yet, `update.lua` would just have an empty `resolv.conf` here, and if it is, a proper `resolv.conf` has been (or is currently being) created by the hotplug script.

neocturne (approved):

PR #2971, opened 2023-09-07
contrib: push Docker build-container to GitHub registry

blocktrron:
This adds a workflow to push the build-container described in Gluons Dockerfile to the freifunk-gluon organizations container registry.

This will create a image

 - For every push to master
 - For every tagged version

### DIFF
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -1,0 +1,42 @@
+# Based on the example from https://docs.github.com/en/actions/publishing-packages/publishing-docker-images
+name: Create and publish a Docker image
+
+on:
+  push:
+    branches:
+      - 'master'
+    tags:
+      - 'v*'
+  pull_request:
+    branches:
+      - 'master'
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: gluon-build
+
+jobs:
+  build-and-push-image:
+    runs-on: ubuntu-latest
+    permissions: write-all
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


skorpy2009 (approved):

.github/workflows/build-container.yml
```diff
@@ -0,0 +1,42 @@
...
+        uses: docker/login-action@65b78e6e13532edd9afa3aa52ac7964289d1a9c1
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@9ec57ed1fcdbf14dcef7dfbe97b2010124a938b7
+        with:
+          images: ${{ env.REGISTRY }}/freifunk-gluon/${{ env.IMAGE_NAME }}
```
herbetom, line 35: This action is also triggered in forks which in itself is fine.

Hoewer, for me that always results in a failed action becuase i don't have the neccesary permissions to push to the freifunk-gluon org.

```
------
 > pushing ghcr.io/freifunk-gluon/gluon-build:master with docker:
------
ERROR: denied: permission_denied: The requested installation does not exist.
```

I don't know what would happen if members of the Org (with the necceary permissions) try the same and if this could unintentionally override "official" images.

I think it would be best if the `freifunk-gluon` part would be an dynamic variable.

herbetom: It seems like it's luckily not possible to push towards other Orgs/Users.

PR #2976, opened 2023-09-10
workflows: build-container: build images for amd64 and arm64

herbetom:


Djfe:
related log:
https://github.com/freifunk-gluon/gluon/actions/runs/6173756219/job/16756823009?pr=2982

blocktrron:
What's the status here? This needs at least a rebase and the Draft state removed.

Did you came around to test the resulting images?

herbetom:
I've rebased and tested an x86-64 as well as an mpc85xx-p1020 image built inside the container on an arm64/aarch64 system. I didn't notice any problems withe the build system not beeing x86-64.

contrib/docker/Dockerfile
```diff
@@ -1,5 +1,10 @@
 FROM debian:bookworm-slim
 
+ARG TARGETPLATFORM
+ARG BUILDPLATFORM
```
blocktrron, line 4: Are these required?

herbetom: I was under the impression that they were required. That doesn't seems to be the case. I've removed them.

PR #2978, opened 2023-09-11
generic: disable building openvswitch

herbetom:
openvswitch fails to build with groff v1.23 on the build system.

This groff version hasn't reached a Debian Release yet but is already present with rolling release distros and causes the gluon build to fail.

Since we don't need openvswitch the best temporary fix seems to disable building it by default until this get's resolved upstream.

(Let's see if opening this or even merging this is followed by an coincidental upstream solution. :))

If this is merged it shoud be backported.

### DIFF
--- a/targets/generic
+++ b/targets/generic
@@ -53,6 +53,18 @@ try_config('PACKAGE_dnsmasq_full_noid', false)
 try_config('PACKAGE_dnsmasq_full_broken_rtc', false)
 try_config('PACKAGE_dnsmasq_full_rtc', false)
 
+try_config('PACKAGE_openvswitch', false) -- fails to build
+try_config('PACKAGE_kmod-openvswitch', false) -- fails to build
+try_config('PACKAGE_kmod-openvswitch-geneve', false) -- fails to build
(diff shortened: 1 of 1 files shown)


targets/generic
```diff
@@ -53,6 +53,18 @@ try_config('PACKAGE_dnsmasq_full_noid', false)
 try_config('PACKAGE_dnsmasq_full_broken_rtc', false)
 try_config('PACKAGE_dnsmasq_full_rtc', false)
 
+try_config('PACKAGE_openvswitch', false) -- fails to build
```
neocturne, line 56: This line is not necessary, as it is only pulled in by the kmods.

neocturne:
For a backport, we should check if the list of kmods is the same in older OpenWrt.

neocturne:
Upstream fix: https://github.com/openvswitch/ovs/commit/0945e1a5fa3d66749facd24f8c60f732f7220f11

I think it's better to backport the fix to OpenWrt.

herbetom:
I agree.

PR #2979, opened 2023-09-12
Update tunneldigger-watchdog

cmarxmeier:
current tunneldigger misses .pid file for tunneldigger-watchdog quick and dirty workaround to avoid tunnel restarts every 5 minutes

### DIFF
--- a/package/gluon-mesh-vpn-tunneldigger/luasrc/usr/bin/tunneldigger-watchdog
+++ b/package/gluon-mesh-vpn-tunneldigger/luasrc/usr/bin/tunneldigger-watchdog
@@ -33,7 +33,9 @@ local function has_mesh_vpn_neighbours()
 end
 
 if uci:get_bool('tunneldigger', 'mesh_vpn', 'enabled') then
-	if io.popen('pgrep -x /usr/bin/tunneldigger'):read('*l') ~= read_pid_file() then
+-- tunneldigger misses .pid file for this script - workaround and simply generate one from process-list
+        os.execute('logger -t tunneldigger-watchdog "generating missing .pid file for gluon...";ps | grep "bin/tunneldigger -f" -m 1 | cut -d " " -f 2 >/var/run/tunneldigger.mesh-vpn.pid')
+        if io.popen('pgrep -x /usr/bin/tunneldigger'):read('*l') ~= read_pid_file() then
 		os.execute('logger -t tunneldigger-watchdog "Process-Pid does not match with pid-File."')
 		restart_tunneldigger()
 		return


neocturne:
The whole point of the first check is to trigger when the PID file doesn't match the process, so rather than this hack, removing that restart condition would have the same effect. It is not clear to me what that check was supposed to accomplish however... so I'd like to gather some more information before deciding on a fix.

neocturne:
With procd, that part of the watchdog is useless, because procd will already restart tunneldigger when it crashes.

cmarxmeier:
a simple quality check in tunneldigger env would have shown this error -  showed up in our firmware test in 5 minutes and took some time to find the reason for crashing tunnels. changed the script and started building firmware for our hoods for release, when testnodes behave brave with the newly generated images.

cmarxmeier:
@neocturne : you're welcome :D

root@rettesichwerkann:/usr/bin# ubus call service list '{"name":"tunneldigger"}'
{
        "tunneldigger": {
                "instances": {
                        "tunneldigger_1": {
                                "running": true,
                                "pid": 4228,
                                "command": [
                                        "/usr/bin/tunneldigger",
                                        "-f",
                                        "-u",
                                        "a42bb0ca86da",
                                        "-i",
                                        "mesh-vpn",
                                        "-t",
                                        "1",
                                        "-b",
                                        "lohmar1.freifunk-rhein-sieg.net:20003",
                                        "-b",
                                        "lohmar2.freifunk-rhein-sieg.net:20003",
                                        "-b",
                                        "lohmar3.freifunk-rhein-sieg.net:20003",
                                        "-I",
                                        "br-wan"
                                ],
                                "term_timeout": 5,
                                "respawn": {
                                        "threshold": 3600,
                                        "timeout": 5,
                                        "retry": 5
                                }
                        }
                }
        }
}

node: http://[2a03:2260:123:300:a62b:b0ff:feca:86da]/cgi-bin/status

package/gluon-mesh-vpn-tunneldigger/luasrc/usr/bin/tunneldigger-watchdog
```diff
@@ -33,7 +33,9 @@ local function has_mesh_vpn_neighbours()
 end
 
 if uci:get_bool('tunneldigger', 'mesh_vpn', 'enabled') then
-	if io.popen('pgrep -x /usr/bin/tunneldigger'):read('*l') ~= read_pid_file() then
+-- tunneldigger misses .pid file for this script - workaround and simply generate one from process-list
+        os.execute('logger -t tunneldigger-watchdog "generating missing .pid file for gluon...";ps | grep "bin/tunneldigger -f" -m 1 | cut -d " " -f 2 >/var/run/tunneldigger.mesh-vpn.pid')
+        if io.popen('pgrep -x /usr/bin/tunneldigger'):read('*l') ~= read_pid_file() then
 		os.execute('logger -t tunneldigger-watchdog "Process-Pid does not match with pid-File."')
 		restart_tunneldigger()
 		return
```
neocturne, line 41: Let's just drop this whole condition (and the function `read_pid_file`) - procd will automatically restart tunneldigger anyways, so it is not needed anymore.

At some point, we could also replace the `uci:get_bool('tunneldigger', 'mesh_vpn', 'enabled')` with a query to procd whether the service is actually supposed to be running (so manually starting and stopping it will be noticed by the watchdog), but that's fairly low priority, so let's leave that for a future PR.

neocturne:
> a simple quality check in tunneldigger env would have shown this error - showed up in our firmware test in 5 minutes and took some time to find the reason for crashing tunnels. changed the script and started building firmware for our hoods for release, when testnodes behave brave with the newly generated images.

Yeah, I wouldn't exactly call the tunneldigger support in Gluon "maintained" ... at some point we might have to move it to community-packages unless someone steps up, so we actually get some testing *before* major releases (but I'm not sure if keeping the support is even desirable for future Gluon releases, given that tunneldigger doesn't even support IPv6, and these connection issues have been unresolved for years...)

cmarxmeier:
Whenever a news gluon release is commited on github, we'll check before releasing/rolling out images. Therefore you'd received this quick-and-dirty patch commit directly to ensure stable working  networks as far as possible.
https://map.freifunk-rhein-sieg.net/#!/de/map
Standard nodes on this map run tunneldigger - some bigger installations have dedicated wireguard connects.
See no reason to enable encryption on weak routers when we blow the traffic to FF Rheinland by means of unencrypted GRE Tunnel. Either enryption all the way or none - therefore we decided to use l2tp implementation of tunneldigger.

neocturne:
> Whenever a news gluon release is commited on github, we'll check before releasing/rolling out images. Therefore you'd received this quick-and-dirty patch commit directly to ensure stable working networks as far as possible. https://map.freifunk-rhein-sieg.net/#!/de/map Standard nodes on this map run tunneldigger - some bigger installations have dedicated wireguard connects. See no reason to enable encryption on weak routers when we blow the traffic to FF Rheinland by means of unencrypted GRE Tunnel. Either enryption all the way or none - therefore we decided to use l2tp implementation of tunneldigger.

*If* we decide to drop tunneldigger support in the future, fastd-l2tp would likely be our recommendation for unencrypted tunnels.

In any case, thanks for the report - do you want to update the PR according to my suggestion, or should we take care of it?

neocturne:
Thanks, I've created #2981 to fix the issue.

rotanid:
> Either enryption all the way or none - therefore we decided to use l2tp implementation of tunneldigger.

have you tried fastd-l2tp ? it wasn't there in the past, so maybe you don't need tunneldigger anymore

PR #2983, opened 2023-09-13
contrib: Dockerfile: install openssh-client

skorpy2009:
Add ssh-client to enable deployment of the build output from CI.


herbetom:
I kinda don't think a CI build action should have access to "credentials" (ssh keys). Hoewer in the proposed workflow that would be required to make it work.

A seperate build step to upload the build output (which could use another image where ssh is included) seems cleaner to me.

PR #2994, opened 2023-09-21
modules: bump to latest HEAD

blocktrron:


rotanid:
@blocktrron CI failed :-/

patches/openwrt/0007-fixup-kernel-bridge-Implement-MLD-Querier-wake-up-calls-Android-bug-workaround.patch
```diff
@@ -0,0 +1,78 @@
...
++ 	BUILD_BUG_ON(sizeof(diff) < __DEV_ATTR_MAX / 8);
++ 
++@@ -1267,8 +1267,8 @@ interface_device_config_changed(struct i
++ 	blobmsg_parse(device_attr_list.params, __DEV_ATTR_MAX, ntb,
++ 		blob_data(if_new->config), blob_len(if_new->config));
++ 
++-	uci_blob_diff(ntb, otb, &device_attr_list, &diff);
++-	return diff;
+++	uci_blob_diff(ntb, otb, &device_attr_list, diff);
+++	return !!(diff[0] | diff[1]);
```
neocturne, line 40: The `!!` can be omitted, the function already returns `bool`.

I'm not entirely convinced that `|` is used instead of `||` when the result is only used as `bool`, but at least it's consistent with the other place in netifd where a diff of length 2 is checked...

PR #2995, opened 2023-09-23
gluon-mesh-batman-adv-brmldproxy: add package

T-X:
Now that we have general support for routable IPv6 multicast address in Gluon master thanks to the newer Linux (bridge) and batman-adv versions it becomes more interesting to also support layer 3 IPv6 multicast routers.

So far this was also not possible with the default settings in Gluon due to filtering MLD into the mesh. This now adds support for brmldproxy, a daemon which proxies MLD reports between bridge ports.

For the Gluon scenario this package adds brmldproxy proxying configuration for the mesh side bat0 bridge port.

The configuration is tuned in a way to enable the usage of layer 3 IPv6 multicast routers for routable IPv6 multicast address ranges. But with a lot smaller MLD overhead compared to the filter_membership_reports=false site.conf option.

If a node has no multicast listener for a routable IPv6 multicast address then this node will emit no MLD report into the mesh. Furthermore, if a node has multiple multicast listening hosts for routable IPv6 multicast addresses then the node will act in deputy and respond with combined, aggregated MLD reports on behalf.

This package is currently incompatible with a
filter_membership_reports=true site.conf option.

---

This pull-request requires: https://github.com/freifunk-gluon/packages/pull/264


T-X:
Changelog v2:
* added necessary firewall rules for the brmldp0 interface through `package/gluon-mesh-batman-adv-brmldproxy/luasrc/lib/gluon/upgrade/400-brmldproxy-firewall` (not much, as the `tc` rules that are set up by `brmldproxy` are applied before iptables)

T-X:
Changelog v3:
* Added a patch / config option to batman-adv to forward MLD reports only to multicast routers instead of always flooding MLD reports, to further reduce overhead.
  * That way, when installing this package it makes no difference in overhead, still no MLD reports in the mesh then if no multicast router is present (as I assume no one has installed any multicast routers yet). On the other hand the batman-adv patch should be inactive / make no difference when gluon-mesh-batman-adv-brmldproxy is not installed.

ecsv:
Regarding the batman-adv patch: I think this a patch which could be integrated in upstream when the actual real world effects are verified/tested in actual setups.

T-X:
As discussed during the Gluon meeting back then it would be great to gather a bit of real world experiences with it.

I've now prepared some packages with all the upcoming batman-adv / Gluon multicast features integrated into Gluon package feeds. So far tested on VMs only, but real hardware tests coming up next. It should be enough to just add it to your site/modules file and update all nodes with it. Should compile+run for Gluon v2021.1.x, Gluon v2023.1 and Gluon master: https://github.com/T-X/gluon-batman-adv-next/

If any community would be interested in testing IPv6 multicast routing between communities (on their own Gluon domains), let me know.

T-X:
With the last firmware update at Freifunk Lübeck the latest gluon-batman-adv-next package feed was rolled out, with brmldproxy integration and the latest batman-adv release (v2024.0). So far we didn't notice any issues/regressions.

Haven't found a second community to test an IPv6 multicast router (pim6sd) on a Gluon based network with, but with using mrdisc (https://github.com/troglobit/mrdisc) on one wifi client I can see the MLD reports for (mostly) routable multicast addresses arriving on the local Gluon node from other nodes in tcpdump. And when stopping mrdisc, I don't see these MLD reports anymore.

So to me seems to work as expected/intended on a real setup.

package/gluon-mesh-batman-adv-brmldproxy/Makefile
```diff
@@ -0,0 +1,36 @@
...
+
+	If a node has no multicast listener for a routable IPv6
+	multicast address then this node will emit no MLD report
+	into the mesh. Furthermore, if a node has multiple multicast
+	listening hosts for routable IPv6 multicast addresses then the
+	node will act in deputy and respond with combined, aggregated
+	MLD reports on behalf.
+
+	This package is currently incompatible with a
+	filter_membership_reports=true site.conf option.
```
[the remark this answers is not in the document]
T-X: will fix this, thanks!

T-X:
Changelog v4:
* rebased to current master
* added option to also use filter_membership_reports=false
* for this removed the static bpfcountd UCI config file and generate it via a Gluon upgrade Lua script instead
* for that renamed /lib/gluon/upgrade/400-brmldproxy-firewall to /lib/gluon/upgrade/400-brmldproxy
* absence of filter_membership_reports now defaults to "true" for gluon-mesh-batman-adv-brmldproxy
* updated Makefile+docs description
* moved "batctl multicast_mld_rtr_only 1" call addition from netifd gluon_bat0.sh to a separate hotplug.d script
* added patch: "gluon-ebtables: don't filter incoming MLD Reports via ebtables"
* added patch: "gluon-mesh-batman-adv-brmldproxy: add to mesh-batman-adv-15 by default"

The bpfcountd concept has now also successfully been tested between Freifunk Lübeck and a domain at Freifunk Vogtland. With pim6sd and bpfcountd multicast audio could be transmitted on one and listened to in the other Gluon site/domain.

package/features
```diff
@@ -38,6 +38,7 @@ when(_'web-advanced' and _'autoupdater', {
 
 when(_'mesh-batman-adv-15', {
 	'gluon-ebtables-limit-arp',
+	'gluon-mesh-batman-adv-brmldproxy',
```
neocturne, line 41: I don't think the feature should be included by default.

package/gluon-mesh-batman-adv-brmldproxy/files/etc/hotplug.d/iface/50-gluon-mesh-batman-adv-brmldproxy
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+([ "$INTERFACE" != "bat0" ] || [ "$ACTION" != "ifup" ]) && exit 0
+
+[ "$(gluon-show-site | jsonfilter -e "@.mesh.filter_membership_reports")" = "false" ] && exit 0
+
+[ -x /etc/init.d/brmldproxy ] && /etc/init.d/brmldproxy enabled && \
```
neocturne, line 7: - There is no need to check for the executability, `gluon-mesh-batman-adv-brmldproxy` depends on `brmldproxy`
- There is no need to check for the `enabled` status (on Gluon, all installed services are always enabled; disabling is not preserved on ugprades, so it is not supported). If disabling should be possible, a `/etc/config/gluon` flag must be introduced.

package/gluon-mesh-batman-adv-brmldproxy/files/etc/hotplug.d/iface/50-gluon-mesh-batman-adv-brmldproxy
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+([ "$INTERFACE" != "bat0" ] || [ "$ACTION" != "ifup" ]) && exit 0
+
+[ "$(gluon-show-site | jsonfilter -e "@.mesh.filter_membership_reports")" = "false" ] && exit 0
```
neocturne, line 5: Printing the whole site and then filtering it inefficient. You can copy the `lookup_site` helper from `package/gluon-mesh-batman-adv/files/lib/netifd/proto/gluon_bat0.sh`.

package/gluon-ebtables/luasrc/lib/gluon/ebtables/105-mcast-drop-igmp-mld
```diff
@@ -13,8 +13,4 @@ if site.mesh.filter_membership_reports(true) then
 	rule('MULTICAST_OUT_ICMPV6 -p IPv6 --ip6-protocol ipv6-icmp --ip6-icmp-type 131 -j DROP') -- MLDv1 Report
 	rule('MULTICAST_OUT_ICMPV6 -p IPv6 --ip6-protocol ipv6-icmp --ip6-icmp-type 132 -j DROP') -- MLDv1 Done
 	rule('MULTICAST_OUT_ICMPV6 -p IPv6 --ip6-protocol ipv6-icmp --ip6-icmp-type 143 -j DROP') -- MLDv2 Report
-
-	rule('MULTICAST_IN_ICMPV6 -p IPv6 --ip6-protocol ipv6-icmp --ip6-icmp-type 131 -j DROP', 'nat') -- MLDv1 Report
```
neocturne, line 17: Should these be allowed unconditionally, or only when the feature is enabled?

neocturne (changes requested):

T-X:
Changelog v5:
* rebased to current master (no conflicts)
* removed the patch "gluon-mesh-batman-adv-brmldproxy: add to mesh-batman-adv-15 by default" to not add brmldproxy by default with batman-adv
* /etc/hotplug.d/iface/50-gluon-mesh-batman-adv-brmldproxy:
  * removed executable/enabled checks on /etc/init.d/brmldproxy, as not required due to dependencies and we don't allow manually disabling it for now
  * use Lua JSON filtering/checking instead of piping the full site through the jsonfilter process
* only remove incoming MLD report ebtables filters if using brmldproxy, to avoid any functional changes to any setups without brmldproxy

rotanid:
@T-X have you seen the comments by neocturne? these are without answers, at least as far as i can see

T-X:
Changelog v6:
* rebased to current main branch (no conflicts)
* replaced the batman-adv mld-rtr-only configuration option with additional tc-DNAT'ing for now, to be able to use brmldproxy with low MLD report overhead even without any changes to batman-adv
  * note: we have to temporarily change the ICMPv6 type from MLD report to something else, here "100 - Private experimentation", to avoid that batman-adv always floods it
* simplified the "don't filter incoming MLD Reports" patch by moving the change from gluon-ebtables to gluon-mesh-batman-adv-brmldproxy
  * these two changes allow to use gluon-mesh-batman-adv-brmldproxy with both a vanilla Gluon and vanilla batman-adv now, making it suitable for any external feed as is
* replaced `( [ ... ] || [ ... ] ) && ...` in `gluon-mesh-batman-adv-brmldproxy` with `{ [ ... ] || [ ...]; } && ...` to avoid a sub-shell invocation, to make the linter happy

T-X:
@rotanid thanks for the comment, but I think all of @neocturne 's previous remarks were addressed in v5 already

package/gluon-mesh-batman-adv-brmldproxy/files/etc/hotplug.d/iface/51-gluon-mesh-batman-adv-brmldproxy
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
+
+{ [ "$INTERFACE" != "client" ] || [ "$ACTION" != "ifup" ]; } && exit 0
```
neocturne, line 3:
```suggestion
if [ "$INTERFACE" != "client" ] || [ "$ACTION" != "ifup" ]; then exit 0; fi
```

package/gluon-mesh-batman-adv-brmldproxy/files/etc/hotplug.d/iface/51-gluon-mesh-batman-adv-brmldproxy
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
+
```
neocturne, line 2:
```suggestion
set -e
```

package/gluon-mesh-batman-adv-brmldproxy/files/etc/hotplug.d/iface/51-gluon-mesh-batman-adv-brmldproxy
```diff
@@ -0,0 +1,37 @@
...
+{ [ "$INTERFACE" != "client" ] || [ "$ACTION" != "ifup" ]; } && exit 0
+
+lookup_site() {
+	local path="$1" default="$2"
+	lua -e "print(require('gluon.site').$path('$default'))"
+}
+
+[ "$(lookup_site 'mesh.filter_membership_reports' 'true')" = "false" ] && exit 0
```
neocturne, line 10:
```suggestion
if [ "$(lookup_site 'mesh.filter_membership_reports' 'true')" = "false" ]; then exit 0; fi
```

package/gluon-mesh-batman-adv-brmldproxy/files/etc/hotplug.d/iface/51-gluon-mesh-batman-adv-brmldproxy
```diff
@@ -0,0 +1,37 @@
...
+wait_for_qdisc "fffe:"
+wait_for_qdisc "ffff:"
+
+# MLD reports, mesh outgoing:
+# 1) DNAT to 33:33:42:4e:f3:14
+# 2) Change ICMPv6 type to 100, keep original type in code field
+# => only send report to IPv6 multicast routers
+tc filter add dev bat0 parent fffe: protocol ipv6 prio 4221 handle 11: u32 divisor 1
+tc filter add dev bat0 parent fffe: protocol ipv6 prio 4221 u32 ht 11: match u8 131 0xff at 48 match u8 0 0xff at 49 action pedit ex munge eth dst set 33:33:42:4e:f3:14 munge offset 0x30 u16 set 0x6483 action pipe classid 1:1
+tc filter add dev bat0 parent fffe: protocol ipv6 prio 4221 u32 ht 11: match u8 132 0xff at 48 match u8 0 0xff at 49 action pedit ex munge eth dst set 33:33:42:4e:f3:14 munge offset 0x30 u16 set 0x6484 action pipe classid 1:1
```
neocturne, line 28: Maybe some helpers could be added for the repetitive parts of these commands?

package/gluon-mesh-batman-adv-brmldproxy/files/usr/sbin/gluon-brmldproxy-router-check
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+BRMLDPROXY_RTR_GROUP="ff12::424e:f314"
+
+update_router_recv() {
+	bridge mdb "$1" dev br-client port local-port grp "$BRMLDPROXY_RTR_GROUP" permanent 2> /dev/null
+}
+
+if batctl mcast_flags | grep -q "^Multicast flags (own flags: \[.*R6\.*\]"; then
```
neocturne, line 9:
```suggestion
if batctl mcast_flags | grep -q '^Multicast flags (own flags: \[.*R6\.*\]'; then
```

Also, maybe the JSON output should be used here?

package/gluon-mesh-batman-adv-brmldproxy/files/usr/sbin/gluon-brmldproxy-router-check
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+BRMLDPROXY_RTR_GROUP="ff12::424e:f314"
```
neocturne, line 3: Maybe add a comment regarding the source of this group address?

Using a group prefix derived from `prefix6` (as specified in some RFC) might also be a an alternative.

neocturne (changes requested):

T-X:
Changelog v7:
* rebased to current Gluon main (no changes)
* 51-gluon-mesh-batman-adv-brmldproxy hotplug script:
  * added "set -e"
  * replaced "{ [ ... ] } && exit 0" patterns with if/else oneliners
  * added helper functions for tc calls to avoid repetitions
* gluon-brmldproxy-router-check:
  * calculated a Gluon multicast router multicast address from domain seed and used a unicast prefix based multicast address
  * added a small comment for the RFC origin of its format
* to ease overall configuration added a patch to auto-enable a mesh-side MLD querier (candidate) if that node's client port features a multicast router

Could not use JSON for the "batctl mcast_flags" check though as the own flags in the reader are not printed in batctl's JSON mode output.

T-X:
Changelog v8:
* make linter happy about missing quotes

T-X:
Changelog v9:
*  use/check the JSON output from "batctl mj" instead of  regex'ing from "batctl mcast_flags" output (wasn't aware until a minute ago that the former existed and outputs the own batman-adv multicast flags in JSON format :D)
   * this should now have addressed all of @neocturne's previous remarks and suggestions

rotanid:
currently, further development happens over at https://github.com/freifunk-gluon/community-packages/pull/151 because more people can help with shellcheck issues.
at a later point, it might be possible to copy it to here, but currently it is NOT 100% the same thing unlike claimed.

T-X:
Changelog v10:
* rebased to current main branch (no changes)
* synced to state in gluon-community packages:
  * linting/style fixes
  * used $(PKG_NAME) in Makefile

package/gluon-mesh-batman-adv-brmldproxy/Makefile
```diff
@@ -0,0 +1,36 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-mesh-batman-adv-brmldproxy
+
+include $(TOPDIR)/../package/gluon.mk
+
+define Package/$(PKG_NAME)
```
neocturne, line 7: OpenWrt's convention is never to use `$(PKG_NAME)` like this, and Gluon follows this convention.

package/gluon-mesh-batman-adv-brmldproxy/Makefile
```diff
@@ -0,0 +1,36 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-mesh-batman-adv-brmldproxy
+
+include $(TOPDIR)/../package/gluon.mk
```
neocturne, line 5:
```suggestion
include ../gluon.mk
```

package/gluon-mesh-batman-adv-brmldproxy/files/etc/hotplug.d/iface/51-gluon-mesh-batman-adv-brmldproxy
```diff
@@ -0,0 +1,60 @@
+#!/bin/sh
+set -e
+
+if [ "$INTERFACE" != "client" ] || [ "$ACTION" != "ifup" ]; then exit 0; fi
```
neocturne, line 4: What happens when the client interface is taken down and up again repeatedly?

package/gluon-mesh-batman-adv-brmldproxy/files/etc/hotplug.d/iface/51-gluon-mesh-batman-adv-brmldproxy
```diff
@@ -0,0 +1,60 @@
...
+
+	# shellcheck disable=SC2086 # Intended splitting of $rule
+	tc filter add dev bat0 \
+		parent "$parent" prio "$prio" handle "$handle" protocol ipv6 \
+		u32 $rule
+}
+
+if [ "$(lookup_site 'mesh.filter_membership_reports' 'true')" = "false" ]; then exit 0; fi
+
+wait_for_qdisc "fffe:"
```
neocturne, line 43: What is this waiting for? Is something else creating these qdiscs?

package/gluon-mesh-batman-adv-brmldproxy/files/etc/hotplug.d/iface/51-gluon-mesh-batman-adv-brmldproxy
T-X: I think I generally get what you might be referring to. And yes I can confirm that generally there is a race, we might be in the middle of this hotplug script while things got torn down already.

I wrote a small script to rapidly call "ifdown client; ifup client" and "ifup client; ifdown client" and checked that the resulting tc filter output would always be the same for the first and empty for the second case. And so far they were for me, eventually.

I think for a final "ifdown client" there should be no issue because stopping brmldproxy will delete the according, dedicated tc-qdisc which brmldproxy had created before. Which will also flush the attached tc-filter rules.

For a race with a final "ifup client" I don't think I see a race right now either. As the script does not delete tc filter rules it created for one thing. And for parallel invocations with ACTION=ifup one of them would error out because of tc-filter additions failing while the other should set things up as intended.

I'm not sure what specific race / issue I could try to produce from here.

package/gluon-mesh-batman-adv-brmldproxy/files/etc/hotplug.d/iface/51-gluon-mesh-batman-adv-brmldproxy
T-X: Yes, brmldproxy is creating and deleting these tc-qdiscs. And in my tests it needed one more second or so before they were there while at this point in this hotplug script. I'd get errors for the tc filter add commands in this hotplug script otherwise.

But granted that's not obvious from here (especially with these nameless tc handles). I'll add a comment for that in this hotplug script.

T-X:
Changelog v11:

* Rebased to current master (small conflict with new batman-adv per hard interface hop penalty documentation)
* Avoid using $(PKG_NAME) in the Gluon/OpenWrt Makefile
* replacing "include $(TOPDIR)/../package/gluon.mk" with "include ../gluon.mk"
* adding a comment regarding why we wait for a qdisc in the hotplug script

package/gluon-mesh-batman-adv-brmldproxy/files/usr/sbin/gluon-brmldproxy-router-check
```diff
@@ -0,0 +1,50 @@
...
+}
+
+# Generates and prints an RFC3306, section 4 style network prefix based
+# multicast address from the site prefix6 with a pseudo-random group-id
+# from the domain seed.
+get_gluon_all_mc_routers_ip6() {
+	local prefix6 prefix6_len group_id
+
+	prefix6="$(lookup_prefix6)"
+	prefix6_len="$(lookup_prefix6_len)"
```
neocturne: Probably a good idea to avoid running `lookup_site` twice, and instead just assign the output of `lookup_site 'prefix6'` to `prefix6`, splitting it in the subsequent uses below.

package/gluon-mesh-batman-adv-brmldproxy/files/etc/hotplug.d/iface/51-gluon-mesh-batman-adv-brmldproxy
neocturne: Hmm yes, even with multiple instances of the script running at the same time, one of them would eventually fail and exit because it is trying to install a rule that already exists, and the other one will continue successfully.

package/gluon-mesh-batman-adv-brmldproxy/files/etc/hotplug.d/iface/51-gluon-mesh-batman-adv-brmldproxy
```diff
@@ -0,0 +1,62 @@
...
+}
+
+add_filter() {
+	local parent="$1"
+	local prio="$2"
+	local handle="$3"
+	local rule="$4"
+
+	# shellcheck disable=SC2086 # Intended splitting of $rule
+	tc filter add dev bat0 \
```
neocturne, line 34: Okay, but do we need the splitting of `rule`? It seems preferable to me to pass in separate arguments and refer to them using `"$@"` (removing the first 3 using `shift`)

T-X:
Changelog v12:
* add a general description of the use-case of brmldproxy to docs/package/gluon-mesh-batman-adv-brmldproxy.rst
* avoid calling lookup_site twice in package/gluon-mesh-batman-adv-brmldproxy/files/usr/sbin/gluon-brmldproxy-router-check
* use shift operator and `"$@"` to avoid word splitting in tc commands in /etc/hotplug.d/iface/51-gluon-mesh-batman-adv-brmldproxy

rotanid:
@neocturne this PR needs https://github.com/freifunk-gluon/packages/pull/264
so we forgot to merge and bump modules before this merge :D
we will do this now.

PR #3006, opened 2023-09-30
[v2023.1.x]: modules: update to latest HEAD

herbetom:
The openssl bump has no relevance. But it's still something I would just chuck in before a v2023.1.1.


rotanid:
> The openssl bump has no relevance. But it's still something I would just chuck in before a v2023.1.1.

why? what's the benefit?

herbetom:
Just so that we're not "behind". But it's not important.

Issue #3003, opened 2023-09-28
D-Link DAP-X1860 A1: Wifi crashes immediately after boot


citronalco:
### Bug report

**What is the problem?**
One of our users has reported non working Wifi on his D-Link DAP-X1860 (A1) with Gluon-v2023.1-13-g473ff9d
He has several of those devices. All of them work properly with Gluon 2022.1-4.

Wifi crashes immediately after boot. 100% reproducible:

dmesg:
```
...
[   43.230583] batman_adv: bat0: IGMP Querier appeared
[   43.235477] batman_adv: bat0: MLD Querier appeared
[   58.140491] mt7915e 0000:02:00.0: Message 00004eed (seq 15) timeout
[   78.620544] mt7915e 0000:02:00.0: Message 000007ed (seq 1) timeout
[  110.620479] mt7915e 0000:02:00.0: Message 00005aed (seq 2) timeout
[  131.100459] mt7915e 0000:02:00.0: Message 00005aed (seq 3) timeout
[  151.580499] mt7915e 0000:02:00.0: Message 00005aed (seq 4) timeout
[  172.060360] mt7915e 0000:02:00.0: Message 00005aed (seq 5) timeout
[  192.540121] mt7915e 0000:02:00.0: Message 00005aed (seq 6) timeout
[  213.019902] mt7915e 0000:02:00.0: Message 00005aed (seq 7) timeout
[  233.499631] mt7915e 0000:02:00.0: Message 00005aed (seq 8) timeout
[  253.979366] mt7915e 0000:02:00.0: Message 00005aed (seq 9) timeout
[  274.459085] mt7915e 0000:02:00.0: Message 00005aed (seq 10) timeout
[  294.938764] mt7915e 0000:02:00.0: Message 00005aed (seq 11) timeout
[  315.418483] mt7915e 0000:02:00.0: Message 00005aed (seq 12) timeout
[  335.898169] mt7915e 0000:02:00.0: Message 00005aed (seq 13) timeout
[  356.377828] mt7915e 0000:02:00.0: Message 00005aed (seq 14) timeout
[  376.857480] mt7915e 0000:02:00.0: Message 00005aed (seq 15) timeout
[  397.337120] mt7915e 0000:02:00.0: Message 00005aed (seq 1) timeout
(continues endlessly)
```

`iwinfo` runs very slow, taking several seconds per line. It displays:
```
client0   ESSID: "Freifunk"
[...]

PR #3011, opened 2023-10-02
[v2023.1.x] patches: openwrt: revert latest mt76 bump

herbetom:
The latest mt76 bump results in broken WLAN on the D-Link DAP-X1860 (#3003).

Reverting them works arround this for now.


Djfe:
This is missing a bump of `PKG_RELEASE`
https://github.com/freifunk-gluon/gluon/pull/3011

Reasons are similar to this:
https://github.com/openwrt/openwrt/commit/91d2ead3c3bf75b279f861ad5d11b64bf31478f0

If we don't bump `PKG_RELEASE`, then we have to ask people having build Gluon from commits newer than tag v2023.1 to run make clean.
Else the buildsystem won't rebuilt the package

wait a second 🙄
@nbd168 forgot to to bump the package as well 😅
That could explain the issues. @herbetom have you tested building from a clean repo?

rotanid:
merging as there were two confirmed working reports in #3003

PR #3014, opened 2023-10-12
Cherry-pick v2023.1.1 release notes to master

neocturne:


skorpy2009 (approved):

herbetom:
Even if it's not an backport but an forport (does that word exist, i don't tink so ... :D ), i would tend to cherry pick it with -x.

neocturne:
> Even if it's not an backport but an forport (does that word exist, i don't tink so ... :D ), i would tend to cherry pick it with -x.

Updated

PR #3015, opened 2023-10-12
CI: Fix target filters, add CI check

neocturne:
The filters generated by `generate-target-filters.py` did not match the actual filters, but the extraneous entry (added in c418b7bb16d54090e468cfb8a89d20655bd466a6) did not do what it was supposed to (as our CI builds do not use the Docker image).

Revert the change and introduce a CI check to prevent future inconsistencies.

### DIFF
--- a/.github/filters.yml
+++ b/.github/filters.yml
@@ -222,7 +222,6 @@
     "targets/targets.mk",
     "targets/x86.inc",
     "contrib/ci/minimal-site/**",
-    "contrib/docker/Dockerfile",
     "package/**"
   ],
   "bcm27xx-bcm2710": [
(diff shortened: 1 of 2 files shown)


neocturne:
Wait, why do we even want to build the x86-64 image when the Dockerfile changes? This was introduced in #2933 by @herbetom, but AFAICT, our CI builds don't use our Docker images at all

herbetom:
mhm, yeah, idk why i was thinking docker and action where in some way related there ...

maybe they shoud be, but yeah, doesn't seem to make sense

neocturne:
> mhm, yeah, idk why i was thinking docker and action where in some way related there ...
> 
> maybe they shoud be, but yeah, doesn't seem to make sense

Thanks for the confirmation, I have now changed the PR to revert the change.

PR #3021, opened 2023-10-14
Prefer packages from Gluon feeds to upstream feeds

neocturne:
In Gluon v2023.1, our tunneldigger package in freifunk-gluon/packages got replaced with a new package in openwrt/packages, causing multiple bugs. To prevent similar issues in the future, reorder the list of feeds, resulting in the following order of precedence:

- OpenWrt base
- Gluon base
- (Site feeds)
- Gluon packages
- OpenWrt packages
- OpenWrt routing

Should be backported to v2023.1.x.

### DIFF
--- a/modules
+++ b/modules
@@ -1,8 +1,11 @@
-GLUON_FEEDS='packages routing gluon'
+GLUON_FEEDS='gluon packages routing'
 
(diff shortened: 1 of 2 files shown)


rotanid:
@neocturne auto-backport failed.

PR #3027, opened 2023-10-17
new target: armsr

Kistelini:
This target supports Arm SystemReady compatible machines and allows to run an offloader on any EFI compatible ARM machine or especially on QEMU on ARM.

This depends on https://github.com/openwrt/openwrt/pull/13719, resp. its backport.

I'm testing the armv8 image on a x86 host.

  - ~[ ] Web interface~
  - ~[ ] TFTP~
  - [X] Other: dd image to disk
- ~[ ] Reset/WPS/... button must return device into config mode~
- Wired network
  - ~[ ] Association with AP must be possible on all radios~
  - ~[ ] Association with 802.11s mesh must work on all radios~ 
  - ~[ ] AP+mesh mode must work in parallel on all radios~
- LED mapping
  - Power/system LED
    - ~[ ] Should display config mode blink sequence~
  - Radio LEDs
    - ~[ ] Should map to their respective radio~
    - ~[ ] Should show activity~
  - Switch port LEDs
    - ~[ ] Should map to their respective port (or switch, if only one led present)~
    - ~[ ] Should show link state and activity~
- Outdoor devices only:
  - ~[ ] Added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`~
- Cellular devices only:
  - ~[ ] Added board name to `is_cellular_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`~
  - ~[ ] Added board name with modem setup function `setup_ncm_qmi` to `package/gluon-core/luasrc/lib/gluon/upgrade/250-cellular`~
- Docs:

### DIFF
--- a/.github/filters.yml
+++ b/.github/filters.yml
@@ -1,4 +1,24 @@
 {
+  "armsr-armv7": [
+    "targets/armsr-armv7",
+    "modules",
+    "Makefile",
+    "patches/**",
+    "scripts/**",
+    "targets/generic",
+    "targets/targets.mk",
+    "targets/armsr.inc"
+  ],
+  "armsr-armv8": [
+    "targets/armsr-armv8",
+    "modules",
+    "Makefile",
+    "patches/**",
+    "scripts/**",
+    "targets/generic",
+    "targets/targets.mk",
+    "targets/armsr.inc"
+  ],
   "ath79-generic": [
     "targets/ath79-generic",
     "modules",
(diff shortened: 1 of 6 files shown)


neocturne:
Thanks! I'm also adding some information regarding the use with qemu to our Wiki.

neocturne:
I have written down how I tested the armsr images here: https://github.com/freifunk-gluon/gluon/wiki/Virtualization

PR #3029, opened 2023-10-18
docs: fix typos, change letter case and remove whitespace characters

bobidle:
Notices some issues in the docs.

Checked some words with the help of an English dictionary, Wiktionary, Wikipedia, other external web resources and the usage in the docs itself.

I hope the suggested changes are helpful.


maurerle (approved):
I think this is a valuable contribution, which helps to streamline the docs appearance.
Even if this touches old release notes - I think it does not harm, as it does not change meaning or information somewhere..?

PR #3030, opened 2023-10-19
ci: update branch-pattern matching

blocktrron:
## [github: restrict branch-triggers on specific branches](https://github.com/freifunk-gluon/gluon/commit/b59048742300142aedca8ba0e36482564477f7d6)

In https://github.com/freifunk-gluon/gluon/pull/2997 it was pointed out the current branch-trigger matches more
branches than the release-branches. Retrict the match-pattern to only
cover release-branches

Also only trigger builds on the next-branch and not branches prefixed
with this name.

Link: https://github.com/freifunk-gluon/gluon/pull/2997#issuecomment-1771076385

---

## [github: add workflow-dispatch trigger for build workflow](https://github.com/freifunk-gluon/gluon/commit/6fee38e372906a9e892f16bf5df89da4922f28d8)

Add a workflow-dispatch trigger to the build-workflow to enable
triggering build-tests on branches not covered from the match
pattern.

---

## [github: trigger ci-build when altering ci](https://github.com/freifunk-gluon/gluon/pull/3030/commits/ae758117ab5fae995307d36493ee5afd3ee734a1)

It makes sense to run CI build-tests in case the CI itself is altered.

### DIFF
--- a/.github/workflows/build-gluon.yml
+++ b/.github/workflows/build-gluon.yml
@@ -3,8 +3,8 @@ on:
   push:
     branches:
       - master
-      - next*
-      - v20*
+      - next
+      - 'v20[2-9][0-9].[0-9].x'
   pull_request:
     types: [opened, synchronize, reopened]
 
--- a/.github/workflows/build-gluon.yml
+++ b/.github/workflows/build-gluon.yml
@@ -7,6 +7,7 @@ on:
       - 'v20[2-9][0-9].[0-9].x'
   pull_request:
     types: [opened, synchronize, reopened]
+  workflow_dispatch:
 
 permissions:
   contents: read
--- a/contrib/actions/generate-target-filters.py
+++ b/contrib/actions/generate-target-filters.py
@@ -10,6 +10,7 @@ import json
 
 # these changes trigger rebuilds on all targets
 common = [
+    ".github/workflows/build-gluon.yml",
     "modules",
     "Makefile",
     "patches/**",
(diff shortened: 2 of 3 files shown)


skorpy2009 (approved):
:+1:  Looks lovely to me! :heart_eyes:

contrib/actions/generate-target-filters.py
```diff
@@ -12,6 +12,7 @@
 common = [
     "modules",
     "Makefile",
+    ".github/workflows/build-gluon.yml",
```
neocturne, line 15: Should be sorted alphabetically

.github/workflows/build-gluon.yml
```diff
@@ -3,8 +3,8 @@ on:
   push:
     branches:
       - master
-      - next*
-      - v20*
+      - next
+      - 'v20[2-9][0-9].[0-9].x'
```
neocturne, line 7: My idea was to just have a single branch here, differing between the next, master and release branches, but I'm fine with this solution, too.

neocturne (changes requested):

.github/workflows/build-gluon.yml
blocktrron: This is the "cleaner" solution, however this adds one more task when branching.

As workflow run is limited to max. 1 for pushes on branches, it has not functional impact when keeping the references in all branches.

neocturne (approved):

PR #3032, opened 2023-10-25
ipq807x-generic: add support for Xiaomi R-3600

Djfe:
also known as Mi AIoT Router AX3600

Thanks to _madez_ for providing the hardware :)

  - [ ] Web interface
  - [ ] TFTP
  - [x] Other: [see wiki for details](https://openwrt.org/toh/xiaomi/ax3600#installation)
       - first obtain ssh access (hack)
       - then flash openwrt `initramfs-factory.ubi` via ubiformat
       - then `sysupgrade -n` to gluon
        xiaomi-ax3600
- [x] Reset/~~WPS/...~~ button must return device into config mode
      28:d1:27:fc:a6:2c
  - ~~When re-adding a device that was supported by an earlier version of Gluon, a
    the setting from the old version is not reset otherwise.~~
- Wired network
- LED mapping
  - Power/system LED
  - ~~Radio LEDs~~
  - Switch port LEDs
- ~~Outdoor devices only:~~
- ~~Cellular devices only:~~
- Docs:


Djfe:
two possible issues:
- Gluon Config mode > Advanced > Info says:
``Switch Typ: none`
- Primary MAC is ok, but device uses random macs for most interfaces
- I want to figure out how to do factory images for the target (if at all), currently building of Netgear WAX218 fails due to missing initramfs even with `factory = false,` being set.

rotanid:
@Djfe have you tested this with latest openwrt changes and do you still want to work on this support?

AiyionPrime:
@Djfe I built a firmware using your PR, wifi looks good.

I know that @cawi2001 would be available to test something, if there's something missing here.
The "IOT" triple radio is not worth waiting for (at least for this device) due to the ridiculous small possible throughput.

Deactivating it is very much fine, I'd say.

Djfe:
👍
should we continue this discussion in #3354?

PR #2799, opened 2023-02-25
Revert "ipq40xx: switch Wave2 firmware to -ct (#2541)"

AiyionPrime:
> This is a temporary measure that is related to #2692.
> 
> This reverts commit 15ef885836907b27b2c0481f7ea83d1dd504c5d5.

Issue #2692, opened 2022-11-03
ipq40xx: issue with WiFi Mesh since v2022.1.x


bschelm:
### Bug report

**Fritz Box 4040 looses mesh connection to Unifi AC Mesh**

- Since upgrade to version vh29, meshing between the two devices is not working properly
- The mesh connection only lasts for a few hours (after start) and then breaks down ~(RX or TX = 0%)~
    - UPDATE: The TQ is good on both ends. See below. (It was only a lack of data of the freifunk map, that showed "0% TQ" at first.)
- Problem started with update to vH29
- I noticed a broken connection between the two devices. Recovery is possible with command "wifi" on FB 4040
- Connection only brakes when FB4040 is > vH29, software version on AC Mesh is irrelevant. All Devices were installed from scratch (forget configuration)?

**What is the expected behaviour?**

- Connections should be stable?
- it works with vH26 and lower?

***vH26:***

https://github.com/freifunkh/site/releases/tag/vH26

_FURTHER INFORMATION FOLLOWS_

**Gluon Version: gluon-v2022.1.1**

**Site Configuration:**

https://github.com/freifunkh/site/releases/tag/vH30

**Custom patches:**

https://github.com/freifunkh/site/tree/vH30/patches

- [0001-fix-when-vlan-crc-is-00000000.patch](https://github.com/freifunkh/site/blob/vH30/patches/0001-fix-when-vlan-crc-is-00000000.patch)
- [0002-patches-uradvd-change-ip-lifetimes-to-150-300-second.patch](https://github.com/freifunkh/site/blob/vH30/patches/0002-patches-uradvd-change-ip-lifetimes-to-150-300-second.patch)
- [0003-patches-openwrt-add-TL-WR841ND-N-8M-and-16M-variants.patch](https://github.com/freifunkh/site/blob/vH30/patches/0003-patches-openwrt-add-TL-WR841ND-N-8M-and-16M-variants.patch)

AiyionPrime:
Maybe I should visit the forum more often.

https://forum.freifunk.net/t/ubiquiti-richtfunk-mit-0-100-uebertragungsqualitaet/22674/18

lemoer:
We traced this issue further. It seems to boil down to the fact that OpenWrt introduced ath10k-ct for the ipq40xx target. We tested if this issue also appears with ath10k, but it doesn't seem so (since a few days).

lemoer:
[...]

PR #3033, opened 2023-10-25
ath79-generic: add support for TP-Link EAP225-Outdoor v3

T-X:
  - [x] Web interface
   * Note: direct vendor firmware to Gluon untested, device was originally flashed from vendor fw to vanilla OpenWrt
  - [ ] TFTP: untested
  - [ ] Other: <specify>
- Wired network
- LED mapping
  - Power/system LED
  - ~~Radio LEDs~~ No radio LED available
  - ~~Switch port LEDs~~ No switch LED available
    - [ ] ~~Should map to their respective port (or switch, if only one led present)~~
    - [ ] ~~Should show link state and activity~~
- Outdoor devices only:
- ~~Cellular devices only:~~
- Docs:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -114,7 +114,7 @@ ath79-generic
   - CPE220 (v3.0)
   - CPE510 (v1.0, v1.1, v2.0, v3.0)
   - CPE710 (v1.0)
-  - EAP225-Outdoor (v1)
+  - EAP225-Outdoor (v1, v3)
   - TL-WDR3500 (v1)
   - TL-WDR3600 (v1)
   - TL-WDR4300 (v1)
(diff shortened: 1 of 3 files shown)


T-X:
Kudos to @yksflip for creating the excellent OpenWrt support for this device.

herbetom (changes requested):
I've only got experience with the v1. Hoewer, as far as I can tell they are identical (except for the ethernet phy chipset).

With the v1 there were huge problems with the 5 GHz WLAN:
- #2541
- #2692
- #2799
- #2827
- #2848

And the current -ct variant we're using isn't without issues. The WiFi randomly stops working with it. At least it doesn't take the whole device into OOM, but it's not stable.

And as far as i was told it might be possible to work arround the issues with some kind of watchdog, but an upstream fix is extremely unlikely.

I think this device should only be added as broken.

herbetom:
I don't think it's arbitrary to say that now, that we're aware of issues with the platform, until further developments, no new devices using the platform should be added (at least as non-broken). That way we we don't increase the size of the problem.

I do agree that the listed options should be evaluated and a universal way forward be found. After that it's possible to reevaluate the addition of new devices / unbreaking this one if it's merged "now".

However, the broken flag isn't particularly granular and there isn't a good way to permanently "revoke" a decision of non-broken. Installed devices would just don't get further updates. For those existing decices a "workaround" would be better. But if that happens to be for example disabling 5 GHz, it would mean that, strictly speaking, the device will no longer pass the checklist. Therefore I do think this shouldn't be merged as is.

PR #3034, opened 2023-10-25
ath79-generic: (re)add support for archer-c60

ambassador86:
This device was gone. I tried to use it with the smallbuffers wifi firmware and the device remained suprisingly stable. I whould also consider to add this at least with the broken flag, because I do not have a crowded test environment.
Test device: https://map.chemnitz.freifunk.net/#!v:m;n:50c7bf2cf5c8

  - [x] Web interface
        (`lua -e 'print(require("platform_info").get_image_name())'`) -> tp-link-archer-c60-v1
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [x] Should show activity
  - Switch port LEDs
- Docs:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -110,6 +110,7 @@ ath79-generic
   - Archer C6 (v2 EU/RU/JP)
   - Archer C7 (v2, v4, v5)
   - Archer C59 (v1)
+  - Archer C60 (v1)
   - CPE210 (v1.0, v1.1, v2.0, v3.0, v3.1, v3.20)
   - CPE220 (v3.0)
   - CPE510 (v1.0, v1.1, v2.0, v3.0)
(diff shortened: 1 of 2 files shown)


ambassador86:
I just saw in the build log, that the factory image becomes too big. Should I change my commit to only support sysupgrades? This whould allow users to install gluon after installing vanilla openwrt first.

blocktrron:
To be precise, it is not the factory image that becomes to big, it's the support-list "partition" / area:

```
2023-10-25T20:52:35.3002277Z support-list partition too big (more than 256 bytes): Success
```

This can be fixed either by expanding the partition, however I'm not sure if this has side-effects depending on the device.

I think it is fine to only build sysupgrade images for now. I just wanted to note this here, as it's not firmware image-size itself that is dangerously close to rendering the device unsupportable.

PR #3037, opened 2023-10-29
gluon-radv-filterd: replace malloc with calloc

Emantor:
Allocation of the router structure for newly detected router advertisements was done using malloc and selectively clearing some of the structure fields. However the redirect boolean was never set to 0, which means that during update_redirect, we would depend on the random initialization value of the structure.

This means that sometimes ebtables-tiny wouldn't add the correct rules for the advertised routes.


neocturne:
Thanks!

PR #3042, opened 2023-10-31
gluon-mesh-vpn-fastd: use CAKE as bandwidth-shaper

blocktrron:
Use CAKE for shaping ingress as well as egress bandwidth shaping.

CAKE allows to better shape egress as well as ingress bandwidth. It not
only performs bandwidth shaping but also flow-queueing.

This in conjunction with the batadv flow dissector allows us to utilize
CAKE for all mesh-implementations currently supported in Gluon.

### DIFF
--- a/package/gluon-mesh-vpn-fastd/luasrc/usr/lib/lua/gluon/mesh-vpn/provider/fastd.lua
+++ b/package/gluon-mesh-vpn-fastd/luasrc/usr/lib/lua/gluon/mesh-vpn/provider/fastd.lua
@@ -4,6 +4,8 @@ local site = require 'gluon.site'
 local util = require 'gluon.util'
 local vpn_core = require 'gluon.mesh-vpn'
 
+local unistd = require 'posix.unistd'
+
 local M = {}
 
 function M.public_key()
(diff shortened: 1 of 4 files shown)


T-X:
Nice! Should this then also provide equal bandwidth per flow over fastd?

Haven't looked at the batman-adv flow dissector yet, nor flows and Cake in general. Would this result in equal bandwidth per TCP session over fastd? Or even equal bandwidth per client device over fastd? Or neither? ("airtime fairness" over VPN?)

And looks like it's only applied on devices with more than 200MB RAM, correct? For all other devices no functional changes? Should this be mentioned in the commit message?

blocktrron:
@T-X while cake does introduce some fairness to concurrent TCP streams, it primarily heps with bufferbloat and thus real-time applications and sparse flows in general. But the effect can be generalized to the point you've assumed.

Yes, it performs not great on devices with low RAM and CPUs lower than mt7621 in general.

Some observations (fastd-l2tp):

 - mpc85xx-p1020: ~60 Mbit/s
 - ramips-mt7621: ~80 Mbit/s
 - ath79-generic: ~20 Mbit/s
 - mediatek-mt7981: ~220 Mbit/s

So it limits the Bandwith quite significantly, however it performs really well. We've deployed it in a location on a VDSL 100 link with a MT7986 based router.

package/gluon-mesh-vpn-fastd/luasrc/usr/lib/lua/gluon/mesh-vpn/provider/fastd.lua
```diff
@@ -25,18 +27,46 @@ function M.active()
...
-		uci:section('simple-tc', 'interface', 'mesh_vpn', {
-			ifname = vpn_core.get_interface(),
-			enabled = true,
-			limit_egress = egress_limit,
-			limit_ingress = ingress_limit,
-		})
+		if not sqm_available() or util.get_mem_total() < 200*1024 then
+			set_limit_simple_tc(ingress_limit, egress_limit)
+		else
+			set_limit_sqm(ingress_limit, egress_limit)
```
skorpy2009, line 64:
```suggestion
		if sqm_available() and util.get_mem_total() > 200*1024 then
			set_limit_sqm(ingress_limit, egress_limit)
		else
			set_limit_simple_tc(ingress_limit, egress_limit)
```

skorpy2009 (approved):

skorpy2009:
LGTM :heart_on_fire:

PR #3052, opened 2023-11-03
mediatek-filogic: add support for GL.iNet GL-MT3000

blocktrron:
  - [ ] Web interface
  - [ ] TFTP: untested
  - [x] Other: SSH from vendor firmware
- Wired network
- LED mapping
  - Power/system LED
  - ~~Radio LEDs~~ No radio LED available
  - ~~Switch port LEDs~~ No switch LED available
    - [ ] ~~Should map to their respective port (or switch, if only one led present)~~
    - [ ] ~~Should show link state and activity~~
- Outdoor devices only:
- ~~Cellular devices only:~~
- Docs:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -285,6 +285,10 @@ mediatek-filogic
 
   - WR3000 (v1)
 
+* GL.iNet
+
+  - GL-MT3000
+
 * NETGEAR
 
   - WAX220
--- a/targets/mediatek-filogic
+++ b/targets/mediatek-filogic
@@ -13,6 +13,13 @@ device('cudy-wr3000-v1', 'cudy_wr3000-v1', {
 })
 
 
+-- GL.iNet
+
+device('gl.inet-gl-mt3000', 'glinet_gl-mt3000', {
+	factory = false,
+})
+
+
 -- NETGEAR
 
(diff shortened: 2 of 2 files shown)


docs/user/supported_devices.rst
```diff
@@ -285,6 +285,10 @@ mediatek-filogic
 
   - WR3000 (v1)
 
+* GL.iNet
+
+  - MT3000
```
herbetom, line 290:
```suggestion
  - GL-MT3000
```

herbetom:
AFAIK is flashing via the Gl.iNet Web-UI also an option for the initial installation.

herbetom:
Is it possible to get the fan improvement Commit for this Device backported to openwrt-23.05?

https://github.com/openwrt/openwrt/commit/5a603c7a316c3fb4bae4c7cb5c666232864126f2

blocktrron:
In case it is verified it works with the device, a backport might be accepted by a target maintainer.

The patch however is not strictly necessary, as it only tweaks the fan curve.

herbetom:
In my testing the untweaked fan curve resulted in constant starts and stopps of the Fan.
I wouldn't expect that this is good for the hardware.
It's probably "fine for now". But with the reality in mind that Gluon Versions tend to stick arround for a bit, it would be better to fix this.

PR #3054, opened 2023-11-05
build: implement flexible site-selection system

blocktrron:
This is currently lacking documentation.
Opening as a draft to obtain feedback whether or not this is moving in the right direction.

---

Implement a flexible system for handling site-defined features as well as packages.

This system is inspired by the existing feature-system and allows for a more flexible approach for selecting specific packages for devices.

Features are now defined in a `features` file in the site-root. The same goes for packages.

These files are sequentially evaluated and the device-package list is evaluated for each device independently.


neocturne:
I like the idea of using a Lua DSL here a lot, but I think it would be nicer if features and packages could be configured in the same file, as they are often related. That would also allow to set related conditional features and packages under a single `when`.

blocktrron:
Great idea! I'll see how this can be done.

blocktrron:
@neocturne I've adressed your request, does this look better?

docs/site-example/packages
```diff
@@ -0,0 +1,17 @@
...
+	Feature('autoupdater'),
+	Feature('ebtables-filter-multicast'),
+	Feature('ebtables-filter-ra-dhcp'),
+	Feature('ebtables-limit-arp'),
+	Feature('mesh-batman-adv-15'),
+	Feature('mesh-vpn-fastd'),
+	Feature('respondd'),
+	Feature('status-page'),
+	Feature('web-advanced'),
+	Feature('web-wizard')
```
neocturne, line 12: How about making this:
```
Features {
	'autoupdater',
	'ebtables-filter-multicast',
	'ebtables-filter-ra-dhcp',
	'ebtables-limit-arp',
	'mesh-batman-adv-15',
	'mesh-vpn-fastd',
	'respondd',
	'status-page',
	'web-advanced',
	'web-wizard',
}
```
(as Lua conveniently allows us to omit the `()` when there is a single table argument)

I'm also wondering if we really need the `default` and `when` functions - maybe something like a toplevel `features`/`packages` function and `if not device_class('tiny')` would also work? I have not really through about the the implementation though.

Finally, is it possible to remove features or packages with this syntax (as supported in `site.mk` with a `-` prefix)?

blocktrron:
@neocturne Please see the updated PR, is this more to your liking?

Also yes, negation is still supported, as this is done in `handle_pkgs` which is not modified.

neocturne:
Looking quite nice now!

Two remarks (obviously bikeshedding - I'm fine with either solution, just stating my preference):

- Is there a reason you kept the `()` in addition to the `{}` for the `Packages` and `Features` arguments in the example?
- The capitalization of `Packages` and `Features` looks a bit weird IMO

blocktrron:
@neocturne 

No special reason for keeping the `()`. I can remove them at wish

For the capitalization - My intention was to keep modifying functions with a leading Uppercase while function for the logic evaluation have lowercase first letters.

blocktrron:
@neocturne I'll opt with all lowercase and maybe revising the docs. Anything from the implementation side i should revise?

scripts/image_customization_lib.lua
```diff
@@ -0,0 +1,132 @@
...
+end
+
+local function evaluate_device(env, dev)
+	local selections = {}
+	local funcs = {}
+	local device_overrides = {}
+
+	local function add_elements(element_type, element_list)
+		for _, element in ipairs(element_list) do
+			if not selections[element_type] then
```
neocturne, line 31: This check could be omitted if `selections` is initialized with `feature` and `package` fields.

scripts/image_customization_lib.lua
```diff
@@ -0,0 +1,132 @@
...
+				selections[element_type] = {}
+			end
+
+			selections[element_type][element] = true
```
neocturne, line 35: The order of keys is unspecified, so this will break with negated elements. The elements will have to be stored as values, not as keys.

scripts/image_customization_lib.lua
```diff
@@ -0,0 +1,132 @@
...
+	end
+
+	function funcs.DisableFactory()
+		add_override('disable_factory', true)
+	end
+
+	function funcs.device(device_names)
+		assert(
+			type(device_names) == 'table',
+			'Incorrect use of device(): pass a list of device-names as argument')
```
neocturne, line 69:
```suggestion
			'Incorrect use of device(): pass a list of device names as argument')
```

scripts/image_customization_lib.lua
```diff
@@ -0,0 +1,132 @@
...
+			end
+		end
+
+		return false
+	end
+
+	function funcs.target(target, subtarget)
+		assert(
+			type(target) == 'string',
+			'Incorrect use of target(): pass a target-name as first argument')
```
neocturne, line 83:
```suggestion
			'Incorrect use of target(): pass a target name as first argument')
```

scripts/target_config_lib.lua
```diff
@@ -102,20 +86,25 @@ local function feature_packages(features)
...
+	site_features = image_customization_lib.get_selection('feature', env, dev_info)
+	site_features = compact_list(site_features, false)
+
+	-- Create List from packages inherited from features
+	feature_inherited_pkgs = feature_packages(site_features)
 
-	local features = site_vars(string.format('$(GLUON_FEATURES) $(GLUON_FEATURES_%s)', class))
-	features = compact_list(split(features), false)
+	-- Read list of packages from site
+	site_packages = image_customization_lib.get_selection('package',  env, dev_info)
```
neocturne, line 102: We are getting the selections of `feature` and `package` right next to each other. It would be more efficient to have a `get_selections`, so `evaluate_device` needs to run only once here.

neocturne (changes requested):

neocturne:
One more idea: How about allowing to specify negated features and packages as `-'name'` instead of `'-name'`? I kinda like the separation of the operator from the name.

A very simple implementation might look like this (using the `-` in the strings would be an implementation detail that we could change in the future):
```lua
getmetatable('').__unm = function(s)
	return '-' .. s
end
```

scripts/target_lib.lua
```diff
@@ -210,18 +235,33 @@ local function as_table(v)
...
-	options = merge(default_options, options)
+local function disable_factory_image(device_info)
+	if device_info.options.deprecated and env.GLUON_DEPRECATED ~= 'full' then
+		return true
+	end
 
-	if not want_device(image, options) then
-		return
+	local overrides = get_device_overrides(device_info)
+	if overrides["disable_factory"] ~= nil and overrides["disable_factory"] then
```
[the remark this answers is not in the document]
blocktrron: Nope, changed. Thanks!

blocktrron:
I have revised the documentation to accurately reflect the image customization feature. As far as I can tell, this pull request is now ready.

Makefile
```diff
@@ -35,6 +35,14 @@ ifneq ($(GLUON_BRANCH),)
   GLUON_AUTOUPDATER_ENABLED ?= 1
 endif
 
+ifneq ($(GLUON_FEATURES)$(GLUON_FEATURES_standard)$(GLUON_FEATURES_tiny),)
+  $(warning *** Warning: GLUON_FEATURES has been deprecated, please use the image-customization framework instead.)
```
neocturne, line 39: Maybe we should just write "image-customization file" rather than framework, like the docs do

contrib/ci/olsr-site/image-customization
```diff
@@ -0,0 +1,20 @@
...
+	'ebtables-limit-arp',
+	'mesh-olsrd',
+	'mesh-vpn-fastd',
+	'respondd',
+	'status-page',
+	'web-advanced',
+	'web-wizard',
+}
+
+packages({
```
neocturne, line 14: A lot of the examples are inconsistent regarding use of `{` vs. `({`. I'd prefer to use `{` to be consistent with package lists in target files.

scripts/image_customization_lib.lua
```diff
@@ -0,0 +1,135 @@
...
+	end
+
+	function funcs.packages(packages)
+		add_elements('package', packages)
+	end
+
+	function funcs.broken(broken)
+		assert(
+			type(broken) == 'boolean',
+			'Incorrect use of Broken(): has to be a boolean value')
```
neocturne, line 47:
```suggestion
			'Incorrect use of broken(): has to be a boolean value')
```

scripts/image_customization_lib.lua
```diff
@@ -0,0 +1,135 @@
...
+
+	if not file_exists(get_customization_file_name(env)) then
+		return return_object
+	end
+
+	local eval_result = evaluate_device(env, dev)
+	return_object = {
+		features = eval_result.selections['feature'],
+		packages = eval_result.selections['package'],
+	}
```
neocturne, line 121: If the fields in the `selections` table were called `features` and `packages`, `get_selections` could directory return `eval_result.selections` without renaming the fields.

Makefile
neocturne: Also, deprecated implies that it still works, which is not the case - it should be called obsolete instead.

It might also make sense to make this a hard error like obsolete `site.conf` entries, as updating the site is required to keep things working as intended.

neocturne (changes requested):

scripts/image_customization_lib.lua
```diff
@@ -0,0 +1,135 @@
...
+	local f = io.open(file)
+	if not f then
+		return false
+	end
+	f:close()
+	return true
+end
+
+local function get_customization_file_name(env)
+	return env.GLUON_SITEDIR .. '/image-customization'
```
maurerle, line 13: This is basically a lua file, right?
Shouldn't we also name it as one, so that highlighting and code linting works better?

grische: Would also make auto-formatting easier

contrib/ci/olsr-site/image-customization
grische: While we are at it, we had examples where we needed brackets:
https://github.com/ffac/site/commit/8cddeadc890b9e6b12ec3a347b875d5846919f79#diff-d79e4f28ca1dc9caa17a36e10310542744140aa58c9ecfd00256c741bca2d7afR133-R135

So I am not sure if a simple `{ 'a' }` is better than a `({ 'a' })` considering that some people use vars like `(list)`.

grische:
After building all images, it seems there is an issue somewhere with the PR:
- https://github.com/freifunkMUC/site-ffm/actions/runs/7253374267/job/19759838511#step:6:1
- https://github.com/freifunkMUC/site-ffm/actions/runs/7253374267/job/19759840204#step:5:1

This one looks like a configuration error on my side:
https://github.com/freifunkMUC/site-ffm/actions/runs/7253374267/job/19759839156#step:6:1

blocktrron:
@grische See https://github.com/grische/site-ffm/blob/feature/image-customizations/image-customization#L192

The single items provided as parameter have to be wrapped in a table (`{ 'pkg-name }`). This is why your mt7621 build fails.

contrib/ci/olsr-site/image-customization
blocktrron: I think I've fixed all inconsistencies with the latest push. Please check if you still find any.

blocktrron:
Updated the PR. I have not build-tested the current state yet.

docs/multidomain-site-example/image-customization.lua
```diff
@@ -0,0 +1,20 @@
...
+	'ebtables-limit-arp',
+	'mesh-batman-adv-15',
+	'mesh-vpn-fastd',
+	'respondd',
+	'status-page',
+	'web-advanced',
+	'web-wizard',
+}
+
+packages{
```
grische, line 14: Small note:
It seems all other examples are using a space before the bracket

maurerle:
It would be quite handy to be able to call:
`is_cellular_device` and `is_outdoor_device` from platform.lua in the image-customization.lua
https://github.com/freifunk-gluon/gluon/blob/9a8ee5733365c0a64531c539a578133bbc6429ac/package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua#L25

blocktrron:
@maurerle I would like to hold this back for further improvement. I see this as possible, but we would ahve to build this meta-information from the same source. Possibly the same way as to handling flash-sizes as well as RAM.

maurerle:
Sure, I am fine with that.
For now I helped myself here - but I don't like the idea of maintaining that list myself:
https://github.com/ffac/site/blob/eb6f98e8ec05957163c3d24115d337cb21ecb97e/image-customization.lua#L25-L36

I just build and tested Firmware from this PR - looks good. I am happy with merging it the way it is.

PR #3060, opened 2023-11-09
ath79-generic: add support for D-Link DAP-3662

s-2:
```
Specifications:
 * QCA9557, 16 MiB Flash, 128 MiB RAM, 802.11n 2T2R
 * QCA9882, 802.11ac 2T2R
 * 2x Gigabit LAN (1x 802.11af PoE)
 * IP68 pole-mountable outdoor case
```

This device has both switch ports declared as LAN in OpenWrt, which results in no VLAN configuration being generated.
As a result, the automagical assignment of the first LAN port as WAN in gluon would cause both ports to be WAN.
To prevent this, an OpenWrt patch is introduced to set the PoE input port to WAN manually. Is there a better way to fix this in gluon without using a patch here? 

The device has been running fine so far: https://hannover.freifunk.net/karte/#/de/map/e46f134b1500

----------------------------------------------------------------------------

  - [x] Web interface
  - [ ] TFTP
  - [ ] Other: <specify>
- Wired network
  - [*see above] must have correct port assignment (WAN/LAN)
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [ ] Should show activity
  - Switch port LEDs
- Outdoor devices only:
- Cellular devices only:
- Docs:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -33,6 +33,7 @@ ath79-generic
   - DAP-1330 A1 [#lan_as_wan]_
   - DAP-1365 A1 [#lan_as_wan]_
   - DAP-2660 A1 [#lan_as_wan]_
+  - DAP-3662 A1 [#lan_as_wan]_
   - DIR-505 A1 [#lan_as_wan]_
   - DIR-505 A2 [#lan_as_wan]_
   - DIR-825 B1
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/020-interfaces
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/020-interfaces
@@ -43,6 +43,10 @@ local lan_ifname = iface_exists(lan_interfaces)
 local wan_ifname = iface_exists(wan_interfaces)
 
 if platform.match('ath79', 'generic', {
+	'dlink,dap-3662-a1',
+}) then
+	lan_ifname, wan_ifname = 'eth0.1', 'eth1.2'
+elseif platform.match('ath79', 'generic', {
 	'tplink,cpe210-v1',
 	'tplink,cpe210-v2',
 	'tplink,cpe510-v1',
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua
@@ -25,6 +25,7 @@ end
 function M.is_outdoor_device()
 	if M.match('ath79', 'generic', {
 		'devolo,dvl1750x',
+		'dlink,dap-3662-a1',
 		'librerouter,librerouter-v1',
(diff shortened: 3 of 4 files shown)


herbetom:
Please check out https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-core/luasrc/lib/gluon/upgrade/020-interfaces for how to reconfigure interfaces with gluon.

As mentioned in the checklist the PoE Input should be WAN.

s-2:
Dropped patch, tried settings in `020-interfaces`.
Now it seems I can't reach the device on any of the ports in config mode at all. Settings were probably guessed wrong, maybe I can waste some further time on this device by the end of this or next week.

herbetom:
`eth0.1` and `eth1.2` seems quite odd. I would expect that it's one of the following:

- `eth0.1` and `eth0.2`
- `eth0.1` and `eth1.1`

Or of course something completely diffrent.

s-2:
iirc, in config mode both wan and lan ports should expose the web config, so I would have expected it would work for one interface at least, so it could be further narrowed down, but I will try further variations. The builds took way too long yesterday, I eventually needed to get some sleep...

I'm still not sure if it would work this way, it seems without any modification it behaves like there is only a single port with an unmanaged switch attached to it, i.e. there is only eth0. So maybe it really is just `eth0.1` and `eth0.2` here.

herbetom:
The Config Mode is only available via LAN.

If the AP doesn't already have some config to seperate the interfaces, the 020-interfaces file won't work. Either DSA (currently not with ath79), seperate PHYs or swconfig would be needed AFAIK. My bad.

My way forward would be to check how 
it's done for 'ubnt,unifiac-mesh-pro' and 'ubnt,unifiac-pro'. From the looks of it in 020-interfaces there probably is some swconfig config in OpenWrt to seperate the ports. I would hope the same is possible for the DAP-3662. But haven't checked.
Adding some config like that in OpenWrt (interfaces roles wouldn't have to be reassigned in OpenWrt, that's something that could be done inside 020-interfaces, Gluon only needs the ports refrencable) would be most likely the preferred way from the Gluon side as it doesn't require a patch.

As a note: eth0.2 typically means that you've got an interface that sends out with the vlan ID 2 tagged traffic via eth0. If you want to use VLANs to seperate interfaces on an managed Switch Chip you need to configure (that's the before mentioned swxonfig part) the chip to do so.

s-2:
Changed to `'eth0.1', 'eth0.2'`, indeed no difference.

I will try to find the device I had originally used for development, with UART attached (don't want to open the IP68 case here to see what's happening, also unbricking is way more comfortable via UART).

> From the looks of it in 020-interfaces there probably is some swconfig config in OpenWrt to seperate the ports.

So we'd basically mimic the behavior of OpenWrt here, and just call the `ucidef_add_switch` script from gluon instead?

PR #3062, opened 2023-11-10
ath79-mikrotik: add support for MikroTik wAPR-2nD (wAP R)

splumhoff:
[X] Must be flashable from vendor firmware
  - [ ] Web interface
  - [X] TFTP
  - [ ] Other: <specify>
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [X] Should show activity
  - Switch port LEDs
- Outdoor devices only:
- Cellular devices only:
- Docs:

### DIFF
--- a/targets/ath79-mikrotik
+++ b/targets/ath79-mikrotik
@@ -1,3 +1,4 @@
 include 'mikrotik.inc'
 
 device('mikrotik-routerboard-951ui-2nd-hap', 'mikrotik_routerboard-951ui-2nd')
+device('mikrotik-routerboard-wapr-2nd', 'mikrotik_routerboard-wapr-2nd')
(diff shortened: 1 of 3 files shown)


targets/ath79-mikrotik
```diff
@@ -1,3 +1,4 @@
 include 'mikrotik.inc'
 
 device('mikrotik-routerboard-951ui-2nd-hap', 'mikrotik_routerboard-951ui-2nd')
+device('mikrotik-routerboard-wapr-2nd', 'mikrotik_routerboard-wapr-2nd')
```
herbetom, line 4:
```suggestion

device('mikrotik-routerboard-wapr-2nd', 'mikrotik_routerboard-wapr-2nd')
```

ich würde da eine Zeile Abstand lassen

herbetom:
sorting should be checked, i think we've had A-z0-9

splumhoff:
> sorting should be checked, i think we've had a-z0-9

I found no guideline on this, but checked in targets/ath79-generic and it seems to be 0-9a-z, e.g. the GL.iNET-Area is sorted:

- gl.inet-6416
- gl.inet-gl-ar150
- ...

PR #3063, opened 2023-11-13
ci: workflow_dispatch: allow usage of a custom site

herbetom:
The intention is to offer an easy way to build useful gluon images via GitHub Actions.

This should lower the barrier towards testing PRs a bit.

### DIFF
--- a/.github/workflows/build-gluon.yml
+++ b/.github/workflows/build-gluon.yml
@@ -8,6 +8,19 @@ on:
   pull_request:
     types: [opened, synchronize, reopened]
   workflow_dispatch:
+    inputs:
+      site_url:
+        description: 'Gluon Site Repo to use'
+        type: string
+      site_branch:
+        description: 'Gluon Site Branch to use'
+        type: string
+      autoupdater_enabled:
+        description: 'Gluon autoupdater activation State'
+        type: boolean
+      autoupdater_branch:
+        description: 'Gluon autoupdater default branch'
+        type: string
 
 permissions:
   contents: read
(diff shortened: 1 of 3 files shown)


.github/workflows/build-gluon.yml
```diff
@@ -8,6 +8,19 @@ on:
...
   workflow_dispatch:
+    inputs:
+      site_url:
+        description: 'Gluon Site Repo to use'
+        type: string
+      site_branch:
+        description: 'Gluon Site Branch to use'
+        type: string
+      autoupdater_enabled:
+        description: 'Gluon autoupdater activation State'
```
blocktrron, line 19: "Enable the autoupdater"

.github/workflows/build-gluon.yml
```diff
@@ -8,6 +8,19 @@ on:
...
+        type: boolean
+      autoupdater_branch:
+        description: 'Gluon autoupdater default branch'
```
blocktrron, line 22: "Default Branch to use for the autoupdater"

.github/workflows/build-gluon.yml
```diff
@@ -8,6 +8,19 @@ on:
   pull_request:
     types: [opened, synchronize, reopened]
   workflow_dispatch:
+    inputs:
+      site_url:
+        description: 'Gluon Site Repo to use'
```
blocktrron, line 13: URL of the repository containing the Site

.github/workflows/build-gluon.yml
```diff
@@ -8,6 +8,19 @@ on:
   pull_request:
     types: [opened, synchronize, reopened]
   workflow_dispatch:
+    inputs:
```
blocktrron, line 11: This misses an option for cases where the site-root is not equal with the root of the repository. In case the site is stored in a subdirectory of the repository, this currently does not work.

blocktrron:
In addition, this currently lacks the option of modifying the strategy for which targets to build.

In case a core-package is modified, this might not build the target you require to test on.

blocktrron:
Will the open issues be addressed or can this be closed?

herbetom:
I plan on adressing them. Hoewer, i'm not sure when. Could be today, could be in a few weeks. Depends on when i've got the time and energy to do so.

Issue #2939, opened 2023-08-22
Read-Only filesystem on Ubiquiti airmax devices


blocktrron:
### Bug report

**What is the problem?**

Several Ubiquiti airmax devices report a Read-only filesystem with OpenWrt 23.05

Upstream ticket: https://github.com/openwrt/openwrt/issues/12882

**What is the expected behaviour?**

The rootfs_data partition is not read-only

**Gluon Version:**

Gluon based on OpenWrt 23.05

**Site Configuration:**
n/a

**Custom patches:**
n/a

blocktrron:
Removing this ticket for the upcoming v2023.2 milestone, as the devices were removed from Gluon. 


Keeping this ticket open for visibility.

rotanid:
@blocktrron this seems to be solved upstream?

blocktrron:
DO you have a commit as reference?

rotanid:
@blocktrron f024f4b1b0380b3b2e18115bd8e4f35393fccc70 was mentioned here https://github.com/openwrt/openwrt/pull/14237

herbetom:
I could validate te fix on the NanoBeam 5AC 19 (XC). Hoewer that device is currently in Darmstadt and it's currently raining which doesn't help with my motivation to do so right now.

Hoewer, in theory we could validate it and reintroduce support before the next release.

blocktrron:
@rotanid I've seen this commit, I'd like to have this verified before reverting though.

nrbffs:
We have M5 Loco XM (5x) and M5 Loco XW (3x). Is the fix already in master? What needs to be done to verify this?

rotanid:
https://github.com/openwrt/openwrt/commit/c55aaa7c9a98e7c0a5e1ea8293a534dc5b395cf3

@blocktrron shouldn't this be part of gluon already which would render this issue fixed?

nrbffs:
As far as I understand, the fix from OpenWRT commit c55aaa7c9a ("ath79: generic: disable SPI-NOR write protect unconditionally") is now in Gluon main. 

To drive this issue forward, I would do the following:

- make a build of latest gluon main, with reverted a0f5b4e9468 ("ath79-generic: remove Ubiquiti AirMax targets (#3067)")
- flash it onto a Ubiquiti Nanostation Loco M (XW) or Ubiquiti Nanostation M (XW)  - does not matter if M2 or M5
[...]

PR #3067, opened 2023-11-24
ath79-generic: remove Ubiquiti AirMax targets

blocktrron:
Remove due to ticket #2939

These devices risk being bricked in a way that can not be recovered without running full TFTP or serial recovery, as the flash ends up in a read-only state with no way of upgrading to a fixed version.

Thus, remove the support for the time being. Don't just flag them as BROKEN, as this would deliberately brick these boards when installating such an image.

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -128,10 +128,6 @@ ath79-generic
 
 * Ubiquiti
 
-  - NanoBeam 5AC 19 (XC)
-  - NanoBeam M5 (XW)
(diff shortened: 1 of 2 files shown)


herbetom:
"disable" instead of "remove" would be more appropriate if the devices are commented out. Might minimize potential for misunderstandings by someone only skimming the commits. But no hard opinion.

blocktrron:
This is functionally equivalent.

skorpy2009 (approved):

PR #3080, opened 2023-12-01
ci: build Gluon in own Docker image

blocktrron:
Build the CI integration test in our own docker container. This way, we can make sure Gluon builds actually succeed there.

This also has the advantage of becoming independent from the host version of GitHubs CI runners.


skorpy2009 (approved):
LGTM

Issue #2980, opened 2023-09-12
NWA50AX Pro regularily crashes during normal operation on master


nrbffs:
### Bug report

**What is the problem?**

NWA50AX Pro crashes regularily (every 1-12 hours). The following stracktrace was collected in ramoops:

```
<1>[34681.152965] Unable to handle kernel access to user memory outside uaccess routines at virtual address 0000000000000008
<1>[34681.152984] Mem abort info:
<1>[34681.152986]   ESR = 0x0000000096000005
<1>[34681.152990]   EC = 0x25: DABT (current EL), IL = 32 bits
<1>[34681.152994]   SET = 0, FnV = 0
<1>[34681.152997]   EA = 0, S1PTW = 0
<1>[34681.153000]   FSC = 0x05: level 1 translation fault
<1>[34681.153004] Data abort info:
<1>[34681.153006]   ISV = 0, ISS = 0x00000005
<1>[34681.153009]   CM = 0, WnR = 0
<1>[34681.153013] user pgtable: 4k pages, 39-bit VAs, pgdp=0000000041781000
<1>[34681.153018] [0000000000000008] pgd=08000000440e7003, p4d=08000000440e7003, pud=08000000440e7003, pmd=0000000000000000
<0>[34681.153032] Internal error: Oops: 96000005 [#1] SMP
[...]

PR #3084, opened 2023-12-03
mac80211: avoid crashing on invalid band info

blocktrron:
Frequent crashes have been observed on MT7916 based platforms. While the root of these crashes are currently unknown, they happen when decoding rate information of connected STAs in AP mode. The rate-information is associated with a band which is not available on the PHY.

Check for this condition in order to avoid crashing the whole system. This patch should be removed once the roout cause has been found and fixed.

Closes: #2980
Link: https://github.com/freifunk-gluon/gluon/issues/2980


blocktrron:
@nrbffs If you wish, provide me with your real-name as well as e-mail address for `Suggested-by:` and `Reported-by:`

PR #3085, opened 2023-12-04
[v2022.1.x] modules: update to latest HEAD

blocktrron:


herbetom:
This will probably cause problems with the DAP-X1860. #3011 would have to be backported.

I don't quite remeber but there might have been other issues.

With v2023.1 beeing by some (at least me :sweat_smile:) described as an v2022.1.5 (just with different minimal version requirements) i think bumping efforts would be more useful for the v2023.1.x branch. If the goal is to get an Gluon with an up to date openwrt-22.03 version :see_no_evil:.

blocktrron:
> This will probably cause problems with the DAP-X1860. https://github.com/freifunk-gluon/gluon/pull/3011 would have to be backported.

I'll look into this.

> i think bumping efforts would be more useful for the v2023.1.x branch

I'm willing to invest time into moving this forward, i don't understand your point.

herbetom:
> I'm willing to invest time into moving this forward, i don't understand your point.

- v2022.1.x is currently bumped further then v2023.1.x
- v2023.1.x is supported, v2022.1.x isn't
- there weren't any major breking changes between the two series
- searching for and backporting neccesary fixes/workarrounds is riskier then using what has been already tested
- working on/with v2023.1.x benefits everyone using an currently supported Gluon Version

You're of course free to do whatever. I just don't see (besides us at ffda beeing able to release it as an v2.6.x instead of having to bump to v2.8.x (and numbers are kinda "free")) whats the benefit of investing time into v2022.1.x when the work has already been done in v2023.1.x (+a few other improvements but no real drawbacks).

But do what you think is right. It's just not what I would do.

rotanid:
>     * there weren't any major breking changes between the two series

repeating this doesn't make it true.
the supported upgrade path changed.

PR #3086, opened 2023-12-05
github: release docker container for next branch

blocktrron:
Build and release a docker container for the next branch.


herbetom (approved):
With #3080 i guess the only other option would be to carry this patch exclusively inside next.

PR #3090, opened 2023-12-08
[WIP] scripts: allow specifying a minimum flash and RAM

grische:


blocktrron:
@grische I'd rather have a list a site can maintain in their repo to exclude specific devices. This can then be used for multiple reasons. I see your approach as an indirection of this goal. What do you think?

blocktrron:
Superseded by `image-customization` framework. Closing.

PR #3091, opened 2023-12-10
build: allow overriding GLUON_SITE_VERSION

blocktrron:
In case a community does not use Git for versioning it's site or using a different versioning scheme that do not have annotated tags on the master branch.

### DIFF
--- a/docs/user/getting_started.rst
+++ b/docs/user/getting_started.rst
@@ -239,7 +239,7 @@ GLUON_RELEASE
 
 GLUON_SITE_VERSION
   Version of the site configuration. This string is displayed in the config mode
-  and `gluon-info`. If not set, Gluon generates a version-string using `git describe`
+  and `gluon-info`. If unset, Gluon generates a version string using `git describe`
   on the site folder.
 
 GLUON_TARGET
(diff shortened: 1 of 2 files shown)


blocktrron:
cc @herbetom

docs/user/getting_started.rst
```diff
@@ -237,6 +237,11 @@ GLUON_RELEASE
   is available. The same GLUON_RELEASE has to be passed to ``make`` and ``make manifest``
   to generate a correct manifest.
 
+GLUON_SITE_VERSION
+  Version of the site configuration. This string is displayed in the config mode
+  and `gluon-info`. If not set, Gluon generates a version-string using `git describe`
+  of the site folder.
```
herbetom, line 243:
```suggestion
  in the site folder.
```

or something like that. of sounds wrong to me

herbetom (changes requested):
looks good

docs/user/getting_started.rst
blocktrron: should be `on`

docs/user/getting_started.rst
```diff
@@ -237,6 +237,11 @@ GLUON_RELEASE
   is available. The same GLUON_RELEASE has to be passed to ``make`` and ``make manifest``
   to generate a correct manifest.
 
+GLUON_SITE_VERSION
+  Version of the site configuration. This string is displayed in the config mode
+  and `gluon-info`. If not set, Gluon generates a version-string using `git describe`
```
neocturne, line 242:
```suggestion
  and `gluon-info`. If not set, Gluon generates a version string using `git describe`
```

PR #3102, opened 2023-12-13
gluon-core: preserve interface MAC-address configuration

blocktrron:
This preserve configuration of MAC-addresses for each interface in case OpenWrt sets them using netifd.

This is done by preserving all device-configuration which does not define a bridge or the MAC-address of a bridge.

Closes ##3100

### DIFF
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/020-interfaces
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/020-interfaces
@@ -90,8 +90,15 @@ else
 	sysconfig.single_ifname = lan_ifname or wan_ifname
 end
 
+-- Delete all UCI device esctions of type 'bridge'
+-- as well as the ones starting with 'br-'.
+-- Preserve all others to apply MAC address stored in UCI
+uci:foreach('network', 'device',function(s)
+	if s.type == 'bridge' or s.name:match('^br-') then
+		uci:delete('network', s['.name'])
+	end
+end)
 
-uci:delete_all('network', 'device')
 uci:delete_all('network', 'interface')
 
 uci:save('network')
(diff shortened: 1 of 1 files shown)


package/gluon-core/luasrc/lib/gluon/upgrade/020-interfaces
```diff
@@ -90,8 +90,15 @@ else
 	sysconfig.single_ifname = lan_ifname or wan_ifname
 end
 
+-- Delete all UCI device esctions of type 'bridge'
```
skorpy2009, line 93:
```suggestion
-- Delete all UCI device sections of type 'bridge'
```

blocktrron:
Tested on ASUS TUF-AX4200

Djfe:
thank you ❤️

package/gluon-core/luasrc/lib/gluon/upgrade/020-interfaces
```diff
@@ -90,8 +90,15 @@ else
...
 end
 
+-- Delete all UCI device sections of type 'bridge'
+-- as well as the ones starting with 'br-'.
+-- Preserve all others to apply MAC address stored in UCI
+uci:foreach('network', 'device',function(s)
+	if s.type == 'bridge' or s.name:match('^br-') then
+		uci:delete('network', s['.name'])
+	end
+end)
```
neocturne, line 100: Using `delete` in `foreach` has caused weird breakage in the past (including segfaulting Lua at some point). `delete_all` with a function argument must be used instead, see `gluon-respondd/luasrc/lib/gluon/upgrade/400-respondd-firewall`

blocktrron: I'll check that

Issue #3023, opened 2023-10-15
[v2023.1.x] Fritzbox 7520/7530 bricked after update to v2023.1.1


maurerle:
### Bug report

**What is the problem?**

The FB7520 and FB7530 do not boot properly since some commit between the tag v2023.1 and v2023.1.1

This bug exists only on this branch, a build from master is working properly.

I don not have serial logs for this bug - the broken firmware can be currently found here:
http://firmware.freifunk-aachen.de/firmware/gluon-ffac-v2023.1.1-1-avm-fritz-box-7530-sysupgrade.bin

The Power LED stays lit in green and the device does nothing.
No Wifi, no Gateway, no ip address..
Power cycle does not help either.

Someone reported that a Computer attached to the switch can talk to another computer on the switch, when the device is on - but the device itself does not boot properly.

We have 13 devices which have this problem.

**What is the expected behaviour?**
The device should work as normal - as bevor v2023.1.1 or on current master.

**Gluon Version:**
v2023.1 and v2023.1.1

**Site Configuration:**
https://github.com/ffac/site

## Recovery
To recover, one has to use the respective AVM tool:
For 7520:
https://download.avm.de/fritzbox/fritzbox-7520/deutschland/recover/
For 7530:
https://download.avm.de/fritzbox/fritzbox-7530/deutschland/recover/
(works with playonlinux/WINE)

And can then run the (improved) fritz flash tool from here:
https://github.com/maurerle/fritz-tools/blob/master/fritzflash.py

Download an initramfs and a working firmware
And run:
`sudo python3 fritzflash.py --dev eno2 --initramfs ./openwrt-22.03.2-ipq40xx-generic-avm_fritzbox-7530-initramfs-fit-uImage.itb --sysupgrade ./gluon-ffac-v2023.1.0-3-avm-fritz-box-7520-sysupgrade.bin`
or execute this command without sudo in an admin prompt on windows.

Djfe:
on windows there are definitely a few more differences, I'll update this post, once I have tested a working command.
One difference: you execute `py` instead of `python3`

Djfe:
Finally got around to doing the upgrade on a box while being connected via serial
The box hangs at `Starting kernel ...`
https://pastebin.com/0NiNpjqr

Next step:
I'm building branch openwrt-22.03 to see whether it's affected or not.
Then I'm going to do a git bisect.

[...]

PR #3104, opened 2023-12-13
[v2023.1.x] ipq40xx-generic: remove support for AVM FRITZ!Box 7520+7530

herbetom:
remove due to issue https://github.com/freifunk-gluon/gluon/issues/3023

There seem to be issues hindering an updated 7520/7530 to boot properly.

As far as we know the issue only affects openwrt-22.03.

Since the issue results in an unusable device remove support entirely for now and don't just mark it as broken.

---

Should be merged before the next v2023.1.x release at the latest if there is no solution by then.

cc @maurerle @grische

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -196,8 +196,6 @@ ipq40xx-generic
 * AVM
 
   - FRITZ!Box 4040 [#avmflash]_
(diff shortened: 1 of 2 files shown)


grische (approved):

maurerle:
Can be closed as the issue does not appear in v22.03.6 and later

PR #3109, opened 2023-12-18
gluon-mesh-vpn-tunneldigger: drop package

herbetom:
The Gluon Tunneldigger protocol handler hasn't received the care and attention a core package, especially a VPN protocol, should.

Due to the works in #2102 it's no longer required to maintain VPN handlers in the Gluon Core and they can be instead included via a package feed.

Users who wish to continue using a supported core Gluon package are encouraged to take a look at fastd in conjunction with it's null@l2tp method. It offers roughly the same performance (it's using the same principle of operation) and supports connections via IPv6 as a bonus.

In general it should be possible to change the VPN used protocol asynchronously with a normal update.

If Gluon users wish to continue using tunneldigger the recomendation is to participate in the development of tunneldigger inside our community-packages: https://github.com/freifunk-gluon/community-packages/pull/77

---

Don't merge this before v2023.2.x has been branched of.


maurerle (approved):
From looking at it, it seems fine, I do not have remarks

skorpy2009 (approved):

package/gluon-mesh-vpn-core/luasrc/lib/gluon/upgrade/500-mesh-vpn
skorpy2009, line 1: Not needed when you rebase to https://github.com/freifunk-gluon/gluon/pull/3113 when it is merged

herbetom: Thanks for looking into this. I already had the suspicion that we could do so but hadn't yet followed up on it. :+1:

PR #3114, opened 2023-12-20
add package gluon-libustream-tls

grische:
This package allows to easily select the correct libustream implementation without all communities having the need to manually change their included packages.

### DIFF
--- a/package/gluon-tls/Makefile
+++ b/package/gluon-tls/Makefile
@@ -1,0 +1,12 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-tls
+
+include ../gluon.mk
+
+define Package/gluon-tls
+  DEPENDS:=+libustream-mbedtls +ca-bundle
+  TITLE:=Dummy Package to select the current TLS implementation
+endef
+
+$(eval $(call BuildPackageGluon,gluon-tls))


neocturne:
Do we want to simply call this `gluon-tls`, so you can add `tls` to GLUON_FEATURES?

maurerle:
@neocturne wouldn't it then also make sense to add `ca-bundle` too?

neocturne:
> > @neocturne wouldn't it then also make sense to add `ca-bundle` too?
> 
> add it to what?

To the dependencies of the new package.

Yes, sounds like a good idea to me.

rotanid:
> > > @neocturne wouldn't it then also make sense to add `ca-bundle` too?
> > 
> > 
> > add it to what?
> 
> To the dependencies of the new package.
> 
> Yes, sounds like a good idea to me.

why? its not always needed and uses additional space

neocturne:
> > > > @neocturne wouldn't it then also make sense to add `ca-bundle` too?
> > > 
> > > 
> > > add it to what?
> > 
> > 
> > To the dependencies of the new package.
> > Yes, sounds like a good idea to me.
> 
> why? its not always needed and uses additional space

So, would you install a custom root certificate then? Are any communities actually doing that?

But if we don't want to make the `ca-bundle` a hard dependency, we can include it by default with the `tls` feature through `package/features`.

rotanid:
ah, sorry for my confusion. i got it now!

so i agree with @neocturne to rename it to "gluon-tls" and i also agree to include ca-bundle as a dependency.
@grische can you do these adjustments, please?

i guess it would also be useful to backport this to v2023.1 (of course with dependency adjusted to wolfssl)

PR #3115, opened 2023-12-20
github: build-gluon: cancel obsolete in progress workflows for PRs

herbetom:
Goal is to cancel Worfklows for Pull Request if they become obsolte due to new changes.
Without this workflows will continue running, wasting resources and delaying the run of relevant workflows.
This situation will typically occour if a author of a PR notices an error/typo/missing bit and pushes new changes to the branch.

This won't affect concurrent workflows for otherwise triggered workflows (push, workflow_dispatch, tag, ...).

Relevant Links:
- https://docs.github.com/en/actions/using-jobs/using-concurrency
- https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

### DIFF
--- a/.github/workflows/build-gluon.yml
+++ b/.github/workflows/build-gluon.yml
@@ -12,6 +12,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref || github.run_id }}-${{ github.event.pull_request.number || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   changed:
     permissions:


.github/workflows/build-gluon.yml
```diff
@@ -12,6 +12,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref || github.run_id }}-${{ github.event.pull_request.number || github.run_id }}
```
blocktrron, line 16: This would also trigger on branches, which should not be interrupted. I'm fine with cancelling PR's though.

herbetom: For branches `github.event.pull_request.number` isn't set, therefore it falls back to `github.run_id` and that will be diffrent for diffent commits inside a branch.

blocktrron: fair, i did not parse that correctly.

skorpy2009 (approved):

PR #3126, opened 2023-12-28
ath79-generic: add support for Sophos AP15

kpanic23:
Basically the same device as a Sophos AP55

  - [ ] Web interface
  - [ ] TFTP
  - [X] Other: Vendor software, theoretically TFTP+UART, but UART header is not accessible externally
  - Device does _not_ have any buttons. 
  Warning: Make sure to set up SSH remote access, otherwise the only option is to open up the case and use the UART header for console access.
- Wired network
- LED mapping
  - Power/system LED
- Docs:

### DIFF
--- a/targets/ath79-generic
+++ b/targets/ath79-generic
@@ -380,6 +380,12 @@ device('siemens-ws-ap3610', 'siemens_ws-ap3610', {
 
 -- Sophos
 
+device('sophos-ap15', 'sophos_ap15', {
+	packages = ATH10K_PACKAGES_QCA9880,
+	factory = false,
+	broken = true, -- no button and no external console port
+})
+
 device('sophos-ap100', 'sophos_ap100', {
 	packages = ATH10K_PACKAGES_QCA9880,
 	factory = false,


targets/ath79-generic
```diff
@@ -380,6 +380,11 @@ device('siemens-ws-ap3610', 'siemens_ws-ap3610', {
 
 -- Sophos
 
+device('sophos-ap15', 'sophos_ap15', {
+	packages = ATH10K_PACKAGES_QCA9880,
+	factory = false,
+})
```
blocktrron, line 386: As Serial is not external but  requires disassembly, this device would be marked as broken (add a comment about this circumstance so we can be more granular about this in the future).

kpanic23: Well, you don't need serial to flash the device. Serial is only needed if you forgot to setup an SSH key/password because due to the device having no button you wouldn't be able to bring it into config mode again.

The device still can be flashed without opening it via the free of cost Home Edition of the Sophos Firewall, it enables you to SSH into the stock firmware and flash the firmware with mtd from there. Exactly the same as with the other Sophos APs.

Therefore I'm unsure if it really needs the broken flag...

blocktrron: @kpanic23 Please add the broken flag, as @herbetom explained the re-entry of the setup-mode is problematic (And will be extended in the future to allow such cases.

docs/user/supported_devices.rst
```diff
@@ -111,6 +111,7 @@ ath79-generic
 
 * Sophos
 
+  - AP15
```
herbetom, line 114:
```suggestion
```

broken devices aren't listed in the docs

kpanic23: Ah, my bad!

blocktrron (approved):

PR #3127, opened 2023-12-29
Improvements to image customization docs

neocturne:
I was a bit unhappy with our current docs, as they didn't include an example. I've also reworded the section a bit and added a link to the example section.

### DIFF
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -836,6 +836,12 @@ site.conf
 .. literalinclude:: ../site-example/site.conf
   :language: lua
 
+image-customization.lua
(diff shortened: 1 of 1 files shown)


neocturne:
Note: I'd prefer to add even more examples (maybe a short example snippet for each of the functions), but this will take me some more time and I don't want to hold up the release more than necessary.

herbetom:
I've created a small wiki entry: https://github.com/freifunk-gluon/gluon/wiki/Image-customization

maurerle (approved):
I think that this is a good addition.

PR #3143, opened 2024-01-04
add device_has_usb function for image-customisation

grische:
This adds the has_usb (manually created based on OpenWRT's ToH) to all devices and adds a function to image-customisation to check if a device has USB support.

I would continue doing this for all targets, not just ath79-generic.


blocktrron:
We've spoken about strucctured metadata at 37c3 and @aparcar is already working on this. I don't think we should track this metadata within Gluon, same for the RAM sizes.

neocturne:
@blocktrron @aparcar Does structured hardware metadata exist for OpenWrt now, or what is the status? Is the toh still be best place to obtain this information?

PR #3148, opened 2024-01-05
Image customization: add include() function

neocturne:
Add a simple way to include image customization Lua snippets.
Can be used to split long files, to include the same options multiple
times in different conditional branches, or even to pass values back to
the including file using return.
    
Relative paths are interpreted relative to the site root (where
image-customization.lua is located). Relative includes would become
confusing when subdirectories are involved (as they are still interpreted
relative to the site root, and proper tracking of current directory for
each include seems fairly complex for a very niche use case), so we simply
reject this case for now; this way, we can implement this however we
want in the future if deemed necessary, without a breaking change.

Suggested by @T-X

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,11 +36,11 @@ ifneq ($(GLUON_BRANCH),)
 endif
 
 ifneq ($(GLUON_FEATURES)$(GLUON_FEATURES_standard)$(GLUON_FEATURES_tiny),)
-  $(error *** Warning: GLUON_FEATURES has been obsolete, please use the image-customization.lua file instead.)
+  $(error GLUON_FEATURES is obsolete, please use the image-customization.lua file instead)
 endif
 
 ifneq ($(GLUON_SITE_PACKAGES)$(GLUON_SITE_PACKAGES_standard)$(GLUON_SITE_PACKAGES_tiny),)
-  $(error *** Warning: GLUON_SITE_PACKAGES has been obsolete, please use the image-customization.lua file instead.)
+  $(error GLUON_SITE_PACKAGES is obsolete, please use the image-customization.lua file instead)
 endif
 
 GLUON_AUTOUPDATER_ENABLED ?= 0
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -790,6 +790,39 @@ disable_factory()
       disable_factory()
     end
 
+include(path)
+  Includes another image customization file. Relative paths are interpreted relative to the site
+  repository root.
+
+  .. code-block:: lua
+    :caption: target-settings.lua
+
+    features {'wireless-encryption-wpa3'}
+
+  .. code-block:: lua
+    :caption: image-customization.lua
+
+    if target('x86', '64') then
+      include 'target-settings.lua'
+    end
+
+  Values returned from the included file become the return value of ``include``:
+
+  .. code-block:: lua
+    :caption: matches-device.lua
+
+    return device {
+      'openmesh-a40',
+      'openmesh-a60',
+    }
+
+  .. code-block:: lua
+    :caption: image-customization.lua
+
+    if include('matches-device.lua') then
+      packages {'iperf3'}
+    end
+
 Technically, the image customization file is evaluated once for each device, allowing
 to make use of regular Lua *if* statements for device-specific configuration as
 can be seen in the examples.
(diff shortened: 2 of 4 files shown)


blocktrron:
@neocturne Would you mind implementing a sample implementation for the example-site? From the documentation one could assume it works similar to sourcing a shell-script. I think an example is the best way of demonstrating a best-practice use.

The implementation itself looks good :+1:

neocturne:
Makes sense, I wanted to add more examples to the image customization docs anyways.

blocktrron:
Shall we wait for these to arrive? I think the PR itself is fine as-is. I'd just wait for a backport to v2023.2 for docs.

neocturne:
I think there is no hurry regarding this PR, and I can work in the doc examples in the next days.

Makefile
```diff
@@ -36,11 +36,11 @@ ifneq ($(GLUON_BRANCH),)
 endif
 
 ifneq ($(GLUON_FEATURES)$(GLUON_FEATURES_standard)$(GLUON_FEATURES_tiny),)
-  $(error *** Warning: GLUON_FEATURES has been obsolete, please use the image-customization.lua file instead.)
```
AiyionPrime, line 39: Does `error` do this, while `warning` in line 33 doesn't?

docs/user/site.rst
```diff
@@ -699,39 +699,131 @@ setup and a device-specific alteration.
...
+  .. code-block:: lua
+    :caption: target-settings.lua
+
+    features {'wireless-encryption-wpa3'}
+
+  .. code-block:: lua
+    :caption: image-customization.lua
+
+    if target('x86', '64') then
+      include 'target-settings.lua'
```
AiyionPrime, line 806:
```suggestion
      include('target-settings.lua')
```

other functions like broken use braces in this line as well as in the description https://github.com/freifunk-gluon/gluon/pull/3148/files#diff-9385c61083912d3fa14a4e9de1769b9487f71f7d52801cdf2a66c7f3216e435eR791

Is this deviation intentional?
Looking at line 820, we might use braces here as well.

scripts/image_customization_lib.lua
```diff
@@ -62,23 +72,49 @@ local function evaluate_device(env, dev)
...
+	function funcs.include(path)
+		assert(
+			type(path) == 'string',
+			'incorrect use of include(): path expected as first argument')
+
+		if string.sub(path, 1, 1) ~= '/' then
+			assert(
+				string.find(path, '/') == nil,
+				'incorrect use of include(): including files from subdirectories is unsupported')
+			path = env.GLUON_SITEDIR .. '/' .. path
```
AiyionPrime, line 111: Just for completeness; path might be anything and further sanitization of this (to e.g. prevent directory traversal out of the sitedir) is out of scope, right?
I think leaving this as is is fine, just wanted to highlight it.

scripts/image_customization_lib.lua
```diff
@@ -62,23 +72,49 @@ local function evaluate_device(env, dev)
...
+		end
+		local f = assert(loadfile(path))
+		setfenv(f, funcs)
+		return f()
```
AiyionPrime, line 115: This looks scary powerful :D

scripts/image_customization_lib.lua
neocturne: Yes, sanitization is kind of pointless, you can already include arbitrary code in site.mk.

Makefile
neocturne: Yes,
```makefile
$(warning Foo.)
$(error Bar)
```
results in
```
Makefile:1: Foo.
Makefile:2: *** Bar.  Stop.
```
... which means, that the `***` in the `$(error)` is also redundant, will fix.

docs/user/site.rst
neocturne: I generally like the syntax without parentheses for functions like `include` and `require` (we use it a lot for `require`), but I don't have a strong opinion in this case.

Makefile
neocturne: I have made some more improvements to these messages.

docs/user/site.rst
AiyionPrime: no strong opinion on my end either, just wanted to ensure you did it on purpose.

AiyionPrime (approved):
Looks better, thanks :)

Issue #3020, opened 2023-10-14
UniFi 6 LR status page - internal server error 500


Djfe:
### Bug report

**What is the problem?**
the status page on a UniFi 6 LR errors like this (probably affects all version, this one is a v3, PR pending)
see http://[2a03:2260:3006:115:6222:32ff:fe6a:37d5]/cgi-bin/status

> 500 Interner Serverfehler
> 
> Entschuldigung, auf dem Server ist ein unerwarteter Fehler aufgetreten.
> 
> Failed to execute dispatcher target for entry '/'.
> The called action terminated with an exception:
> /usr/lib/lua/gluon/web/template.lua:43: Failed to execute template 'status-page'.
> A runtime error occurred: /usr/lib/lua/gluon/wireless.lua:26: attempt to concatenate field '?' (a nil value)

**What is the expected behaviour?**
Status page should render

**Gluon Version:**
https://github.com/freifunk-gluon/gluon/commit/1b6ddac3c608961399c4877198dd552c12c3f07b (master 2 weeks ago)

**Site Configuration:**
https://github.com/ffac/site/tree/v2023.2.0-5

**Custom patches:**
quite a few, but none affecting gluon-core
https://github.com/ffac/site/tree/v2023.2.0-5/patches

I started debugging the issue, line 26 is effectively this one:
https://github.com/freifunk-gluon/gluon/blob/a1370198de864c56dda892f8c5ff80f317bb9095/package/gluon-core/luasrc/usr/lib/lua/gluon/wireless.lua#L40

The issue appears to stem from a line above:
`local phy_idx = v:match('^phy(%d+)$')`

If I understand it correctly, it expects wiphys to contain `phy` in their name in order to extract the correct phy_idx.
This device has (atleast since OpenWrt 23) the phy names `wl0` and `wl1`

```
root@DN-Viktoria-Str-46:/sys/devices# find . | grep ieee80211
./platform/18000000.wmac/ieee80211
./platform/18000000.wmac/ieee80211/wl0
./platform/18000000.wmac/ieee80211/wl0/uevent
./platform/18000000.wmac/ieee80211/wl0/power
./platform/18000000.wmac/ieee80211/wl0/power/runtime_active_time
./platform/18000000.wmac/ieee80211/wl0/power/runtime_status
./platform/18000000.wmac/ieee80211/wl0/power/autosuspend_delay_ms
./platform/18000000.wmac/ieee80211/wl0/power/runtime_suspended_time
[...]

PR #3155, opened 2024-01-09
remove hardcoded 'phy' to get radio device name

maurerle:
This works for the  Ubiquiti UniFi 6 LR v3 and my COVR-X1860 - so it handles the wl name properly.

I am not very happy with this fix, because I do not know if this catches all possible cases.

This just adds a second counter to count the wl interfaces separately.
It is probably only ever the case to have either phy or wl, but I am not sure.

#### previous workaround

Before that, I tested using the helper function from netif, which did not work on the COVR X1860 (because it does not have a `wlan` section in the board.json)
```
local handle = io.popen(string.format('. /lib/netifd/wireless/mac80211.sh && get_board_phy_name "%s"', path))
return handle:read("*line")
```
This took the information from netifd which reads the board.json, which does not look that beautiful either.

relates #3020


neocturne:
I think a better solution would be to do what `mac80211.sh` does - which is calling `iwinfo nl80211 phyname`. I have a patch locally that implements this using libiwinfo-lua, but I still need to test my fix.

neocturne:
I have opened a draft PR for my untested patch: https://github.com/freifunk-gluon/gluon/pull/3223

PR #3162, opened 2024-01-17
gluon-ebtables-limit-arp: stop for autoupdater

blocktrron:
Stop and start the gluon-arp-limiter service before the network is brought down when commencing update.

Start the service after the network is brought back up when aborting.

### DIFF
--- a/package/gluon-ebtables-limit-arp/files/usr/lib/autoupdater/abort.d/15start-arp-limiter
+++ b/package/gluon-ebtables-limit-arp/files/usr/lib/autoupdater/abort.d/15start-arp-limiter
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+. /lib/gluon/autoupdater/lib.sh
+
+# Start after network
+start_enabled gluon-arp-limiter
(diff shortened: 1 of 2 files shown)


rotanid:
can you add the reason for the change?

blocktrron:
We are currently investigating cases where the autoupdater would start the sysupgrade procedure which never finishes.

The node is in a state with stopped network, firmware placed in `/tmp` but no updater running.

In this state, the arp-limiter is spamming the log, complaining about a missing bat0 interface, rotating away potential informations about the reason for the failing autoupdater process.

skorpy2009 (approved):

maurerle:
so this is eventually fixes #3171

blocktrron:
No, this is a different problem.

T-X (approved):
gluon-arp-limiter does not have a proper shutdown procedure / signal handler right now. So the ebtables rules it has added will stay even after its shutdown. However when gluon-arp-limit is then started again through abort.d, it will flush the according two tables during its initalization, before adding its stuff again. So generally looks fine to me, as long as it's shutdowned stated does not stay for too long.

Also: Tested these two, added scripts on a Plasmacloud PA2200. ebtables rules look fine to me afterwards.

PR #3186, opened 2024-02-09
Omit obsolete manifest-aliases / cleanups

Djfe:
This PR is mostly about some cleanups.
I propose dropping manifest aliases for unsupported upgrade paths (devices running pre OpenWrt 22.03)
This makes target files better to read and makes the manifests smaller.

I also make OEM naming a bit more consistent across targets and documentation and add a few new lines here and there.
And I gave two existing manifest-aliases their respective code comment (last OpenWrt release with that model name).
There were further manifest-aliases without code comments but I was able to drop those. (introduced for OpenWrt 21 or prior)

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -71,7 +71,7 @@ ath79-generic
 
   - LibreRouter v1 [#missing_radios]_
 
-* Netgear
+* NETGEAR
 
   - WNDR3700 (v1, v2)
   - WNDR3800
@@ -177,7 +177,7 @@ ath79-nand
   - GL-AR750S
   - GL-XE300
 
-* Netgear
+* NETGEAR
 
   - WNDR3700 (v4)
   - WNDR4300 (v1)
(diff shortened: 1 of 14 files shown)


docs/user/supported_devices.rst
```diff
@@ -316,7 +316,7 @@ mediatek-filogic
 
   - UniFi 6 Plus
 
-* ZyXEL
+* ZYXEL
```
blocktrron, line 319: Wrong: 
image

Djfe: interesting, you are right. I had looked at their Logos mostly not continous text
[old](https://de.m.wikipedia.org/wiki/Datei:Zyxel_Logo.svg)
[new](https://commons.m.wikimedia.org/wiki/File:Z_logo_1.png#mw-jump-to-license)

Djfe: I addressed your suggestion in my latest rebase

PR #3190, opened 2024-02-16
ath79-generic: add support for D-Link DAP-2680

s-2:
The DAP-2680 is an AC1750 Wave 2 business Access Point (not previously added along with DAP-2660 since there was an issue with ath10k firmware package choice, meanwhile gluon points to a version of OpenWrt [that includes the fix](https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=0dc056eb66e1b3a4a6797bdf91f7362df6ced9c3)).

```
Specifications:
 * QCA9558, 16 MiB Flash, 256 MiB RAM, 802.11n 3T3R
 * QCA9984, 802.11ac Wave 2 3T3R
 * Gigabit LAN Port (AR8035), 802.3at PoE
```

https://hannover.freifunk.net/karte/#!/de/map/f48ceb8a2cb0

  - [x] Web interface
  - [-] TFTP
  - [x] Other: recovery webinterface (keep reset button pressed during power-on, until upload started)
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [-] Should map to their respective radio
    - [- Should show activity
  - Switch port LEDs
    - [-] Should map to their respective port (or switch, if only one led present)
    - [-] Should show link state and activity
- Outdoor devices only:
  - [-] Added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -43,6 +43,7 @@ ath79-generic
   - DAP-1330 A1 [#lan_as_wan]_
   - DAP-1365 A1 [#lan_as_wan]_
   - DAP-2660 A1 [#lan_as_wan]_
+  - DAP-2680 A1 [#lan_as_wan]_
   - DIR-505 A1 [#lan_as_wan]_
   - DIR-505 A2 [#lan_as_wan]_
   - DIR-825 B1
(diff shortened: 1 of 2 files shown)


s-2:
> I think it would be good to add only one model per PR.

Looking back at #2414, indeed there was only that single device added back then, sorry for the confusion with OpenWrt (where similar devices within the same PR, but of course separate commits, are more common since the PRs would be dropped in the long term anymways).

Issue #3191, opened 2024-02-17
vxlan layers learns broadcast / multicast addresses from incoming packets


blocktrron:
When a Gluon node receives a vxlan packet where the encapsulated ethernet source-address is a broadcast or multicast address, the Gluon node adds this source-address to it's fdb. Subsequently, all broadcast packets and therefor batman-adv OGMs are exclusively sent to this node as unicast.

![image](https://github.com/freifunk-gluon/gluon/assets/16539439/b4efb8a6-30dc-4b10-89be-3fffb1202521)

```
de12d533e19496afbf99ea4486dd6000000000a61140fe8000000000000094afbffffe99ea44fe80000000000000dc12d5fffe33e1949acb12b500a60000080000003418f400ffffffffffff96afbf99ea444305000f31000aaf98e1de12d533e193de12d533e19400ce002c0401001c014700039abfab1780010000d383d63080000000c750872600000000060200041900000002010000000f3000bcdd26e99629afe64f0bde12d533e19400c2002c0401001c01fd0003419655e18001000008aa28c680000000a2b78f4a00000000060200041900000002010000
```

I'm unsure if this behavior is okay (or how it happens organically). However, as this allows a DoS on a wired-mesh network, we should at least apply a hotfix. I will open a pull-request for this.

I assume meshing over wireguard using vxlan should not be affected from my understabding, as it's a 1:1 peer connection. Someone with more insight into this architecture should evaluate if this is also affected.

blocktrron:
I will keep this issue open until we've backported the Fix to v2023.2 and v2023.1.

PR #3192, opened 2024-02-17
net vxlan: don't learn non-unicast L2 destinations

blocktrron:
See Gluon #3191

Link: https://github.com/freifunk-gluon/gluon/issues/3191


T-X:
Sounds like a valid approach. For batman-adv we explicitly don't learn from ethernet multicast source addresses (while I think we still forward them?): https://elixir.bootlin.com/linux/v6.7.5/source/net/batman-adv/soft-interface.c#L247

For the Linux bridge I'm not quite sure. I'm only aware that it disallows learning from the zero MAC address and then even refrains from forwarding it: https://elixir.bootlin.com/linux/v6.7.5/source/net/bridge/br_fdb.c#L1225. (while batman-adv does learn zero MAC address sources like any other unicast MAC address...) I haven't yet tested what the bridge does for multicast ethernet source addresses.

---

My general take for any protocol/driver on layer 2 is: Learn from any unicast ethernet source MAC address, but must not learn from the zero MAC address. Must not learn any multicast ethernet source MAC address.

Should (probably) drop / refrain from forwarding frames with a multicast source address / should consider them as broken/invalid. (Or at least for the IPv4/IPv6 multicast ranges, 01:00:5E:00:00:00-01:00:5E:FF:FF:FF 33:33:00:00:00:00-33:33:FF:FF:FF:FF?)

Might be worth checking with 802.1Q-2022, should specify how switches/bridges should handle this (does anyone have access to it?).

blocktrron:
@T-X 

For the learning process, it seems like it should not be learned:

image

For the forwarding process depicted in the spec, no step talks about dropping non-unicast source-addresses:

image

PR #3193, opened 2024-02-18
docs: add preferred English to guidelines

bobidle:
After several chats about the preferred English (American or British) used in Gluon I decided to add it to the documentation.

My suggestion is to add American or British English as preferred English language to the documentation.

My commit adds American English as the preferred one because of the feedback I received from the chats.

### DIFF
--- a/docs/dev/basics.rst
+++ b/docs/dev/basics.rst
@@ -83,3 +83,6 @@ To that end we provide a ``.editorconfig`` configuration, which is supported by 
 of the editors out there.
 
 If you add Lua scripts to gluon, check formatting with ``luacheck``.
+
+Only English should be used for naming, comments, documentation and so on.
+The use of American English is preferred.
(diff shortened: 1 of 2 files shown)


blocktrron:
> After several chats about the preferred English (American or British) used in Gluon I decided to add it to the documentation.

Where did they happen? For me this feels like overcomplicating things for no benefit, especially since none of us are native speakers of either dialect.

maurerle (commented):
This is about consistency when using
optimization/optimisation, behavior/behaviour or neighbor/neighbour (former being American - latter one being British).

I think it would be good to have some consistency in Code and documentation.

Though we currently have the british `gluon-neighbour-info` and `neighbor` in the i18n files  - as well as a mixed usage of behavior and behaviour.

As we do not enforce this and probably should not either - I would not like to add these guidelines and just live with the drawback of having to `grep` twice when searching code which contains one of these words.

blocktrron:
Closing due to lack of intereset.

I still think we do not have the proficiency and should invest our resources on more important technical matters.

Issue #3059, opened 2023-11-08
batman-adv: Multicast eventually not forwarded from/to clients on supernodes on v2023.x?


maurerle:
### Bug report

In FFAC I experienced broken ipv6 neighbor discovery after some time for predominantly wifi clients on gluon v2023.x (not entirely sure about affected versions - see below).

First I suspected a wrong supernode configuration being the case.
But this issue also appeared on the local mesh ipv6 (being `fdac::/64` in FFAC), inbetween the mesh.
Pinging `fe80:` addresses from the supernode to the client did work fine though.

The IP-Addresses of the Gluon-Nodes themselves have never been affected yet.
I could also reproduce this issue on a client behind a gluon v2022.1.x - connected to a supernode running batman-adv 2023.x (v2023.0 or v2023.2 don't remember it)

It does not affect IPv4 traffic, only ipv6 neighbor discovery (neighbor advertisement/soliciation).
This results in broken ipv6 for clients, which leads to a buggy web connection.

An eventually related problem could be #2854
I did not see Neighbor discovery packages of the router sent to the client on the gluon-node in bat0 though. So this looks differently (or should I have looked into `local-node`/`any`?).

### Fixes - Disabling multicast optimizations

This issue has been discussed on IRC. @T-X asked me to disable [multicast optimizations in batman](https://www.open-mesh.org/doc/batman-adv/Multicast-optimizations.html) using `batctl multicast_forceflood 1` which did help.

To get it working throughout the mesh domain I had to set `multicast_forceflood` on all nodes inbetween to receive the multicast from my laptop

1. I could only ping the client's `fdac::` from its nextnode
2. I did set multicast_forceflood on its nextnode (not sure if required)
3. I did set multicast_forceflood on the mesh-vpn node connected through mesh-lan
4. Now I could also ping client's `fdac::` from the mesh-vpn node - but not from my laptop
5. the supernode already had `multicast_forceflood` set
6. after I did set `multicast_forceflood` on my nextnode too - I could ping the clients `fdac::`

### Reproducing

To try to reproduce, I would use a >=v2023.1 node, connect through wifi as client and ping ipv6 addresses or fdac addresses of other nodes in the same mesh domain/segment. If this works throughout multiple >3 days, I don't think you have this issue.
It helps to use freifunk as the main connection while being in home-office, to monitor that.

### Affected Versions/Devices

[...]

PR #3200, opened 2024-02-25
docs readme: Gluon v2023.2.2 

blocktrron:
Let's do a Gluon v2023.2.2 and resolve the potentially mesh-breaking VxLAN bug.

### DIFF
--- a/docs/releases/v2023.2.2.rst
+++ b/docs/releases/v2023.2.2.rst
@@ -1,0 +1,43 @@
+Gluon 2023.2.2
+==============
+
+Bugfixes
+--------
+
+* Fixed a bug where nodes connected using VXLAN would lose connectivity to all but one neighbor 
+  (`#3191 <https://github.com/freifunk-gluon/gluon/issues/3191>`_)  
+
+* fastd-l2tp VPN did not re-generate the flow-label for the inner-tunnel traffic,
+  breaking QoS mechanisms 
+  (`#3203 <https://github.com/freifunk-gluon/gluon/issues/3203>`_)  
+
+* Fixed silent-failures when image-customization files were not valid
+  (`#3218 <https://github.com/freifunk-gluon/gluon/issues/3218>`_)
+
+* Multicast-snooping did stop working after extended uptime periods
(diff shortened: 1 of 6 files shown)


neocturne:
As the VXLAN bug is not new, I'd rather wait until we have Linux 5.15.149, which contains the patch from #3176 (@T-X Is that the same issue as #3059?)

T-X:
@neocturne I'm not quite sure that #3176 and #3059 are related. The former should only lead to increased multicast overhead, not packet loss (unless without snooping the air were fully congested?). At least in Lübeck with ~200 nodes I didn't notice any packet loss with the #3176 issue.

neocturne:
> @neocturne I'm not quite sure that #3176 and #3059 are related. The former should only lead to increased multicast overhead, not packet loss (unless without snooping the air were fully congested?). At least in Lübeck with ~200 nodes I didn't notice any packet loss with the #3176 issue.

Ah, in that case, the patch seems low-priority and I'm fine with releasing without it.

docs/releases/v2023.2.2.rst
```diff
@@ -0,0 +1,33 @@
+Gluon 2023.2.2
+==============
+
+Bugfixes
+--------
+
+* Fixed a bug where nodes connected using VxLAN would lose connectivity to all but 1 neighbors
```
neocturne, line 7:
```suggestion
* Fixed a bug where nodes connected using VXLAN would lose connectivity to all but 1 neighbors 
```

blocktrron:
@neocturne These patches are not related, @T-X is right with that.

docs/releases/v2023.2.2.rst
```diff
@@ -0,0 +1,43 @@
+Gluon 2023.2.2
+==============
+
+Bugfixes
+--------
+
+* Fixed a bug where nodes connected using VXLAN would lose connectivity to all but 1 neighbors
```
herbetom, line 7:
```suggestion
* Fixed a bug where nodes connected using VXLAN would lose connectivity to all but one neighbor
```

PR #3202, opened 2024-02-26
kernel net l2tp: drop flow hash on forward

blocktrron:
Drop the flow-hash of the skb when forwarding to the L2TP netdev.

This avoids from the L2TP qdisc using the flow-hash from the outer packet, which is in case of a VPN connection always identical.

This allows queuing mechanisms to correctly identify flows within the mesh-vpn on platforms which include L4 into the flow-hash receives from hardware. This is (at least) the case for MediaTek Filogic platforms.


rotanid:
obsolete due to #3205 3205

PR #3208, opened 2024-02-29
mac80211: add AQL support for broadcast packets

blocktrron:
## Rationale

A Gluon node usually has multicast and broadcast filtering towards the mesh, however the local traffic is still unfiltered.
This can be an issue when a rouge node sends excessive broadcast packets. As broadcast WiFI traffic is currently not
under AQL limits, this can easily flood the AP, effectively DoSing the wireless network.

Backport the commit which introduces AQL limits on non-unicast WiFi queues to ease this issue.

## Original commit message

Excessive broadcast traffic with little competing unicast traffic can easily
flood hardware queues, leading to throughput issues. Additionally, filling
the hardware queues with too many packets breaks FQ for broadcast data.

Fix this by enabling AQL for broadcast packets.

### DIFF
--- a/patches/openwrt/0009-mac80211-add-AQL-support-for-broadcast-multicast-packets.patch
+++ b/patches/openwrt/0009-mac80211-add-AQL-support-for-broadcast-multicast-packets.patch
@@ -1,0 +1,241 @@
+From: Felix Fietkau <nbd@nbd.name>
+Date: Fri, 9 Feb 2024 20:47:39 +0100
+Subject: mac80211: add AQL support for broadcast/multicast packets
+
+Should improve performance/reliability with lots of mcast packets
+
+Signed-off-by: Felix Fietkau <nbd@nbd.name>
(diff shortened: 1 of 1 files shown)


T-X:
> A Gluon node usually has multicast and broadcast filtering towards the mesh, however the local traffic is still unfiltered.

One correction, this does not seem 100% correct to me: We have WiFi AP isolation on and the bridge hairpin mode on, which reflects packets back to the WiFi AP inerface.

However in between, bridge multicast snooping is still applied. So if someone floods to a multicast group with no listener locally, it won't be retransmitted locally.

Would adding this patch potentially break things like RTP audio/video multicast streams?

blocktrron:
@T-X You can definitely flood broadcast traffic on the AP interface (client0 to client0), I've did this multiple times to trigger an mt76 bug.

No, this does not break RTP, at least not when the RTP traffic does not exceed it's fair share of airtime (in which case this patch does break a stream which would otherwise work by interrupting anybody else on the radio)

PR #3212, opened 2024-03-03
mt76 mt7915: restart target on MCU timeout

blocktrron:
Please test this. It is far from perfect. 

The message-wait is not aborted, thus the recovery process might be blocking for a good minute. However, it can recover the MCU crashes  reported in #3154 

----

This commit adds a patch to restart the MCU firmware of MediaTek MT7915 chips.

This can happen when the TX queue ist overwhelmed by non-unicast packets. The same issue also happens on the MT7612 chip, where the chip is reset upon the same failure.


blocktrron:
CLosing for now, as this seems to only work on the first reset

PR #3219, opened 2024-03-07
scripts: image_customization_lib.lua: fail build without valid customization file

### LINKED ISSUE #3218: Invalid or missing image-customization.lua is ignored
The build should abort when no valid image-customization.lua is found, rather than continuing as if the file were empty.

neocturne:
image-customization.lua has replaced GLUON_FEATURES and GLUON_SITE_PACKAGES, so the file is always required to build a useful Gluon image. In addition, not only 'file not found' errors were ignored, but also all other Lua load errors, resulting in a build with empty image customization when the file was not valid.

Closes #3218

### DIFF
--- a/scripts/image_customization_lib.lua
+++ b/scripts/image_customization_lib.lua
@@ -90,26 +90,11 @@ local function evaluate_device(env, dev)
 end
 
 function M.get_selections(dev)
-	local return_object = {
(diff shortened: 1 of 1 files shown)


blocktrron (approved):
This should be backported to v2023.2.x

PR #3227, opened 2024-03-15

herbetom:
contains the patches for #3092


maurerle (approved):
commit messages match - commits match - looks good

herbetom:
bumped it again to also include dropbear patches

PR #3079, opened 2023-12-01
[v2021.1.x]: update OpenWrt 19.07 feeds and backport modules priority changes

T-X:
This updates the OpenWrt feeds to their latest / final version in 19.07. And backports the Gluon module priority changes.

The latter would be useful to have to ease the usage of https://github.com/T-X/gluon-batman-adv-next/ (which is to update the batman-adv version also on Gluon v2021.1.x, to avoid that new features in batman-adv get disabled due to nodes running an old batman-adv version).

PR #3230, opened 2024-03-16
[v2023.1.x] modules: update to latest HEAD + backport container script fixes

herbetom:


blocktrron:
NACK, please don't cherry-pick all consecutive changes from dependabot PRs. Please update to the latest version.

Also i don't think we should rebuild the whole CI to match the current state in master, given v2023.1.x will be EOL soon anyhow.

herbetom:
> NACK, please don't cherry-pick all consecutive changes from dependabot PRs. Please update to the latest version.

That is a much more manual proccess which means more chances of misshapps while applying the cherry picks and resolving merge conflicts with the main benefit beeing a "nicer" commit history for some actions bump commits.

> Also i don't think we should rebuild the whole CI to match the current state in master, given v2023.1.x will be EOL soon anyhow.

There are a few changes, but I wouldn't call it a rebuild. More like a few improvements here and there. And these have been used for some time now without any problems and it seems to be working with v2023.1.x as well.
And just because v2023.1.x will soon be EoL doesn't mean that there won't ever be changes again ([1](https://github.com/freifunk-gluon/gluon/pull/3079#issuecomment-1836917608)). Accordingly, I think adjustments to keep the actions running make sense.
Things like the use of ubuntu-latest, actions with node16 (where I don't know when and if GitHub will remove it completely, but they once said they want to have the migration completed by spring 2024) in v2023.1.x etc. are more problematic in this context.

blocktrron:
I'm fine with pinning the Host-version but i really dislike throwing the whole build-process from master on top of it. This is a stable Release and will build in it's setup for the two years OpenWrt still ships updates.

I will certainly not merge this part.

For the action version-bumps: We get no benefit from cherry picking this all the way to current master, you backport 4 commits for `login-action`, if this does not clutter the commit history i don't know what does. I'm fine picking changes that actually change related configuration (like the labeler config), but we gain nothing in value from bumping a single action a zillion times.

And in general - Updating the OpenWrt base version has nothing to do with throwing30 commits against the CI. Rather keep both separate in order to maintain progress on the individual pieces.

herbetom:
Quite a guessing game if something went wrong or it's just some node16 notice:

image

blocktrron:
Please re-read my [comment](https://github.com/freifunk-gluon/gluon/pull/3230#issuecomment-2015213489) and take the time to understand.

 - Only create a single commit for the update to the latest version of the respective action
 - Keep this separate from the base-modules update
 - Skip the CI adaption to build in it's own docker image
   -  You can pin the Ubuntu Release if you want to

I don't understand the message you intend to convey with the screenshot you've just posted. The required steps to push this forward are all outlined in my comment. The only thing where I've stated i will not merge it has nothing to do with the issue you are outlining in your screenshot.

PR #3233, opened 2024-03-22

blocktrron:
Update the mt76 driver to the latest Version of OpenWrt 23.05.


herbetom:
i think [234f1a2efa558074dfa29382cf86f0024d71c524](https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=234f1a2efa558074dfa29382cf86f0024d71c524) should be included as well in the bump

blocktrron:
Interesting, I did update the hash and invoking hash-fixup it still resets to the hash from my commit.

Nevertheless, I'll bump to the latest commit including the release.

PR #3235, opened 2024-03-23
github: add action for OpenWrt base update

blocktrron:
Create a GitHub action to automatically create updates to the OpenWrt base-version on manual run.

This allows to create base-modules updates where Gluon patches did not change from the GitHub UI.

Skip PR creation. Automatically created PRs do not invoke subsequent actions, thus not creating build-tests.

### DIFF
--- a/.github/workflows/bump-gluon.yml
+++ b/.github/workflows/bump-gluon.yml
@@ -1,0 +1,51 @@
+---
+name: "Update OpenWrt base"
+
+on:
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: "Branch to create update for"
+        required: true
+        default: "master"
+
+jobs:
+  update-openwrt:
+    runs-on: ubuntu-22.04
+    env:
+      COMMIT_NAME: Gluon CI Bot
+      COMMIT_EMAIL: bot@freifunk-gluon.github.io
+    steps:
+      - name: Clone Gluon
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.inputs.branch }}
+      
+      - name: Configure Git User and E-Mail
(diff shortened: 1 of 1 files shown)


blocktrron:
Example of Workflow run: https://github.com/blocktrron/gluon/actions/workflows/bump-gluon.yml

.github/workflows/bump-gluon.yml
```diff
@@ -0,0 +1,48 @@
...
+        run: echo "branch-name=update-openwrt-${{ github.event.inputs.branch }}-$(date +%s)" >> $GITHUB_OUTPUT
+
+      - name: Link example Site
+        run: ln -s docs/site-example site
+      
+      - name: Invoke update-modules
+        run: make update-modules
+      
+      - name: Refresh patches
+        run: make refresh-patches
```
herbetom, line 38: Does the call of refresh-patches actually change anything?

I would expect it to either add the patches, create a notice or abort the branch creation.

From what i can tell it probably does nothing?

blocktrron: Thanks, missed emitting a warning. Updated. Demo run: https://github.com/blocktrron/gluon/actions/runs/8413011733

maurerle (approved):

T0biii:
> Skip PR creation. Automatically created PRs do not invoke subsequent actions, thus not creating build-tests.

Automatically created PRs trigger actions only if you use a personal token or GitHub Apps Token
Like here:
https://github.com/freifunkMUC/site-ffm/blob/5dc2224ee9965158dfa368641d067b8a013a7b6c/.github/workflows/bump-modules.yml#L44

https://github.com/freifunkMUC/site-ffm/pull/516
https://github.com/freifunkMUC/site-ffm/pull/572

if you like i can try to create a pr for this but then you need to add Repo secrets/var for this
For Github App private-key or for the PAT

PR #3245, opened 2024-04-09

herbetom:


maurerle:
You could have also created this by clicking the button here:
https://github.com/freifunk-gluon/gluon/actions/workflows/bump-gluon.yml

Which should be visible for people with write permission on this repo:
https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow

Right?

herbetom:
That Workflow would have had two problems:1. I don't have write permissions for the Gluon repo (okay, could have synced my master branch in my fork and then ran the action there, but that are also quite a few clicks)2. It wouldn't have updated the patches, manual intervention would have been still been requiredAm 12.04.2024 12:13 schrieb Florian Maurer ***@***.***>:
You could have also created this by clicking the button here:
https://github.com/freifunk-gluon/gluon/actions/workflows/bump-gluon.yml
Which should be visible for people with write permission on this repo:
https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow
Right?

PR #3251, opened 2024-05-02
ramips/mt7621: add D-Link Exo AC2600 DIR-882

misanthropos:
- [x] Must be flashable from vendor firmware
  - [x] Web interface
  - [x] TFTP (according to openwrt)

- [x] Must support upgrade mechanism
  - [x] Must have working sysupgrade
    - [x] Must keep/forget configuration (`sysupgrade [-n]`, `firstboot`)
  - [x] Gluon profile name matches autoupdater image name
        (`lua -e 'print(require("platform_info").get_image_name())'`)
- [x] Reset/WPS/... button must return device into config mode
- [x] Primary MAC address should match address on device label (or packaging)
      (https://gluon.readthedocs.io/en/latest/dev/hardware.html#hardware-support-in-packages)
- Wired network
  - [x] should support all network ports on the device
  - [x] must have correct port assignment (WAN/LAN)
    - if there are multiple ports but no WAN port:
      - the PoE input should be WAN, all other ports LAN
      - otherwise the first port should be declared as WAN, all other ports LAN
- Wireless network (if applicable)
  - [x] Association with AP must be possible on all radios
  - [x] Association with 802.11s mesh must work on all radios 
  - [x] AP+mesh mode must work in parallel on all radios
- LED mapping
  - Power/system LED
    - [x] Lit while the device is on
    - [x] Should display config mode blink sequence 
          (https://gluon.readthedocs.io/en/latest/features/configmode.html)
  - Radio LEDs
    - [x] Should map to their respective radio
    - [x] Should show activity
  - Switch port LEDs
    - [x] Should map to their respective port (or switch, if only one led present) 
    - [x] Should show link state and activity

- Docs:
  - [x] Added Device to `docs/user/supported_devices.rst`

PR #3252, opened 2024-05-06
ramips/mt7621: add D-Link 878 A1

misanthropos:
  - [x] Web interface
  - [x] TFTP (according to OpenWRT https://openwrt.org/toh/d-link/dir-878_a1)
  - [x] Other:  D-Link Recovery GUI 
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [x] Should show activity
  - Switch port LEDs
- Outdoor devices only:

- Docs:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -425,6 +425,7 @@ ramips-mt7621
   - COVR-X1860 (A1)
   - DAP-X1860 (A1)
   - DIR-860L (B1)
+  - DIR-878 (A1)
   - DIR-882 (A1)
 
 * GL.iNet
--- a/targets/ramips-mt7621
+++ b/targets/ramips-mt7621
@@ -44,6 +44,8 @@ device('d-link-dap-x1860-a1', 'dlink_dap-x1860-a1')
 
 device('d-link-dir-860l-b1', 'dlink_dir-860l-b1')
 
+device('d-link-dir-878-a1', 'dlink_dir-878-a1')
+
 device('d-link-dir-882-a1', 'dlink_dir-882-a1')
 
 


s-2:
> * Must be flashable from vendor firmware

With OpenWrt 23.05, the factory image for this device was still unencrypted and can only be flashed via the bootloader recovery, which is very cumbersome with current browsers and TCP stacks.

The firmware-util `dlink-sge-image` is used in OpenWrt master to encrypt factory images for this device, so they can be flashed via the regular OEM web updater.

The tool was also backported to 23.05 (e.g. used in gluon by D-Link COVR-X1860).

Maybe we should also backport factory image encryption (as a patch in gluon), until switching to OpenWrt 24?
https://github.com/openwrt/openwrt/commit/4c0dc68f466588bb2ef60b29a4d8052b34112ca6

This also affects DIR-882 which was added in #3251

blocktrron:
So the claim it is flashable from web-if is wrong in the current state present in Gluon?

s-2:
To me it always seemed a bit ambiguous what exactly `Web interface` would refer to; when adding devices to gluon I had previously checked that box as well, even if only recovery was available, considering the recovery system does indeed use a web interface after all.

Only in recent times, this turned to be more complicated, e.g. on some systems only Chromium based browsers work, or the user should reboot the device into recovery after loading the page and selecting firmware (so the upload would be the first http request the devices would see) etc.

However, my concern was mostly about the change having happened in OpenWrt already, so the format of `factory` would change as soon as gluon would switch to OpenWrt 24, and also the new `recovery` image could be generated (e.g. as `other` image type).

Maybe it could prevent some confusion for a newly-added device, if we did it consistently right from the beginning and not change it later? Or maybe it even causes more confusion since people knowing about this would assume gluon would behave the same way as OpenWrt, not sure which way is better in the end :innocent:
Just wanted to mention this here.

PR #3254, opened 2024-05-10
docs: site: pubkey_privacy works for wireguard

Djfe:
I removed a comment from the site documentation citing, that fastd is the only vpn where `pubkey_privacy` has an effect.
I just verified by editing /lib/gluon/site.json and running gluon-reconfigure and reboot, that this settings also works for wireguard VPNs.

I'm not sure whether it works for **Tunneldigger**, so this documentation change requires further feedback.
Is there a fourth VPN provider we support?


maurerle (approved):
We only support fastd and wireguard.
So the removal of this line is correct, as all two supported VPNs do support this option.

blocktrron:
Please remove the Draft state and we can get this merged.

PR #3258, opened 2024-05-12
mesh-vpn-wireguard: actually set the mtu from the site.conf on the wireguard interface

maurerle:
Before, we did only set the MTU on mesh-vpn (the vxlan interface), which has to be smaller than the set value on the wg_mesh (wireguard interface).

On WAN interfaces with an MTU of 1500, the default wireguard MTU of 1420 is optimal (as wireguard takes 80 Bytes).
Though it would be better for PPPoE interfaces (which typically have a MTU of 1496) to use a MTU of 1416 or less (1406 being a typical value, currently used by FFMUC, FFH and FFAC).

[The documentation](https://gluon.readthedocs.io/en/latest/user/mtu.html) mentions an optimal wireguard MTU of 1376.
This would then also work on WAN-interfaces with an MTU of 1436 on IPv4 respective 1462 for IPv6).

This PR changes the behavior for mesh-vpn-wireguard to set the correct MTU from the site.conf.

Curiously, this did never lead to any known problems in one of the stated communities, and was found when debugging a fragmentation problem.

Eventually, one should also set the MTU on the wireguard gateway/supernode to something less than 1420..?

### DIFF
--- a/package/gluon-mesh-vpn-wireguard/luasrc/lib/gluon/upgrade/400-mesh-vpn-wireguard
+++ b/package/gluon-mesh-vpn-wireguard/luasrc/lib/gluon/upgrade/400-mesh-vpn-wireguard
@@ -60,7 +60,8 @@ uci:section('network', 'interface', 'wg_mesh', {
 	proto = 'wireguard',
 	fwmark = 1,
 	private_key = wg_private_key,
-	mtu = active_vpn.mtu(),
+	-- Add 70 bytes for IPv6 VXLAN overhead
+	mtu = active_vpn.mtu() + 70,
 })
 
 uci:section('network', 'interface', 'mesh_wg_mesh', {
--- a/package/gluon-mesh-vpn-wireguard/luasrc/lib/gluon/upgrade/400-mesh-vpn-wireguard
+++ b/package/gluon-mesh-vpn-wireguard/luasrc/lib/gluon/upgrade/400-mesh-vpn-wireguard
@@ -6,6 +6,8 @@ local util = require('gluon.util')
 local site = require 'gluon.site'
 local sp = util.subprocess
 local wait = require 'posix.sys.wait'
+local vpn_core = require('gluon.mesh-vpn')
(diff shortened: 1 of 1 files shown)


neocturne:
We discussed this issue today, and found that the MTU in `site.conf` should always be applied to `mesh-vpn`, as it already is for all VPN providers; changing this would make the VPN handling less consistent. The MTU of the Wireguard interface should be `active_vpn.mtu() + 70` to account for the VXLAN overhead.

Unfortunately, the documentation is wrong and adds the 70 bytes at the different point than what Gluon does, but this must be fixed in the docs, not on the code.

package/gluon-mesh-vpn-wireguard/luasrc/lib/gluon/upgrade/400-mesh-vpn-wireguard
```diff
@@ -58,6 +60,7 @@ uci:section('network', 'interface', 'wg_mesh', {
 	proto = 'wireguard',
 	fwmark = 1,
 	private_key = wg_private_key,
+	mtu = active_vpn.mtu(),
```
neocturne, line 63:
```suggestion
	-- Add 70 bytes for IPv6 VXLAN overhead
	mtu = active_vpn.mtu() + 70,
```

maurerle:
Thanks for taking care of this! @blocktrron

blocktrron:
@maurerle  Can you provide a short explanation for the release notes of an upcoming release?

maurerle:
I'd suggest something like this:
`* Fixed setting MTU on vxlan interface correctly by adding 70 to it when using mesh-vpn-wireguard-vxlan (https://github.com/freifunk-gluon/gluon/pull/3258)`

neocturne:
As the MTU was previously not set at all for wireguard, I don't think the 70 need to be mentioned anywhere.

We still need to fix the MTU doc page though.

blocktrron:
Ah, thanks for the explanation, I misread the previous comments.

PR #3261, opened 2024-05-14
mt76: work around dropped TX status reporting

blocktrron:
These fixes were developed for MT7603 and MT7612. Originally, we've discovered the TX-status reporting mishandles cases where continuous action frames are sent over a mesh-link but no actual data traffic is transmitted over the link.

This leads to a client "sticking" to the AP even if it already left the covered area.

In the process, we've discovered this patchset might(!) also help mitigate issues seen on MT7916 based platforms. Therefore, i think it is a good idea to open this PR as a draft to serve as a starting point for discussions on this matter.

I've rebased this patchset on current master, build testing by the CI runners now. We've tested fixes with the mt76 patches based on v2023.2.1 releases.

### First example

image

Dashboard: https://stats.darmstadt.freifunk.net/d/000000021/router-meshviewer-export?var-node=3894edf5ff9b&orgId=1&from=1708931806122&to=1711420462467

This Node was the one we've originally diagnosed the issue. You can see the devices accumulating on the OWE VAP for the 5 GHz radio, as the driver incorrectly marks all action-frames sent to the device as acked. They are dropped after 24h for rekeying.

You can observe a sticky client on the 2.4 GHz radio (MT7603). This does not matter if it happens on the OWE or unencrypted VAP. The node will in this case pollute the airtime and be unavailable to other clients.

### Second example

I have no access to this node, but it shows the same symptoms:

image

Dashboard: https://stats.darmstadt.freifunk.net/d/000000021/router-meshviewer-export?var-node=3894edf70c46&orgId=1&from=1713058330554&to=1715097473774

## MT7986 Nodes

We've seen cases where MT7986 and MT7981 radios might experience degraded performance. Coincidentally, this patchset seems to mitigate similar issues as well (We only have some days of testing on this hypothesis).

image

image

Dashboard: https://stats.darmstadt.freifunk.net/d/000000021/router-meshviewer-export?var-node=c87f54231ad8&orgId=1&from=1713058547177&to=1715642058103

### DIFF
--- a/patches/openwrt/0010-mt76-include-fixes-for-MT7603-MT7612.patch
+++ b/patches/openwrt/0010-mt76-include-fixes-for-MT7603-MT7612.patch
@@ -1,0 +1,175 @@
+From: David Bauer <mail@david-bauer.net>
+Date: Thu, 14 Mar 2024 09:39:22 +0100
+Subject: mt76: include fixes for MT7603 / MT7612
+
+diff --git a/package/kernel/mt76/patches/0001-tx-add-limit-for-TXS-ack-override.patch b/package/kernel/mt76/patches/0001-tx-add-limit-for-TXS-ack-override.patch
+new file mode 100644
+index 0000000000000000000000000000000000000000..568c590f24c251dda70522865af32b3753cd5beb
+--- /dev/null
++++ b/package/kernel/mt76/patches/0001-tx-add-limit-for-TXS-ack-override.patch
+@@ -0,0 +1,79 @@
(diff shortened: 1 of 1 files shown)


blocktrron:
Results for MT7612 and MT7603 look very promising (no sticky clients anymore).

For MT7986 it also significantly improves the situation, however it does not completely fix all current issues.

grische:
@blocktrron should this potentially be backported to v2023.2.x?

blocktrron:
@grische lets give it some time and me for discussing upstreaming it.

PR #3264, opened 2024-05-15
gluon-autoupdater: allow https mirror urls

kevin-olbrich:


kevin-olbrich:
> Features[](https://gluon.readthedocs.io/en/latest/releases/v2023.2.html#features)
> TLS support[](https://gluon.readthedocs.io/en/latest/releases/v2023.2.html#tls-support)
> Gluon now provides HTTPS client support when the tls feature is included in the site configuration, allowing nodes to establish encrypted connections to autoupdater mirrors, opkg repositories and other HTTPS servers.
Source: https://gluon.readthedocs.io/en/latest/releases/v2023.2.html

TLS support can be enabled but the site check prevents https mirror urls in site.conf.

neocturne:
Hmm, the change is working as intended, but optimally we'd only allow HTTPS if the gluon-tls package is installed, to prevent accidentally building non-updateable firmwares (which could happen by removing the `tls` feature to save space, without thinking about updating the URL).

This could be implemented by having `gluon-tls` create a marker file like `gluon-wireless-encryption-wpa3` does, and check for the existence of the marker in the site check script (similar to the check for the contents of the `default_branch` file).

neocturne:
Hmm, or does libustream fall back to HTTP anyways when HTTPS is requested, but not available?

T0biii:
maybe auto select the correct protocol if the `tls` feature is there or not.
if you want explicitly `http://` set the URL to it. if you want `http://` and/or `https://` remove the prefix?
and if you have `https://` set, try to fallback to `http://` if all `https://` mirrors are unreachable

Djfe:
In that case https has no advantages. Just drop all packets on port 443 and nodes will fall back to port 80.
either the community offers https or it doesn't.

the bigger issue imo are root certificates that need to be replaced from time to time.
if a node is offline for atleast x years then it won't be able to update anymore, right?
unless you build self-signed certificates into the firmware that won't expire

neocturne:
> In that case https has no advantages. Just drop all packets on port 443 and nodes will fall back to port 80. either the community offers https or it doesn't.

Fallback might still be useful if it only happens when no TLS library is installed. A TLS connection failing must not trigger a fallback.

Maybe we should support scheme-less URLs like browsers (`//server/path`) to have the node automatically use HTTP or HTTPS depending on `gluon-tls` being installed.

> the bigger issue imo are root certificates that need to be replaced from time to time. if a node is offline for atleast x years then it won't be able to update anymore, right? unless you build self-signed certificates into the firmware that won't expire

Such a long downtime would likely also mean skipping several Gluon versions, so automated updates wouldn't be supported anyways.

---

Anyways, I'm not sure why this was tagged as "blocked" and not "waiting-on-author". @kevin-olbrich: Would you like to help with the discussed improvements, or should we take over from here?

Djfe:
Thanks, for taking a look at this again 👍
I like the schema-less approach a lot.
Using a fallback for images built without `gluon-tls` is sensible and should be considered.

My last comment was probably quite the downer, I want to apologize for that.
I just wanted to make clear, which downsides this change could have. And bring some counterarguments to the table.

I kind of adapted my standpoint by now. https is something we **should** want to pursue here. So I support the change.
We should do it carefully though.

If we do this, then this needs to be communicated **very clearly** to Communities building gluon!
In addition to the release notes (and mails via `gluon@luebeck.freifunk.net`) I suggest that the site check should error for non schema-less URLs and inform builders about this change. [0]

Honestly I would like us to even require https to be implemented server-side regardless of `gluon-tls` being built in. But considering how the Tunneldigger debate went, I assume backlash for this. So I suggest a more diplomatic approach here:
Tell builders that if their updateserver offers https:// then it is required to be working. (if `gluon-tls` is built into the image) [1]

As an example:
Our Update-URLs in Aachen aren't offering valid certificates at the moment because historically our Updateservers are caching servers running on different infrastructure and we never set up Let's Encrypt to also add these domains to the certificates used by our Nginx running on our Firmwareserver.
This can be easily done of course, but wasn't necessary till now, because https couldn't be used, yet. We will happily adapt the change.

As another example:
I know of communities running their Updateservers on ipv6-only inside their Freifunk on Unique-Local-Addresses (ULA). Afaik such addresses can never be run with official certificates by CAs (Certificate Authoritiy) and have to be self-signed.
Some of these also run or ran on custom top-level-domains like .ffe (Freifunk Essen) which would also require self-signed certificates.

I digress. While it's not necessary for this PR, I think, it could be useful to have an official way to build self-signed certificates into gluon images. [2]

I know my post is getting way too long already for something so simple 😅

some nitpicky comments that aren't super relevant to the PR at hand:
[0]
For opkg as well as autoupdater paths.

[1] If we ever face out 8MiB flash devices we could still think about including `gluon-tls` by default and requiring update servers to offer https. But right now this is overkill imho.

[2] Even if self-signed certificates are just deployed as a fallback for the autoupdater in-case root certificates do eventually expire.
I agree that this occasion is rare but we do already have a few communities still running devices on Gluon v2015.1 so it's something to keep in mind. We don't know what the next decade brings.
> so automated updates wouldn't be supported anyways.

Yes, but you can just simply chain updates and offer images for every second gluon release till the devices are up-to-date. We adjust our update paths for every major gluon release. (it's also possible to interpret the HTTP headers instead now to offer different images server-side depending on the connected client)

neocturne:
@Djfe I have no idea what you think needs to be communicated. My plan would be the following:

- If you have a http:// URL in site.conf, HTTP will be used. No TLS happening. Period. This is currently the only supported case, nothing changes about this. I don't have any plans to drop this. And as long you you don't modify site.conf, Gluon will never start using HTTPS on its own.
- If you have a https:// URL in site.conf, HTTPS will be used. No fallback. Build-time check for gluon-tls being installed.
- If you have a // URL in site.conf, HTTP or HTTPS will be used depending on the availability of gluon-tls.

It should be obvious that you need to take care of keeping your HTTPS certificates valid if you use HTTPS.

Djfe:
ah, well that clears up what I was worrying about.
I foolishly expected schema-less to become the only supported case. nvm

> Build-time check for gluon-tls being installed.

great solution :)

neocturne:
Superseded by #3463.

PR #3286, opened 2024-06-15
linux: net: dsa: mt7530: add support for bridge port isolation

neocturne:
Adds bridge port isolation to MT7621-based devices like the Ubiquiti EdgeRouter-X.

Part of #2679.

### DIFF
--- a/patches/openwrt/0010-linux-net-dsa-mt7530-add-support-for-bridge-port-isolation.patch
+++ b/patches/openwrt/0010-linux-net-dsa-mt7530-add-support-for-bridge-port-isolation.patch
@@ -1,0 +1,325 @@
+From: Matthias Schiffer <mschiffer@universe-factory.net>
+Date: Sat, 15 Jun 2024 01:09:41 +0200
+Subject: linux: net: dsa: mt7530: add support for bridge port isolation
+
+Backport of from net-next (and one older patch) to Linux 5.15.
+
+Signed-off-by: Matthias Schiffer <mschiffer@universe-factory.net>
+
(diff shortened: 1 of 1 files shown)


neocturne:
Updated to the version accepted for net-next and OpenWrt main

rotanid:
for me, stuff that was accepted in net-next and openwrt main doesnt need a further review by gluon maintainers, it's only a question about whether we want to backport it.
i'd say yes.

neocturne:
I did have to make some changes to the patches for 5.15, but yes, the main question is whether we want the backport.

herbetom:
Is there a reason to track this backport downstream and not in the OpenWrt repo? Seems like something that's not gluon specific.

Otherwise it seems justifiable to backport it.

neocturne:
Arguably it's a breaking change, so I'd rather not backport it to OpenWrt stable branches.

herbetom:
From my understanding bridge port isolation is only enabled if that setting has been enabled and AFAIK it's not by default in openwrt (only gluon enables it by default).
Therefore it should only break stuff if a user enabled it in the past, forgot about it and then it suddenly starts working. That doesn't really seem like a problem openwrt should care about.

neocturne:
The first patch affects all drivers, not just those that actually support the offloading; it also doesn't only affect port isolation, but also other bridge flags. It's difficult to say what the exact effects might be.

rotanid:
let's test in master, we can revert later if we have issues with this

rotanid:
let's test in master, we can revert later if we have issues with this

@neocturne i suggest you try you build and if necessary refresh patch with current master branch and merge it youself

neocturne:
Status: I want to include the qca8k patches as well (not sure if they have been backported to OpenWrt main already)

neocturne:
Closing, as mt7530 bridge port isolation has been ported to OpenWrt already. Support for other switch drivers is tracked in https://github.com/freifunk-gluon/gluon/issues/2679.

PR #3288, opened 2024-06-19
openwrt/patches: ipq40xx: disable DMA for fritzbox 4040

Emantor:
We saw read/write errors in conjunction with the onboard flash on Fritzbox 4040 and observed devices unable to update that required a recovery via fritz flash. After disabling DMA via the internal SPI we no longer observed these issues.


blocktrron:
Can you submit this patch upstream? We generally don't carry patches like this downstream.

blocktrron:
Ping me when you've opened the upstream PR. Maybe my statement was a bit misleading - We con't carry such patches downstream when they are not contained in a future OpenWrt release.

However, in my opinion the patch should be picked to the release branch for OpenWrt, thus we can get it by updating the respective OpenWrt commit we track for our base.

PR #3294, opened 2024-06-27
private-wifi: rename ifname to wan_wifiX to not collide with openwrt devices with more than one wan port

### LINKED ISSUE #3290: DSA interface name overlap: wan1 could be ethernet or private wifi
Switching to DSA introduced new interface names like wan1, wan2, lan1, lan2, lan3
This could cause conflicts, once devices like this one end up in Gluon: MQMaker - WiTi Board
`ucidef_set_interfaces_lan_wan "lan1 lan2 lan3 lan4" "wan1 wan2"`
https://github.com/openwrt/openwrt/blob/66edb2f9ad89206ac6f608ec76fad161b53683a7/target/linux/ramips/mt7621/base-files/etc/board.d/02_network#L128

wan1 is currently used if you enable the private wifi feature (for the second band)

If this is an actual conflict, then we should rename the private wifi interfaces eventually.

maurerle:
Currently, no device is affected, though it makes sense to improve the current situation by giving more meaningful ifnames anyway.

closes #3290

### DIFF
--- a/package/gluon-web-private-wifi/luasrc/lib/gluon/config-mode/model/admin/privatewifi.lua
+++ b/package/gluon-web-private-wifi/luasrc/lib/gluon/config-mode/model/admin/privatewifi.lua
@@ -81,7 +81,7 @@ function f:write()
 				ssid       = ssid.data,
 				key        = key.data,
 				macaddr    = macaddr,
-				ifname     = suffix and 'wan' .. suffix,
+				ifname     = suffix and 'wl-wan' .. suffix,
 				disabled   = false,
 			})
 
--- a/package/gluon-private-wifi/luasrc/lib/gluon/upgrade/325-gluon-private-wifi
+++ b/package/gluon-private-wifi/luasrc/lib/gluon/upgrade/325-gluon-private-wifi
@@ -12,7 +12,7 @@ wireless.foreach_radio(uci, function(radio)
 		return
 	end
 
-	uci:set('wireless', name, 'ifname', suffix and 'wan' .. suffix)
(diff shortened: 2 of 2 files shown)


package/gluon-web-private-wifi/luasrc/lib/gluon/config-mode/model/admin/privatewifi.lua
```diff
@@ -81,7 +81,7 @@ function f:write()
 				ssid       = ssid.data,
 				key        = key.data,
 				macaddr    = macaddr,
-				ifname     = suffix and 'wan' .. suffix,
+				ifname     = suffix and 'wan_wifi' .. suffix,
```
blocktrron, line 84: Even though the package is called "private-wifi", we agreed some time ago to only continue using the term "WLAN", as "WiFi" is associated with the WiFi Alliance.

neocturne: How about making it a "wl" prefix, naming the interface "wlwanX"? Could also be used like "wlclient", "wlmesh", ... in the future as a common naming scheme.

maurerle: Sounds good, what about `wl_wanX` and `wl_client`/`wl_mesh` in the future?

neocturne: If we add a separator (which I do not have strong opinions on), I'd rather use `wl-` to match OpenWrt's `br-`.

PR #3301, opened 2024-07-01
scripts: update-modules: read feeds from modules

grische:
A small improvement derived from work we are doing on our side


maurerle (approved):
It makes sense to be more generic so that customized feeds are also recognized by the script.

PR #3309, opened 2024-07-07
gluon-core: fix swconfig detection in ethernet module

herbetom:
Switch config for swconfig can be found in the network, not the system config.

This resulted in `Switch type: none` beeing reported for swconfig devices.

cc @mkg20001


maurerle (approved):

maurerle:
Tested on a `	TP-Link Archer C7 v2` - works fine. This shows `swconfig` now.

PR #3310, opened 2024-07-08
scripts: add --tags to git describe in getversion.sh to include lightweighted tags

misanthropos:
This flag allows to get the latest non annotated tag in e.g. site repositories.

### DIFF
--- a/scripts/getversion.sh
+++ b/scripts/getversion.sh
@@ -8,6 +8,6 @@ fi
 cd "$1" || exit 1
 
 cat .scmversion 2>/dev/null && exit 0
-git --git-dir=.git describe --always --abbrev=7 --dirty=+ 2>/dev/null && exit 0
+git --git-dir=.git describe --tags --always --abbrev=7 --dirty=+ 2>/dev/null && exit 0
 
 echo unknown


neocturne:
We only use annotated tags in Gluon, which are always used by `git describe`. So I guess this is about site repos which use lightweight tags?

misanthropos:
> We only use annotated tags in Gluon, which are always used by `git describe`. So I guess this is about site repos which use lightweight tags?

Indeed! I stumbled upon this while using gluon-info which reported a rather old site config.

maurerle (approved):
I found this confusing in the beginning as well.
However, I quickly learned to always use annotated tags in the site repository as well.

I would favor having this working with other tags as well - as it does not harm and makes things easier for beginners.

neocturne:
Hmm, I'm currently reconsidering this - using lightweight tags for releases is a bad idea IMO, as they don't contain date and author metainfo - and forcing the use of annotated tags frees up the lightweight tags for any non-release use when you just want to point at a commit (which I guess is the rationale for them being ignored by `git describe` by default). So maybe we should revert this?

misanthropos:
Fine with me - question remains how to propagate the necessity of an annotated commit so gluon-info does give correct information.

neocturne:
I think the only place where we document that tags are relevant at all is this: https://github.com/freifunk-gluon/gluon/blob/main/docs/dev/build.rst#helper-scripts - this should be extended, but maybe adding a note outside of the developer section of the docs would be a good idea as well.

T0biii:
Just FYI why that feature was quite good for us back then.

For new releases, we often used GitHub’s built-in (Web UI) function for creating releases, which also creates a tag. However, this tag is unfortunately lightweight.

image

PR #3312, opened 2024-07-09
fritzbox-7530/7520: utilize lan1 as wan, others as lan 

grische:
> With the switch to DSA on this device in openwrt 23.05, we can now utilize the ports of a 7520/7530 better, by splitting them according to their usage.
> Before, they had all LAN ports used as WAN - a behavior which can also be easily configured in config mode nowadays.
> 
> If I remember correctly, this also changes an existing configuration to LAN1 -> WAN and LAN2,3,4 -> LAN
> Therefore, the existing router owner need to make sure to plug the cable into Port1.
> 
> This patch has been tested with latest gluon and already is deployed on all affected FB7520/7530 in FFAC.

This is a new version of #3118 that includes a uci migration path for devices that were configured pre-DSA.

cc @maurerle

### DIFF
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/020-interfaces
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/020-interfaces
@@ -63,6 +63,10 @@ elseif platform.match('ath79', 'generic', {
 	'ubnt,unifiac-pro',
 }) then
 	lan_ifname, wan_ifname = 'eth0.2', 'eth0.1'
+elseif platform.match('ipq40xx', 'generic', {
+	'avm,fritzbox-7530',
+}) then
+	lan_ifname, wan_ifname = 'lan2 lan3 lan4', 'lan1'
 elseif platform.match('ramips', 'mt7621', {
 	'netgear,wac104',
 }) then
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/021-interface-roles
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/021-interface-roles
@@ -43,6 +43,7 @@ if uci:get('gluon', 'iface_wan') then
 end
 if uci:get('gluon', 'iface_single') then
 	roles.wan = uci:get_list('gluon', 'iface_single', 'role')
+	roles.lan = uci:get_list('gluon', 'iface_single', 'role')
 end
 
 -- Non-existing interfaces are nil, so they will not be added to the table


herbetom:
The change isn't that dsa is introduced for the device, the change is that ports can be split thanks to dsa or something like that. 

Also i don't see the point in gatekeeping the migration for specific devices. If `/wan` + `/lan` is expected and `/single` is configured it always seems like a sensible default to copy the `/single` config to `/wan` / `/lan`.

maurerle:
With 
https://github.com/freifunk-gluon/gluon/blob/3784d24c60332f5e0fec8cf1bc7fd61d5e03b9bb/package/gluon-core/luasrc/lib/gluon/upgrade/021-interface-roles#L58
and the following coede, I also think it is fine to do this generally for all devices :)

Though I understand that @grische wanted to reduce the impact of the PR :)

blocktrron:
I'm opposed to adding special cases conveying a opinionated default roleset of the port assignment.

In the end, this adds further complexity. The user itself can re-role these DSA ports now and I'd rather like to see this configuration possibility in the Web-UI instead of configuring whatever an individual feels well suited differing from the OpenWrt default.

grische:
@blocktrron I understand that it is not ideal that we deviate from OpenWRT here, but [we have done so in the past](https://github.com/freifunk-gluon/gluon/blob/f49f90f81c042f32d02e038cfd1db4a23cf101f2/package/gluon-core/luasrc/lib/gluon/upgrade/020-interfaces#L59-L86) and users are used to it:
Many devices that do not have a dedicated WAN port use LAN1 as their WAN port and all others for "LAN" ports, making it the de-facto default for our communities.

Out of habit, many (of at least our) users just come and plug their uplink into LAN1 and are then wondering why there is no client network on LAN2. Even worse, they might enable Mesh (on WAN) and then bridge the Mesh and Client networks accidentally.

While I think the suggestion to be able to fully configure ports separately is a nice idea, I would treat this independently and focus in this PR on easy-of-use (or "good defaults") for our users.

neocturne:
The commit message "gluon-core: fix undefined lan role after migration" is inaccurate - it is not undefined, it just defaults to the `site.conf` setting.

I agree with @grische that this should be merged - Gluon's device checklist explicitly states that we should differ from OpenWrt's defaults for such devices.

grische:
@neocturne thanks, I changed the commit message to be more generic

Djfe:
> In the end, this adds further complexity. The user itself can re-role these DSA ports now and I'd rather like to see this configuration possibility in the Web-UI instead of configuring whatever an individual feels well suited differing from the OpenWrt default.

That's part of the point: You can't reconfigure singular ports in the Web-UI since it's a bridge and mkg's PR hasn't been merged, yet.
What's presented here is a useful default for the case Gluon. Also for this particular device it's actually the default to have WAN on LAN1 (if it isn't used to do DSL itself, which can be done in OpenWrt nowadays)
Also when the device was introduced into OpenWrt and Gluon it wasn't possible to split the bridge, right?

Regardless I support the change. And getting MKG's PR finished up and merged would benefit users a lot.

Djfe (approved):

Djfe:
I suggest adding these as well, we run these in Aachen just fine:
```
elseif platform.match('lantiq', 'xrx200', {
	'avm,fritz3370-rev2-hynix',
	'avm,fritz3370-rev2-micron',
	'avm,fritz7360sl',
	'avm,fritz7360-v2',
	'avm,fritz7362sl',
}) then
	lan_ifname, wan_ifname = 'lan2 lan3 lan4', 'lan1'
elseif platform.match('lantiq', 'xrx200', {
	'tplink,tdw8970',
	'tplink,tdw8980',
}) then
	lan_ifname, wan_ifname = 'lan1 lan2 lan3', 'lan4'
```
LAN4 on the TP-Link TDW89x0 devices is explicitly marked as a WAN port
grafik

maurerle:
I would rather get this PR through as is and handle other devices afterward if needed in a separate PR.

PR #3334, opened 2024-08-31
ath79-generic: add support for DLAN (pro) 1200+ WiFi ac

maurerle:
The DLAN pro 1200+ WiFi ac is the same as the DLAN 1200+ WiFi ac (non-pro).
It is a wall plug device which still exposes the wall plug.

The powerline functionality does not work without additional firmware, which can not be shipped due to licensing issues - yet someone from our community wanted to use this device for FF and did test it.

Looks fine to me, except that there is no Power LED, just the House-LED (used for powerline) and the WIFI-LED (which works as expected)

  - [ ] Web interface
  - [x] TFTP see: https://openwrt.org/toh/devolo/dlan_pro_1200_wifi_ac#oem_installation_using_the_tftp_method
  - [ ] Other: <specify>
        `devolo-dlan-pro-1200+-wifi-ac`
- Wired network
- LED mapping
  - Power/system LED (non exists)
  - Radio LEDs (only 1 wifi LED)
    - [x] Should show activity
  - Switch port LEDs (non)
    ~~- [ ] Should map to their respective port (or switch, if only one led present)~~
    ~~- [ ] Should show link state and activity~~
- Docs:

### DIFF
--- a/targets/ath79-generic
+++ b/targets/ath79-generic
@@ -86,6 +86,12 @@ device('buffalo-wzr-hp-g300nh-rtl8366s', 'buffalo_wzr-hp-g300nh-s')
 
 -- devolo
 
+device('devolo-dlan-pro-1200+-wifi-ac', 'devolo_dlan-pro-1200plus-ac', {
+	packages = ATH10K_PACKAGES_QCA9880,
+	factory = false,
+	broken = true, -- no power led
+})
+
 device('devolo-wifi-pro-1200e', 'devolo_dvl1200e', {
 	packages = ATH10K_PACKAGES_QCA9880,
 	factory = false,


targets/ath79-generic
```diff
@@ -86,6 +86,11 @@ device('buffalo-wzr-hp-g300nh-rtl8366s', 'buffalo_wzr-hp-g300nh-s')
 
 -- devolo
 
+device('devolo-dlan-pro-1200+-wifi-ac', 'devolo_dlan-pro-1200plus-ac', {
+	packages = ATH10K_PACKAGES_QCA9880,
+	factory = false,
+})
+
```
blocktrron, line 93: As the LED is not working, please add broken here.

maurerle: There is no Power LED which should work that way in the first place.
1. house LED for Powerline (DLAN)
2. wifi LED for wireless activity

I am happy adding this as broken.

PR #3353, opened 2024-10-10
kirkwood-generic: update BROKEN reason, add Linksys EA4500

neocturne:
The hardware is identical to the E4200 v2, and it's broken, so only minimal testing was done - this is added mostly because I have one lying around and its good to have some "supported" device for testing ARMv5.


grische (approved):
LGTM

PR #3355, opened 2024-10-13
gluon-web-network: support handling of active-low PoE passthrough

blocktrron:
The Aruba AP-303H offers a switchable PSE output on ETH3. The power state is controlled using a GPIO pin like seen on CPE210 and Nanostation alike.

However, the PSE state is active low, which the current code handles as "Enable PoE Passthrough disable". This is not intuitive, thus support the active low case in the Gluon Web-UI.

### DIFF
--- a/package/gluon-web-network/luasrc/lib/gluon/config-mode/model/admin/network.lua
+++ b/package/gluon-web-network/luasrc/lib/gluon/config-mode/model/admin/network.lua
@@ -102,15 +102,28 @@ uci:foreach("system", "gpio_switch", function(si)
 		end
 
 		local texts = {
-			["^PoE Power Port(%d*)$"] = function(m) return translatef("Enable PoE Power Port %s", m[1]) end,
-			["^PoE Passthrough$"] = function() return translate("Enable PoE Passthrough") end,
+			["^PoE Power Port(%d*)$"] = {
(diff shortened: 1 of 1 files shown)


blocktrron:
Tested-on: Aruba AP11D (303H)

package/gluon-web-network/luasrc/lib/gluon/config-mode/model/admin/network.lua
```diff
@@ -102,15 +102,28 @@ uci:foreach("system", "gpio_switch", function(si)
 		end
 
 		local texts = {
-			["^PoE Power Port(%d*)$"] = function(m) return translatef("Enable PoE Power Port %s", m[1]) end,
-			["^PoE Passthrough$"] = function() return translate("Enable PoE Passthrough") end,
+			["^PoE Power Port(%d*)$"] = {
+				["func"] = function(m) return translatef("Enable PoE Power Port %s", m[1]) end,
+				["active_low"] = false,
```
neocturne, line 107: How about the following naming? (and not using the `[""]` syntax without reason)
```suggestion
				get_title = function(m) return translatef("Enable PoE Power Port %s", m[1]) end,
				active_low = false,
```

blocktrron: Works for me

PR #2769, opened 2023-01-24
ipq40xx-generic: add linksys-mr8300

Djfe:
Notes on this device:
The bootloader only loads up to `3MB` of Kernel in stock config.
So users need to edit a bootloader variable to `5MB` for Kernel 5.15 and onwards (master/OpenWRT 2023)
[They are raising compat version for this](https://github.com/openwrt/openwrt/blob/7c04e80f87f343bde6d137d813d694f40f3ec481/target/linux/ipq40xx/image/generic.mk#L639-L648), so maybe let's only add this device in Gluon 2023 at the earliest to avoid a future failure of autoupdate? On the other hand there are communities with older gluon's that might want to have access to this device now while IPQ4019 is still considered good. (OT: IPQ807x was just added and new device support is flooding into OpenWRT)

The router has three WiFi radios. 5GHz channels are split like here: #1661 / #1666
My solution for now: remove kernelpackets for QCA9888-support until triple radio support is implemented. We could flag the device as broken until then.
Freifunk doesn't require dfs channels anyways.
If there is a way to disable _outdoor support_ (aka remove the config entry), then pls tell me where :)

The same thing could be done to add the **Fritz!Repeater 3000**, so it could atleast be deployed by Freifunk Communities.
Other devices that use the same packages-string: openmesh-a62 and plasma-cloud-pa2200. Does this also fix their 5GHz support? Looking at their dts files, they are triple radio devices as well, but I don't want to break anything accidentally.

How do I remove "dallas" from the expected image name? It might say dallas in the code but that name is not found in OpenWRT's firmware selector and it's also not in the image file (I looked at strings at start and end of the image file).
I found [this](https://github.com/freifunk-gluon/packages/tree/master/libs/libplatforminfo) but there is no exceptions  for singular devices in there only for broadcom devices as a whole.

Not sure whether this is important:
When I enter config mode all LEDs are supposed to flash once. All of them are except the power led which stays lit throughout the whole reboot. The power led is on top of the router, every other led is on the back.
Also: there is no WiFi led.

- [x] Must be flashable from vendor firmware
  - [x] Web interface
  - [ ] TFTP
  - [ ] Other: <specify>
- [ ] Must support upgrade mechanism
  - [x] Must have working sysupgrade
    - [x] Must keep/forget configuration (`sysupgrade [-n]`, `firstboot`)
  - [ ] Gluon profile name matches autoupdater image name
        (`lua -e 'print(require("platform_info").get_image_name())'`)
[...]

PR #3057, opened 2023-11-06
ipq40xx-generic: add support for AVM Fritz!Repeater 3000

maurerle:
@j0chn1 tested this device.

Eventually, the device can get full support if the third radio is used instead of the second.
But I don't know how we want to handle this.

- [X] Must be flashable from vendor firmware
  - [ ] Web interface
  - [X] TFTP
  - [ ] Other: <specify>
- [X] Must support upgrade mechanism
  - [X] Must have working sysupgrade
  - [X] Must keep/forget configuration (`sysupgrade [-n]`, `firstboot`)
  - [X] Gluon profile name matches autoupdater image name
        (`lua -e 'print(require("platform_info").get_image_name())'`)
- [X] Reset/WPS/... button must return device into config mode
- [X] Primary MAC address should match address on device label (or packaging)
      (https://gluon.readthedocs.io/en/latest/dev/hardware.html#hardware-support-in-packages)
  - When re-adding a device that was supported by an earlier version of Gluon, a
    factory reset must be performed before checking the primary MAC address, as
    the setting from the old version is not reset otherwise.
- Wired network
  - [X] should support all network ports on the device
  - [X] must have correct port assignment (WAN/LAN)
    - if there are multiple ports but no WAN port:
      - the PoE input should be WAN, all other ports LAN
      - otherwise the first port should be declared as WAN, all other ports LAN
- Wireless network (if applicable)
  - [X] Association with AP must be possible on all radios
  - [X] Association with 802.11s mesh must work on all radios 
  - [X] AP+mesh mode must work in parallel on all radios
	
- LED mapping
  - Power/system LED
    - [X] Lit while the device is on
    - [x] Should display config mode blink sequence 
          (https://gluon.readthedocs.io/en/latest/features/configmode.html)
  - Radio LEDs (none)
	The RSSI Led, does not work, others are not present
    - [ ] ~~Should map to their respective radio~~
    - [ ] ~~Should show activity~~
  - Switch port LEDs
	No switch port LEDs available
    - [ ] ~~Should map to their respective port (or switch, if only one led present)~~ 
    - [ ] ~~Should show link state and activity~~
- ~~Outdoor devices only:~~
  - [ ] ~~Added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`~~
- ~~Cellular devices only:~~
  - [ ] ~~Added board name to `is_cellular_device` function in~~
[...]

PR #3357, opened 2024-10-14
Triband radio support

maurerle:
I'd like to reopen the changes from #1666 to make way to support #1661

This works quite well and exposes the three wifis correctly on the status-page:

image

radio0, 5GHz, 144 (client wifi)
radio1, 2.4GHz, 11 (client + mesh)
radio2, 5GHz, 44 (mesh)

and also the config mode shows the expected default config for such devices:
image

We might want to:

- [x] grey out the mesh network operation if a radio does only support DFS channels (removed, as the site api does not directly support disabling items)
- [ ] adjust config to detect how wifis should be distributed best across the channels (similar to how it is done here https://github.com/freifunk-gluon/gluon/blob/fc25b4940a7da61e4a60292324a3e2615122d810/package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless#L15 )
- [ ] move whole wireless configuration in uci to `gluon.wireless` in the future (far out of scope here)

This is currently focusing on having two 5GHz bands, where at least one does not support all channels.

---

This blocks the following PRs:

- #2769
- #3418

### DIFF
--- a/package/gluon-web-wifi-config/i18n/de.po
+++ b/package/gluon-web-wifi-config/i18n/de.po
@@ -47,6 +47,9 @@ msgstr "Knoten wird im Außenbereich betrieben"
 msgid "Outdoor Installation"
 msgstr "Outdoor-Installation"
 
+msgid "The configured channel for meshing is unsupported by this WLAN radio."
+msgstr "Der ausgewählte Mesh-Kanal wird von diesem WLAN-Gerät nicht unterstützt."
+
 msgid "Transmission power"
 msgstr "Sendeleistung"
 
--- a/package/gluon-web-wifi-config/luasrc/lib/gluon/config-mode/model/admin/wifi-config.lua
+++ b/package/gluon-web-wifi-config/luasrc/lib/gluon/config-mode/model/admin/wifi-config.lua
@@ -95,7 +95,7 @@ uci:foreach('wireless', 'wifi-device', function(config)
 
 	vif_option('client', {'client', 'owe'}, translate('Enable client network (access point)'))
 
-	-- disable setting mesh iface setting if radio does not support configured mesh channel
+	-- only show mesh iface setting on radios which support the configured mesh channel
 	if wireless.supports_channel(config, tonumber(config.channel)) then
 		local mesh_vif = vif_option('mesh', {'mesh'}, translate("Enable mesh network (802.11s)"))
 
@@ -104,8 +104,8 @@ uci:foreach('wireless', 'wifi-device', function(config)
 		end
 	else
 		p:element('model/warning', {
-			content = translate("Mesh deactivated due to unsupported channel"),
-		}, 'warning')
+			content = translate("The configured channel for meshing is unsupported by this WLAN radio."),
+		}, 'notice')
 	end
 
 	local phy = wireless.find_phy(config)
--- a/package/gluon-web-model/files/lib/gluon/web/view/model/notice.html
+++ b/package/gluon-web-model/files/lib/gluon/web/view/model/notice.html
@@ -1,0 +1,14 @@
+<%- if not self.hide then -%>
+<div class="gluon-section-descr"<%=
+	attr("id", id) ..
+	attr("data-index", self.index) ..
+	attr("data-depends", self:deplist(self.deps))
+%>>
+	<%- if self.content then -%>
+	<%= self.content %>
+	<%- else -%>
+	<b><%= self.title %></b><br>
+	<%= self.description %>
+	<%- end -%>
+</div>
+<%- end -%>
--- a/package/gluon-web-wifi-config/i18n/gluon-web-wifi-config.pot
+++ b/package/gluon-web-wifi-config/i18n/gluon-web-wifi-config.pot
@@ -33,6 +33,9 @@ msgstr ""
 msgid "Outdoor Installation"
 msgstr ""
 
+msgid "The configured channel for meshing is unsupported by this WLAN radio."
+msgstr ""
+
 msgid "Transmission power"
 msgstr ""
 


herbetom:
I think the mac address problem is as of now unresolved? https://github.com/freifunk-gluon/gluon/pull/1666#issuecomment-1232273562

T-X:
Would love to see the triband support! Devices with potential triband support and Gluon support I'm aware of: Plasmacloud PA2200 and the LibreRouter. Any other devices people would look forward to for triband support? Also, I guess this was untested with 6GHz so far? (I'd be looking forward to the Banana Pi BPI-R4, for instance, seems to have a dedicated phy for 6GHz for the WiFi7 iPA NIC module)

maurerle:
I think @Djfe  gave a quite extensive list of affected devices in https://github.com/freifunk-gluon/gluon/issues/1661#issuecomment-1441244292

For now, this does not include 6GHz. I have some ideas to add counters to the 6g band as well, but this would create to much noise and untested code (and I am not to familiar with 6GHz frequencies) , so I think we should do this one at a time.

From the list of open todos, I think I can handle the first two eventually, but would definitely need help for a complete overhaul (I think this way stopping the old PR mainly?). If there is someone willing to draft something where I can built on, improve and test, I would be very happy :)

maurerle:
There are concerns here regarding the probability of collision due to the reservation of the additional MAC addresses.
Steps to improve this are:

1. in batman-adv the mac address should be independent - should be checked: https://git.open-mesh.org/batman-adv.git/commit/5317331775c316
2. reconsider the assignment of mac addresses to avoid collisions
3. if we reserve more mac-addresses we have to see whether how we use them
@neocturne 

Regarding the assignment of the interfaces:
We don't need to solve this dynamically - write down the existing cases once (see all the triband devices)
and see if we have a solution for all cases.

maurerle:
If the last mac-address byte was `254` then the next mac-addresses could overflow leading to something like:

`root@ffac-lothringer-flomi:~# lua -e 'print(require("gluon.util").generate_mac(9))'`
`0a:a2:8d:fb:f9:102`

One way to solve this is to zero more than 3 bits in
`m6 = bit.band(m6, 0xF8) -- zero the last three bits (space needed for counting)`
or just let the counter overflow.

See: https://github.com/freifunk-gluon/gluon/pull/3357/commits/af539fc577f691984c08d984b81de685032bac3c

I think one can even remove the `m6 = bit.band(m6, 0xF8)` part now? @neocturne

package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
```diff
@@ -13,13 +14,19 @@ local function is_disabled(config, name)
 	return config.disabled(false)
 end
 
+local has_client_radio = false
```
maurerle, line 17: should be renamed to `has_client_only_radio` as this checks for a radio which can not be used for mesh

maurerle:
Looking at the device list provided by @djfe in https://github.com/freifunk-gluon/gluon/issues/1661#issuecomment-1441244292 I came to two possible operation modes:

a - there is 1 radio which supports only a subset of the channels while the other radio supports all unsupported channels (or more)
b - there are 2 radios with the same support channels, eventually with the same quality, eventually one radio would be preferred (though I neither know how to find the preference or set a preference to use the "better" one for mesh vs client :shrug: )

 Devices                               | 2.4GHz | 5GHz                  | Preferred Usage
---------------------------------------|--------|-----------------------|-----------------
 OpenMesh A62                          | 1      | 2                     | a
 [Plasma Cloud PA2200](https://openwrt.org/toh/hwdata/plasma_cloud/plasma_cloud_pa2200)                   | 1      | 2                     | a
 [AVM Fritz!Repeater 3000](https://github.com/freifunk-gluon/gluon/pull/3057) and 6000      | 1      | 2                     | a
 AVM Fritz!Box 4060                    | -      | no support            |
 LibreRouter v1                        | 1      | 2 (both working)      | b
 Linksys [MR8300](https://github.com/freifunk-gluon/gluon/pull/2769)/EA8300, (EA7500)       | 1      | 2                     | a
[ Xiaomi AX3600](https://openwrt.org/toh/xiaomi/ax3600)                         | 1      | 1+ 1not really usable | ?
 [Xiaomi AX9000](https://openwrt.org/toh/xiaomi/ax9000)                         | 1      | 2 (both working?)     | b
[ GL.iNet GL-B2200](https://openwrt.org/toh/gl.inet/gl-b2200)                      | 1      | 2 (both working?)     | b
 Linksys Velop (WHW01 v1 and [WHW03 v2](https://openwrt.org/toh/linksys/whw03_v2)) | 1      | 2 (different channel) | a
 Meraki MR33                           | 2      | 2                     | ?

This PR should work for the usage of a and b by default, while the user can configure different behavior.

TODO:

- [ ] test on different three-radio routers
- [ ] disable option from `gluon-web-wifi-config` to configure a wifi on an unsupported channel if `wireless.supports_channel(radio, config.channel())` is False

package/gluon-web-wifi-config/luasrc/lib/gluon/config-mode/model/admin/wifi-config.lua
```diff
@@ -95,9 +95,13 @@ uci:foreach('wireless', 'wifi-device', function(config)
 
 	vif_option('client', {'client', 'owe'}, translate('Enable client network (access point)'))
 
-	local mesh_vif = vif_option('mesh', {'mesh'}, translate("Enable mesh network (802.11s)"))
-	if is_5ghz then
-		table.insert(mesh_vifs_5ghz, mesh_vif)
+	-- disable setting mesh iface setting if radio does not support configured mesh channel
```
Djfe: Just a suggestion to use positive wording here. Instead of double negative.

> -- only show mesh iface setting if said radio supports the configured mesh channel

package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
```diff
@@ -112,4 +119,17 @@ wireless.foreach_radio(uci, function(radio, index, config)
 	configure_owe_transition_mode(config, radio_name)
 end)
```
Djfe, line 131: maybe add a code comment, what this is supposed to achieve

> -- Assign mesh radio and client radio to separate 5GHz radios if at least two exist and one of them doesn't support mesh channels. Let the one(s) that doesn't support mesh channels run as client radio and run mesh radio on all remaining 5GHz radios. 5GHz radios will have one WiFi interface each in this case.

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -205,10 +204,8 @@ function M.generate_mac(i)
 	-- a hardware MAC filter. (e.g 'rt305x')
 
 	m6 = bit.band(m6, 0xF8) -- zero the last three bits (space needed for counting)
-	m6 = m6 + i                   -- add virtual interface id
+	m6 = math.fmod(m6 + i, 256)   -- add virtual interface id
```
neocturne, line 207: This commit needs to be squashed into the previous.

Instead of `fmod`, it's better to use `m6 = bit.band(m6 + i, 0xFF)` for consistency with the rest of the code.

package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
```diff
@@ -112,4 +120,16 @@ wireless.foreach_radio(uci, function(radio, index, config)
...
 
+if not sysconfig.gluon_version and has_client_radio then
```
neocturne: Checks for `sysconfig.gluon_version` are almost always wrong. It's better to delete and recreate as much config as possible on each reconfigure instead of trying to handle the first run differently.

All new config that needs to persist should be stored in `/etc/config/gluon` - eventually I'd like to make that the only file that is preserved on upgrades at all.

package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
```diff
@@ -13,13 +15,19 @@ local function is_disabled(config, name)
 	return config.disabled(false)
 end
 
+local has_client_radio = false
```
neocturne: The rationale behind `has_client_radio` could use an explanatory comment.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -53,7 +53,7 @@ local function is_outdoor()
 end
 
 local function get_channel(radio, config)
-	if radio.band == '5g' and is_outdoor() then
+	if radio.band == '5g' and (is_outdoor() or not util.supports_channel(radio, config.channel())) then
```
neocturne: I'm wondering if something like the following would make sense:
```lua
	if (radio.band == '5g' and is_outdoor()) or not util.supports_channel(radio, config.channel()) then
```
But I guess having radios for distinct ranges isn't really a thing for 2.4GHz...

package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
```diff
@@ -24,7 +23,7 @@ local function configure_ap(radio, index, config, radio_name)
 	local ap = config.ap
 	local disabled = is_disabled(ap, name)
 
-	if not util.supports_channel(radio, config.channel()) then
+	if not wireless.supports_channel(radio, config.channel()) then
```
neocturne, line 26: Another commit that needs to be squashed. Always squash fixes, so the resulting commits make sense by themselves for review.

package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
```diff
@@ -14,7 +14,7 @@ local function is_disabled(config, name)
 	return config.disabled(false)
 end
 
-local has_client_radio = false
+local has_client_only_radio = false
```
neocturne, line 17: This commit appears to include an unrelated variable rename.

package/gluon-web-wifi-config/luasrc/lib/gluon/config-mode/model/admin/wifi-config.lua
```diff
@@ -95,9 +95,13 @@ uci:foreach('wireless', 'wifi-device', function(config)
...
+	if wireless.supports_channel(config, tonumber(config.channel)) then
+		local mesh_vif = vif_option('mesh', {'mesh'}, translate("Enable mesh network (802.11s)"))
+
+		if is_5ghz then
+			table.insert(mesh_vifs_5ghz, mesh_vif)
+		end
```
neocturne, line 104: Maybe add a notice that it's unsupported instead of just removing the checkbox?

neocturne (changes requested):

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
maurerle: Thanks!
Just a question: why is it needed to "zero the last three bits (space needed for counting)"?
I think this is just convenient as one knows which MAC is which, but it would not be required if I am not mistaken..?

package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
maurerle: I left the commits splitted, as parts were developed by blocktrron and I wanted to keep this state - but I squashed it now, for easier reviewing

herbetom: In that case "co-authored-by" should be added to the commits this affects.

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
neocturne: For some radios, all VIFs need to use MAC addresses that only differ in the last 3 bits. This was achieved by setting the last 3 bits to 0 and counting up for up to 8 addresses.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
maurerle: Yes, I think that your suggestion still makes sense - I adapted it.

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
maurerle: Thanks, then I will leave it as it is now

package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
neocturne: I'd add an optional 3rd argument to `is_disabled()` to extend the logic in the case where the UCI section does not exist yet.

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -185,9 +185,11 @@ end
 -- 5: mesh1
 -- 6: owe1
 -- 7: wan_radio1 (private WLAN); mesh VPN
+-- 8: client2
+-- 9: mesh2
+-- A: ibss2
```
neocturne, line 190: We don't support IBSS, should be owe2

I'm also pondering if we should change `generate_mac` to take 2 arguments - a radio index + an address offset 0..3, where each radio index would result in a different hash seed (and the non-WLAN callers of `generate_mac` would pass 0 or 1 as the radio index)

blocktrron: I think this was going into the direction of where i did propose this years ago.

I have another thought - In addition to calling this by indexes which effectively translate to an offset in the assigned block of n (8?) addresses, how about we use named indexes?

This way, we can track all handed out addresses by name. This has the upside of allowing 3rd party community packages to obtain unique per-radio addresses.

package/gluon-web-wifi-config/luasrc/lib/gluon/config-mode/model/admin/wifi-config.lua
maurerle: I don't quite now how this should be done with our config mode. So this is still open

neocturne: See https://github.com/freifunk-gluon/gluon/blob/main/package/gluon-web-private-wifi/luasrc/lib/gluon/config-mode/model/admin/privatewifi.lua#L29-L35 for a conditionally displayed warning. This refers to the template in gluon-web-model/files/lib/gluon/web/view/model/warning.html - a similar template could be added for a "notice".

T-X:
One point I briefly mentioned during the last meetup yesterday:

Especially if we were using higher channel bandwidths on 5GHz at some point and with BATMAN V, we might be able to achieve higher, overall throughput with mesh enabled on the secondary 5GHz, too. As then a) BATMAN V with its throughput metric would likely/often prefer just meshing on 5GHz, not 2.4GHz. And b) would be using interface alternating between the two 5GHz interfaces to avoid the half-duplex nature of WLAN: https://www.open-mesh.org/projects/batman-adv/wiki/Multi-link-optimize#Interface-alternating

But as noted during the meetup, it might maybe be difficult to agree on the exact channel for the secondary 5GHz wifi. Having this configurable in the config mode seems great for now, though.

Djfe:
@T-X suggestion to have mesh interfaces on both 5GHz phys for batman-adv interface alternating when DFS allows for the secondary channels to be used, is intriguing and a nice afterthought how to continue this work after this was merged.
(I wrote with T-X over IRC just now to clarify how his statement is meant)

It would increase throughput when no radar etc. is present and allow forwarding mesh traffic in a full duplex manner over 5GHz only.

package/gluon-web-wifi-config/luasrc/lib/gluon/config-mode/model/admin/wifi-config.lua
maurerle: Thanks, then it would just look like this:
image

What do you think of it @neocturne ? :)

Djfe: Two alternative phrasings:
- This WLAN radio doesn't support the channel range required for meshing.
- This WLAN radio doesn't support the channel selected by your community for meshing.

I like that your variant is shorter but it might leave people less knowledgeable people a bit puzzled. (yes, this is in advanced settings after all. but still)

maurerle: I added a notice as you suggested now. I think this looks better:

image

package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
maurerle: Thank you! I fixed this by checking 

`if is_disabled_wireless(radio_name) == nil then` for every radio instead.
Seemed easier in this case :)

maurerle:
I tested this. It works as expected.
Some strings should maybe adjusted, but I don't have strong feelings on that.
@neocturne can you please take another look?

Sneak preview of the channel distribution on the status-page:
image

I also tested the private-wlan package. It creates a private ssid on all three radios. Which is fine I guess.

package/gluon-web-wifi-config/luasrc/lib/gluon/config-mode/model/admin/wifi-config.lua
```diff
@@ -95,9 +95,17 @@ uci:foreach('wireless', 'wifi-device', function(config)
...
+	else
+		p:element('model/warning', {
+			content = translate("Mesh deactivated due to unsupported channel"),
```
AiyionPrime:
```suggestion
			content = translate("Mesh deactivated due to unsupported channel"),
```
```suggestion
			content = translate("Mesh deactivated due to unsupported channel."),
```

Just a nitpick; the other box texts ends in a dot :)

AiyionPrime (changes requested):
@lemoer and I took a brief look at this.
In general this looks great.

Nitpicks aside, the only thing I cannot wrap my head around, are the implications regarding mac overflow cutoffs due to the overflow prevention `band(..., 0xff)`.

Therefore the discussion I'm most interested in rn is this one _posted by @neocturne in https://github.com/freifunk-gluon/gluon/pull/3357#discussion_r2004467601_.

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
maurerle: This is now fixed with #3539 - @neocturne can you take another look please?

package/gluon-web-wifi-config/i18n/de.po
```diff
@@ -47,6 +47,9 @@ msgstr "Knoten wird im Außenbereich betrieben"
 msgid "Outdoor Installation"
 msgstr "Outdoor-Installation"
 
+msgid "The configured channel for meshing is unsupported by this WLAN radio."
+msgstr "Der ausgewählte Mesh-Kanal wird von diesem WLAN radio nicht unterstützt."
```
neocturne: Hmm, "radio" is not very common in this context in German. "WLAN-Gerät"?

neocturne (changes requested):

neocturne:
Hmm, I guess recovering the user intent regarding enabling or disabling certain radios/VIFs becomes more difficult with this PR. One idea I've been thinking of for some time - in particular in the context of moving user intent to `/etc/config/gluon` - is to remove the option to configure individual radios through Gluon's UI, instead just storing "client on 2.4GHz", "mesh on "2.4GHz" etc. and letting the config scripts deal with the details.

Djfe:
honestly that's a good point and it probably should be done. but since this only affects devices with radios with limited channel range: for now we could move that discussion to a follow up PR.
Sure that's not ideal and it could lead to said PR to not being merged before the release.
But it's out of scope and makes the discussion more convoluted, no?
I really feel like keeping the user's intent saved is important though, please don't get me wrong. But for now the user doesn't even have the option in config mode so it pretty much is the user's intent. Sure this can be enhanced later on and for old installations the intent would stay the same and transferred into the new entries in '/etc/config/gluon'

Offtopic:
Why I feel that it makes sense to brainstorm @neocturne's idea further before it can take shape:
- How do we want to structure the new wireless config option within gluon so it fits a schema for further ideas on the table?
- How do we communicate to ssh users that one wIf currently cannot be enabled? It feels weird/like a bug when there are two disabled options for the same style interface and gluon-reconfigure doesn't transfer the disabled=0 from gluon to wireless. Or is there only one general option per band and gluon assigns it to one radio and disables it for the other one on the same band? That could work actually. Though in the future we might want to have the option to assign mesh radio to both radios on the same band even though one is a non-standard frequency (for a local installation)

maurerle:
@neocturne
> I guess recovering the user intent regarding enabling or disabling certain radios/VIFs becomes more difficult with this PR.

Why do you think so? This basically sets a config for the third radio, but would respect user changes once set up?

Would you consider solving config movement to a gluon section as  outlined in #3550 mandatory for this PR or as something we can still see separately?

neocturne:
> @neocturne
> 
> > I guess recovering the user intent regarding enabling or disabling certain radios/VIFs becomes more difficult with this PR.
> 
> Why do you think so? This basically sets a config for the third radio, but would respect user changes once set up?

Hmm true, the code preserves the `disabled` values. But I wonder what would happen during a channel change - due to upgrade or domain change - that would change `has_client_only_radio` from true to false? I think we would end up with the other client radio disabled, but that does not represent user intent.

maurerle:
@neocturne 
I introduced `if is_disabled_wireless(radio_name) == nil then` instead of checking for `sysconfig.gluon_version`

So disabling of a client wifi is only done, if the config for this radio does not yet exist.
See also this comment/thread: https://github.com/freifunk-gluon/gluon/pull/3357#discussion_r2162705572

neocturne:
> @neocturne I introduced `if is_disabled_wireless(radio_name) == nil then` instead of checking for `sysconfig.gluon_version`
> 
> So disabling of a client wifi is only done, if the config for this radio does not yet exist. See also this comment/thread: [#3357 (comment)](https://github.com/freifunk-gluon/gluon/pull/3357#discussion_r2162705572)

Exactly! And it will remain disabled when the mesh channel changes, even if a configuration from scratch would enable it. In this case, it is currently impossible to recover the user intent because the information whether the setting was done by the script or the user is simply not available.

maurerle:
True, thanks.

Though in reality I found no community which has different channels in their domain.conf  - ~~all did have it in the `site.conf`.~~
And I would not expect a community to have a configured wifi5 mesh channel above 100 leading to such problems..
Or are there other scenarios in which the mesh channel could suddenly not work?

herbetom:
> Though in reality I found no community which has different channels in their domain.conf - all did have it in the site.conf.

When looking at a multitude of sites communities do define channel info inside their domains. Sometimes it's redundant, sometimes they probably changed channels during a scheduled domain switch but didn't remove the old domain and sometimes there might be other reasons:

<details>

<summary>Show grep</summary>

```bash
$ git clone https://github.com/herbetom/ff-sites.git --recurse-submodules
$ grep -ri "channel =" | grep domains
site-ffrgb/domains/ffrgb_uml.conf:    channel = 1,
site-ffrgb/domains/ffrgb_uml.conf:    channel = 44,
site-ffrgb/domains/ffrgb_cty.conf:    channel = 1,
site-ffrgb/domains/ffrgb_cty.conf:    channel = 44,
site-ffrgb/domains/ffrgb_tst.conf:    channel = 1,
site-ffrgb/domains/ffrgb_tst.conf:    channel = 44,
site-ffs/gluon/docs/multidomain-site-example/domains/alpha_centauri.conf:      channel = 1,
site-ffs/gluon/docs/multidomain-site-example/domains/alpha_centauri.conf:      channel = 44,
site-fffl/domains/Niebuell_v1.conf:        channel = 1,
site-fffl/domains/Niebuell_v1.conf:        channel = 44,
site-fffl/domains/Flensburg_v1.conf:        channel = 1,
site-fffl/domains/Flensburg_v1.conf:        channel = 44,
site-ff3l/domains/ssww.conf:		channel = 1,
site-ff3l/domains/ssww.conf:		channel = 44,
site-ff3l/domains/ffng.conf:		channel = 1,
site-ff3l/domains/ffng.conf:		channel = 44,
site-ff3l/domains/wiese.conf:		channel = 1,
site-ff3l/domains/wiese.conf:		channel = 44,
site-ff3l/domains/sswo.conf:		channel = 1,
site-ff3l/domains/sswo.conf:		channel = 44,
site-ff3l/domains/ref.conf:		channel = 1,
site-ff3l/domains/ref.conf:		channel = 44,
site-ff3l/domains/nalb.conf:		channel = 1,
site-ff3l/domains/nalb.conf:		channel = 44,
site-ff3l/domains/swb.conf:		channel = 1,
site-ff3l/domains/swb.conf:		channel = 44,
site-ff3l/domains/saek.conf:		channel = 1,
site-ff3l/domains/saek.conf:		channel = 44,
site-ff3l/domains/rhf.conf:		channel = 1,
site-ff3l/domains/rhf.conf:		channel = 44,
site-ff3l/domains/loe.conf:		channel = 1,
site-ff3l/domains/loe.conf:		channel = 44,
site-ff3l/domains/3land.conf:		channel = 1,
site-ff3l/domains/3land.conf:		channel = 44,
site-ff3l/domains/hotz.conf:		channel = 1,
site-ff3l/domains/hotz.conf:		channel = 44,
site-ff3l/domains/ffem.conf:		channel = 1,
site-ff3l/domains/ffem.conf:		channel = 44,
site-ff3l/domains/bh.conf:		channel = 1,
site-ff3l/domains/bh.conf:		channel = 44,
site-ff3l/domains/default.conf:		channel = 1,
site-ff3l/domains/default.conf:		channel = 44,
site-ff3l/domains/fftut.conf:		channel = 1,
site-ff3l/domains/fftut.conf:		channel = 44,
site-ff3l/domains/fffr.conf:		channel = 1,
site-ff3l/domains/fffr.conf:		channel = 44,
site-ff3l/domains/wtk.conf:		channel = 1,
site-ff3l/domains/wtk.conf:		channel = 44,
site-ffc/domains/neukirchen.conf:            channel = 1,
site-ffc/domains/neukirchen.conf:            channel = 36,
site-ffc/domains/zwickau.conf:            channel = 11,
site-ffc/domains/zwickau.conf:            channel = 36,
site-ffc/domains/hartmannsdorf.conf:            channel = 1,
site-ffc/domains/hartmannsdorf.conf:            channel = 36,
site-ffc/domains/penig.conf:            channel = 1,
site-ffc/domains/penig.conf:            channel = 36,
site-ffc/domains/eventnet.conf:            channel = 1,
site-ffc/domains/eventnet.conf:            channel = 36,
site-ffc/domains/hainichen.conf:            channel = 1,
site-ffc/domains/hainichen.conf:            channel = 36,
site-ffc/domains/frankenberg.conf:            channel = 1,
site-ffc/domains/frankenberg.conf:            channel = 36,
site-ffc/domains/floeha.conf:            channel = 11,
site-ffc/domains/floeha.conf:            channel = 36,
site-ffc/domains/umland.conf:            channel = 1,
site-ffc/domains/umland.conf:            channel = 36,
site-ffc/domains/chemnitz.conf:            channel = 11,
site-ffc/domains/chemnitz.conf:            channel = 36,
site-ffc/domains/mittweida.conf:            channel = 1,
site-ffc/domains/mittweida.conf:            channel = 36,
site-ffhl/gluon/docs/multidomain-site-example/domains/alpha_centauri.conf:      channel = 1,
site-ffhl/gluon/docs/multidomain-site-example/domains/alpha_centauri.conf:      channel = 44,
site-ffm/domains/ffdon_mitte.conf:    channel = 1,
site-ffm/domains/ffdon_nordwest.conf:    channel = 1,
site-ffm/domains/ffdon_sued.conf:    channel = 1,
site-ffef/gluon/docs/multidomain-site-example/domains/alpha_centauri.conf:      channel = 1,
site-ffef/gluon/docs/multidomain-site-example/domains/alpha_centauri.conf:      channel = 44,
site-ffef/domains/ffil_v1.conf:		channel = 6,
site-ffef/domains/ffil_v1.conf:		channel = 44,
site-ffef/domains/ffwsee_v1.conf:		channel = 1,
site-ffef/domains/ffwsee_v1.conf:		channel = 44,
site-ffef/domains/ffthw_v1.conf:		channel = 1,
site-ffef/domains/ffthw_v1.conf:		channel = 44,
site-ffef/domains/ffen_v1.conf:		channel = 1,
site-ffef/domains/ffen_v1.conf:		channel = 44,
site-ffef/domains/ffef_v1.conf:		channel = 1,
site-ffef/domains/ffef_v1.conf:		channel = 44,
site-ffef/domains/ffgth_v1.conf:		channel = 1,
site-ffef/domains/ffgth_v1.conf:		channel = 44,
site-fftr/domains/fftr_l2tp.conf:		channel = 1,
site-fftr/domains/fftr_l2tp.conf:		channel = 44,
site-fftr/domains/fftr.conf:		channel = 1,
site-fftr/domains/fftr.conf:		channel = 44,
site-ffv/domains/ffv_tri.conf:		channel = 11,
site-ffv/domains/ffv_tri.conf:		channel = 36,
site-ffv/domains/ffv_sbg.conf:		channel = 11,
site-ffv/domains/ffv_sbg.conf:		channel = 36,
site-ffv/domains/ffv_b.conf:		channel = 11,
site-ffv/domains/ffv_b.conf:		channel = 36,
site-ffv/domains/ffv_boe.conf:		channel = 11,
site-ffv/domains/ffv_boe.conf:		channel = 36,
site-ffv/domains/ffv_le.conf:		channel = 11,
site-ffv/domains/ffv_le.conf:		channel = 36,
site-ffv/domains/fferz.conf:		channel = 11,
site-ffv/domains/fferz.conf:		channel = 36,
site-ffv/domains/ffv_oel.conf:		channel = 11,
site-ffv/domains/ffv_oel.conf:		channel = 36,
site-ffv/domains/ffv_mtl.conf:		channel = 11,
site-ffv/domains/ffv_mtl.conf:		channel = 36,
site-ffv/domains/ffv_wei.conf:		channel = 11,
site-ffv/domains/ffv_wei.conf:		channel = 36,
site-ffv/domains/ffv_mh.conf:		channel = 11,
site-ffv/domains/ffv_mh.conf:		channel = 36,
site-ffv/domains/ffv_mkn.conf:		channel = 11,
site-ffv/domains/ffv_mkn.conf:		channel = 36,
site-ffv/domains/ffv_k.conf:		channel = 11,
site-ffv/domains/ffv_k.conf:		channel = 36,
site-ffv/domains/ffv_gb.conf:		channel = 11,
site-ffv/domains/ffv_gb.conf:		channel = 36,
site-ffv/domains/ffv_t.conf:		channel = 11,
site-ffv/domains/ffv_t.conf:		channel = 36,
site-ffv/domains/ffv_els.conf:		channel = 11,
site-ffv/domains/ffv_els.conf:		channel = 36,
site-ffv/domains/ffv_n.conf:		channel = 11,
site-ffv/domains/ffv_n.conf:		channel = 36,
site-ffv/domains/ffv_eic.conf:		channel = 11,
site-ffv/domains/ffv_eic.conf:		channel = 36,
site-ffv/domains/ffv_hdg.conf:		channel = 11,
site-ffv/domains/ffv_hdg.conf:		channel = 36,
site-ffv/domains/ffv_rc.conf:		channel = 11,
site-ffv/domains/ffv_rc.conf:		channel = 36,
site-ffv/domains/ffv_s.conf:		channel = 11,
site-ffv/domains/ffv_s.conf:		channel = 36,
site-ffv/domains/ffv_wer.conf:		channel = 11,
site-ffv/domains/ffv_wer.conf:		channel = 36,
site-ffv/domains/ffv_a.conf:		channel = 11,
site-ffv/domains/ffv_a.conf:		channel = 36,
site-ffv/domains/ffv_tr.conf:		channel = 11,
site-ffv/domains/ffv_tr.conf:		channel = 36,
site-ffv/domains/ffv_bb.conf:		channel = 11,
site-ffv/domains/ffv_bb.conf:		channel = 36,
site-ffv/domains/ffv_l.conf:		channel = 11,
site-ffv/domains/ffv_l.conf:		channel = 36,
site-ffv/domains/ffv_ell.conf:		channel = 11,
site-ffv/domains/ffv_ell.conf:		channel = 36,
site-ffv/domains/ffv_rbh.conf:		channel = 11,
site-ffv/domains/ffv_rbh.conf:		channel = 36,
site-ffv/domains/ffv_nst.conf:		channel = 11,
site-ffv/domains/ffv_nst.conf:		channel = 36,
site-ffv/domains/ffv_nsz.conf:		channel = 11,
site-ffv/domains/ffv_nsz.conf:		channel = 36,
site-ffv/domains/ffv_nmk.conf:		channel = 11,
site-ffv/domains/ffv_nmk.conf:		channel = 36,
site-ffv/domains/ffv.conf:		channel = 11,
site-ffv/domains/ffv.conf:		channel = 36,
site-ffv/domains/ffv_be.conf:		channel = 11,
site-ffv/domains/ffv_be.conf:		channel = 36,
site-ffv/domains/ffv_poe.conf:		channel = 11,
site-ffv/domains/ffv_poe.conf:		channel = 36,
site-ffv/domains/ffv_fst.conf:		channel = 11,
site-ffv/domains/ffv_fst.conf:		channel = 36,
site-ffv/domains/ffv_tdf.conf:		channel = 11,
site-ffv/domains/ffv_tdf.conf:		channel = 36,
site-ffv/domains/ffv_pmf.conf:		channel = 11,
site-ffv/domains/ffv_pmf.conf:		channel = 36,
site-ffv/domains/ffv_pl.conf:		channel = 11,
site-ffv/domains/ffv_pl.conf:		channel = 36,
site-ffv/domains/ffv_rdw.conf:		channel = 11,
site-ffv/domains/ffv_rdw.conf:		channel = 36,
site-ffv/domains/ffv_ae.conf:		channel = 11,
site-ffv/domains/ffv_ae.conf:		channel = 36,
site-fffd/gluon/docs/multidomain-site-example/domains/alpha_centauri.conf:      channel = 1,
site-fffd/gluon/docs/multidomain-site-example/domains/alpha_centauri.conf:      channel = 44,
site-ffin/domains/ffin.conf:        channel = 5,
site-ffin/domains/ffin.conf:        channel = 44,
site-ffhb/gluon/docs/multidomain-site-example/domains/alpha_centauri.conf:      channel = 1,
site-ffhb/gluon/docs/multidomain-site-example/domains/alpha_centauri.conf:      channel = 44,
site-ffhb/domains/ffhb_batv15.conf:		channel = 6,
site-ffhb/domains/ffhb_batv15.conf:		channel = 44,
site-ffmk/domains/ffmk_ost.conf:		channel = 11,
site-ffmk/domains/ffmk_ost.conf:		channel = 44,
site-ffmk/domains/ffmk_nord.conf:		channel = 11,
site-ffmk/domains/ffmk_nord.conf:		channel = 44,
site-ffmk/domains/ffmk_west.conf:		channel = 11,
site-ffmk/domains/ffmk_west.conf:		channel = 44,
```

</details>

For example Chemnitz has different 2,4 GHz channels for different domains:
https://gitlab.com/FreifunkChemnitz/site-ffc/-/blob/4964884c08763c26ddbc8d256ded9117b885c3c7/domains/floeha.conf#L12
https://gitlab.com/FreifunkChemnitz/site-ffc/-/blob/4964884c08763c26ddbc8d256ded9117b885c3c7/domains/mittweida.conf#L12

rotanid:
> Though in reality I found no community which has different channels in their domain.conf - all did have it in the `site.conf`. 

we have one domain with a different channel.
such things can happen if communities merge and/or decide to use the same firmware base.

AiyionPrime:
We talked about it two weeks back, our community intends to use it as experimental domains.
Different channel, different SSID, (different Mesh ID for good measure).

maurerle:
@herbetom thanks for doing a thorough check, I should not have generalized from a handful, sorry.

> And I would not expect a community to have a configured wifi5 mesh channel above 100 leading to such problems..
Or are there other scenarios in which the mesh channel could suddenly not work?

I still don't see where this would be an issue.
The devices have a client-only Wi-Fi because they only support outdoor channels. Therefore, this would only be an issue when switching to meshing on outdoor channels, which is also not supported with the outdoor role.

Though I see that we want to spare ourselves from doing migrations of such devices where we can't know what the original user intent is.
In that case, this PR is blocked until we rebuild the wireless config from /etc/config/gluon (#3550)

maurerle:
superseded by #3563

PR #3359, opened 2024-10-28
[v2023.2.x] scripts: add --tags to git describe in getversion.sh to include lightweighted tags (#3310)

grische:
Backport of https://github.com/freifunk-gluon/gluon/pull/3310

As this has been bothering us on our end with the latest "show version" commands, we would like to see this backported at least to v2023.2.x
image

### DIFF
--- a/scripts/getversion.sh
+++ b/scripts/getversion.sh
@@ -8,6 +8,6 @@ fi
 cd "$1" || exit 1
 
 cat .scmversion 2>/dev/null && exit 0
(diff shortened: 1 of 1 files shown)


herbetom:
Please cherry-pick with the `-x` parameter. And with there beeing a "Signed-Off-By" i would probably also do `-s` but that's nothing Gluon enforces.

grische:
Updated with `-sx`

PR #3367, opened 2024-11-04
scripts: target_config_lib: make opkg a device-specific package

### LINKED ISSUE #3365: removing opkg package not possible per-device
if you e.g. want to remove the opkg package from all tiny devices, this doesnt work

```
GLUON_SITE_PACKAGES_tiny := \
    -opkg
```

```
Refusing to remove essential package opkg.
    Removing an essential package may lead to an unusable system, but if
    you enjoy that kind of pain, you can force opkg to proceed against
    its will with the option: --force-removal-of-essential-packages
No packages removed.
```

neocturne:
If a package is installed target-wide, but removed for specific devices (or through image-customization), it is first installed during target rootfs generation, and then uninstalled again for the derived device rootfs.

For packages that are marked as essential, removal of the package will be refused by default, which is usually the correct thing to do to avoid building broken images.

For opkg, we know that it is not required for regular operation of Gluon, and removing it may be desirable for devices with small flash. By never adding opkg to the target list of packages and instead only installing it as a device-specific package, the uninstall issue can be avoided.

Removing opkg saves ~38KiB of the squashfs on mips_24kc. Stripping the opkg database is not supported however, as there is no device-specific flag for that.

---

I'm undecided if 38KiB are worth a hack like this... but at least the change is very simple.

cc @rotanid 

Fixes #3365

### DIFF
--- a/scripts/target_config_lib.lua
+++ b/scripts/target_config_lib.lua
@@ -173,7 +173,11 @@ lib.include(target)
 
 lib.check_devices()
 
-handle_target_pkgs(concat_list(get_default_pkgs(), lib.target_packages))
+-- Hack: opkg is added removed from the target packages and instead added
+-- to the devices packages to make it removable via image-customization
+local target_packages = concat_list(get_default_pkgs(), lib.target_packages)
+target_packages = concat_list(target_packages, {'-opkg'})
+handle_target_pkgs(target_packages)
 
 for _, dev in ipairs(lib.devices) do
 	local device_pkgs = {}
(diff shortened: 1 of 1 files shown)


rotanid:
> I'm undecided if 38KiB are worth a hack like this... but at least the change is very simple.

i think because of this i shouldn't be the one to decide whether to merge,
@blocktrron ?

neocturne:
Discussed at meetup; decided that 38K aren't worth a hack (and opkg is on the way to be replaced by apk in OpenWrt anyways)

PR #3376, opened 2024-11-18
ci: push image for all next branches

blocktrron:
Currently the docker image is published for the next branch. If we have two next branches (e.g. during a pending OpenWrt release), a branch like "next-24.10" would not be pushing a docker image.

Match the push action to all branches starting with next. This is helpful if a downstream CI builds using the Gluon docker image.

### DIFF
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -5,7 +5,7 @@ on:
   push:
     branches:
       - 'main'
-      - 'next'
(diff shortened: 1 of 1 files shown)


neocturne:
Not needed anymore for now, and `next*` would also match branches like `next-updates`, `next-debug`, etc. which might be unintended. We can revisit this when we have need for a secondary next branch again.

PR #3377, opened 2024-11-19
[v2021.1.x] batman-adv: update to version 2024.3

T-X:
This updates batman-adv to the latest version, as provided in the openwrt/routing feed and its master branch. The goal is to prevent the deprecated 4/32 devices stuck on OpenWrt 19.07 to disable newer batman-adv features.

The main difference to the openwrt/routing upstream version is:

* noflood mark addition, as used in Gluon (patch 2002)
* always assume absence of multicast routers (patch 9001) -> to avoid disabling multicast for packets with a routeable multicast destination address
* compat patches to be able to build on OpenWrt 19.07 / Linux 4.14 (the batman-adv compat hacks approach is too complicated due to the compat hacks from wireless-backports, therefore these patches apply/revert code inside the net/batman-adv/*.{c,h} directly, as a safer way)

This has been tested via https://github.com/T-X/gluon-batman-adv-next/ at Freifunk Lübeck for several months now (with the addition of patches 2001, 2003 and 2004).

### DIFF
--- a/patches/packages/routing/0005-batman-adv-update-to-version-2024.3.patch
+++ b/patches/packages/routing/0005-batman-adv-update-to-version-2024.3.patch
@@ -1,0 +1,5773 @@
+From: Linus Lüssing <linus.luessing@c0d3.blue>
+Date: Tue, 19 Nov 2024 14:30:33 +0100
+Subject: batman-adv: update to version 2024.3
+
+This updates batman-adv to the latest version, as provided in the
+openwrt/routing feed and its master branch. The goal is to prevent the
+deprecated 4/32 devices stuck on OpenWrt 19.07 to disable newer
+batman-adv features.
+
+The main difference to the openwrt/routing upstream version is:
+
+* noflood mark addition, as used in Gluon (patch 2002)
+* always assume absence of multicast routers (patch 9001)
+  -> to avoid disabling multicast for packets with a
+     routeable multicast destination address
+* compat patches to be able to build on OpenWrt 19.07 / Linux 4.14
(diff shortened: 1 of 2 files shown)


blocktrron:
My opinion here - We should abandon this release, it is based on a more than 5 year old OpenWrt release. The CI is also not working anymore.

T-X:
Meeting today also voted to not merge this. (There was a new question/discussion started by me if it could be added to the community repository instead then. Tendency to not add it there either, but no final decision there yet. Community packages repo README does not address this at the moment.)

Djfe:
Something I just noticed:
batman-adv dropped support for 4.14 with v2024.0 and you wanted to backport v2024.3 to OpenWrt 19.07 here (Which is kernel 4.14)
But I also haven't looked at all commits since then whether they dropped support for stuff that was necessary for older Kernels as 4.14 (not necessary since this was closed months ago)
I just wanted to leave this here, in-case this was missed back then. Since Lübeck is using this release on an unsupported older Kernel version.
https://git.open-mesh.org/batman-adv.git/blobdiff/ecdc425a87d40eba44c9c094293d50d3e9a38b45..3f2b64e972b9a491bc417c41ac8e3e14267dea15:/CHANGELOG.rst

https://git.open-mesh.org/batman-adv.git/commit/3a12789c116f05560b3e288967d19a585f79a7ce

T-X:
@Djfe right, that's how it is currently done in upstream batman-adv, kernel compatibility is there only assured until the last supported LTS kernel. But I had reverted the compat code removal that dropped 4.14 support in this fork.

PR #3384, opened 2024-12-09
mediatek-filogic: add Cudy AP3000 Outdoor v1

blocktrron:
  - [x] Web interface
  - [ ] TFTP
  - [ ] Other: <specify>
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [ ] Should show activity
  - Switch port LEDs
- Outdoor devices only:
- Cellular devices only:
- Docs:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -310,6 +310,7 @@ mediatek-filogic
 
 * Cudy
 
+  - AP3000 Outdoor (v1)
   - WR3000 (v1)
 
 * GL.iNet
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua
@@ -64,6 +64,11 @@ function M.is_outdoor_device()
 	}) then
 		return true
 
+	elseif M.match('mediatek', 'filogic', {
+		'cudy,ap3000outdoor-v1'
+	}) then
+		return true
+
 	elseif M.match('ramips', 'mt7621', {
 		'wavlink,ws-wn572hp3-4g',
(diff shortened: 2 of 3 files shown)


Djfe:
I don't think it's wise to keep adding further (mt7981) devices while their wifi is unusable with Gluon after just 1-3hours of uptime.
TQ in wifi mesh drops to 1-3% (when it was 100% in the first hour of uptime)

Please keep using the device and its wifi for a couple of days to confirm, whether it doesn't happen for this one.
We would love to roll out filogic in Aachen but the target's wifi (atleast mt7981) proofed to be too unstable for that.

autinerd:
We already have a node to test: https://map.freifunk-stuttgart.de/#!/en/map/80afca55801f, which is running for 30 days now and seems to work without issues, also the meshing TQ looks good.

herbetom:
Until there is a completed checklist i would propose marking this as a draft.

Djfe:
> We already have a node to test: https://map.freifunk-stuttgart.de/#!/en/map/80afca55801f, which is running for 30 days now and seems to work without issues, also the meshing TQ looks good.

okay TQ doesn't look particularly bad but I assume your firmware runs modprobe every 12hours to reload the wifi firmware mt7915 onto the mcu chip.

> seems to work without issues

Sounds like you haven't used the device yourself so there might've been issues. Like here.
There are no error messages in the log for these things sadly. You have to use the AP yourself to confirm whether it runs smoothly for days.

Yes it does traffic, I can see that. Awesome! :)
but that doesn't mean the user experience can't be bad at the same time.
We got very weary regarding these because they ran unstable for us. (unlike ramips-mt7621 which only had outages from time to time (mcu timeout because it hangs), filogic runs unstable for us without actual outages, which is worse)

PR #3387, opened 2024-12-12

herbetom:


maurerle:
I would wait for the backport of https://github.com/openwrt/openwrt/pull/17265

rotanid:
> I would wait for the backport of [openwrt/openwrt#17265](https://github.com/openwrt/openwrt/pull/17265)

we can always bump again.

Issue #3389, opened 2024-12-13
Fritz!Box 4040 eth0 mac is random on fresh installs


Djfe:
@herbetom reported the eth0 mac to be random since merging gluon next
https://paste.linuxlounge.net/QA

it might be related to this commit:
https://github.com/openwrt/openwrt/commit/b22d382ae4eaa1af42930115d91855f402314cac

maybe something broke the label mac since then or the label mac never worked at all

Djfe:
Update:
thanks to @neocturne
```
[13:26] <neocturne> djfe: the problem with the MAC address on ipq40xx is the missing alias ethernet0 in the Device Tree, which is used by the bootloader to fill in MAC addresses. It was previously incorrectly set in the SoC DTSI, but it belongs in in the board DTS(I). When it got removed from the SoC DTSI due to the update to Linux 6.6, nobody fixed it in the board DTS.
```

Djfe:
related
https://github.com/chunkeey/FritzBox-4040-UBOOT/pull/11

rotanid:
this might fix it:
https://github.com/openwrt/openwrt/pull/17432

maurerle:
I could not reproduce this issue on a FB7530, as my MAC-Address stays the same throughout boots and fresh installs.

Though I found a different MAC-Address related issue for my FB7530, mentioned in #3410

rotanid:
fixed in OpenWrt main, waiting for backport to 24.10 probably?
https://github.com/openwrt/openwrt/pull/17442

PR #3396, opened 2024-12-17
ipq40xx-generic: add extreme-networks-ws-ap3915i

seilpforte:
  - [ ] Web interface
  - [x] TFTP
  - [ ] Other: <specify>
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [x] Should show activity
  - Switch port LEDs
- Outdoor devices only:
- Cellular devices only:
- Docs:

Primary MAC Adress is: ca:b2:ef:60:c6:63
Eth0 MAC is: dc:b8:08:72:a2:ad
Label MAC is: dc:b8:08:72:a2:ad

Keeps forgetting config after any sysupgrade

### DIFF
--- a/targets/ipq40xx-generic
+++ b/targets/ipq40xx-generic
@@ -154,3 +154,9 @@ device('zyxel-nbg6617', 'zyxel_nbg6617')
 device('8devices-jalapeno', '8dev_jalapeno', {
 	factory_ext = '.ubi',
 })
+
+-- Extreme Networks
+
+device('extreme-networks-ws-ap3915i', 'extreme-networks_ws-ap3915i', {
+	factory = false,
+})


maurerle:
This looks like a related thread with a possible solution:
https://forum.openwrt.org/t/ws-ap3915i-settings-lost-upgrading-23-05-0-23-05-2/179770

setting `BLOCKSIZE := 64k` for the sysupgrades instead..?

blocktrron:
Primary MAC  also needs to be fixed in order to consider merging.

maurerle:
Should this be fixed through label-mac in openwrt or through 010-mac-address in gluon?

blocktrron:
Preferably this should be fixed in openwrt if possible.

maurerle:
We tried a few things in openwrt:
https://github.com/openwrt/openwrt/pull/17305

We could fix the BLOCKSIZE issue - making sysupgrades keep the config.
However I could not set the label-mac based on the OpenWRT output.

I tried:
`label_mac="$(mtd_get_mac_ascii CFG1 ethaddr)"` - during the running system, this shows the correct macaddress.
This is also how it was done for the `ws-ap3710i`
https://github.com/openwrt/openwrt/blob/e321e70ddcd82e0b4a5ad695b83a88d44cde02b4/target/linux/mpc85xx/base-files/etc/board.d/02_network#L21

Snippet:
```
. /lib/functions.sh
mtd_get_mac_ascii CFG1 ethaddr
```

I also tried different variants of adding `label-mac-device = &eth0;` to the dts, which somehow did not exist though.

Eventually this is related to something being broken for ipq40xx (as mentioned in #3389) - so I could need a hint here what to try :)

rotanid:
@maurerle can you please rebase and test again, there seems to be a linter issue

maurerle:
Tested again after rebase - fixed the wrong spacing.
Is ready from my side :)

Djfe:
@seilpforte can you update the check list, please? (primary mac and must/keep forget configuration are both working now)

I just want to avoid confusion as this checklist counts as some form of documentation

maurerle:
More important, we forgot to add it to the supported devices list :(

rotanid:
i edited the checklist. can't do the supported devices atm.

rotanid:
thanks, @maurerle fixed it with #3413

PR #3401, opened 2024-12-23
Update openwrt main

maurerle:
Makes way for support of ap3915i


rotanid:
@maurerle it seems a refresh of a patch is missing for the CI check to succeed

rotanid:
@maurerle xrx200 build failed again though.

maurerle:
The TD-W8970 and TD-W8980 are affected from the lantiq-xrx200 to lantiq-xrx200_legacy migration.
I am unsure how our process to handle the target migration is:

1. just add the new xrx200_legacy target and readd the supported devices
2. remove support for now and readd after checking device compatibility again

Only 4 devices of the `TP-Link TD-W89x0` kind using gluon are currently active https://stats.darmstadt.freifunk.net/d/CA5PRFmMz/gluon-census3a-dashboard?orgId=1&from=now-30d&to=now&var-community=Darmstadt&var-version_major=v2023.2&var-version=v2023.2.4&var-version=v2023.2.3&var-models=TP-LINK%20TD-W8970&refresh=15m&timezone=browser&viewPanel=panel-22

For now, I just readded the device in a separate commit, as this is more a formal removal and did not look like a greater technical change to me.

targets/lantiq-xrx200_legacy
```diff
@@ -0,0 +1,30 @@
...
+
+device('tp-link-td-w8970', 'tplink_tdw8970', {
+	factory = false,
+})
+
+device('tp-link-td-w8980', 'tplink_tdw8980', {
+	factory = false,
+	aliases = {'tp-link-td-w9980'},
+	broken = true, -- 5GHz unsupported
+})
```
rotanid, line 30: GitHub shows a "(-)" sign after line 30, maybe some kind of problem with line ending here?

rotanid:
thanks for your effort, build of the "new" target is now what is failing (at least, not all builds finished by the time of writing this) :-/

targets/lantiq-xrx200_legacy
maurerle: A missing empty newline at the end was the reason for this (and for the lint job failure).
But the build failure came from target_config_lib.lua which did replace all _ in the openwrt target name, instead of only the first one.. :)

blocktrron, line 30: This also required a documentation change of the supported target assigments.

rotanid:
@Djfe mentioned on IRC, there might be issues with the two devices we move with this PR to legacy, https://github.com/openwrt/openwrt/pull/16555 https://github.com/openwrt/openwrt/pull/9775

targets/lantiq-xrx200_legacy
rotanid: @maurerle https://github.com/freifunk-gluon/gluon/blob/main/docs/user/supported_devices.rst

Djfe:
I had the same question as @maurerle had: how do we handle target migration? (I hadn't seen this thread, yet)
Or do we drop support instead? (depends on whether the effort makes sense here)

#16555 is just clean up within OpenWrt's main branch. Openwrt devs are cleaning up dts files and transferring manual configs into dts files wherever possible. this will reduce maintenance and allows those files to be upstreamed into the Linux kernel at some point.

#9775 is just a PR I saw some time back then which could elleviate certain issues here and restructured the dts file iirc.
Devs upstream already took a look at the old PR and brought it up to speed with current dts changes. I could help in getting it tested or so.

I have a TD-W8970 in my flat and could take a look (redo our checklist and help in finalizing #9775 )
I myself don't require the device to be working but since upstream already did lots of work it's likely not that hard 😅

I assume both TD-W devices will be dropped at some point anyways due to little amount of flash (in a later release)

also in that same motion I could upstream this patch of ours to Gluon:
https://github.com/ffac/site/blob/v2025.1.x/patches/020-interfaces.patch#L28
It separates LAN and WAN ports in Gluon (one is labeled LAN4/WAN and used as LAN in OpenWrt)

blocktrron:
@rotanid Depends on the outlook whether or not additional devices will be moved in the future.

That being said, I'm fine wiht migrating for now and postponing the decision towards the next release.

PR #3406, opened 2024-12-27
core: remove old migrations

maurerle:
v2023.2 did only support upgrades from v2022.x
we can therefore remove these old migrations for the next release

I hope, that I did not miss anything, but to me these were the only occurances of migrations to be deleted.


rotanid:
@maurerle please have a look at the linter error :)
` package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless:3:7: (W211) unused variable util`

rotanid:
i think it looks good now, but i would be more happy if @blocktrron or @neocturne decide about the merge

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -223,11 +222,6 @@ wireless.foreach_radio(uci, function(radio, index, config)
 		uci:set('wireless', radio_name, 'legacy_rates', false)
 		configure_mesh_wireless(radio, index, config)
 	elseif (band == '5g') then
-		-- ToDo: Remove in v2024.x
-		local hostapd_options = uci:get_list('wireless', radio_name, 'hostapd_options')
```
neocturne, line 227: This migration was added in 2023.1, I think it is too early to remove it.

Djfe: What do you mean?
v2023.1 means: The migration is for v2022.1 and v2021.1
from v2022.1 you can upgrade to v2023.2 but not till gluon main/v2025.1

Removing this should therefore be fine, right? (Even the code comment says so)

rotanid: @Djfe whats the problem with leaving it for another release?

Djfe: conclusion in the meeting was:
let's keep these lines (don't remove them) and then this PR will be merged :)
They don't harm.
We will remove them later

rotanid: @maurerle :)

maurerle: Done :) LGTM :)

package/gluon-core/luasrc/lib/gluon/upgrade/190-preserve-wireless-channels
```diff
@@ -1,23 +0,0 @@
...
-	end
-
-	uci:delete('gluon-core', core_wireless)
-	uci:save('gluon-core')
-end
-
-uci:section('gluon', 'wireless', 'wireless', {
-	preserve_channels = preserve_channels or false,
-})
-uci:save('gluon')
```
neocturne, line 23: Hmm, maybe we want to keep the initial setup of `gluon.wireless.preserve_channels` and only delete lines 8..18?

maurerle: Fine for me as well, so that we keep setting the default value if it does not exist.
I just pushed an update to this.

PR #3412, opened 2025-01-05
gluon-private-wifi: rename psk3 to sae as it is used in openwrt that way

maurerle:
I had a hard time finding information about psk3 in openwrt documentation as it is not mentioned here:

https://openwrt.org/docs/guide-user/network/wifi/basic#encryption_modes

I think the naming changed some time, so it makes sense to use the documented value of sae instead of psk3.

The functionality in openwrt is the same, so this does not change any functionality ~~and there is no good reason to migrate existing configs~~.
Though this helps to better find related documentation to the value set in `uci show wireless`

### DIFF
--- a/package/gluon-private-wifi/luasrc/lib/gluon/upgrade/325-gluon-private-wifi
+++ b/package/gluon-private-wifi/luasrc/lib/gluon/upgrade/325-gluon-private-wifi
@@ -13,6 +13,11 @@ wireless.foreach_radio(uci, function(radio)
 	end
 
 	uci:set('wireless', name, 'ifname', suffix and 'wl-wan' .. suffix)
+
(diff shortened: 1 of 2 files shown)


blocktrron:
You need a migration as the current state display in the frontend is used as the identifier.

maurerle:
Isn't it enough to replace the string when setting the default like I do in this PR?

neocturne:
> Isn't it enough to replace the string when setting the default like I do in this PR?

Would be cleaner to migrate it once in an upgrade script instead of keeping the string replacement in the UI forever.

blocktrron:
I'm with @neocturne on this - Your approach would need to be kept indefinitely while a migration can be removed at one point in the future.

PR #3416, opened 2025-01-08
mac80211: silence mesh rate mismatch warning

blocktrron:
Printing a warning everytime a rate mismatch is detected is a bit to much. With many mesh neighbors, this quickly results in the kernel log being spammed with these messages.

### DIFF
--- a/patches/openwrt/0007-net-mac80211-override-incompatible-basic-rates-for-mesh.patch
+++ b/patches/openwrt/0007-net-mac80211-override-incompatible-basic-rates-for-mesh.patch
@@ -2,14 +2,29 @@ From: David Bauer <mail@david-bauer.net>
 Date: Mon, 6 Jan 2025 08:30:35 +0100
 Subject: net: mac80211: override incompatible basic-rates for mesh
 
+This is a dirty hack for Gluon.
+
+We assume basic rate setup only affects the rate-controller on the TX
+side. As all devices we support have at least a 802.11n radio and thus
+cover 802.11b as well as 802.11g on 2.4 GHz, they are compatible with
+each other.
+
+As the basic rate was incorrectly set for mesh interfaces in the past,
+connections between mesh neighbors would fail when altering the basic
+rate.
+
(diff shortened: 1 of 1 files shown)


maurerle:
I have a node with 2 mesh partners, which is directly spammed very much with these records - like 30 messages per second.

`ignoring basic rate mismatch for peer (local=40 peer=1)`

Is this normal without silencing it?

What is needed to fix the mismatch warning? If both covr-x1860 are using latest gluon main - they should not have this warning anymore right?
Do you need to restart wifi or reboot or something to fix this, after both have the main version and the third router without it is removed?

maurerle:
reboot did not help - there are only two COVR-X1860 seeing each other with gluon-main firmware.
Both are spamming the `ignoring basic rate mismatch for peer (local=40 peer=1)` message.

patches/openwrt/0007-net-mac80211-override-incompatible-basic-rates-for-mesh.patch
```diff
@@ -35,24 +50,15 @@ index 0000000000000000000000000000000000000000..19ca64d5012d2974ab2a48e9363ecaa3
...
 ++++ b/net/mac80211/mesh.c
-+@@ -92,8 +92,11 @@ bool mesh_matches_local(struct ieee80211_sub_if_data *sdata,
++@@ -94,9 +94,6 @@ bool mesh_matches_local(struct ieee80211
 + 	ieee80211_sta_get_rates(sdata, ie, sband->band,
 + 				&basic_rates);
 + 
 +-	if (sdata->vif.bss_conf.basic_rates != basic_rates)
 +-		return false;
-++	if (sdata->vif.bss_conf.basic_rates != basic_rates) {
-++		wiphy_warn(sdata->wdev.wiphy,
```
grische, line 49: Would it help to use `wiphy_warn_ratelimited` instead?
See https://github.com/torvalds/linux/blob/v6.6/include/net/cfg80211.h#L9094-L9097

PR #3418, opened 2025-01-09
ipq40xx-generic: add support for linksys-whw03v2

maurerle:
  - [x] Web interface
  - [ ] TFTP
  - [ ] Other: <specify>
        (`lua -e 'print(require("platform_info").get_image_name())'`) - `linksys-whw03-v2-velop`
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [ ] Should show activity
  - Switch port LEDs
- Outdoor devices only:
- Cellular devices only:
- Docs:

This device relies on #3357  and upstream merging of https://github.com/openwrt/openwrt/pull/17535


maurerle:
The device works fine with latest gluon and #3563.
The device responds rather slow when pressing the reset button until entering the config mode. So it takes 3-4 minutes to enter setup-mode and have working ports there.
The LED works just fine - but after pressing to enter config-mode it still takes ~30 seconds until the device reboots.
When calling `reboot` from the terminal, it also takes similarly long.
After the lights are turned off, the device is 90 seconds without any status sign turned off - then takes 20 seconds fast blinking until it reaches the config-mode blinking sequence.

Other than that, the device works just fine.
With ath10k instead of ath10k-ct, the log is also clean

PR #3419, opened 2025-01-09
mac80211: only override advertised basic-rate on 2.4 GHz

blocktrron:
Previously the basic-rate was set to 1 Mbit/s 802.11b on VIFs which operate on non-g band. This broke meshes on 5 GHz.


maurerle:
I do get `ERROR: package/kernel/mac80211 failed to build (build variant: regular)` when adding this patch.
Build with V=sc is in progress

maurerle:
I still get

```
build_dir/target-mipsel_24kc_musl/linux-ramips_mt7621/mac80211-regular/backports-6.12.6/net/mac80211/mesh.c:1090:25: error: comparison between 'enum nl80211_band' and 'enum ieee80211_radiotap_channel_flags' [-Werror=enum-compare]
 1090 |         if (sband->band == IEEE80211_CHAN_2GHZ) {
      |                         ^~
cc1: all warnings being treated as errors
```

maurerle:
Hi @blocktrron, I just tested this patch after `s/IEEE80211_CHAN_2GHZ/NL80211_BAND_2GHZ/` which compiled fine, but has the same behavior as before somehow.
Nodes on v2023.2.3 are seen on the status page, but no link is established in `batctl n`

maurerle:
I just tested the latest push to pretend `BIT(4) | BIT(6) | BIT(8)` as basic rate for 5GHz mesh, but unfortunately it did not make a COVR-X1860 on gluon-main see its v2023.2.3 neighbours on the 5GHz wifi again :(

blocktrron:
@maurerle should be fixed now.

maurerle:
Yes, works fine! Great :)

PR #3420, opened 2025-01-09
ramips-mt7620: Add support for Netgear EX6120

fabianh95:
  - [x] Web interface
  - [ ] TFTP
  - [x] nmrpflash
       ->  netgear-ex6120
      (https://gluon.readthedocs.io/en/latest/dev/hardware.html#hardware-support-in-packages) -> MAC-Address match eth0 and br-wan but not br-client (last digit changed)
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [ ] Should map to their respective radio ->  led labeled "router" and led labeled "device" was controllable in openwrt but is off with gluon
    - [ ] Should show activity ->  led labeled "router" and led labeled "device" was controllable in openwrt but is off with gluon
    - [ ] Should map to their respective port (or switch, if only one led present) ->  no extra led
    - [ ] Should show link state and activity ->  no extra led
- Outdoor devices only:
- Cellular devices only:
- Docs:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -405,6 +405,7 @@ ramips-mt7620
 
   - EX3700
   - EX3800
+  - EX6120
   - EX6130
 
 * Nexx
--- a/targets/ramips-mt7620
+++ b/targets/ramips-mt7620
@@ -30,6 +30,10 @@ device('netgear-ex3700', 'netgear_ex3700', {
 	factory_ext = '.chk',
 })
 
+device('netgear-ex6120', 'netgear_ex6120', {
+	factory_ext = '.chk',
+})
+
 device('netgear-ex6130', 'netgear_ex6130', {
 	factory_ext = '.chk',
 })


herbetom:
Looks good to me, just one clarification:

> MAC-Address match eth0 and br-wan but not br-client (last digit changed)

The primary mac is the one one the status page or in the output of `gluon-info`. And that one matches with the one printed on the device? The mention of those interface names is just you wondering about that?

maurerle:
@fabianh95 ping

could you report if the information from the status page or in the output of gluon-info matches the label-mac written on the device?

fabianh95:
@herbetom Thanks for your answer and sorry for the late response.

I had a lot to do and I bought an EX6150 for testing.

XX:XX:XX:XX:XX:39  <--- printed on the device label and is the MAC-Address of eth0 and br-wan (via ifconfig)

gluon-info print XX:XX:XX:XX:XX:38, witch is the MAC-Address of client0, br-client and bat0 (via ifconfig)

On the EX6150 there is the same behavior, but on the device label are both MAC-Addresses printed. On the label from the EX6120 there is only one MAC-Address printed.

maurerle:
okay @fabianh95 - can you build the firmware for both devices with this patch applied and test if the mac is set correctly:
https://github.com/openwrt/openwrt/pull/17760.patch

The snippet below should then show the correct label-mac as printed on the device:

```
. /lib/functions.sh
. /lib/functions/system.sh
echo $(get_mac_label)
```
or just use `firstboot`, which regenerates the device mac as well :)

blocktrron:
@fabianh95 can you report back if yo uwere able to test with the patch provided? If you are short on time, I owuld close this PR for now and you can reopen it once you have time to test.

AiyionPrime:
@fabianh95 First of all, thanks for your work so far.
I hope you understand our desire for a clean PR section, so that we can focus on what's actively developed.

As you did not respond for months now, I'll follow up on @blocktrron's suggestion and close this for the time being.

As soon as you find the time to get back to this, feel free to reopen the PR, so we can continue, where you left off.

All the best until then,
Aiyion

PR #3424, opened 2025-01-15
docs: user/getting_started: add dependencies

herbetom:
Changing the docs as well was missed in b1cdcc8784aa06fe22eff9aeb215244b8c429563

Reported-in: https://github.com/freifunk-gluon/gluon/pull/3414
Reported-by: 132701936+ChristianMiddendorf@users.noreply.github.com

### DIFF
--- a/docs/user/getting_started.rst
+++ b/docs/user/getting_started.rst
@@ -29,8 +29,10 @@ freshly installed Debian Bullseye system the following packages are required:
 
 * `git` (to get Gluon and other dependencies)
 * `python3`
+* `python3-dev`
+* `python3-pyelftools`
+* `python3-setuptools`
 * `build-essential`
-* `ecdsautils` (to sign firmware, see `contrib/sign.sh`)
 * `gawk`
 * `unzip`
 * `libncurses-dev` (actually `libncurses5-dev`)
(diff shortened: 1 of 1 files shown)


herbetom:
Rendered version appears to look fine: https://gluon--3424.org.readthedocs.build/en/3424/user/getting_started.html#dependencies

blocktrron:
Please refer the dockerfile in the documentation instead of duplicating.

I also assume one would rather want a install command he can just copy and paste.

AiyionPrime:
Just a general minor gesture, but I'd try to add Christian as author of the commit, if it's his contribution.
Feel free to add yourself as a co-author or the other way around, so that others can infer whose idea this was by looking at the git log :)

This might show a little more appreciation for the work they've put into spotting this.

herbetom:
I think i added Christian as "Reported-" because i took the idea of fixing that from there but didn't actually use those changes and with the commit message also not beeing what we need i didn't really felt comfortable authoring that in someone elses name. But adding a "Co-Authored-By" seems like a good idea.
---

Regarding referring to the Dockerfile: I'm not going to look into that further because while it seems like a sensible idea it either would be something without a link (which also isn't that user friendly) or it would probabbly end up pointing to the main branch (at least I can't think of a good way to make the link dynamic) which also would be supoptimal. So in that regard it's a "take it or leave it" regarding my involvement :see_no_evil:

PR #3430, opened 2025-01-22
gluon-core: fixes wifi on mediatek-mt7622 through a fallback of renamed wireless phys

maurerle:
tested on a Netgear WAX206

though I think that backporting the commit which fixes renaming of phy devices might fix this as well:
https://github.com/openwrt/openwrt/commit/ed34e337a959919048e393f4b80d22aceffc3218
(No, the files do not exist on openwrt-24.10)

@RolandoMagico can you test my PR on your device as well?

Maybe there is a better way to do this, but this is just a quick fix to get things working again.

`iwinfo.nl80211.phyname(config['.name']:gsub("radio", "phy"))` returns phy0 again.

On my device `iwinfo phy0 htmodelist` shows correct values, while

```
root@ffac-000c43266078:~# uci show wireless
wireless.radio0=wifi-device
wireless.radio0.type='mac80211'
wireless.radio0.phy='wl0'
wireless.radio0.band='2g'
wireless.radio0.channel='11'
wireless.radio0.htmode='HT20'
wireless.radio0.country='DE'
wireless.radio0.legacy_rates='0'
wireless.radio1=wifi-device
wireless.radio1.type='mac80211'
wireless.radio1.phy='wl1'
wireless.radio1.band='5g'
wireless.radio1.channel='44'
wireless.radio1.htmode='HE20'
wireless.radio1.country='DE'
```

### DIFF
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/wireless.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/wireless.lua
@@ -9,7 +9,11 @@ local iwinfo = require 'iwinfo'
 local M = {}
 
 function M.find_phy(config)
-	return iwinfo.nl80211.phyname(config['.name'])
+	local phyname = iwinfo.nl80211.phyname(config['.name'])
+	if not phyname then
+		phyname = iwinfo.nl80211.phyname(config['.name']:gsub("radio", "phy"))
+	end
+	return phyname
 end
 
 local function get_addresses(radio)


maurerle:
@neocturne  could you take a look at this, please?

A fix like this is required for #3431 and #3332

RolandoMagico:
@maurerle: Thanks for the PR. Wifi setttings are now available in config mode and Wifi also works in normal mode on my R32

package/gluon-core/luasrc/usr/lib/lua/gluon/wireless.lua
```diff
@@ -9,7 +9,11 @@ local iwinfo = require 'iwinfo'
 local M = {}
 
 function M.find_phy(config)
-	return iwinfo.nl80211.phyname(config['.name'])
+	local phyname = iwinfo.nl80211.phyname(config['.name'])
+	if not phyname then
+		phyname = iwinfo.nl80211.phyname(config['.name']:gsub("radio", "phy"))
```
maurerle, line 14: This is the same as:
`phyname = config['.name']:gsub("radio", "phy")`

Though such a fallback was not necessary on openwrt

neocturne: PHY and radio numbering should be considered unrelated, there is no guarantee that PHY numbers don't change (on upgrades or possibly reboots).

I believe that this must be fixed in iwinfo itself.

maurerle: I think, that something goes wrong here:
https://github.com/openwrt/openwrt/blob/1e9966a63ae430a5c48039fd9752bf1c465325f3/package/network/config/wifi-scripts/files/lib/netifd/wireless/mac80211.sh#L619

As I think I saw the `Could not find PHY for device ..` message in logread or something similar, which is issued by `find_phy`

maurerle: There is this is the logread:

```
Thu Jan 23 04:39:30 2025 daemon.notice netifd: radio1 (2778): Command failed: Not found
Thu Jan 23 04:39:30 2025 daemon.notice hostapd: Set new config for phy wl0:
Thu Jan 23 04:39:30 2025 daemon.notice netifd: radio0 (2774): Command failed: Not found
Thu Jan 23 04:39:31 2025 daemon.notice hostapd: Set new config for phy wl0: /var/run/hostapd-wl0.conf
Thu Jan 23 04:39:31 2025 daemon.notice hostapd: Restart interface for phy wl0
Thu Jan 23 04:39:31 2025 daemon.notice hostapd: Configuration file: data: driver=nl80211 logger_syslog=127 logger_syslog_level=2 logger_stdout=127 logger_stdout_level=2 country_code=DE ieee80211d=1 hw_mode=g supported_rates=60 90 120 180 240 360 480 540 basic_rates=60 120 240 beacon_int=100 stationary_ap=1 chanlist=11 noscan=1 #num_global_macaddr=1 #macaddr_base= ieee80211n=1 ht_coex=0 ht_capab=[LDPC][SHORT-GI-20][SHORT-GI-40][TX-STBC][RX-STBC1] channel=11  interface=client0 bssid=72:90:4e:18:f7:30 ctrl_interface=/var/run/hostapd ap_isolate=1 bss_load_update_period=60 chan_util_avg_period=600 disassoc_low_ack=1 skip_inactivity_poll=0 preamble=1 wmm_enabled=1 ignore_broadcast_ssid=0 uapsd_advertisement_enabled=1 utf8_ssid=1 multi_ap=0 auth_algs=1 wpa=0 ssid=Freifunk bridge=br-client wds_bridge= snoop_iface=br-client qos_map_set=0,0,2,16,1,1,255,255,18,22,24,38,40,40,44,46,48,56 nas_identifier=72904e18f730  (phy wl0) --> new PHY
Thu Jan 23 04:39:31 2025 daemon.notice hostapd: Set new config for phy wl1: /var/run/hostapd-wl1.conf
Thu Jan 23 04:39:31 2025 daemon.notice hostapd: Restart interface for phy wl1
```

Does this help somehow @neocturne ?

maurerle: The `iface_update_supplicant_macaddr` call on reload was added in
https://github.com/openwrt/openwrt/commit/20c667cc88176552e4f5769051dff63819cd8177

I think that this is what causes the "Not found" error and probably is the root cause of this issue, why reloading fails.

Eventually this is a different error, as I do not really see, how this impacts that the renaming fails

maurerle:
Alternatively, one can run:

```
. /lib/netifd/wireless/mac80211.sh 
phy=wl0
find_phy
phy=wl1
find_phy
gluon-reconfigure
```

in config-mode. I don't know why this is skipped in gluon, while it runs in openwrt.
find_phy should be running in `drv_mac80211_setup`

neocturne:
> Alternatively, one can run:
> 
> ```
> . /lib/netifd/wireless/mac80211.sh 
> phy=wl0
> find_phy
> phy=wl1
> find_phy
> gluon-reconfigure
> ```
> 
> in config-mode. This manually triggers the renaming, after which everything works as expected

Interesting. We'll have to check if it makes sense to duplicate more code from the shell script in Gluon, or what a good place to run the shell script is.

maurerle:
Normally, the `find_phy` funciton is called in `drv_mac80211_setup` in mac80211.sh - I did not find any occurance, where this is called.
Do you know how this works @neocturne ? How does this work for gluon in contrast?

maurerle:
The board.json of the Netgear WAX 206 is given here:

<details>

```json
{
	"model": {
		"id": "netgear,wax206",
		"name": "Netgear WAX206"
	},
	"led": {
		"wan": {
			"name": "WAN",
			"sysfs": "inet:blue",
			"type": "netdev",
			"device": "wan",
			"mode": "link tx rx"
		}
	},
	"network": {
		"lan": {
			"ports": [
				"lan1",
				"lan2",
				"lan3",
				"lan4"
			],
			"protocol": "static"
		},
		"wan": {
			"device": "wan",
			"protocol": "dhcp"
		}
	},
	"wlan": {
		"wl0": {
			"path": "platform/18000000.wmac",
			"info": {
				"antenna_rx": 15,
				"antenna_tx": 15,
				"bands": {
					"2G": {
						"ht": true,
						"max_width": 40,
						"modes": [
							"NOHT",
							"HT20",
							"HT40"
						],
						"default_channel": 1
					}
				},
				"radios": [
				]
			}
		},
		"wl1": {
			"path": "1a143000.pcie/pci0000:00/0000:00:00.0/0000:01:00.0",
			"info": {
				"antenna_rx": 15,
				"antenna_tx": 15,
				"bands": {
					"5G": {
						"ht": true,
						"vht": true,
						"he": true,
						"max_width": 160,
						"modes": [
							"NOHT",
							"HT20",
							"VHT20",
							"HE20",
							"HT40",
							"VHT40",
							"HE40",
							"VHT80",
							"HE80",
							"HE160"
						],
						"default_channel": 36
					}
				},
				"radios": [
				]
			}
		},
		"wl2": {
			"path": "1a145000.pcie/pci0001:00/0001:00:01.0/0001:01:00.0"
		}
	}
}
```
</details>

Djfe:
I just looked at the device again. Gluon only defines 112MiB of the 256MiB flash in it's mtd partitions. I feel like we should adjust this now, so everyone can profit from this and so it won't be done later (after we add the device). because adjusting a ubi partition causes trouble and requires a factory install for example with nmrpflash (we need to increase major compat for this).
Gluon doesn't require the rest of the flash to be usable, but OpenWrt users profit.

rotanid:
will be fixed with https://github.com/openwrt/openwrt/pull/17821 probably

maurerle:
This PR does not try to solve the root problem but fixes the issue, that the phy is not renamed during config mode for mt7622.

The upstream fix for this problem in openwrt https://github.com/openwrt/openwrt/pull/17821 seems to take a little more time.

The other attempts (#3408, and #3438) are more complex, so I would go for the easiest successful workaround until an upstream fix is available..?

maurerle:
As the upstream PR seems not be needed anymore in OpenWrt, as the issue seems to be resolved:
https://github.com/openwrt/openwrt/issues/14207#issuecomment-3276242343

I took a further look into the remaining issue in gluon.

@neocturne had concerns, that radioX should not depend on phyX/wlX. This is fine.
Instead, phyX and wlX depend on each other.
Therefore, we are looking for phyX if we do not find anything with the given (wlX) name.

This is done about 4 times in openWrt iwinfo ucode as well:

https://github.com/openwrt/openwrt/blob/f5655704a7892c772e1d32c6884a8aef5092040c/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/iwinfo.uc#L60

https://github.com/openwrt/openwrt/blob/f5655704a7892c772e1d32c6884a8aef5092040c/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/iwinfo.uc#L264

So I actually think that this is the right fix for the mt7622 issue in gluon..? Until someone else tells me differently.

Tested with latest main on a unifi 6-lr v3

rotanid:
small remark: the commit message and the comment in the code need some polish

maurerle:
superseded by #3598

PR #3431, opened 2025-01-22
mediatek-mt7622: add support for netgear wax206

maurerle:
  - [ ] Web interface
  - [ ] TFTP
  - [ ] Other: <specify>
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [ ] Should show activity
  - Switch port LEDs
- Outdoor devices only:
- Cellular devices only:
- Docs:

## Caveats

* on the test device, the internet wan port did not work until `ethtool --set-eee internet wan off tx-lpi off` is issued. This happens on some devices on 24.10 - see https://github.com/openwrt/openwrt/issues/17351
* the LEDs are working once https://github.com/openwrt/openwrt/pull/17694 is merged and backported

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -349,6 +349,10 @@ mediatek-mt7622
 
   - E8450
 
+* Netgear
+
+  - WAX206
+
 * Ubiquiti
 
   - UniFi 6 LR (v1)
--- a/targets/mediatek-mt7622
+++ b/targets/mediatek-mt7622
@@ -6,6 +6,13 @@ device('linksys-e8450-ubi', 'linksys_e8450-ubi', {
 })
 
 
+-- Netgear
+
+device('netgear-wax206', 'netgear_wax206', {
+	factory_ext = '.img',
+})
+
+
 -- Ubiquiti
 
 device('ubiquiti-unifi-6-lr-v1', 'ubnt_unifi-6-lr-v1', {


Djfe:
I just looked at the device again. Gluon only defines 112MiB of the 256MiB flash in it's mtd partitions. I feel like we should adjust this now, so everyone can profit from this and so it won't be done later (after we add the device). because adjusting a ubi partition causes trouble and requires a factory install for example with nmrpflash (we need to increase major compat for this).
Gluon doesn't require the rest of the flash to be usable, but OpenWrt users profit.

blocktrron:
@Djfe  112M are plenty, so id o not see the need.

@maurerle Is https://github.com/openwrt/openwrt/pull/17694 requiring a backport or is it already being take care of?

Regarding the link flapping - can you check if adding the `eee-broken-...` properties (https://www.kernel.org/doc/Documentation/devicetree/bindings/net/ethernet-phy.yaml) fixes the initital link state?

maurerle:
@blocktrron  https://github.com/openwrt/openwrt/pull/17694 is already backported.

See https://github.com/openwrt/openwrt/commit/0ba00ec205a449fd341a7b094363048a8884da63

Regarding eeee-broken, I will try that soon, thanks

rotanid:
@maurerle are you still having the device at hand and considering to rebase/finalize/re-test it?

maurerle:
So the WAN port works fine, the LEDs work fine.
Thanks for merging! :)

PR #3435, opened 2025-01-26
gluon-core: lua: wireless: add support for PHYs named in board.json

neocturne:
On mediatek-mt7622, board.json contains a PHY -> path mapping (which confusingly also uses the `phy` UCI option). Look up the mapping and pass the path query into iwinfo, like OpenWrt's mac80211.sh does.

---

Supersedes https://github.com/freifunk-gluon/gluon/pull/3430

Untested on mt7622. Tested on ipq40xx to check that it doesn't break other platforms.

PR #3438, opened 2025-02-01
gluon-setup-mode: rename phys before starting setup mode

maurerle:
Works on WAX206

somehow the default_radio of openwrt is not deleted on the device..?
And I still had an issue for privatewifi, otherwise this worked on a monkeypatched wax206.

---

The `iwinfo.nl80211.phyname('path=' .. path)` behaves differently than the `iwinfo` cli command - so we use the iwinfo cli command to rename the phy as long as the networks are not up with netifd.

Outside of the setup-mode, netifd starts the interfaces correctly and does not lead to issues.
So we might want to have this patch until an upstream fix to apply the renaming before is created and backported to 24.10.

Just another attempt to fix this @neocturne @RolandoMagico @Djfe 

supersedes #3435 
supersedes #3430

### DIFF
--- a/package/gluon-setup-mode/files/lib/gluon/setup-mode/rc.d/S20network
+++ b/package/gluon-setup-mode/files/lib/gluon/setup-mode/rc.d/S20network
@@ -48,6 +48,8 @@ start_service() {
 	init_switch
 	iw reg set "$(lua -e 'print(require("gluon.site").regdom())')"
 
+    /usr/bin/lua -e 'require("gluon.setup-mode").rename_phys()'
+
 	procd_open_instance
 	procd_set_param command /sbin/netifd -c /var/gluon/setup-mode/config
 	procd_set_param respawn
--- a/package/gluon-setup-mode/luasrc/usr/lib/lua/gluon/setup-mode.lua
+++ b/package/gluon-setup-mode/luasrc/usr/lib/lua/gluon/setup-mode.lua
@@ -1,4 +1,5 @@
 local platform = require 'gluon.platform'
+local json = require 'jsonc'
 
 
 local M = {}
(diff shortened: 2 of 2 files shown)


RolandoMagico:
I am not a Openwrt or gluon expert but I want to understand why the problem exists and why it must be fixed in gluon.
Is the WiFi handling different among the Openwrt targets? Are there other targets with similar issues or is it only related to MT7622?

maurerle:
So me neither, but my take on an explanation is @RolandoMagico :
`mediatek-mt7622` has a board.json which contains phy mappings to rename phy0 to something else.
Why this is important for this target is not clear to me either, but I can accept this.

now OpenWrt renames the phys accordingly in `rename_board_phy_by_path`, which is called in `find_phy` which is called at the start of netifd for the interface:
https://github.com/openwrt/openwrt/blob/c1d5de0c59efd25ecb5922cb11e5d38b072cf993/package/network/config/wifi-scripts/files/lib/netifd/wireless/mac80211.sh#L619

So OpenWrt renames the phys on start of the radio correctly.

In Gluon we do not activate the radio yet in setup-mode and therefore still have the old interface names.
Though we need the correct interfaces names in setup mode to create the correct configuration.

There are two ways to fix this:

1. rename the phys ourselves when we are in config mode (this PR)
2. create an upstream fix to this issue, so that the renaming takes place regardless of starting the radio/netifd as mentioned in https://github.com/freifunk-gluon/gluon/pull/3435#issuecomment-2618096860
3. ~~using a fallback to `phy` instead of `wl`, though you can not guarantee the ordering is correct, this is therefore faulty~~ (#3430)

Note that no change is required in the normal mode, as the radio is then correctly renamed by netifd - just as in OpenWrt

neocturne:
@maurerle I do not believe that a workaround in Gluon is a proper solution - this should be fixed in OpenWrt. The same issue might also affect LuCI when the WLAN hasn't been brought up yet.

Here's my new solution: https://github.com/neocturne/openwrt/commit/121ef230d2134675349db3f9c9dc8e14bcc6e4e1 - still needs a proper commit message and more testing before I open a PR, but it appears to work correctly on my FB4040 when I modify board.d to set fixed PHY names.

RolandoMagico:
@neocturne : I tested your commit on my R32 (MT7622).
It fixes https://github.com/openwrt/openwrt/issues/14207 and WiFi can be set up in gluon config mode as well.

neocturne:
PR at https://github.com/openwrt/openwrt/pull/17821

PR #3446, opened 2025-03-05
mediatek-filogic: add support for OpenWRT One

ambassador86:
Specification:
 - MT7981 CPU using 2.4GHz and 5GHz WiFi (both AX)
 - 1GB RAM
 - 16MB NOR
 - 128MB NAND
 - 3 LEDs (red, green, blue, white)
 - 2 buttons (reset, user defined)
 - 1 2.5Gbit WAN port (Airoha EN8811h)
 - 1 1Gbit LAN ports
 - 1 single lane M.2 SSD slot
 - 1 mikroBus socket
 - externel HW WDT (25s refresh time)
 - i2c RTC (with battery backup)

Serial Interface
 - UBS-C CDC-ACM
 - 3 Pins GND, RX, TX
 - Settings: 115200, 8N1

Test Device: https://map.chemnitz.freifunk.net/#!v:m;n:2005b6ff13f0

  - [X] Web interface
  - [X] TFTP
  - [X] Other: see wiki at https://openwrt.org/toh/openwrt/one
        (`lua -e 'print(require("platform_info").get_image_name())'`) = openwrt-one
- Wired network
    - 2.5G WAN / 1G LAN 
- LED mapping
  - Power/system LED - green
  - Switch port LEDs
- Docs:

### DIFF
--- a/targets/mediatek-filogic
+++ b/targets/mediatek-filogic
@@ -1,3 +1,6 @@
+config('TARGET_ROOTFS_INITRAMFS', true)
+config('TARGET_INITRAMFS_COMPRESSION_LZMA', true)
+
 -- ASUS
 
 device('asus-tuf-ax4200', 'asus_tuf-ax4200', {
@@ -48,6 +51,13 @@ device('gl.inet-gl-mt3000', 'glinet_gl-mt3000', {
 
 device('netgear-wax220', 'netgear_wax220', {
 	factory_ext = '.img',
+})
+
+-- OpenWRT
+device('openwrt-one', 'openwrt_one',{
+	factory = '-factory',
+	factory_ext = '.ubi',
+	sysupgrade_ext = '.itb',
 })
 
 
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -334,6 +334,10 @@ mediatek-filogic
 
   - WAX220
 
+* OpenWrt
+
+  - One
+
 * Ubiquiti
 
   - UniFi 6 Plus


targets/mediatek-filogic
```diff
@@ -50,6 +53,21 @@ device('netgear-wax220', 'netgear_wax220', {
...
+	sysupgrade_ext = '.itb',
+	extra_images = {
+		{'-nor-bl31-uboot', '-nor-bl31-uboot', '.fip'},
+		{'-nor-factory', '-nor-factory', '.bin'},
+		{'-nor-preloader', '-nor-preloader', '.bin'},
+		{'-snand-bl31-uboot', '-snand-bl31-uboot', '.fip'},
+		{'-snand-factory', '-snand-factory', '.bin'},
+		{'-snand-preloader', '-dnand-preloader', '.bin'}
+	},
+})
```
blocktrron, line 69: I prefer we skip the bootloader images and refer for installation to the latest OpenWrt release to obtain images from there.

ambassador86: Hello @rotanid, should I just stick to the sysugrade image then and remove all other images?

AiyionPrime: I'd be fine with your current state.

targets/mediatek-filogic
```diff
@@ -50,6 +53,17 @@ device('netgear-wax220', 'netgear_wax220', {
 	factory_ext = '.img',
 })
 
+-- OpenWRT
+device('openwrt-one', 'openwrt_one',{
+	factory = '-factory',
+	factory_ext = '.ubi',
+	sysupgrade_ext = '.itb',
+	extra_images = {
+		{'-nor-factory', '-nor-factory', '.bin'},
```
AiyionPrime, line 62: @blocktrron, @ambassador86 Is the NOR intended to be used by us, or just to be a recovery means?

AiyionPrime (approved):
What an annoyance :D
Just borrowed an OpenWrt this week to get this in, only to be beaten by @ambassador86 :D

Anyway, the logs looked good on my end.
The performance lies around 333mbit/s with WireGuard and the MAC address matches.
https://hannover.freifunk.net/karte/#!/en/map/2005b6ff00bc


Thanks for your work :)

AiyionPrime:
Just gotta say, the USB C console port is one smooth experience I don't wanna miss on other devices anyore already.

targets/mediatek-filogic
blocktrron: NOR is for recovery, so I'd also suggest dropping it.

Background: NOT image is for reflashing NAND, which Gluon currently does not support. One could detect the boot-source in gluon and only offer reflashing from there, but I'd reserve this for later.

AiyionPrime:
I can rebuild this tonight, but'd be fine with the results and looking forward for some more experiences from deployments.
@blocktrron any further suggestions?
Or is this good to go in your opinion as well?

rotanid:
@ambassador86 thanks for the changes, your build seems to be failing though, please investigate

ambassador86:
Hello @rotanid
I already looked through the logs. The only thing i found so far: 'No space space left on device'
This sounds like a runner problem to me and is not related to the PR

AiyionPrime:
I can confirm, the build ran without issues for us yesterday on our machines.

rotanid:
i rebuilt again, anyone has an idea on how to deal/fix the problem in the CI ?

blocktrron:
GitHub provides runners with quite an opinionated package-set (See https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md )

Let me just push something so the CI removes these default packages and see if this fixes the build issue.

herbetom:
Is it possible that were just reached the point where the 21G of usable space on / of the runners isn't enough anymore?

herbetom:
What i find sometimes quite helpful for debugging is adding something like https://github.com/marketplace/actions/debugging-with-tmate to the workflow so that one can take a look at the state of the runner interactively. And also watch for example for disk space usage. But probably better to do that in the action run of someones local fork of this branch.

blocktrron:
@rotanid Test-build with this PR on top is here, as "allow changes by maintainers" was not checked for this PR: https://github.com/blocktrron/gluon/actions/runs/13743697613

Once this completed, feel free to merge #3448  before this and we should be set.

rotanid:
@blocktrron thanks

PR #3447, opened 2025-03-06
modules: update to latest Git HEAD

blocktrron:
Update the OpenWrt base for Gluon master.

Include LLVM ttolchain into docker container. Use Host LLVM toolchain from container to conserve build-time.

Remove mt76 patches. I did some testing and the issues once discovered did not emerge again. They may have been mitigated at one point upstream. First patch of the series is definitely not needed anymore, second patch may be required if problem re-emerges under specific conditions.

Supersedes #3441

PR #3436, opened 2025-01-28
mt76: fix system recovery routine for MT7915

blocktrron:
This is a draft-PR to fix the longstanding "Message timed out" errors with mt7915 based radios.

I've had positive experience with this patchset over the last 3 weeks and appreciate feedback from communities which frequently face this problem.

Sometimes, the node itself might crash when attempting recovery. I'm not entirely sure where this is happening, but it seems to be somewhere else in the mac80211 stack. The node comes back online by itself, so i still consider this an improvement over the current situation.

Log from a recovered node:

```
[1029591.814363] br-client: received packet on bat0 with own address as source address (addr:b8:ec:a3:e1:75:4f, vlan:0)
[1029592.855657] br-client: received packet on bat0 with own address as source address (addr:b8:ec:a3:e1:75:4f, vlan:0)
[1029592.856005] br-client: received packet on bat0 with own address as source address (addr:b8:ec:a3:e1:75:4f, vlan:0)
[1031792.254655] mt7915e 0000:02:00.0: Retry message 0000aded (seq 7)
[1031794.334830] mt7915e 0000:02:00.0: Message 0000aded (seq 7) timeout
[1031794.479827] mt7915e 0000:02:00.0: HW/SW Version: 0x8a108a10, Build Time: 20240429200716a
[1031794.479827] 
[1031794.495189] mt7915e 0000:02:00.0: WM Firmware Version: ____000000, Build Time: 20240429200752
[1031794.518127] mt7915e 0000:02:00.0: WA Firmware Version: DEV_000000, Build Time: 20240429200812
[1031800.085604] ieee80211 phy0: Hardware restart was requested
[1031800.085669] ieee80211 phy1: Hardware restart was requested
[1206189.188440] br-client: received packet on bat0 with own address as source address (addr:b8:ec:a3:e1:75:4f, vlan:0)
[1206189.188719] br-client: received packet on bat0 with own address as source address (addr:b8:ec:a3:e1:75:4f, vlan:0)
[1211397.565504] br-client: received packet on bat0 with own address as source address (addr:b8:ec:a3:e1:75:4f, vlan:0)
[1211397.565790] br-client: received packet on bat0 with own address as source address (addr:b8:ec:a3:e1:75:4f, vlan:0)
[1275209.524711] Ignoring NSS change in VHT Operating Mode Notification from 90:9c:4a:ba:e4:a4 with invalid nss 3
[1278619.260856] br-client: port 6(owe1) entered disabled state
[...]

PR #3449, opened 2025-03-09
patches: refresh-patches

herbetom:
After the removal of the mt76 patches in (#3447) the new mt76 patch (#3436) needs a lower patch file number.


rotanid:
thanks @herbetom

PR #3451, opened 2025-03-10
[v2023.2.x] modules: update to latest Git HEAD

herbetom:


herbetom:
Apparently the readthedocs failure is readthedocs beeing flaky. A simple `git commit --amend --no-edit` makes it work. :/

The failue doesn't seem related to the version beeing used. I have left the Sphinx Bump in anyway. We didn't have any problems with it on main as far as I remember, so it probably won't do any harm.

PR #3454, opened 2025-03-16
gluon-core: add hop penalty per hardif setting

### LINKED ISSUE #2754: batman-adv: per interface hop_penalty
As a result of #1942 `gluon.mesh_batman_adv.hop_penalty` was implemented. This allows setting the hop penealty for the meshif (the complete node).

Apparently it's also possible to set an hop penealty for the hardif. Those are the interfaces where bat0 is the master interface:

```
        meshif <netdev> hop_penalty|hp             [penalty]            display or modify hop_penalty setting
        hardif <netdev> hop_penalty|hp             [penalty]            display or modify hop_penalty setting
```

Currently the hop penealty for all of those hardifs is 0:

```
# batctl interface
mesh-vpn: active
primary0: active
mesh1: active
mesh0: active
vx_mesh_other: active
# batctl mesh-vpn hop_penalty
0
# batctl primary0 hop_penalty
0
# batctl mesh1 hop_penalty
0
# batctl mesh0 hop_penalty
0
# batctl vx_mesh_other hop_penalty
0
```

I haven't yet played with this but could imagine scenarios where it could be helpful to set an higher hop penealty on an interface.

maurerle:
set this using

    uci set gluon.batadv_hop_penalty=batadv_hop_penalty
    uci set gluon.batadv_hop_penalty.mesh_vpn='20'
    uci set gluon.batadv_hop_penalty.mesh_radio0='20'
    uci set gluon.batadv_hop_penalty.mesh_other='20'
    uci commit gluon

run

    gluon-reconfigure
    service network restart

and check using

    batctl hardif mesh0 hop_penalty

- [x] docs are currently missing
- [ ] also documentation of `gluon.mesh_batman_adv.hop_penalty` for the general hop_penalty is not available yet
- [x] requests for improvements on the usage of uci would be cool, but I do not really know, how to do this using a list without the fuzzy string replacement

closes #2754

### DIFF
--- a/docs/package/gluon-mesh-batman-adv.rst
+++ b/docs/package/gluon-mesh-batman-adv.rst
@@ -168,3 +168,32 @@ or to at least manually mark the port to the Gluon router as a
 Alternatively, the filtering of IGMP/MLD reports can be disabled via
 the site.conf (which is not recommended in large meshes though).
 See :ref:`site.conf mesh section <user-site-mesh>` for details.
+
+Tweaking Hop Penalty
+^^^^^^^^^^^^^^^^^^^^
+
+In general, the usage of a directly connected uplink is preferred over a faster more distant connection.
+In situations where the uplink of a device should only be used as a fallback (e.g. metered connection or slower encryption), this can be tweaked using a hop_penalty in gluon.
+Examples of this are shown below:
+
+.. code-block:: sh
+
+  # use mesh-vpn as fallback only
+  uci set gluon.mesh_vpn.batadv_hop_penalty='120'
+  # optional penalty for wired mesh
+  uci set gluon.iface_lan.batadv_hop_penalty=10
+  # optional for mesh on wan
+  uci set gluon.iface_wan.batadv_hop_penalty=10
+  # don't forget to commit changes
+  uci commit gluon
+
+To apply the changes, run the following commands:
+
+.. code-block:: sh
+
+  gluon-reconfigure
+  reboot
+
+Further documentation of the hop penalty can be found here:
+
+https://www.open-mesh.org/doc/batman-adv/Tweaking.html#hop-penalty
--- a/package/gluon-core/files/lib/netifd/proto/gluon_mesh.sh
+++ b/package/gluon-core/files/lib/netifd/proto/gluon_mesh.sh
@@ -8,14 +8,15 @@ init_proto "$@"
 
 proto_gluon_mesh_init_config() {
 	proto_config_add_boolean fixed_mtu
+	proto_config_add_int hop_penalty
 }
 
 proto_gluon_mesh_setup() {
 	export CONFIG="$1"
 	export IFNAME="$2"
 
-	local fixed_mtu
-	json_get_vars fixed_mtu
+	local fixed_mtu hop_penalty
+	json_get_vars fixed_mtu hop_penalty
 
 	export FIXED_MTU="${fixed_mtu:-0}"
 
@@ -27,6 +28,7 @@ proto_gluon_mesh_setup() {
 
 	proto_add_data
 	json_add_boolean fixed_mtu "$FIXED_MTU"
+	[ -n "${hop_penalty}" ] && json_add_int hop_penalty "${hop_penalty}"
 	[ "$IFNAME" != 'br-wan' ] && json_add_string zone 'mesh'
 	proto_close_data
 	proto_send_update "$CONFIG"
--- a/package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
+++ b/package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
@@ -49,8 +49,8 @@ proto_gluon_wired_setup() {
 
 	local meshif="$config"
 
-	local func vxlan vxpeer6addr
-	json_get_vars func vxlan vxpeer6addr
+	local hop_penalty func vxlan vxpeer6addr
+	json_get_vars hop_penalty func vxlan vxpeer6addr
 
 	# default args
 	[ -z "$vxlan" ] && vxlan=1
@@ -82,6 +82,7 @@ proto_gluon_wired_setup() {
 	json_add_string ifname "@${meshif}"
 	json_add_string proto 'gluon_mesh'
 	json_add_boolean fixed_mtu 1
+	[ -n "${hop_penalty}" ] && json_add_int hop_penalty "${hop_penalty}"
 	json_close_object
 	ubus call network add_dynamic "$(json_dump)"
 }
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/210-interface-mesh
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/210-interface-mesh
@@ -4,35 +4,52 @@ local site = require 'gluon.site'
 local uci = require('simple-uci').cursor()
 local util = require 'gluon.util'
 
-local mesh_interfaces = util.get_role_interfaces(uci, 'mesh')
-local uplink_interfaces = util.get_role_interfaces(uci, 'uplink')
+local mesh_interfaces = util.get_role_interfaces_with_options(uci, 'mesh', {'batadv_hop_penalty'})
+local uplink_interfaces = util.get_role_interfaces_with_options(uci, 'uplink', {})
 
 local mesh_interfaces_uplink = {}
 local mesh_interfaces_other = {}
-for _, iface in ipairs(mesh_interfaces) do
-	if util.contains(uplink_interfaces, iface) then
-		table.insert(mesh_interfaces_uplink, iface)
+for iface, options in pairs(mesh_interfaces) do
+	if uplink_interfaces[iface] ~= nil then
+		table.insert(mesh_interfaces_uplink, {iface = iface, options = options})
 	else
-		table.insert(mesh_interfaces_other, iface)
+		table.insert(mesh_interfaces_other, {iface = iface, options = options})
 	end
 end
 
+local function extract_ifaces(interfaces)
+	local ifnames = {}
+	local max_hop_penalty = 0
+	for _, iface in ipairs(interfaces) do
+		table.insert(ifnames, iface.iface)
+		local hop_penalty = tonumber(iface.options.batadv_hop_penalty) or 0
+		if max_hop_penalty < hop_penalty  then
+			max_hop_penalty = hop_penalty
+		end
+	end
+	table.sort(ifnames)
+	return max_hop_penalty, ifnames
+end
+
 if #mesh_interfaces_uplink > 0 then
+	local max_hop_penalty, _ = extract_ifaces(mesh_interfaces_uplink)
 	uci:section('network', 'interface', 'mesh_uplink', {
 		ifname = 'br-wan',
 		proto = 'gluon_wired',
 		func = 'wan',
+		hop_penalty = max_hop_penalty,
 		vxlan = site.mesh.vxlan(true),
 	})
 end
 
 if #mesh_interfaces_other > 0 then
+	local max_hop_penalty, ifnames = extract_ifaces(mesh_interfaces_other)
 	local iftype, ifname
-	if #mesh_interfaces_other == 1 then
-		ifname = mesh_interfaces_other[1]
+	if #ifnames == 1 then
+		ifname = ifnames[1]
 	else
 		iftype = 'bridge'
-		ifname = mesh_interfaces_other
+		ifname = ifnames
 
 		for _, iface in ipairs(ifname) do
 			uci:section('network', 'device', nil, {
@@ -49,6 +66,7 @@ if #mesh_interfaces_other > 0 then
 		igmp_snooping = false,
 		proto = 'gluon_wired',
 		func = 'mesh_other',
+		hop_penalty = max_hop_penalty,
 		vxlan = site.mesh.vxlan(true),
 	})
 end
--- a/package/gluon-mesh-batman-adv/files/lib/netifd/proto/gluon_bat0.sh
+++ b/package/gluon-mesh-batman-adv/files/lib/netifd/proto/gluon_bat0.sh
@@ -29,10 +29,16 @@ proto_gluon_bat0_renew() {
 
 	lock /var/lock/gluon_bat0.lock
 
-	ubus call network.interface dump | jsonfilter \
-		-e "@.interface[@.proto='gluon_mesh' && @.up=true].device" \
-	| xargs -r -n 1 batctl interface add
-
+	ubus call network.interface dump | \
+		jsonfilter -e "@.interface[@.proto='static' && @.up=true]['device', 'hop_penalty']" | \
+		while read -r device
+	do
+		batctl interface add "$device"	
+		read -r hop_penalty
+		if [ -n "$hop_penalty" ]; then
+			batctl hardif "$device" hop_penalty "$hop_penalty"
+		fi
+	done
 	lock -u /var/lock/gluon_bat0.lock
 }
 
--- a/package/gluon-mesh-batman-adv/files/lib/netifd/proto/gluon_bat0.sh
+++ b/package/gluon-mesh-batman-adv/files/lib/netifd/proto/gluon_bat0.sh
@@ -24,21 +24,34 @@ lookup_uci() {
 	uci -q get "$path" || echo "$default"
 }
 
+handle_mesh_interface() {
+	local proto up device hop_penalty
+
+	# Enter item in the array by index
+	json_select "$2"
+
+	json_get_vars proto up device
+	json_select "data"
+	json_get_var hop_penalty hop_penalty
+	json_select ".."
+
+	if [ "$proto" = "gluon_mesh" ] && [ "$up" = 1 ]; then
+		batctl interface add "$device"
+		batctl hardif "$device" hop_penalty "${hop_penalty:-0}"
+	fi
+
+	json_select ".."
+	# exit array item at the end of the loop
+}
+
 proto_gluon_bat0_renew() {
 	local config="$1"
 
 	lock /var/lock/gluon_bat0.lock
 
-	ubus call network.interface dump | \
-		jsonfilter -e "@.interface[@.proto='static' && @.up=true]['device', 'hop_penalty']" | \
-		while read -r device
-	do
-		batctl interface add "$device"	
-		read -r hop_penalty
-		if [ -n "$hop_penalty" ]; then
-			batctl hardif "$device" hop_penalty "$hop_penalty"
-		fi
-	done
+	json_load "$(ubus call network.interface dump)"
+
+	json_for_each_item handle_mesh_interface interface
 	lock -u /var/lock/gluon_bat0.lock
 }
 
--- a/package/gluon-mesh-vpn-core/luasrc/lib/gluon/upgrade/500-mesh-vpn
+++ b/package/gluon-mesh-vpn-core/luasrc/lib/gluon/upgrade/500-mesh-vpn
@@ -12,6 +12,7 @@ uci:section('network', 'interface', 'mesh_vpn', {
 	ifname = vpn_core.get_interface(),
 	proto = 'gluon_mesh',
 	fixed_mtu = true,
+	hop_penalty = uci:get('gluon', 'mesh_vpn', 'batadv_hop_penalty'),
 	macaddr = util.generate_mac_by_name('mesh_vpn'),
 	mtu = active_vpn.mtu(),
 })


rotanid:
comment by @neocturne on IRC:

> Hmm, our batadv setup is pretty asynchronous. May make more sense to add it to the renew handler in  gluon-mesh-batman-adv/files/lib/netifd/proto/gluon_bat0.sh

neocturne:
Overall, I think the following would make sense:

- Hop penalty is defined in the interface role sections in `/etc/config/gluon`
- Config scripts copy hop penalty to `gluon_wired` sections
- `gluon_wired` protocol forwards setting to `gluon_mesh`
- `gluon_mesh` protocol adds setting to protocol data
- `gluon_bat0` protocol is extended to get the info from the `ubus call network.interface dump` and apply it after `batctl interface add`

maurerle:
Thanks for the suggestions.

> Hop penalty is defined in the interface role sections in /etc/config/gluon

Looking at this output:

```console
root@ffac:~# uci show gluon
gluon.core=core
gluon.iface_single=interface
gluon.iface_single.name='/single'
gluon.iface_single.role='uplink' 'mesh'
gluon.wireless=wireless
gluon.wireless.outdoor='0'
gluon.wireless.preserve_channels='0'
gluon.wireless.outdoor_radio1_htmode='HE80'
gluon.mesh_vpn=mesh_vpn
gluon.mesh_vpn.enabled='1'
gluon.mesh_vpn.limit_enabled='0'
root@ffac:~# batctl if
br-wan: active
primary0: active
mesh-vpn: active
mesh0: active
```

I don't see an easy way to include this in /etc/config/gluon, as it is unclear what name the br-wan has and whether wireless mesh is currently active.
Furthermore, there does not exist `gluon_wired` for the wireless links at all, which only exist in /etc/config/wireless :/

Currently, there already is the rather unknown option `uci set gluon.mesh_batman_adv.hop_penalty='8'`. So I think it is fine to specifiy all hop_penalty settings in `gluon.mesh_batman_adv` as well..

To apply your suggestions, I definitely need some time and help.. :)

blocktrron:
@maurerle Did you consider using configuration structure such as:

```
config gluon_meshif_hop_penalty 'wan0'
    option 'network' 'lan' // Resolve ifname of interface section from /etc/config/networks
    option 'gluon_interface' '/wan' // Resolve ifname of /etc/config/gluon interface section
    option 'radio' // Resolve meshif of radio-idx
    option 'penalty' '20'
```

In all cases, a big fat warning that this functionality is not expected to preserve over upgrades. Writing this, I like the idea of moving this to a community-package if this functionality stays under this premise.

maurerle:
> this functionality is not expected to preserve over upgrades
why is that? You mean from a technical perspective or that gluon would not guarantee that?
Having an upgrade-safe way for this is basically the main idea of this PR..

This functionality likely requires adjustments to `gluon_bat0.sh` and therefore would not easily work as a community-package..?

neocturne:
Yea, I consider this a feature that should be implemented in `gluon_bat0.sh`, not in a separate package.

Regarding the config storage: For wired mesh, the hop penalty would make sense in the sections like `iface_single`.

For wireless, just putting it in the mesh iface sections in `/etc/config/wireless` would make sense for now - with my proposal above this would not require any additional work, as the attribute in `gluon_mesh` interfaces would already be handled. Moving this to `/etc/config/gluon` would definitely make sense once we start resetting `/etc/config/wireless` completely on reconfigure - which would be nice, but out of scope for this PR (and it would not change anything about the underlying implementation, `/etc/config/wireless` would just be populated based on `/etc/config/gluon` then.

maurerle:
So using `/lib/netifd/proto/gluon_bat0.sh` works fine without a sleep statement, though I though that my post-setup.d script would have been called afterwards anyway.

Now improving the uci setting flow would be nice.
I don't really know how  to do this properly, as it seems that I have to touch many files for it..

neocturne:
> So using `/lib/netifd/proto/gluon_bat0.sh` works fine without a sleep statement, though I though that my post-setup.d script would have been called afterwards anyway.

The name `post-setup.d` isn't very fitting in the case of batman-adv, because it will only be "post" sending the interface configuration to netifd, but the actual setup (`gluon_bat0.sh`) happens asynchronously.

T-X:
One more syntax idea came to my mind. I typically would want to use this to designate some interface as a backup. And in that sense it seems like a kind of (adjusted) role to me.

What would people think if you could add the penalty to a role? `gluon.iface_single.role='uplink:30'`, meaning the according interface is penalized by 70%?

One could also add aliases, where `gluon.iface_single.role='backuplink'`` were an alias for the uplink:30.

That way it would be super easy to add a penalty manually in UCI, without much room for typos.

(However we don't allow setting roles for WLAN interfaces in /etc/config/gluon yet. But I think it would make sense to do so, for consistency reasons, having all Gluon settings in one place and to rebuild a clean /etc/config/wireless.)

neocturne:
@T-X: That doesn't make much sense in my opinion - the only role that would ever have a penalty would be "mesh". We should not conflate "uplink" (meaning that the interface is added to `br-wan`) with the VPN connection running over that interface.

I also think overloading a string with more functions is in every way worse than just adding a separate UCI option to the section that just contains the single number - and that could be named something like `batadv_hop_penalty`, so we don't have to figure out how this should interact with other protocols right now.

(Hmm, using something similar to the roles to express which VIFs should be enabled on a radio/band is an interesting idea... I'm unsure if it wouldn't be too confusing because the effect would be so different from the roles for wired interfaces; the interaction with triband radio must also be considered - let's make a new ticket for that discussion)

maurerle:
I updated this PR according to the comment here: https://github.com/freifunk-gluon/gluon/pull/3454#issuecomment-2738237717

In `package/gluon-core/luasrc/lib/gluon/upgrade/210-interface-mesh` the interfaces are created through `util.get_role_interfaces(uci, 'mesh')` - this only provides a list of iface names per role.

1. the section for mesh-vpn is created at a different place
2. I don't feel like I should add my batman_hop_penalty to the very generic get_role_interfaces function

Todo:
* Hop penalty is defined in the interface role sections in `/etc/config/gluon`
* Config scripts copy hop penalty to `gluon_wired` sections

Done:
* `gluon_wired` protocol forwards setting to `gluon_mesh`
* `gluon_mesh` protocol adds setting to protocol data
* `gluon_bat0` protocol is extended to get the info from the `ubus call network.interface dump` and apply it after `batctl interface add`

This makes it possible to set the hop_penalty in network, but not yet on wireless, mesh-vpn or gluon roles.

package/gluon-core/luasrc/lib/gluon/upgrade/210-interface-mesh
```diff
@@ -49,6 +50,7 @@ if #mesh_interfaces_other > 0 then
 		igmp_snooping = false,
 		proto = 'gluon_wired',
 		index = 4,
+		hop_penalty = 0,
```
neocturne: All values set in this section are overwritten on every upgrade/reconfigure. Let's not set the value in the section at all by default.

package/gluon-mesh-batman-adv/files/lib/netifd/proto/gluon_bat0.sh
```diff
@@ -29,10 +29,14 @@ proto_gluon_bat0_renew() {
 
 	lock /var/lock/gluon_bat0.lock
 
-	ubus call network.interface dump | jsonfilter \
-		-e "@.interface[@.proto='gluon_mesh' && @.up=true].device" \
-	| xargs -r -n 1 batctl interface add
-
+	for device in $(ubus call network.interface dump | jsonfilter -e "@.interface[@.proto='gluon_mesh' && @.up=true].device"); do
```
neocturne: If we can rely on every `gluon_mesh` interface having a `hop_penalty` set (and I believe we can, because `gluon_mesh.sh` always sets it), this can be handled much more efficiently with a single ubus and jsonfilter:
```sh
ubus call network.interface dump | \
    jsonfilter -e "@.interface[@.proto='static' && @.up=true]['device', 'hop_penalty']" | \
    while read device
do
    read hop_penalty
    ...
done
```

package/gluon-mesh-batman-adv/files/lib/netifd/proto/gluon_bat0.sh
```diff
@@ -29,10 +29,14 @@ proto_gluon_bat0_renew() {
...
+		batctl interface add "$device"
+		local device_sanitized="$(echo "$device" | sed 's/[^a-zA-Z0-9]/_/g')"
+		local hop_penalty="$(ubus call network.interface dump | jsonfilter -e "@.interface[@.device='$device'].data.hop_penalty")"
+		if [ -n "$hop_penalty" ] && [ "$hop_penalty" -ne 0 ]; then
+			batctl hardif "$device" hop_penalty "$hop_penalty"
```
neocturne: This should set the hop penalty unconditionally, so changing it back from non-zero to zero will work (not really relevant to Gluon, as we will never change the state of a configured interface, but simpler and more correct)

neocturne (changes requested):

maurerle:
Adding the config to the "interface role sections" is somewhat wrong in the concept as the following example shows:

```
config interface 'iface_lan'
	option name '/lan'
	list role 'mesh'
        hop_penalty 17

config interface 'iface_vlan17'
	option name 'lan2.17'
	list role 'mesh'
        hop_penalty 20
```

would both affect `br-mesh-other` and conflict each other.

So I think @neocturne  suggested the following in [this comment](https://github.com/freifunk-gluon/gluon/pull/3454#issuecomment-2815081165): 

```
gluon.batadv_hop_penalty=batadv_hop_penalty
gluon.batadv_hop_penalty.mesh_other=17
gluon.batadv_hop_penalty.mesh_vpn=20
gluon.batadv_hop_penalty.wan=23
gluon.batadv_hop_penalty.mesh_radio0=23
```

This is now implemented.
I did not test this yet though.

package/gluon-mesh-batman-adv/files/lib/netifd/proto/gluon_bat0.sh
maurerle: addressed

package/gluon-mesh-batman-adv/files/lib/netifd/proto/gluon_bat0.sh
maurerle: sounds good - I did it that way

package/gluon-core/luasrc/lib/gluon/upgrade/210-interface-mesh
maurerle: This is now correctly read from gluon config using:

`hop_penalty = uci:get('gluon', 'batadv_hop_penalty', name) or 0`

maurerle:
So I tested this on two different devices, it works fine for different kind of meshs.
I checked that the value is set through `batctl hardif mesh-vpn hop_penalty` after configuring the values as documented.
I also checked that the value is set back to 0 if the value is removed from the gluon section again.

It might be good to also migrate the less known undocumented node-wide `gluon.mesh_batman_adv.hop_penalty` call to something like `gluon.batadv_hop_penalty.all='20'` but I am undecided if this should be "all" or something else.

So far this is ready from my side.
Thanks @neocturne , I learned a lot about the proto interface and config of gluon in this PR.

package/gluon-mesh-batman-adv/files/lib/netifd/proto/gluon_bat0.sh
```diff
@@ -29,10 +29,19 @@ proto_gluon_bat0_renew() {
...
-	| xargs -r -n 1 batctl interface add
-
+	ubus call network.interface dump | \
+		jsonfilter -e "@.interface[@.proto='gluon_mesh' && @.up=true].device" | \
+		while read -r device
+	do
+		batctl interface add "$device"	
+		
+		hop_penalty="$(ubus call network.interface dump | \
+			jsonfilter -e "@.interface[@.device='$device'].data.hop_penalty")"
```
neocturne: Ah, my previously suggested command won't work because there's the extra level `data` above `hop_penalty`. I'll need to try a few things to figure out how to get rid of that.

We definitely want to avoid running the dump for each loop iteration. Optimally, we would also avoid running jsonfilter for each loop iteration.

neocturne:
Hmm, I liked the previous structure with the hop penalties in the interface sections better - just calling the field `batadv_hop_penalty`:
```
config interface 'iface_lan'
	option name '/lan'
	list role 'mesh'
        batadv_hop_penalty 17

config interface 'iface_vlan17'
	option name 'lan2.17'
	list role 'mesh'
        batadv_hop_penalty 20
```

package/gluon-mesh-vpn-core/luasrc/lib/gluon/upgrade/500-mesh-vpn
```diff
@@ -12,6 +12,7 @@ uci:section('network', 'interface', 'mesh_vpn', {
 	ifname = vpn_core.get_interface(),
 	proto = 'gluon_mesh',
 	fixed_mtu = true,
+	hop_penalty = uci:get('gluon', 'batadv_hop_penalty', 'mesh_vpn') or 0,
```
neocturne: Why is the `or 0` here necessary? Can we not leave the field unset and handle the missing value in `gluon_mesh.sh`? Ideally, there should be only one piece of code that sets the default to 0, and all other places just forward the (set or unset) value the the next layer.

package/gluon-mesh-batman-adv/files/lib/netifd/proto/gluon_bat0.sh
neocturne: Hmm, looks like jsonfilter can't do what I want. `json_load` + `json_for_each_item` may be the way to go...

maurerle:
@neocturne 
Yes i would have liked that better as well, but see
https://github.com/freifunk-gluon/gluon/pull/3454#issuecomment-3026658832
Why i did not do it.
There is only one bridge for different interfaces.
 If you have a solution for me, i am all ears

package/gluon-mesh-vpn-core/luasrc/lib/gluon/upgrade/500-mesh-vpn
maurerle: I need to double check this, but will put my testing efforts into triband at first.

neocturne:
@maurerle 

> @neocturne Yes i would have liked that better as well, but see [#3454 (comment)](https://github.com/freifunk-gluon/gluon/pull/3454#issuecomment-3026658832) Why i did not do it. There is only one bridge for different interfaces. If you have a solution for me, i am all ears

For the bridge situation, there isn't really a good solution - we'll need to pick the minimum or maximum hop penalty of the member interfaces (I would go with maximum - for the simple reason that an interface with hop penalty unset and set to 0 would always behave the same then). I'd like to see `br-mesh_other` to disappear in the future, solving that issue (edit: and I see `br-mesh_other` as an implementation detail that should not be reflected in our config structure). While br-wan will still exist, at least having multiple WAN interfaces is rare. Maybe an ebtables/nftables-based "brouting" approach could be considered in the future.

For getting the value from the interface sections, this is my proposal: https://github.com/freifunk-gluon/gluon/pull/3551

maurerle:
@neocturne I did use the new function for wired mesh, which has this side effect, that with mesh and uplink on the WAN port:

```
gluon.core=core
gluon.iface_wan=interface
gluon.iface_wan.name='/wan'
gluon.iface_wan.role='mesh' 'uplink'
gluon.iface_wan.batadv_hop_penalty='7'
```

we can not decide to have a different hop penalty for mesh vs lan mesh.

```
network.mesh_radio0.hop_penalty='0'
network.mesh_radio1.hop_penalty='0'
network.mesh_uplink.hop_penalty='7'
network.mesh_other.hop_penalty='0'
network.mesh_vpn.hop_penalty='7'
```

I think that this is fine/manageable, but is not ideal as well.

Other than that, we need to have a draft for #3550 for the wireless section right?

neocturne:
> @neocturne I did use the new function for wired mesh, which has this side effect, that with mesh and uplink on the WAN port:
> 
> ```
> gluon.core=core
> gluon.iface_wan=interface
> gluon.iface_wan.name='/wan'
> gluon.iface_wan.role='mesh' 'uplink'
> gluon.iface_wan.batadv_hop_penalty='7'
> ```
> 
> we can not decide to have a different hop penalty for mesh vs lan mesh.

@maurerle I would not expect this config to have an effect on the mesh VPN. There is already a `mesh_vpn` section in `/etc/config/gluon`, let's add `batadv_hop_penalty` there (but that can also happen in a follow-up PR)

maurerle:
@neocturne right, thanks!
So we do now have the option to set the hop penalty for all wired meshs through the gluon section.

If everything is fine, I will squash my latest two commits.
Then this is ready :)

batadv_hop_penalty for wireless mesh will be done in a follow-up PR for #3550

AiyionPrime:
I'd suggest marking resolved conversations as such, but I like the idea :)

package/gluon-core/files/lib/netifd/proto/gluon_mesh.sh
```diff
@@ -8,16 +8,18 @@ init_proto "$@"
...
 	export CONFIG="$1"
 	export IFNAME="$2"
 
-	local fixed_mtu
-	json_get_vars fixed_mtu
+	local fixed_mtu hop_penalty
+	json_get_vars fixed_mtu hop_penalty
 
 	export FIXED_MTU="${fixed_mtu:-0}"
+	export HOP_PENALTY="${hop_penalty:-0}"
```
neocturne: We aren't using the hop penalty in any `setup.d` scripts, so this line can be omitted.

package/gluon-core/files/lib/netifd/proto/gluon_mesh.sh
```diff
@@ -27,6 +29,7 @@ proto_gluon_mesh_setup() {
 
 	proto_add_data
 	json_add_boolean fixed_mtu "$FIXED_MTU"
+	json_add_int hop_penalty "$HOP_PENALTY"
```
neocturne: As mentioned before, I'd prefer to have one single place of code that sets the default value "0" - all other places should just forward the value, or the non-existance of a value:
```suggestion
	[ -n "${hop_penalty}" ] && json_add_int hop_penalty "${hop_penalty}"
```

package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
```diff
@@ -82,6 +83,7 @@ proto_gluon_wired_setup() {
 	json_add_string ifname "@${meshif}"
 	json_add_string proto 'gluon_mesh'
 	json_add_boolean fixed_mtu 1
+	json_add_int hop_penalty "${hop_penalty}"
```
neocturne:
```suggestion
	[ -n "${hop_penalty}" ] && json_add_int hop_penalty "${hop_penalty}"
```

package/gluon-mesh-batman-adv/files/lib/netifd/proto/gluon_bat0.sh
```diff
@@ -24,15 +24,37 @@ lookup_uci() {
...
+	json_select "data"
+	json_get_var hop_penalty hop_penalty
+	json_select ".."
+
+	if [ "$proto" = "gluon_mesh" ] && [ "$up" = 1 ]; then
+		batctl interface add "$device"
+
+		if [ -n "$hop_penalty" ]; then
+			batctl hardif "$device" hop_penalty "$hop_penalty"
+		fi
```
neocturne:
```suggestion
		batctl hardif "$device" hop_penalty "${hop_penalty:-0}"
```

docs/package/gluon-mesh-batman-adv.rst
```diff
@@ -168,3 +168,30 @@ or to at least manually mark the port to the Gluon router as a
 Alternatively, the filtering of IGMP/MLD reports can be disabled via
 the site.conf (which is not recommended in large meshes though).
 See :ref:`site.conf mesh section <user-site-mesh>` for details.
+
+Tweaking Hop Penalty
+^^^^^^^^^^^^^^^^^^^^
+
+In general, the usage of a directly connected uplink is preferred over a faster more distant connection.
+In situations where the uplink of a device should only be used as a fallback (e.g. metered connection or slower encryption), this can be tweaked using a hop_penalty in gluon.
+Examples of this are shown below
```
neocturne:
```suggestion
Examples of this are shown below:
```

docs/package/gluon-mesh-batman-adv.rst
```diff
@@ -168,3 +168,30 @@ or to at least manually mark the port to the Gluon router as a
...
+
+
```
neocturne: .. code-block:: sh

docs/package/gluon-mesh-batman-adv.rst
```diff
@@ -168,3 +168,30 @@ or to at least manually mark the port to the Gluon router as a
...
+  uci set gluon.mesh_vpn.batadv_hop_penalty='120'
+  # optional penalty for wired mesh
+  uci set gluon.iface_lan.batadv_hop_penalty=10
+  # optional for mesh on wan
+  uci set gluon.iface_wan.batadv_hop_penalty=10
+  # don't forget to commit changes
+  uci commit gluon
+
+To apply the changes one should running
+
```
neocturne, line 191: .. code-block:: sh

docs/package/gluon-mesh-batman-adv.rst
```diff
@@ -168,3 +168,30 @@ or to at least manually mark the port to the Gluon router as a
...
+  gluon-reconfigure
+  reboot
+
+Further documentation of the hop penalty can be found here:
+
+* https://www.open-mesh.org/doc/batman-adv/Tweaking.html#hop-penalty
```
neocturne: Let's remove the `*`, as this is not a list.

docs/package/gluon-mesh-batman-adv.rst
```diff
@@ -168,3 +168,30 @@ or to at least manually mark the port to the Gluon router as a
...
+  # use mesh-vpn as fallback only
+  uci set gluon.mesh_vpn.batadv_hop_penalty='120'
+  # optional penalty for wired mesh
+  uci set gluon.iface_lan.batadv_hop_penalty=10
+  # optional for mesh on wan
+  uci set gluon.iface_wan.batadv_hop_penalty=10
+  # don't forget to commit changes
+  uci commit gluon
+
+To apply the changes one should running
```
neocturne:
```suggestion
To apply the changes, run the following commands:
```
(or maybe just combine everything into one code block?)

package/gluon-core/luasrc/lib/gluon/upgrade/210-interface-mesh
```diff
@@ -4,35 +4,52 @@ local site = require 'gluon.site'
 local uci = require('simple-uci').cursor()
 local util = require 'gluon.util'
 
-local mesh_interfaces = util.get_role_interfaces(uci, 'mesh')
-local uplink_interfaces = util.get_role_interfaces(uci, 'uplink')
+local mesh_interfaces = util.get_role_interfaces_with_options(uci, 'mesh', {'batadv_hop_penalty'})
+local uplink_interfaces = util.get_role_interfaces_with_options(uci, 'uplink', {'batadv_hop_penalty'})
```
neocturne: The extra option of `uplink_interfaces` is unused:
```suggestion
local uplink_interfaces = util.get_role_interfaces_with_options(uci, 'uplink', {})
```

package/gluon-core/luasrc/lib/gluon/upgrade/210-interface-mesh
```diff
@@ -4,35 +4,52 @@ local site = require 'gluon.site'
...
 
 local mesh_interfaces_uplink = {}
 local mesh_interfaces_other = {}
-for _, iface in ipairs(mesh_interfaces) do
-	if util.contains(uplink_interfaces, iface) then
-		table.insert(mesh_interfaces_uplink, iface)
+for iface, options in pairs(mesh_interfaces) do
+	if uplink_interfaces[iface] ~= nil then
+		mesh_interfaces_uplink[iface] = options
```
neocturne: Let's keep this an integer-indexed table, so we can sort this by iface name (`pairs` iterates in random order, making generation nondeterministic)
```suggestion
		table.insert(mesh_interfaces_uplink, {iface = iface, options = options})
```

package/gluon-core/luasrc/lib/gluon/upgrade/210-interface-mesh
```diff
@@ -4,35 +4,52 @@ local site = require 'gluon.site'
...
 	else
-		table.insert(mesh_interfaces_other, iface)
+		mesh_interfaces_other[iface] = options
```
neocturne: Let's keep this an integer-indexed table, so we can sort this by iface name (`pairs` iterates in random order, making generation nondeterministic)
```suggestion
		table.insert(mesh_interfaces_other, {iface = iface, options = options})
```

package/gluon-core/files/lib/netifd/proto/gluon_mesh.sh
maurerle: Should this be `json_add_int hop_penalty "${hop_penalty:-0}"` then below?
I am not sure what happens with `json_add_int` for empty values?

maurerle: Well, you said it below. Thanks.

`[ -n "${hop_penalty}" ] && json_add_int hop_penalty "${hop_penalty}"`

package/gluon-core/luasrc/lib/gluon/upgrade/210-interface-mesh
maurerle: Or should I use `get_role_interfaces` here and check with contains again (like it was before)?

neocturne: `get_role_interfaces_with_options` is more efficient (less iteration, and `get_role_interfaces` uses `get_role_interfaces_with_options` internally anyways.

package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
neocturne: You applied this change only in `gluon_mesh.sh` and forgot `gluon_wired.sh`.

package/gluon-core/luasrc/lib/gluon/upgrade/210-interface-mesh
```diff
@@ -4,35 +4,51 @@ local site = require 'gluon.site'
...
-	if util.contains(uplink_interfaces, iface) then
-		table.insert(mesh_interfaces_uplink, iface)
+for iface, options in pairs(mesh_interfaces) do
+	if uplink_interfaces[iface] ~= nil then
+		table.insert(mesh_interfaces_uplink, {iface = iface, options = options})
 	else
-		table.insert(mesh_interfaces_other, iface)
+		table.insert(mesh_interfaces_other, {iface = iface, options = options})
 	end
 end
```
neocturne, line 18: As `pairs(mesh_interfaces)` iterates in arbitrary order, you need to sort the tables after the loop:
```lua
local function sort_ifaces(ifaces)
	table.sort(ifaces, function(a, b)
		return a.iface < b.iface
	end)
end
sort_ifaces(mesh_interfaces_uplink)
sort_ifaces(mesh_interfaces_other)
```
(untested)

neocturne: Oh wait, you don't actually need a custom comparator - you can just sort `table.sort(ifnames)` before returning from `extract_ifaces`, then it doesn't matter that `mesh_interfaces_uplink` and `mesh_interfaces_other` are unsorted.

package/gluon-core/luasrc/lib/gluon/upgrade/210-interface-mesh
```diff
@@ -4,35 +4,51 @@ local site = require 'gluon.site'
...
 
+local function extract_ifaces(interfaces)
+	local ifnames = {}
+	local max_hop_penalty = 0
+	for _, iface in ipairs(interfaces) do
+		table.insert(ifnames, iface['iface'])
+		local options = iface['options']
+		if options['batadv_hop_penalty'] and tonumber(options['batadv_hop_penalty']) > max_hop_penalty then
+			max_hop_penalty = options['batadv_hop_penalty']
+		end
```
neocturne: You only `tonumber()` the hop penalty for the comparison, but the value assigned to `max_hop_penalty` must be converted as well. I'd write the code like this:

```suggestion
		table.insert(ifnames, iface.iface)
		let hop_penalty = tonumber(iface.options.batadv_hop_penalty) or 0
		if max_hop_penalty < hop_penalty  then
			max_hop_penalty = hop_penalty
		end
```
(untested)

package/gluon-core/luasrc/lib/gluon/upgrade/210-interface-mesh
```diff
@@ -4,35 +4,51 @@ local site = require 'gluon.site'
...
 		func = 'wan',
+		hop_penalty = max_hop_penalty,
 		vxlan = site.mesh.vxlan(true),
 	})
 end
 
 if #mesh_interfaces_other > 0 then
+	local max_hop_penalty, ifnames = extract_ifaces(mesh_interfaces_other)
 	local iftype, ifname
 	if #mesh_interfaces_other == 1 then
```
neocturne: Maybe also change the condition to match the change of the lines below:
```suggestion
	if #ifnames == 1 then
```

package/gluon-core/files/lib/netifd/proto/gluon_wired.sh
```diff
@@ -49,10 +49,11 @@ proto_gluon_wired_setup() {
 
 	local meshif="$config"
 
-	local func vxlan vxpeer6addr
-	json_get_vars func vxlan vxpeer6addr
+	local hop_penalty func vxlan vxpeer6addr
+	json_get_vars hop_penalty func vxlan vxpeer6addr
 
 	# default args
+	[ -z "$hop_penalty" ] && hop_penalty=0
```
neocturne: The default is applied in `gluon_bat0.sh`, it is not needed here.

package/gluon-core/luasrc/lib/gluon/upgrade/210-interface-mesh
```diff
@@ -4,35 +4,51 @@ local site = require 'gluon.site'
...
+		table.insert(mesh_interfaces_other, {iface = iface, options = options})
 	end
 end
 
+local function extract_ifaces(interfaces)
+	local ifnames = {}
+	local max_hop_penalty = 0
+	for _, iface in ipairs(interfaces) do
+		table.insert(ifnames, iface.iface)
+		local hop_penalty = tonumber(iface.options.batadv_hop_penalty)
```
neocturne:
```suggestion
		local hop_penalty = tonumber(iface.options.batadv_hop_penalty) or 0
```
(otherwise the following comparison will crash for `nil`)

package/gluon-core/luasrc/lib/gluon/upgrade/210-interface-mesh
```diff
@@ -4,35 +4,51 @@ local site = require 'gluon.site'
...
+		if max_hop_penalty < hop_penalty  then
+			max_hop_penalty = hop_penalty
+		end
+	end
+	return max_hop_penalty, ifnames
```
neocturne, line 31: Insert before the return for determinism:
```lua
table.sort(ifnames)
```

maurerle: I had this, but lost it somehow. Sorry.
I hope this is finally finished now :)

neocturne (approved):

PR #3460, opened 2025-03-27

herbetom:


herbetom:
Seems like the Read the Docs action just doesn't want to work right now because it has some issues with cloning. As this PR doesn't affect docs i'll leave it at that.

neocturne:
Retriggering fixed the RTD build.

PR #2707, opened 2022-11-26
realtek-rtl838x: add support for D-Link DGS-1210-10P

AiyionPrime:
> as well as the target itself

- [x] Must be flashable from vendor firmware
  - [x] Web interface (instructions are somewhat complex, not recommended)
  - [x] TFTP
  - ~~Other: <specify>~~
- [x] Must support upgrade mechanism
  - [x] Must have working sysupgrade
    - [x] Must keep/forget configuration (`sysupgrade [-n]`, `firstboot`)
  - [x] Gluon profile name matches autoupdater image name
        (`lua -e 'print(require("platform_info").get_image_name())'`)
- [x] Reset/WPS/... button must return device into config mode
- [x] Primary MAC address should match address on device label (or packaging)
      (https://gluon.readthedocs.io/en/latest/dev/hardware.html#hardware-support-in-packages)
  - When re-adding a device that was supported by an earlier version of Gluon, a
    factory reset must be performed before checking the primary MAC address, as
    the setting from the old version is not reset otherwise.
- Wired network
  - [x] should support all network ports on the device (ports nine and ten *sfp* reportedly work in OpenWrt, but not tested)
  - [x] must have correct port assignment (WAN/LAN)
    - On devices supplied via PoE, there is usually no explicit WAN/LAN labeling on the hardware.
      The PoE input should be the WAN port in this case.
- Wireless network (if applicable)
  - ~~Association with AP must be possible on all radios~~
  - ~~Association with 802.11s mesh must work on all radios~~ 
  - ~~AP+mesh mode must work in parallel on all radios~~
- LED mapping
  - Power/system LED
    - [x] Lit while the device is on
    - [x] Should display config mode blink sequence 
          (https://gluon.readthedocs.io/en/latest/features/configmode.html)
  - Radio LEDs
    - ~~Should map to their respective radio~~
    - ~~Should show activity~~
  - Switch port LEDs
    - [x] Should map to their respective port (or switch, if only one led present) 
    - [x] Should show link state and activity
- Outdoor devices only:
  - ~~Added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`~~
- Docs:
  - [x] Added Device to `docs/user/supported_devices.rst`

PR #3461, opened 2025-03-28
hardware: remove support for D-Link DGS-1210-10P and realtek-rtl838x target

### LINKED ISSUE #3459: Verify D-Link DGS-1210-10P device name
Is the device name `d-link_dgs-1210-10p` correct (matches `lua -e 'print(require("platform_info").get_image_name())'`)? For most targets it should be impossible for an underscore to appear because of the way the name is derived from the model name.

_Originally posted by @neocturne in https://github.com/freifunk-gluon/gluon/pull/2707#discussion_r2010814717_

neocturne:
This device should never have been supported as non-broken. It is not a useful target for batman-adv (as the CPU is too slow for reasonable throughput; it may be more useful with L3 routing protocols); also, the autoupdater never worked because of an incorrect image name.

As there it nobody actually running this device (or planning to do so at the moment), just drop the device instead of fixing the name. The realtek-rtl838x target becomes empty and is dropped.

Fixes #3459


AiyionPrime (approved):

maurerle:
Just adding a link to the initial PR, which added broken support for D-Link DGS-1210-10P: #2707

PR #3469, opened 2025-04-03
gluon-setup-mode: namespace-based implementation for password-less login

neocturne:
Start dropbear in a mount namespace with a readonly overlay mount
replacing passwd and shadow to allow password-less login. This allows us
to drop our custom -f flag.

A login shell wrapper is introduced to move back into the init namespace
for the session, so the overlay mount is invisible to the user and
passwd modifies the right shadow file.


neocturne:
I have pushed a few minor fixes to override `$SHELL` and use a login shell again.

Djfe:
awesome, one less downstream patch is always great 💪

neocturne:
Will need to be rebased after #3471 (or vice-versa, whichever is merged first)

Djfe:
Just a side-note for v2025.2 or some future release:
Dropbear 2025.87 apparently requires permissions of directory `/etc` to be set to 755 (for whatever reason)

`/lib/gluon/setup-mode/dropbear` might require those permissions, too in that case.

If you agree with the change you could give this directory those permissions as well on your initial PR (here) to prevent a future regression in config mode.
Because this way it won't be forgotten before the next release.

schuettecarsten:
> Just a side-note for v2025.2 or some future release: Dropbear 2025.87 apparently requires permissions of directory `/etc` to be set to 755 (for whatever reason) [openwrt/openwrt#18664](https://github.com/openwrt/openwrt/pull/18664)

The reason is: https://github.com/openwrt/openwrt/pull/18606#issuecomment-2843311993

package/gluon-setup-mode/files/lib/gluon/setup-mode/dropbear/etc/shadow
AiyionPrime, line 1: Is this a busybox enforced quirk?

root::**0**:0:99999:7:::

this zero has a special meaning: "the user should change her password the next time she will log in the system"

An empty field on the other hand would express password aging was disabled, which might allow cleaner values for the fields maximum and warn (`99999:7`) as well.

Possible that you already checked that :)

package/gluon-setup-mode/files/lib/gluon/setup-mode/dropbear/ash
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+export SHELL=/bin/ash
+exec nsenter --target=1 --mount --wd="$PWD" "$SHELL" --login "$@"
```
AiyionPrime, line 4: So one enters the router via ssh, meets dropbear in its own namespace, which has and `/etc` with no password configured and as a login shell, we enter the context of PID 1 (our regular `procd`) and as in its context the setupmode `/etc` overlay is not mounted, one can see the normal `/etc` again?

package/gluon-setup-mode/files/lib/gluon/setup-mode/rc.d/S50dropbear
```diff
@@ -2,22 +2,14 @@
...
 
 start_service() {
-	[ -x /etc/init.d/dropbear ] || return 0
-
-	. /etc/init.d/dropbear
-
 	hk_generate_as_needed
 
-	. /lib/functions.sh
-	. /lib/functions/network.sh
```
AiyionPrime, line 17: These two are unrelated unused omissions, right?

package/gluon-setup-mode/files/lib/gluon/setup-mode/ash-login
```diff
@@ -1,3 +0,0 @@
-#!/bin/sh
```
AiyionPrime, line 1: When would on have met this file before?
Only upon SSH access or via console login as well?

targets/generic
```diff
@@ -81,6 +81,10 @@ try_config('OONF_GENERIC_HTTP', true)
 
 config('COLLECT_KERNEL_DEBUG', true)
 
+config('BUSYBOX_CUSTOM', true)
+config('BUSYBOX_CONFIG_NSENTER', true)
```
AiyionPrime, line 89: If these were for some reason not configurable options for some target, this would not fail silently, right?

package/gluon-setup-mode/files/lib/gluon/setup-mode/dropbear/etc/shadow
neocturne: Hmm, I guess it doesn't really matter and we could just make the value empty.

package/gluon-setup-mode/files/lib/gluon/setup-mode/dropbear/ash
neocturne: Correct.

package/gluon-setup-mode/files/lib/gluon/setup-mode/ash-login
neocturne: It was used in the past for telnet, and we forgot to drop it when switching to dropbear.

targets/generic
neocturne: `config` will fail configuration if a symbol is not available, `try_config` is for settings that are allowed to differ.

package/gluon-setup-mode/files/lib/gluon/setup-mode/rc.d/S50dropbear
neocturne: This was copied from the original initscript, but it is not actually needed.

package/gluon-setup-mode/files/lib/gluon/setup-mode/dropbear/etc/shadow
```diff
@@ -0,0 +1 @@
+root::0:0:99999:7:::
```
AiyionPrime:
```suggestion
root::::::::
```

Just tested it, omitting them all is fine on an ipq40xx-chromium device.

AiyionPrime (approved):
The PR works like a charm. Thanks for your work :)

package/gluon-setup-mode/files/lib/gluon/setup-mode/dropbear/etc/shadow
AiyionPrime: tested as https://github.com/freifunk-gluon/gluon/pull/3469#discussion_r2216943477

AiyionPrime (approved):
@neocturne I can test this again, if you did not do already.
But codewise this still looks fine.

neocturne:
@AiyionPrime I just applied your suggestion and refreshed the patches (1 rename without changes), so I don't think another test will be necessary.

neocturne:
Rebased again because of changes in the `patches` directory.

AiyionPrime:
I hope I'm not overstepping, but it might take a while until @blocktrron would get to this again.

PR #3470, opened 2025-04-03
perl: drop patch to disable parallel build

neocturne:
The upstream issue has been resolved:
https://github.com/openwrt/packages/issues/8238


rotanid:
@neocturne isn't https://github.com/openwrt/packages/pull/23602 more relevant? that one got fixed in 24.10.x

neocturne:
> @neocturne isn't [openwrt/packages#23602](https://github.com/openwrt/packages/pull/23602) more relevant? that one got fixed in 24.10.x

That is precisely the PR that fixed the issue I linked.

PR #3477, opened 2025-04-10
gluon-ebtables-filter-multicast: block packets with Gluon VXLAN multicast destination

### LINKED ISSUE #3025: Prevent VXLAN packages from Wired mesh from entering BATMAN
### Bug report

**What is the problem?**

Due to miscabling/misconfiguration, it is possible for packets coming from the mesh to be encapsulated in VXLAN packets, sent out of a wired mesh port, arriving at a LAN port of another node and then and being fed back into the mesh with the VXLAN encapsulation as multicast.

A node receiving VXLAN packets with the correct VNI and destination address for its domain does not forward them into the mesh, even if received on a port not configured for wired mesh.

**Gluon Version:**

v2021.1.2 (but probably later as well?)

**Site Configuration:**

https://github.com/FreifunkBremen/gluon-site-ffhb

**Custom patches:**

neocturne:
An incorrectly connected node may send wired-mesh packets to another node's client port. For raw batman-adv, this is filtered by batman-adv, but this does not work for VXLAN-encapsulated frames.

While such packets are not flooded through the mesh because of the batman-adv noflood flag, they may still reach other node that is connected in the same incorrect way, which could result in loops and other connectivity issues.

Filter out packets sent to our VXLAN multicast group address to avoid this issue.

Fixes #3025

---

Completely untested, but seems straightforward enough.

Another point for discussion: Do we want to remove some protocols from the gluon-ebtables-filter-multicast allowlist? Running Babel, OSPF or RIP-NG over a regular mesh seems like a bad idea, and unusual enough that requiring to disable the filter may be justified.


rotanid (approved):

rotanid:
> Another point for discussion: Do we want to remove some protocols from the gluon-ebtables-filter-multicast allowlist? Running Babel, OSPF or RIP-NG over a regular mesh seems like a bad idea, and unusual enough that requiring to disable the filter may be justified.

next meetup or via an issue?

T-X:
> Another point for discussion: Do we want to remove some protocols from the gluon-ebtables-filter-multicast allowlist? Running Babel, OSPF or RIP-NG over a regular mesh seems like a bad idea, and unusual enough that requiring to disable the filter may be justified.

I like removing this with a different, egoistic motivation ;-). Each of them uses its own multicast destination address anyway, not ff02::1. With an up-to-date mesh network they should be allowed automatically anyway. And can make use of the new batman-adv multicast packet type. So if someone wanted to use this, it might be a nice incentive to update their network first.

Issue #3479, opened 2025-04-13
ipq806x: broken network configuration in specific scenarios


blocktrron:
### Bug report

**What is the problem?**

With current (04/2025) Gluon master, assigning the Gluon networks zones "uplink" and "client" with identical roles (in this case WAN) renders the ethernet ports of the device unusable.

Device in Question: Netgear R7800

**What is the expected behaviour?**

Zones apply correctly to the device and render a usable configuration.

The issue is with the switch to OpenWrt 24.10, the ipq806x target now uses DSA for configuring the switch. The ipq806x target has the specialty of connecting to two CPU ports in parallel. By default, WAN ports are assigned to a different CPU port as the LAN ports.

This leads to (yet  non-investigated) issues, probably while performing offload of the bridge in hardware.

Changing the DSA conduit of all ports to the same CPU port mitigates the issue.

**Gluon Version:**

60734d56ee3ffb08ff1f5c99ee8367f642a2c843

**Site Configuration:**
Not relevant

**Custom patches:**
None

Djfe:
there is a pending patch upstream to solve this properly that would allow us to revert #3480 
I haven't tested it, yet.

https://github.com/openwrt/openwrt/pull/21317

neocturne:
Workarounded, closing - workaround should be reevaluated on major OpenWrt updates.

PR #3480, opened 2025-04-13
gluon-core: use single conduit interface for DSA topology

blocktrron:
As explained in issue #3479, asisgning the same role over interfaces using different conduit interfaces can lead the network setup rendered in an dysfunctional state.

As a stop-gap solution, use the same conduit interface for all devices with the problematic topology.

Link: https://github.com/freifunk-gluon/gluon/issues/3479


package/gluon-core/luasrc/lib/gluon/upgrade/205-dsa-conduit
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/lua
+
+local uci = require('simple-uci').cursor()
+local platform = require 'gluon.platform'
+
+-- This is a workaround to reult in a configuration where all ports
```
rotanid, line 6: result

Djfe:
This only fixes the issue on second sysupgrade, right? so that's likely too late since it renders WAN unusable, right?

neocturne:
@Djfe I don't understand your concern - it runs as an upgrade script, just like everything else that creates configuration on Gluon?

Djfe:
I guess that means I'm tired. sorry
upgrade scripts run after rebooting and aren't related to sysupgrade at all.

neocturne (approved):

PR #3488, opened 2025-04-18
generic: enable dnsmasq ed25519 keys regardless of target

blocktrron:
OpenWrt disabled ed25519 SSH keys for SMALL_FLASH targets back in 2020, however this now affects ramips-mt76x8 targets.

ALways compile dnsmasq with ed25519 support. Without this patch, previously working ed25519 keys would be unusable on said devices without this patch.

Link: https://github.com/openwrt/openwrt/commit/d0f295837a03f7f52000ae6d395827bdde7996a4
Link: https://github.com/openwrt/openwrt/commit/fd46680966e9b9bc282b665282e2e6a22a27c26e

### DIFF
--- a/targets/generic
+++ b/targets/generic
@@ -58,6 +58,10 @@ try_config('PACKAGE_dnsmasq_full_rtc', false)
 
 try_config('TARGET_SQUASHFS_BLOCK_SIZE', 256)
 
+-- OpenWrt disables ed25519 keys on SMALL_FLASH targets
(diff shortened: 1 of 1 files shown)


neocturne:
Ugh, the second after I merged I realized that the commit description is wrong - of course, this is about dropbear, not dnsmasq.

neocturne:
I have updated the PR description, so at least that is correct.

blocktrron:
Yeah, sorry about that.

PR #3493, opened 2025-04-26
ramips-mt76x8: add support for Cudy TR1200 v1

blocktrron:
  - [ ] Web interface
  - [ ] TFTP
  - [x] Other: Intermediate OpenWrt firmware
        `cudy-tr1200-v1`
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs (No Radio LEDs preset)
    - [ ] Should show activity
  - Switch port LEDs (No switch port LEDs present)
- Outdoor devices only: (No outdoor device)
- Cellular devices only: (No cellular device)
- Docs:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -506,6 +506,7 @@ ramips-mt76x8
 
 * Cudy
 
+  - TR1200 (v1)
   - WR1000 (v1)
 
 * GL.iNet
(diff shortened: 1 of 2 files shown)


neocturne:
For unchecked boxes, it would be helpful to add a comment to the checklist explaining the reason - for #3492, I had to look up the device myself to figure out that it doesn't have WLAN.

Here, there radio LED boxes are not checked. Does that mean that the device does not have such LEDs?

blocktrron:
I've added the reasons for not checking the boxes in the initial description.

The device does not have radio nor switch port LEDs.

PR #3499, opened 2025-05-05
gluon-setup-mode: provide LED feedback for setup-mode activation

blocktrron:
This adds a fast-blink animation when Gluon started the procedure to enter setup mode.

This is helpful, as some devices might not provide feedback by the status LED on reboot.
This can lead to unwanted activation of a bootloader recovery procedure in case the reset button is not released prior.

### DIFF
--- a/package/gluon-setup-mode/files/etc/hotplug.d/button/50-gluon-setup-mode
+++ b/package/gluon-setup-mode/files/etc/hotplug.d/button/50-gluon-setup-mode
@@ -6,6 +6,7 @@ wait=3
 
 wait_setup_mode() {
 	sleep $wait
+	/lib/gluon/setup-mode/led.sh confirm
 	gluon-enter-setup-mode
 }
 
--- a/package/gluon-setup-mode/files/lib/gluon/setup-mode/led.sh
+++ b/package/gluon-setup-mode/files/lib/gluon/setup-mode/led.sh
@@ -1,0 +1,37 @@
+#!/bin/sh
+
+# shellcheck disable=SC1091,SC2154
+
+get_setup_mode_led() {
+	. /etc/diag.sh
+	get_status_led 2> /dev/null
+
+	if [ -z "$status_led" ]; then
+		status_led="$running"
+	fi
+
+	if [ -z "$status_led" ]; then
+		status_led="$boot"
+	fi
+
+	custom_led="$(lua -e 'print(require("gluon.setup-mode").get_status_led() or "")')"
+	if [ -z "$status_led" ]; then
+		status_led="$custom_led"
+	fi
+}
+
+MODE="$1"
+case "$mode" in
+	confirm)
+		get_setup_mode_led 2> /dev/null
+		status_led_set_timer 100 100
+		;;
+	running)
+		get_setup_mode_led 2> /dev/null
+		status_led_set_timer 1000 300
+		;;
+	*)
+		echo "Usage: $0 <running|confirm>"
+		exit 1
+		;;
+esac
--- a/package/gluon-setup-mode/files/lib/gluon/setup-mode/rc.d/S96led
+++ b/package/gluon-setup-mode/files/lib/gluon/setup-mode/rc.d/S96led
@@ -1,30 +1,8 @@
 #!/bin/sh /etc/rc.common
 
-# shellcheck disable=SC1091,SC2154
-
 START=96
 
 start() {
-	local custom_led
-
 	/etc/init.d/led start
-
-	. /etc/diag.sh
-
-	get_status_led 2> /dev/null
-
-	if [ -z "$status_led" ]; then
-		status_led="$running"
-	fi
-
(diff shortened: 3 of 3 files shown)


package/gluon-setup-mode/files/lib/gluon/setup-mode/led.sh
```diff
@@ -0,0 +1,35 @@
...
+	custom_led="$(lua -e 'print(require("gluon.setup-mode").get_status_led() or "")')"
+	if [ -z "$status_led" ]; then
+		status_led="$custom_led"
+	fi
+}
+
+MODE="$1"
+if [ "$MODE" = "running" ]; then
+	# runing - slow pulse
+	get_setup_mode_led 2> /dev/null
```
neocturne, line 26: Why is only one of these redirecting stderr?

package/gluon-setup-mode/files/lib/gluon/setup-mode/led.sh
```diff
@@ -0,0 +1,35 @@
...
+	fi
+
+	custom_led="$(lua -e 'print(require("gluon.setup-mode").get_status_led() or "")')"
+	if [ -z "$status_led" ]; then
+		status_led="$custom_led"
+	fi
+}
+
+MODE="$1"
+if [ "$MODE" = "running" ]; then
```
neocturne, line 24:
```
case "$MODE" in
...
```

package/gluon-setup-mode/files/etc/hotplug.d/button/50-gluon-setup-mode
```diff
@@ -6,6 +6,7 @@ wait=3
 
 wait_setup_mode() {
 	sleep $wait
+	/lib/gluon/setup-mode/led.sh confirm
```
neocturne, line 9: There is a race condition here; the shell process could be killed between setting the led and running `gluon-enter-setup-mode`. Maybe we should move the LED trigger into `gluon-enter-setup-mode`?

neocturne (changes requested):

package/gluon-setup-mode/files/etc/hotplug.d/button/50-gluon-setup-mode
blocktrron: The problem is the target-specific implementation of the `diag.sh` script which throws unrelated errors for our case.

On pure DT platforms we would not need it (only real consumer for us is x86 with apu)

neocturne: I don't understand how this answers my comment

Djfe:
I like this idea a lot actually. Since I did it many times I know by now that I can release the button way earlier but this is actual useful feedback for our end-users!

package/gluon-setup-mode/files/etc/hotplug.d/button/50-gluon-setup-mode
blocktrron: Hey, sorry i did not finish it.

So the race condition is for the cause of `gluon-enter-setup-mode` setting `set -eu`.

The LED definition is sourced from `diag.sh` which is included in other contexts where additional variables are set.

Including it within the `gluon-enter-setup-mode` script results in the script aborting early.

package/gluon-setup-mode/files/lib/gluon/setup-mode/led.sh
blocktrron: This is an oversight.

package/gluon-setup-mode/files/etc/hotplug.d/button/50-gluon-setup-mode
neocturne: The race condition I meant was `wait_setup_mode` being killed after `/lib/gluon/setup-mode/led.sh confirm`, but before `gluon-enter-setup-mode`, if the button is released at exactly the wrong moment.

blocktrron: Now i get it. I'll think about how to tackle this.

neocturne: The easiest way to fix this would be to call `/lib/gluon/setup-mode/led.sh confirm` from a subprocess together with `gluon-enter-setup-mode` instead of `wait_setup_mode` - either by moving the call into `gluon-enter-setup-mode` itself, or by introducing a new shell using `sh -c` for the two commands.

rotanid:
@blocktrron are you still interested in getting this merged, then it would be good to follow up with the remark by @neocturne and fixing the linting :-)

maurerle:
The Aruba AP-32x devices trigger a bootloader recovery as well, if the button is still pressed during boot. This bricks the device and requires serial access (via the dedicated serial port, so not a huge issue).

This PR would help a lot with this.

PR #3501, opened 2025-05-07
mvebu-cortexa9: add support for CZ.NIC Turris Omnia

maurerle:
Hi, someone new in our community has this device and wants to use it for Freifunk.
While it is a few years old, it is still an interesting device.

We successfully flashed the firmware but did not yet run through most parts of the checklist.

  - [ ] Web interface
  - [ ] TFTP
  - [x] Other: https://openwrt.org/toh/turris/turris_omnia
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [ ] Should show activity
  - Switch port LEDs
- ~~Outdoor devices only:~~
- ~~Cellular devices only:~~
- Docs:

### DIFF
--- a/targets/mvebu-cortexa9
+++ b/targets/mvebu-cortexa9
@@ -1,3 +1,8 @@
 device('linksys-wrt1200ac', 'linksys_wrt1200ac', {
 	factory_ext = '.img',
 })
+
+device('cznic-turris-omnia', 'cznic_turris-omnia', {
+	factory = false,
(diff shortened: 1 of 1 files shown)


herbetom:
If a PR isn't ready to be merged i would suggest marking it as an draft. That way it's much easier to see when the author of an PR thinks something is actually ready and needs attention. Thanks :)

maurerle:
This is currently not moving forward, due to missing device access in the community.
Closing it for now.

PR #3505, opened 2025-05-12
ramips-mt7621: add support for Genexis Pulse EX400

blocktrron:
  - [ ] Web interface
  - [ ] TFTP
  - [x] Other: Bootloader WebIf --> OpenWrt factory --> Gluon sysupgrade
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [x] Should show activity
  - Switch port LEDs
- Outdoor devices only:
- Cellular devices only:
- Docs:

### DIFF
--- a/targets/ramips-mt7621
+++ b/targets/ramips-mt7621
@@ -53,6 +53,14 @@ device('d-link-dir-878-a1', 'dlink_dir-878-a1')
 device('d-link-dir-882-a1', 'dlink_dir-882-a1')
 
 
+-- Genexis
+
+device('genexis-pulse-ex400', 'genexis_pulse-ex400', {
+	factory = false,
+	sysupgrade = '-sysupgrade',
+})
+
+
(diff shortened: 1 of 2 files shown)


targets/ramips-mt7621
```diff
@@ -53,6 +53,14 @@ device('d-link-dir-878-a1', 'dlink_dir-878-a1')
 device('d-link-dir-882-a1', 'dlink_dir-882-a1')
 
 
+-- Genexis
+
+device('genexis-inteno-pulse-ex400', 'genexis_pulse-ex400', {
```
herbetom, line 58: With https://github.com/openwrt/openwrt/commit/ea97c49766ad55cdfedcf7bd3eaf3b2a66d29730 this should probably be just the one name:

```suggestion
device('genexis-pulse-ex400', 'genexis_pulse-ex400', {
```

PR #3510, opened 2025-05-15
mediatek-filogic: add Wavlink WL-WN573HX3 / 7Links WLR-1300

s-2:
  - [x] Web interface
  - ~TFTP~
  - ~Other: <specify>~
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - ~Should map to their respective radio~
    - ~Should show activity~
  - Switch port LEDs
- Outdoor devices only:
- Cellular devices only:
  - ~Added board name to `is_cellular_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`~
  - ~Added board name with modem setup function `setup_ncm_qmi` to `package/gluon-core/luasrc/lib/gluon/upgrade/250-cellular`~
- Docs:

------------------------------

This device is also known as 7Links WLR-1300, sold in Europe by Pearl (https://www.pearl.de/a-ZX5612-1131.shtml).
There is an alias set in OpenWrt, is any additonal configuration required for gluon regarding autoupdate?

OEM Web interface uses sysupgrade without the option to forget uci, which results in broken network config.
A script to address this by calling `firstboot` was rejected in OpenWrt logn time ago (https://github.com/openwrt/openwrt/pull/4174#issuecomment-842434735), but maybe we could add it downstream to improve user experience.

What is the ideal way of dealing with this, maybe we could reset uci configuration upon first boot directly, when the presence of a particular key is detected, so that no additional reboot would be required?

I'm not sure what would be the best way to implement this though, but it also affects other devices (e.g. COVR-C1200 and COVR-P2500 from the `ath79` target, presumably further new devices in the future).

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -345,6 +345,10 @@ mediatek-filogic
 
   - UniFi 6 Plus
 
+* Wavlink
+
+  - WL-WN573HX3 [#lan_as_wan]_
+
 * Zyxel
 
   - NWA50AX Pro
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua
@@ -65,7 +65,8 @@ function M.is_outdoor_device()
 		return true
 
 	elseif M.match('mediatek', 'filogic', {
-		'cudy,ap3000outdoor-v1'
+		'cudy,ap3000outdoor-v1',
+		'wavlink,wl-wn573hx3',
 	}) then
 		return true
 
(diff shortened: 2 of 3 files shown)


s-2:
Indeed, gluon does not seem to care about previous OEM settings in UCI, initial setup works even without the script.

So this probably just affected OpenWrt in the first place (initially, the issue was discovered with ath79 devices, coming with OEM firmware based on ar71xx; I'll try those devices (e.g. COVR-P2500) with gluon as well, since the network config from ath71xx to ath79 was solved downstream in gluon, those might as well work with gluon).

Dropped `package/gluon-core/luasrc/lib/gluon/upgrade/900-discard-oem-config`.

AiyionPrime (approved):

AiyionPrime:
Thanks for your contribution!

PR #3511, opened 2025-05-16
scripts: Alter shebangs to /usr/bin/env bash

AiyionPrime:
> in order to increase portability across systems where /bin/ does not contain bash.

### DIFF
--- a/contrib/actions/free-runner-space.sh
+++ b/contrib/actions/free-runner-space.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # For a List of pre-installed packages on the runner image see
 # https://github.com/actions/runner-images/tree/main?tab=readme-ov-file#available-images
--- a/contrib/actions/show-system-info.sh
+++ b/contrib/actions/show-system-info.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 echo "-- CPU --"
 cat /proc/cpuinfo
--- a/contrib/depdot.sh
+++ b/contrib/depdot.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # Script to output the dependency graph of Gluon's packages
 # Limitations:
--- a/contrib/lsupgrade.sh
+++ b/contrib/lsupgrade.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -e
 # Script to list all upgrade scripts in a clear manner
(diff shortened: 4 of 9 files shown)


T-X:
I don't fully understand this yet. Could you elaborate on what kind of systems are affected? And what would fail in these scripts exactly otherwise? How would I know when I would need to add /usr/bin/env to the shebang of a script?

herbetom (approved):

herbetom:
@T-X using whichever bash is available in PATH via `#!/usr/bin/env bash` is what makes it much more portable.

PATH can be easibly modified to whatever is needed. In contrast `/bin/bash` is hardcoding where bash needs to be placed.
This means without some form of chroot/container/whatever it wouldn't be possible to use different variants of bash for different purposes.

One Linux Distro which is taking advantage of this (which i'm aware of) is NixOS. `/bin/` is really quite empty with it (okay, `/bin/sh` exists, but that's it):

```bash
# ls -l /bin/
total 4
lrwxrwxrwx 1 root root 74 May 17 01:04 sh -> /nix/store/byps33s26fp9hq3my3m7s79xavzk32dr-bash-interactive-5.2p37/bin/sh
# ls -lah /usr/bin/
total 12K
drwxr-xr-x 2 root root 4.0K May 17 01:04 .
drwxr-xr-x 3 root root 4.0K May 14 20:27 ..
lrwxrwxrwx 1 root root   65 May 17 01:04 env -> /nix/store/440q5scq8paszj2sdgz98hxl1rz12i88-coreutils-9.5/bin/env
```

This makes it possbile to have different packages per user, application or even terminal.

T-X:
@herbetom and for /usr/bin/env it is safe to assume that it exists? Or could another distro come along with only having /bin/env?

neocturne:
While there is no standard for any of these paths, /usr/bin/env is usually a safe bet. It is also often recommended to usr `/usr/bin/env python3` as Shebang for Python scripts.

AiyionPrime:
Hey there, this went faster than anticipated, thanks 😅

@T-X I'm rather new to it, but yeah, as tom said my goal was to be able ton run our scripts on Nix.

I went down the rabbit hole of which systems do not have /usr/bin/env, but they appear to be some ancient sun OS versions.

I tried finding a general advise on whether to use /bin/bash or env, but it actually depends on their availability during e.g. boot, where /usr might be mounted later, or your usage of arguments in the shebang.

All I can currently say is it doesn't hurt our build-process and decreases my local patch count :/

So thanks everyone!

blocktrron:
> and for /usr/bin/env it is safe to assume that it exists?

No, it is not. It is kind of "expected" to be available on Linux, but it is not defined in any POSIX specification and the "portability improvement" is basically a "NixOS requires us to change it".

PR #3512, opened 2025-05-24
[RFC] gluon-core: activate mesh_nolearn

Djfe:
Activate mesh_nolearn property for 802.11s mesh networks

What it does:
https://git.sipsolutions.net/iw.git/commit/mesh.c?h=v3.2&id=94ded888ba696a1ec03571d596e96cf944ccc0c8
This feature was added in by T-X in 2020 to be used for 802.11s meshes with separate mesh protocols like batman-adv but Gluon never activated it, yet.

Another recent commit was added upstream lately that might've tried to prevent what this patch/feature was meant to solve initially.
https://github.com/openwrt/openwrt/pull/18657

_____________________

I think this might've been forgotten about and should be activated for batman-adv networks, right?

Please correct my commit message if something isn't accurate.


T-X:
Generally I'm not a fan of adding changes without a proven benefit. Otherwise we might end up accumulating diffs to the OpenWrt/Linux defaults with a homeopathic effect.

I can't 100% remember, but I think I had made this change at my previous job for a slightly different scenario. For one where every second to reestablish a mesh connection would count. However with our 5 seconds batman-adv orig-interval, I'm not quite sure if this change alone has much use.

Also, I'm not sure how many people are actively using this setting so far. I've never promoted it anywhere.

However if someone tests this and can show in a deployment scenario that in this or that scenario x and y works better with Gluon with this setting, it'd be a different story.

Djfe:
Actually mesh_nolearn is recommended in the openwrt wiki in some place for batman-adv mesh setups
https://openwrt.org/docs/guide-user/network/wifi/mesh/802-11s#setting_mesh_parameters_automatically_using_hotplug
Though it's not mentioned here
https://openwrt.org/docs/guide-user/network/wifi/mesh/batman
So I assume atleast some people already use it

> Generally I'm not a fan of adding changes without a proven benefit.

Agreed, but now that I know that it needs to be proven first:
What could be useful test procedures for this?

### step 1

1. Select a mesh cloud and reboot all nodes (for reducing issues due to other complications)
2. Wait until mesh stabilizes again (high TQ)
3. Select one of the the wifi nodes with mesh-vpn in said cloud
4. Restart wifi and measure via ssh how long it takes to see neighbor nodes via batctl, also measure how long it takes for the node to have high TQ (TQ reaching a certain value like 80 or 90 that is not 100)

### step 2

5. Activate mesh_nolearn on several mesh nodes in a mesh cloud
6. Reboot them
7. Wait a few minutes
8. Run wifi on one of the nodes with mesh-vpn and then repeat the same measurements as in step 4

### notes

Do atleast 10 measurements (after restarting wifi) in each scenario and compare how they perform
Repeat the same thing for different wifi plattforms (atleast ath9k, ath10k, mt76)
This could show whether mesh_nolearn would actually reduce latency in reestablishing connections or not. Even though you assumed that this pales in comparison to the 5 seconds rig-interval it might still add significant latency but who knows.

### Alternative test scenario:
(somehow) measure whether mesh_nolearn makes a wifi setup more stable.
For example higher mesh TQ in a high density environment or when nodes are further away from each other. Or is not expected to even have an influence on mesh TQ at all?

### further notes:
Also compare TQ for setups where TQ is close to zero so mesh learning might run more frequently (?) like in the setup @benzea described.

I'd argue to choose mostly mesh setups where nodes are on the same channel (default) and where mesh_learning actually wants to figure out additional routes even if batman-adv would never consider them due to low TQ

These are just some ideas from a noobish kind of perspective though and your input on them would be valuable.

benzea:
Honestly, it seems like a no-brainer to set mesh_nolearn. From a quick glance at the code, this is exactly what we want if batman-adv is running anyway.

We'll have a Freifunk meeting tomorrow night. Could try to get a build with this patch rather than my OpenWRT one (or just both) and see what happens. I believe my patch is still correct and (independently) makes sense. In my view, the mesh path algorithm is just really bad when there are multiple paths available but all of them are bad (as the metric gets reset when the path changes). That said, nobody actually wants to use it that way anyway, so we should just set mesh_nolearn and be done with it.

Djfe:
Would you be interested in joining the gluon dev meeting on June 10th 20:00 CEST on mumble.freifunk.net? we could discuss this topic there.

benzea:
> Would you be interested in joining the gluon dev meeting on June 10th 20:00 CEST on mumble.freifunk.net? we could discuss this topic there.

Not sure I have much to say, but sure, I can make that.

benzea:
Some context (as it is spread around). The issues is that due to a mac80211 path handling bug, nodes that are not forwarding are still responding saying that they have a path to a requested node and will forward packets.

You can see this happening because the PATH_CHANGE counter goes up if you run `iw dev meshX mpath dump`.

You can see if the option worked by running `iw dev meshX mesh_param dump`.

I said it was a "no brainer" because mesh nolearn simply disables the whole path discovery logic where the bug is.

AiyionPrime:
Good afternoon from brunswick. We spent the better part of the evening exploring the background of this PR.

@Djfe I hope you don't mind me rewriting the commit message.
We tried to put it as simple as possible.

The changes were tested, do affect the interface properly and survive reboots.
As soon as the CI passes, I'd merge this.

AiyionPrime (approved):
We couldn't have written it better ourselves :D

Djfe:
No I don't mind at all ^^
Glad this got merged, great to hear from Brunswick and how productive it was.

> This also works around known bugs in mac80211, which we can avoid
altogether by disabling the learning.

I would love to read more on this tbh.
But not super important right now.

PR #3516, opened 2025-05-28
ipq806x-generic: add Extreme Networks WS-AP3935i

grische:
  - [ ] Web interface
  - [ ] TFTP
  - [x] Other: RJ45 serial console like other Extreme Network/Enterasys devices
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [x] Should show activity (both mesh and client traffic lights up LED)
  - Switch port LEDs
- ~~Outdoor devices only:~~
  - ~~[ ] Added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`~~
- ~~Cellular devices only:~~
  - ~~[ ] Added board name to `is_cellular_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`~~
  - ~~[ ] Added board name with modem setup function `setup_ncm_qmi` to `package/gluon-core/luasrc/lib/gluon/upgrade/250-cellular`~~
- Docs:

---

### LAN port numbering / assignment 

The ports at the back look like this:
```
  +---------+  +---------+  +---------+
  | Console |  |  LAN 1  |  |  LAN 2  |
  |   [ ]   |  |   [ ]   |  |   [ ]   |
  +---------+  +---------+  +---------+
```

Stock OpenWRT:
- LAN 1 = LAN
- LAN 2 = WAN

At the same time, the front has two LAN LEDs labelled "1" and "2"
- LAN "1" LED = WAN (=LAN 2)
- LAN "2" LED = LAN (=LAN 1)

Should this LAN confusion be fixed in upstream OpenWRT or as part of this PR here?

### DIFF
--- a/targets/ipq806x-generic
+++ b/targets/ipq806x-generic
@@ -8,8 +8,17 @@ local QCA9984_PACKAGES = {'kmod-ath10k', '-kmod-ath10k-ct', 'ath10k-firmware-qca
 
 
 --
--- QCA9980
+-- QCA9980 / QCA9990
 --
+
+-- Extreme Networks
+
+device('extreme-networks-ap3935', 'extreme_ap3935', {
+	factory = false,
+	sysupgrade = '-squashfs-nand-sysupgrade',
+	packages = QCA9980_PACKAGES,
+	broken = true,	-- https://github.com/openwrt/openwrt/issues/17635
+})
 
 -- TP-Link
 device('tp-link-archer-c2600', 'tplink_c2600', {


Djfe:
The edgecore also needed to switch eth0 and eth1 in 02_network upstream, I assume a kernel update changed the order
https://github.com/openwrt/openwrt/commit/84a489b7cfc4c98b8747c9965826b2588eb3a57d

the meraki 52 might be affected as well but Noone noticed

patches/openwrt/0011-ipq806x-fix-Extreme-Networks-AP3935-LAN-WAN-ports.patch
```diff
@@ -0,0 +1,28 @@
...
+ 		;;
+ 	qcom,ipq8064-ap161)
+@@ -63,7 +64,6 @@ ipq806x_setup_interfaces()
+ 	meraki,mr42)
+ 		ucidef_set_interface_lan "eth0"
+ 		;;
+-	extreme,ap3935 |\
+ 	meraki,mr52)
+ 		ucidef_set_interfaces_lan_wan "eth0" "eth1"
+ 		;;
```
blocktrron, line 28: Please send this with a proper description upstream.

blocktrron (dismissed):

patches/openwrt/0011-ipq806x-fix-Extreme-Networks-AP3935-LAN-WAN-ports.patch
grische: https://github.com/openwrt/openwrt/pull/19077

grische: @blocktrron should these persist as (to-be-updated) patches here or should I ask upstream to backport the changes to 24.10?

blocktrron: @grische I have some pending commits, I'll backport them in the coming days.

grische:
While testing, I noticed that the LAN1 (now WAN) port does not work if the system has been booted without LAN1 plugged in (Or not plugged in within a shirt time after boot?). 

Similar to bug  https://github.com/openwrt/openwrt/issues/17635

patches/openwrt/0011-ipq806x-fix-Extreme-Networks-AP3935-LAN-WAN-ports.patch
grische: patches have been upstreamed

grische:
Except for the above issue, where LAN1 does not get up if no cable is plugged in during boot there are no blockers. 
Even the LAN1 [has a simple workaround](https://github.com/openwrt/openwrt/issues/17635#issuecomment-3065909100): restarting networking powers up LAN1 again.

I propose we merge it as-is, as this should not be a blocker for deployment of the APs.

docs/user/supported_devices.rst
```diff
@@ -269,6 +269,10 @@ ipq40xx-mikrotik
 ipq806x-generic
 ---------------
 
+* Extreme Networks
+
+  - AP3935
+
```
AiyionPrime:
```suggestion
```

targets/ipq806x-generic
```diff
@@ -8,9 +8,18 @@ local QCA9984_PACKAGES = {'kmod-ath10k', '-kmod-ath10k-ct', 'ath10k-firmware-qca
...
+-- QCA9980 / QCA9990
 --
 
+-- Extreme Networks
+
+device('extreme-networks-ap3935', 'extreme_ap3935', {
+	factory = false,
+	sysupgrade = '-squashfs-nand-sysupgrade',
+	packages = QCA9980_PACKAGES,
+	broken = true,	-- 802.11s works, but https://github.com/openwrt/openwrt/issues/17635
```
AiyionPrime:
```suggestion
	broken = true,	-- https://github.com/openwrt/openwrt/issues/17635
```

AiyionPrime (changes requested):
Looks good, but it won't be supported as broken.
Also no need to describe what's working.

AiyionPrime (approved):

AiyionPrime (approved):

rotanid:
https://github.com/openwrt/openwrt/issues/17635
https://github.com/openwrt/openwrt/pull/23208

PR #3519, opened 2025-06-01
ath79-generic: disable TP-Link Archer C6 (v2 EU/RU/JP) factory image

herbetom:
The flash is quite small and apparently building the factory will fail prior to the sysupgrade.

Disabling the factory image seems like the best course of action as installing via an OpenWrt factory image should be resonably easy to do.

Reference: https://github.com/freifunk-darmstadt/site-ffda/pull/128


blocktrron:
This depends on the included package set, which is pretty much in Darmstadt.

herbetom:
Yes, and i've seen that you've implemented a reduced package set for Darmstadt instead.

Hoewer, i do think giving communities more briefing room also wouldn't be a bad thing. But since i didn't gather data on how much this really would improve the situation i've got no hard opinion wheter this really would improve it that much.

PR #3527, opened 2025-06-20
gluon-mesh-layer3-common: get rid of gsub substitution count on prefix

dg0tm:
because is failed in gluon-mesh-olsrd: 360-gluon-mesh-olsrd-setup-intf on line: 132

### DIFF
--- a/package/gluon-mesh-layer3-common/luasrc/usr/lib/lua/gluon/l3.lua
+++ b/package/gluon-mesh-layer3-common/luasrc/usr/lib/lua/gluon/l3.lua
@@ -7,7 +7,7 @@ local M = {}
 function M.node_client_prefix6()
 	local key = "gluon-l3roamd.node_client_prefix6"
 	local prefix_seed = util.domain_seed_bytes(key, 7)
-	return ("fd" .. prefix_seed):gsub(("(%x%x%x%x)"):rep(4), "%1:%2:%3:%4" .. "::/64")
+	return (("fd" .. prefix_seed):gsub(("(%x%x%x%x)"):rep(4), "%1:%2:%3:%4" .. "::/64"))
 end
 
 return M


package/gluon-mesh-layer3-common/luasrc/usr/lib/lua/gluon/l3.lua
```diff
@@ -7,7 +7,9 @@ local M = {}
 function M.node_client_prefix6()
 	local key = "gluon-l3roamd.node_client_prefix6"
 	local prefix_seed = util.domain_seed_bytes(key, 7)
-	return ("fd" .. prefix_seed):gsub(("(%x%x%x%x)"):rep(4), "%1:%2:%3:%4" .. "::/64")
+	local prefix = ("fd" .. prefix_seed):gsub(("(%x%x%x%x)"):rep(4), "%1:%2:%3:%4" .. "::/64")
+
+	return prefix
```
neocturne: Parentheses are sufficient to drop the second return value:
```suggestion
	return (("fd" .. prefix_seed):gsub(("(%x%x%x%x)"):rep(4), "%1:%2:%3:%4" .. "::/64"))
```

neocturne:
The commit message is confusing if you don't remember that gsub returns a substitution count. Replacing "gsub substitution count" with "substitution could returned by gsub" would make that clearer.

neocturne:
Oops, I had a typo in my commit message description, that should be "substitution count" instead of "substitution could" of course.

T-X:
@dg0tm sorry, a bit off-topic my following question but asking it here anyway because I don't see you in the Gluon IRC/Matrix chat room: Are you actively using Gluon with OLSR? The issue we have is that it too has the same risk like the Babel implementation we had in the passed, that is missing users+maintainers. I think @mkg20001 is currently not working on this anymore.
It would also be great if someone were adding a respondd export for OLSR in Gluon so that we could better track it in the Gluon Census:
* https://github.com/freifunk-gluon/census-exporter/
* https://stats.darmstadt.freifunk.net/d/CA5PRFmMz/gluon-census?orgId=1

leoss:
@T-X I will give a short answer to not clutter this PR. @dg0tm and me, we are from Opennet Initiative (www.opennet-initiative.de). We are using OLSR for a very long time. We also began to transition to OLSRv2 (a long time ago).
We have a custom firmware based on OpenWrt. We think about switching to Gluon because lack of developer manpower.
To answer your question: No, we do not yet use Gluon with OLSR.

PR #3528, opened 2025-06-20
gluon-state-ntpd-check: Add package

AiyionPrime:
> a hotplug for busybox ntpd, which provides the current ntp sync state as a gluon-state-check.
> 
> This check is particularly interesting in WireGuard based communities.
> 
> Originally suggested by hexa, thanks!

This package provides two state-checks:
- `has_ntp_sync` in case the last NTP sync was successful
- `has_lost_ntp_sync` in case it wasn't.

Furthermore the modification times get updated by subsequent events
and this package provides `gluon-ntp-info` a command which interprets the creation and modification times of the two flags and produces human readable output like this:

```console
NTP has been unsynced for 706 seconds (last checked 46 seconds ago).
```

### DIFF
--- a/package/gluon-state-ntpd-check/files/etc/hotplug.d/ntp/21-gluon-state-check
+++ b/package/gluon-state-ntpd-check/files/etc/hotplug.d/ntp/21-gluon-state-check
@@ -1,0 +1,14 @@
+#!/bin/sh
+sync_flag="/var/gluon/state/has_ntp_sync"
+unsync_flag="/var/gluon/state/has_lost_ntp_sync"
+
+# shellcheck disable=SC2154 # stratum is set by busybox ntpd.
+if [ "$stratum" = 16 ]
+then
+	touch -a "$unsync_flag"
+	rm -f "$sync_flag"
+elif [ "$stratum" -lt 16 ]
+then
+	touch -a "$sync_flag"
+	rm -f "$unsync_flag"
+fi
--- a/package/gluon-state-ntpd-check/Makefile
+++ b/package/gluon-state-ntpd-check/Makefile
@@ -1,0 +1,12 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-state-ntpd-check
+
+include ../gluon.mk
+
+define Package/gluon-state-ntpd-check
+  TITLE:=Hotplug for ntpd into gluon-state-check
+  DEPENDS:=+gluon-core +gluon-state-check
+endef
+
+$(eval $(call BuildPackageGluon,gluon-state-ntpd-check))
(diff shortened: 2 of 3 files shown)


blocktrron (changes requested):
You need to fix the linter

AiyionPrime:
@blocktrron Thanks captain :kissing_closed_eyes:

AiyionPrime:
@blocktrron and I skipped the build, as the package is not part of the minimal site.
I'm building the image locally now.

AiyionPrime:
The local build succeeded, greetings from Karlsruhe!

```bash
gluon-root@watchdog-demo:~# gluon-ntp-info
NTP state is unknown, ntpd hotplugs might not have been invoked, yet.
```

and after more than 11? minutes:

```bash
root@watchdog-demo:~# uptime
 03:57:15 up 14 min,  load average: 0.22, 2.10, 2.54
root@watchdog-demo:~# gluon-ntp-info
NTP has been unsynced for 6 seconds (last checked 6 seconds ago).
```

package/gluon-state-ntpd-check/files/etc/hotplug.d/ntp/21-gluon-state-check
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+sync_flag="/var/gluon/state/has_ntp_sync"
+unsync_flag="/var/gluon/state/has_lost_ntp_sync"
+
+# shellcheck disable=SC2154 # stratum is set by busybox ntpd.
+[ "$stratum" = 16 ] && touch -a "$unsync_flag" && rm -f "$sync_flag"
+# shellcheck disable=SC2154 # stratum is set by busybox ntpd.
+[ "$stratum" -lt 16 ] && touch -a "$sync_flag" && rm -f "$unsync_flag"
```
blocktrron: Update so removal is done regardless of exit code of the previous operation

AiyionPrime: resolved in 4753f1c

AiyionPrime:
One thing to keep in mind:
NTP changes the time. So the `last checked` might jump to a point in time ages ago or in the future.
For anything important take into account how often the hotplug is supposed to be called (e.g. very 11 minutes) and ignore the output if the time is long ago or even in the future.

blocktrron (approved):

neocturne:
Ah, I missed this PR. Some late thoughts:
- This should not be a separate package - let's merge it into gluon-state-check - or even merge both into gluon-core
- Having two complementary files seems redundant to me, I'd prefer just to have the positive one

neocturne:
> One thing to keep in mind: NTP changes the time. So the `last checked` might jump to a point in time ages ago or in the future. For anything important take into account how often the hotplug is supposed to be called (e.g. very 11 minutes) and ignore the output if the time is long ago or even in the future.

A better solution might be to just write monotonous timestamps (from /proc/uptime) to the files instead of relying on filesystem timestamps.

AiyionPrime:
>     * Having two complementary files seems redundant to me, I'd prefer just to have the positive one

I'm afraid it's not redundant. There are three states NTP sync can have.

- synced
- unsynced
- unchecked

I'd like to be able to distinguish the three and not have one of them shadow the third.

PR #3532, opened 2025-06-24
gluon-l3roamd: startup fix

dg0tm:
after commit 085250c l3roamd does not start anymore
and remove the quotes seems fix the issue.

### DIFF
--- a/package/gluon-l3roamd/files/etc/init.d/gluon-l3roamd
+++ b/package/gluon-l3roamd/files/etc/init.d/gluon-l3roamd
@@ -55,7 +55,7 @@ start_service () {
 	procd_set_param stderr 1
 	procd_set_param respawn "${respawn_threshold:-3660}" "${respawn_timeout:-5}" "${respawn_retry:-0}"
 	# shellcheck disable=SC2086
-	procd_set_param command "$PROG" -s /var/run/l3roamd.sock "$prefix4" "$prefix6" $interfaces -t 254 -a "$localip" -b br-client "$roamingprefix"
+	procd_set_param command "$PROG" -s /var/run/l3roamd.sock $prefix4 $prefix6 $interfaces -t 254 -a "$localip" -b br-client "$roamingprefix"
 	procd_close_instance
 }
 


AiyionPrime (changes requested):
Interesting.
Have you inspected the variables contents and could provide them here?

The quotes prevent word splitting.
Meaning if a variable contains a space in quotes it is considered one word, not two.

This prevents all sorts of problems and plainly reverting this is not the right way, I think.

But thanks for spotting the issue, I'm sure we can work out what the problem and the cleaner solution for this is :)

dg0tm:
hey thanks for take time!

with quotes procd parse as follow:

`{ "name": "gluon-l3roamd", "script": "\/etc\/init.d\/gluon-l3roamd", "instances": { "instance1": { "stdout": true, "stderr": true, "respawn": [ "3660", "5", "0" ], "command": [ "\/usr\/sbin\/l3roam
d", "-s", "\/var\/run\/l3roamd.sock", "", " -p fd32:d8d3:87da:1000::\/64", "-m", "eth0.101", "-i", "local-node", "-t", "254", "-a", "fd32:d8d3:87da:1001:7a8a:20ff:fe09:2cb2", "-b", "br-client", " -
P fd08:b475:52b6:d6c5::\/64" ] } }, "triggers": [ [ "interface.*", [ [ "run_script", "\/etc\/init.d\/gluon-l3roamd", "reload" ] ], 0 ] ] }`

without quotes:

`{ "name": "gluon-l3roamd", "script": "\/etc\/init.d\/gluon-l3roamd", "instances": { "instance1": { "stdout": true, "stderr": true, "respawn": [ "3660", "5", "0" ], "command": [ "\/usr\/sbin\/l3roamd", "-s", "\/var\/run\/l3roamd.sock", "-p", "fd32:d8d3:87da:1000::\/64", "-m", "eth0.101", "-i", "local-node", "-t", "254", "-a", "fd32:d8d3:87da:1001:7a8a:20ff:fe09:2cb2", "-b", "br-client", "-P", "fd08:b475:52b6:d6c5::\/64" ] } }, "triggers": [ [ "interface.*", [ [ "run_script", "\/etc\/init.d\/gluon-l3roamd", "reload" ] ], 0 ] ] }`

remove the space in https://github.com/freifunk-gluon/gluon/blob/main/package/gluon-l3roamd/files/etc/init.d/gluon-l3roamd#L45-L46 does also not help.

neocturne:
Oh! I totally missed the `-P` printed by Lua. The correct fix is to move the `-P` out of the Lua command and into the `procd_set_param command`.

neocturne:
Hmm, it's not that easy, there are multiple arguments that need to be split, and in the case of `prefix4` the `-p` is conditional. In that case removing the quotes (and disabling the shellcheck lint for that line) may actually be the best solution. `$localip` can keep the quotes.

dg0tm:
okay, should $interfaces than also been quoted?

by the way commit 59b7625e542b29a477b7d789ee071ff7eed86da8 remove the option -4 0:0:0:0:0:ffff::/96 and now l3roamd complain about that

`2025-06-24T20:01:13+02:00 oni-788a20092cb2 l3roamd[4533]: -4 was not specified. Defaulting to 0:0:0:0:0:ffff::/96
2025-06-24T20:01:13+02:00 oni-788a20092cb2 l3roamd[4533]: Initializing unix socket: /var/run/l3roamd.sock
2025-06-24T20:01:13+02:00 oni-788a20092cb2 l3roamd[4533]: initializing detection of wifistations
2025-06-24T20:01:13+02:00 oni-788a20092cb2 l3roamd[4533]: nl80211 not found.`

neocturne:
`$interfaces` also consists of multiple arguments and needs to be split, so no quotes.

leoss:
@AiyionPrime Yes, this still resolves the startup issue. We tested it.

AiyionPrime:
@dg0tm Thanks for your contribution!

PR #3536, opened 2025-06-30
gluon-scheduled-domain-switch: Replace ping with gluon-state checks

### LINKED ISSUE #2231: gluon-scheduled-domain-switch uses ping check
As it was decided in IRC Gluon should not perform ping checks. See #2228...

Unfortunately ping checks are already performed by https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-check-connection

I suggest moving the scheduled-domain-switch into the community package repository where such code trash belongs.

AiyionPrime:
I'd like to touch a relict again, now that we've got ntp checks.

As @mweinelt put it a few days ago:

> [...] irgendeine Kombination aus den folgenden Dinge prüfen
> 
> kein ntp sync
> kein batman gateway
> keine ipv6 route

- [x] check the three above
- [x] dependency on gluon-state-check (has_default_gw6)
- [x] dependency on batman_adv (has_default_gw4)
- [x] dependency on ntpd-check (has_lost_ntp_sync)
- [x] see if there might an absence check in check_site.lua *yes, 1702 introduced the feature.*
- [x] either make sure the outdated section has been removed using check_site lua ~or warn about it during runtime~
- [x] take a look at the docs
- [x] test all the things 

This intends to resolve #2231 eventually.

This was accepted in https://github.com/freifunk-gluon/gluon/issues/2231#issuecomment-857192174.

### DIFF
--- a/docs/package/gluon-scheduled-domain-switch.rst
+++ b/docs/package/gluon-scheduled-domain-switch.rst
@@ -10,6 +10,9 @@ powered off while this was supposed to happen, it might not be able to acquire t
 correct time. In this case, the node will switch after it has not seen any gateway
 for a given period of time.
 
+In older versions ping was used against an array of endpoints to determine mesh-connectivity.
+Nowadays *gluon-state-check* is used for this and evaluates mesh-(VPN) connectivity and NTP states.
+
 site.conf
 ---------
 All those settings have to be defined exclusively in the domain, not the site.
@@ -21,9 +24,6 @@ domain_switch : optional (needed for domains to switch)
     - amount of time without reachable gateway to switch unconditionally
   switch_time :
     - UNIX epoch after which domain will be switched
-  connection_check_targets :
-    - array of IPv6 addresses which are probed to determine if the node is
-      connected to the mesh
 
 Example::
 
@@ -31,8 +31,4 @@ Example::
     target_domain = 'new_domain',
     switch_after_offline_mins = 120,
     switch_time = 1546344000, -- 01.01.2019 - 12:00 UTC
-    connection_check_targets = {
-      '2001:4860:4860::8888',
-      '2001:4860:4860::8844',
-    },
   },
--- a/package/gluon-scheduled-domain-switch/Makefile
+++ b/package/gluon-scheduled-domain-switch/Makefile
@@ -6,7 +6,7 @@ include ../gluon.mk
 
 define Package/gluon-scheduled-domain-switch
   TITLE:=Allows scheduled migrations between domains
-  DEPENDS:=+gluon-core @GLUON_MULTIDOMAIN
+  DEPENDS:=+gluon-core +gluon-state-check @GLUON_MULTIDOMAIN
 endef
 
 $(eval $(call BuildPackageGluon,gluon-scheduled-domain-switch))
--- a/package/gluon-scheduled-domain-switch/check_site.lua
+++ b/package/gluon-scheduled-domain-switch/check_site.lua
@@ -2,5 +2,6 @@ if need_table(in_domain({'domain_switch'}), nil, false) then
 	need_domain_name(in_domain({'domain_switch', 'target_domain'}))
 	need_number(in_domain({'domain_switch', 'switch_after_offline_mins'}))
 	need_number(in_domain({'domain_switch', 'switch_time'}))
-	need_string_array_match(in_domain({'domain_switch', 'connection_check_targets'}), '^[%x:]+$')
+	obsolete({'domain_switch', 'connection_check_targets'},
+	'Connections to NTP servers and gateways are used for this now.')
 end
--- a/package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-check-connection
+++ b/package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-check-connection
@@ -5,7 +5,14 @@ local util = require 'gluon.util'
 local site = require 'gluon.site'
 
 local offline_flag_file = "/tmp/gluon_offline"
-local is_offline = true
+
+local ntp_negative_state_file = "/var/gluon/state/has_lost_ntp_sync"
+
+
+local function is_offline()
+	return unistd.access(ntp_negative_state_file)
+end
+
 
 -- Check if domain-switch is scheduled
 if site.domain_switch() == nil then
@@ -13,16 +20,7 @@ if site.domain_switch() == nil then
 	os.exit(0)
 end
 
--- Check reachability of pre-defined targets
-for _, ip in ipairs(site.domain_switch.connection_check_targets()) do
-	local exit_code = os.execute("ping -c 1 -w 10 " .. ip)
-	if exit_code == 0 then
-		is_offline = false
-		break
-	end
-end
-
-if is_offline then
+if is_offline() then
 	-- Check if we were previously offline
 	if unistd.access(offline_flag_file) then
 		os.exit(0)
--- a/package/gluon-autoupdater/check_site.lua
+++ b/package/gluon-autoupdater/check_site.lua
@@ -29,7 +29,7 @@ local branches = table_keys(need_table({'autoupdater', 'branches'}, function(bra
 		return good_signatures <= #pubkeys
 	end, nil, string.format('be less than or equal to the number of public keys (%d)', #pubkeys))
 
-	obsolete(in_site(extend(branch, {'probability'})), 'Use GLUON_PRIORITY in site.mk instead.')
+	obsolete(extend(branch, {'probability'}), 'Use GLUON_PRIORITY in site.mk instead.')
 end))
 
 need_one_of(in_site({'autoupdater', 'branch'}), branches, false)


AiyionPrime:
check site works great using the site from the current stable docs.
`*** domains/legacy.conf error: domain_switch.connection_check_targets is obsolete. Connections to NTP servers and gateways are used for this now.`

AiyionPrime:
The package dependencies work as expected, after 11 minutes this oneliner:

`while :; do uptime; ls /var/gluon/state; sleep 1; done`

emits

```console
 19:37:27 up 11 min,  load average: 0.00, 0.00, 0.00
 19:37:32 up 11 min,  load average: 0.00, 0.00, 0.00
 19:37:37 up 11 min,  load average: 0.00, 0.00, 0.00
has_lost_ntp_sync
[...]
```

So after 11 minutes the expected `stratum 16` event hits and all conditions are met.

~But in practice the current code might not even reach this point.~
*Since b98467cccd4aa01a3dcfacd2f860259fed43b812 it does.*

---

In domains with a configured domain switch (in my case dom99) every minute `/usr/bin/gluon-check-connection` is executed via `micron.d`.

After another minute this leads to the creation of `/tmp/gluon_offline` and a cron hitting every 5 minutes finds this and initiates the domain-switch.in case file contains an uptime smaller than `current_uptime-switch_after_offline_mins`.

Which worked out just fine and landed my router in `dom14`.

```console
root@scheduleddomainswitch:~# gluon-info 
Hostname:               	scheduled_domain_switch
MAC address:            	52:54:00:12:34:57
Hardware model:         	QEMU Standard PC (i440FX + PIIX, 1996)
Gluon version / Site version:	v2023.2-483-gb9bec9c / vH35-37-gab5d876+
Firmware release:       	exp-20250630
Site:                   	Freifunk Hannover
Domain:                 	dom14
[...]
```

That is, if the device is offline.

I think that's all there is to test.

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-check-connection
```diff
@@ -5,24 +5,31 @@ local util = require 'gluon.util'
...
+local function is_offline()
+	if not unistd.access(gw6_state_file) then
+		return true
+	elseif not unistd.access(gw4_state_file) then
+		return true
+	elseif unistd.access(ntp_negative_state_file) then
+		return true
+	end
+	return false
+end
```
AiyionPrime: Coming with a small downside, I'd suggest altering my current approach
`(any of the three conditions)` to `no NTP and (no GW4 or no GW6)`

```suggestion
local function is_offline()
        if unistd.access(ntp_negative_state_file) then
		if not unistd.access(gw6_state_file) then
			return true
		elseif not unistd.access(gw4_state_file) then
			return true
		end
	end
	return false
end
```

Which ensures: the device has tried to sync time via NTP, but failed.

The cost this imbues is: The stratum event hits after about 11 minutes.
Which means domain switches would not happen after minimal 5 minutes but 15 minutes of downtime upon reboot.

Especially in WireGuard domains this could decrease a false-positive rate, but is not that urgent.

Any comments on the suggestion?

AiyionPrime: resolved in b98467cccd4aa01a3dcfacd2f860259fed43b812

docs/package/gluon-scheduled-domain-switch.rst
```diff
@@ -10,6 +10,9 @@ powered off while this was supposed to happen, it might not be able to acquire t
 correct time. In this case, the node will switch after it has not seen any gateway
 for a given period of time.
 
+In older versions ping was used agaisnt an array of endpoints to determine mesh-connectivity.
+Nowadays *gluon-state-check* is used for this and evaluates mesh-(VPN) connectivity and NTP states.
```
maurerle, line 14: against typo

maurerle (approved):

docs/package/gluon-scheduled-domain-switch.rst
AiyionPrime: resolved in 5a038ebc2ef0479133208f1d6dece87efc2a5cf7

neocturne:
I would prefer just to have the NTP check, as the other event for switching domains is time-based. I also don't think there is any reason to restrict this to batman-adv.

See also https://github.com/freifunk-gluon/gluon/pull/3528#issuecomment-3025324759.

AiyionPrime:
If we just use the NTP check, the batman-adv limit can go.
batman-adv is currently the only one implementing has_default_gw4, iirc.

AiyionPrime:
regarding the comments in the other PR:
the NTP check is indeed cheap enough, we can merge it gluon_state.

regarding mergin both into core: No hard feelings on my end, but it was a rather harsh discussion which cost a lot of time back then.
If possible, I'd rather not touch it again in the forseeable future :D

AiyionPrime:
reduced to the offline check to the NTP check in da445b0242b6f0f0795610b482749e2fc32d66d5

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-check-connection
```diff
@@ -5,24 +5,24 @@ local util = require 'gluon.util'
 local site = require 'gluon.site'
 
 local offline_flag_file = "/tmp/gluon_offline"
-local is_offline = true
+
+local gw4_state_file = "/var/gluon/state/has_default_gw4"
+local gw6_state_file = "/var/gluon/state/has_default_gw6"
```
neocturne: Unused variables, as linted.

package/gluon-scheduled-domain-switch/Makefile
```diff
@@ -6,7 +6,7 @@ include ../gluon.mk
 
 define Package/gluon-scheduled-domain-switch
   TITLE:=Allows scheduled migrations between domains
-  DEPENDS:=+gluon-core @GLUON_MULTIDOMAIN
+  DEPENDS:=+gluon-core +gluon-state-ntpd-check  @GLUON_MULTIDOMAIN
```
neocturne:
```suggestion
  DEPENDS:=+gluon-core +gluon-state-check @GLUON_MULTIDOMAIN
```

package/gluon-scheduled-domain-switch/check_site.lua
```diff
@@ -2,5 +2,6 @@ if need_table(in_domain({'domain_switch'}), nil, false) then
 	need_domain_name(in_domain({'domain_switch', 'target_domain'}))
 	need_number(in_domain({'domain_switch', 'switch_after_offline_mins'}))
 	need_number(in_domain({'domain_switch', 'switch_time'}))
-	need_string_array_match(in_domain({'domain_switch', 'connection_check_targets'}), '^[%x:]+$')
+	obsolete(in_domain({'domain_switch', 'connection_check_targets'}),
```
neocturne:
```suggestion
	obsolete({'domain_switch', 'connection_check_targets'},
```
Hmm, there is probably no reason to keep the `in_domain` here... although there's another `obsolete` where we left an `in_site`.

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-check-connection
```diff
@@ -5,24 +5,24 @@ local util = require 'gluon.util'
...
+local ntp_negative_state_file = "/var/gluon/state/has_lost_ntp_sync"
+
+
+local function is_offline()
+	return unistd.access(ntp_negative_state_file)
```
neocturne, line 13: Will this ever trigger if a node comes up and the NTP server is already unreachable?

package/gluon-scheduled-domain-switch/check_site.lua
AiyionPrime: i think this is more explicit, which I'd favour.

package/gluon-scheduled-domain-switch/Makefile
AiyionPrime: resolved since 815cf6608c27a3d658349fe80ad8309c33ca652a

package/gluon-scheduled-domain-switch/luasrc/usr/bin/gluon-check-connection
AiyionPrime: Yes. It will appear, if the NTP synchronization is unsuccessful, which happens around 11 minutes of uptime.

package/gluon-scheduled-domain-switch/check_site.lua
neocturne: The innermost modifier matches first. So with `in_domain`, you'd get a message that the field should be in the domain if it appears in a multidomain `site.conf`, rather than the message that it is obsolete altogether. I would avoid the unnecessary `in_domain` because of that.

AiyionPrime: In that case I'd fix both occurrences?

neocturne: Makes sense

AiyionPrime: resolved in https://github.com/freifunk-gluon/gluon/compare/7d4b70fa194399342e7846933dedb2e3bf26f899..788a1c14f7df240c73a510daebc8e62aa1048322

neocturne:
Commit message states "Omit in_domain filter", but the commit actually removes `in_site` filter.

AiyionPrime:
@neocturne fixed.

neocturne (approved):

PR #3539, opened 2025-07-01
gluon-core: create mac address by name

maurerle:
This includes my first draft of the suggestion mentioned in https://github.com/freifunk-gluon/gluon/pull/3357#discussion_r2023746381 to refactor the `generate_mac` function to be more generic regarding mac generation.

>  In addition to calling this by indexes which effectively translate to an offset in the assigned block of n (8?) addresses, how about we use named indexes?
>
> This way, we can track all handed out addresses by name. This has the upside of allowing 3rd party community packages to obtain unique per-radio addresses. ~ blocktrron

I think that this should not be part of #3357, so I am creating a new draft PR for this to track.

### DIFF
--- a/docs/features/private-wlan.rst
+++ b/docs/features/private-wlan.rst
@@ -27,7 +27,7 @@ You may also enable a private WLAN using the command line::
   uci set wireless.wan_radio$RID.ssid="$SSID"
   uci set wireless.wan_radio$RID.key="$KEY"
   uci set wireless.wan_radio$RID.disabled=0
-  uci set wireless.wan_radio$RID.macaddr=$(lua -e "print(require('gluon.util').generate_mac(3+4*$RID))")
+  uci set wireless.wan_radio$RID.macaddr=$(lua -e "print(require('gluon.wireless').get_wlan_mac('wan_radio', $RID))")
   uci commit
   wifi
 
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/wireless.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/wireless.lua
@@ -39,13 +39,26 @@ local function get_wlan_mac_from_driver(radio, vif)
 	return addresses[vif+1]
 end
 
-function M.get_wlan_mac(radio, index, vif)
-	local addr = get_wlan_mac_from_driver(radio, vif)
-	if addr then
-		return addr
+local radio_mac_offsets = {
+	client = 0,
+	mesh = 1,
+	owe = 2,
+	wan_radio = 3,
+}
+
+function M.get_wlan_mac(func, index, radio)
+	local offset = radio_mac_offsets[func]
+	if offset == nil then
+		return nil
+	end
+	if radio then
+		local addr = get_wlan_mac_from_driver(radio, offset)
+		if addr then
+			return addr
+		end
 	end
 
-	return util.generate_mac(4*index + vif)
+	return util.generate_mac(4*index + offset)
 end
 
 -- Iterate over all radios defined in UCI calling
--- a/package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
+++ b/package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
@@ -22,7 +22,7 @@ local function configure_ap(radio, index, config, radio_name)
 
 	uci:delete('wireless', name)
 
-	local macaddr = wireless.get_wlan_mac(radio, index, 0)
+	local macaddr = wireless.get_wlan_mac('client', index, radio)
 
 	if not ap.ssid() or not macaddr then
 		return
@@ -55,7 +55,7 @@ local function configure_owe(radio, index, config, radio_name)
 		return
 	end
 
-	local macaddr = wireless.get_wlan_mac(radio, index, 2)
+	local macaddr = wireless.get_wlan_mac('mesh', index, radio)
 
 	if not ap.owe_ssid() or not macaddr then
 		return
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
@@ -122,7 +122,7 @@ local function configure_mesh(config, radio, index, suffix, disabled)
 		return
 	end
 
-	local macaddr = wireless.get_wlan_mac(radio, index, 1)
+	local macaddr = wireless.get_wlan_mac('mesh', index, radio)
 	if not macaddr then
 		return
 	end
@@ -155,7 +155,7 @@ local function fixup_wan(radio, index)
 		return
 	end
 
-	local macaddr = wireless.get_wlan_mac(radio, index, 3)
+	local macaddr = wireless.get_wlan_mac('wan_radio', index, radio)
 	if not macaddr then
 		return
 	end
--- a/package/gluon-web-private-wifi/luasrc/lib/gluon/config-mode/model/admin/privatewifi.lua
+++ b/package/gluon-web-private-wifi/luasrc/lib/gluon/config-mode/model/admin/privatewifi.lua
@@ -71,7 +71,7 @@ function f:write()
 		local name   = "wan_" .. radio_name
 
 		if enabled.data then
-			local macaddr = wireless.get_wlan_mac(radio, index, 3)
+			local macaddr = wireless.get_wlan_mac('wan_radio', index, radio)
 
 			uci:section('wireless', 'wifi-iface', name, {
 				device     = radio_name,


package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -175,18 +175,20 @@ end
...
+	client = 0,
+	mesh = 1,
+	owe = 2,
+	wan_radio = 3,
+	primary = 3,
+	mesh_other = 4,
+	mesh_vpn = 7,
+}
+
+function M.generate_mac(id, use)
```
neocturne: I would prefer the following signature:
```suggestion
function M.generate_mac(use, radio)
```
where there are separate `interface_ids` definitions for `radio == nil` and `radio ~= nil`. Or maybe separate functions with and without radio argument would be less confusing? Both could internally call the current implementation (which would become private to this module).

I'm also not entirely happy with the name 'use' - maybe 'function' would be better in UCI (shortened to 'func' for the argument name, as 'function' is a keyword in Lua).

maurerle: Thanks, I tried to improve this PR accordingly.

maurerle:
Tested and working fine.
- mesh, client and primary have the right address
- generate_mac works correctly

did not check yet
- private wlan
- lan mesh

So this is waiting for a review. I will squash the commits when reviews are done, to make the history and changes transparent.

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
neocturne: There are still some pieces of code left that are using the name `use`

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
```diff
@@ -175,18 +175,22 @@ end
...
--- 2: owe0
--- 3: wan_radio0 (private WLAN); batman-adv primary address
--- 4: client1; LAN
--- 5: mesh1
--- 6: owe1
--- 7: wan_radio1 (private WLAN); mesh VPN
-function M.generate_mac(i)
-	if i > 7 or i < 0 then return nil end -- max allowed id (0b111)
+
+M.radio_mac_offsets = {
```
neocturne: Hmm, maybe move `radio_mac_offsets` into `wireless.lua`, make both offset tables `local`. `wireless.lua` would be the only place that deals with radio numbers then.

`M.generate_mac` would just have a `func` argument then, `generate_mac_by_idx` is made public again and called both by `generate_mac` and `get_wlan_mac`.

package/gluon-mesh-batman-adv/luasrc/lib/gluon/upgrade/330-gluon-mesh-batman-adv-mac-addresses
```diff
@@ -7,9 +7,9 @@ local uci = require('simple-uci').cursor()
 
 -- fix up potentially duplicate MAC addresses (for meshing)
 if not site.mesh.vxlan(true) then
-	uci:set('network', 'wan', 'macaddr', util.generate_mac(0))
+	uci:set('network', 'wan', 'macaddr', util.generate_mac('wan'))
 end
 if uci:get('network', 'mesh_other') then
-	uci:set('network', 'mesh_other', 'macaddr', util.generate_mac(4))
+	uci:set('network', 'mesh_other', 'macaddr', util.generate_mac('mesh_other'))
```
neocturne: I wonder if setting the MAC address here is even necessary... as far as I can tell, it's always set in `210-interface-mesh`?

maurerle: Same goes for the `generate_mac('wan')` above right?

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
maurerle: Then I would import wireless in gluon_bat0.sh, 330-gluon-mesh-batman-adv-mac-addresses, 500-mesh-vpn and gluon_wired.sh

which do not have anything to do with wireless.. Otherwise this suggestion would be fine to me

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
maurerle: Thanks. Fixed

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
neocturne: Why would you import that? Only `radio_mac_offsets` would be moved, `if_mac_offsets` should remain in `util.lua`.

package/gluon-mesh-batman-adv/luasrc/lib/gluon/upgrade/330-gluon-mesh-batman-adv-mac-addresses
neocturne: No, the setting of the `wan` MAC address is batman-adv-specific and not done in the generic code.

package/gluon-core/luasrc/usr/lib/lua/gluon/util.lua
maurerle: Thanks, totally agree. That's a good refactoring step.

package/gluon-mesh-batman-adv/luasrc/lib/gluon/upgrade/330-gluon-mesh-batman-adv-mac-addresses
maurerle: Okay, I removed setting the mac there - I did not test this yet though.

maurerle:
If this is okay code-wise I will clean up the history into useful chunks.
But I will wait with this to see if there are more changes upcoming :)

maurerle:
Hi @neocturne ,
sorry for the noise/many force-pushes.
I hope that this PR is now ready:

* gluon-core: introduce generate_mac_by_name
* use correct radio and mac index in get_wlan_mac - also removes unused uci param from get_wlan_mac
  * this fixes radio index in `foreach_radio` being of-by-one as well as the mac index in `get_wlan_mac` as they come from a `ipairs` iteration which starts at 1
* gluon-core: update get_wlan_mac to use named mac address

I think this is very clean now and also did test it already.
- [x] interfaces initialize correctly
- [x] mac address of mesh0, mesh1, client0, client1, wan matches the expected values
- [x] mesh on wan works and has correct mac
- [x] br-mesh_other exists - discussion about mac address without vxlan: https://github.com/freifunk-gluon/gluon/pull/3539#discussion_r2203629619

The old way of using `generate_mac` is still compatible, while the docs are updated for new usage with `get_wlan_mac`.

Side note regarding compatibility: the `foreach_radio` index param and `get_wlan_mac` is only used in the community-repo at https://github.com/freifunk-gluon/community-packages/blob/4a90cd221e8eba254ffc8dc5addbf3374b830415/ffgraz-private-ap/luasrc/lib/gluon/upgrade/325-gluon-private-ap which I hope is unused since this feature was upstreamed.

docs/features/private-wlan.rst
```diff
@@ -27,7 +27,7 @@ You may also enable a private WLAN using the command line::
   uci set wireless.wan_radio$RID.ssid="$SSID"
   uci set wireless.wan_radio$RID.key="$KEY"
   uci set wireless.wan_radio$RID.disabled=0
-  uci set wireless.wan_radio$RID.macaddr=$(lua -e "print(require('gluon.util').generate_mac(3+4*$RID))")
+  uci set wireless.wan_radio$RID.macaddr=$(lua -e "print(require('gluon.wireless').get_wlan_mac('wan_radio', $RID)")
```
maurerle:
```suggestion
  uci set wireless.wan_radio$RID.macaddr=$(lua -e "print(require('gluon.wireless').get_wlan_mac('wan_radio', $RID))")
```

package/gluon-mesh-batman-adv/luasrc/lib/gluon/upgrade/330-gluon-mesh-batman-adv-mac-addresses
AiyionPrime: You commented below: "I think this is very clean now and also did test it already.", this change included, right?

maurerle: I just tested a mesh_other interface and found that the previous behavior is required.
Without manually setting the macaddr, the mac of `br-mesh_other` was `d8:84:66:8c:9c:60`.
With setting it to `util.generate_mac('mesh_other')` in `network.mesh_other.macaddr` it is correctly: `9a:5b:5b:3c:9c:5c`.

The reason for this is, that the `json_add_string macaddr` is currently only set in `if [ "$vxlan" -eq 1 ]; then`.

FFAC currently does not have VXLAN set for mesh, so this is not executed.
Should I just readd this line or do you think there is some other workaround? @neocturne

maurerle: If there is only one `mesh_interfaces_other` (e.g. br-mesh_other is not created), there is no iface with the mesh_other mac address at all (except for clien1 of course, which has the same offset).

So I wonder why the correct MAC of br-mesh_other is important, while it is not important when having only one port..?

maurerle: If there is only one `mesh_interfaces_other` (e.g. br-mesh_other is not created), there is no iface with the mesh_other mac address at all (except for clien1 of course, which has the same offset).

So I wonder why the correct MAC of br-mesh_other is important, while it is not important when having only one port..?

package/gluon-core/luasrc/usr/lib/lua/gluon/wireless.lua
```diff
@@ -61,10 +61,11 @@ function M.foreach_radio(uci, f)
 	for index, radio in ipairs(radios) do
 		local band = radio.band
 
+		-- radio index is zero-based
```
neocturne, line 77: A more descriptive commit message would be great - you are not "correcting" the index, you are changing the interpretation of the index from 1-based to 0-based.

package/gluon-core/luasrc/usr/lib/lua/gluon/wireless.lua
```diff
@@ -39,13 +39,26 @@ local function get_wlan_mac_from_driver(radio, vif)
...
+	owe = 2,
+	wan_radio = 3,
+}
+
+function M.get_wlan_mac(func, index, radio)
+	local offset = radio_mac_offsets[func]
+	if offset == nil then
+		return nil
+	end
+	if radio then
```
neocturne, line 54: Why is this `if radio` needed? It wasn't before

package/gluon-mesh-batman-adv/luasrc/lib/gluon/upgrade/330-gluon-mesh-batman-adv-mac-addresses
neocturne: Oh, I appear to have misread the code... yea, it needs to be readded, as one piece of code sets the VXLAN interface address, and one the address of the underlying bridge.

Generally we just want to avoid random (or nondeterministically chosen) MAC addresses being visible anywhere. So if there is no bridge, there already is only a single interface with a fixed address. With the bridge, we want to set an address.

package/gluon-core/luasrc/usr/lib/lua/gluon/wireless.lua
maurerle: The `private-wlan.rst` now uses `$(lua -e "print(require('gluon.wireless').get_wlan_mac('wan_radio', $RID))")`.

If we would not check for radio being nil, this would fail - or we would have to provide an empty table to get_wlan_mac.

I liked to have the user facing call simpler, so I went to make the radio table optional.
This is new, as we called generate_mac before at this place.

package/gluon-core/luasrc/usr/lib/lua/gluon/wireless.lua
maurerle: Done - if there is nothing else, this is ready to merge.

package/gluon-core/luasrc/usr/lib/lua/gluon/wireless.lua
neocturne: > The `private-wlan.rst` now uses `$(lua -e "print(require('gluon.wireless').get_wlan_mac('wan_radio', $RID))")`.
> 
> If we would not check for radio being nil, this would fail - or we would have to provide an empty table to get_wlan_mac.
> 
> I liked to have the user facing call simpler, so I went to make the radio table optional. This is new, as we called generate_mac before at this place.

Mmmh, the old code is simply wrong though, and the new one needs a proper radio. If the WLAN adapter provides multiple MAC addresses, those *must* be used instead of the generated ones (I know at least ralink doesn't allow generated MAC addresses, maybe also (some) Mediatek)

Edit: It would be good to know which targets actually need this - rt305x was dropped in 2021, and I'm not sure which Mediatek chips might share this behavior.

maurerle: So the old code (generating just by entering 4*radio+3) is wrong/incomplete, as it does not respect the wlan adapter MAC.
The new code mimics the same (incomplete) behavior.

So this is only problematic if a user creates a private wifi themselves on a target which does not allow generated MACs.
But should also be generally improved to prefer the the driver macs.

The `D-Link DIR-860L B1 (ramips/mt7621)` does have multiple mac addresses (in `/sys/class/ieee80211/phy0/addresses`) but does not complain using a generated one instead.

I added a fix though in this commit: https://github.com/freifunk-gluon/gluon/commit/cfc825661fc0777d02ae5a65d0752fe394953b54
It introduces a new user facing function which iterates using `foreach_radio` to get the correct radio config to ask the driver as well .
@neocturne

neocturne: Hmm... we're duplicating more and more complex logic, but it would be better to just generate the whole section, using the same code for both the web UI and manually. This could be achieved by moving the actual settings to `/etc/config/gluon` and regenerating the whole of `/etc/config/wireless` on every reconfigure - but that is definitely out of scope for this PR. So maybe we should just keep the incorrect code for now...

AiyionPrime: I'd second that :/

maurerle: Then we are back to the state before my last commit.
https://github.com/freifunk-gluon/gluon/commit/47f176e9d6b846b4558a9076028d1e0a410d9d1a

I will create a draft PR for this additional commit then to track the discussion here, once this is merged.

AiyionPrime: Thanks for your patience :)

PR #3540, opened 2025-07-01
gluon-core: remove check on sysconfig.gluon_version for wireless

maurerle:
The idea for this came up in https://github.com/freifunk-gluon/gluon/pull/3357#discussion_r1989966938

> Checks for sysconfig.gluon_version are almost always wrong. It's better to delete and recreate as much config as possible on each reconfigure instead of trying to handle the first run differently. ~ neocturne

What do you think of this? Especially @neocturne ?

### DIFF
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
@@ -3,13 +3,28 @@
 local util = require 'gluon.util'
 local wireless = require 'gluon.wireless'
 local site = require 'gluon.site'
-local sysconfig = require 'gluon.sysconfig'
 local iwinfo = require 'iwinfo'
 
 local uci = require('simple-uci').cursor()
 
--- Initial
-if not sysconfig.gluon_version then
+local function is_disabled_wireless(name)
+	if uci:get('wireless', name) then
+		return uci:get_bool('wireless', name, 'disabled')
+	end
+
+	return nil
+end
+
+local initial_exists = true
+-- check if initial config exists
+wireless.foreach_radio(uci, function(radio)
+	local radio_name = radio['.name']
+	if is_disabled_wireless('client_' .. radio_name) == nil then
+		initial_exists = false
+	end
+end)
+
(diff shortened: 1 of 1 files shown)


package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -3,15 +3,28 @@
...
 local wireless = require 'gluon.wireless'
 local site = require 'gluon.site'
-local sysconfig = require 'gluon.sysconfig'
 local iwinfo = require 'iwinfo'
 
 local uci = require('simple-uci').cursor()
 
--- Initial
-if not sysconfig.gluon_version then
-	uci:delete_all('wireless', 'wifi-iface')
```
neocturne, line 13: This `delete_all` removes OpenWrt's default wifi-iface, it must not be dropped completely. Maybe that could also be used to find unconfigured radios instead?

maurerle: I like my current approach better, as it does assume less about the existing wifi configuration we might got coming from a preconfigured openwrt?

But I agree on keeping the `delete_all` and readded it.

maurerle:
Tested:
* add default_wifi without existing client_radio config -> gets removed after run of `gluon-reconfigure`
* add default_wifi section with existing client_radio config -> does not get deleted
* have wireless section from openwrt -> correctly migrated
* have empty wireless section -> nothing created, because radio definition missing (just like before)
* wireless section only with defined radio section -> wifi-iface correctly created
* gluon-reconfigure on running node -> no change of wireless config

So this PR is ready as well.

neocturne:
Hmm, I'm undecided if I like this change - maybe it would be better to drop the check at the same time as moving the user intent to `/etc/config/gluon`.

maurerle:
Then I'll just close this PR for now and mention it in #3550 ..?
And hope to take it up again later.

neocturne:
Yes, let's close this in favor of https://github.com/freifunk-gluon/gluon/issues/3550.

PR #3542, opened 2025-07-02
gluon-state-check: Merge with gluon-state-ntpd-check

AiyionPrime:
as requested in #3528


package/gluon-state-check/Makefile
```diff
@@ -20,8 +20,12 @@ define Package/gluon-state-check/description
 
   This package provides the following checks:
   - `has_default_gw6` - check whether the router has a default IPv6-route on br-client.
+  - `has_ntp_sync` - check whether the last stratum event of busybox ntpd was <16
+  - `has_lost_ntp_sync` - check whether the last stratum event of busybox was 16
```
lemoer, line 24: Maybe, we should somehow explain all three possible states? Like what it means when there is none of these two flags available?

lemoer (approved):

package/gluon-state-check/Makefile
neocturne: Hmm, I think we can merge it like this for now and improve with a later iteration.

The ntp check does more than we originally intended for the gluon-state-check package (both in that it tries to track the duration of the state, and that the state is more complex than a simple boolean), and I would like to build a better solution that is not specific to the ntp check.

AiyionPrime: Thanks for the fast responses :)

I think we did indeed intend to use the durations using the different times of the state from the beginning;
but I'm not opposed to finding a better solution later on :)

PR #3543, opened 2025-07-07
docs: Add info about ECDSA keys

### LINKED ISSUE #3247: explain site.conf signature pubkeys in docs
### Bug report

**What is the problem?**

Using the autoupdater feature of Gluon requires specifying pubkeys at build time, but the docs don't explain what kind of pubkeys we're talking about.

This is the stanza in question, in [https://gluon.readthedocs.io/en/latest/user/site.html](https://gluon.readthedocs.io/en/latest/user/site.html):
```
autoupdater = {
    -- Default branch (optional), can be overridden by setting GLUON_AUTOUPDATER_BRANCH when building.
    -- Set GLUON_AUTOUPDATER_ENABLED to enable the autoupdater by default for newly installed nodes.
    branch = 'stable',

    -- List of branches. You may define multiple branches.
    branches = {
      stable = {
        name = 'stable',

        -- List of mirrors to fetch images from. IPv6 required!
        mirrors = {'http://1.updates.services.ffhl/stable/sysupgrade'},

        -- Number of good signatures required.
        -- Have multiple maintainers sign your build and only
        -- accept it when a sufficient number of them have
        -- signed it.
        good_signatures = 2,

        -- List of public keys of maintainers.
        pubkeys = {
          'xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx', -- Alice
          'xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx', -- Bob
          'xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx', -- Mary
        },
      },
    },
``` 
Perusing "site configurations in the wild" give me an inkling that we _might_ be talking about gpg keys, and specifically the fingerprint of the gpg key. I'm not sure at all though, and would like clarification :)
Building gluon using SSH public keys fails for me, at least.

Docs explain clearly what kind of public keys are expected, and links to a guide for how to get/produce it

AiyionPrime:
> and how to use them to sign manifests.

Eventually, this resolves #3247.

### DIFF
--- a/docs/features/autoupdater.rst
+++ b/docs/features/autoupdater.rst
@@ -50,6 +50,32 @@ The header contains the following information:
   the number of days over which the update should be stretched out after ``DATE``. Nodes will calculate a probability
   based on the time left to determine when to update.
 
+Signing images
+--------------
+
+As noted above, manifest files can be signed by an arbitrary amount of ECDSA keys.
+The amount of valid signatures required to have the autoupdater accept an image is configured using the :ref:`site.conf <user-site-autoupdater>`.
+
+A secret key (that like an SSH private key must never be shared) can be generated like this:
+
+.. code-block:: sh
+
+    mkdir ~/.ecdsa/
+    ( umask 077 && ecdsautil generate-key > ~/.ecdsa/id_y25519 )
+
+This is then used to derive a public key, meant to be placed in the ``site.conf``.
+
+.. code-block:: sh
+
+    ( umask 033 && ecdsautil show-key < ~/.ecdsa/id_y25519 > ~/.ecdsa/id_y25519.pub )
+
+A manifest can then be signed using the helper in gluons `contrib` directory.
+
+.. code-block:: sh
+
+    ./contrib/sign.sh ~/.ecdsa/id_y25519 output/images/sysupgrade/stable.manifest
+
+In the manifest file only the content above the three dashes is signed, not other signatures that might exist.
 
 Automated nightly builds
 ------------------------
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -438,6 +438,8 @@ interfaces \: optional
 poe_passthrough \: optional
   Enable PoE passthrough by default on hardware with such a feature.
 
+.. _user-site-autoupdater:
+
 autoupdater \: package
   Configuration for the autoupdater feature of Gluon.
 
@@ -483,6 +485,13 @@ autoupdater \: package
 
   Note that the validity period of TLS certificates is checked as well, so care must be taken
   to provide working NTP servers in addition to the update mirrors when using HTTPS.
+
+  The autoupdater will only accept images signed by as many keys as `good_signatures` specifies.
+
+  The pubkeys section must in turn contain at least that many pubkeys.
+  It's advised to provide more than that to account for human or e.g. hard-drive errors.
+
+  See :doc:`../features/autoupdater` for more information about the process.
 
 .. _user-site-config_mode:
 


docs/features/autoupdater.rst
```diff
@@ -50,6 +50,35 @@ The header contains the following information:
   the number of days over which the update should be stretched out after ``DATE``. Nodes will calculate a probability
   based on the time left to determine when to update.
 
+Signing images
+--------------
+
+As noted above, manifest files can be signed by an arbitrary amount of ECDSA keys.
+The amount of valid signatures required to have the autoupdater acknowledge an image is configured using the ``site.conf``.
```
herbetom:
```suggestion
The amount of valid signatures required to have the autoupdater accept an image is configured using the ``site.conf``.
```

Same term is used here https://gluon.readthedocs.io/en/latest/user/site.html#configuration

But no strong opinion

Also it might make sense to actually link to that page instead of just saying `site.conf`?

herbetom (commented):
Thanks for following up on that issue and creating a PR! :)

docs/features/autoupdater.rst
AiyionPrime: resolved in af3723a2771ffd0e9d4d9de9f7507d7ab5bae060, thanks.

docs/features/autoupdater.rst
```diff
@@ -50,6 +50,35 @@ The header contains the following information:
...
+    umask 077
+    mkdir ~/.ecdsa/
+    ecdsautil generate-key > ~/.ecdsa/id_y25519
+
+This is then used to derive a public key, meant to be placed in the ``site.conf``.
+
+.. code-block:: sh
+
+    cd ~/.ecdsa/
+    umask 033
```
neocturne: Changing the umask here doesn't make much sense if `~/.ecdsa` has been created with `077` before. Maybe just set the umask for the `generate-key` command itself using a `( ... )` subshell?

docs/features/autoupdater.rst
```diff
@@ -50,6 +50,35 @@ The header contains the following information:
...
+
+    umask 077
+    mkdir ~/.ecdsa/
+    ecdsautil generate-key > ~/.ecdsa/id_y25519
+
+This is then used to derive a public key, meant to be placed in the ``site.conf``.
+
+.. code-block:: sh
+
+    cd ~/.ecdsa/
```
neocturne: The previous code block does not `cd`, this one does. I would keep it consistent (probably removing the `cd` here in favor of writing the full patch twice)

docs/features/autoupdater.rst
AiyionPrime: Mh. Unsure about this.
I tried to replicate the permissions of the `.ssh/` directory.
So far most distros I encountered use `drwx-----`.

The only exception I found is one endeavourOS setup.

Your choice.

docs/features/autoupdater.rst
AiyionPrime: resolved 104917a50b9113eb358d52ea6858ae6549fee7cf

docs/features/autoupdater.rst
AiyionPrime: resolved 104917a50b9113eb358d52ea6858ae6549fee7cf

PR #3549, opened 2025-07-21

herbetom:


herbetom:
Dropped `patches/openwrt/0009-mt76-import-MT7915-recovery-fixes.patch` since those three patches have been merged to the OpenWrt openwrt-24.10 branch:

https://github.com/openwrt/openwrt/commit/ad765d3cc4035406a4cd748f3339301ecc19dba8 (via https://github.com/openwrt/openwrt/pull/19239)

PR #3551, opened 2025-07-22
gluon-core: gluon.util: extend get_role_interfaces() to retrieve additional fields

neocturne:
Add an argument to get_role_interfaces() to optionally pass in a list of extra fields to obtain.

For a UCI config like

    config interface 'iface_eth0'
        option name 'eth0'
        list role 'mesh'

    config interface 'iface_eth1'
        option name 'eth1'
        list role 'mesh'
        option batadv_hop_penalty '10'

a regular call of `util.get_role_interfaces(uci, 'mesh')` would still return a list like the following:

    {
        'eth0',
        'eth1',
    }

`util.get_role_interfaces(uci, 'mesh', nil, {'batadv_hop_penalty'})` can be used to get a result like the following instead:

    {
        ['eth0'] = {},
        ['eth1'] = {['batadv_hop_penalty'] = '10'},
    }

---

My idea is that this could be used to implement https://github.com/freifunk-gluon/gluon/pull/3454#issuecomment-3063349232 - @maurerle


neocturne:
Yea, I agree that the differing return type is a bit iffy. Maybe we should just make it two separate functions.

neocturne:
@maurerle `ipairs` will not loop through anything with the "extended" variant, because there are no integer keys.

neocturne:
Updated to use a separate function.

maurerle:
This works fine with

```
config interface 'iface_lan'
	option name '/lan'
	list role 'mesh'
        option batadv_hop_penalty '1'
```

the following script

```lua
local ifaces_options = util.get_role_interfaces_with_options(uci, "mesh", {'batadv_hop_penalty'}, nil)
local ifaces = util.get_role_interfaces(uci, "mesh")

print("ifaces options", ifaces_options)
for idx, iface in pairs(ifaces_options) do
    print(idx, iface)
    for idx, option in pairs(iface) do
        print(idx, option)
    end
end

print()
print("ifaces", ifaces)
for idx, iface in ipairs(ifaces) do
    print(idx, iface)
end
```

prints

```
ifaces options	table: 0xb6eaada0
lan	table: 0xb6eabbb0
batadv_hop_penalty	1

ifaces	table: 0xb6eabee0
1	lan

```

AiyionPrime (commented):
forgot to send the approval tonight. Sorry.

PR #3554, opened 2025-07-26
status-page: Emit VPN pubkey

AiyionPrime:
This is the result:

Screenshot 2025-07-26 at 16-51-26 x86-aiyion - Status

### DIFF
--- a/package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn.lua
+++ b/package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn.lua
@@ -44,4 +44,19 @@ function M.get_active_provider()
 	return nil, nil
 end
 
+function M.get_enabled_public_key()
+	if M.enabled() ~= true then
+		return nil
+	end
+
+	local _, active_vpn = M.get_active_provider()
+
+	local pubkey
+	if active_vpn ~= nil then
+		pubkey = active_vpn.public_key()
+	end
+
+	return pubkey
+end
+
 return M
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/info.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/info.lua
@@ -9,12 +9,8 @@ local has_vpn, vpn = pcall(require, 'gluon.mesh-vpn')
 local ethernet = require 'gluon.ethernet'
 
 local pubkey
-if has_vpn and vpn.enabled() then
-	local _, active_vpn = vpn.get_active_provider()
-
-	if active_vpn ~= nil then
-		pubkey = active_vpn.public_key()
-	end
+if has_vpn then
+	pubkey = vpn.get_enabled_public_key()
 end
 
 local M = {}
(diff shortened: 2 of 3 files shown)


AiyionPrime:
@neocturne I see two possible extensions.

One is a setup-mode script, which is pretty similar.
The other is to take the configuration `pubkey_privacy` from the site into account.

herbetom:
> The other is to take the configuration pubkey_privacy from the site into account.

This seems more like a must have to me. :see_no_evil: 

---

I'm not so sure wheter this really is something which has to be on the status page? For most people this will be just a bunch of random characters without any explanation.

Some may even wonder wheter their node, browser or computer was "hacked" or there is some other corruption ...

If you need it getting it via either respondd or a http request to `/cgi-bin/dyn/neighbours-nodeinfo?mesh-vpn` seems at first glance "enough"? :see_no_evil:

SmithChart:
In our *ffbs-mesh-vpn-parker* the WireGuard pubkey is used as the device-id during the configuration phase and in the backend data structure.
One could, for example, want to pin a specific node to a concentrator. In that case the user needs to be able to tell the devices WireGuard pubkey.

AiyionPrime:
@herbetom Thanks for raising your concerns.
If we find it irritates users more than the pubkey emitted in setup mode, we could add additional information.

For starters I'll start evaluating whether this privacy feature itself is helpful beyond communities using fastd.
If I read the WireGuard paper correctly, it is not.

T-X:
> In our _ffbs-mesh-vpn-parker_ the WireGuard pubkey is used as the device-id during the configuration phase and in the backend data structure. One could, for example, want to pin a specific node to a concentrator. In that case the user needs to be able to tell the devices WireGuard pubkey.

What's the advantage of using the WireGuard pubkey instead of node-id for this? Wouldn't the latter be more protocol agnostic? Is Parker currently incompatible with fastd? (though maybe this is getting a bit off-topic for this ticket)

AiyionPrime:
> If I read the WireGuard paper correctly, it is not.

I followed up with more research and find it really isn't useful for WireGuard
and opened #3578 to address the issue.

AiyionPrime:
This does now respect `pubkey_privacy`.

jluebbe:
> What's the advantage of using the WireGuard pubkey instead of node-id for this? Wouldn't the latter be more protocol agnostic?

As wireguard binds each peer's identity to its pubkey, it can't be impersonated by another node. We accept/reject nodes based on their wireguard pubkeys, so it's very useful to show them on the status page.

> Is Parker currently incompatible with fastd? (though maybe this is getting a bit off-topic for this ticket)

Parker uses wireguard instead of fastd.

@AiyionPrime Perhaps prefix the key by `public key: `, to avoid possible user confusion?

AiyionPrime:
Looking at the other fields, we could append "` (public key)`", to match their appearance.

package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn.lua
```diff
@@ -44,4 +44,19 @@ function M.get_active_provider()
 	return nil, nil
 end
 
+function M.get_enabled_public_key()
+	if M.enabled() ~= true or uci:get_bool('gluon', 'mesh_vpn', 'pubkey_privacy') ~= true then
```
neocturne, line 48: This code is incorrect. `pubkey_privacy` is a site field, it never exists in UCI.

neocturne: https://github.com/freifunk-gluon/gluon/issues/3624

PR #3556, opened 2025-07-27
gluon-mesh-batman-adv: improve has_default_gw4 for parker

maurerle:
if we have batctl gw_mode server configured, we also have a `default_gw` for it.

This does not occur for traditional setups and therefore does not change existing behavior

The motivation for this is the FFBS setup, which otherwise does not have a working `has_default_gw4`.

@SmithChart

### DIFF
--- a/package/gluon-mesh-batman-adv/files/lib/gluon/state/check.d/has_default_gw4
+++ b/package/gluon-mesh-batman-adv/files/lib/gluon/state/check.d/has_default_gw4
@@ -1,2 +1,19 @@
 #!/bin/sh
-out=$(batctl gwl -H 2>/dev/null) && [ -n "$out" ]
+
+# shellcheck disable=SC3040
+set -o pipefail
+
+if ! out=$(batctl gwl -H 2>/dev/null); then
+	# batctl gwl command failed
+	exit 1
+fi
+
+if [ -n "$out" ]; then
+	# no batctl gw available
+	exit 2
+fi
+
+if ! batctl gw_mode | grep -qw '^server'; then
+	# server is also not in gw_mode
+	exit 3
+fi


SmithChart:
On parker, I see the following behavior:

For a node that is `batman gateway mode: server`:
```
root@cir6120:/# /lib/gluon/state/check.d/has_default_gw4; echo $?
0
```

For a node that is `batman gateway mode: client`:
```
root@cir6120:/# /lib/gluon/state/check.d/has_default_gw4; echo $?
3
```

Not sure what your intended behavior is :-)

maurerle:
Hi @SmithChart ,

if the node has a default_gw4 - the script should exit 0 - which will create the `/var/gluon/state/has_default_gw4`.
In other cases, the file is not created/removed.

That way, other packages can determine if a connection is established (e.g. ffac-ssid-changer).

SmithChart:
> Hi @SmithChart ,
> 
> if the node has a default_gw4 - the script should exit 0 - which will create the `/var/gluon/state/has_default_gw4`. In other cases, the file is not created/removed.

I am still not sure about the specification here. The script does not touch nor remove files.
I guess I am missing the big picture here.

maurerle:
The missing link is the gluon-state-check:
https://github.com/freifunk-gluon/gluon/blob/main/package/gluon-state-check/luasrc/usr/sbin/gluon-state-check
This creates or removes the files for check scripts in /lib/gluon/state/check.d

Further documentation about the gluon-state-check will be added in #3558

package/gluon-mesh-batman-adv/files/lib/gluon/state/check.d/has_default_gw4
```diff
@@ -1,2 +1,17 @@
-#!/bin/sh
-out=$(batctl gwl -H 2>/dev/null) && [ -n "$out" ]
+#!/bin/busybox sh
```
neocturne: We don't use this shebang anywhere else in Gluon.

package/gluon-mesh-batman-adv/files/lib/gluon/state/check.d/has_default_gw4
```diff
@@ -1,2 +1,17 @@
...
+	# batctl gwl command failed
+	exit 1
+fi
+
+if [[ -n "$out" ]]; then
+	# no batctl gw available
+	exit 2
+fi
+
+if ! batctl gw_mode | grep -q server; then
```
neocturne: How exactly does the output of `batctl gw_mode` look in server mode? Should the grep be more specific?

package/gluon-mesh-batman-adv/files/lib/gluon/state/check.d/has_default_gw4
```diff
@@ -1,2 +1,17 @@
...
-out=$(batctl gwl -H 2>/dev/null) && [ -n "$out" ]
+#!/bin/busybox sh
+set -o pipefail
+
+if ! out=$(batctl gwl -H 2>/dev/null); then
+	# batctl gwl command failed
+	exit 1
+fi
+
+if [[ -n "$out" ]]; then
```
neocturne: Is there any reason why you switched from [ to [[?

package/gluon-mesh-batman-adv/files/lib/gluon/state/check.d/has_default_gw4
maurerle: It looks like this:
```
> batctl gw_mode
server (announced bw: 10.0/2.0 MBit)
```

I don't think that it should be more specific, what do you think?

vs client:

```
> batctl gw_mode
client (selection class: 20)
```

neocturne: I would use `grep -qw '^server'`.

PR #3557, opened 2025-07-27
openwrt hostapd: disable WPS button handling

blocktrron:
Pressing the WPS button when wpa-supplicant is installed for encrypted mesh operation triggers WPS in the wpa_supplicant process.

This results in a deactivation of the wireless mesh.

Disabling the WPS handling in OpenWrt to keep wireless mesh operational even when pressing the WPS key.


AiyionPrime (approved):
Unsure whther there are any, but this does not impose problems on devices, which use the button as reset button, right?

(Thinking of DECT buttons on FB or similar)

blocktrron:
Gluon Setup Mode handling of the WPS button is working fine with this patch.

AiyionPrime (approved):
Thanks for the follow up.
That's currently all I can think of :)

PR #3560, opened 2025-07-31
openwrt: ath9k: Add RX inactivity detection and reset chip when it occurs

citronalco:
Backport of this patch:
https://github.com/torvalds/linux/commit/b5f871ab4913b2403a7cdcbcde16d39d0b071fb3


citronalco:
I'm not sure if my backport of this patch fixes the issue with the deaf wifi interface.

PR #3356, opened 2024-10-13
Update filters.yml >>>adding ipq807x-generic to support Xiaomi AX3600 belongs to #3354

ChristianMiddendorf:
The added target ipq807x-generic is in the first step used for the Xiaomi AX3600.

In the future it maybe support Xiaomi AX9000 as well.

PR #3354, opened 2024-10-12
Adding new target ipq807x-generic and new device Xiaomi AX3600

ChristianMiddendorf:
- [ ] Must be flashable from vendor firmware
  - [ ] Web interface
  - [ ] TFTP
  - [X] Other: <supgrade from OpenWRT but do not save configuration over reflash... "sysupgrade -n filenamefromsysupgrade.bin">
- [X] Must support upgrade mechanism
  - [X] Must have working sysupgrade
    - [X] Must keep/forget configuration (`sysupgrade [-n]`, `firstboot`)
  - [X] Gluon profile name matches autoupdater image name
        (`lua -e 'print(require("platform_info").get_image_name())'`) 
- [X] Reset/WPS/... button must return device into config mode
- [X] Primary MAC address should match address on device label (or packaging)
      (https://gluon.readthedocs.io/en/latest/dev/hardware.html#hardware-support-in-packages)
  - When re-adding a device that was supported by an earlier version of Gluon, a
    factory reset must be performed before checking the primary MAC address, as
    the setting from the old version is not reset otherwise.
- Wired network
  - [ ] should support all network ports on the device
  - [X] must have correct port assignment (WAN/LAN)
    - if there are multiple ports but no WAN port:
      - the PoE input should be WAN, all other ports LAN
      - otherwise the first port should be declared as WAN, all other ports LAN
- Wireless network (if applicable)
  - [X] Association with AP must be possible on all radios
  - [X] Association with 802.11s mesh must work on all radios 
  - [X] AP+mesh mode must work in parallel on all radios
- LED mapping
  - Power/system LED
    - [X] Lit while the device is on
    - [X] Should display config mode blink sequence 
          (https://gluon.readthedocs.io/en/latest/features/configmode.html)
  - Radio LEDs
    - [ ] Should map to their respective radio
    - [X] Should show activity
  - Switch port LEDs
    - [ ] Should map to their respective port (or switch, if only one led present) 
    - [ ] Should show link state and activity
- Outdoor devices only:
  - [ ] Added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`
- Cellular devices only:
  - [ ] Added board name to `is_cellular_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`
  - [ ] Added board name with modem setup function `setup_ncm_qmi` to `package/gluon-core/luasrc/lib/gluon/upgrade/250-cellular`
- Docs:
[...]

PR #3561, opened 2025-07-31
qualcommax-ipq807x: Add Xiaomi AX3600

AiyionPrime:
>  as supported despite the lack of the disabled third radio (QCA9889 1×1
>  ac/n) due to explicit maintainer wish.

  - ~Web interface~
  - ~TFTP~
  - [x] Other:
    1. https://openwrt.org/toh/xiaomi/ax3600#obtain_ssh_access
    2. https://openwrt.org/toh/xiaomi/ax3600#install_the_firmware
        (`lua -e 'print(require("platform_info").get_image_name())'`) **xiaomi-ax3600**
- Wired network
  - [x] AP+mesh mode must work in parallel on all radios **the two enabled ones do**
  *the one that has been disabled does not*
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [x] Should show activity
  - Switch port LEDs
- Outdoor devices only:
  - ~Added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`~
- Cellular devices only:
  - ~Added board name to `is_cellular_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`~
  - ~Added board name with modem setup function `setup_ncm_qmi` to `package/gluon-core/luasrc/lib/gluon/upgrade/250-cellular`~
- Docs:
  - [x] Added Device to `docs/user/supported_devices.rst` **despite not supported due to third radio being unsupported**


AiyionPrime:
@cawi2001 lent me a device to perform the checks on.
I expect everything to still work fine, but the target has changed in the meantime.
The third radio is disabled and as 1x1 it isn't worth being added imho, anyway.

I'll follow up on this on Saturday.

AiyionPrime:
Previous iterations of this were:
- #3032
- #3344
- #3347 
- #3354 
- #3356

rotanid:
@AiyionPrime hm, it is marked as BROKEN only because of a stupid third radio we don't want to support anyway?

AiyionPrime:
Not sure whether we would not support it, but it appears to be insuffcient for the two classical use-cases, client and mesh.
It could be used for low-bandwidth devices though.

I think the current decision is to follow the checklist strictly -.-' .
Once we resolve our pending tri-radio problem, I'd gladly re-iterate this and mark it as non broken.

AiyionPrime:
@rotanid I hope you don't mind me assigning this to you.
This is mostly about the decision like I wrote on IRC.

Whether we want to add it as supported or broken is up to you.
I'd be happy either way - just let me know what you'd favor.

rotanid:
i'd say it's a corner case... i would say merge without broken, but i'd like to have a 2nd opinion by another maintainer.

targets/qualcommax-ipq807x
```diff
@@ -0,0 +1,13 @@
+-- Xiaomi
+
+device('xiaomi-ax3600', 'xiaomi_ax3600', {
+	-- third radio is not supported
+	broken = true,
+	factory = false,
+	packages = {
+		'-ath10k-firmware-qca9887-ct',
+		'-ath10k-board-qca9887',
```
lemoer:
```suggestion
		'-ath10k-firmware-qca9887-ct', -- block the 3rd "iot" radio, which is unsuitable for gluon
		'-ath10k-board-qca9887',       -- block the 3rd "iot" radio, which is unsuitable for gluon
```

lemoer:
I'd say, let's merge this as non-broken.

So far, it's unclear if the "iot" interface could be useful with triband support. If we have triband support merged and somebody finds out, that the third interface is actually useful, we can reenable it. However, until proven otherwise, I would stick with the theory that the interface would not be useful for gluon.

targets/qualcommax-ipq807x
AiyionPrime: resolved in 247803ac3ef5044fe86898e6fd653b6e48acc452

PR #3562, opened 2025-08-11
cortexa53: add support for GL.iNet GL-MV1000

T0biii:
We recently got an request to create a gluon firmware for this device from a node owner of our community. the node owner will test the whole checklist as soon as possible

Test Device: https://map.ffmuc.net/#!/de/map/9483c400f884
OpenWrt sites: 
https://openwrt.org/toh/hwdata/gl.inet/gl.inet_gl-mv1000_brume
https://openwrt.org/toh/gl.inet/gl-mv1000

  - [ ] Web interface
  - [ ] TFTP
  - [x] Other: Web interface: 
	- https://forum.gl-inet.com/t/installing-openwrt-23-05-on-brume-1/33477 			
	- https://openwrt.org/toh/gl.inet/gl-mv1000#uboot_method
	- [x] sysupgrade keeps configuration
	- firstboot
image

        (`lua -e 'print(require("platform_info").get_image_name())'`) -> `gl.inet-gl-mv1000`

image

- Wired network
- ~~Wireless network (if applicable)~~ we dont got a GL-MV1000W so no Wireless Features can be tested
  - [ ] ~~Association with AP must be possible on all radios~~
  - [ ] ~~Association with 802.11s mesh must work on all radios~~
  - [ ] ~~AP+mesh mode must work in parallel on all radios~~
- LED mapping
  - Power/system LED
  - ~~Radio LEDs~~
  - Switch port LEDs
- ~~Outdoor devices only:~~
- ~~Cellular devices only:~~
- Docs:

https://github.com/freifunkMUC/site-ffm/tree/add-cortexa53

### DIFF
--- a/targets/mvebu-cortexa53
+++ b/targets/mvebu-cortexa53
@@ -1,4 +1,4 @@
-device('glinet-gl-mv1000', 'glinet_gl-mv1000', {
+device('gl.inet-gl-mv1000', 'glinet_gl-mv1000', {
 	factory = false,
 	sysupgrade = '-ext4-sdcard',
 	sysupgrade_ext = '.img.gz',
--- a/targets/mvebu-cortexa53
+++ b/targets/mvebu-cortexa53
@@ -1,5 +1,5 @@
 device('gl.inet-gl-mv1000', 'glinet_gl-mv1000', {
 	factory = false,
-	sysupgrade = '-ext4-sdcard',
+	sysupgrade = '-squashfs-sdcard',
 	sysupgrade_ext = '.img.gz',
 })
--- a/targets/mvebu-cortexa53
+++ b/targets/mvebu-cortexa53
@@ -1,3 +1,6 @@
 device('glinet-gl-mv1000', 'glinet_gl-mv1000', {
-	factory_ext = '.img',
+	factory = false,
+	sysupgrade = '-ext4-sdcard',
+	sysupgrade_ext = '.img.gz',
+	manifest_aliases = {'gl-inet-gl-mv1000'},
 })
--- a/targets/mvebu-cortexa53
+++ b/targets/mvebu-cortexa53
@@ -2,5 +2,4 @@ device('glinet-gl-mv1000', 'glinet_gl-mv1000', {
 	factory = false,
 	sysupgrade = '-ext4-sdcard',
 	sysupgrade_ext = '.img.gz',
-	manifest_aliases = {'gl-inet-gl-mv1000'},
 })
--- a/targets/mvebu-cortexa53
+++ b/targets/mvebu-cortexa53
@@ -1,0 +1,3 @@
+device('glinet-gl-mv1000', 'glinet_gl-mv1000', {
+	factory_ext = '.img',
+})
--- a/targets/targets.mk
+++ b/targets/targets.mk
@@ -32,6 +32,5 @@ $(eval $(call GluonTarget,bcm27xx,bcm2711)) # BROKEN: No 11s support, no reset b
 $(eval $(call GluonTarget,ipq40xx,chromium)) # BROKEN: Devices cannot be flashed without opening the case
 $(eval $(call GluonTarget,kirkwood,generic)) # BROKEN: No devices with 11s support
 $(eval $(call GluonTarget,mvebu,cortexa9)) # BROKEN: No 11s support
-$(eval $(call GluonTarget,mvebu,cortexa53)) # BROKEN:
-
+$(eval $(call GluonTarget,mvebu,cortexa53)) # BROKEN: Untested
 endif
--- a/targets/targets.mk
+++ b/targets/targets.mk
@@ -13,6 +13,7 @@ $(eval $(call GluonTarget,lantiq,xrx200_legacy))
 $(eval $(call GluonTarget,lantiq,xway))
 $(eval $(call GluonTarget,mediatek,filogic))
 $(eval $(call GluonTarget,mediatek,mt7622))
+$(eval $(call GluonTarget,mvebu,cortexa53))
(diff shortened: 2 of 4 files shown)


T0biii:
Most of it works. What we can't test is wireless as we dont got a `GL-MV1000W`.
firstboot doesn't work. Do you have any idea what must be done to fix this:
image

neocturne:
> Most of it works. What we can't test is wireless as we dont got a `GL-MV1000W`. firstboot doesn't work. Do you have any idea what must be done to fix this: image

firstboot requires an overlay, which `ext4` images generally don't provide; `sunxi-cortexa7` likely has the same problem. This does not seem critical to me and shouldn't prevent the device from being considered "supported".

herbetom:
Do we need to use the ext4 image? It looks like squashfs is also an option and probably would make more sense?

targets/mvebu-cortexa53
```diff
@@ -0,0 +1,5 @@
+device('gl.inet-gl-mv1000', 'glinet_gl-mv1000', {
+	factory = false,
+	sysupgrade = '-ext4-sdcard',
```
T0biii: @herbetom do you mean like so?

```suggestion
	sysupgrade = '-squashfs-sdcard',
```

was me first time creating a new image for the community for us so therefore, I am grateful for any suggestions for improvement

herbetom: yes, no worries :)

Obviously needs to be tested but if that works it's probably the way to go.

T0biii: as soon as the image build i guess the person in our community can test the build again

neocturne:
> Do we need to use the ext4 image? It looks like squashfs is also an option and probably would make more sense?

You are right, I missed that in the code. In that case squashfs is definitely preferable, and we should migrate `sunxi-cortexa7` as well.

T0biii:
with the new image the Problem with firstboot is gone :)

image

~~the rest will probably be tested this morning~~
the rest got tested and worked as bevor

targets/targets.mk
```diff
@@ -32,4 +32,5 @@ $(eval $(call GluonTarget,bcm27xx,bcm2711)) # BROKEN: No 11s support, no reset b
 $(eval $(call GluonTarget,ipq40xx,chromium)) # BROKEN: Devices cannot be flashed without opening the case
 $(eval $(call GluonTarget,kirkwood,generic)) # BROKEN: No devices with 11s support
 $(eval $(call GluonTarget,mvebu,cortexa9)) # BROKEN: No 11s support
+$(eval $(call GluonTarget,mvebu,cortexa53)) # BROKEN: Wireless untested
```
herbetom: Since the device in question doesn't have Wireless capabilities and the checklist looks fine to me, I would argue the broken classification isn't necessary here.
The requirement is simply that if a device has those capabilities, then it has to work. That way no false expectations are created. However, if it doesn't, that is completely fine.

If it turns out that WLAN is a problem with that target it should probably be flagged for the device in question and not the whole target.

So if you would like to add it as non-broken just move it in the correct alpabetically place in the list above and then create the docs entry.

T0biii: The reason I’m thinking about it is that there’s another device that also has Wi-Fi and the same image.

the: GL-MV1000W

https://openwrt.org/toh/gl.inet/gl-mv1000

> The Brume-W, GL-MV1000W, is the same hardware plus a WLAN card, and will run on the same image. It contains an additional Wireless USB chip. Please see the section wireless for details.

Djfe: yeah but the image you generated isn't called GL-MV1000W.
So MV1000W is untested. Yes, you could run the image of the MV1000 on it, but you would need to install packages manually for wifi to work. effectively it's the wrong image and a user error in this case.
OpenWrt would need a new dts and image/*.mk entry for the wireless entry anyways imo. (that way wifi works out of the box)

Marking the target as broken or this particular device doesn't make any sense :)

ambassador86:
Hello @T0biii,
I recently got a GL-MV1000W. I will build an image for my community this weekend based on your commit and report if wifi works as expected.

T0biii:
> Hello @T0biii, I recently got a GL-MV1000W. I will build an image for my community this weekend based on your commit and report if wifi works as expected.

when i understood @Djfe correctly the GL-MV1000W needs his own dts and image/*.mk in openwrt so that wifi can work out of the box 

for the GL-MV1000 i think this pr ready

ambassador86:
I can confirm, that the wifi driver is missing in this image and the radio is not recognized for the "GL-MV1000W". After building another image with the drivers included the device does not meet the criteria for "non-broken" devices anymore.

1. The mac address from the wifi interface becomes the primary mac and does not match the adress printed on the case.
2. The wifi is not able to work in 'mesh point' mode
3. The wifi is not able to create multiple APs

Overall I would strongly suggest to merge this PR as is. Go on with the suggestion from @Djfe to create another dts file for the version with wifi. But the the device with wifi is not supportable for gluon with the current driver.

My test device without wifi: https://map.chemnitz.freifunk.net/#!/de/map/9483c41005df
My test device after including wifi: https://map.chemnitz.freifunk.net/#!/de/map/d4d2d667ce40

herbetom:
Thanks, Good to know :)

(JFTR: as long as the wifi mac is printed on the device itself that isn't really an issue in itself but it's also something that could just be overriden: [package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac#L72-L157](https://github.com/freifunk-gluon/gluon/blob/6d3ceffce2fd866dc1b71f65ff7151b3720ffdc8/package/gluon-core/luasrc/lib/gluon/upgrade/010-primary-mac#L72-L157).
But the wifi does indeed sound broken and not fit for a non broken device. Creating a seperate DTS does sound like a good plan to me.)

Issue #3550, opened 2025-07-22
RFC: configure wifi through gluon config


maurerle:
A thing which comes up when changing wifi related code a lot is the complete recreation of the `uci show wireless` section.

Instead, one should be able to configure everything with `/etc/config/gluon` alone.

Currently, we already have

```
gluon.wireless=wireless
gluon.wireless.outdoor='0'
gluon.wireless.preserve_channels='1'
```

## Requirements

- [ ] set htmode #2947 
- [ ] set txpower in `/etc/config/gluon`
- [x] allow additional interface like private-wlan
- [x] set hop_penalty per wireless
- [x] some option to persist custom wireless iface sections?
- [ ] allow configuration overrides per radio

## Concept

```
config core 'core'

config interface 'iface_wan'
	option name '/wan'
	list role 'uplink'

config interface 'iface_lan'
	option name '/lan'
	list role 'client'

config mesh_vpn 'mesh_vpn'
	option enabled '1'
	option limit_enabled '0'

config wireless_band '2g'
	option enabled '1'
	list role 'client'
	list role 'mesh'
	list role 'private'


config wireless_band '5g'
	option enabled '1'
	list role 'client'
	list role 'mesh'
	option htmode '80'

config wireless 'wireless'
	option outdoor '0'
	option preserve_channels '0'
	option private_ssid 'MyWireless'
	option private_key 'mypassword'
	option private_encryption 'psk2'
	option private_mfp '0'
```

The above has the disadvantage of having multiple information in one setting.
Furthermore, one does not have an option to configure the htmode/txpower of multiple 5ghz radios differently.

## Motivation

Currently, this idea has been mentioned in

* https://github.com/freifunk-gluon/gluon/pull/3454#issuecomment-2741101763
    > Moving this to /etc/config/gluon would definitely make sense once we start resetting /etc/config/wireless completely on reconfigure - which would be nice, but out of scope for this PR
* https://github.com/freifunk-gluon/gluon/pull/3539#discussion_r2214162134
    > Hmm... we're duplicating more and more complex logic, but it would be better to just generate the whole section, using the same code for both the web UI and manually. This could be achieved by moving the actual settings to /etc/config/gluon and regenerating the whole of /etc/config/wireless on every reconfigure - but that is definitely out of scope for this PR. So maybe we should just keep the incorrect code for now...

* https://github.com/freifunk-gluon/gluon/pull/3357#issuecomment-3102311884
[...]

PR #3563, opened 2025-08-16
gluon-core: reconfigure wireless config to support triple radio configs

maurerle:
This implemens the concept proposed in #3550

Supersedes #3357

It includes configuration of a default roles section, as well as a wireless configuration.
This solves the limitations of the current implementation, mentioned in #3550 quite well and also adds support for devices with multiple radios on the same band.

```
config wireless_band '2g'
	list role 'client'
	list role 'mesh'
	list role 'private'

config wireless_band '5g'
	list role 'client'
	list role 'mesh'

config wireless 'wireless'
	option outdoor '0'
	option preserve_channels '0'
	option private_ssid 'MyWireless'
	option private_key 'mypassword'
	option private_encryption 'psk2'
	option private_mfp '2'
	option htmode_radio1 'VHT80'
	option txpower_radio1 '11'

```

The settings from the gluon section are written to the wireless section.
The wireless interface section is recreated with every gluon-reconfigure (the radio configuration is kept).

Having multiple radios on the same band automatically configures one of them mesh-only if there is at least one client only radio (and both, mesh and client role are selected).

**TODO**
- [x] migration if config does not yet exist
- [x] adjustments to private wifi
- [x] adjustments to client bridge 
- [x] adjustments to `gluon-web-wifi-config`
- [x] addition of `htmode` and per radio settings (using `gluon.wireless.htmode_radio1`)
- [ ] persist txpower per radio (was available in config-mode before) - now in `gluon.wireless.txpower_radio1`)
- [x] allow setting `gluon_preserve` in the wireless section for additional radios
- [x] testing
  - [x] new install
  - [x] upgrade
  - [x] changes to config from config-mode or `uci set gluon...`
  - [x] `preserve_channels` still functional, per radio settings are not touched

### DIFF
--- a/package/gluon-core/files/lib/gluon/upgrade/001-reset-uci
+++ b/package/gluon-core/files/lib/gluon/upgrade/001-reset-uci
@@ -6,6 +6,9 @@ NETWORK_SAVED="${NETWORK_CFG}_gluon-old"
 SYSTEM_CFG='/etc/config/system'
 SYSTEM_SAVED="${SYSTEM_CFG}_gluon-old"
 
+WIRELESS_CFG='/etc/config/wireless'
+WIRELESS_SAVED="${WIRELESS_CFG}_gluon-old"
+
 # Make sure everything is saved before we move away the config files
 uci commit
 
@@ -16,6 +19,10 @@ if [ -s "$NETWORK_CFG" ] && ! [ -s "$NETWORK_SAVED" ]; then
 fi
 if [ -s "$SYSTEM_CFG" ] && ! [ -s "$SYSTEM_SAVED" ]; then
 	mv -f "$SYSTEM_CFG" "$SYSTEM_SAVED"
+fi
+if [ -s "$WIRELESS_CFG" ] && ! [ -s "$WIRELESS_SAVED" ]; then
+	# TODO wireless should be moved as well after migrating radio specific settings
+	cp "$WIRELESS_CFG" "$WIRELESS_SAVED"
 fi
 
 # Generate a new network config
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/997-migrate-preserved
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/997-migrate-preserved
@@ -2,7 +2,7 @@
 
 local uci = require('simple-uci').cursor()
 
-for _, config in ipairs({'system', 'network'}) do
+for _, config in ipairs({'system', 'network', 'wireless'}) do
 	uci:foreach(config .. '_gluon-old', nil, function(s)
 		if s.gluon_preserve ~= '1' then return end
 


maurerle:
This is now ready for review.
I tested this and it works quite well.

Here is a screenshot of the adjustments in the wifi config:

Screenshot From 2025-08-19 22-56-39

private-wifi can be configured independently as usual, if the package is installed.


**The addition of settings for htmode on a per radio basis will be added in the future and is not part of this PR.**

maurerle:
I fixed a config check in the 022-wireless-roles to correctly generate the new config from the existing config in my latest force-push.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -12,19 +12,24 @@ local uci = require('simple-uci').cursor()
 if not sysconfig.gluon_version then
 	uci:delete_all('wireless', 'wifi-iface')
 
+	local radio_band_count = {band2g=0, band5g=0}
```
neocturne: Commit message: "improve" is very unspecific, and I'd usually expect some functional change when I read it. I would use "refactor" or "clean up" for such a style change.

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
```diff
@@ -0,0 +1,92 @@
...
+		if wan_disabled == '0' then
+			local wan_name = 'wan_radio' .. suffix
+			local ssid = uci:get('wireless', wan_name, 'ssid')
+			local key = uci:get('wireless', wan_name, 'key')
+			local encryption = uci:get('wireless', wan_name, 'encryption')
+			local mfp = uci:get('wireless', wan_name, 'ieee80211w')
+			uci:set('gluon', 'wireless', 'private_ssid', ssid)
+			uci:set('gluon', 'wireless', 'private_key', key)
+			uci:set('gluon', 'wireless', 'private_encryption', encryption)
+			uci:set('gluon', 'wireless', 'private_mfp', mfp)
```
neocturne: It would make sense to migrate the private-wifi settings regardless of its enable status.

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
```diff
@@ -0,0 +1,92 @@
...
+		end
+		local radio_name = radio['.name']
+		local suffix = radio_name:match('^radio(%d+)$')
+
+		local client_disabled = uci:get('wireless', 'client_radio' .. suffix, 'disabled')
+		local mesh_disabled = uci:get('wireless', 'mesh_radio' .. suffix, 'disabled')
+		local wan_disabled = uci:get('wireless', 'wan_radio' .. suffix, 'disabled')
+		add_or_remove_role(roles, 'client', client_disabled)
+		add_or_remove_role(roles, 'mesh', mesh_disabled)
+		add_or_remove_role(roles, 'private', wan_disabled)
```
neocturne: Hmm, generally I think the logic should work like the following:
- For each band and each role, iterate over all matching `<role>_radioX` sections
- If any such sections exist for a given band and role, the enabled status is taken from these sections (if at least one of these sections is enabled, enable the role; disable it otherwise)
- If no such sections exist, use the default from the site config
- If no radio of a given band exists, do not generate the `wireless_band` at all

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -176,7 +164,6 @@ local function configure_mesh_wireless(radio, index, config, disabled)
 	configure_mesh(config.mesh(), radio, index, suffix,
 		first_non_nil(
 			disabled,
-			is_disabled('mesh_' .. radio_name),
```
neocturne, line 174: The functions `first_non_nil` can be dropped as well - `<config>.mesh.disabled` should have an effect on the initial role list for a band, but here, only whether the role is enabled affects whether the mesh is enabled or not.

We could even omit generating `wifi-iface` sections for disabled roles.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -7,7 +7,9 @@ local iwinfo = require 'iwinfo'
 
 local uci = require('simple-uci').cursor()
 
-uci:delete_all('wireless', 'wifi-iface')
+if not uci:get_bool('gluon', 'wireless', 'preserve_channels') then
```
neocturne: `preserve_channels` should not preserve `wifi-iface` sections.

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
maurerle: I think this matches the current behavior, except for 
1. The default config is generated first, and then adjusted
2. `If no radio of a given band exists, do not generate the wireless_band at all` is done with the latest commit in https://github.com/freifunk-gluon/gluon/pull/3563/commits/6fe3596975061be611c23f0254067341184ae209 - I would squash this commit into the initial 022-wireless-roles commit?

Would you like it better to only add the roles to an empty list and add the default roles to it if nothing exists?
I'll try this then.

neocturne: As far as I can tell, the current behavior in the PR is such that the last iteration for a given band "wins", because `add_or_remove_role` may remove a role that has been added by an earlier iteration. While such setups aren't well-supported at the moment, I think it may also make the code clearer by only adding roles and never removing them.

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
maurerle: resolved in the latest push

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
maurerle: Thanks, yes.
In the current implementation, we can't omit generating mesh configs, as this would keep the wrong setting - if "preserve_channels" aka "keep existing config" is set.

So I first need to improve how "preserve_channels" works.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
maurerle: Now I can either
1. remove “preserve_channels” and set the channel via gluon.wireless.channel_radio1
2. or build a map that preserves the existing settings before discarding it and rebuilding the channel and `chanlist` setting from this..
3. rename `preserve_channels` to `preserve_additional_ifaces_htmode_and_channels` :laughing:

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
maurerle: I improved this in the latest iteration. Thanks!

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
maurerle: Well - the channels are in the radio section and not in the wifi-iface, so I could just remove this line, without breaking the preserve_channels mechanism.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
maurerle: `omit generating wifi-iface sections for disabled roles` is still open, otherwise this is fixed as well

maurerle: This is now implemented as well.

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
```diff
@@ -0,0 +1,100 @@
...
+local site = require 'gluon.site'
+local uci = require('simple-uci').cursor()
+local util = require 'gluon.util'
+local wireless = require 'gluon.wireless'
+
+-- Create wifi roles if they do not exist yet
+local default_wifi_roles = {
+	band2g = {'client'},
+	band5g = {'client'},
+}
```
neocturne, line 12: I don't think these defaults are correct. If no configuration for a band is found in the old config, the site config will define the defaults (based on the existence and `disabled` status of various sections); `client` is not special.

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/lua
+
+local site = require 'gluon.site'
+local uci = require('simple-uci').cursor()
+local util = require 'gluon.util'
+local wireless = require 'gluon.wireless'
+
+-- Create wifi roles if they do not exist yet
+local default_wifi_roles = {
+	band2g = {'client'},
```
neocturne: Let's add an underscore to the section names:

```suggestion
	band_2g = {'client'},
```

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
```diff
@@ -0,0 +1,100 @@
...
+	band5g = {'client'},
+}
+
+-- enable mesh if it is not disabled
+if not site.wifi24.mesh.disabled(false) then
```
neocturne: If no mesh section exists at all, we must not enable the mesh role either:
```suggestion
if site.wifi24.mesh() and not site.wifi24.mesh.disabled(false) then
```
(same for 5g)

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
```diff
@@ -0,0 +1,100 @@
...
+	wireless.foreach_radio(uci, function(radio)
+		if radio.band ~= band then
+			return
+		end
+		local radio_name = radio['.name']
+		local suffix = radio_name:match('^radio(%d+)$')
+
+		for role, iface_prefix in pairs(role_iface_names) do
+			local iface_name = iface_prefix .. suffix
+			local role_disabled = uci:get('wireless', iface_prefix .. suffix, 'disabled')
```
neocturne: If no `disabled` option exists, it should be treated as '0':
```suggestion
			local role_disabled = uci:get('wireless', iface_prefix .. suffix, 'disabled') or '0'
```

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
```diff
@@ -0,0 +1,100 @@
...
+uci:foreach('gluon', 'wireless_band', function(band)
+	local function has_role(role)
+		if band.role == nil then
+			return false
+		end
+
+		return util.contains(band.role, role)
+	end
+
+	if band['.name'].name=='5g' and has_role('mesh') and is_outdoor() then
```
neocturne:
```suggestion
	if band['.name'].name=='band5g' and has_role('mesh') and is_outdoor() then
```
(or rather `band_5g` with my suggestion above applied)

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
```diff
@@ -0,0 +1,100 @@
...
+		band_roles = default_wifi_roles["band" .. band]
+	end
+
+	uci:section('gluon', 'wireless_band', band, {
+		role = band_roles,
+	})
+end
+
+for band_name, _ in pairs(default_wifi_roles) do
+	local raw_band = band_name:gsub("band", "")
```
neocturne:
```suggestion
	local raw_band = band_name:gsub("^band", "")
```
(or rather `^band_` with my suggestion above applied)

package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
```diff
@@ -20,6 +21,10 @@ local function configure_ap(radio, index, config, radio_name)
 	local ap = config.ap
 	local disabled = is_disabled(ap, name)
 
+	if disabled then
```
neocturne: It is not clear to me why this `if disabled` is introduced here - especially with the `is_disabled` implementation unchanged in this commit, which doesn't really make sense anymore, as the disabled status should be derived from the role list, and that is only changed in a later commit.

package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
```diff
@@ -103,8 +112,29 @@ local function configure_owe_transition_mode(config, radio_name)
...
+
 wireless.foreach_radio(uci, function(radio, index, config)
 	local radio_name = radio['.name']
+	-- now we assign radios to roles
+	local band = radio.band
+	local roles = uci:get('gluon', band, 'role')
+
+	-- if there is a radio which does not support the mesh channel, we can configure it on a different channel
+	-- in this case, the other radio on the same band supporting the channel turns off its client radio and is mesh only
+	local disable_client = wireless.supports_channel(radio, config.channel()) and client_only_radio['band' .. band] > 0
```
neocturne: The `client_only_radio` logic should go in a separate commit - the switch to role-based configuration of VIFs is already a very big diff without introducing new logic.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -90,24 +86,6 @@ local function get_htmode(radio)
...
-local function first_non_nil(first, ...)
-	if first ~= nil then
-		return first
-	end
-
-	return first_non_nil(...)
-end
-
-
 local function delete_ibss(radio_name)
```
neocturne: `delete_ibss` can be dropped, as we're deleting all ifaces now.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -220,28 +187,32 @@ wireless.foreach_radio(uci, function(radio, index, config)
 	uci:delete('wireless', radio_name, 'basic_rate')
 
 	local band = radio.band
+	-- now we assign radios to roles
+	local roles = uci:get('gluon', band, 'role')
```
neocturne:
```suggestion
	local roles = uci:get('gluon', 'band_' .. band, 'role')
```

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -127,8 +111,11 @@ local function configure_mesh(config, radio, index, suffix, disabled)
 		return
 	end
 
+	local batadv_hop_penalty = uci:get('gluon', radio.band, 'batadv_hop_penalty') or 0
```
neocturne:
```suggestion
	local batadv_hop_penalty = uci:get('gluon', 'band_' .. radio.band, 'batadv_hop_penalty') or 0
```
There should probably be a helper that you pass in a radio to retrieve role assignment and other options like `batadv_hop_penalty` - that way we only need to adjust the helper when we introduce radio-specific overrides.

package/gluon-web-private-wifi/luasrc/lib/gluon/config-mode/model/admin/privatewifi.lua
```diff
@@ -61,41 +61,31 @@ if wireless.device_supports_mfp(uci) then
...
+		util.add_to_set(roles_5g, 'private')
+	else
+		util.remove_from_set(roles_2g, 'private')
+		util.remove_from_set(roles_5g, 'private')
+	end
+	uci:set_list('gluon', '2g', 'role', roles_2g)
+	uci:set_list('gluon', '5g', 'role', roles_5g)
 
+	uci:commit('gluon')
+	os.execute('/lib/gluon/upgrade/325-gluon-private-wifi')
```
neocturne: I think we should just do a full reconfigure here. We may want to assign VIF MAC address dynamically in the future based on the enabled roles.

package/gluon-web-wifi-config/Makefile
```diff
@@ -5,7 +5,7 @@ PKG_NAME:=gluon-web-wifi-config
 include ../gluon.mk
 
 define Package/gluon-web-wifi-config
-  DEPENDS:=+gluon-web-admin +libiwinfo-lua
+  DEPENDS:=+gluon-web-admin +gluon-client-bridge +libiwinfo-lua
```
neocturne: No, the wifi config should not care whether the `gluon-client-bridge` package is enabled (even though currently all supported network topologies use it).

maurerle: This was needed to execute:
`os.execute('/lib/gluon/upgrade/320-gluon-client-bridge-wireless')` on `f:write()` to directly create the wireless section.
So instead also do a full gluon-reconfigure..?
I think this could be quite much..?

package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
maurerle: I can sure add this change to a later commit.
The idea of this was to not generate the ap/mesh config at all if it is disabled - as suggested by you in the last review.
Would you rather not call the `configure_ap` function at all in this case?

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
maurerle: This is at wrong for the private wifi - this iface does not exist and is therefore disabled.
I am currently checking for existence of the iface and the disabled status in one go.

I think I should surround this with `if uci:get('wireless', iface_prefix .. suffix) then` to do nothing if the iface does not exist.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
maurerle: I would leave this for the radio-specific overrides PR for now..?

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
```diff
@@ -0,0 +1,109 @@
...
+	uci:section('gluon', 'wireless_band', "band_" .. band, {
+		role = band_roles,
+	})
+end
+
+for band_name, _ in pairs(default_wifi_roles) do
+	local raw_band = band_name:gsub("^band_", "")
+	if not uci:get('gluon', band_name) and wireless.device_uses_band(uci, raw_band) then
+		-- remove when upgrades are only possible from v2025.1 onwards
+		migrate_wifi_to_roles(raw_band)
```
neocturne: This is also responsible for setting the defaults, so it can't be removed entirely in the future.

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
```diff
@@ -0,0 +1,109 @@
...
+	end
+
+	uci:section('gluon', 'wireless_band', "band_" .. band, {
+		role = band_roles,
+	})
+end
+
+for band_name, _ in pairs(default_wifi_roles) do
+	local raw_band = band_name:gsub("^band_", "")
+	if not uci:get('gluon', band_name) and wireless.device_uses_band(uci, raw_band) then
```
neocturne: Hmm, I wonder if `device_uses_band` is the right choice here - it might be a good idea generate band configs for all bands that *could* be supported, allow switching the band when there is a single multi-band radio.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -220,28 +176,28 @@ wireless.foreach_radio(uci, function(radio, index, config)
...
-			uci:set('wireless', radio_name, 'country3', '0x4f')
-			configure_mesh_wireless(radio, index, config, true)
-		else
-			uci:delete('wireless', radio_name, 'country3')
+	end
+
+	if roles and util.contains(roles, 'mesh') then
+		if band == '2g' then
 			configure_mesh_wireless(radio, index, config)
+		elseif (band == '5g') then
```
neocturne:
```suggestion
		elseif band == '5g' then
```

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -225,27 +197,30 @@ wireless.foreach_radio(uci, function(radio, index, config)
...
+			else
+				uci:delete('wireless', radio_name, 'country3')
+				configure_mesh_wireless(radio, index, config)
+			end
 		end
 	end
 
 	uci:set('wireless', radio_name, 'beacon_int', beacon_interval)
 
+	uci:set('wireless', "wan_" .. radio_name, 'disabled', util.contains(roles, 'private'))
```
neocturne: While this line is removed again in a later commit, the logic seems to be inverted.

package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
```diff
@@ -1,24 +1,38 @@
...
+end)
 
-	return config.disabled(false)
+local function is_disabled(radio, config)
+	-- now we assign radios to roles
+	local band = radio.band
+	local roles = uci:get('gluon', 'band_' .. band, 'role')
+	-- if there is a radio which does not support the mesh channel, we can configure it on a different channel
+	-- in this case, the other radio on the same band supporting the channel turns off its client radio and is mesh only
+	local disable_client = wireless.supports_channel(radio, config.channel()) and client_only_radio['band' .. band] > 0
```
neocturne:
```suggestion
	local disable_client = wireless.supports_channel(radio, config.channel()) and client_only_radio['band_' .. band] > 0
```

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -220,28 +176,28 @@ wireless.foreach_radio(uci, function(radio, index, config)
 	uci:delete('wireless', radio_name, 'basic_rate')
 
 	local band = radio.band
+	-- now we assign radios to roles
+	local roles = uci:get('gluon', 'band_' .. band, 'role')
```
neocturne:
```suggestion
	local roles = uci:get_list('gluon', 'band_' .. band, 'role')
```
This will set roles to `{}` if no roles are set, so you can omit the `if roles` checks later.

package/gluon-private-wifi/luasrc/lib/gluon/upgrade/325-gluon-private-wifi
```diff
@@ -2,22 +2,55 @@
...
+		uci:delete('wireless', name, 'ieee80211w')
 	end
+end
+
+wireless.foreach_radio(uci, function(radio, index)
+	local radio_name = radio['.name']
+	local name   = "wan_" .. radio_name
 
-	uci:set('wireless', name, 'ifname', suffix and 'wl-wan' .. suffix)
+	local roles = uci:get('gluon', 'band_' .. radio.band, 'role')
```
neocturne:
```suggestion
	local roles = uci:get_list('gluon', 'band_' .. radio.band, 'role')
```
This will set roles to `{}` if no roles are set, so you can omit the `if roles` checks later.

package/gluon-web-wifi-config/luasrc/lib/gluon/config-mode/model/admin/wifi-config.lua
```diff
@@ -38,15 +38,47 @@ f:section(Section, nil, translate(
...
+local function add_or_remove_role(roles, role, disabled)
+	if disabled == true then
+		util.remove_from_set(roles, role)
+	elseif disabled == false then
+		util.add_to_set(roles, role)
+	end
+end
+
+local function vif_option(section, role_name, band, msg)
+	local config_roles = uci:get('gluon', band, 'role') or {}
```
neocturne:
```suggestion
	local config_roles = uci:get_list('gluon', band, 'role')
```

package/gluon-web-wifi-config/luasrc/lib/gluon/config-mode/model/admin/wifi-config.lua
```diff
@@ -38,15 +38,47 @@ f:section(Section, nil, translate(
...
+	local o = section:option(Flag, band .. '_' .. role_name .. '_enabled', msg)
+	o.default = util.contains(config_roles, role_name)
+
+	function o:write(data)
+		-- this does race for multiple options without additional read before write in o:write
+		local roles = uci:get('gluon', band, 'role') or {}
```
neocturne:
```suggestion
		local roles = uci:get_list('gluon', band, 'role')
```

package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
```diff
@@ -1,24 +1,38 @@
...
+	if not wireless.supports_channel(radio, config.channel()) then
+		count_band("band_" .. radio.band)
 	end
+end)
 
-	return config.disabled(false)
+local function is_disabled(radio, config)
+	-- now we assign radios to roles
+	local band = radio.band
+	local roles = uci:get('gluon', 'band_' .. band, 'role')
```
neocturne:
```suggestion
	local roles = uci:get_list('gluon', 'band_' .. band, 'role')
```
This will set roles to `{}` if no roles are set, so you can omit the `if roles` checks later.

package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
```diff
@@ -80,7 +80,7 @@ local function configure_owe(radio, index, config, radio_name)
 		ssid = ap.owe_ssid(),
 		macaddr = macaddr,
 		ifname = suffix and 'owe' .. suffix,
-		disabled = disabled or false,
+		disabled = false,
```
neocturne: All the `disabled = false` lines can be dropped where the whole section is recreated on every reconfigure anyways.

package/gluon-web-private-wifi/luasrc/lib/gluon/config-mode/model/admin/privatewifi.lua
```diff
@@ -61,41 +61,31 @@ if wireless.device_supports_mfp(uci) then
...
+	else
+		util.remove_from_set(roles_2g, 'private')
+		util.remove_from_set(roles_5g, 'private')
+	end
+	uci:set_list('gluon', 'band_2g', 'role', roles_2g)
+	uci:set_list('gluon', 'band_5g', 'role', roles_5g)
 
+	uci:commit('gluon')
+	os.execute('/usr/bin/gluon-reconfigure')
+	uci:commit('network')
```
neocturne: No commits are needed after the reconfigure.

neocturne (changes requested):

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
maurerle: I deleted this line therefore, as it is not needed. Thanks!

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
maurerle: I added this commit which should tackle this:
https://github.com/freifunk-gluon/gluon/pull/3563/commits/032a09fdecbf3da4c3998960567d3ceddd9e1c72

I implemented all the suggested changes.

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
```diff
@@ -0,0 +1,109 @@
...
+
+local function migrate_wifi_to_roles(band)
+	local band_roles = {}
+	local config_exists = false
+	wireless.foreach_radio(uci, function(radio)
+		if radio.band ~= band then
+			return
+		end
+		local radio_name = radio['.name']
+		local suffix = radio_name:match('^radio(%d+)$')
```
neocturne: You don't really need this `match` - you can just change `role_iface_names` to contain values like `client_` and concatenate that with `radio_name` directly.

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
```diff
@@ -0,0 +1,109 @@
...
+
+		for role, iface_prefix in pairs(role_iface_names) do
+			local iface_name = iface_prefix .. suffix
+			if uci:get('wireless', iface_prefix .. suffix) then
```
neocturne:
```suggestion
			if uci:get('wireless', iface_name) then
```

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
```diff
@@ -0,0 +1,109 @@
...
+				config_exists = true
+				local role_disabled = uci:get('wireless', iface_prefix .. suffix, 'disabled') or 0
```
neocturne:
```suggestion
				local role_disabled = uci:get('wireless', iface_name, 'disabled') or 0
```

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
```diff
@@ -0,0 +1,109 @@
...
+uci:foreach('gluon', 'wireless_band', function(band)
+	local function has_role(role)
+		if band.role == nil then
+			return false
+		end
+
+		return util.contains(band.role, role)
+	end
+
+	if band['.name'].name=='5g' and has_role('mesh') and is_outdoor() then
```
neocturne:
```suggestion
	if band['.name'] =='band_5g' and has_role('mesh') and is_outdoor() then
```
Do you have hardware to test this code path?

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
```diff
@@ -0,0 +1,109 @@
...
+end
+
+-- Fix invalid mesh role configurations for outdoor
+uci:foreach('gluon', 'wireless_band', function(band)
+	local function has_role(role)
+		if band.role == nil then
+			return false
+		end
+
+		return util.contains(band.role, role)
```
neocturne: I prefer the or {} to explicit nil checks:
```suggestion
		return util.contains(band.role or {}, role)
```

package/gluon-core/luasrc/usr/lib/lua/gluon/wireless.lua
```diff
@@ -135,4 +145,18 @@ function M.device_uses_11a(uci)
 	return ret
 end
 
+function M.device_supports_band(uci, band)
+	local ret = false
+	M.foreach_radio(uci, function(radio)
+		local hwmodes = iwinfo.nl80211.hwmodelist(M.find_phy(radio))
+
+		if band == '2g' and (hwmodes.g) then
```
neocturne:
```suggestion
		if band == '2g' and hwmodes.g then
```

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -169,17 +146,11 @@ local function fixup_wan(radio, index)
...
 	local suffix = radio_name:match('^radio(%d+)$')
 
-	configure_mesh(config.mesh(), radio, index, suffix,
-		first_non_nil(
-			disabled,
-			is_disabled('mesh_' .. radio_name),
-			config.mesh.disabled(false)
-		)
-	)
+	configure_mesh(config.mesh(), radio, index, suffix)
```
neocturne: `configure_mesh_wireless` is trivial now, it can be merged with `configure_mesh`.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -225,22 +196,24 @@ wireless.foreach_radio(uci, function(radio, index, config)
...
-			uci:delete('wireless', radio_name, 'country3')
+	end
+
+	if util.contains(roles, 'mesh') then
+		if band == '2g' then
 			configure_mesh_wireless(radio, index, config)
+		elseif band == '5g' then
+			if is_outdoor() then
+				-- enforce outdoor channels by filtering the regdom for outdoor channels
+				uci:set('wireless', radio_name, 'country3', '0x4f')
```
neocturne: `country3` must be added/removed for `band == 5g and is_outdoor()` regardless of the existence of the 'mesh' role. I wonder if it could just be set for `is_outdoor()` regardless of band, thus also future-proofing the code for other bands?

package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
```diff
@@ -17,20 +17,22 @@ wireless.foreach_radio(uci, function(radio, _, config)
...
-	return config.disabled(false)
+local function is_disabled(radio, config)
+	-- now we assign radios to roles
+	local band = radio.band
+	local roles = uci:get_list('gluon', 'band_' .. band, 'role')
+	-- if there is a radio which does not support the mesh channel, we can configure it on a different channel
+	-- in this case, the other radio on the same band supporting the channel turns off its client radio and is mesh only
+	local disable_client = wireless.supports_channel(radio, config.channel()) and client_only_radio['band_' .. band] > 0
+	local client_enabled = util.contains(roles, 'client') and not disable_client
+	return config.ap.disabled(not client_enabled)
```
neocturne: `config.ap.disabled` is just the default for the role assignment - it must not be taken into account here.

package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
```diff
@@ -117,20 +115,12 @@ end
...
 	local radio_name = radio['.name']
-	-- now we assign radios to roles
-	local band = radio.band
-	local roles = uci:get('gluon', band, 'role')
-
-	-- if there is a radio which does not support the mesh channel, we can configure it on a different channel
-	-- in this case, the other radio on the same band supporting the channel turns off its client radio and is mesh only
-	local disable_client = wireless.supports_channel(radio, config.channel()) and client_only_radio['band' .. band] > 0
+	if not is_disabled(radio, config) then
+		configure_ap(radio, index, config, radio_name)
```
neocturne, line 114: Ideally we'd move all wireless setup of `gluon-client-bridge` into `200-wireless` and only leave the actual bridge setup in `gluon-client-bridge` - I feel like that would allow for a lot of deduplication and clearer code. Probably best left for a follow-up PR though.

package/gluon-private-wifi/luasrc/lib/gluon/upgrade/325-gluon-private-wifi
```diff
@@ -2,22 +2,48 @@
...
-	if not uci:get('wireless', name, 'device') then
-		return
+	local encryption = uci:get('gluon', 'wireless', 'private_encryption')
+	local ssid = uci:get('gluon', 'wireless', 'private_ssid')
+	local key = uci:get('gluon', 'wireless', 'private_key')
+	local mfp = wireless.device_supports_mfp(uci) and uci:get('gluon', 'wireless', 'private_mfp')
+
+	-- migrate encryption from Gluon v2023.2.x or older
+	-- remove this in 2027 or on first release supporting only upgrades from >=v2025.1.x
+	encryption = encryption and encryption:gsub("psk3", "sae")
```
neocturne: This migration must happen when setting `gluon.wireless.private_encryption`, it's not useful in the wifi-iface setup.

package/gluon-private-wifi/luasrc/lib/gluon/upgrade/325-gluon-private-wifi
```diff
@@ -2,22 +2,48 @@
...
+	local macaddr = wireless.get_wlan_mac('wan_radio', index, radio)
+	uci:section('wireless', 'wifi-iface', name, {
+		device     = radio_name,
+		network    = 'wan',
+		mode       = 'ap',
+		encryption = encryption,
+		ssid       = ssid,
+		key        = key,
+		macaddr    = macaddr,
+		ifname     = suffix and 'wl-wan' .. suffix,
```
neocturne: We should just exit early from `configure_private_wifi` if we have no `suffix`.

package/gluon-web-private-wifi/luasrc/lib/gluon/config-mode/model/admin/privatewifi.lua
```diff
@@ -61,42 +61,31 @@ if wireless.device_supports_mfp(uci) then
...
+		uci:set('gluon', 'wireless', 'private_ssid', ssid.data)
+		uci:set('gluon', 'wireless', 'private_key', key.data)
+		uci:set('gluon', 'wireless', 'private_encryption', encryption.data)
+		uci:set('gluon', 'wireless', 'private_mfp', mfp.data)
+	end
 
-			-- hostapd-mini won't start in case 802.11w is configured
-			if wireless.device_supports_mfp(uci) then
-				uci:set('wireless', name, 'ieee80211w', mfp.data)
+	for _, band in ipairs({'band_2g', 'band_5g'}) do
```
neocturne: This must iterate over the existing bands in the wireless config, then you can drop the `device_supports_band` here.

package/gluon-web-private-wifi/luasrc/lib/gluon/config-mode/model/admin/privatewifi.lua
```diff
@@ -38,12 +38,12 @@ warning:depends(enabled, true)
 local ssid = s:option(Value, "ssid", translate("Name (SSID)"))
 ssid:depends(enabled, true)
 ssid.datatype = "maxlength(32)"
-ssid.default = uci:get('wireless', primary_iface, "ssid")
```
neocturne, line 41: There are remaining uses of `primary_iface` in this file that must be replaced as well.

package/gluon-web-wifi-config/luasrc/lib/gluon/config-mode/model/admin/wifi-config.lua
```diff
@@ -38,15 +38,47 @@ f:section(Section, nil, translate(
 
 local mesh_vifs_5ghz = {}
 
+local function add_or_remove_role(roles, role, disabled)
```
neocturne: Passing `disabled` rather than `enabled` here feels backwards.

package/gluon-web-wifi-config/luasrc/lib/gluon/config-mode/model/admin/wifi-config.lua
```diff
@@ -38,15 +38,47 @@ f:section(Section, nil, translate(
...
+		add_or_remove_role(roles, role_name, not data)
+		uci:set('gluon', band, 'role', roles)
+	end
+
+	return o
+end
+
+local bands = {}
 
 uci:foreach('wireless', 'wifi-device', function(config)
```
neocturne: Per-band and per-device config should be separated completely, iterating over bands instead of devices for the per-band config. Most config is per-band, with just things like txpower that aren't reflected in /etc/config/gluon yet remaining per-device.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
maurerle: if is_outdoor is set, all radios only allow setting outdoor channels right..?
I guess, this generally might need some adjustments and should be set per radio in the future?

package/gluon-web-private-wifi/luasrc/lib/gluon/config-mode/model/admin/privatewifi.lua
maurerle: I guess this requires to introduce a checkbox per band so that one can decide the role per band.

This is generally useful, as the current implementation on main did have the issue, that saving private-wifi settings in config mode did enable it for both bands and could not be enabled only on one band through the UI.

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
maurerle: Yes, I can just set `uci set gluon.wireless.outdoor=1`, enable mesh role on 5g and make sure it is disabled after running `gluon-reconfigure`.
Works fine with your fix - I forgot to adjust/test this case in the mean time though.

maurerle:
This is how the wifi-config looks like for the whw-03 v2:

image


For private-wifi it could look like this:

** draft **:
image

Though I don't know how the further options can depend on both checkboxes conditionally.
So I just kept it at the current approch and took the bug (can't set private enabled per band) with us to fix in the future.

neocturne:
I like the screenshots of the new config interface a lot!

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
```diff
@@ -0,0 +1,102 @@
...
+	end
+end
+
+
+local function is_outdoor()
+	return uci:get_bool('gluon', 'wireless', 'outdoor')
+end
+
+-- Fix invalid mesh role configurations for outdoor
+uci:foreach('gluon', 'wireless_band', function(band_config)
```
neocturne: This could just `uci:get_list('gluon', 'band_5g', 'role')` directly instead of iterating over the bands and retrieving all options.

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
```diff
@@ -0,0 +1,102 @@
...
+
+local site = require 'gluon.site'
+local uci = require('simple-uci').cursor()
+local util = require 'gluon.util'
+local wireless = require 'gluon.wireless'
+
+-- Create wifi roles if they do not exist yet
+local default_wifi_roles = {
+	band_2g = {},
+	band_5g = {},
```
neocturne: There are multiple pieces of code in this script that would become simpler if the keys in `default_wifi_rules` were just `2g` and `5g`.

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
```diff
@@ -75,7 +75,7 @@ end
 
 for band_name, _ in pairs(default_wifi_roles) do
 	local raw_band = band_name:gsub("^band_", "")
-	if not uci:get('gluon', band_name) and wireless.device_uses_band(uci, raw_band) then
+	if not uci:get('gluon', band_name) and wireless.device_supports_band(uci, raw_band) then
```
neocturne: Hmm, I guess we should leave this commit out for now, as we don't handle switching a radio to one band or the other based on enabled roles anyways (and that would also require some more design work)

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -90,40 +86,17 @@ local function get_htmode(radio)
...
-local function first_non_nil(first, ...)
-	if first ~= nil then
-		return first
-	end
-
-	return first_non_nil(...)
-end
-
-
-local function delete_ibss(radio_name)
```
neocturne, line 106: The same commit that removes `delete_ibss` should also remove its call.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -90,40 +86,17 @@ local function get_htmode(radio)
...
-local function configure_mesh(config, radio, index, suffix, disabled)
+local function configure_mesh_wireless(radio, index, mesh_config)
 	local radio_name = radio['.name']
 	local name = 'mesh_' .. radio_name
 
 	local macfilter = uci:get('wireless', name, 'macfilter')
 	local maclist = uci:get('wireless', name, 'maclist')
+	local suffix = radio_name:match('^radio(%d+)$')
 
 	uci:delete('wireless', name)
```
neocturne: This `uci:delete()` is also redundant now.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -90,40 +86,17 @@ local function get_htmode(radio)
...
-
-
-local function delete_ibss(radio_name)
-	local name = 'ibss_' .. radio_name
-
-	uci:delete('wireless', name)
-end
-
-local function configure_mesh(config, radio, index, suffix, disabled)
+local function configure_mesh_wireless(radio, index, mesh_config)
```
neocturne: Probably better to call this `configure_mesh` - the wireless suffix seems redundant, as we're already in `200-wireless`.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -140,48 +113,18 @@ local function configure_mesh(config, radio, index, suffix, disabled)
...
-		disabled = disabled,
+		basic_rate = { mesh_config.mcast_rate },
+		mcast_rate = mesh_config.mcast_rate,
+		ifname = 'mesh' .. suffix,
 		macfilter = macfilter,
 		maclist = maclist,
 	})
 end
 
-local function fixup_wan(radio, index)
```
neocturne, line 151: The same commit that removes `fixup_wan` should also remove its call.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -177,7 +177,7 @@ wireless.foreach_radio(uci, function(radio, index, config)
 			configure_mesh_wireless(radio, index, config.mesh())
 		end
 	elseif band == '5g' then
-		if is_outdoor() then
+		if is_outdoor() or not wireless.supports_channel(radio, config.channel()) then
```
neocturne: This change does not make sense - there is no reason to set `country3=0x4f` if we're not outdoors, regardless of usage of a radio.

package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
```diff
@@ -105,6 +117,15 @@ end
...
 	local radio_name = radio['.name']
+	-- now we assign radios to roles
+	local band = radio.band
+	local roles = uci:get('gluon', band, 'role')
+
+	-- if there is a radio which does not support the mesh channel, we can configure it on a different channel
+	-- in this case, the other radio on the same band supporting the channel turns off its client radio and is mesh only
+	local disable_client = wireless.supports_channel(radio, config.channel()) and client_only_radio['band' .. band] > 0
+
+	uci:set('wireless', 'client_' .. radio_name, 'disabled', util.contains(roles, 'client') and not disable_client)
```
neocturne: This can't work - the `client_...` section doesn't exist at this point, and it does not affect the OWE interface.

 I think it would make sense to move this commit to the very end of the series, in particular after the change that makes it so the client ifaces are only created when enabled.

package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
```diff
@@ -17,26 +17,27 @@ wireless.foreach_radio(uci, function(radio, _, config)
...
 end
 
 local function configure_ap(radio, index, config, radio_name)
 	local name = 'client_' .. radio_name
 	local suffix = radio_name:match('^radio(%d+)$')
 
 	local ap = config.ap
-	local disabled = is_disabled(ap, name)
 
 	uci:delete('wireless', name)
```
neocturne: Can be dropped, we delete all ifaces now

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
```diff
@@ -52,6 +52,11 @@ local function migrate_wifi_to_roles(band)
 						local ssid = uci:get('wireless', iface_name, 'ssid')
 						local key = uci:get('wireless', iface_name, 'key')
 						local encryption = uci:get('wireless', iface_name, 'encryption')
+
+						-- migrate encryption from Gluon v2023.2.x or older
+						-- remove this in 2027 or on first release supporting only upgrades from >=v2025.1.x
+						encryption = encryption and encryption:gsub("psk3", "sae")
```
neocturne, line 58: This change should be moved to the commit creating `022-wireless-roless`.

package/gluon-private-wifi/luasrc/lib/gluon/upgrade/325-gluon-private-wifi
```diff
@@ -2,22 +2,48 @@
...
+		key        = key,
+		macaddr    = macaddr,
+		ifname     = 'wl-wan' .. suffix,
+	})
+
+	-- hostapd-mini won't start in case 802.11w is configured
+	if mfp then
+		uci:set('wireless', name, 'ieee80211w', mfp)
+	else
+		uci:delete('wireless', name, 'ieee80211w')
```
neocturne: This `delete` is not needed anymore, we remove all ifaces.

package/gluon-web-wifi-config/luasrc/lib/gluon/config-mode/model/admin/wifi-config.lua
```diff
@@ -38,15 +38,38 @@ f:section(Section, nil, translate(
 
 local mesh_vifs_5ghz = {}
 
+local function add_or_remove_role(roles, role, enabled)
+	if enabled == true then
+		util.add_to_set(roles, role)
+	elseif enabled == false then
```
neocturne: I don't see how `enabled` could ever be `nil` here - so I guess this could just be `if enabled then ... else ... end`?

package/gluon-web-wifi-config/luasrc/lib/gluon/config-mode/model/admin/wifi-config.lua
```diff
@@ -38,15 +38,38 @@ f:section(Section, nil, translate(
...
 
-uci:foreach('wireless', 'wifi-device', function(config)
-	local radio = config['.name']
+local function vif_option(section, role_name, band, msg)
+	local config_roles = uci:get_list('gluon', band, 'role')
+	local o = section:option(Flag, band .. '_' .. role_name .. '_enabled', msg)
+	o.default = util.contains(config_roles, role_name)
+
+	function o:write(data)
+		-- this does race for multiple options without additional read before write in o:write
```
neocturne: This comments sounds like there is something wrong with this code, but it's only supposed to explain why the code is written like this, right? Maybe reword it to something like "Without the additional read, this would race [...]"?

package/gluon-web-wifi-config/luasrc/lib/gluon/config-mode/model/admin/wifi-config.lua
```diff
@@ -55,51 +78,16 @@ uci:foreach('wireless', 'wifi-device', function(config)
...
-		end
-
-		return o
-	end
+	vif_option(p, 'client', band, translate('Enable client network (access point)'))
 
-	vif_option('client', {'client', 'owe'}, translate('Enable client network (access point)'))
-
-	local mesh_vif = vif_option('mesh', {'mesh'}, translate("Enable mesh network (802.11s)"))
+	local mesh_vif = vif_option(p, 'mesh', band, translate("Enable mesh network (802.11s)"))
```
neocturne: I think it would make sense to pass the `band_config` into `vif_option()`, so it doesn't have to retrieve options from UCI again.

neocturne:
As mentioned before, I would also prefer not to call this "triband" in commit descriptions or the PR title, as this does not add support for any third band. "Triple radio" would be more appropriate.

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
maurerle: I moved it to my other 6ghz testing branch - as it would be quite good to not generate a 6ghz config for all devices - but unrelated for now.
(checking for uses is alright here - woulb be bad if not check is present)

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
maurerle: The original idea was that the spare wifi is configured on different channels than the 5ghz with mesh configured.
The most simple option to this is to configure it as outdoor..?
We would need to set the channel to 'auto' as well though.

So you propose to not set the country3 and fix the setting of auto channel? :blush:

neocturne: country3 must be set exactly when `is_outdoor()` is true - it just causes the device to respect regulatory limitations for outdoor operation in its channel choice. This seems orthogonal to the channel being different from the mesh channel.

package/gluon-web-wifi-config/luasrc/lib/gluon/config-mode/model/admin/wifi-config.lua
maurerle: Yes, this was from a time, where it was nil if unchanged. Thanks!

package/gluon-web-wifi-config/luasrc/lib/gluon/config-mode/model/admin/wifi-config.lua
maurerle: It must have the values from uci to set them in write, right?

neocturne: For write yes, probably easiest to get and set individually in `o:write()` (the alternative would be to use a global write for the whole section, completely replacing the roles lists with the submitted settings). For `o.default` you can use the value from `band_config` instead of retrieving it again.

dekarl:
I'm currently looking how to mix standard Gluon with two private SSID / VLAN. Does a config section like this look ok as extension on top of your PR?

```diff
- config wireless_private 'priv1'
- 	option private_ssid 'MyOtherWireless'
- 	option private_key 'mypassword'
- 	option private_encryption 'psk2'
- 	option private_mfp '2'
- 	option private_interface 'lan.45'
- 
- config wireless_private 'priv2'
- 	option private_ssid 'MyWireless'
- 	option private_key 'myotherpassword'
- 	option private_encryption 'psk2'
- 	option private_mfp '2'
- 	option private_interface 'lan.48'
```

Edit: After reading more... I'll look into splitting out a "role other" for interfaces and private_ssid of upstream or said other interfaces

package/gluon-web-wifi-config/luasrc/lib/gluon/config-mode/model/admin/wifi-config.lua
maurerle: Done :)

maurerle:
https://github.com/freifunk-gluon/gluon/compare/d3ccef6852d8423c0ca618a74e242b79e8bf0dc2..909c63235daf6fde5fb8070550f601210fb09f53
These should be the latest changes.

Tested on linksys-whw03v2

neocturne:
@dekarl Hmm, I guess we should support the `gluon_preserve` option for `wireless` sections like we do in `system` and `network`, allowing to create additional VIFs not touched by the upgrade scripts.

maurerle:
I implemented gluon_preserve for the wifi section as well in https://github.com/freifunk-gluon/gluon/pull/3563/commits/803cf9baa644defb66f35aae6256e9b90b666f34

Tested it and works fine.

@neocturne I can't use `mv -f "$WIRELESS_CFG" "$WIRELESS_SAVED"` just yet, as the radio settings are not yet migrated.
I will fix this to `cp` in a follow-up PR when we introduce the radio-specific `wireless_radio` sections


Adding:

```
config wifi-iface 'iot_radio'
	option macaddr '56:36:7a:aa:de:bf'
	option ifname 'wl-iot0'
	option network 'wan'
	option device 'radio0'
	option gluon_preserve '1'
	option ssid 'IoT'
	option encryption 'psk2'
	option key 'meiniotnetzwerk'
	option mode 'ap'
```

is preserved here. @neocturne I am not too happy with the MAC assignment in this case - which is created at random by the user..?

dekarl:
> I am not too happy with the MAC assignment in this case - which is created at random by the user..?

I don't like it either. I took the logic from https://gluon.readthedocs.io/en/latest/features/private-wlan.html but instead of 4*$RID I used the running count of private SSIDs instead, to get unique MAC addresses per radio. (the private SSID is out-of-scope for batman, so the MAC address only needs to be unique per radio, not globally)

I saw some people looking to bring up separate additional interfaces to connect PTP links to batman without them being directly bridged. So I wondered if a "role" "other" makes sense, that brings up an interface but does not do anything with it.

Then one could generalized the private SSID logic from only WAN to upstream/other interface. So the MAC selection is not in the hands of the user.

My current solution is to put a custom script into /lib/gluon/upgrade. This is good enough for me. I prefer not to derail this PR. So if this is more then 5 lines extra to generalize, its something for later.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
neocturne: This change is still added and then removed again in the following commit. It should not be added in the first place.

maurerle: Thanks! This is now fixed

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
maurerle: a helper to get the roles from a radio is implemented/handled in #3601, so I am resolving this discussion.

PR #3567, opened 2025-08-20
gluon-mesh-batman-adv: fix has_default_gw4

maurerle:
before, the script did land in exit 2 as $out is not empty

this fixes a bug introduced in #3556 which led to the ssid-changer always showing the offline SSID on current gluon main.

Also checking the `gw_mode` is only of interest if no `batctl gwl` exists.

And if the gw_mode is set to server, we should return with exit 0.

### DIFF
--- a/package/gluon-mesh-batman-adv/files/lib/gluon/state/check.d/has_default_gw4
+++ b/package/gluon-mesh-batman-adv/files/lib/gluon/state/check.d/has_default_gw4
@@ -8,12 +8,8 @@ if ! out=$(batctl gwl -H 2>/dev/null); then
 	exit 1
 fi
 
-if [ -n "$out" ]; then
-	# no batctl gw available
+# node does not have any gateways listed
+# and is also not in gw_mode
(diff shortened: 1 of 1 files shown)


package/gluon-mesh-batman-adv/files/lib/gluon/state/check.d/has_default_gw4
```diff
@@ -8,12 +8,10 @@ if ! out=$(batctl gwl -H 2>/dev/null); then
...
-	# no batctl gw available
-	exit 2
-fi
-
-if ! batctl gw_mode | grep -qw '^server'; then
-	# server is also not in gw_mode
-	exit 3
+if [ -z "$out" ]; then
+	# node does not have any gateways listed
+	if ! batctl gw_mode | grep -qw '^server'; then
```
rotanid: is there a need for a second if, why not combine them in one "if" statement?

maurerle: True

maurerle:
Yes, tested and working well.
Sorry again for breaking this before.

PR #3568, opened 2025-08-22
 ramips-mt76x8: add support for TP-Link Archer C50 v6

GoliathLabs:
  - [X] Web interface
  - [X] TFTP
  - [X] Other: https://github.com/openwrt/openwrt/commit/e9ac1b19e0f3d383ab83373c261bfb5527a29521
```
root@ffmuc-9c5322801069:~# lua -e 'print(require("platform_info").get_image_name())'
tp-link-archer-c50-v6-ca-eu-ru
```
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [X] Should show activity
  - Switch port LEDs
- Outdoor devices only:
  - ~Added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`~ **device is no outdoor device**
- Docs:


GoliathLabs:
I finally got my hands on a C50 v6 again. Everything is working as expected. Even meshing on both radios works.

The only problem I currently face is, that the lua -e 'print(require("platform_info").get_image_name())' yields a different device name than specified by OpenWrt & in this PRs target file. How should we deal with that?

GoliathLabs:
(See https://map.ffmuc.net/#!/en/map/9c5322801069)

GoliathLabs:
I also got a C50 v6.20 on hand. The hardware appears to be identical to the v6, but it ships with TP-Link's proprietary TPOS instead of the usual Linux-based firmware.
It boots fine using the v6 OpenWrt/Gluon image, but it does not report the hardware revision correctly as printed on the sticker on the back.

Should we treat the v6.20 differently upstream, or simply document it as a v6 variant that uses TPOS by default?

targets/ramips-mt76x8
```diff
@@ -62,6 +62,10 @@ device('tp-link-archer-c50-v4', 'tplink_archer-c50-v4', {
 	factory = false,
 })
 
+device('tp-link-archer-c50-v6', 'tplink_archer-c50-v6', {
```
AiyionPrime: Maybe I just miss the bigger problem. But can't you just set this to the value you found using platforminfo?

```suggestion
device('tp-link-archer-c50-v6-ca-eu-ru', 'tplink_archer-c50-v6', {
```

https://gluon.readthedocs.io/en/latest/dev/hardware.html#device-definitions

docs/user/supported_devices.rst
```diff
@@ -552,7 +552,7 @@ ramips-mt76x8
 * TP-Link
 
   - Archer C20 (v4, v5)
-  - Archer C50 (v3, v4)
+  - Archer C50 (v3, v4, v6)
```
AiyionPrime:
```suggestion
  - Archer C50 (v3, v4, v6 CA/EU/RU)
```

The US variant is not supported, but shares a lot - if not all - hardware details with the C50v5.
Anyway, not what you're currently aiming for.

We used the same description before: https://github.com/freifunk-gluon/gluon/blob/main/docs/user/supported_devices.rst?plain=1#L132

GoliathLabs:
> The OpenWrt wiki suggests flashing is not as easy as you marked in the checklist? I think others would appreciate a link to the instructions :)
> 
> either [openwrt/openwrt@e9ac1b1](https://github.com/openwrt/openwrt/commit/e9ac1b19e0f3d383ab83373c261bfb5527a29521)
> 
> or [https://openwrt.org/toh/tp-link/archer-c50?s[]=more&s[]=host#v6_instructions](https://openwrt.org/toh/tp-link/archer-c50?s%5B%5D=more&s%5B%5D=host#v6_instructions)

Fixed. Thank you!

Djfe:
this device is not going to last many releases (just a warning)
64MiB RAM for dual band wifi is cutting it close. 8MiB flash even closer.
and mt7613 is cool but does not support outdoor 5ghz channels

AiyionPrime:
@GoliathLabs despite being a weak device, ave you verified, the profile name does now match the autoupdater name?

trixmoe:
> this device is not going to last many releases (just a warning) 64MiB RAM for dual band wifi is cutting it close. 8MiB flash even closer. and mt7613 is cool but does not support outdoor 5ghz channels

Very fair. I'm personally in a situation where I've ended up with quite a few of these devices in my hands, and I can either use them, or somehow dispose of them (selling, etc.).

GoliathLabs:
> @GoliathLabs despite being a weak device, ave you verified, the profile name does now match the autoupdater name?

I'll probably be able to do that tomorrow 👍

GoliathLabs:
Gluon profile name now matches the autoupdater image name. Reflashed the device with the new PR build

```
root@9c5322801069:# lua -e 'print(require("platform_info").get_image_name())'
tp-link-archer-c50-v6-ca-eu-ru
```

AiyionPrime:
Thanks for your effort and following up :)

GoliathLabs:
> Thanks for your effort and following up :)

Thank you for your patience and support! It's highly appreciated! :)

PR #3570, opened 2025-08-25
gluon-mesh-layer3-common: Allow lifetime configuration in site

AiyionPrime:
> as optional section:
> uradvd = {
>     preferred_lifetime = 14400,
>     valid_lifetime = 86400,
> },

this is currrently **blocked** by #3569.

### DIFF
--- a/package/gluon-mesh-layer3-common/luasrc/lib/gluon/radvd/arguments
+++ b/package/gluon-mesh-layer3-common/luasrc/lib/gluon/radvd/arguments
@@ -5,3 +5,11 @@ io.write("-i local-node --default-lifetime 900 -a " .. site.prefix6())
 if site.dns() and site.dns.servers() then
 	io.write(" --rdnss " .. site.next_node.ip6())
 end
+if site.uradvd() then
+	if site.uradvd.preferred_lifetime() then
+		io.write(" --preferred-lifetime " .. site.uradvd.preferred_lifetime())
+	end
+	if site.uradvd.valid_lifetime() then
+		io.write(" --valid-lifetime " .. site.uradvd.valid_lifetime())
+	end
+end


AiyionPrime:
With a site containing `300` and `150` like described in the commit messages:

```lua
uradvd = {
        preferred_lifetime = 150,
        valid_lifetime = 300,
},
```

`uradvd` is launched accordingly.

```console
root@platzhalter-525400123457:~# ps | grep uradvd
 2517 root       808 S    /usr/sbin/uradvd -i local-node -p fdca:ffee:8:14::/64 --rdnss fdca:ffee:8:14::1 --preferred-lifetime 150 --valid-lifetime 300
 5007 root      1184 S    grep uradvd
 ```

**Comment by herbetom** [comment]
I think i would prefer if it would use `radvd` as config section name (instead of `uradvd`).

I've got two arguments for it:
- the package is called gluon-radvd
- radvd is a more generic term (switching to another package seems somewhat unlikely but i don't think it's neccesary to expose the underlying package name to gluon users)

But these are just my current thoughts on the matter.

File: package/gluon-mesh-batman-adv/luasrc/lib/gluon/radvd/arguments [hunk 1]
```diff
@@ -7,3 +7,11 @@ io.write("-i local-node -p " .. site.prefix6())
 if site.dns.servers() and site.next_node.ip6() then
 	io.write(" --rdnss " .. site.next_node.ip6())
 end
+if site.uradvd() then
```
neocturne: This check is redundant, as gluon-site will just treat fields of non-existent tables as non-existent as well instead of throwing an error.

neocturne (dismissed):

neocturne:
> I think i would prefer if it would use `radvd` as config section name (instead of `uradvd`).

I agree, this should be called `radvd`.

package/gluon-mesh-batman-adv/luasrc/lib/gluon/radvd/arguments
AiyionPrime: Sorry for the dealay; I've been on a festival. I'll resolve this this afternoon.

AiyionPrime: resolved in 47609871c441a34af0e1dfc9e3f8637598a9ed52

docs/package/gluon-radvd.rst
```diff
@@ -0,0 +1,36 @@
...
+
+This package provides skeleton service files for ``uradvd``.
+It requires another package to provide the arguments ``uradvd`` is supposed to
+be launched with.
+
+:doc:`gluon-mesh-batman-adv` and ``gluon-mesh-layer3-common`` are two packages
+providing such arguments.
+
+This allows to have ``uradvd`` announce default routes for layer three meshes,
+while only announcing prefixes for layer two meshes.
```
herbetom:
```suggestion
This allows to have ``uradvd`` announce default routes for layer 3 meshes,
while only announcing prefixes for layer 2 meshes.
```

I think writing the number down is the more typical way and what we've used in the past for layer 2 or layer 3. But no strong opionion.

docs/package/gluon-radvd.rst
```diff
@@ -0,0 +1,36 @@
...
+
+site.conf
+---------
+
+radvd.preferred_lifetime : optional
+    - the span of time during which the address can be freely used as a source
+      and destination for traffic. Should be less or equal valid-lifetime.
```
herbetom, line 23:
```suggestion
      and destination for traffic. Should be less or equal valid_lifetime.
```

Probably makes sense to use the exact spelling of the other option.

herbetom (commented):
Just saw the following two minor things. Sorry for only noticing it now. Feel free to ignore.

docs/package/gluon-radvd.rst
AiyionPrime: The texts currently match their description in `uradvd`.
If this is a strong opinion, we can remove the dash in both docs,
but I'd just leave it as it is.

docs/package/gluon-radvd.rst
AiyionPrime: I think it would be correct to refer to the layer with the number three, and spell it out as it's a small number, or use the OSI naming and refer to it by its title.
In that case we can use the digit, but would capitalize it as name `OSI Layer 3` or `Layer 3`.

Looking at out code base I think this is personal preference at best and we should spend less time on such things in general.

AiyionPrime: resolved in 0805f8015859703ae94a9334f67eb796d2d879b8.
Now matches the VXLAN section in the VPN docs.

docs/package/gluon-radvd.rst
herbetom: (I would argue that gluon-radvd is an abstraction layer and therefore should reference itself where possible. But this really isn't something i want to spend time on arguing. So fine with me. :) )

docs/package/gluon-radvd.rst
```diff
@@ -0,0 +1,36 @@
...
+
+This package provides skeleton service files for ``uradvd``.
+It requires another package to provide the arguments ``uradvd`` is supposed to
+be launched with.
+
+:doc:`gluon-mesh-batman-adv` and ``gluon-mesh-layer3-common`` are two packages
+providing such arguments.
+
+This allows to have ``uradvd`` announce default routes for Layer 3 meshes,
+while only announcing prefixes for Layer 2 meshes.
```
lemoer:
```suggestion
:doc:`gluon-mesh-batman-adv` and ``gluon-mesh-layer3-common`` are two packages
providing such arguments in order to announce default routes for Layer 3 meshes,
while only announcing prefixes for Layer 2 meshes.
```

lemoer (approved):
LGTM. Just a minor comment for the docs.

docs/package/gluon-radvd.rst
AiyionPrime: resolved in f549310fde133eb97d86c5cfc5cda53f28118cd6

PR #3571, opened 2025-08-26
mesh-layer3-common: Align fixes to mesh-batman-adv

AiyionPrime:
While working on #3570, I found two unnecessary discrepancies between the variants of the arguments file in the two mesh packages.


AiyionPrime:
After the merge of this, the only deviation is setting the defaultroute, which is desired for babel and not for batman-adv.

I'm fine with leaving these as two different files for the moment; other solutions might require more code.

herbetom (approved):
LGTM

PR #3578, opened 2025-09-04
mesh-vpn: Remove `pubkey_privacy` setting for WireGuard

AiyionPrime:
> which implies a usefulness for WireGuard.
> 
> Contrary to fastd WireGuard itself is identity hiding; which means: As long as neither of the two peers is compromised, the identity if the initiator will not be revealed to third parties.
> 
> The claim can be found on https://www.wireguard.com/protocol/ under 'Key Exchange and Data Packets':
> `- Identity Hiding`
> 
> The explanation of it and the whole exchange in detail can be found in this blog post:
> https://icandothese.com/docs/tech/networking/noise_and_wireguard/#identity-hiding
> 
> In order to support mixed setups, where one domain is using e.g. fastd - where the privacy setting is useful - and another domain is using WireGuard, this commit removes the limitation to only set `pubkey_privacy` in the `site.conf`.

### DIFF
--- a/docs/features/multidomain.rst
+++ b/docs/features/multidomain.rst
@@ -134,7 +134,6 @@ site.conf only variables
   - setup_mode.skip
   - autoupdater.branch
   - mesh_vpn.enabled
-  - mesh_vpn.pubkey_privacy
   - mesh_vpn.bandwidth_limit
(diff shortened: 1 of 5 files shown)


AiyionPrime:
thx to @s-2 for the typo

PR #3593, opened 2025-10-18
gluon-setup-mode: fix non-interactive dropbear sessions

neocturne:
Our ash wrapper was passing `--login` unconditionally, however this is incorrect for non-interactive sessions - it does not only result in the OpenWrt banner to be included in the output of commands like `ssh node ... > output.log`, but also completely breaks scp.

The correct behavior is determined by the value of argv[0] set by dropbear, which is '-ash' for interactive sessions and 'ash' otherwise, and interpreted by busybox accordingly. This however poses a problem for our shell-based wrapper: argv[0] is lost due to the indirection when passing the script to its interpreter defined by the shebang.

To solve the problem, replace our shell wrapper with a tiny C program, which preserves the whole argv array when running ash.

Fixes: 18879e28fe74 ("gluon-setup-mode: namespace-based implementation for password-less login")

### DIFF
--- a/package/gluon-setup-mode/Makefile
+++ b/package/gluon-setup-mode/Makefile
@@ -10,7 +10,7 @@ include ../gluon.mk
 define Package/gluon-setup-mode
   TITLE:=Setup mode
   DEPENDS:=+gluon-core +gluon-lock-password +ubus +dnsmasq-full \
-    +!BUSYBOX_CONFIG_NSENTER:nsenter +!BUSYBOX_CONFIG_UNSHARE:unshare
+    +!BUSYBOX_CONFIG_UNSHARE:unshare
 endef
 
 define Package/gluon-setup-mode/description
@@ -41,6 +41,8 @@ define Package/gluon-setup-mode/install
 	for link in $(init_links); do \
 		$(LN) "/etc/init.d/$$$${link:3}" "$(1)/lib/gluon/setup-mode/rc.d/$$$${link}"; \
 	done
+
+	$(INSTALL_BIN) $(PKG_BUILD_DIR)/ash-wrapper $(1)/lib/gluon/setup-mode/dropbear/ash
 endef
 
 $(eval $(call BuildPackageGluon,gluon-setup-mode))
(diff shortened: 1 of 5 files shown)


AiyionPrime:
I'll be looking into it tomorrow, when I'm back again.

maurerle:
On debian 13 trixie, `OpenSSH_10.0p2 Debian-7, OpenSSL 3.5.1 1 Jul 2025` - I have no problems calling:

`ssh node gluon-info > output.txt` - and also no additional headers.
running `scp node:/tmp/board.json .` works as well just fine.

Running on latest main branch without this fix, but with `openssh-sftp-server - 9.9_p2-r1` installed on the node as an additional package (initially to avoid requiring `scp -O` for legacy scp without sftp).

neocturne:
@maurerle Hmm, is it unclear to me how the client version could have an effect (but I'm using OpenSSH_10.2p1, OpenSSL 3.6.0 1 Oct 2025 on Arch Linux). I have only tested `scp` with `-O`. Note that this issue only affects dropbear in config mode.

maurerle:
Oh okay - I did not get that this issue only happens in config-mode.
That's then why I never encountered it.

Djfe:
good catch! Would've bitten us later on, if we hadn't catched this before the release.

rotanid:
@AiyionPrime maybe you might have time soon to review this? :)

AiyionPrime:
Indeed, sorry for the delay and thanks for your patience.

maurerle:
@AiyionPrime do you have time for a review of this PR in the next few days?

maurerle (approved):
This replaces the previous ash-wrapper with a C-script.

We therefore do not use `nsenter` and remove it from the build options (the `BUSYBOX_CONFIG_NSENTER:nsenter`).
Took me a little bit to understand how the previous solution worked.

AiyionPrime:
Thanks @blocktrron.

I'm unsure how much I'll be able to contribute in the next weeks,
commuting and the new job consume more time than anticipated -.-'

PR #3595, opened 2025-10-20
ipq40xx-generic: add support for Linksys WHW01 / VLP01

achterin:
  - [x] Web interface
  - [ ] TFTP
  - [ ] Other: <specify>
        (`lua -e 'print(require("platform_info").get_image_name())'`) -> `linksys-whw01`
- Wired network
    - port closest to the power jack is configured as WAN.
- LED mapping
  - Power/system LED
  - Radio LEDs
    - no radio LEDS
  - Switch port LEDs
    - no switch LEDs
- Docs:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -238,6 +238,7 @@ ipq40xx-generic
 * Linksys
 
   - EA6350 (v3)
+  - VLP01
   - WHW01
 
 * NETGEAR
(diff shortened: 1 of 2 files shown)


achterin:
The module update is required for the correct label-mac assignment and for the status led to work during upgrade/failsafe.

The Linksys VLP01 is the same device with a different label. It uses the same image as the WHW01 (the header also just says WHW01) and shares the FCCID.

rotanid:
thanks @achterin !
let's separate the bump though, i created #3597

achterin:
removed the module bump and rebased on main

PR #3596, opened 2025-10-22
ath79-generic: additional reason for BROKEN device Archer C58

### LINKED ISSUE #3425: [ath79][ipq40xx] verify 5GHz on devices using ath10k QCA9888
This a crossreference to an upstream ticket.
The nvmem-cell rework caused certain devices to not detect (some of) their 5GHz radios correctly. Specifically devices with a chip requiring QCA9888 firmware (wifi5 wave2).
https://github.com/openwrt/openwrt/issues/14541

Affected are likely (untested so far):
TP-Link
- EAP225 variants (outdoor for example)
- Archer C58 v1, C59 v1, C6 v2

ZTE MF281

OpenMesh A62

Plasma Cloud PA2200

(and Linksys Velop WHW03 v1/v2, which we were planning on adding, though I haven't noticed the issue on these while we were testing)

rotanid:
closes #3425

### DIFF
--- a/targets/ath79-generic
+++ b/targets/ath79-generic
@@ -434,7 +434,7 @@ device('tp-link-archer-c5-v1', 'tplink_archer-c5-v1', {
 
 device('tp-link-archer-c58-v1', 'tplink_archer-c58-v1', {
 	packages = ATH10K_PACKAGES_QCA9888,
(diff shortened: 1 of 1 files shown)


rotanid:
i cancelled the CI build as it's only about a comment.

PR #3598, opened 2025-10-23
patches: base-files: disable wphy renaming

### LINKED ISSUE #3482: mt7622: new installs have no wifi-config
### Bug report

**What is the problem?**
This is an issue to track the following upstream PR for our next [milestone](https://github.com/freifunk-gluon/gluon/milestone/24)
https://github.com/openwrt/openwrt/pull/17821

Gluon should be able to generate a working wifi-config on this target

**Gluon Version:**
main

**Site Configuration:**
[irrelevant](https://github.com/ffac/site/tree/v2025.1.x)

**Custom patches:**
[irrelevant](https://github.com/ffac/site/tree/v2025.1.x/patches)

maurerle:
The phy renaming happens too late for us in openwrt-24.10 and should be postponed until the next release therefore. Disable it therefore, so that mt7622 uses phy0 and phy1 as well.

Tested on a ubiquiti-unifi-6-lr-v3.

If the device was running a broken renamed version of gluon-main/openwrt-24.10 before, one has to remove the wireless section (and regenerate using `wifi config`). After this, a `gluon-reconfigure` works.

Tested:
- config-mode toggles work, private wifi settings work
- private wifi works
- wifi works

fixes #3482 
blocks #3431

### DIFF
--- a/patches/openwrt/0010-base-files-disable-wiphy-renaming.patch
+++ b/patches/openwrt/0010-base-files-disable-wiphy-renaming.patch
@@ -1,0 +1,23 @@
+From: Florian Maurer <f.maurer@outlook.de>
+Date: Thu, 23 Oct 2025 14:22:01 +0200
+Subject: base-files: disable wiphy renaming
+
+The phy renaming is currently only done when the wifi iface is used.
+This is too late for gluon - disable it therefore, so that all targets use phyX as well.
+This is currently needed for the mediatek-mt7622 target.
+Only needed for openwrt-24.10 branch as later releases will be able to handle renaming early enough.
+
+Signed-off-by: Florian Maurer <f.maurer@outlook.de>
+
+diff --git a/package/base-files/files/lib/functions/uci-defaults.sh b/package/base-files/files/lib/functions/uci-defaults.sh
+index 5293ce12c0e08f2d211999e4f00fcbac6639a774..c55ff3fd681150adc9456b992a3c315c0950fb4f 100644
+--- a/package/base-files/files/lib/functions/uci-defaults.sh
++++ b/package/base-files/files/lib/functions/uci-defaults.sh
+@@ -757,6 +757,7 @@ ucidef_set_poe() {
+ }
+ 
+ ucidef_add_wlan() {
++	return 0
+ 	local path="$1"; shift
+ 
+ 	ucidef_wlan_idx=${ucidef_wlan_idx:-0}


neocturne:
I don't think the commit title should mention the target name, as the change affects all targets (mt7622 just happens to be the one in Gluon). The target name can be included in the further explanation in the commit message.

Djfe:
Please add a mention into the patch that OpenWrt 24.10.x is the last major OpenWrt release that needs this fix (25.x or 26.x or however it will be called will come with a proper fix)
edit: mostly as documentation for ourselves in order to remove the patch once openwrt main is added into gluon next and then gluon main at some point

neocturne:
- The title of the patch should be "base-files: disable wphy renaming"
- There is a typo in the target name ("mediathek" -> "mediatek")
- "will handle renaming" -> "will be able to handle renaming" would be more accurate (there is still a lot of work necessary in Gluon to get the renaming to work)
- The Gluon commit should have the same commit message as the OpenWrt patch

rotanid:
> * The title of the patch should be "base-files: disable wphy renaming"
> 
>     * There is a typo in the target name ("mediathek" -> "mediatek")
> 
>     * "will handle renaming" -> "will be able to handle renaming" would be more accurate (there is still a lot of work necessary in Gluon to get the renaming to work)
> 
>     * The Gluon commit should have the same commit message as the OpenWrt patch

@neocturne  he addressed those, can you check again and maybe merge?

Djfe:
too early, I'll report back early on Monday whether this worked

maurerle:
This is already tested as mentioned in the PR description.
@neocturne what do you think?

maurerle:
This Patch removes the phy renaming for mt7622 devices (and other targets which rename phy names).
The issue for it is described in: https://github.com/freifunk-gluon/gluon/issues/3482

A proper fix for this must be added in openwrt. There was a proposal: https://github.com/openwrt/openwrt/pull/17821 but https://github.com/openwrt/openwrt/commit/4b7323e3bfa82ed8279f1e757f82d4961d8e3ae7 made it into openwrt-v25.12

Unfortunately, gluon based on openwrt-25.12 still does not work without the here mentioned patch.
As described in: https://github.com/freifunk-gluon/gluon/pull/3757#issuecomment-4721349850

A fix for this exists with: https://github.com/openwrt/openwrt/pull/23860

PR #3601, opened 2025-10-28
gluon-core: add override setting for htmode per radio

### LINKED ISSUE #2947: allow configuration of wlan radio bandwidth
While I can configure it on the device itself, currently only 20 MHz of channel bandwidth is used by Gluon and is not configurable per-band in the site configuration.

To achieve this, the site should be able to indicate the desired maximum bandwith (20/40/80/160) while Gluon should try to pair the widest supported channel bandwith with the most efficient HT-mode.

We mainly have to extend https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless#L64 and update https://github.com/freifunk-gluon/gluon/blob/master/package/gluon-core/check_site.lua#L34

### LINKED ISSUE #3550: RFC: configure wifi through gluon config

maurerle:
This is a follow up PR from #3563 and implements the missing parts of #3550

Fixes #3550
Fixes #2947 

This allows to set a custom bandwidth, which should be supported by all devices with that band.

add a txpower and htmode per radio section, which can be set using
`uci set gluon.wradio0.htmode="HE40"`

Similarly, `gluon.wradio0.txpower` can be adjusted.
This is based on the radio name which should stay the same once created with wifi config
The migration of the existing txpower value takes already place in `022-wireless-roles`.

There might be a better option to tie this setting to, but radio0 seems more consistent than "phy0" as we have seen.


package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
```diff
@@ -66,6 +67,23 @@ local function migrate_wifi_to_roles(band)
...
+
+		uci:section('gluon', 'wireless_radio', radio_name, {
+		})
+
+		local txpower = uci:get('wireless', radio_name, 'txpower')
+		if txpower ~= nil then
+			uci:set('gluon', radio_name, 'txpower', txpower)
+		end
+
+		-- remove in 2028 or when upgrades from v2025 are unsupported
```
rotanid: i'm not sure i understand this sentence. and shouldn't we make it more obvious to mark it for later removal with a more eye-catching or searchable keyword?

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
```diff
@@ -79,7 +97,7 @@ local function migrate_wifi_to_roles(band)
 end
 
 for band, _ in pairs(default_wifi_roles) do
-	if not uci:get('gluon', 'band_' .. band) and wireless.device_uses_band(uci, band) then
+	if not uci:get('gluon', 'band_' .. band) and wireless.device_supports_band(uci, band) then
 		-- should only create defaults when upgrades are only possible from v2025.1 onwards
```
rotanid: i know it's not part of this PR, but i'm not sure i understand this sentence.

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
maurerle: Yes, this is similar as the sentence below.
Actually both can be removed when upgrades from v2024 are not supported anymore.
Do you have a proposal? Maybe `-- remove_after v2024`?

I updated the strings to be consistent when searching for `>=v2025.1.x`.

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
```diff
@@ -79,7 +79,7 @@ local function migrate_wifi_to_roles(band)
 end
 
 for band, _ in pairs(default_wifi_roles) do
-	if not uci:get('gluon', 'band_' .. band) and wireless.device_uses_band(uci, band) then
+	if not uci:get('gluon', 'band_' .. band) and wireless.device_supports_band(uci, band) then
```
neocturne: This commit still doesn't make sense until we create the infrastructure to switch radios between bands. We should remove it for now.

package/gluon-core/luasrc/usr/lib/lua/gluon/wireless.lua
```diff
@@ -159,4 +159,21 @@ function M.device_supports_band(uci, band)
...
+	local roles = uci:get_list('gluon', 'band_' .. band, 'role')
+	if next(radio_roles) == nil then
+		return roles
+	else
+		return radio_roles
+	end
+end
+
+function M.radio_hop_penalty(uci, radio)
+	return uci:get('gluon', 'band_' .. radio.band, 'batadv_hop_penalty') or 0
```
neocturne: This helper should be called `radio_option` and accept the option name as an argument - the string `batadv` should not appear in `gluon-core`. The `or 0` should also be handled by the caller.

The same helper will then be used for htmode/txpower.

(I appear to have missed the batadv reference in `200-wireless` as well... we can leave it like that for now, but we should move it into the gluon-mesh-batman-adv package in the future)

package/gluon-core/luasrc/usr/lib/lua/gluon/wireless.lua
```diff
@@ -159,4 +159,21 @@ function M.device_supports_band(uci, band)
 	return ret
 end
 
+function M.radio_roles(uci, radio)
+	local band = radio.band
+	local radio_name = radio['.name']
+	local radio_roles = uci:get_list('gluon', radio_name, 'role')
```
neocturne: The whole point of moving this config from `/etc/config/wireless` to `/etc/config/gluon` was to make it independent of the radio indices. We can't add sections with radio names to `gluon` now, they should be identified by their paths (or PHY names in the future, when renaming works...)

It may make sense to disallow this for now and wait for working PHY renaming.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -71,22 +71,37 @@ local function get_htmode(radio)
...
-
 	local htmode = uci:get('gluon', radio['.name'], 'htmode')
 	if htmode ~= nil then
 		return htmode
 	end
 
+	local bandwidth
+
+	if radio.band == '2g' then
+		bandwidth = site.wifi24.bandwidth('20')
```
neocturne: Types are inconsistent - `check_site.lua` enforces a number, but here the default value is a string.

neocturne: Also, shouldn't this be called `channel_width`?

neocturne (changes requested):

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
neocturne: Also, the abstraction should be the same in `site.conf` and `/etc/config/gluon` (with values 20, 40, ...). The conversion to `HE20` etc. should happen when applying the configuration to `/etc/config/wireless.

neocturne:
I also remember meshing getting broken because of dynamic switching of primary and secondary channel by hostapd with HT40 modes (if another radio is seen on the primary channel, which will... always happen when meshing), which could only be workarounded with the potentially non-compliant `noscan` setting. We must make sure that this issue has been resolved before allowing wider channels.

maurerle:
Thanks! I did not have any issue with this so far. Tested with:
20-20 MHz
20-40 MHz
20-80 MHz
40-80 MHz

I guess that I should have run into this issue, if it would still be a problem.
I deployed this on around 30 nodes already.

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
```diff
@@ -66,6 +67,23 @@ local function migrate_wifi_to_roles(band)
 				end
 			end
 		end
+
+		uci:section('gluon', 'wireless_radio', radio_name, {
+		})
+
+		local txpower = uci:get('wireless', radio_name, 'txpower')
+		if txpower ~= nil then
+			uci:set('gluon', radio_name, 'txpower', txpower)
```
maurerle: This sets the txpower by radio_name (radioX).
This is compatible to how it currently is set with:

`uci:get('wireless', radio, 'txpower')`
`'gluon', 'wireless', 'outdoor_' .. radio .. '_htmode'`

@neocturne maybe we should also migrate this to phynames (once they are stable..?)

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
maurerle: I would like to have the option to set 'HT20' instead of 'HE20' in the /etc/config/gluon as well..?
I only need this to debug an issue from time to time and could adjust /etc/config/wireless myself.
So I guess it is fine to only have the "highest" mode configured in gluon

maurerle:
The `uci show wireless.radioX.path` is always the same to determine a radio. So this should be used as a mapping from radioX names instead.

I guess I have to somehow implement this somewhat like this:

1. If no gluon.wradioX exists, create according entries from the existing /etc/config/wireless and migrate the existing settings
2. then remove the existing /etc/config/wireless
3. run `wifi config` to generate new radio names
4. adjust the /etc/config/wireless with settings in /etc/config/gluon, using the configured parts

I could also just create the wifi-device section myself, but I don't know how I will know about phy renaming then..?

I guess this works fine, but also is a little bit of work.
This removes all "preserve_channels" options as well

I wonder how this could adhere to wifi path changes like this:
`migrate_radio 'platform/18000000.wifi' 'platform/soc/18000000.wifi'`
https://github.com/openwrt/openwrt/blob/9934c716edbc68d9d223c6290fd7f073d0715351/target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/05-wifi-migrate#L56

package/gluon-core/luasrc/usr/lib/lua/gluon/wireless.lua
maurerle: I fixed this in the latest push (not cleaned up though)

package/gluon-core/luasrc/lib/gluon/upgrade/022-wireless-roles
maurerle: I did migrate this to path as well. So resolving this

maurerle:
Until I have some guidance to this change, the htmode site setting (#3659) is unrelated from this and should work without introducing regrettable changes.
(unlike this PR which really should get the identification of a radio right throughout all cases to not change this again)

PR #3602, opened 2025-10-29
add support for 6ghz radios - mediatek-filogic: acer-vero-w6m

### LINKED ISSUE #2906: WiFi 6E / 6GHz WiFi
6 GHz WiFi is a thing apparently

Are there plans to support it? Does OpenWRT support it yet?

maurerle:
This PR adds support for 6GHz radios as available in the Acer Vero W6m.
Addition of configurable channel_width as proposed in #3601 would be very good.

But it works using 20MHz channel width as well.
Closes #2906

### DIFF
--- a/package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
+++ b/package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
@@ -236,8 +236,12 @@
 					<tbody>
 						<tr><th><%:Total%></th><td><%= statistics('clients/total') %></td></tr>
 						<tr><th><%:Wireless 2.4 GHz%></th><td><%= statistics('clients/wifi24') %></td></tr>
+					<% if wireless.device_uses_band(uci, '5g') then -%>
 						<tr><th><%:Wireless 5 GHz%></th><td><%= statistics('clients/wifi5') %></td></tr>
+					<%- end %>
+					<% if wireless.device_uses_band(uci, '6g') then -%>
 						<tr><th><%:Wireless 6 GHz%></th><td><%= statistics('clients/wifi6') %></td></tr>
+					<%- end %>
 					</tbody>
 				</table>
 				<% if radios[1] then -%>


package/gluon-core/check_site.lua
```diff
@@ -48,6 +48,12 @@ for _, config in ipairs({'wifi24', 'wifi5'}) do
 			need_one_of({config, 'channel'}, channels)
 			need_chanlist({config, 'outdoor_chanlist'}, channels, false)
 			need_one_of({config, 'outdoors'}, {true, false, 'preset'}, false)
+		elseif config == 'wifi6' then
+			local channels = {
+				1, 3, 5, 7, 9, 11, 13, 15, 17, 19, 21, 23, 25, 27, 29, 31, 33,
+				35, 37, 39, 41, 43, 45, 47, 49, 51, 53, 55, 57, 59, 61, 63,
+				65, 67, 69, 71, 73, 75, 77, 79, 81, 83, 85, 87, 89, 91, 93 }
```
maurerle, line 56: These are taken from here:
https://en.wikipedia.org/wiki/List_of_WLAN_channels#6_GHz_(802.11ax_and_802.11be)

maurerle:
~~Still needs support for respondd wifi6 section~~

Ready for a first review

maurerle:
Comment from gluon meetup: The respondd implementation should be changed to only add the clients6 field to devices which have a 6ghz radio so that the size of the respondd messages is not increased

rotanid (approved):

rotanid:
FYI: my approval is limited to the non-C parts, not really any C-knowledge here

maurerle:
This PR is deployed on 1120 Nodes in FFAC: https://map.aachen.freifunk.net/#/en/map?node.firmware=v2025.1.0-1
And working well. Though I would still like to get a review on the C parts.

package/gluon-respondd/src/respondd-statistics.c
```diff
@@ -214,13 +251,17 @@ static void count_iface_stations(size_t *wifi24, size_t *wifi5, const char *ifna
 			continue;
 
 		(*wifi)++;
```
rotanid: why not check the pointer here also while it is checked in line 255 for owe case?

maurerle: yes, if the `freq` would ever be something unknown, we need this check.
Fixed - thanks! :)

maurerle:
any obligations against backporting this to v2025.1.x? @neocturne

PR #3609, opened 2025-11-10

herbetom:


herbetom:
Would be nice if someone with the necessary permissions could retrigger the failed "build_firmware (ath79-nand)" workflow run.
My only option would be to retrigger them all which wouldn't be really resource friendly.

It seems to be just a GitHub "Internal Server Error" glitch.

PR #3610, opened 2025-11-10
mt7915: detect and purge stuck PLE queues

blocktrron:
When stations leave the area of service while there are transmissions
pending inside the hardware, these queues can become stuck.

This results in missing TXRX free / TXS events to the host. Also the
throughput of other connected stations substantially decreases and
latency massively increases.

Periodically poll the queue state of connected station from hardware and
purge queues detected as stuck.

### DIFF
--- a/patches/openwrt/0013-mt7915-detect-and-purge-stuck-PLE-queues.patch
+++ b/patches/openwrt/0013-mt7915-detect-and-purge-stuck-PLE-queues.patch
@@ -1,0 +1,336 @@
+From: David Bauer <mail@david-bauer.net>
+Date: Tue, 2 Dec 2025 01:25:42 +0100
+Subject: mt7915: detect and purge stuck PLE queues
+
+When stations leave the area of service while there are transmissions
+pending inside the hardware, these queues can become stuck.
+
+This results in missing TXRX free / TXS events to the host. Also the
+throughput of other connected stations substantially decreases and
+latency massively increases.
+
+Periodically poll the queue state of connected station from hardware and
+purge queues detected as stuck.
+
+Signed-off-by: David Bauer <mail@david-bauer.net>
+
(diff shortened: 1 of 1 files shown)


rotanid:
@blocktrron are there related issues in gluon's or openwrt's tracker?

blocktrron:
@rotanid See https://github.com/openwrt/mt76/issues/1009

maurerle:
I can confirm this solves the problem on the NWA55AXE.
I haven't seen this issue in combination with mesh interfaces.
I do have a mesh connection which varies between 0% and 4% (due to a large distance) on the NWA55AXE, but this does not seem to cause issues (as the issue does not appear less often if I disable mesh).

rotanid:
@blocktrron we discusses this PR shortly in our meetup and came to the conclusion, that if you agree and can fix the CI issue, we would include it in the upcoming release

blocktrron:
@rotanid  I'd also consider to backport this patch.

That being said, I would like to have some more feedback from other communities about this patch. @maurerle did help me to find two rough edges, but if someone tests this patch in a busy environment I deeply appreciate it.

To back with (some) data, traffic graphs of my test node (pedestrian zone) look healthier since ~1 week with this patch:

image

In addition (not related to this patch) I'd like to extrapolate some information we can extract from the devices into e.g. respondd to help improve visibility for such issues.

blocktrron:
Received positive long-term feedback from @maurerle - Rebased and will merge after CI succeeds.

neocturne:
@blocktrron I'm okay with merging this in Gluon for now - although I can't really say much about the code - but I'd be even happier if we just pushed such code to OpenWrt (or further upstream) first instead of carrying it downstream.

blocktrron:
@neocturne I'm with you on this one. I send all these downstream patches upstream eventually, but having them downstream presents be with the chance to get additional feedback on it before iterating upstream.

csharper2005:
@blocktrron hi! Why are you reverted the commit?

Djfe:
> @blocktrron hi! Why are you reverted the commit?

He is working on an actual solution that is upstreamable and that branch's first commit is reverting the current workaround (in order to replace it with a solution). #3673

Issue #2756, opened 2023-01-10
software-controlled LEDs spanning multiple LAN ports (DSA)


AiyionPrime:
I think there is no need to stall #2716 much longer.

> I think we have to take a look at software-controlled LEDs spanning multiple LAN ports in conjunction with DSA, as this is currently not handled nicely upstream. I assume some other devices have lost sane handling with migrating to DSA also.

_Originally posted by @blocktrron in https://github.com/freifunk-gluon/gluon/issues/2716#issuecomment-1345694190_

Djfe:
Also: LED handles for MT7621 devices (MT7530 switch) were removed by https://github.com/openwrt/openwrt/commit/fc33c41c21362b7186aa051a2140623943fa3143

So there is only a hacky way right now to disable them if you want to by using mdio manually.
https://forum.openwrt.org/t/openwrt-support-for-wax206/127523/294?u=djfe

Why where they removed? Because registering them as mdio and firing them with netdev doesn't show the offloaded traffic between lan ports on a switch and because they cause many interrupts to be fired on mdio which reduces performance massively.
Discussion was here (it spans over several posts):
https://github.com/openwrt/openwrt/pull/10829#issuecomment-1281930122 

This is how I got aware of it:
https://github.com/openwrt/openwrt/pull/11601#issuecomment-1364692035
https://github.com/openwrt/openwrt/pull/11601#issuecomment-1365076527

OpenWRT maintainers are waiting for this to be resolved upstream (Linux), so they don't have to maintain a solution downstream that belongs upstream. The solution would be to allow setting registers on the switch to enable and disable it from there (or something like that). Right now the switch handles blinking on packets itself. (in hardware/offloaded)
This was a first try to send in a patch upstream, I think:
https://lwn.net/ml/linux-kernel/20220503151633.18760-1-ansuelsmth@gmail.com/

Issue in OpenWRT:
https://github.com/openwrt/openwrt/issues/11010

namiltd:
I created a module to support LEDs for all LAN interfaces (ledtrig-netdev-lan): https://forum.openwrt.org/t/multiple-devices-ports-status-single-led/239622/3
The patch also includes a similar module for WiFi (ledtrig-netdev-phy).
Tests and any comments are welcome.

namiltd:
Check out https://github.com/openwrt/openwrt/pull/19903
The ledtrig-network module does what you need.

namiltd:
[...]

PR #3611, opened 2025-11-12
mediatek-filogic: add cudy-wr3000e

misanthropos:
  - [x] Web interface *one needs to install a signed OWRT from cudy first*
  - [x] TFTP *needs a signed firmware*

- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [x] Should show activity
  - Switch port LEDs
        *wan will both work but waits for PR* 
        *lan has one led represening 4 lan ports of which only one of them can be assigned at a time*

- Docs:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -331,6 +331,7 @@ mediatek-filogic
   - AP3000 Outdoor (v1)
   - TR3000 (v1)
   - WR3000 (v1)
+  - WR3000e (v1)
 
 * D-Link
 
--- a/targets/mediatek-filogic
+++ b/targets/mediatek-filogic
@@ -47,6 +47,9 @@ device('cudy-wr3000-v1', 'cudy_wr3000-v1', {
 	factory = false,
 })
 
+device('cudy-wr3000e-v1', 'cudy_wr3000e-v1', {
+	factory = false,
+})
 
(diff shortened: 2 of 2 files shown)


maurerle:
Hi @misanthropos 
can you take a look at `openwrt/target/linux/mediatek/filogic/base-files/etc/board.d/01_leds` - you should be able to set your LED using `ucidef_set_led_netdev` somehow.
This should be patched upstream in openwrt - otherwise, this could only be added as broken.

I guess you need to specify `blue:wan-online` and `blue:lan`, just as it was done for the other WR3000 devices:

https://github.com/openwrt/openwrt/blob/501f4edb04c434ed966a9443788b2f968abca38f/target/linux/mediatek/filogic/base-files/etc/board.d/01_leds#L63

misanthropos:
WAN led should be fixed with this PR: [assign WAN netdev trigger to blue:wan](https://github.com/openwrt/openwrt/pull/20808)
Thank you @maurerle for your suggestion!

maurerle (approved):
It was mentioned on IRC that the functional lan iface LED is not considered a hard requirement for such devices.

I support handling it like this, as this requires larger fixes upstream (see #2756).

rotanid:
@misanthropos we merge this now but rely on you following up with fixing it upstream

PR #3612, opened 2025-11-14
gluon-core: add migration for Ethernet driver load order change

neocturne:
To account for the change module load order between OpenWrt 23.05 and 24.10 for some Ethernet drivers, resulting in eth0 and eth1 (and therefore LAN and WAN) interfaces to switch, also switch the LAN and WAN roles to keep the correct roles on each physical port.

Partial solution for #3576 - a proper solution should ensure such a migration won't be necessary again in the future.

### DIFF
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/019-migrate-interface-order
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/019-migrate-interface-order
@@ -1,0 +1,112 @@
+#!/usr/bin/lua
+
+local platform = require 'gluon.platform'
+local sysconfig = require 'gluon.sysconfig'
+
+local libgen = require 'posix.libgen'
+local unistd = require 'posix.unistd'
+local uci = require('simple-uci').cursor()
+
+if not platform.match('x86') then
+	return
+end
+
+if sysconfig.migrated_interface_order then
+	return
+end
+sysconfig.migrated_interface_order = ''
+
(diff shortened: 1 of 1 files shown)


maurerle:
This does not account for situations in which there are more than 2 nics.. There is at least one offloader with wan,mesh,client in ffmuc
but I guess this will be okay for 99% of installations

According to https://github.com/freifunk-gluon/gluon/issues/3576#issuecomment-3418318601
of the prior supported modules, only `r8169` and `igc` changed in order.

neocturne:
Only offloaders with more than 2 interfaces that use multiple different Ethernet drivers would not be covered. I'm somewhat reluctant to introduce even more complexity into an already fragile migration...

One alternative would be to revert the load order change, so we have more time to develop a proper path-based solution.

package/gluon-core/luasrc/lib/gluon/upgrade/019-migrate-interface-order
```diff
@@ -0,0 +1,97 @@
...
+
+local libgen = require 'posix.libgen'
+local unistd = require 'posix.unistd'
+local uci = require('simple-uci').cursor()
+
+if not platform.match('x86') then
+	return
+end
+
+if sysconfig.migrate_interface_order then
```
neocturne: Misleading name:
```suggestion
if sysconfig.migrated_interface_order then
```

neocturne:
VLAN interfaces are migrated as well now.

PR #3614, opened 2025-11-17
net/l2tp: reset skb control buffer on xmit

blocktrron:
I will send this one upstream.

But I think it should be applied to Gluon, given this can potentially DoS the tunnel by crafting specific IPv6 packets from the client network.

This might be target specific, but I've had success at breaking the tunnel by performing larger IPv6 downloads on mpc85xx, but only with a station  connected directly to the WiFi of the device establishing the VPN tunnel.

This leads to massive fragmentation on the uplink port, breaking the tunnel for everyone:

image

I'm not aware if other platforms are affected as well.

---

The L2TP stack did not reset the skb control buffer before handing off the package to the lower netdev.

In a setup with an ath10k radio and batman-adv over an L2TP tunnel massive fragmentations happen sporadically if the L2TP tunnel is established over IPv4.

L2TP might reset some of the fields in the IP control buffer, but L2TP assumes the type of the control buffer to be of an IPv4 packet, while due to it being Layer 2, this is not a given.

Clear the entire control buffer to avoid such mishaps altogether.

### DIFF
--- a/patches/openwrt/0011-net-l2tp-reset-skb-control-buffer.patch
+++ b/patches/openwrt/0011-net-l2tp-reset-skb-control-buffer.patch
@@ -1,0 +1,68 @@
+From: David Bauer <mail@david-bauer.net>
+Date: Mon, 17 Nov 2025 01:50:08 +0100
+Subject: net/l2tp: reset skb control buffer
+
+The L2TP stack did not reset the skb control buffer before handing off
+the package to the lower netdev.
+
+In a setup with an ath10k radio and batman-adv over an L2TP tunnel
+massive fragmentations happen sporadically if the L2TP tunnel is
+established over IPv4.
+
+L2TP might reset some of the fields in the IP control buffer, but L2TP
+assumes the type of the control buffer to be of an IPv4 packet, while
+due to it being Layer 2, this is not a given.
+
+Clear the entire control buffer to avoid such mishaps altogether.
+
+Signed-off-by: David Bauer <mail@david-bauer.net>
+
+diff --git a/target/linux/generic/pending-6.6/985-net-l2tp-reset-skb-control-buffer.patch b/target/linux/generic/pending-6.6/985-net-l2tp-reset-skb-control-buffer.patch
+new file mode 100644
+index 0000000000000000000000000000000000000000..5c8f66300082cac7ee8637613513bd618ded5593
+--- /dev/null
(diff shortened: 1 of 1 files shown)


patches/openwrt/0011-net-l2tp-reset-skb-control-buffer.patch
```diff
@@ -0,0 +1,68 @@
...
+++++ b/net/l2tp/l2tp_core.c
++@@ -1246,9 +1246,9 @@ static int l2tp_xmit_core(struct l2tp_session *session, struct sk_buff *skb, uns
++ 	else
++ 		l2tp_build_l2tpv3_header(session, __skb_push(skb, session->hdr_len));
++ 
++-	/* Reset skb netfilter state */
++-	memset(&(IPCB(skb)->opt), 0, sizeof(IPCB(skb)->opt));
++-	IPCB(skb)->flags &= ~(IPSKB_XFRM_TUNNEL_SIZE | IPSKB_XFRM_TRANSFORMED | IPSKB_REROUTED);
+++	/* Reset control buffer */
+++	memset(skb->cb, 0, sizeof(skb->cb));
```
neocturne, line 61: I wonder if this should be made conditional on the L2TP version, assuming IP for v2 and clearing the whole buffer for v3. I don't know the possible downsides of clearing the whole buffer for v2 however, so this may be fine.

blocktrron: I think this applies to both protocols, as IPCB assumes the packet being IPv4 while regardless of L2TP protocol version it can be either IPv4 or IPv6?

neocturne: Okay, makes sense.

blocktrron:
Looks good or do i miss something?

neocturne:
Ah oops, was comparing with patch numbers of a local branch. Going to merge when the CI has finished.

blocktrron:
Okay. Do we want to backport the fix or wait for it to be applied in the kernel stable branch?

neocturne:
Hmm, it's been a while since the last kernel bump in 23.05, so a backport may be in order... the question is, in Gluon or in OpenWrt. I would opt for OpenWrt, as the issue is not Gluon-specific.

blocktrron:
I tend to apply the patch to us downstream and just proceed with upstream kernel for OpenWrt. I assume the patch gets picked to stable kernels once approved.

Issue #3613, opened 2025-11-15
gluon-wan-dnsmasq not resolving due to needing CAP_NET_RAW for libpacketmark.so


benzea:
There is a dnsmasq instance running on port 54 to handle WAN DNS requests. To correctly route them, the outgoing socket should use setsockopt(SO_MARK) set using the libpacketmark.so preload library. However, setting SO_MARK requires `CAP_NET_RAW` (or `CAP_NET_ADMIN`).

dnsmasq does the right thing and drops the capability unless they are needed. As gluon-wan-dnsmasq does not use DHCP, etc. it doesn't actually keep the required capability. As such, the libpacketmark.so setsockopt fails with -EPERM and the wrong routing table may be used (in upstream gluon this will affect IPv6, in the parker fork it also affects IPv4).

Possible solutions:
* Use network namespaces (likely infeasible in the short term)
* Use VRFs instead of packet marking for routing purposes
* Make dnsmasq keep the privileges:
  - Simple workaround is to pass a dummy `--enable-ubus=gluon-wan-dnsmasq` parameter, as the capabilities are kept then
  - Patching dnsmasq could be an option
  - A hook in libpacketmark.so to set the capabilities seems like it would be very fragile at best

neocturne:
Decision from meetup:
- For 2025.1, we will implement one of the mentioned workarounds
- For 2026.1, a VRF will be used

neocturne:
The `--enable-ubus` workaround is probably the least invasive, only resulting in the following syslog messages once on start (as the ujail prevents the ubus connection):
```
daemon.info dnsmasq[1]: UBus support enabled: bus connection pending
daemon.err dnsmasq[1]: Cannot set UBus listeners: no connection
```

neocturne:
Workaround can be dropped if #3803 is implemented.

PR #3617, opened 2025-11-18
gluon-wan-dnsmasq: work around missing capabilities for libpacketmark + minor cleanup

neocturne:
Workaround for #3613, as a VRF-based solution is too invasive for a timely 2025.1 release.

### DIFF
--- a/package/gluon-wan-dnsmasq/files/etc/init.d/gluon-wan-dnsmasq
+++ b/package/gluon-wan-dnsmasq/files/etc/init.d/gluon-wan-dnsmasq
@@ -22,7 +22,10 @@ start_service() {
 		--keep-in-foreground \
 		--pid-file=/var/run/dnsmasq/gluon-wan-dnsmasq.pid \
 		--cache-size=0 \
+		--enable-ubus=gluon-wan-dnsmasq \
 		--resolv-file="$RESOLV_CONF"
+	# --enable-ubus must be set as a workaround to avoid dnsmasq dropping
+	# CAP_NET_RAW, which would break libpacketmark
 	procd_set_param env LD_PRELOAD=libpacketmark.so LIBPACKETMARK_MARK=1
 	procd_set_param respawn 60 5 5
 


benzea:
All the changes look sane to me.

package/gluon-wan-dnsmasq/files/etc/init.d/gluon-wan-dnsmasq
```diff
@@ -22,6 +22,7 @@ start_service() {
 		--keep-in-foreground \
 		--pid-file=/var/run/dnsmasq/gluon-wan-dnsmasq.pid \
 		--cache-size=0 \
+		--enable-ubus=gluon-wan-dnsmasq \
 		--resolv-file="$RESOLV_CONF"
 	procd_set_param env LD_PRELOAD=libpacketmark.so LIBPACKETMARK_MARK=1
```
benzea, line 29: Please add a comment that we add `--enable-ubus` purely so that dnsmasq does not drop `CAP_NET_RAW`. And that we need this for libpacketmark.so. I guess it is reasonable to put it right above setting the LD_PRELOAD (which is also pretty close to the `--enable-ubus` anyway).

blocktrron:
This will not work on ramips-mt76x8 (and potentially other targets) as these don't have procd-ujail installed.

neocturne:
@blocktrron What exactly will not work? The ubus connection will succeed on these targets, but do you think that is a problem in spite of the unique service name?

blocktrron:
Ah okay. I misunderstood the commit message in terms of the failing ubus connection is a requirement. In that case, can you test if this works when you remove the jail from the initscript? This should resemble the situation on mt76x8.

neocturne:
@blocktrron Yep, still works. I would have been very surprised if it didn't as the success or failure of the ubus connection is irrelevant for the workaround.

blocktrron (approved):
LGTM

PR #3620, opened 2025-11-24
gluon-core: limit size of wireless buffers

blocktrron:
By default, the kernel configures the buffers for wireless PHYs to 4MB for 11n radios and 16MB for 11ac and above.

For our RAM constrained devices, this has the effect of the buffers potentially filling up to cause a OOM oops. This makes it harder to bisect issues, as non-resumed TX queues often lead to such a crash on resource-constrained boards.

Limit wireless queue size for devices with 64MB of RAM to 512kB. For 128MB RAM limit to 2MB accordingly. The 512kB was chosen to provide ~100 Mbit/s sustained traffic total to 3 clients on a more powerful board. This should totally suffice the needs of these lower-tier routers.

### DIFF
--- a/package/gluon-core/files/etc/hotplug.d/ieee80211/01-gluon-core-codel-memusage
+++ b/package/gluon-core/files/etc/hotplug.d/ieee80211/01-gluon-core-codel-memusage
@@ -1,8 +1,17 @@
 #!/bin/sh
 
+LIMIT=""
+
 if [ "${ACTION}" = 'add' ]; then
 	RAM=$(awk '/MemTotal/ {print $2}' /proc/meminfo)
-	if [ "$RAM" -le $((32*1024)) ]; then
-		echo 'fq_memory_limit 262144' > "/sys/kernel/debug/ieee80211/$DEVICENAME/aqm"
+	if [ "$RAM" -le $((64*1024)) ]; then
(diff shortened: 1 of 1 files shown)


package/gluon-core/files/etc/hotplug.d/ieee80211/01-gluon-core-codel-memusage
```diff
@@ -1,8 +1,17 @@
 #!/bin/sh
 
+LIMIT=""
+
 if [ "${ACTION}" = 'add' ]; then
 	RAM=$(awk '/MemTotal/ {print $2}' /proc/meminfo)
-	if [ "$RAM" -le $((32*1024)) ]; then
-		echo 'fq_memory_limit 262144' > "/sys/kernel/debug/ieee80211/$DEVICENAME/aqm"
```
neocturne, line 6: I was confused by this removal at first. Let's mention in the commit message that this also removes obsolete logic for  32MiB RAM devices.

Djfe:
a bit OT but on the note of improving low ram devices:
https://github.com/openwrt/openwrt/pull/19992#issuecomment-3590187999
we could adjust values for `vm.min_free_kbytes` upstream too maybe, what do you think @blocktrron ?

PR #3625, opened 2025-11-30
gluon-core: fix pubkey_privacy being in site instead of uci

### LINKED ISSUE #3624: status-page: VPN public key display ignores pubkey_privacy
https://github.com/freifunk-gluon/gluon/pull/3554 documented that it respects pubkey_privacy, but the implementation is incorrect: https://github.com/freifunk-gluon/gluon/blob/8f38662f44f5df69357ced93f166000716f018b3/package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn.lua#L48 checks a non-existent UCI value, while it is actually set in the site.

maurerle:
This readds the setting for wireguard as well, to configure the behavior on the status page

The alternative - if one does not want to have this setting for wireguard at all, would be to extend the mesh-vpn provider as done in:

https://github.com/ffac/gluon/commit/7df8f6324c458950ad4b1483ada8f4789ae40eff

From my perspective, this is the cleaner solution to not add more complexity into this feature.
What do you think about this @aiyionprime

fixes #3624

### DIFF
--- a/package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn.lua
+++ b/package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn.lua
@@ -45,7 +45,7 @@ function M.get_active_provider()
 end
 
 function M.get_enabled_public_key()
-	if M.enabled() ~= true or uci:get_bool('gluon', 'mesh_vpn', 'pubkey_privacy') ~= true then
+	if M.enabled() ~= true then
 		return nil
 	end
 
@@ -53,7 +53,9 @@ function M.get_enabled_public_key()
 
 	local pubkey
 	if active_vpn ~= nil then
-		pubkey = active_vpn.public_key()
+		if not active_vpn.pubkey_privacy() then
+			pubkey = active_vpn.public_key()
+		end
 	end
 
 	return pubkey
--- a/package/gluon-mesh-vpn-fastd/luasrc/usr/lib/lua/gluon/mesh-vpn/provider/fastd.lua
+++ b/package/gluon-mesh-vpn-fastd/luasrc/usr/lib/lua/gluon/mesh-vpn/provider/fastd.lua
@@ -73,4 +73,8 @@ function M.mtu()
 	return site.mesh_vpn.fastd.mtu()
 end
 
+function M.pubkey_privacy()
+	return site.mesh_vpn.pubkey_privacy(true)
+end
+
 return M
--- a/package/gluon-mesh-vpn-wireguard/luasrc/usr/lib/lua/gluon/mesh-vpn/provider/wireguard.lua
+++ b/package/gluon-mesh-vpn-wireguard/luasrc/usr/lib/lua/gluon/mesh-vpn/provider/wireguard.lua
@@ -43,4 +43,9 @@ function M.mtu()
 	return site.mesh_vpn.wireguard.mtu()
 end
 
+function M.pubkey_privacy()
+	-- WireGuard inherently doesn't leak the pubkey to third parties upon initiating connections.
+	return false
+end
+
 return M


package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn.lua
```diff
@@ -45,7 +46,7 @@ function M.get_active_provider()
 end
 
 function M.get_enabled_public_key()
-	if M.enabled() ~= true or uci:get_bool('gluon', 'mesh_vpn', 'pubkey_privacy') ~= true then
+	if M.enabled() ~= true or site.mesh_vpn.pubkey_privacy ~= false then
```
neocturne: This doesn't work, you need to resolve the site lookup expression by calling it:
```suggestion
	if M.enabled() ~= true or site.mesh_vpn.pubkey_privacy(false) then
```

package/gluon-mesh-vpn-wireguard/check_site.lua
```diff
@@ -8,5 +8,3 @@ end
 
 need_table({'mesh_vpn', 'wireguard', 'peers'}, check_peer)
 need_number({'mesh_vpn', 'wireguard', 'mtu'})
-obsolete({'mesh_vpn', 'pubkey_privacy'},
-'WireGuard inherently doesn\'t leak the pubkey to third parties upon initiating connections.')
```
neocturne, line 12: There is no need for this change (or the doc change). Setting `pubkey_privacy` for Wireguard should still be disallowed.

package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn.lua
maurerle: `site.mesh_vpn.pubkey_privacy(false)` would change the default to "false", while the docs say, that the default is true.

package/gluon-mesh-vpn-wireguard/check_site.lua
maurerle: I would also like to have the option to show the pubkey on the status page for wireguard.
As the default is `pubkey_privacy=true` which has to explicitly disabled, I did not see a better option than the current force push, which is cumbersome...?

neocturne: Ah, you are right, I misread something. I think your new approach is good, just move the check of the `pubkey_privacy` site option to the fastd provider.

maurerle:
Tested working on wireguard gluon-nodes in FFAC.
The status page shows the public key of the node as well.

package/gluon-mesh-vpn-core/luasrc/usr/lib/lua/gluon/mesh-vpn.lua
```diff
@@ -45,15 +45,17 @@ function M.get_active_provider()
...
+	if M.enabled() ~= true then
 		return nil
 	end
 
 	local _, active_vpn = M.get_active_provider()
 
 	local pubkey
 	if active_vpn ~= nil then
-		pubkey = active_vpn.public_key()
+		if active_vpn.pubkey_privacy and active_vpn.pubkey_privacy() == false then
```
neocturne: I don't think we need to account for providers without pubkey_privacy function, as we don't have any stable API defined for providers.
```suggestion
		if not active_vpn.pubkey_privacy() then
```

maurerle: Okay. I only know about ffmuc's wireguard, ffbs-parker and the community-tunneldigger anyway.
All of which I can handle in the community-packages repository.
https://github.com/freifunk-gluon/community-packages/pull/178

PR #3631, opened 2025-12-05
docs: user/supported_devices: move EAX11/EAX12/EAX15 to correct vendor

neocturne:
These are NETGEAR, not TP-Link.


maurerle (approved):
Thanks for fixing this!

PR #3632, opened 2025-12-05
Revert `git describe --tags` change, improve docs

neocturne:
This reverts #3310, as suggested in https://github.com/freifunk-gluon/gluon/pull/3310#issuecomment-3589149103.

### DIFF
--- a/docs/dev/build.rst
+++ b/docs/dev/build.rst
@@ -95,7 +95,7 @@ getversion.sh
     */lib/gluon/gluon-version* and */lib/gluon/site-version*.
 
     By default, this uses ``git describe`` to generate a version number based
-    on the last git tag. This can be overridden by putting a file called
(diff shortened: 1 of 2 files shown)


maurerle (approved):

grische:
@neocturne @maurerle As we cannot use annotated tags in our site repos (due to our release process), what is the recommended way to generate the scm version file?

Are we meant to run the command as part of our CI process?
```sh
git --git-dir=.git describe --tags --always --abbrev=7 --dirty=+ 2>/dev/null >.scmversion
```

If yes, I think it might be worth adding this to the docs as well.

maurerle:
As this change was already part of v2023.2.5, we should have kept it on v2025.1.x as well as there seems to be no easy way to set full tags from github releases:
https://github.com/freifunk-gluon/gluon/pull/3310#issuecomment-3702174812

PR #3635, opened 2025-12-06
treewide: wrap netlink flag value access

blocktrron:
Building Gluon based on OpenWrt 24.10 currently fails due to an implicit
cast to json_object_new_boolean.

Wrap the access in order to return a boolean int value.


neocturne:
Shouldn't the commit messages refer to OpenWrt 25.xx? We've been on 24.10 for a while...

Issue #3520, opened 2025-06-03
Warning: Xiaomi AX3200 Sysupgrade from 23.05 base and earlier to 24.10 fails


ChristianMiddendorf:
OpenWRT changes the Flash Layout for Xiaomi AX3200 (AX6)with v24.10.
Actually there is no easy way to Sysupgrade from actual Gluon to the experimental 24.1 based Version.
A downgrade from Gluon based on 24.1 to the stable Series via Sysupgrade fails, too.
See https://openwrt.org/toh/xiaomi/ax3200#upgrading_from_2305_and_earlier_to_upcoming_2410_or_snapshot

The only workaround actually is to restore the Original Firmware and installing the New Gluon from scratch.

Would be great if you find a good Solution . At least the auto updater should not crash the devices.
If crashed… TFTP recovery works.

herbetom:
This shouldn't really cause issues with Gluon apart from devices just not updating.

With the autopdater only minor compat version changes are ignored. The process therefore should stop around here: 

https://github.com/freifunk-gluon/packages/blob/bae34a0f53b25dbd691c57c8ab2d0b7cfe811517/admin/autoupdater/src/autoupdater.c#L396

AiyionPrime:
@ChristianMiddendorf Lets start with the easy part:
Gluon does not support downgrades.

And for the other part, I think not crashing is the only thing we can offer at this point.
There won't be a feasible way to autoupdate this.

You'd need a two staged update process, where your device loads an initramfs, puts its config somewhere(tm) for safekeeping, then updates to the new firmware with the new flash layout and restores its config from somewhere(tm).

We encountered this before with e.g. the EdgeRouter X, which needs layout changes every so often (in fact it might be due for another one soon?) and with our current stack there is little we can do about it.

As @rotanid put it by adding the respective labels, we should mark this as known issue upon Releasing the first affected firmware. But beside that there's little we can do.

If I misread your issue and the device does in fact crash while updating to the latest master, let me know and I'll investigate further.

AiyionPrime:
I hope you don't mind: I updated the title to match the projects scope.

rotanid:
device is currently not in gluon:
https://github.com/freifunk-gluon/gluon/commit/35b70fcea274b847cc1d46bd058868312748790c

ChristianMiddendorf:
[...]

PR #3636, opened 2025-12-06
docs: releases: add v2025.1 release notes

neocturne:
TODO:

- ~https://github.com/freifunk-gluon/gluon/issues/3520 needs to be mentioned~
- What is the status of https://github.com/freifunk-gluon/gluon/issues/3481 with the latest fixes?

### DIFF
--- a/docs/releases/v2025.1.rst
+++ b/docs/releases/v2025.1.rst
@@ -1,0 +1,428 @@
+Gluon 2025.1
+============
+
+Major updates and notices
+-------------------------
+
+This release is based on the latest OpenWrt 24.10 release branch. It ships with Linux kernel 6.6.y,
+wireless-backports 6.12.y and batman-adv 2024.3.
+
+Updates are only supported from v2022.1 and later.
+
+The following devices can't be updated automatically due to breaking changes in OpenWrt, requiring
+manual steps to adjust the flash layout:
+
+- Ubiquiti EdgeRouter-X (`upgrade instructions <https://github.com/darkxst/erx-migration>`__)
+- Xiaomi AX3200
+  (`upgrade instructions <https://openwrt.org/toh/xiaomi/ax3200#upgrading_from_2305_and_earlier_to_upcoming_2410_or_snapshot>`__)
+
+Added hardware support
+----------------------
+
+ath79-generic
+~~~~~~~~~~~~~
+
+- Buffalo
+
+  - WZR-HP-G450H / WZR-450HP
+
+- D-Link
+
+  - DAP-2680 A1
+  - DAP-2695 A1
+
+- GL.iNet
+
+  - GL-AR300M16
+
+- TP-Link
+
+  - CPE710 v2
+
+- Ubiquiti
+
+  - NanoBeam 5AC 19 (XC) [#readded]_
+  - NanoBeam M5 (XW) [#readded]_
+  - NanoStation Loco M2/M5 (XW) [#readded]_
+  - NanoStation M2/M5 (XW) [#readded]_
+
+.. [#readded]
+  This previously supported device had been removed in an earlier release and was reintroduced in
+  v2025.1.
+
+ipq40xx-generic
+~~~~~~~~~~~~~~~
+
+- Extreme Networks
+
+  - WS-AP3915i
+
+- Linksys
+
+  - MR8300
+  - VLP01
+  - WHW01
+  - WHW03 v1, v2
+
+- NETGEAR
+
+  - RBR50 v1
+  - RBS50 v1
+  - SRS60
+  - SRR60
+
+ipq806x-generic
+~~~~~~~~~~~~~~~
+
+- Ubiquiti
+
+  - UniFi AC HD
+
+lantiq-xrx200
+~~~~~~~~~~~~~
+
+- AVM
+
+  - FRITZ!Box 7430
+
+lantiq-xrx200_legacy
+~~~~~~~~~~~~~~~~~~~~
+
+This is a new target.
+
+The following device previously found in the *lantiq-xrx200* target is now built as part of
+*lantiq-xrx200_legacy*:
+
+- TP-Link
+
+  - TD-W8970 v1
+
+mediatek-filogic
+~~~~~~~~~~~~~~~~
+
+- ASUS
+
+  - RT-AX52
+  - TUF-AX6000
+
+- Cudy
+
+  - AP3000 Outdoor v1
+  - TR3000 v1
+  - WR3000e v1
+
+- D-Link
+
+  - AQUILA PRO AI M30 A1
+  - AQUILA PRO AI M60 A1
+
+- GL.iNet
+
+  - GL-MT2500
+
+- OpenWrt
+
+  - OpenWrt One
+
+- Wavlink
+
+  - WL-WN573HX3
+
+- Xiaomi
+
+  - AX3000T (Stock, ubootmod)
+
+mediatek-mt7622
+~~~~~~~~~~~~~~~
+
+- NETGEAR
+
+  - WAX206
+
+mvebu-cortexa53
+~~~~~~~~~~~~~~~
+
+This is a new target.
+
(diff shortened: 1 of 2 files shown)


maurerle:
Looks good.

- [ ] #3620 is currently missing in the bugfix/stability section.
- [x] I would like to mention that devices with more than two radios (typically triple radio) can now get support as well as a new feature?
- [ ] Mention mesh_nolearn and adjusted basic rate in mesh which improves compatibility for mesh between different devices
- [x] And the problem that Ubiquiti ER-X can not be upgraded to 24.10 without reinstall: https://github.com/darkxst/erx-migration

docs/releases/v2025.1.rst
```diff
@@ -0,0 +1,394 @@
...
+
+Removed Features
+----------------
+
+.. _#3105: https://github.com/freifunk-gluon/gluon/pull/3105
+.. _#3109: https://github.com/freifunk-gluon/gluon/pull/3109
+
+- Tunneldigger Mesh VPN support has been dropped (`#3109`_)
+
+  Tunnerdigger support was deprecated in Gluon v2023.2 and has been subsequently moved to
```
DasSkelett:
```suggestion
  Tunneldigger support was deprecated in Gluon v2023.2 and has been subsequently moved to
```

T-X:
I think it could make sense to mention the newly added batman-adv multicast packet type and its implications. Like:


> batman-adv is updated to 2024.3. This introduces a new [batman-adv multicast packet type](https://www.open-mesh.org/projects/batman-adv/wiki/Multicast-Packet-Type), which will become active once all nodes in the domain were updated to this version. Once active, more Gluon nodes than the previous limit of 16 can potentially be addressed for a specific multicast group before it is filtered. For instance in a large domain this can lead to mDNS capable services to become discoverable again.

neocturne:
@T-X Hmm, should we make it configurable whether non-broadcast multicasts are allowed or not? I find it very misleading that "gluon-ebtables-filter-multicast" does not filter multicast anymore...

Edit: Ah, it's been like this for a few years already. So definitely nothing to change just before the 2025.1 release, but I would like to reconsider our approach to multicast filtering in the future.

neocturne:
@maurerle I did not mention two of the points you listed on purpose - they seem too minor (in particular, less important than the changes introduced by the kernel and OpenWrt updates, which are not mentioned either).

docs/releases/v2025.1.rst
```diff
@@ -0,0 +1,414 @@
+Gluon 2025.1 (unreleased)
```
maurerle:
```suggestion
Gluon 2025.1
```

Remove this before merging, as we release right after - right?

docs/releases/v2025.1.rst
```diff
@@ -0,0 +1,414 @@
...
+Major updates and notices
+-------------------------
+
+This release is based on the latest OpenWrt 24.10 release branch. It ships with Linux kernel 6.6.y,
+wireless-backports 6.12.y and batman-adv 2024.3.
+
+Updates are only supported from v2022.1 and later.
+
+The following devices can't be updated automatically due to breaking changes in OpenWrt, requiring
+manual steps do adjust the flash layout:
```
maurerle:
```suggestion
manual steps to adjust the flash layout:
```

maurerle (approved):

docs/releases/v2025.1.rst
neocturne: Make sense, I have removed it.

maurerle:
Regarding the open question to mt7915e becoming deaf issue:
I never experienced this bug with over 30 devices with months of runtime on gluon main firmware, so I guess the frequency of this bug is small enough to not be considered in release planning. Though it might be beneficial to have more information to better triage it if it occurs for anyone.

neocturne:
@T-X I have added a section for the new batadv multicast packet type.

T-X:
> @T-X Hmm, should we make it configurable whether non-broadcast multicasts are allowed or not? I find it very misleading that "gluon-ebtables-filter-multicast" does not filter multicast anymore...
> 
> Edit: Ah, it's been like this for a few years already. So definitely nothing to change just before the 2025.1 release, but I would like to reconsider our approach to multicast filtering in the future.

Right. It just might become more likely for some specific networks to switch to this new behaviour, too.

Had been pondering about this for some time, as in Lübeck I also got mixed feedback. I think my current take on this goes in a similar (but not as strict) direction as with the [client isolation PR](https://github.com/freifunk-gluon/gluon/pull/2714): To me it's not in the spirit of Freifunk to strictly filter multicast by default. And it's still a very memorable moment for me from our early Freifunk days in Lübeck where Fluse put a printer+scanner at home into our network with a pizza flyer on it and anyone finding and printing their order on it would get a free pizza at our upcoming Freifunk meeting back then :D. And of course I also have strong bias towards non-strict multicast filtering because of the time I've spend to enable this :-). But similar to the client isolation PR, I wouldn't be against a site config / package option to choose the strict or non-strict filtering. I wouldn't feel inclined to implement it myself though.

On the other hand I would like to add a site config option later to add more strict broadcast flooding, to entirely disable any broadcast flooding. As I think we are now at a point where batman-adv's features would allow operating an IPv4/IPv6 network without batman-adv broadcast flooding packets.

But I also wouldn't mind renaming some packages (or adding new packages) later to have non-misleading names, I think that could make sense, indeed.

docs/releases/v2025.1.rst
```diff
@@ -0,0 +1,428 @@
...
+
+The :doc:`../package/gluon-ebtables-filter-multicast` package filters many kinds of multicast
+packets when they would be flooded through the whole mesh, but it allows packets that are only
+forwarded to a limited number of nodes. In earlier version of Gluon, this allowed multicasts with up
+to 16 destination nodes, as these were sent as separate unicasts instead of broadcasting them.
+
+batman-adv 2024.3 has introduced a new
+`multicast packet type <https://www.open-mesh.org/projects/batman-adv/wiki/Multicast-Packet-Type>`__,
+which adds support for more directed forwarding of multicast packets without duplicating them for
+each destination, allowing packets with a higher number of distinations not to be filtered by the
```
T-X: typo: distinations -> destinations

maurerle:
Hi, I would like to get this out within this year. Any chance to do so? :)
I think there is not much missing.

docs/releases/v2025.1.rst
neocturne: Thanks, fixed.

neocturne:
> Hi, I would like to get this out within this year. Any chance to do so? :)

@maurerle With the 39C3 happening, it seems very likely that we'll finalize the release this year (even though I'm not attending ^^)

fsffunk:
Non working things can be fixed in a point release sometime next year. It would make sense to release the first version of v2025 in the year 2025.

rotanid (approved):

AiyionPrime:
@neocturne @blocktrron intended to ask; would you merge this?
We're fine with the current suggestion. Thanks for its high quality!

neocturne:
From my side this is ready, but I would prefer if the merger of a PR is never its creator.

rotanid:
> Looks good.
> 
>     * [ ]  [gluon-core: limit size of wireless buffers #3620](https://github.com/freifunk-gluon/gluon/pull/3620) is currently missing in the bugfix/stability section.
> 
>     * [x]  I would like to mention that devices with more than two radios (typically triple radio) can now get support as well as a new feature?
> 
>     * [ ]  Mention mesh_nolearn and adjusted basic rate in mesh which improves compatibility for mesh between different devices
> 
>     * [x]  And the problem that Ubiquiti ER-X can not be upgraded to 24.10 without reinstall: https://github.com/darkxst/erx-migration

two of those are still not in the merged release notes, are they?

sorry i only noticed it now @neocturne @maurerle :-(

fsffunk:
Why have my post been marked as off-topic? Its on topic about the release v2025.

PR #3639, opened 2025-12-14
Update modules: bump openwrt

maurerle:
The openwrt commit:

`ath11k: fix transmit queue flushing through flush_sta implementation`

improves stability of ath11k like the Xiaomi AX3600


neocturne:
Ah, of course I only noticed after merging, but the PR title should have been something like "Update modules".

maurerle:
Thanks! Will remember for the next time :)

PR #3640, opened 2025-12-15
ramips-mt7621: add support for Cudy AP1300 Outdoor v1

ambassador86:
  - [X] Other: via intermediate firmware supplied by Cudy
        (`lua -e 'print(require("platform_info").get_image_name())'`) -> cudy-ap1300-outdoor-v1
- [x] Primary MAC address should match address on device label (or packaging) -> one above the actual printed MAC
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [X] Should show activity
  - Switch port LEDs
- Outdoor devices only:
- Docs:

Hardware:
 - SoC: MediaTek MT7621DAT
 - Flash: 16 MiB XM25QH128C
 - RAM: 128 MiB
 - WLAN: 2.4 GHz (MT7603E, 11n), 5 GHz (MediaTek MT7613BEN, 11ac)
 - Ethernet: 1x10/100/1000 Mbps LAN
 - Buttons: 1 Reset button, 1 WPS button
 - LEDs: 5x Green
 - Serial Console: unpopulated header 115200 8n1 (silkscreen on PCB)
 - Power: POE 802.3af (37-57V DC)

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -488,6 +488,7 @@ ramips-mt7621
 
 * Cudy
 
+  - AP1300 Outdoor (v1)
   - WR1300 (v1)
   - WR2100
   - X6 (v1, v2)
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua
@@ -76,6 +76,7 @@ function M.is_outdoor_device()
 		return true
 
 	elseif M.match('ramips', 'mt7621', {
+		'cudy,ap1300-outdoor-v1',
 		'wavlink,ws-wn572hp3-4g',
 		'zyxel,nwa55axe',
 	}) then
--- a/targets/ramips-mt7621
+++ b/targets/ramips-mt7621
@@ -12,6 +12,10 @@ device('asus-rt-ax53u', 'asus_rt-ax53u')
 
 -- Cudy
 
+device('cudy-ap1300-outdoor-v1', 'cudy_ap1300-outdoor-v1',{
+	factory = false,
+})
+
 device('cudy-wr1300-v1', 'cudy_wr1300-v1', {
 	factory = false,
 	manifest_aliases = {


ambassador86:
Actually no interface has the mac address written on the label. It is off by one. The test device is here https://map.chemnitz.freifunk.net/#!/de/map/80afca5f9b99

rotanid:
> Actually no interface has the mac address written on the label. It is off by one. The test device is here https://map.chemnitz.freifunk.net/#!/de/map/80afca5f9b99

@blocktrron @maurerle how to deal with this?

maurerle:
@ambassador86 is the MAC 1 too high or too low?

Can you try to use `<&macaddr_bdinfo_de00 0>;` for gmac0 here instead - this should fix the number being too high:
https://github.com/openwrt/openwrt/blob/8fb644e9e8416df0d7051d8e35b57b06a030c05e/target/linux/ramips/dts/mt7621_cudy_ap1300-outdoor-v1.dts#L195

Looks like this should be fixed in openwrt..?

rotanid:
> @ambassador86 is the MAC 1 too high or too low?
> 
> Can you try to use `<&macaddr_bdinfo_de00 0>;` for gmac0 here instead - this should fix the number being too high: https://github.com/openwrt/openwrt/blob/8fb644e9e8416df0d7051d8e35b57b06a030c05e/target/linux/ramips/dts/mt7621_cudy_ap1300-outdoor-v1.dts#L195
> 
> Looks like this should be fixed in openwrt..?

what about these two lines then?
https://github.com/openwrt/openwrt/blob/8fb644e9e8416df0d7051d8e35b57b06a030c05e/target/linux/ramips/dts/mt7621_cudy_ap1300-outdoor-v1.dts#L178
https://github.com/openwrt/openwrt/blob/8fb644e9e8416df0d7051d8e35b57b06a030c05e/target/linux/ramips/dts/mt7621_cudy_ap1300-outdoor-v1.dts#L188
how are they related to what you linked to?

rotanid:
> @ambassador86 is the MAC 1 too high or too low?

his PR says "one above the actual printed MAC" - so 1 too high like in #3663

maurerle:
This `macaddr_bdinfo_de00` node defines an NVMEM cell in the device tree that exposes a base MAC address stored in flash at an offset.
The base MAC is then incremented by an index.

As the usage starts with index 1 - the base mac is not assigned on any interface.
I would expect to see the base mac (index 0) as gmac0.
Using the incremented macs (+1 and +2) for the other radios is fine and does not affect gluon.
So one should only need to adjust the `gmac0` line in:
https://github.com/openwrt/openwrt/blob/8fb644e9e8416df0d7051d8e35b57b06a030c05e/target/linux/ramips/dts/mt7621_cudy_ap1300-outdoor-v1.dts#L195

ambassador86:
Sorry for the delayed answer @maurerle and thanks for pointing out where to fix this. I will check if your suggestion works and file an PR upstream.

ambassador86:
With the gmac definition adjusted to offset 0 the integration checklist is done. Test device: https://map.chemnitz.freifunk.net/#!/de/map/80afca5f9b98

rotanid:
as the upstream fix got merged and backported to 25.12, we only need a backport to 24.10 now, could you also do/provide this upstream?

PR #3642, opened 2025-12-18
[v2023.2.x] modules: update to latest HEAD

herbetom:


neocturne:
Hmm, I feel like the gluon/packages module is now at a point where it should not be updated to the latest main anymore for 2023.2.x - if any commits need to be backported, we can create a release branch (but likely that won't be necessary).

herbetom:
No longer bumping that sounds reasonable. But i would propose going ahead and just creating that release branch and switching over.
Doing it is quite easy and without it i see the potential that it will be bumped accidentally next time.
Also if backports are needed non maintainers would then be able to PR them without prior action. Not that i necessarily expect that to happen but i think it's just good practice as long as v2023.2.x is still "active".

rotanid:
@herbetom at least in the past we only created the branch as soon as it was necessary.
but as the changes look very unproblematic, i'll merge now.

PR #3618, opened 2025-11-19
Fix cellular IPv6 DNS servers + related cleanup

neocturne:
qmi interfaces only create an interface with _6 suffix when the dhcpv6 option is set - otherwise, IPv6 addresses (and presumably DNS servers) are found on the cellular interface itself.

This is currently entirely untested on hardware with cellular support, and the previously checked device where I found that no `cellular_6` interface exists did not have any IPv6 DNS servers assigned anyways.

Fixes #3616

PR #3647, opened 2025-12-29
core: fix ipv6 config for cellular devices

maurerle:
The mark for uplink traffic from cellular to be used as wan for gluon-wan-dnsmasq and wireguard was not set.
Without this ipv6 uplink does not work for cellular devices

fixes the issue mentioned in https://github.com/freifunk-gluon/gluon/pull/3618#issuecomment-3696686621


rotanid:
merging without CI build because the missing CI check is not influenced by the change in this PR

Issue #3655, opened 2026-01-04
Updating to Gluon 2025.1 on the MikroTik SXTsq 5 ac only works via sysupgrade -n, meaning that all settings will be lost


ChristianMiddendorf:
Sysupgrade indicates that no update is possible while retaining the settings. This means that nothing bad will happen when the image is made available for auto-update. However, no update will take place either.
A sysupgrade -n is possible. Afterwards, however, all node settings must be reconfigured.
The device then functions normally, as proof of concept: https://map.freifunk-ingolstadt.de/#!/de/map/cc2de04a5b4c

rotanid:
closing, reason/explanation/workaround see:
https://github.com/freifunk-gluon/gluon/pull/3656#issuecomment-3753358888

PR #3656, opened 2026-01-04
Update v2025.1.rst > auto-update is not possible with the MikroTik SXTsq 5 ac

ChristianMiddendorf:
https://github.com/freifunk-gluon/gluon/issues/3655

### DIFF
--- a/docs/releases/v2025.1.rst
+++ b/docs/releases/v2025.1.rst
@@ -15,6 +15,7 @@ manual steps to adjust the flash layout:
 - Ubiquiti EdgeRouter-X (`upgrade instructions <https://github.com/darkxst/erx-migration>`__)
 - Xiaomi AX3200
   (`upgrade instructions <https://openwrt.org/toh/xiaomi/ax3200#upgrading_from_2305_and_earlier_to_upcoming_2410_or_snapshot>`__)
+- MikroTik SXTsq 5 ac (Only a sysupgrade -n is possible, meaning that the device settings will be lost and must be reset.)
 
 Added hardware support
 ----------------------


maurerle:
Can you please give some more information on the error you got when running sysupgrade?

Eventually this affects other mikrotik devices as well if it is this commit which caused it:
https://github.com/openwrt/openwrt/commit/a60721f2ed3b0152821d2a82e5e7ed115a8ae9cc
 

Eventually, a minor compat version bump is missing for ipq40xx as done for ath79 in:
https://github.com/openwrt/openwrt/commit/d97143fadc65058754497d8757c0d63fc22b503a

ChristianMiddendorf:
> Can you please give some more information on the error you got when running sysupgrade?
> 
> Eventually this affects other mikrotik devices as well if it is this commit which caused it: [openwrt/openwrt@a60721f](https://github.com/openwrt/openwrt/commit/a60721f2ed3b0152821d2a82e5e7ed115a8ae9cc)
> 
> Eventually, a minor compat version bump is missing for ipq40xx as done for ath79 in: [openwrt/openwrt@d97143f](https://github.com/openwrt/openwrt/commit/d97143fadc65058754497d8757c0d63fc22b503a)

It was just a Minor Compact Version difference and it directly gave the hint that just sysupgrade -n is supported, not the normal sysupgrade with keeping the config. In before the device used v2023.2.5-3+139.
As this One is the only Microtic Device in our Community we can not test if others are affected as well.

herbetom:
The Gluon autoupdater ignores any minor compat-version mismatches: https://github.com/freifunk-gluon/packages/commit/b80428166454d184f356e5a14a144911d9ee11aa

(Same goes for Upgrades via the Admin Interface: https://github.com/freifunk-gluon/gluon/commit/75ceb0d963b53ddde3b0ef2039f8a57ef6116eaf)

So from that regard it should just auto update. I kinda would suspect this also isn't a problem in this case but that's just a feeling from what i've read and seen so far.

Hoewer i'm also still unsure where the DEVICE_COMPAT_VERSION increase is actually happening in this case.

@ChristianMiddendorf It would be generally quite helpful if you could copy such error messages into such reports. That would make it much easier to dig stuff up. The more details the easier the triage. :see_no_evil: :sweat_smile:

ChristianMiddendorf:
You're absolutely right. During the update, I was unfortunately only concerned with overcoming all obstacles quickly. The thought of giving you feedback only came later :)
But I'll ask the user if he would be willing to downgrade again and then take a screenshot of the error message during the upgrade.

ChristianMiddendorf:
If you Reverse from v2025.1 to 2023.2.5 there is no warning and now you can easy switch between both Versions via sysupgrade. So it seems to me that stepping up to 2025.1 fixes something that is not reverted if you downgrade.

maurerle:
> It was just a Minor Compact Version difference and it directly gave the hint that just sysupgrade -n is supported

I am quite sure, that this would not have been an issue when updating through the autoupdater instead of sysupgrade and would therefore like to close this, as well as the referenced issue..?
Still thanks for reporting and taking the efforts!

BTW: it is also possible to run `sysupgrade --ignore-minor-compat-version` - which is what the autoupdater does.

ChristianMiddendorf:
O.K., closing…

PR #3141, opened 2024-01-03
Fix readonly ubiquiti airmax

herbetom:
possible fix for #2939

PR #3664, opened 2026-01-11
ath79-generic: re add support for ubiquiti-rocket-m-xm

maurerle:
This device comes in 3 differenct configurations:
* dual-band (M XM)
* 2.4 G only (M2 XM)
* 5 GHz only (M5 XM)

Was removed in #2925
Since then, devices affected by the read-only issue were added in #3141

More information in https://forum.freifunk.net/t/ubiquiti-rocket-m2-m5-xm-besitzer-gesucht/23762/79

Tested by  murxmaster from forum.freifunk.net

## Checklist

  - [X] Web interface
  - [ ] TFTP
  - [ ] Other: <specify>
      
        ubiquiti-rocket-m-xm
- [x] When re-adding a device that was supported by an earlier version of Gluon, a
- Wired network
- LED mapping
  - Power/system LED
          No. Last RSSI-LED has blink sequence
          maurerle: this is fine
  - Radio LEDs
    - [X] Should show activity
  - Switch port LEDs
- Outdoor devices only:
- Cellular devices only:
- Docs:

### DIFF
--- a/targets/ath79-generic
+++ b/targets/ath79-generic
@@ -548,6 +548,8 @@ device('ubiquiti-nanostation-loco-m-xw', 'ubnt_nanostation-loco-m-xw', {
 
 device('ubiquiti-nanostation-m-xw', 'ubnt_nanostation-m-xw')
 
+device('ubiquiti-rocket-m-xm', 'ubnt_rocket-m')
+
 device('ubiquiti-unifi-ac-lite', 'ubnt_unifiac-lite', {
 	factory = false,
 	packages = ATH10K_PACKAGES_QCA9880,
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -159,6 +159,7 @@ ath79-generic
   - NanoBeam M5 (XW)
   - NanoStation Loco M2/M5 (XW)
   - NanoStation M2/M5 (XW)
+  - Rocket M2/M5 (XM)
   - UniFi AC Lite
   - UniFi AC LR
   - UniFi AC Mesh
--- a/package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua
+++ b/package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua
@@ -46,6 +46,7 @@ function M.is_outdoor_device()
 		'ubnt,nanobeam-ac-xc',
 		'ubnt,nanostation-loco-m-xw',
 		'ubnt,nanostation-m-xw',
+		'ubnt,rocket-m',
 		'ubnt,uk-ultra',
 		'ubnt,unifi-ap-outdoor-plus',
 		'ubnt,unifiac-mesh',


maurerle:
This device is not part of v2022.1.x. The latest release supporting these devices is v2021.1.x - according to the diagram:
https://github.com/freifunk-gluon/gluon/wiki/Release-life-cycle#migration-paths-diagram

The greatest jump from there is to v2023.1.x - we would have to backport this to v2023.2.x and v2023.1.x as well, if we want to create a supported update path for these devices.

This currently affects 36 devices:
https://stats.darmstadt.freifunk.net/d/CA5PRFmMz/gluon-census3a-dashboard?orgId=1&from=now-12h&to=now&var-models=Ubiquiti Rocket M XW&var-models=Ubiquiti Rocket M2&var-models=Ubiquiti Rocket M5&var-models=Ubiquiti Rocket M (XM)&refresh=15m&timezone=browser&viewPanel=panel-22

targets/ath79-generic
```diff
@@ -548,6 +548,14 @@ device('ubiquiti-nanostation-loco-m-xw', 'ubnt_nanostation-loco-m-xw', {
 
 device('ubiquiti-nanostation-m-xw', 'ubnt_nanostation-m-xw')
 
+device('ubiquiti-rocket-m-xm', 'ubnt_rocket-m', {
+	manifest_aliases = {
+		'ubiquiti-rocket-m', -- upgrade from OpenWrt 19.07
+		'ubiquiti-rocket-m2', -- upgrade from OpenWrt 19.07
+		'ubiquiti-rocket-m5', -- upgrade from OpenWrt 19.07
```
maurerle: if we decide against supporting this device in v2023.1.x as well, we do not have a valid update path and can drop the manifest aliases

herbetom: Shouldn't they be dropped for main/v2025.1.x either way and just be added to the branches during the backport where it's relevant?

rotanid: @herbetom hm, you might be right. v2021.1.x can only update to v2023.1.x so only v2023.1.x might need the aliases?

rotanid: @maurerle then they should be removed from this PR, right?

rotanid (changes requested):

targets/ath79-generic
maurerle: I just wondered what would possibly break:

- VPN will be site default, bandwidth_limit will be lost: https://github.com/freifunk-gluon/gluon/pull/3113
- The interface roles will be reset to site default: https://github.com/freifunk-gluon/gluon/pull/3406
- hwmode setting and preserve_channels is lost/reset to default: https://github.com/freifunk-gluon/gluon/pull/3406
- a firewall config is not removed anymore since v2023.2: https://github.com/freifunk-gluon/gluon/pull/3040

So this is quite crucial, yes.

The mentioned deprecations were all introduced to v2023.2.x.
So it should be enough to add support to v2023.1.x to create a full update chain.

rotanid (approved):

PR #3662, opened 2026-01-11
mediatek-filogic: add support for Cudy RE3000 v1

mbaumga:
- [X] Must be flashable from vendor firmware
  - [ ] Web interface
  - [ ] TFTP
  - [X] Other: Web Based via Intermediate Firmware. [German Instructions](https://forum.darmstadt.freifunk.net/t/installation-cudy-tr3000/1010) | [English Instructions](https://openwrt.org/toh/cudy/tr3000#oem_easy_installation) (Similar to TR3000 v1)
- [X] Must support upgrade mechanism
  - [X] Must have working sysupgrade
    - [X] Must keep/forget configuration (`sysupgrade [-n]`, `firstboot`)
  - [X] Gluon profile name matches autoupdater image name
        (`lua -e 'print(require("platform_info").get_image_name())'`)
- [X] Reset/WPS/... button must return device into config mode
- [X] Primary MAC address should match address on device label (or packaging)
      (https://gluon.readthedocs.io/en/latest/dev/hardware.html#hardware-support-in-packages)
  - When re-adding a device that was supported by an earlier version of Gluon, a
    factory reset must be performed before checking the primary MAC address, as
    the setting from the old version is not reset otherwise.
- Wired network
  - [X] should support all network ports on the device
  - [X] must have correct port assignment (WAN/LAN)
    - if there are multiple ports but no WAN port:
      - the PoE input should be WAN, all other ports LAN
      - otherwise the first port should be declared as WAN, all other ports LAN
- Wireless network (if applicable)
  - [X] Association with AP must be possible on all radios
  - [X] Association with 802.11s mesh must work on all radios 
  - [X] AP+mesh mode must work in parallel on all radios
- LED mapping
  - Power/system LED
    - [X] Lit while the device is on
    - [X] Should display config mode blink sequence 
          (https://gluon.readthedocs.io/en/latest/features/configmode.html)
  - Radio LEDs
    - [X] Should map to their respective radio (Only one LED for both)
    - [X] Should show activity
  - Switch port LEDs
    - [-] Should map to their respective port (or switch, if only one led present) 
    - [X] Should show link state and activity (Amber/Green LED directly at Ethernet Port)
- Outdoor devices only:
  - [-] Added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`
- Cellular devices only:
  - [-] Added board name to `is_cellular_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`
[...]

PR #2927, opened 2023-07-15
gluon-core: Split check-site.lua and Introduce Lua Module: gluon.validation

lemoer:
This is a preparatory step. The newly introduced validation class is to be used within the REST API later on.

PR #3667, opened 2026-01-15
[v2025.1.x] backport current latest development from main

maurerle:
This are non breaking changes, which I would not like to wait for until v2026.1.x to get merged.
Currently, we are able to fast-forward these without separate backport commits.

Includes:
#2927
#3618 
#3662
#3654

### DIFF
--- a/package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
+++ b/package/gluon-status-page/files/lib/gluon/status-page/view/status-page.html
@@ -50,10 +50,15 @@
 
 		wireless.foreach_radio(uci, function(radio)
 			local channel = iwinfo.nl80211.channel(wireless.find_phy(radio))
+			local radio_name = radio['.name']
+			local suffix = radio_name:match('^radio(%d+)$')
+			local ssid = iwinfo.nl80211.ssid('client' .. suffix) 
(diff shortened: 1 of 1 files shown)


T-X:
Thanks for the effort once more, @maurerle!

Not as a complaint, more as a remark: I'm a big fan of how for instance the Linux kernel does minor releases, how they backport: Only backport fixes (typ. only commits with a specific "Fixes: \<commitid\>" line). This greatly reduces the risk of running into regressions in minor version releases. (and I often find it very frustrating that OpenWrt does not follow that rule that strictly, therefore I regularly run into regressions with OpenWrt on minor release updates... and it's a bit painful to bisect on OpenWrt due to the build+deploy times...)

For Gluon one exception to the "only backport fixes" rule was made for backporting device support, I think. I guess with the reason that it is highly unlikely that adding a new device causes a regression for any of the other ones.

But anyway, as I said, don't want to complain, maybe not really a problem as v2025.1 was just released and a v2025.1.1 will probably follow soon. And at a glance the backported changes don't look that extensive/complicated.

(And I know, with the long major release cycles in Gluon and OpenWrt it's always difficult to patiently wait for a new major release :-). )

PR #2369, opened 2022-01-07
targets/ramips-mt7621: add device Totolink X5000R

RobWei:
Running device:
[http://[2a03:2260:115:100:5e92:5eff:fea3:1510]/cgi-bin/status](http://[2a03:2260:115:100:5e92:5eff:fea3:1510]/cgi-bin/status)
https://meshviewer.freifunk-muensterland.de/map/#!/de/map/5c925ea31510

WIP:
- [ ] must be flashable from vendor firmware
  - [x] ~~webinterface~~ (In refer to https://openwrt.org/toh/totolink/x5000r#installation this should be possible. In my case I was only able to do a "cloud"-update.)
  - [ ] tftp
  - [x] other: "Image Flashing mode: See https://openwrt.org/toh/totolink/x5000r#back_to_stock_firmware"
- [ ] must support upgrade mechanism
  - [ ] must have working sysupgrade
    - [ ] must keep/forget configuration (if applicable)
      *think `sysupgrade [-n]` or `firstboot`*
  - [ ] must have working autoupdate
    *usually means: gluon profile name must match image name*
- [x] reset/wps/phone button must return device into config mode
- [x] primary mac should match address on device label (or packaging) (https://gluon.readthedocs.io/en/latest/dev/hardware.html#notes)
- wired network
  - [x] should support all network ports on the device
  - [x] must have correct port assignment (WAN/LAN)
    - On devices supplied via PoE, there is usually no explicit WAN/LAN labeling on the hardware.
      The PoE input should be the WAN port in this case.
- wifi (if applicable)
  - [x] association with AP must be possible on all radios
  - [x] association with 802.11s mesh must be working on all radios 
  - [x] ap/mesh mode must work in parallel on all radios
- led mapping
  - power/sys led (_critical, because led definitions are setup on firstboot only_)
    - [x] lit while the device is on
    - [x] should display config mode blink sequence 
(https://gluon.readthedocs.io/en/latest/features/configmode.html)
  - ~~radio leds~~
    - [x] ~~should map to their respective radio~~
    - [x] ~~should show activity~~
  - switchport leds
    - [x] should map to their respective port (or switch, if only one led present) 
    - [x] should show link state and activity
- ~~outdoor devices only~~
  - [x] ~~added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`~~

PR #3668, opened 2026-01-15
Adding Totolink X5000R

Adorfer:
  - [x] Web interface
  - [ ] TFTP
  - [ ] Other: <specify>
- [x] "Reset-WPS" button must return device into config mode
- Wired network
- LED mapping
  - Power/system LED
    - [x] Should display config mode blink sequence (1Hz blink works)
  - Radio LEDs (HW has no radio LEDs)
    - [ ] Should show activity
  - Switch port LEDs
<s>- Outdoor devices only:
  - [ ] Added board name to `is_outdoor_device` function in `package/gluon-core/luasrc/usr/lib/lua/gluon/platform.lua`</s>
<s>- Cellular devices only:
  - [ ] Added board name with modem setup function `setup_ncm_qmi` to `package/gluon-core/luasrc/lib/gluon/upgrade/250-cellular`</s>
- Docs:

- Upgrade from Stockfw to gluon-sysupgrade v2025.1 (ffac-image) works, same for openwrt->gluon.

grafik
grafik


maurerle (approved):
For reference, there was already #2369 which tried to add support.
This looks good now.

PR #3692, opened 2026-02-10
[v2023.2.x] modules: update to latest HEAD

herbetom:


herbetom:
seems like there were some backports in the last kernel bump which now need further backports on the openwrt side ... openwrt/openwrt#21967

maurerle:
Is this bump needed to fix a known issue?
Otherwise, I would like to release a last v2023.2.6 after merging #3699 ..?

rotanid:
not knowing anything specific, but generally speaking every kernel bugfix release potentially fixes (security) bugs and the included kernel bump contains half a year of bugfixes...

herbetom:
Yeah, it's nothing specific on my end. But i generally consider kernel bumps something you would want to have.

With the backported patches everything apart from mvebu-cortexa9 seems to build (mediatek-filogic is just an github ci failure and should be a success the next time it runs).

I think the question is kinda whether someone is willing to merge the OpenWrt PR (tracking this downstream seems ofc out of the question). It would just be follow up fixes for the kernel bump. Yeah, it's EoL, but if the bump is okay, then so should follow up fixes.

I can try to look into fixing mvebu-cortexa9 too. But there probably should be an indication into wheter it would be merged. 

But then, i also don't really currently intend on using v2023.2.6 and therefore i'm also fine with keeping it as it is.

maurerle:
> But then, i also don't really currently intend on using v2023.2.6 and therefore i'm also fine with keeping it as it is.

Same for me.

I don't know if it will get merged upstream. If so, I am happy to merge this (after fixing mvebu-cortexa9 as well).
Though I think this will not happen in the next few weeks and still is a lot of work for an old release..

I would vote to just release a v2023.2.6 soon.

maurerle:
As blocktrron found a simpler fix for the upstream openwrt-23.05 build issue, which is available upstream, I'll close this :)

PR #3694, opened 2026-02-10
add gl-mifi and _unifi-6-lr-v2 to targets

cmarxmeier:
gl-mifi target - for LTE additional kernel and openwrt packages needed
unifi-6-lr-v2 - V3 already added

PR #3698, opened 2026-02-11
Add GL-Mifi

cmarxmeier:
- [ X] Must be flashable from vendor firmware
  - [X ] Web interface
  - [ ] TFTP
  - [ ] Other: <specify>
- [ X] Must support upgrade mechanism
  - [ X] Must have working sysupgrade
    - [X ] Must keep/forget configuration (`sysupgrade [-n]`, `firstboot`)
  - [ X] Gluon profile name matches autoupdater image name
- [X ] Reset/WPS/... button must return device into config mode
- [X ] Primary MAC address should match address on device label (or packaging)
- Wired network
  - [X ] should support all network ports on the device
  - [ X] must have correct port assignment (WAN/LAN)
  - [ X] Association with AP must be possible on all radios
  - [ X] Association with 802.11s mesh must work on all radios 
  - [ X] AP+mesh mode must work in parallel on all radios
- LED mapping
  - Power/system LED
    - [ X] Lit while the device is on
    - [X ] Should display config mode blink sequence 
  - Radio LEDs
    - [X ] Should map to their respective radio
    - [X ] Should show activity
  - Switch port LEDs
    - [ X] Should map to their respective port (or switch, if only one led present) 
    - [ X] Should show link state and activity
- Outdoor devices only:
- Cellular devices only:
  - [ X] no LTE modem support in base image, needs additininal kernel and openwrt distro packages to work
  -  [X] LTE performance is terrible slow on this device
- Docs:
  - [X ] Added Device to `docs/user/supported_devices.rst`

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -400,7 +400,7 @@ mediatek-mt7622
 
 * Ubiquiti
 
-  - UniFi 6 LR (v1, v3)
+  - UniFi 6 LR (v1, v2, v3)
 
 mvebu-cortexa53
 ---------------


ambassador86:
Hello @cmarxmeier,
thanks for the contribution. The additional packages can be added with a "packages" directive in the device definition to get the LTE working, even if it is terribly slow.
One example of adjusting the packages can be found here: https://github.com/freifunk-gluon/gluon/blob/main/targets/ipq40xx-generic#L63

cmarxmeier:
Thanks - i'm aware of the package defs opportunity but actually i only have a gl-mifi with broken sim-Slot for testing, so let's keep ist standard: WAN Interface is ethernet . GLK-mifi was to be ordered with different LTE modems for Europe or USA. 
For modem-support see https://openwrt.org/docs/guide-user/network/wan/wwan/ltedongle.
this also affects USB LTE-modem on AR-150.
markings
AR150 with LTE-Modem and USB-HUB (without, modem operation interrupts sequential)
DSC_0010

targets/mediatek-mt7622
```diff
@@ -20,6 +20,10 @@ device('ubiquiti-unifi-6-lr-v1', 'ubnt_unifi-6-lr-v1', {
 	manifest_aliases = {'ubiquiti-unifi-6-lr'}, -- Upgrade from OpenWrt 22.03
 })
 
+device('ubiquiti-unifi-6-lr-v2', 'ubnt_unifi-6-lr-v2', {
+        factory = false,
+})
+
```
blocktrron, line 26: This is not the GL-MiFi, please create a PR per-device as requested here: https://github.com/freifunk-gluon/gluon/pull/3694#issuecomment-3881049161

maurerle:
@cmarxmeier it would be best to create a separate branch for each device you want to add.
Then you can create a pr for this branch.

Do you need help with this?
The best way is to clone and commit to the repository locally and only use the ui to open the PR.

PR #3699, opened 2026-02-21
 [v2023.2.x] batman-adv: avoid double-rtnl_lock ELP metric worker

ecsv:
`batadv_v_elp_get_throughput()` might be called when the RTNL lock is already held. This could be problematic when the work queue item is canceled via `cancel_delayed_work_sync()` in `batadv_v_elp_iface_disable()`. In this case, an `rtnl_lock()` would cause a deadlock.

To avoid this, `rtnl_trylock()` was used in this function to skip the retrieval of the ethtool information in case the RTNL lock was already held.

But for cfg80211 interfaces, `batadv_get_real_netdev()` was called - which also uses `rtnl_lock()`. The approach for `__ethtool_get_link_ksettings()` must also be used instead and the lockless version `__batadv_get_real_netdev()` has to called.

Fixes: c9d482f65196 ("modules: update routing")
Reported-by: @grische
Tested-by: @Nordm4nn
Bug: https://github.com/freifunkMUC/site-ffm/issues/776

### DIFF
--- a/patches/packages/routing/0004-batman-adv-avoid-double-rtnl_lock-ELP-metric-worker.patch
+++ b/patches/packages/routing/0004-batman-adv-avoid-double-rtnl_lock-ELP-metric-worker.patch
@@ -1,0 +1,133 @@
+From: Sven Eckelmann <sven@narfation.org>
+Date: Sat, 21 Feb 2026 09:58:32 +0100
+Subject: batman-adv: avoid double-rtnl_lock ELP metric worker
+
+batadv_v_elp_get_throughput() might be called when the RTNL lock is already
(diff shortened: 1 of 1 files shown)


blocktrron:
@ecsv Is master / v2025 not affected?

ecsv:
@blocktrron for newer branches, please merge the changes from `routing.git` in `modules`. But I couldn't integrate this change (from this PR) in `routing.git` because the branch openwrt-23.05 is closed (because OpenWrt 23.05 is EOL).


See https://github.com/freifunk-gluon/gluon/pull/3703 for gluon main.

PR #3700, opened 2026-02-21
[v2021.1.x] Backport build-container action

herbetom:
This backports the build-container action to v2021.1.x so that there is an compatible image to build v2021.1.x images with a container from the the registry.

I'm somewhat torn with this PR. I currently "need" such a container so there is that and others could probably also find it helpful down the line as well. But the Release is certainly EoL and there is a good argument to be made that it shouldn't be touched.
Since i'm doing the work anyway i thought hoewer i might as well open the PR and hear what others have to say.

Unfortunately i can't just use one of the existing images because dependencies differ.

Note: The build-container action builds and publishes as expected from my testing. I haven't looked at the other actions of the branch but i also don't think they should be touched this far down the line without a very good reason. This PR therefore doesn't get a green tick from the CI. And neither would the merged commit.

### DIFF
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -20,7 +20,8 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v3
       - name: Log in to the Container registry
-        uses: docker/login-action@65b78e6e13532edd9afa3aa52ac7964289d1a9c1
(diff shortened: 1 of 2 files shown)


rotanid:
@herbetom can you figure out the CI fail? and amend your commits?

herbetom:
I don't think i'll be able to look into this within a somewhat reasonable timeframe. Closing therefore.

PR #3712, opened 2026-02-27
scripts: update-modules: use git worktree for generating changelog

blocktrron:
Use a git worktree with the already cloned modules.

This avoids re-cloning all the modules for every bump, substantially speeding up the process.

### DIFF
--- a/scripts/update-modules.sh
+++ b/scripts/update-modules.sh
@@ -9,65 +9,98 @@ cd "$(dirname "$0")/.." || exit 1
 # shellcheck source=./modules
 source ./modules
 
-git diff --quiet ./modules || {
-	1>&2 echo "Your modules file is dirty, aborting."
-	exit 1
+function switch_to_branch() {
+	local BRANCH="$1"
+	git switch -c "${BRANCH}" || git switch "${BRANCH}"
 }
 
-LOCAL_BRANCH=$(git branch --show-current)
-[[ $LOCAL_BRANCH != *-updates ]] && LOCAL_BRANCH+=-updates
+function get_default_branch() {
+	local REPO_URL="$1"
+	git ls-remote --symref "${REPO_URL}" HEAD | awk '/^ref:/ { sub(/refs\/heads\//, "", $2); print $2 }'
+}
 
-for MODULE in "OPENWRT" ${GLUON_FEEDS}; do
-	if [[ $MODULE != "OPENWRT" ]]; then
-		MODULE=PACKAGES_${MODULE^^}
-	fi
+function get_module_info() {
+	local MODULE_NAME=$1
+	local MODULE_DATA=$2
 
-	_REMOTE_URL=${MODULE}_REPO
-	_REMOTE_BRANCH=${MODULE}_BRANCH
-	_LOCAL_HEAD=${MODULE}_COMMIT
+	_DATA_KEY=${MODULE_NAME^^}_${MODULE_DATA}
+	echo "${!_DATA_KEY}"
+}
 
-	REMOTE_URL="${!_REMOTE_URL}"
-	REMOTE_BRANCH="${!_REMOTE_BRANCH}"
-	LOCAL_HEAD="${!_LOCAL_HEAD}"
-
-	# get default branch name if none is set
-	[ -z "${REMOTE_BRANCH}" ] && {
-		REMOTE_BRANCH=$(git ls-remote --symref "${REMOTE_URL}" HEAD | awk '/^ref:/ { sub(/refs\/heads\//, "", $2); print $2 }')
-	}
-
-	# fetch the commit id for the HEAD of the module
-	REMOTE_HEAD=$(git ls-remote --heads "${REMOTE_URL}" "${REMOTE_BRANCH}" | awk '{ print $1 }')
-
-	# skip ahead if the commit id did not change
-	[ "$LOCAL_HEAD" == "$REMOTE_HEAD" ] && continue 1
-
-	# switch to local working branch, if we found changes
-	[ "$(git branch --show-current)" != "${LOCAL_BRANCH}" ] && {
-		git switch -c "${LOCAL_BRANCH}" || git switch "${LOCAL_BRANCH}"
-	}
-
-	CHECKOUT=$(mktemp -d)
-
-	# clone the target branch
-	git clone --bare "${REMOTE_URL}" --branch="${REMOTE_BRANCH}" "${CHECKOUT}"
+function get_commit_message() {
+	local MODULE_NAME="$1"
+	local MODULE_PATH="$2"
+	local LOCAL_HEAD="$3"
+	local REMOTE_HEAD="$4"
 
 	# prepare the commit message
 	# shellcheck disable=SC2001
-	MODULE=$(echo "${MODULE,,}" | sed 's/packages_//')
+	MODULE=$(echo "${MODULE_NAME,,}" | sed 's/packages_//')
 	TITLE="modules: update ${MODULE}"
 	MESSAGE="$(mktemp)"
 	{
 		echo "${TITLE}"
 		printf '\n\n'
-		git -C "${CHECKOUT}" log --oneline --no-decorate --no-merges "${LOCAL_HEAD}..${REMOTE_HEAD}" | cat
+		git -C "${MODULE_PATH}" log --oneline --no-decorate --no-merges "${LOCAL_HEAD}..${REMOTE_HEAD}" | cat
 	} > "$MESSAGE"
+
+	cat "$MESSAGE"
+}
+
+function update_module() {
+	local MODULE_NAME="$1"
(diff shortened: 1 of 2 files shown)


neocturne:
Why do we need a worktree at all? `fetch` doesn't even update the checked out files.

blocktrron:
@neocturne You are right.

I did it with a worktree differently and then adapted to build the log against the FETCH_HEAD reference.

Updated.

scripts/update-modules.sh
```diff
@@ -18,7 +18,9 @@ LOCAL_BRANCH=$(git branch --show-current)
 [[ $LOCAL_BRANCH != *-updates ]] && LOCAL_BRANCH+=-updates
 
 for MODULE in "OPENWRT" ${GLUON_FEEDS}; do
+	MODULE_PATH="${MODULE,,}"
 	if [[ $MODULE != "OPENWRT" ]]; then
+		MODULE_PATH="packages/${MODULE,,}"
```
neocturne: Hmm, `MODULE` is already lowercase for regular modules...
```suggestion
		MODULE_PATH="packages/${MODULE}"
```

I don't really like the inconsistent handling of "OPENWRT" and feeds - it would be much cleaner to refactor the loop body into a helper function that accepts `MODULE` and `MODULE_PATH` as arguments, so the handling of OPENWRT can be pulled out of the loop over `GLUON_FEEDS`.

blocktrron: Done.

blocktrron:
I also added a timestamp suffix to the -updates branch created.

scripts/update-modules.sh
```diff
@@ -1,73 +1,104 @@
 #!/usr/bin/env bash
 # shellcheck enable=check-unassigned-uppercase
 
-set -eo pipefail
+set -xeo pipefail
```
maurerle: Should this really be this verbose..?

maurerle (commented):
Looks good in general - and works very well!
Did not know about `FETCH_HEAD` before.

Can you also adjust the PR title to match the fetch?

scripts/update-modules.sh
blocktrron: No, forgot to remove pre-commit

maurerle (approved):

scripts/update-modules.sh
```diff
@@ -9,65 +9,96 @@ cd "$(dirname "$0")/.." || exit 1
...
-	}
+	[ -z "${REMOTE_BRANCH}" ] && REMOTE_BRANCH=$(get_default_branch "${REMOTE_URL}")
 
 	# fetch the commit id for the HEAD of the module
-	REMOTE_HEAD=$(git ls-remote --heads "${REMOTE_URL}" "${REMOTE_BRANCH}" | awk '{ print $1 }')
+	REMOTE_HEAD="$(git ls-remote --heads "${REMOTE_URL}" "${REMOTE_BRANCH}" | awk '{ print $1 }')"
 
 	# skip ahead if the commit id did not change
-	[ "$LOCAL_HEAD" == "$REMOTE_HEAD" ] && continue 1
+	[ "$LOCAL_HEAD" == "$REMOTE_HEAD" ] && return 0
```
neocturne: `[` uses `=` for string comparison, not `==`.

scripts/update-modules.sh
```diff
@@ -9,65 +9,96 @@ cd "$(dirname "$0")/.." || exit 1
...
+LOCAL_BRANCH=$(git branch --show-current)
+[[ $LOCAL_BRANCH != *-updates-${CURRENT_DATE} ]] && LOCAL_BRANCH+=-updates-${CURRENT_DATE}
 
-	git commit -F "${MESSAGE}"
+update_module "openwrt" "./openwrt"
 
-	# remove the checkout
-	rm -fr "${CHECKOUT}"
+for MODULE in ${GLUON_FEEDS}; do
+	update_module "PACKAGES_${MODULE^^}" "packages/${MODULE,,}"
```
neocturne:
```suggestion
	update_module "PACKAGES_${MODULE}" "packages/${MODULE,,}"
```

scripts/update-modules.sh
```diff
@@ -9,65 +9,96 @@ cd "$(dirname "$0")/.." || exit 1
...
-	LOCAL_HEAD="${!_LOCAL_HEAD}"
+	REMOTE_URL="$(get_module_info "${MODULE_NAME}" "REPO")"
+	REMOTE_BRANCH="$(get_module_info "${MODULE_NAME}" "BRANCH")"
+	LOCAL_HEAD="$(get_module_info "${MODULE_NAME}" "COMMIT")"
 
 	# get default branch name if none is set
-	[ -z "${REMOTE_BRANCH}" ] && {
-		REMOTE_BRANCH=$(git ls-remote --symref "${REMOTE_URL}" HEAD | awk '/^ref:/ { sub(/refs\/heads\//, "", $2); print $2 }')
-	}
+	[ -z "${REMOTE_BRANCH}" ] && REMOTE_BRANCH=$(get_default_branch "${REMOTE_URL}")
```
neocturne: I don't think we need to know the default branch name at all. Something like the following should work:
```sh
if [ "${REMOTE_BRANCH}" ]; then
  REMOTE_REF="refs/heads/${REMOTE_BRANCH}"
else
  REMOTE_REF=HEAD
```
Now we can just use `REMOTE_REF` for both `ls-remote` (now without `--heads`) and `fetch`.

scripts/update-modules.sh
```diff
@@ -9,65 +9,96 @@ cd "$(dirname "$0")/.." || exit 1
...
+
+git diff --quiet ./modules || {
+	1>&2 echo "Your modules file is dirty, aborting."
+	exit 1
+}
+
+# Checkout update branch, create if it does not exist
+CURRENT_DATE=$(date +%Y%m%d)
+LOCAL_BRANCH=$(git branch --show-current)
+[[ $LOCAL_BRANCH != *-updates-${CURRENT_DATE} ]] && LOCAL_BRANCH+=-updates-${CURRENT_DATE}
```
neocturne, line 101: If the goal is to make the branch names unique, it seems like a good idea to use a timestamp with second precision instead of only using the date.

scripts/update-modules.sh
```diff
@@ -9,65 +9,96 @@ cd "$(dirname "$0")/.." || exit 1
...
-	} > "$MESSAGE"
+	# get the commit message
+	COMMIT_MESSAGE="$(get_commit_message "${MODULE_NAME}" "${MODULE_PATH}" "${LOCAL_HEAD}" "${REMOTE_HEAD}")"
 
 	# modify modules file
 	sed -i "s/${LOCAL_HEAD}/${REMOTE_HEAD}/" ./modules
+
+	# commit the changes
 	git add ./modules
+	git commit -F <(echo "${COMMIT_MESSAGE}")
```
neocturne: `add` and `commit` can be combined.

The `echo` is redundant, the following can be used instead:
```suggestion
	git commit -m "${COMMIT_MESSAGE}" -- ./modules
```
or
```suggestion
	git commit -F - -- ./modules <<<"${COMMIT_MESSAGE}"
```
(the version using `-F -` may be preferable for very long commit messages which don't fit on the command line)

neocturne (changes requested):

rotanid:
@blocktrron linter complains about REMOTE_REF, either there's something wrong or it needs the linting annotation to make it silent...

maurerle:
@neocturne I'd like to get this merged. Are there any objections against?

scripts/update-modules.sh
```diff
@@ -9,65 +9,98 @@ cd "$(dirname "$0")/.." || exit 1
...
-	# switch to local working branch, if we found changes
-	[ "$(git branch --show-current)" != "${LOCAL_BRANCH}" ] && {
-		git switch -c "${LOCAL_BRANCH}" || git switch "${LOCAL_BRANCH}"
-	}
+function get_module_info() {
+	local MODULE_NAME=$1
+	local MODULE_DATA=$2
 
-	CHECKOUT=$(mktemp -d)
+	_DATA_KEY=${MODULE_NAME^^}_${MODULE_DATA}
```
neocturne:
```suggestion
	local _DATA_KEY=${MODULE_NAME}_${MODULE_DATA}
```
- Add `local` or drop the extra variable that is only used once
- There are still `^^` and `,,` all over the place - preferably `update_module` would get its two parameters in the required capitalization (so `update_module OPENWRT openwrt` for example), and no piece of the script except for the part that calls `get_module` would ever need to change the case.

neocturne (changes requested):

scripts/update-modules.sh
maurerle: @blocktrron can you please take care of this? :)

maurerle:
@neocturne could you please take a look here? I would like to merge this soonish.

neocturne (approved):

PR #3717, opened 2026-03-04
gluon-core: set global DHCPv6 DUID

blocktrron:
Upstream requires a global DUID option to be set for DHCPv4 and DHCPv6. This option shall be stable for a given device. As Gluon resets the network uci configuration, this id gets lost.

Set the DUID based on the primary MAC-address on every reconfigure.

Fixes: https://github.com/freifunk-gluon/gluon/issues/3707

---

Strictly this is only required when switching to `next`. No impact for the current OpenWrt base as this key is not used anywhere.

### DIFF
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/015-dhcp-duid
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/015-dhcp-duid
@@ -1,0 +1,18 @@
+#!/usr/bin/lua
+
+-- Configure global DUID for DHCPv4/DHCPv6 based on the primary MAC address.
+-- OpenWrt generates the DUID randomly based on a UUID. As we regenerate the configuration
+-- for every "gluon-reconfigure" run, we need to ensure the DUID remains identical across reconfigurations.
+
+-- In the past (OpenWrt 25.12) DUID was based on the interface MAC address, so this was not necessary.
+
+-- Even though DUID is IPv6 specific, the set value from UCI is also used as the client-id for DHCPv4.
+
+local uci = require('simple-uci').cursor()
+local sysconfig = require 'gluon.sysconfig'
+
+local primary_mac = sysconfig.primary_mac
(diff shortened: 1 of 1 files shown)


maurerle:
@blocktrron should we also create the `network.globals.ula_prefix` created in https://github.com/openwrt/openwrt/blob/main/package/base-files/files/etc/uci-defaults/12_network-generate-ula or is this never needed?

blocktrron:
@maurerle this is used for ULA addresses distributed by the router. We don't do this, so we can ignore its existence.

maurerle (approved):
Looks good.

Does not hurt on openwrt-24.10, but is required for openwrt-25.10 as it fixes #3707

neocturne:
Would it not make more sense to use OpenWrt's logic and just preserve the DUID on upgrades instead of introducing our own logic to generate one?

maurerle:
Good idea, then we would have something like `996-migrate-duid` which copies it  (or whole `network.globals`..?) from gluon-old if it exists?

I guess that it is created after the sysupgrade, but immediately gets overwritten/removed by gluon-reconfigure

blocktrron:
> Would it not make more sense to use OpenWrt's logic and just preserve the DUID on upgrades instead of introducing our own logic to generate one?

Thought about this.

 - Reason for the OpenWrt mechanic is to have a per-device, not per-interface identifier
   - We have this with the primary MAC
 - DUID is set with uci-defaults script (which is deleted afterwards)
   - Could refactor and introduce a script to re-apply on upgrade
 - Create migration logic

All in all, i see the implementation of this PR as the simplest one achievable in Gluon independent on what OpenWrt itself does. --> I see no benefit of creating this dependency.

neocturne:
Copying the whole globals section might actually be preferable - that way we would need to know even less about the fields OpenWrt expects.

maurerle:
I would rather like to merge this as is and switch to the whole globals section in the future if there is more to come.
But for now I would keep it simple and only set what we really need @neocturne

PR #3731, opened 2026-04-09
v2025.1.x: various backports

maurerle:
This backports various device support PRs to v2025.1.x:

backports:
#3706 
#3720 
#3729 
#3727


rotanid:
i'm not sure, whether the order of commits is right, maybe it's just displayed confusing me, but the bumps should come before the devices, right?

maurerle:
Yes it is displayed confusing (e.g. oldest commit at the top and newer ones at the bottom) in Github.
Same view as in `git rebase -i`.
In contrast to "newest at the top" (as seen in git log)

The openwrt bumps are before the addition of new devices.

PR #3736, opened 2026-04-10
add Batman V support

### LINKED ISSUE #1726: status-page does not show neighbours with BATMAN_V
### Bug report

Mesh partners are shown.

**What is the problem?**
Mesh-Partners are not shown - neither for mesh-vpn nor other.

**Gluon Version:**
v2018.2.1

**Site Configuration:**
Site: ffmuc
Domain: ffmuc_welt
Link: https://github.com/freifunkMUC/site-ffm/blob/0c84212/domains/ffmuc_welt.conf

**Custom patches:**
None

Some hints which we (hexa- and I) to while debugging this:
- We have the strong feeling that the issue here is that BATMAN_V is not supported yet.
- The "wrong" response came from /cgi-bin/neighbours-batadv which is implemented [here](https://github.com/freifunk-gluon/gluon/blob/v2018.2.1/package/gluon-status-page-mesh-batman-adv/src/neighbours-batadv.c)
- One node where this is failing currently is [here](http://[2001:608:a01:109:12fe:edff:fe57:f6a6]/cgi-bin/status)

grische:
This is a first iteration trying to upstream the Batman V changes we have been using for multiple years.

The original patchset can be found here: 
https://github.com/freifunk-gluon/gluon/compare/v2025.1.x...freifunkMUC:gluon:v2025.1.x-batmanv

Looking for feedback.

Fixes #1726


package/libgluonutil/src/libgluonutil.c
```diff
@@ -337,3 +338,15 @@ struct json_object * gluonutil_load_site_config(void) {
...
+	return throughput >= 54000 ? 255 :
+		throughput < 417 ? 0 :
+		(uint8_t)(
+			(
+				1.42459274279287898080 * log2(throughput) -
+				12.39555493934044793479
+			)
+			* 25.5
+		);
+}
```
T-X: Style: Just use if's and else's here, there's nothing wrong with using three return statements. A comment explaining the rational between these values might be nice?

---

Using a logarithm sounds like a nice idea to confine it to a value.

Would it make sense to use an exponential function instead, to naturally confine the output to < 255? Ideas:

https://www.wolframalpha.com/input?i=-2%5E%28-x%2F10000%2B10%29%2F4.016+%2B+255+from+x+%3D+0+to+54000%2C+from+y+%3D+0+to+255

(The 4.016 was figured out via trial and error to kind of converge to 0 for a throughput of 0. An exact value could probably be calculated. The `/10000` was chosen to get a similar bulge to your logarithmic function.)

Compared to yours: https://www.wolframalpha.com/input?i2d=true&i=%5C%2840%291.4245927427+*+log2%5C%2840%29x%5C%2841%29+-+12.39555493%5C%2841%29+*+25.5+from+x+%3D+0+to+54000

Also, do we need to stick to a uint8_t in Gluon? Or could we increase that for more precision? I know that 0-255 range comes from BATMAN IV, but do we need to stick to that in Gluon?

T-X: Similarly, for inspiration: MLDv2 queries encode a maximum response code into 16 bits with an exponent and a mantissa as follows: https://www.rfc-editor.org/rfc/rfc3810#section-5.1.3

package/gluon-mesh-batman-adv/files/lib/gluon/core/mesh/post-setup.d/40-gluon-mesh-batman-adv-wired-tpo
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+# throughput_override is only relevant for Batman V
+algo=$(batctl routing_algo 2>/dev/null)
+[ "$algo" = "BATMAN_V" ] || exit 0
+
+if [ "$IFNAME" = "vx_mesh_uplink" ] || [ "$IFNAME" = "vx_mesh_other" ] || [ "$IFNAME" = "mesh-vpn" ]; then
+	sleep 3
+	batctl hardif "${IFNAME}" throughput_override 1000mbit
```
T-X: For mesh-vpn I think the override value should be much lower by default. 100 Mbit/s would make more sense to me, looking at this chart for the average internet speed in Germany:

https://en.wikipedia.org/wiki/List_of_countries_by_Internet_connection_speeds

---

(Ideally, this would also be configurable via a site.conf option. And then further configurable/overwritten via the config-mode. But that could be a later optimization, extra PR.

Also ideally and by default vx_mesh_other/vx_mesh_uplink would get the link speed from their underlying interface. Overwriteable via config-mode. But again could be an extra PR.)

T-X: (And the mesh-vpn BATMAN V throughput override should probably also take the site+config-mode mesh_vpn.bandwidth_limit into account eventually.)

package/gluon-mesh-batman-adv/files/lib/gluon/core/mesh/post-setup.d/40-gluon-mesh-batman-adv-wired-tpo
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+# throughput_override is only relevant for Batman V
+algo=$(batctl routing_algo 2>/dev/null)
+[ "$algo" = "BATMAN_V" ] || exit 0
+
+if [ "$IFNAME" = "vx_mesh_uplink" ] || [ "$IFNAME" = "vx_mesh_other" ] || [ "$IFNAME" = "mesh-vpn" ]; then
+	sleep 3
```
T-X: The sleep 3 does not look nice, looks potentially racy. Could we find another option?

package/gluon-mesh-batman-adv/src/respondd-neighbours.c
```diff
@@ -127,14 +149,32 @@ static struct json_object * get_batadv(void) {
...
+	}
+
+	if (strcmp(algoname, "BATMAN_V") == 0) {
+		opts.is_batman_v = true;
+		ret = batadv_genl_query("bat0", BATADV_CMD_GET_NEIGHBORS,
+					parse_orig_neigh_list_netlink_cb, NLM_F_DUMP,
+					&opts.query_opts);
+	} else if (strcmp(algoname, "BATMAN_IV") == 0) {
+		opts.is_batman_v = false;
+		ret = batadv_genl_query("bat0", BATADV_CMD_GET_ORIGINATORS,
```
T-X, line 196: Shouldn't BATADV_CMD_GET_NEIGHBORS work for BATMAN_IV these days, too? Which would simplify this part? Like this?

```
	if (strcmp(algoname, "BATMAN_V") == 0) {
		opts.is_batman_v = true;
	} else if (strcmp(algoname, "BATMAN_IV") == 0) {
		opts.is_batman_v = false;
	} else {
		json_object_put(opts.interfaces);
		return NULL;
	}
	
	ret = batadv_genl_query("bat0", BATADV_CMD_GET_NEIGHBORS,
				parse_orig_neigh_list_netlink_cb, NLM_F_DUMP,
				&opts.query_opts);
```

package/libgluonutil/src/libgluonutil.c
T-X: .oO(Or maybe we should just scrap this pseudo-TQ and change all the uint8_t tq's to uint32_t, to be able to accommodate both an 8bit BATMAN IV TQ or a 32bit BATMAN V throughput value? And then just let UIs figure out what to interpret that, together with a provided algorithm name?)

package/gluon-mesh-batman-adv/files/lib/gluon/core/mesh/post-setup.d/40-gluon-mesh-batman-adv-wired-tpo
blocktrron: In the wires mesh case, this does ignore the lower layer link speed? I'm not using Batman V, but in a scenario where the lower layer has multiple links over a DSA port, a 10 Mbit link should be wiegted less than a 1G link, doesn't it?

package/gluon-mesh-batman-adv/src/respondd-neighbours.c
```diff
@@ -20,6 +21,7 @@
 
 struct neigh_netlink_opts {
 	struct json_object *interfaces;
+	bool is_batman_v;
```
blocktrron: `uint8_t batman_version;`

Define Versions as a enum.

package/gluon-mesh-batman-adv/src/respondd-neighbours.c
```diff
@@ -44,22 +46,30 @@ static struct json_object * ifnames2addrs(struct json_object *interfaces) {
...
 	BATADV_ATTR_ORIG_ADDRESS,
 	BATADV_ATTR_NEIGH_ADDRESS,
 	BATADV_ATTR_TQ,
 	BATADV_ATTR_HARD_IFINDEX,
 	BATADV_ATTR_LAST_SEEN_MSECS,
 };
 
-static int parse_orig_list_netlink_cb(struct nl_msg *msg, void *arg)
+/* Batman V mandatory attrs */
+static const enum batadv_nl_attrs parse_neigh_list_v_mandatory[] = {
```
blocktrron: The `v` just as above does not indicate the version without context. Rather suffix `parse_neigh_list_mandatory_batadv_v`

Same above for iv.

package/gluon-mesh-batman-adv/src/respondd-neighbours.c
```diff
@@ -75,32 +85,44 @@ static int parse_orig_list_netlink_cb(struct nl_msg *msg, void *arg)
...
 
-	orig = nla_data(attrs[BATADV_ATTR_ORIG_ADDRESS]);
-	dest = nla_data(attrs[BATADV_ATTR_NEIGH_ADDRESS]);
-	tq = nla_get_u8(attrs[BATADV_ATTR_TQ]);
 	hardif = nla_get_u32(attrs[BATADV_ATTR_HARD_IFINDEX]);
 	lastseen = nla_get_u32(attrs[BATADV_ATTR_LAST_SEEN_MSECS]);
 
-	if (memcmp(orig, dest, 6) != 0)
-		return NL_OK;
-
```
blocktrron, line 97: Why is this check removed? It might make more sense to implement different callback ufnctions for IV and V to avoid complications in the future.

package/gluon-mesh-batman-adv/src/respondd-neighbours.c
```diff
@@ -127,14 +149,32 @@ static struct json_object * get_batadv(void) {
...
 	if (!opts.interfaces)
 		return NULL;
 
-	ret = batadv_genl_query("bat0", BATADV_CMD_GET_ORIGINATORS,
-				parse_orig_list_netlink_cb, NLM_F_DUMP,
-				&opts.query_opts);
+	if (batadv_genl_get_algoname("bat0", algoname, sizeof(algoname)) < 0) {
+		json_object_put(opts.interfaces);
+		return NULL;
+	}
```
blocktrron: Put this in a separate function, set the version enum field and use this to decide which query typ to use.

package/gluon-mesh-batman-adv/src/respondd-statistics.c
```diff
@@ -34,16 +34,23 @@ struct clients_netlink_opts {
...
 
 
-static const enum batadv_nl_attrs gateways_mandatory[] = {
+static const enum batadv_nl_attrs gateways_iv_mandatory[] = {
 	BATADV_ATTR_ORIG_ADDRESS,
 	BATADV_ATTR_ROUTER,
 	BATADV_ATTR_TQ,
 };
 
+static const enum batadv_nl_attrs gateways_v_mandatory[] = {
```
blocktrron: Same as above for versioning

package/gluon-radv-filterd/src/gluon-radv-filterd.c
```diff
@@ -267,9 +275,7 @@ static void parse_cmdline(int argc, char *argv[]) {
 				threshold = strtoul(optarg, &endptr, 10);
 				if (*endptr != '\0')
 					exit_errmsg("Threshold must be a number: %s", optarg);
-				if (threshold >= LOCAL_TQ)
-					exit_errmsg("Threshold too large: %ld (max is %d)", threshold, LOCAL_TQ);
-				G.hysteresis_thresh = (uint16_t) threshold;
```
blocktrron, line 272: Adapt this check for the new upper limits depending on the version?

blocktrron (commented):
Thoughts so far

package/gluon-mesh-batman-adv/src/respondd-neighbours.c
grische: renamed it

package/gluon-mesh-batman-adv/src/respondd-neighbours.c
grische: I added it as an enum. I hope it's better now.

package/gluon-radv-filterd/src/gluon-radv-filterd.c
grische: I restored the check and added one for Batman V

package/gluon-mesh-batman-adv/src/respondd-statistics.c
grische: renamed it

package/gluon-mesh-batman-adv/src/respondd-neighbours.c
grische: I tried to split it and have the shared code in a separate function.

package/gluon-mesh-batman-adv/files/lib/gluon/core/mesh/post-setup.d/40-gluon-mesh-batman-adv-wired-tpo
grische: This is a very old piece of code. I asked around and the problem in the past was (rough quote)
> that gateways were detected with 1MBit and hence were rejected in favour of WiFi Meshes with a proper throughput (which are of course much worse).
> So the traffic was routed within the local mesh until eventually it did go through the gateway into the internet.

cc @krombel

blocktrron: I understand, but in case you have this topology, it would lead to a suboptimal path selection:

image

I could imagine this is a bigger can of worms as determining lower link state assumes we don't use a bridge with multiple interfaces.

grische: I understand your concern, but in our cases we never have realistically any cable based connection <100Mbps, even 100Mbps is very very rare and almost all are 1000Mbps (with a few 2500Mbps exceptions).

So hardcoding it to 1000Mbps is much reliable than "hoping" that no device detects its (LAN) speed of 1Mbps or 3Mbps.

What do you think about a compromise like this?
``` 
if speed < 100Mbps:
   speed = 1000Mbps
endif
```

blocktrron: @grische the vxlan interface is bound to a interface, so at least attempting to set the correct link rates on bringup should entirely be possible.

(assuming its not in a bridge with multiple interfaces - In which case the entire approach of having a bridge in the first place seems incompatible with the protocol, but this is the can of worms I've been talking about) 

Forcing the link-rate to a fixed value for all interface is just a hack avoiding the situation you've seen with the vxlan tunnel, not a proper fix.

awlx: @T-X as you might have more insights here on what the issue was.

awlx: I wonder if this could already fix the issue:
```
diff --git a/net/batman-adv/bat_v_elp.c b/net/batman-adv/bat_v_elp.c
index fdc2abe9..0b685f7a 100644
--- a/net/batman-adv/bat_v_elp.c
+++ b/net/batman-adv/bat_v_elp.c
@@ -170,6 +170,23 @@ static bool batadv_v_elp_get_throughput(struct batadv_hardif_neigh_node *neigh,
         * ethtool (e.g. an Ethernet adapter)
         */
        ret = __ethtool_get_link_ksettings(hard_iface->net_dev, &link_settings);
+
+       /* Virtual/stacked interfaces (VXLAN, VLAN, bridges, ...) often fail
+        * the ethtool query or report SPEED_UNKNOWN. Try to resolve the
+        * underlying real net device and query that instead.
+        */
+       if (ret != 0 ||
+           link_settings.base.speed == 0 ||
+           link_settings.base.speed == SPEED_UNKNOWN) {
+               real_netdev = __batadv_get_real_netdev(hard_iface->net_dev);
+               if (real_netdev) {
+                       if (real_netdev != hard_iface->net_dev)
+                               ret = __ethtool_get_link_ksettings(real_netdev,
+                                                                  &link_settings);
+                       dev_put(real_netdev);
+               }
+       }
+
        rtnl_unlock();
        if (ret == 0) {
                /* link characteristics might change over time */
```

So we try to check if there is a realdev in the chain to read the speed from. cc @T-X       

From my understanding this is already done for Wifi interfaces but not for all other interfaces because of [line 109](https://github.com/open-mesh-mirror/batman-adv/blob/main/net/batman-adv/bat_v_elp.c#L109) in that code.

T-X: I think the batman-adv approach so far was when in doubt about the correctness of the throughput then choose a conservative value. Otherwise it'd get annoying if a too optimistic node beats a node with a realistic value.

For VXLAN or mesh-vpn over the internet choosing the lower physical device speed might be too optimistic. So I'm not sure if that would be a patch acceptable for upstream.

However for the Gluon use-case I think we can quite likely say that with mesh-on-lan/wan the lower device of the vxlan one does provide a realistic throughput in most scenarios. So that would make me think that it might be better to implement that lower device check in Gluon scripts? And for mesh-vpn I don't think we can get a realistic value from the lower device via ethtool at the moment.

---

I think I would also be fine to start with 1gbit/s on vx_mesh_uplink / vx_mesh_other and 100mbit/s on mesh-vpn in this PR to start with in this PR. 1gbit/s is the typical / most common LAN ethernet link speed right now. And 100mbit/s the average internet uplink speed in Germany right now.

(But if someone feels like working on this then getting/setting the link speed in Gluon from the lower dev for  vx_mesh_uplink / vx_mesh_other would be great, too.)

grische: After a lot of discussions across several chats, let's remove this from the PR for now and try to sort this independently.

blocktrron: @awlx For Gluon WAN / Single is always ensalved in a bridge (necessity due to private-WiFi) thus the proposed kernel patch will probably not cut it.

Another point - This assumes we do not deal with a legacy switch where link-state metrics only describe the link between SoC and switch mac.

ambassador86:
Hello @grische,
thanks for your PR. I built a firmware for our network with your branch.

At first glance it looks good to me. I can switch between our BATMAN IV and BATMAN V domain without problems. In a small mesh cloud with cable and wifi mesh it also behaves correctly. Only the VPN connection falls back to the default of 1Mbit/s.

Thanks for your effort.

grische:
As @neocturne suggested in the Gluon meet, I added commits which drop the fake-TQ and instead use Batman V's throughput everywhere. 

I also started downstream PRs to support throughput there as well:
- https://codeberg.org/FreifunkBremen/yanic/pulls/250
- https://github.com/freifunk/meshviewer/pull/345

package/libbatadv/src/batadv-genl.c
```diff
@@ -361,3 +361,104 @@ int batadv_genl_query(const char *mesh_iface, enum batadv_nl_commands nl_cmd,
...
+/**
+ * batadv_genl_get_algo() - Query the routing algorithm of a mesh as enum
+ * @mesh_iface: name of the batman-adv mesh interface
+ * @algo: pointer to store the algorithm (BATADV_ALGO_BATMAN_IV or
+ *  BATADV_ALGO_BATMAN_V)
+ *
+ * Return: 0 on success or negative error value otherwise
+ */
+__attribute__ ((visibility ("default")))
+int batadv_genl_get_algo(const char *mesh_iface, uint8_t *algo)
```
T-X:
```suggestion
enum batadv_algo batadv_genl_get_algo(const char *mesh_iface)
```

And add a "BATADV_ALGO_BATMAN_ERROR = -1" or "BATADV_ALGO_BATMAN_UNDEF = -1" to the end of the enum definition.

When possible/feasible try to avoid pointer stuff.

package/gluon-status-page-mesh-batman-adv/src/neighbours-batadv.c
```diff
@@ -89,14 +124,29 @@ static json_object *neighbours(void) {
 		},
 	};
 	int ret;
+	uint8_t algo;
```
T-X:
```suggestion
	enum batadv_algo algo;
```

T-X (changes requested):

package/libbatadv/src/batadv-genl.h
```diff
@@ -29,10 +29,16 @@
...
 #include <netlink/genl/ctrl.h>
 #include <stddef.h>
+#include <stdint.h>
 #include <stdbool.h>
 
 #include "batman_adv.h"
 
+enum batadv_algo {
+	BATADV_ALGO_BATMAN_IV,
+	BATADV_ALGO_BATMAN_V,
```
T-X, line 44:
```suggestion
	BATADV_ALGO_BATMAN_V,
	BATADV_ALGO_BATMAN_ERROR = -1,
```

package/gluon-status-page-mesh-batman-adv/src/neighbours-batadv.c
T-X: And in various other places, too. Prefer the more specific enum type.

grische: This conflicts with what blocktrron asked to do
https://github.com/freifunk-gluon/gluon/pull/3736#discussion_r3069287499

package/gluon-mesh-batman-adv/files/lib/gluon/core/mesh/post-setup.d/40-gluon-mesh-batman-adv-wired-tpo
grische: This was removed from the PR as discussed above. Needs a fix in Batman V instead.

package/gluon-mesh-batman-adv/files/lib/gluon/core/mesh/post-setup.d/40-gluon-mesh-batman-adv-wired-tpo
grische: This was removed from the PR as discussed above. Needs a fix in Batman V instead.

package/libbatadv/src/batadv-genl.c
grische: A negative value won't work with the uint8 requested by blocktrron. https://github.com/freifunk-gluon/gluon/pull/3736#discussion_r3069287499

package/gluon-mesh-batman-adv/src/respondd-neighbours.c
grische: My goal was to not touch the Batman IV code as I have no reliable way to test this. 
If you're sure, I can change it.

ambassador86: The neighbor list under BATMAN IV is working, but the TQ information is not part of the output. Only under BATMAN V you get the speed too.

package/gluon-mesh-batman-adv/src/respondd-neighbours.c
T-X: @blocktrron would it make sense to define it as `enum batadv_algo algo;` instead, too? To have better type checking? (though granted, enum type checking seems rather "new" / WiP: https://stackoverflow.com/a/69294052)

uint8_t typically only makes sense to me in packet formats / socket communication.

package/gluon-mesh-batman-adv/src/respondd-neighbours.c
T-X: Ah, good point, missed that! Would be nice to add metric information to BATMAN IV neighbor tables, too, for compatibility. But yeah, for now needs more code here then.

package/gluon-mesh-batman-adv/src/respondd-neighbours.c
T-X: Or clang can also check if any number assignment is outside of the actually used, specific enum type: https://stackoverflow.com/a/49587809

```
$ clang -Wenum-conversion -Wassign-enum /tmp/foo.c -o /tmp/foo
/tmp/foo.c:10:26: warning: integer constant not in range of enumerated type 'enum batadv_algo'
      [-Wassign-enum]
   10 |         enum batadv_algo algo = 42;
      |                                 ^
1 warning generated.
```

```
$ cat /tmp/foo.c
#include <stdio.h>

enum batadv_algo {
	BATADV_ALGO_BATMAN_IV,
	BATADV_ALGO_BATMAN_V,
	BATADV_ALGO_BATMAN_UNDEF = -1,
};

int main() {
	enum batadv_algo algo = 42;
	printf("algo: %i\n", algo);
}
```

package/libbatadv/src/batadv-genl.c
grische: addressed in the latest push

package/libbatadv/src/batadv-genl.h
grische: I introduced an "UNKNOWN" algorithm. I think that should resolve it for now.

package/gluon-status-page-mesh-batman-adv/src/neighbours-batadv.c
grische: addressed in the latest push

grische:
Rebased on top of latest main and updated the series:

- replaced uint8_t with enum batadv_algo. Also has an BATADV_ALGO_UNKNOWN = 0; now
- fixed batadv_genl_get_algo() to return errno (-EINVAL / propagated negative) instead of -1
- use nla_strlcpy instead of custom strncpy + manual NUL 
- use int64 for throughput: both respondd's gateway_throughput and status-page's tp field
- drop the C-side display formatter for status-page's TP and emit raw kbit/s instead, and add a bitrate formatter for it
- LOCAL_THROUGHPUT maxed at 10 * 1000 * 1000 kbit/s = 10 Gbit/s

package/gluon-radv-filterd/src/gluon-radv-filterd.c
```diff
@@ -52,8 +52,10 @@
 // max execution time of a single ebtables call in nanoseconds
 #define EBTABLES_TIMEOUT 500000000 // 500ms
 
-// TQ value assigned to local routers
+// Local-router metric: must dominate any real neighbour value.
+// IV TQ max is 255; V reports throughput in kbit/s.
 #define LOCAL_TQ 512
+#define LOCAL_THROUGHPUT (10 * 1000 * 1000)
```
neocturne: Can we just use something like `UINT32_MAX` for both IV and V?

grische: If you think we don't need a hysteresis threshold for Batman V, I am fine setting it to UINT32_MAX. I can't speak for Batman IV though.

grische: As discussed, dropped the hysteresis check for Batman V and made it Batman IV-only.
Also changed the local TQ for Batman IV+V to UINT32_MAX.

grische:
Rebased on top of latest main and updated the series:

* Dropped the hysteresis check for Batman V (it is Batman IV-only now)
* Changed the local TQ for Batman IV+V to UINT32_MAX instead
* Renamed "TP" in the dashboard to "Throughput" to make it easier for people to understand the meaning

package/gluon-status-page-mesh-batman-adv/luasrc/lib/gluon/status-page/mesh.lua
```diff
@@ -1,12 +1,22 @@
...
+local attrs
+if site.mesh.batman_adv.routing_algo() == 'BATMAN_V' then
+	attrs = {
+		{'tp', 'Throughput', 'bit/s', 'bitrate'},
+	}
+else
+	attrs = {
+		{'tq', 'TQ', ' %'},
+	}
+end
```
neocturne: I would prefer something like the following:
```lua
local algo_attrs = {
	['BATMAN_IV'] = {
		{'tq', 'TQ', ' %'},
	},
	['BATMAN_V'] = {
		{'tp', 'Throughput', 'bit/s', 'bitrate'},
	},
}
local attrs = algo_attrs[site.mesh.batman_adv.routing_algo()] or {}
```

package/gluon-status-page-mesh-batman-adv/luasrc/lib/gluon/status-page/mesh.lua
```diff
@@ -1,12 +1,22 @@
...
 	-- List of mesh-specific attributes, each a tuple of
 	-- 1) the internal identifier (JSON key)
 	-- 2) human-readable key (not translatable yet)
 	-- 3) value suffix (optional)
 	-- 4) formatter name (optional) -- key into status-page.js formats{}
-	attrs = {
-		{'tq', 'TQ', ' %'},
-		{'tp', 'Throughput', 'bit/s', 'bitrate'},
-	},
+	attrs = attrs,
```
neocturne, line 20: Like many LLM-generated commit descriptions, this one is way too verbose and adds detail that is more confusing than helpful. Just the first paragraph would be much more appropriate.

I'm aware that we currently don't have an LLM policy for Gluon, but my personal one is "LLMs are destroying our planet, our society, our sense of truth, our skills, undermine the idea of open source by normalizing code theft without attribution, and their use supports the worst people on earth and their techno-fascists agendas". As such, I would not be very happy to accept submissions that were supported by LLMs into Gluon, even if it's "just" the commit messages (which I'm not sure about - it just says Co-authored-by, without going into detail what exactly the LLM was used for).

package/gluon-status-page-mesh-batman-adv/luasrc/lib/gluon/status-page/mesh.lua
grische: added

maurerle (approved):
I reviewed the changes. Looks good to me.
This finally solves the long missing compatibility of BATMAN_V - without drawbacks.

rotanid:
@grische For BATMAN_V, the -t threshold is in kbit/s with no documented reference value. Might be worth adding an example to the help text so operators have a sense of scale.
e.g. "~50000 -> 50 Mbit/s"
not sure, but maybe even rejecting very small or very low values might be an option.

grische:
@rotanid added the doc sections via 41733d7286b57d6bfffdb8e03f3f3c64125826ba and squashed them into the respective commit.

rotanid:
@grische One nit: the HEAD commit says "use UINT16_MAX as local TQ" but the code assigns UINT32_MAX ?
or maybe later squashing will "solve" this...

grische:
@rotanid these are two separate things. One is increasing the Batman 4 limit to UINT16_MAX (tq is only 16bit) before adding Batman V support. The Batman V support simply then expands this to 32bit as throughput is 32bit. Splitting this out makes the latter commits imho easier to understand.

PR #3738, opened 2026-04-14
mediatek-mt7622: re-add Xiaomi AX3200 / Redmi AX6S

### LINKED ISSUE #3520: Warning: Xiaomi AX3200 Sysupgrade from 23.05 base and earlier to 24.10 fails

kpanic23:
  - [ ] Web interface
  - [ ] TFTP
  - [x] Other: https://openwrt.org/toh/xiaomi/ax3200#installation
- Wired network
- LED mapping
  - Power/system LED
  - Radio LEDs
    - [ ] Should show activity
  - Switch port LEDs
- Outdoor devices only:
- Cellular devices only:
- Docs:

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -409,6 +409,10 @@ mediatek-mt7622
 
   - UniFi 6 LR (v1, v2, v3)
 
+* Xiaomi
+
+  - AX3200 / Redmi AX6S
+
 mvebu-cortexa53
 ---------------
 
--- a/targets/mediatek-mt7622
+++ b/targets/mediatek-mt7622
@@ -27,3 +27,11 @@ device('ubiquiti-unifi-6-lr-v2', 'ubnt_unifi-6-lr-v2', {
 device('ubiquiti-unifi-6-lr-v3', 'ubnt_unifi-6-lr-v3', {
 	factory = false,
 })
+
+
+-- Xiaomi
+
+device('xiaomi-redmi-router-ax6s', 'xiaomi_redmi-router-ax6s', {
+	sysupgrade_ext = '.itb',
+	factory = '-factory',
+})


kpanic23:
I've added factory image generation, so you can upgrade from older versions directly to Gluon without needing an extra step flashing vanilla OpenWrt.

My amended **upgrade instructions from Gluon <2025.1** are:

- SSH into the router.
- Run `sysupgrade -b /tmp/backup.tgz` to backup the configuration
- On your client, use scp to download the backup file ` scp root@[...]:/tmp/backup.tgz .`
- Copy the factory image to the router's /tmp directory (wget/scp/...)
- On the router, run the following lines:

```
mount -o remount,ro /
mount -o remount,ro /overlay
dd if=/tmp/gluon-*-mediatek-mt7622-xiaomi_redmi-router-ax6s-factory.bin bs=1M count=4 | mtd write - kernel
dd if=/tmp/gluon-*-mediatek-mt7622-xiaomi_redmi-router-ax6s-factory.bin bs=1M skip=4 | mtd -r write - ubi
```

The router should reboot into setup mode.
- SSH into the router: `ssh -oUserKnownHostsFile=/dev/null root@[…]` (Host ID has changed, this ignores the host ID)
- Now copy the backup file back to the router's /tmp directory
- Run `sysupgrade -r /tmp/backup.tgz` to restore the previous configuration.
- Run `uci set system.@system[0].compat_version='2.0';uci commit` to make future upgrades work.
- Run gluon-reconfigure (not sure if needed, but just for keeping it safe…)
- Reboot the router

Done.

maurerle (approved):

rotanid:
there was a segfault in batman-adv compile in github actions, i triggered a re-build

AiyionPrime (approved):
lgtm, once build passes in ci

Issue #3580, opened 2025-09-06
6 GHz and mesh


maurerle:
This is a tracking issue for experience with 6GHz wifi on openwrt/gluon.

The idea is to add 6GHz support after #3563 

First of all:
* 6GHz requires WPA3, so we need to have an OWE interface running.
* Starting an unsecured client interface on the 6ghz radio does simply not come up.
* 6GHz therefore needs OWE without transition mode on the 6ghz wifi
* the private wifi is only available with WPA3
* the mesh channel requires usage of SAE

The question is, which other parts are required for a supported 6GHz mesh radio.

maurerle:
Just for others to find help with similar problems:

After installing `wpa-supplicant-mesh-mbedtls` one must reboot the device, otherwise, the new interface is not set up correctly:

```
Sat Sep  6 02:43:29 2025 daemon.notice netifd: radio1 (14317): Interface  setup failed: WPA_SUPPLICANT_FAILED
Sat Sep  6 02:43:29 2025 daemon.notice netifd: radio1 (14317): Failed to parse json data: unexpected end of data
Sat Sep  6 02:43:29 2025 daemon.notice hostapd: Set new config for phy phy2: /var/run/hostapd-phy2.conf
Sat Sep  6 02:43:29 2025 daemon.notice hostapd: Restart interface for phy phy2
Sat Sep  6 02:43:29 2025 daemon.notice netifd: radio1 (14317): WARNING (wireless_add_process): executable path /usr/sbin/wpa_supplicant does not match process  path (/proc/exe)
Sat Sep  6 02:43:29 2025 daemon.notice netifd: radio1 (14317): Command failed: ubus call network.wireless notify { "command": 2, "device": "radio1", "data": { "pid": 0, "exe": "\/usr\/sbin\/wpa_supplicant", "required": true, "keep": true } } (Invalid argument)
Sat Sep  6 02:43:29 2025 daemon.notice netifd: radio1 (14317): Usage: ubus [<options>] <command> [arguments...]
Sat Sep  6 02:43:29 2025 daemon.notice netifd: radio1 (14317): Options:
Sat Sep  6 02:43:29 2025 daemon.notice netifd: radio1 (14317):  -s <socket>:		Set the unix domain socket to connect to
Sat Sep  6 02:43:29 2025 daemon.notice netifd: radio1 (14317):  -t <timeout>:		Set the timeout (in seconds) for a command to complete
[...]

PR #3743, opened 2026-04-27
docs: user/site: improve wifi docs

herbetom:
Preview: https://gluon--3743.org.readthedocs.build/en/3743/user/site.html

### DIFF
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -104,7 +104,7 @@ regdom \: optional
 wifi24 \: optional
   WLAN configuration for 2.4 GHz devices.
   ``channel`` must be set to a valid wireless channel for your radio.
-  ``bandwidth`` can be specified to set a custom bandwidth per radio band in MHz. Can be 20, 40, 80 or 160.
+  ``channel_width`` can be specified to set a custom channel width per radio band in MHz. Can be 20, 40, 80 or 160.
   ``beacon_interval`` can be specified to set a custom beacon interval in
   time units (TU). A time unit is equivalent to 1024 µs.
   If not set, the default value of 100 TU (=102.4 ms) is used.
--- a/docs/user/site.rst
+++ b/docs/user/site.rst
@@ -184,6 +184,19 @@ wifi5 \: optional
       [...]
     },
 
+.. _user-site-wifi6:
+
+wifi6 \: optional
+  Same as `wifi24` but for the 6 GHz radio.
+
+  ::
+
+    wifi6 = {
+      channel = 33,
+
+      [...]
+    },
+
 next_node \: package
   Configuration of the local node feature of Gluon
 


docs/user/site.rst
```diff
@@ -184,6 +184,19 @@ wifi5 \: optional
       [...]
     },
 
+.. _user-site-wifi6:
+
+wifi6 \: optional
+  Same as `wifi24` but for the 6 GHz radio.
```
maurerle, line 190: Maybe add:

Devices using the 6GHz band require `wireless-encryption-wpa3` and `gluon-mesh-wireless-sae` as WPA3/OWE/SAE is a requirement on that band.

There is currently also lacking a link which explains this better I guess..?

docs/user/site.rst
```diff
@@ -104,7 +104,7 @@ regdom \: optional
 wifi24 \: optional
   WLAN configuration for 2.4 GHz devices.
   ``channel`` must be set to a valid wireless channel for your radio.
-  ``bandwidth`` can be specified to set a custom bandwidth per radio band in MHz. Can be 20, 40, 80 or 160.
+  ``channel_width`` can be specified to set a custom channel width per radio band in MHz. Can be 20, 40, 80 or 160.
```
maurerle, line 107: fixes #3659 - thanks.

T-X: Should channel_width also be specified for a wifi5 section? And I guess in practice 80 or 160 wouldn't work on 2.4GHz (but it should fall back to 40MHz then there?)

And maybe it would also be nice to mention that a specific device would choose a lower channel bandwidth than specified if it doesn't have that capability or if it is limited by DFS?

docs/user/site.rst
herbetom: At least a missing `gluon-mesh-wireless-sae` doesn't seem to throw an error. Neither during build nor in operation (as far as i've seen without looking too closely).
Not sure however whether it would be actually working without it because i don't have two 6 GHz capable APs.

From an operations perspecive i'm also somewhat reluctant of generally enabling sae mesh so maybe we should consider having an option to only do SAE mesh on 6 GHz or the like. But that ofc goes way beyond this small PR which mainly was meant as a drive by contribution of two things i noticed :sweat_smile: :joy:

maurerle: Yes, it does not throw an error in the config, but you should see the mesh interface not coming up and complaining in the logs: `key_mgmt for mesh network in 6 GHz should be SAE`
https://github.com/freifunk-gluon/gluon/issues/3580#issuecomment-3261798201

if the SAE package is available, it is automatically enabled on 6GHz, even if there is no config in the site for it. The package alone does not do anything for other devices:
https://github.com/freifunk-gluon/gluon/blob/73a7d570aec541546ad2cdb600c67ff4cc30edf5/package/gluon-mesh-wireless-sae/luasrc/lib/gluon/upgrade/205-wireless-mesh-sae#L24

docs/user/site.rst
maurerle: Yes, channel_width can also be specified for 5GHz and 6GHz section, but the 5GHz section already states `Same as `wifi24` but for the 5 GHz radio.`, so no need to document twice.

Yes, the fallback is not documented yet. If a higher bandwidth is configured, it should fall back to the largest htmode available:
https://github.com/freifunk-gluon/gluon/blob/73a7d570aec541546ad2cdb600c67ff4cc30edf5/package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless#L100-L110

maurerle:
Should we merge this and create a follow-up issue for the open docs-improvements ideas?

maurerle (approved):

herbetom:
Yeah, maybe for the better (also backport this then so that it'll be part of the v2025.1.1 notes). I'm currently just really strapped for time and don't know when i'll get to it. :see_no_evil:

PR #3758, opened 2026-05-12
docs: suggest distrobox in getting started

lemoer:
If you intend to develop something in gluon, the container can sometimes be a bit too restrictive. E.g. you can not change out of the cwd, editors are not installed, push_pkg.sh does not work, …

For this case, we suggest to use debian inside distrobox. Simply put, distrobox is a fancy wrapper around podman, docker, or lilipod to create and start containers which are highly integrated with the hosts.

For reference: https://github.com/89luca89/distrobox

### DIFF
--- a/docs/user/getting_started.rst
+++ b/docs/user/getting_started.rst
@@ -40,6 +40,27 @@ We also provide a container environment that already tracks all these dependenci
 
   ./scripts/container.sh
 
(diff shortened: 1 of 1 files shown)


rotanid (approved):

lemoer:
Just fixed a minor typo.

PR #3761, opened 2026-05-24

T-X:


T-X:
Btw. a slightly different diff for patches/openwrt/0018-base-files-move-config_generate-to-preinit.patch is created with `make update-patches` on Debian Sid compared to Debian stable/trixie/13. Went for the one from Debian stable, to make the CI happy.

PR #3769, opened 2026-06-10
docs: sort supported devices by supported Wi-Fi standard

T-X:
We are accumulating more and more devices (fortunately). This also makes it harder to find devices of interest though. Therefore adding a new subsection type "WiFi [4-6|-less]".

Sorting things by [Wi-Fi X] -> [Target] -> [Vendor] -> [Model] instead of [Target] -> [Wi-Fi X] -> [Vendor] -> [Model] because I'm assuming that the Wi-Fi standard is of higher interest than the target for the average Gluon user.

While at it also fixing the Totolink X5000R entry which wrongly ended up in the ath79-generic target section instead of the ramips-mt7621 target section.

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -1,38 +1,151 @@
 Supported Devices & Architectures
 =================================
 
-armsr-armv7
------------
+Wi-Fi 6E (IEEE 802.11ax-2021)
+-----------------------------
 
-* Arm SystemReady (EFI) 32-bit
+* Acer
 
-armsr-armv8
------------
+  - Vero-W6M
 
-* Arm SystemReady (EFI) 64-bit
+
+Wi-Fi 6 (IEEE 802.11ax-2021)
+----------------------------
+
+mediatek-filogic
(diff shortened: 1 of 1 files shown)


maurerle:
I think this is fine in general, though I am not sure if the supported_devices should be used like this.
At that point, a filtered toh.openwrt.org for gluon devices would be cool somehow..

But I am not opposed to merging this.

T-X:
> I think this is fine in general, though I am not sure if the supported_devices should be used like this. At that point, a filtered toh.openwrt.org for gluon devices would be cool somehow..
> 
> But I am not opposed to merging this.

Agreed, had been wondering about that, too. Would be nice if readthedocs had such a filter feature, similar to the OpenWrt ToH or also like Meshviewer has (but I think such a feature does not exist in readthedocs yet?). Or if it would make sense and be accepted to filter for Gluon supported devices in the official OpenWrt ToH.

Yeah, this is more a quick patch to allow to more easily find out which modern devices are supported by Gluon, or which are old.

PR #3771, opened 2026-06-11
gluon-core: fix htmode on devices with channel widths < site ch-width

T-X:
On a TP-Link TL-WDR3600 v1 running Gluon v2025.1.1 with an 80MHz channel width in the site config for the 5GHz radio it was observed that the 5GHz radio would not be created anymore since the addition of the channel width setting. The issue could be narrowed down to the UCI htmode setting wrongly being set to HE40 instead of HT40.

```
$ uci get wireless.radio1.htmode
HE40
```

HE40 is not available on this 802.11n/Wi-Fi 4/5GHz radio though:

```
$ iwinfo phy1 htmodelist
HT20 HT40
```

Further investigation showed that in the get_htmode() function it would eventually find and set the "last_mode = HT40". However unless it is not exactly matching the site.conf channel width then it would overwrite it again and finally return the last htmode it checked, even if it was not available on the radio. The order in which get_htmode() checked modes: HE80+80, HE20, VHT20, EHT80, EHT160, HE80, NOHT, HT20, HT40, VHT40, VHT160, EHT20, VHT80+80, EHT320, EHT80+80, VHT80, EHT40, HE160, HE40. Resulting in the last item, HE40 being picked.

Fixing this by only overwriting the last found, chosen mode if the new one is available on this hardware for one thing. For another ensure to select the widest channel width allowed by site config, as the output of iwinfo.nl80211.htmodelist(phy), as can be seen above, is not ensured to be sorted by channel width.

### DIFF
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
@@ -98,11 +98,13 @@ local function get_htmode(radio)
 	local last_mode
 
 	for mode, available in pairs(iwinfo.nl80211.htmodelist(phy)) do
-		if available and mode == htmode then
-			return htmode
-		end
+		if available then
+			if mode == htmode then
+				return htmode
+			end
 
-		last_mode = mode
+			last_mode = mode
+		end
 	end
 
 	-- if preferred htmode is not available
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
@@ -88,28 +88,33 @@ local function get_htmode(radio)
 	local phy = wireless.find_phy(radio)
 	local htmode
 	if iwinfo.nl80211.hwmodelist(phy).ax then
-		htmode = 'HE' .. channel_width
+		htmode = 'HE'
 	elseif iwinfo.nl80211.hwmodelist(phy).ac then
-		htmode = 'VHT' .. channel_width
+		htmode = 'VHT'
 	else
-		htmode = 'HT' .. channel_width
+		htmode = 'HT'
 	end
 
-	local last_mode
+	local last_width
 
 	for mode, available in pairs(iwinfo.nl80211.htmodelist(phy)) do
 		if available then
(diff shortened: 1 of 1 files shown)


T-X:
Btw. with this change the fallback/promise to "if preferred htmode is not available select the last one we got" is not available anymore. But I couldn't quite see when this could legitimately happen anyway. And I didn't want to add even more complexity for such a fallback.

maurerle:
Does your device report
`iwinfo.nl80211.hwmodelist(phy).ax` as true even though it does not support HE?
Thats weird.. Otherwise HE could/should not be selected?

T-X:
> Does your device report `iwinfo.nl80211.hwmodelist(phy).ax` as true even though it does not support HE? Thats weird.. Otherwise HE could/should not be selected?

No, `iwinfo.nl80211.hwmodelist(phy).ax` is false. It does end up in the else-clause and initially sets HT80 [here](https://github.com/freifunk-gluon/gluon/blob/5af124b1712280d38dc39b8d36cc319f33ff1141/package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless#L95).

The HE40 comes from a `(mode = HE40, available = false)` tuple [here](https://github.com/freifunk-gluon/gluon/blob/5af124b1712280d38dc39b8d36cc319f33ff1141/package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless#L100). And then `last_mode = mode` sets that to `HE40` [here](https://github.com/freifunk-gluon/gluon/blob/5af124b1712280d38dc39b8d36cc319f33ff1141/package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless#L105). And `last_mode` is then returned [here](https://github.com/freifunk-gluon/gluon/blob/5af124b1712280d38dc39b8d36cc319f33ff1141/package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless#L110).

Hm, maybe I'll split the patch into two fixes, to make it easier to follow, give me a second.

T-X:
Changelog v2:
* split into two commits (still with the same, final result for 200-wireless)
* adding "Fixes:" lines to both commits

maurerle (approved):
Looks good. Thanks!

PR #3775, opened 2026-06-17
ipq806x-generic: remove broken flag for Extreme Networks WS-AP3935i

grische:
Commit 29623c63c526c18123447460abef154f8c10997c updated OpenWrt with the included fix for LAN1:

8c3c253692 ipq806x: ap3935: disable hibernation on LAN1

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -295,6 +295,10 @@ ipq40xx-mikrotik
 ipq806x-generic
 ---------------
 
+* Extreme Networks
+
+  - WS-AP3935i
+
 * NETGEAR
 
   - R7800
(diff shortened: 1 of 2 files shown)


rotanid (dismissed):
please also add it to the supported device docs when unbreaking and change commit message as suggested

herbetom:
(Conventionally the commit title would be probably more along the lines of `ipq806x-generic: remove broken flag for Extreme Networks WS-AP3935i` (especially the first part with the target instead of targets, the latter is whatever)


Great to see an model loose the broken flag :)

grische:
Added as supported device and renamed exactly as suggested

docs/user/supported_devices.rst
```diff
@@ -61,6 +61,7 @@ ath79-generic
 * Extreme Networks
 
   - WS-AP3805i
+  - WS-AP3935i
```
blocktrron: This is not ath79-generic, move to ipq806x-generic.

blocktrron (dismissed):

docs/user/supported_devices.rst
grische: Sorry for that. Fixed it.

maurerle (approved):

PR #3776, opened 2026-06-17
[v2025.1.x] ipq806x-generic: remove broken flag for Extreme Networks WS-AP3935i

grische:
Commit 0d7f2b1ec8b04a07dac7eb40a93a837f02ca0018 updated OpenWrt with the included fix for LAN1:

8c3c253692 ipq806x: ap3935: disable hibernation on LAN1

### DIFF
--- a/docs/user/supported_devices.rst
+++ b/docs/user/supported_devices.rst
@@ -295,6 +295,10 @@ ipq40xx-mikrotik
 ipq806x-generic
 ---------------
 
+* Extreme Networks
+
+  - WS-AP3935i
+
 * NETGEAR
 
   - R7800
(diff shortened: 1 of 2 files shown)


rotanid (dismissed):
please also add it to the supported device docs when unbreaking

herbetom:
Ideally this should also be a cherry-picked backport with the -x flag once #3775 is merged. That way it's later a little bit easier to trace across branches.

(The Bump PRs itself are somewhat of a special case that's why that didn't happen there (they're not always the same, it's arguably a little bit saver to not fuck up when using automation, backporting all main branch modules bumps to the release branch creates noice without much benefit, ...)

grische:
The commit explanation body of this commit is different as they refer to different commits, that's why I re-created it.

herbetom:
Another Reason for going the classic backport route (and ideally only opening the backport once the primary PR has happened) is that it wastes less CI cycles, is less effort for all sides if changes are needed and doesn't split focus.

If you want to change the reference you can still do that during cherry picking. :sweat_smile: 

This is effectively blocked until #3775 is merged anyway. :see_no_evil:

PR #3778, opened 2026-06-20
contrib: CI config mode - live deployment

maurerle:
This adds a script to create a setup mode preview on Github Pages which can be referred to and should always provide the latest state of the configuration.

This is especially useful when wanting to show what Gluon is about and how the configuration can be done - without having an actual device in config mode at hand.

~~see: https://ffac.github.io/gluon/wizard.html via github pages.~~
https://github.com/ffac/gluon-setup-mode-preview
https://ffac.github.io/gluon-setup-mode-preview/


maurerle:
Closed in favor of a separate repository:
- https://ffac.github.io/gluon-setup-mode-preview/

Source: https://github.com/ffac/gluon-setup-mode-preview

PR #3792, opened 2026-07-29
[v2025.1.x] docs: releases: add v2025.1.2 release notes

herbetom:
Rendered Release Notes: https://gluon--3792.org.readthedocs.build/en/3792/releases/v2025.1.2.html

Diff since last Release: https://github.com/freifunk-gluon/gluon/compare/v2025.1.1...v2025.1.x

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ the future development of Gluon.
 
 Please refrain from using the `main` branch for anything else but development purposes!
 Use the most recent release instead. You can list all releases by running `git tag`
-and switch to one by running `git checkout v2025.1.1 && make update`.
+and switch to one by running `git checkout v2025.1.2 && make update`.
 
 If you're using the autoupdater, do not autoupdate nodes with anything but releases.
 If you upgrade using random main commits the nodes *might break* eventually.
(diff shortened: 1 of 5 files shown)


maurerle:
I would like to add something like "bump openwrt base to 24.10.8" as well, which is the most significant change and intention to update to v2025.1.2?

But i think we never really did this before either..?

T-X:
#3771 wasn't backported yet, right? There was some automatization for backports, right?

herbetom:
Yes, there is an automation. It's currently not working. I've opened a manual backport: #3797

Regarding mentioning the openwrt bump: I thought about this too but haven't really come up with anything good. Usually openwrt and gluon releases don't align and i really don't want  to give anyone the impression that would or should be the case.
I mean we still might bump it beyond 24.10.8 given that 8 commits have landed in the openwrt-24.10 branch since: https://github.com/openwrt/openwrt/compare/5204715d14207efa956553d9d7079f6fa1aaea40...openwrt-24.10

neocturne (approved):

maurerle (approved):
Sounds good! 😊

PR #3799, opened 2026-08-06
gluon-core: add dnsmasq meta-package for deselecting dnsmasq-full

blocktrron:
Gluon switched to dnsmasq-full back when the ath79-tiny target was dropped.

Some years later we see flash getting constrained on some targets again. dnsmasq-full pulls in crypto libraries for DNSSEC validation which result in ~500kB larger images after squashfs compression.

Add a meta-package which allows configuring the selected dnsmasq package per-target. Adapt the cachesize to disable caching on devices where DNSSEC capability is missing. 

Size-difference:

```
dbauer@r20:~/gluon$ ls -lh output_patch | grep plasma
-rw-r--r-- 1 dbauer dbauer 6.3M Aug  6 01:04 gluon-fnsh-4.1~20260806-plasma-cloud-pa300e-sysupgrade.bin
-rw-r--r-- 1 dbauer dbauer 6.3M Aug  6 01:04 gluon-fnsh-4.1~20260806-plasma-cloud-pa300-sysupgrade.bin
dbauer@r20:~/gluon$ ls -lh output/images/sysupgrade/ | grep plasma
-rw-r--r-- 1 dbauer dbauer 6.8M Aug  6 01:19 gluon-fnsh-4.1~20260806-plasma-cloud-pa300e-sysupgrade.bin
-rw-r--r-- 1 dbauer dbauer 6.8M Aug  6 01:19 gluon-fnsh-4.1~20260806-plasma-cloud-pa300-sysupgrade.bin

```


blocktrron:
@maurerle I assume cutting corners for the OpenWrt 25.12 switch is more worthwhile there instead of the package manager

maurerle (approved):
Thats a great idea! I wonder if this helps on next, as I think that apk also did depend on mbedtls and removing crypto meant removing apk..?

Either way, looks good.
I am currently unavailable - if you could validate this for next as well, would be really cool!

blocktrron:
@maurerle this removed libnettle and its dependnencies which do not intersect with mbedtls. image size are w/o dnsmasq-full but with mbedtls.

PR #3801, opened 2026-08-08
gluon-core: adapt sysctl for 64M RAM

blocktrron:
Upstream lower fragmentation buffer limits and min_free_kbytes for devices with 32M of RAM to 384k low / 512k high. Apply the same limits to 64M boards in Gluon.

This should not make a big difference, as fragmentation should only affect L3 traffic directed to the node, which is limited to SSH / status-page in normal operation.

Also lower vm.min_free_kbytes to 2048k. From my testing, this reduces system load in operation massively, as the device less-aggressively reclaims fragmented pages. It also significantly improves stability and throughput on a Netgear R6120.

OpenWrt sets this value to 8192k. The default value by the kernel calculation is 1024k.


blocktrron:
Here are statistics of a specimen which has this patch applied.

System load massively decreased while throughput stays comparable (Upstream / Downstream limits are unchanged in this example)

image

image

maurerle (approved):
Nice Tweaks!

PR #3805, opened 2026-08-12
gluon-core: fix mistyped channel-width comparison

blocktrron:
The channel-width returned by string matching is of Lua type string while the value from site is a number.

Convert the value read from the site to a string. Make it this way, as i do not have a 80+80 device at hand and avoid resolving split channel htmodes, as they are not currently proper supported by OpenWrt.

Fixes: e734119eb808 ("gluon-core: fix potentially too low channel_bandwidth")

---

How this manifests: For me all devices do select a 80 MHz HTMode even as 20 MHz are configured in the site. With this fix applied, the expected behavior of selecting a 20 MHz mode is realized.

### DIFF
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
@@ -103,7 +103,7 @@ local function get_htmode(radio)
 			local width = mode:match("%d+")
 
 			if hwmode .. width == mode and hwmode == htmode then
-				if width == channel_width then
+				if width == tostring(channel_width) then
 					return hwmode .. width
 				end
 


blocktrron:
@T-X I don't get this block in the file:

```lua
			local hwmode = mode:match("%a+")
			local width = mode:match("%d+")

			if hwmode .. width == mode and hwmode == htmode then
```

Isn't the `hwmode .. width == mode` check redundant as they are parsed from `mode` itself? Or is this because of 80+80 modes?

T-X:
> @T-X I don't get this block in the file:
> 
> ```lua
> 			local hwmode = mode:match("%a+")
> 			local width = mode:match("%d+")
> 
> 			if hwmode .. width == mode and hwmode == htmode then
> ```
> 
> Isn't the `hwmode .. width == mode` check redundant as they are parsed from `mode` itself? Or is this because of 80+80 modes?

Yep, for the 80+80 case. I think `width` would be chopped to 80 then. (Not that it would matter much, as if the device supports 80+80 it very likely also has an entry for just 80, I guess.)

T-X:
@blocktrron the commit message confuses me a bit. I don't quite get what this fix has to do with 80+80 mode. Also, why is the issue manifestation not part of the commit message? The commit message of a fix should describe what it fixes and how to reproduce the issue in my opinion.

I'm also wondering if the variable type, width being a string, is the issue why/if the `width > last_width` check a few lines later would work then. Or is that broken, too? (or does Lua auto-convert that to a number there?)

T-X:
> I'm also wondering if the variable type, width being a string, is the issue why/if the `width > last_width` check a few lines later would work then. Or is that broken, too? (or does Lua auto-convert that to a number there?)

Hm, yeah, that part seems broken, too. I think it would unintentionally evaluate to false in `"160" > "80"`, due to doing a lexicographical comparison instead of a number comparison.

---

Edit:

So I think it might make more sense to do this instead?

```
local width = tonumber(mode:match("%d+"))
```

Edit2:

And it would also be great mentioning in the commit message that Lua always evaluates comparing a string to a number to false. Had to look that up and try it (and confusingly, for the ".." operator Lua happily auto-converts a number to a string, it seems...).

Right now the commit message says "Here we have an apple while there we have an orange. Let's turn the orange into an apple." And the commit message does not explain that the issue is comparing apples with oranges in Lua.

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
maurerle, line 110: This suggests, that both must be an integer - and `width` should not be string (as "160"  > "80" is false)

package/gluon-core/luasrc/lib/gluon/upgrade/200-wireless
```diff
@@ -103,7 +103,7 @@ local function get_htmode(radio)
 			local width = mode:match("%d+")
```
maurerle, line 103: This uses the first digits:
so `VHT80+80` would become 80.

So I suggest the following instead:

```suggestion
			local width = tonumber(mode:match("%d+"))
```

maurerle:
Allowed channel widths in the site are also just integer (20,40,80,160) - so I would stick with numbers here and ignore "80+80"?

blocktrron:
@T-X sorry, i was under the impression yesterday 80+80 would evaluate to `80` if converted to a number. Thats why i did the string conversion.

#3806 offers the more straight forward fix, so closing here in favor.

PR #3806, opened 2026-08-13
Fix channel_width selection

drkhsh:
Regression from e734119e, shipped in v2025.1.2.

get_htmode() compares a string (from mode:match) against a number (from
site config), so the exact-match branch is dead and every radio takes the
fallback path. The fallback's max comparison is also string-based, so
"80" > "160".

This hits sites that never set channel_width too, since the default of 20
takes the same path.

Noticed in the Freifunk Innsbruck community while upgrading, currently
shipping a patched release.


maurerle (approved):
Alternative to #3805 - I think that this looks good.

drkhsh:
@maurerle hah, haven't seen that one, too much going on right now :smile:

PR #3809, opened 2026-08-19
gluon-core: extract wireless helpers (radio_roles, is_outdoor)

rotanid:
Extracts two small, behavior-neutral helpers into `gluon.wireless`,
consolidating duplicated expressions across packages.

## `radio_roles(uci, radio)`
Returns `uci:get_list('gluon', 'band_' .. radio.band, 'role')`.
Replaces the identical inline expression at three call sites

## `is_outdoor(uci)`
Returns `uci:get_bool('gluon', 'wireless', 'outdoor')`. The same local
was defined identically in both `200-wireless` and `022-wireless-roles`;
hoisted into `wireless.is_outdoor` and all seven call sites repointed.

## Behavior
Both are pure refactors. byte-equivalent substitutions, no behavior change.
No new dependencies (`wireless` already imported at all affected sites).

## Context
Split out from #3601 (per-radio htmode/txpower overrides), which bundles
several independent concerns and is currently blocked on the radio-identity
decision (indexed `wradioN` vs path-keyed sections). This PR isolates the
uncontroversial, behavior-neutral helper extractions so they can land
independently and shrink the surface of the larger effort.

Builds on the `wireless_band`/roles migration from #3563 and is compatible
with the `channel_width` site option from #3659.

### DIFF
--- a/package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
+++ b/package/gluon-client-bridge/luasrc/lib/gluon/upgrade/320-gluon-client-bridge-wireless
@@ -19,7 +19,7 @@ end)
 local function is_disabled(radio, config)
 	-- now we assign radios to roles
 	local band = radio.band
-	local roles = uci:get_list('gluon', 'band_' .. band, 'role')
+	local roles = wireless.radio_roles(uci, radio)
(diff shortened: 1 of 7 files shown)


maurerle (approved):
Looks good. Thanks!

PR #3815, opened 2026-08-23
contrib: ci: drop batman-only ebtables features from olsr site

mkg20001:
This makes the site config build again


maurerle:
Thanks @mkg20001 
Can you please write a sentence on the rational behind this change? :)

maurerle (approved):

PR #3817, opened 2026-08-24
firewall: switch to nftables

mkg20001:
New PR superseeding https://github.com/freifunk-gluon/gluon/pull/2783

Test script for running on the node: https://github.com/freifunk-gluon/gluon/pull/3817#issuecomment-5554234942


mkg20001:
@neocturne I've created a new PR. Regarding the uci-less way: I would almost prefer wiping /etc/config/firewall on reconfigure and only keeping user rules with a special prefix user_* or similar, as we already have packages adding firewall rules and removing the package keeps the firewall rule arround.

We can't really avoid uci as we need to inject into chains of firewall4 occasionally and those _could_ be injected via extra rules in /etc/nftables.d but since firewall4 manages cleanup of it's own table I'm not sure how stable that would be

mkg20001:
I've somewhat tested this in freifunk lübeck's network, everything important works.

T-X:
Not as a critique but more out of curiosity: Why was a new PR opened instead of force-pushing on the back then still open, old PR? Also, is there a changelog/diff to the old PR? To maybe make it easier for previous reviewers where to pay extra attention, where to continue from where they left off (or is that difficult to say because it has been generated from the scratch via LLMs?).

.oO(I'm also wondering if it would make sense to create a checklist for crowdsourced human runtime testing/verification of the rules. Similar to how we did it for the ar71xx to ath79 migration back then in #2413.)

mkg20001:
The firewall rules were completely rewritten by the LLM as I ended up making a bunch of mistakes in translating them, the rest is mostly from the old version. I made a new PR to have a clean cut.

As for the checklist: The main ones are
- clients can reach the internet
- arp limiter works
- node can mesh correctly with other nodes
- traffic from other nodes goes through the node correctly
- multicast gets blocked

Most of those could be automatically tested. For most cases, the ~~old~~ new testing framework could be used #3818 (in fact off-branch I already have nftables tests), although it doesn't support on-device testing yet, but I think this doesn't matter so much because this isn't very device specific (unless I overlooked some nftables quirks)

T-X:
17 commits is quite a lot. I'm wondering if it could make sense to keep this PR as a summary but then file multiple, smaller PRs? And mark this summary PR as draft? For instance the renaming commit could be nice and easy to start with as an extra PR maybe. (or do other potential reviewers think that's unnecessary?)

Also, several community packages seem to need changes, not quite sure how to proceed with that:

```
$ git grep -l -E "(iptable|ebtable)"
ffbs-collect-debug-info/src/collect-debug-info
ffbs-mesh-vpn-parker/Makefile
ffbs-mesh-vpn-parker/luasrc/usr/share/lua/noderoute.lua
ffbs-parker-nextnode/Makefile
ffbs-parker-nextnode/README.md
ffmuc-ipv6-ra-filter/Makefile
ffmuc-ipv6-ra-filter/files/lib/gluon/ffmuc-ipv6-ra-filter/ipv6-ra-filter
ffmuc-ipv6-ra-filter/src/respondd.c
$ git grep -l -E "\+gluon-ebtables"
ffbs-mesh-vpn-parker/Makefile
ffmuc-ipv6-ra-filter/Makefile
```

Maybe informing ffbs + ffmuc would be enough? (It shouldn't be your/our job here to update community packages in my opinion. But maybe we'd need to add branches to the community packages repository to target different Gluon versions?)

In the gluon-packages repo I don't see anything needing updates right now. brmldproxy just uses tc, which should always happen before nftables/iptables/ebtables.

EDIT: though maybe it's tricky to split the PR, hm.

mkg20001:
Yes, splitting would be tricky as migrating one half and not the other would lead to ebtables + nftables being mixed and this doesn't make things easier for testing. My approach would be to get the testing framework in and then add tests for everything (partly done) so we can easily do before-after comparisons like that and then test around in other freifunk networks or at least with their site configs. As for splitting the rename I'm open to it if requested.

maurerle:
@T-X community-packages has branches: https://github.com/freifunk-gluon/community-packages/branches/all used for v2023.1.x and v2023.2.x - we can create one for v2025.1.x and add the nftables change then to main.
So this would not be a big problem.

docs/dev/uplink.rst
```diff
@@ -36,7 +36,7 @@ sockets created by a process in accordance with the ``LIBPACKETMARK_MARK`` envir
 the packet mark for processes which don't support this themselves. The process must run as root (or at least
 with ``CAP_NET_ADMIN``) for this to work.
 
-Unfortunately there's no nice way to set the packet mark via iptables for outgoing packets. The iptables will
+Unfortunately there's no nice way to set the packet mark via netfilter for outgoing packets. The netfilter hooks will
```
mkg20001, line 39: This may have changed

Tbh I'd leave changing this for another PR

scripts/feature_lib.lua
```diff
@@ -45,6 +57,30 @@ function M.get_packages(files, features)
 
 	end
 
+	function funcs.deprecated(old, new)
```
mkg20001, line 60: Tell me your opinion on this

mkg20001:
[nftables-selftest.sh](https://github.com/user-attachments/files/32264499/nftables-selftest.sh)



Made a script for testing things like the arp-limiter and correct dropping of packets

Edit: Download the script and `scp -O nftables-selftest.sh root@router:/tmp/nftables-selftest.sh && ssh root@router sh /tmp/nftables-selftest.sh` to get the test running

mkg20001:
I've implemented regenerating the firewall completely

mkg20001:
Also commit descriptions should be shorter now

maurerle:
@mkg20001 please rebase onto main and target main branch in this PR.

mkg20001:
Tested again on Lübeck's mesh

PR #3824, opened 2026-09-04
gluon-status-page: allow update via HTTP POST'ed manifest file

T-X:
Allow uploading an autoupdater manifest file via HTTP POST and running the autoupdater on it. For instance via curl like this:

```
  $ curl --data-binary "@/tmp/experimental.manifest" \
        "http://<gluon-node-ipv6-addr>/cgi-bin/firmware-upload"
```

The autoupdater is still performing the usual signature checks etc. on the manifest file.

The advantage of this are the following two: In more complicated, incompatible update scenarios it might useful to be able to perform node updates in a more curated order. Simply using curl and pushing the update to specific nodes should be easier than updating filters on an autoupdater mirror and waiting for a node's micron to trigger.

The second advantage is that this avoids the issue of being unable to update if an autoupdater server becomes inaccessible to signers and/or nodes. Technically the autoupdater server side, including its DNS can currently have a separate governance. And has the power to overrule or disrupt any democratic decisions from the signers, challenging their souvereignity. In the worst case a single person running the autoupdater HTTP server and/or its DNS entries can overrule the signers collective decision. Also due to a misconfiguration or hardware failure an autoupdater server might accidentally become unreachable.

However the primary security promise is supposed to lie in a signed manifest file. And its n-of-m signatures approach to facilitate a democratic security architecture.

---

This PR requires https://github.com/freifunk-gluon/packages/pull/289.

Also posting this as draft for now as I'm wondering if some sort of rate-limiting or similar should be added to avoid a random person potentially blocking legitimate autoupdater processes by continuously sending bogus manifest files.

### DIFF
--- a/package/gluon-status-page/files/lib/gluon/status-page/www/cgi-bin/firmware-upload
+++ b/package/gluon-status-page/files/lib/gluon/status-page/www/cgi-bin/firmware-upload
@@ -1,0 +1,24 @@
+#!/bin/sh
+
+badrequest() {
+	echo 'Status: 400 Bad Request'
+	echo
+	exit 1
+}
+
+case "${REQUEST_METHOD}" in
+	POST)
+		;;
+	*)
+		badrequest
+		;;
+esac
+
+if ! [ -e /usr/sbin/autoupdater ]; then
+	badrequest
+fi
+
+echo ""
+
+logger -t autoupdater "Receiving manifest via HTTP POST, starting autoupdater"
+exec autoupdater - > /dev/null 2>&1


T-X:
It seems uhttpd has its own limits already. In Gluon we set max_requests=32 (or 16 on devices with less than 48MB of RAM) via package/gluon-status-page/luasrc/lib/gluon/upgrade/500-status-page. Compared to the default setting of 3 max_requests=3 in OpenWrt. And max_connections is set to 100 by default in OpenWrt.

I'm wondering if that would be sufficient to avoid an attacker DoS'ing a device or at least denying it from running its micrond based autoupdate. Or should we run a firmware-upload via a separate uhttpd instance with stricter max_requests limits? That way we could also rate-limit it via netfilter separately (later?).

neocturne:
I think this is a bad idea. Unauthenticated users should not have *any* control over users' nodes, not even just to trigger an update that is authenticated otherwise.

In particular, this would allow an attacker to circumvent gradual upgrades with a GLUON_PRIORITY of a few days - the purpose of which is to catch bugs *before* they have been rolled out on all nodes. It would also render the additional transport encryption added by the recently fixed TLS feature meaningless (that would otherwise ensure that not only the signatures match, but also that the correct trusted upgrade server is used - allowing an attacker to continue an upgrade that has been removed from the upgrade server to rescind it). Even without TLS such an attack would otherwise require hijacking the TCP connection to the upgrade server, while is becomes trivial when you can just POST the old manifest.

While these security properties are not documented well, I don't think we should weaken them without a very strong rationale.

T-X:
Ah, you mean if people were (intending) to update the usual, gradual way, triggered via micrond. Then an attacker, any other person could fetch the same, signed manifest from the public mirror and trigger immediate updates. Hm, good point, hadn't thought of that :/.

My initial thought/response was to then only allow POST'ed, immediate updates if GLUON_PRIORITY=0 was set in the manifest. But of course that would be a rather weak safety measure. Usually with a POST'ed mechanism you'd still want to push the firmware gradually, selectively. But to ensure that you'd rely on keeping the manifest you're using a secret...

One could add a (list of) NODEID= attribute(s) to the manifest to select which node should only accept the pushed, signed update. But then the whole signing process would become a bit tedious, if you'd have to sign a different manifest file for each batch of nodes (or even for every node) to update?

(Or the other way would be to do it similar to client certificates,  I guess. To only allow any one of the signers to push an update.)

---

For the current TLS approach/addition I'm not sure I like it that much in a Freifunk context. It opens up a different kind of denial attacks, as I tried to describe in the commit message. I would prefer a way which doesn't rely on central servers in the Freifunk context.

T-X:
The other option I had in mind initially for a serverless update approach: Allowing a node to fetch an image from another node. That is saving and republishing the previously used manifest (like via #3823). But also reconstructing and publishing the original sysupgrade image. I have these not quite cleaned up scripts for that so far: https://gist.github.com/T-X/e33dc2189d6ff0a92bf771d1fdfbb01b (but I'm wondering if using dmsetup would be a nicer way to stitch the sysupgrade file snippets + live kernel + live rootfs together, e.g. to be able to make it seekable, instead of just piping stuff).

But then the remaining part would be how to find/select another node to fetch the image. We could use s.th. like a device model specific multicast address for discovery. But not quite sure how well that would work with gluon-parker or gluon-olsrv2...

T-X:
Another option that would come to my mind: If GLUON_PRIORITY is larger than 0 and the signatures are valid+sufficient: store the manifest file to flash storage but don't update immediately. Instead consider the stored manifest as an alternative source on the next micrond run(s). That would ensure that GLUON_PRIORITY still worked as before.

maurerle:
@T-X I agree with @neocturne on this and would go with
> Unauthenticated users should not have any control over users' nodes, not even just to trigger an update that is authenticated otherwise.

This would mean that this could be a feature extension as community-package but I don't think that this should be part of the update methodology of gluon core..?

PR #3825, opened 2026-09-05
gluon-mesh-olsrd: switch to v1

mkg20001:
oonf-olsrd (olsrd2) had multiple issues and as of right now is not available in OpenWRT (25.12)¹

This commit changes the default implementation from oonf-olsrd to olsrd v1

Additionally this switches to the new method of finding and starting interfaces

Note: OLSR requires an address on every interface and of each family, therefore this commit adds the IPv4 address (randomly generated from prefix4 + primary mac) and an ip addr replace command that adds or replaces an address on a given interface

¹is being readded, but the issues (missing documentation, spurious crashes, etc) still persist

LLM-Notice:
- Upgrade code was written by hand in 2023, partially rewritten by LLMs, then iterated upon by me
- gluon-olsrd-sync-interfaces was entierly generated by LLMs then iterated upon by me

I'd need feedback for whether the IPv4 approach is correct or whether it's problematic due to conflicts and what alternatives there could be

Also this is a split-out of my current changes on upstream-olsr1 branch ~~and I'd need to test it aswell~~ (tested, works so far)


patches/packages/routing/0007-olsrd-give-the-IPv6-daemon-a-ubus-object-of-its-own.patch
```diff
@@ -0,0 +1,28 @@
+From: Maciej Krüger <mkg20001@gmail.com>
```
mkg20001, line 1: will be upstreamed later

patches/packages/routing/0006-olsrd-add-l3roamd-patch.patch
```diff
@@ -0,0 +1,462 @@
+From: Maciej Krüger <mkg20001@gmail.com>
```
mkg20001, line 1: generic patch for any prefix import is being upstreamed: https://github.com/OLSR/olsrd/pull/140

patches/packages/routing/0007-olsrd-give-the-IPv6-daemon-a-ubus-object-of-its-own.patch
mkg20001: https://github.com/openwrt/packages/pull/30454

mkg20001:
Alterenatives to the "assign random IPv4"
- require a certain sized prefix (/16 or bigger)
- add wizard section for setting ipv4, only enable if set
- wizard, but with a default (generated) value
- olsr plugin that detects conflicts and in case there is a conflict, runs a script that generates a new IP, until one is found that doesn't collied
- leave ipv4 olsrd off upstream and require a community package that sets the IP (for example by getting it from a server)

PR #3826, opened 2026-09-05
gluon-core: fix sysctl customization for 64MB boards

blocktrron:
The original PR contained an outdated fix which was broken.

 - File extension did not match the expectation of the sysctl init-file
 - Quotation marks broke sysctl application

Fix both issues to ensure the application works as expected.

Fixes: a505f767ab86 ("gluon-core: adapt sysctl for 64M RAM (#3801)")

### DIFF
--- a/package/gluon-core/luasrc/lib/gluon/upgrade/550-sysctl
+++ b/package/gluon-core/luasrc/lib/gluon/upgrade/550-sysctl
@@ -3,11 +3,11 @@
 local util = require "gluon.util"
 local unistd = require "posix.unistd"
 
-if util.get_mem_total() >= 64 * 1024 * 1024 then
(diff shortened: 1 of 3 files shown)


package/gluon-core/luasrc/lib/gluon/upgrade/550-sysctl
maurerle:
```suggestion
if util.get_mem_total() >= 64 * 1024 then
```

`lua -e 'local util = require "gluon.util"; print(util.get_mem_total())'`
returns values in kbytes - so we want to run this on devices smaller than 64MB RAM - not on all..?

blocktrron: Good catch, fixed.

maurerle (approved):

PR #3828, opened 2026-09-13
Makefile: prevent command-line variable overrides

maurerle:
When Gluon is invoked from wrapper scripts, sub-makes, or CI environments, variables like GLUON_TARGETS, BOARD and SUBTARGET may already be passed via make command-line arguments or MAKEOVERRIDES.

This issue exists in various community site Makefiles:
https://github.com/ffac/site/blob/4d9f26ea7965ff66e51db0d9919e53de75c2ac8d/Makefile#L147-L151

https://github.com/freifunkMUC/site-ffm/blob/75a08aebd2ed00578ce1526ee7318642e8484c1e/Makefile#L8

Beyond distorting list-targets, overriding GLUON_TARGETS and its derived board mappings causes CheckTarget validation to fail (aborting the build) and results in incomplete autoupdater manifests during make manifest, as images for unlisted targets are omitted from the generated manifest file.

For consistency, it makes sense to ensure that `BOARD` and `SUBTARGET` variable are not overwritten from outside as well.


neocturne:
For `BOARD` and `SUBTARGET` this is not enough - the original value will still be passed into the OpenWrt make call and break things. The same is true for any other variable that may affect OpenWrt that Gluon doesn't care about. I also consider the use of `override` a code smell...

While probably not a short-term solution, I think we should get rid of our Makefile - we aren't really using it like a Makefile anyways, with most targets being PHONY.  Having a Python script as the toplevel entry point to the Gluon build would probably be easier to maintain and avoid many of Make's problems.

Issue #3783, opened 2026-07-02
ipq806x: zero UDP checksums are silently discarded - mesh on lan with VXLAN not working


maurerle:
### Bug report

**What is the problem?**
It seems that the ipq806x GMAC (ipq806x-gmac-dwmac) RX checksum-offload engine silently discards inbound IPv6 UDP packets that carry a zero UDP checksum.

Mesh on LAN on a Aruba-AP325 does not work when using VXLAN for gluon mesh on lan.
Luckily, other targets seem to be unaffected.

A tested workaround is to disable the txcsum=0 setting in:
https://github.com/freifunk-gluon/gluon/blob/edfc71fa986d05d072d3c2c8e5e40bf537c1d19b/package/gluon-core/files/lib/netifd/proto/gluon_wired.sh#L75

I wonder why we rely on disabled udp checksums in general?

Related issue is #1321 - maybe this is the cause?

**What is the expected behaviour?**
Mesh-on-LAN should also work with VXLAN on ipq806x target

**Gluon Version:**
probably any, but confirmed on main/v2025.1.x

**Site Configuration:**
https://github.com/freifunkMUC/site-ffm

maurerle:
A patch which would help in Gluon, but affect all targets is:
https://github.com/grische/gluon/commit/7d8331b0fab881289afa1e10201e78b4b3eda6a8

An upstream linux patch which solves the root problem in Linux to add support for empty checksums in stmmac-dwmac1000 driver would be:
https://github.com/grische/linux/commits/dwmac1000-rx-csum-vxlan/

grische:
This apparently also affects devices with VLAN configs and the above kernel patches also fix those cases.

blocktrron:
@maurerle did you send the linked patches upstream? If so, can you provide a link to the mailing list thread?

maurerle:
I did not have time for this yet. Maybe @grische can take care of upstreaming the patches to linux? :)

grische:
I didn't get around to build and test the net-next kernel with the applied patches for the device. I'll start upstreaming after that.

PR #3829, opened 2026-09-15
ipq806x-generic: add support for aruba-ap32x

maurerle:
Known as AP325/AP324.
Indoor Accesspoint from Aruba with 4×4 802.11ac on both 2.4 and 5GHz.

Can be flashed with a patched APBoot bootloader via the dedicated RJ45 Serial Port.
See: https://github.com/maurerle/ap325-flashtool
German Instructions: https://ffmuc.net/wiki/knb:flash-aruba-ap325

  - [ ] Web interface
  - [ ] TFTP
  - [x] Other: serial (dedicated port without opening) and TFTP
- [x] Reset/WPS/... button must return device into config mode (device resets itself when still pressing during reboot)
- Wired network
  - [x] should support all network ports on the device (issue with VXLAN on the target, see #3783 - unrelated to the device)
- LED mapping
  - Power/system LED
    - [x] Should display config mode blink sequence (same frequency is used while in bootloader, but nothing we can change)
  - Radio LEDs
    - [x] Should show activity
  - Switch port LEDs
- Outdoor devices only:
- Cellular devices only:
- Docs:


rotanid (changes requested):
see linter

PR #3831, opened 2026-09-17
migrate interface order for OpenWrt 25.12 e1000e load order change

maurerle:
OpenWrt 25.12 loads the e1000e driver at boot time (openwrt/openwrt@f22f533,
"kernel: modules: load e1000e driver during pre-init stage"). This moves it
in front of r8169, e1000, 8139cp, e100 and a few other drivers, so on x86
devices combining an e1000e NIC with one of those, eth0 and eth1 get
switched when upgrading from v2025.1.x to v2026.x - the same class of
problem the migration in 019-migrate-interface-order already handles for
the r8169/igc change between OpenWrt 23.05 and 24.10.

That migration is a one-shot guarded by sysconfig.migrated_interface_order,
which every device that ran v2025.1.x already has set, so it cannot cover
the new change. Instead of chaining another one-shot script, make it
generation-aware: keep a list of the known driver load orders (23.05, 24.10,
25.12) and store the index of the order the interface roles were last
aligned to in migrated_interface_order.

- unset (never migrated, coming from v2023.2.x or older) -> compare
  against the 23.05 order, as before
- '' (written by the v2025.1.x version of this script) -> compare against
  the 24.10 order
- otherwise the stored index; unparseable values are treated as current

The 24.10 and 25.12 orders were taken from /proc/modules of a
gluon-ffac x86-64 v2025.1.3 and v2026.0.2 nightly image. Apart from
e1000e, no other Ethernet driver changed its relative position.

See freifunk-gluon#3576


maurerle:
superseded by #3832 which reverts the openwrt change for v2026.1.x once and provides a solution for all future situations

PR #3837, opened 2026-09-21
ipq40xx-generic: add support for Netgear RBK20/RBK40

achterin:
  - [ ] Web interface
  - [ ] TFTP
  - [x] Other: nmrpflash
- Wired network
- LED mapping
  - Power/system LED
- Docs:


maurerle (approved):

docs/user/supported_devices.rst
```diff
@@ -264,7 +264,11 @@ ipq40xx-generic
 
   - EX6100 (v2)
   - EX6150 (v2)
+  - RBR20
+  - RBR40
   - RBR50 (v1)
+  - RBR20
+  - RBR40
```
maurerle: Should only be mentioned once here as RBR, and as RBS right?

maurerle (changes requested):
I think that this should probably split out into different PRs for RBK20 and RBK40?
Do you have all 4 devices and check that everything matches (e.g. MAC)?

For reference:

Netgear's naming scheme:

* RBK = Kit (the retail bundle)
* RBR = Router (the unit with WAN port)
* RBS = Satellite (the mesh extender unit)

So:

* RBK20 = [RBR20](https://openwrt.org/toh/netgear/rbr20) (router) + [RBS20](https://openwrt.org/toh/netgear/rbs20) (satellite)
* RBK40 = RBR40 (router) + RBS40 (satellite)

docs/user/supported_devices.rst
achterin: whoopsi, thats a typo indeed.

achterin:
> I think that this should probably split out into different PRs for RBK20 and RBK40? Do you have all 4 devices and check that everything matches (e.g. MAC)?

yes, i tested this with all 4 device variants.
RBR40: https://map.chemnitz.freifunk.net/#/de/map/a040a07009e6
RBS40: https://map.chemnitz.freifunk.net/#/de/map/b0395697aeda
RBR20: https://map.chemnitz.freifunk.net/#/de/map/78d294a0f457
RBS20: https://map.chemnitz.freifunk.net/#/de/map/78d294a11b0d

@ambassador86 can confirm that i've tested every unit, he was sitting across the table ;)

i will split the two kits into two commits.

> 
> For reference:
> 
> Netgear's naming scheme:
> 
> * RBK = Kit (the retail bundle)
> * RBR = Router (the unit with WAN port)
> * RBS = Satellite (the mesh extender unit)
> 
> So:
> 
> * RBK20 = [RBR20](https://openwrt.org/toh/netgear/rbr20) (router) + [RBS20](https://openwrt.org/toh/netgear/rbs20) (satellite)
> * RBK40 = RBR40 (router) + RBS40 (satellite)

that is correct. apart from the lan/wan port labeling, the satellite units ship with half the flash. also rbk20 uses nor flash and rbk40 uses emmc - all other hardware specs are identical.

ambassador86:
As @achterin stated above, we tested the devices all together.

maurerle (approved):

